PR #10247, opened 2022-08-04
[No QA] Use more CPU cores to speed up unit tests

roryabraham:
### Details
Following [this documentation](https://jestjs.io/docs/next/troubleshooting), detect the number of cpus and pass that to jest.

### Fixed Issues
$ n/a – just a quick experiment to hopefully speed up unit tests.

### Tests
Compare the time it takes for unit tests to complete in this PR vs other PRs. If it's notably faster, we should merge this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,8 +23,12 @@ jobs:
             exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


roryabraham:
This doesn't seem to make a notable difference.

PR #10306, opened 2022-08-08
[No QA] Fix staging gradle build

roryabraham:
### Details
Should fix failing Android deploys ([example](https://github.com/Expensify/App/runs/7730172776?check_suite_focus=true))

- [This PR](https://github.com/Expensify/App/pull/10294) updated the app version on main to 1.1.88-4
- When CP'd to staging, that PR somehow introduced [this code](https://github.com/Expensify/App/blame/staging/android/app/build.gradle#L157-L183) on staging, which should not have happened.

So even though we're not seeing a detailed log, I'm certain that is our problem. We are attempting to use the isNewArchitectureEnabled function when it is not defined on staging

### Fixed Issues
$ n/a

### Tests
Going to live-test this bad-boy

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -154,33 +154,6 @@ android {
         multiDexEnabled rootProject.ext.multiDexEnabled
         versionCode 1001018806
         versionName "1.1.88-6"
-        buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()
-
-        if (isNewArchitectureEnabled()) {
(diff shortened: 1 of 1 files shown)


luacmartins (approved):

roryabraham:
Also we don't need the `CP Staging` label because we're merging this directly into `staging`.

roryabraham:
Deploy triggered [here](https://github.com/Expensify/App/actions/runs/2820193761)

roryabraham:
That didn't work because it wasn't an automated pull request

roryabraham:
Not sure if it will use the latest staging code or not, but it's worth a shot

Issue #10419, opened 2022-08-17
Web - Forgot password - If the username is long, the password input field is broken


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Go to the sign in page of new Expensify
2. Enter your login details but not your password (use long login name)
3. Press Forgot? on the password page

## Expected Result:
The structure of the password input window should not be disturbed

## Actual Result:
If the username is long, the password input field is broken

## Workaround:
Unknown

## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.1.88.13

**Reproducible in staging?:** Yes

**Reproducible in production?:** No

**Email or phone of affected tester (no customers):** any

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/93399543/185015600-553d95cc-543e-490a-a391-80079e2cc165.mp4

**Expensify/Expensify Issue URL:** 

**Issue reported by:** Applause - Internal Team

**Slack conversation:** 

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

yuwenmemon:
Offending PR: https://github.com/Expensify/App/pull/10265
Offending Commit: https://github.com/Expensify/App/commit/a3e8e2e68affb7264e64efb9df032c326dc95907

Issue #9817, opened 2022-07-09
IOS - Sign In - Page size is changing after use wrong 2FA and clicking on Save button


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/docs/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
Precondition: need to use 2FA account
1. Launch the app
2. On Sign in user your username
3. Type your password
4. Type wrong 2FA
5. Tap on Sign in

## Expected Result:
Page size is not changing after use wrong 2FA and clicking on Save button

## Actual Result:
Page size is changing after use wrong 2FA and clicking on Save button

## Workaround:
Unknown

## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.1.82.5

**Reproducible in staging?:** Yes

**Reproducible in production?:** Yes

**Email or phone of affected tester (no customers):** any 2FA accounts

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/93399543/178118382-13089550-7051-41f3-9950-7b689930da91.mp4

**Expensify/Expensify Issue URL:** 

**Issue reported by:** Applause - Internal Team

**Slack conversation:** 

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

mvtglobally:
Issue not reproducible during KI retests. (First week)

laurenreidexpensify:
Doublechecking if I need to add this to Upwork before it's reproducible

laurenreidexpensify:
Holding on this to be reproduced in next week's KI testing

laurenreidexpensify:
Holding on this to be reproduced in this week's KI testing

laurenreidexpensify:
Looks like it was reproducible this week so moving forward

laurenreidexpensify:
https://www.upwork.com/jobs/~014f53fc107128911d

errahsoufiane:
### **Proposal**
https://github.com/Expensify/App/blob/151ec1ad3355cfa162ad490030f19a7134bbc076/src/pages/signin/SignInPageLayout/SignInPageContent.js#L51
[...]

PR #10425, opened 2022-08-17
Fix overflow issue on login screen

yuwenmemon:
@techievivek Please review
cc @aimane-chnaif

### Details
I don't believe we need this conditional here as the SignIn page content will always be narrow.

### Fixed Issues
$ https://github.com/Expensify/App/issues/10419

### Tests
1. Go to the sign in page of new Expensify
2. Enter your login details but not your password (use long login name)
3. Press Forgot? on the password page

Screen Shot 2022-08-17 at 12 16 41 PM

### DIFF
--- a/src/pages/signin/SignInPageLayout/SignInPageContent.js
+++ b/src/pages/signin/SignInPageLayout/SignInPageContent.js
@@ -49,7 +49,7 @@ const SignInPageContent = props => (
         <View style={[styles.flex1, styles.flexRow]}>
             <View style={[
                 styles.flex1,
-                props.isSmallScreenWidth && styles.signInPageNarrowContentContainer,
+                styles.signInPageNarrowContentContainer,
             ]}
             >
                 <SignInPageForm style={[


aimane-chnaif:
@yuwenmemon Your solution LGTM! https://github.com/Expensify/App/pull/10425/files#diff-dd48053a4b4ca8a948af29c248e8fee97c8d962017338228d8bec523842df230R52

**Another solution:**
The issue was similar to this: https://stackoverflow.com/questions/35491851/alignitems-center-prevents-flex-row-text-wrapping-in-react-native
So we can also fix this by updating this line:
https://github.com/Expensify/App/blob/78bb1d5f9239e23ef1b6c21a0223e355c0155b1b/src/pages/signin/SignInPageLayout/SignInPageContent.js#L45
To:
```
            props.isSmallScreenWidth && styles.alignItemsCenter,
```
This fixes both issues: https://github.com/Expensify/App/issues/9817, https://github.com/Expensify/App/issues/10419

I think either solution is fine

techievivek (approved):
Looks good, thanks for the changes.

Screenshot 2022-08-17 at 11 27 43 AM

PR #10436, opened 2022-08-17
[No QA] method to randomly generate 64 bit int as string

srikarparsi:
### Details

New Dot section of design doc: https://docs.google.com/document/d/125xZPDi7v6jmS-GKEcvcpdAwpDRPPGsq_iEOCl9RQFw/edit#bookmark=id.3ifommrba4g8

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/223137

### Tests

    - [ ] iOS / native
    - [ ] Android / native
    - [ ] iOS / Safari
    - [ ] Android / Chrome
    - [ ] MacOS / Chrome
    - [ ] MacOS / Desktop
    - [ ] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`)

<details>

</summary>

    - [ ] iOS / native
    - [ ] Android / native
    - [ ] iOS / Safari
    - [ ] Android / Chrome
    - [ ] MacOS / Chrome
    - [ ] MacOS / Desktop

</details>

### QA Steps


src/CONST.js
```diff
@@ -790,6 +790,9 @@ const CONST = {
         INVITE: 'invite',
         LEAVE_ROOM: 'leaveRoom',
     },
+
+    MAX_64BIT_TOP_HALF: 9223372036,
```
flodnv, line 794: Isn't 9223372036 the bottom half? 😕 Or is this name confusing (me)?

flodnv (commented):
Can we add unit tests to this? 🤔

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,22 @@
+import CONST from "../CONST";
+
+function rand64() {
```
arosiclair, line 3: Let's add a jsdoc here that explains that this generates a positive 64 bit number and returns it as a `string` and that this is used for generating transaction IDs. Otherwise looks good!

arosiclair (dismissed):

src/libs/NumberUtils.js
flodnv: s/transaction IDs/IDs

(we are about to reuse this for reportActionIDs)

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,26 @@
...
+ * @returns {string} randomly generated 64 bit string
+ */
+function rand64() {
+    // Max 64-bit signed:
+    // 9,223,372,036,854,775,807
+    // The top part of the max 64-bit number *+1* because we're flooring it
+    let top = Math.floor(Math.random() * CONST.MAX_64BIT_TOP_HALF + 1);
+  
+    // The bottom part of the max 64-bit number *+1* for the same reason
+    let rest = Math.floor(Math.random() * CONST.MAX_64BIT_BOTTOM_HALF + 1);
```
stitesExpensify, line 14: Let's either make this into a ternary, or an else below so that we don't have to generate a number twice for any number that doesn't start with the max first half

src/CONST.js
flodnv: Heh, we must think of it differently, I see it as the bottom half (the bottom of the number) 😄 Left and right sounds good to me

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,26 @@
...
+  
+    // If the top is any number but the highest one, we can actually have any value for the rest
+    if (top != CONST.​​MAX_64BIT_TOP_HALF) {
+      rest = Math.floor(Math.random() * 1000000000);
```
stitesExpensify, line 18: @flodnv @luacmartins you can double check my snake challenge knowledge, but if we want to skip the first 1000000 for reportIDs (or more for reportActionIDs) we could:

1. Change the method to rand64(minimum)
2. Update this to `Math.floor(Math.random() * (1000000000 - minimum)) + minimum`
3. Update the other bottomHalf to `Math.floor(Math.random() * (CONST.MAX_64BIT_BOTTOM_HALF - minimum) + 1) + minimum`;

stitesExpensify: And then when getting an ID for reportIDs just pass `100000000`, transactions will pass its minimum based on our current max in the DB, etc

src/CONST.js
luacmartins: Yea, we can change it to left and right 👍

src/libs/NumberUtils.js
luacmartins: Agree with JSDocs and yes, the usage will be more than just transactionIDs

src/libs/NumberUtils.js
srikarparsi: Would this way work? Let's say the left half is 8223372036, if minimum is 1000000, then you wouldn't be able to reach 8223372036 099999999. 

Also, if min is 1000000000, then rest would always be 0 which should not be the case I think.

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,26 @@
+import CONST from "../CONST";
+
+/**
+ * Generates a random positive 64 bit integer by randomly generating the left half and right half and concatenating them
```
luacmartins, line 4:
```suggestion
 * Generates a random positive 64 bit numeric string by randomly generating the left half and right half and concatenating them. Used to generate client-side ids.
```

src/libs/NumberUtils.js
luacmartins: I actually think we don't need to worry about using a minimum.  We are not doing it for transactionIDs and we have a lot more transactions in the DB.

Since the 64bit distribution is so sparse, the 85m reports we have make almost no difference in increasing the chances of a collision.

stitesExpensify: > ? Let's say the left half is 8223372036, if minimum is 1000000, then you wouldn't be able to reach 8223372036 099999999 for example because rest is always greater than minimum which is 1000000 in this case.

Thinking through this:
1. We have left = 8223372036
2. We generate a number between 0 and 999,000,000 (so max would be 998,999,999)
3. We add 1,000,000 making the max right side 999,999,999, and min right side 1,000,000

> Also, if min is 1000000000, then rest would always be 0 which should not be the case I think.

Good point about the minimum being higher than the right side (it is for reportActionID), we will need conditional logic for that.

stitesExpensify: @luacmartins we have over a billion transactionIDs. I know that's still small compared to 9 quintillion, but it's not nothing /shrug

srikarparsi: I'm trying to come up with a good way but I can't yet, here's my thoughts. The idea is to split the string into the left half, the left of the minimum and then the rest of the right half. Then we could use Brandon's logic of subtracting the min, multiplying it by the random number and adding the min back to the right half. But the problem with this is if the minimum is so big it goes into the right half, then we have the same initial problem since we can only generate 32 bit random numbers. To fix this we could add a max minimum which is equal to CONST.MAX_64BIT_BOTTOM_HALF (right half). I thought this was hacky but I think we have to have this condition regardless even if we do it the other way. The other way I believe I read in the thread is to generate random numbers until one is greater than the min. But to do this, the min has to be below the bottom half since if it goes into the top half, then we are comparing a >32 bit number with a >32 bit number which I don't think is natively possible. I'm re-reading this and I realized it's kind of confusing so let me know if I can elaborate on any portion.

Basically, my proposal is to add a constraint that the min has to be less than the CONST.MAX_64BIT_BOTTOM_HALF and then split the right string into two parts and if the left part is all 0's then do (rand * (right half magnitude - min)) + min and concatenate all three portions.

stitesExpensify: Can't we just do:

```
if (minimum > MAX_64BIT_BOTTOM_HALF) {
   leftMinimum = minimum.toString().substring(0,minimum.toString().length - 9);
   // code I had up there but using leftMinimum instead of minimum
} else {
   // code I had up there
}
```

luacmartins: Well 1 billion is only 0.0000000108% of the whole available range. I think the chances of having a collision with or without setting a minimum would not change much 🤷 

We decided against it in this thread - https://docs.google.com/document/d/125xZPDi7v6jmS-GKEcvcpdAwpDRPPGsq_iEOCl9RQFw/edit?disco=AAAAdh4GcbM

stitesExpensify: Okay let's just not bother then. Sorry for the confusion @srikarparsi

stitesExpensify (dismissed):

stitesExpensify (dismissed):
LGTM! We can ignore the linter error since we know it's not getting used yet

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Generates a random positive 64 bit numeric string by randomly generating the left half and right half and concatenating them. Used to generate client-side ids.
+ * @returns {string} randomly generated 64 bit string
+ */
+function rand64() {
+    // Max 64-bit signed:
+    // 9,223,372,036,854,775,807
+    // The left part of the max 64-bit number *+1* because we're flooring it
+    const left = Math.floor(Math.random() * CONST.MAX_64BIT_LEFT_HALF + 1);
```
luacmartins, line 11:
```suggestion
    const left = Math.floor(Math.random() * (CONST.MAX_64BIT_LEFT_HALF + 1));
```

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,27 @@
...
+
+    // The right part of the max 64-bit number *+1* for the same reason
+    let right;
+
+    // If the top is any number but the highest one, we can actually have any value for the rest
+    if (left !== CONST.MAX_64BIT_LEFT_HALF) {
+        right = Math.floor(Math.random() * 1000000000);
+    } else {
+        right = Math.floor(Math.random() * CONST.MAX_64BIT_RIGHT_HALF + 1);
```
luacmartins, line 20:
```suggestion
        right = Math.floor(Math.random() * (CONST.MAX_64BIT_RIGHT_HALF + 1));
```

src/libs/NumberUtils.js
luacmartins: We should +1 before multiplying

src/libs/NumberUtils.js
```diff
@@ -0,0 +1,27 @@
+import CONST from '../CONST';
+
+/**
+ * Generates a random positive 64 bit numeric string by randomly generating the left half and right half and concatenating them. Used to generate client-side ids.
+ * @returns {string} randomly generated 64 bit string
+ */
+function rand64() {
```
luacmartins, line 7: We should suppress the lint error for now `/* eslint-disable no-unused-vars */`

luacmartins (changes requested):

src/libs/NumberUtils.js
stitesExpensify: We should actually leave it how it is, or move it the other way right?  Otherwise we can get 0

`Math.floor((Math.random() * CONST.MAX_64BIT_LEFT_HALF) + 1);`

luacmartins: Ah that's right! I guess I failed the snake challenge 😅 FWIW Tyler's [original solution](https://expensify.slack.com/archives/C02HWMSMZEC/p1659726344024529?thread_ts=1659032751.555699&cid=C02HWMSMZEC) had the same mistake

luacmartins: I edited the suggestion above to `Math.floor((Math.random() * CONST.MAX_64BIT_LEFT_HALF) + 1);`

luacmartins:
@flodnv I tried to think of some unit tests for this, but couldn't come up with anything useful. Did you have anything in mind?

flodnv:
We could at least verify that the generated number is in the range we expect it to be (and is a string)

luacmartins:
Ooh I thought it would be tricky to test the range, but string comparison seems to work!

Screen Shot 2022-08-19 at 9 44 18 AM

src/libs/NumberUtils.js
srikarparsi: do we need the parentheses? I think it does the same operation regardless but if it helps for clarity I can commit the suggestion.

luacmartins: I see this warning:
Screen Shot 2022-08-19 at 9 48 45 AM

srikarparsi:
I think the lowest number we can get with this function is 1,000,000,000. I don't think it matters that much since 64 bits is very very big but I don't know if this was discussed so I wanted to bring it up.

src/libs/NumberUtils.js
srikarparsi: I was thinking about this but whoever uses the function later would have to remember to take out the eslint that we would add. I can add it though if you think it's better to do it this way.

luacmartins: Yea, I think we should add it, otherwise tests would break for others.

luacmartins: Created the issue here. I'll update the link once we merge this PR.

luacmartins:
Good point! I think we should add first and then multiply on the left part, so the range is 0-MAX_64BIT_LEFT_HALF. For the right part we could multiply and then add so the range is 1-MAX_64BIT_RIGHT_HALF

luacmartins:
@srikarparsi can we add unit tests as Flo suggested?

JmillsExpensify:
@srikarparsi I wanted to check in on this PR? We've had some great process thus far, so let's do everything we can to keep that same momentum this week. When do you think it'll be ready for re-review?

srikarparsi:
I'll add unit tests, clean up the warnings and change the bounds by the end of today :)

src/libs/NumberUtils.js
tylerkaraszewski: A result of 0 should be valid here. Particularly for either `left` or `right`, so long as both aren't 0.

```
const left = Math.floor(Math.random() * (CONST.MAX_64BIT_LEFT_HALF + 1));
```

Is correct, it will give a result between 0 and `MAX_64BIT_LEFT_HALF` (9223372036) inclusive of both those numbers. The same pattern is true below for `right`.

If we're concerned about getting `0`, we can check `if (left === 0 and right === 0)` and if so, change `right` to `1`. However, the chance of this happening is orders of magnitude less than this number colliding with another ID in the DB, so we should be able to ignore it, really.

tylerkaraszewski (changes requested):

luacmartins:
Thank you @tylerkaraszewski! I agree, let's make those changes @srikarparsi!

srikarparsi:
yup @luacmartins! I fixed the lint errors and the bounds so that it allows for 0 in the right or left half but had a question about the test. Is this the right location to add a new file that tests NumberUtils.js https://github.com/Expensify/App/tree/srikar-64bitRandom/tests/unit Also, I believe this is the logic we came up with but I couldn't figure out how to pass or fail the test based on this as none of the unit tests I looked through in the tests/unit had this structure. Most of the other test code I looked at as examples used the [describe, test and expect methods](https://github.com/Expensify/App/blob/srikar-64bitRandom/tests/unit/GitUtilsTest.js#L43.). so I wanted to ask if there was somewhere I can see how to use those methods and if that is the right way to go about writing this unit test.

Screen Shot 2022-08-23 at 3 45 42 AM

tylerkaraszewski (dismissed):

luacmartins:
@srikarparsi yes, that's the correct path for the tests. As for the test themselves, I'd do something like the code below:

```
describe('libs/NumberUtils', () => {
  it('should generate a random 64-bit numeric string', () => {
    const id = NumberUtils.rand64();
    expect(id).toBe(...);
    ...
  });
})

File: tests/unit/NumberUtilsTest.js [hunk 9]
```diff
@@ -0,0 +1,14 @@
+const NumberUtils = require('../../src/libs/NumberUtils');
+
+const result = NumberUtils.rand64();
+
+const max64 = 9223372036854775807;
+
+describe('libs/NumberUtils', () => {
+    it('should generate a random 64-bit numeric string', () => {
+        const id = NumberUtils.rand64();
+        expect(isNaN(result)).toBe(false);
```
luacmartins, line 10: I think this is what you are looking for
```suggestion
        expect(isNaN(id));
```

tests/unit/NumberUtilsTest.js
```diff
@@ -0,0 +1,14 @@
...
+        expect(result > max64).toBe(false);
```
luacmartins, line 11: We should check if `.toBeGreaterOrEqual` works with strings.

```suggestion
        expect(id).toBeGreaterOrEqual(max64);
```

luacmartins (changes requested):

tests/unit/NumberUtilsTest.js
luacmartins: It seems like we can only pass number or bigint - https://jestjs.io/docs/expect#tobegreaterthanorequalnumber--bigint

```suggestion
        expect(result >= max64).toBe(true);
```

src/libs/NumberUtils.js
stitesExpensify: @luacmartins there is no link on your "here" 😄

luacmartins: Classic. Edited and added the link. Thanks!

tests/unit/NumberUtilsTest.js
```diff
@@ -0,0 +1,14 @@
+const NumberUtils = require('../../src/libs/NumberUtils');
+
+const result = NumberUtils.rand64();
+
+const max64 = 9223372036854775807;
```
tylerkaraszewski, line 5: How does this work? As far as I know, this will result in a floating point number with a loss of precision.

tests/unit/NumberUtilsTest.js
tylerkaraszewski: `result` (or `id`, which one are we using?) should be *less* than the max 64-bit number. But this is an inadequate test case (especially if `max64` is wrong). We should probably check it's >0 as well.

tylerkaraszewski (changes requested):

tests/unit/NumberUtilsTest.js
stitesExpensify: Yeah, I think this needs to be a BigInt

tests/unit/NumberUtilsTest.js
roryabraham, line 10:
```suggestion
        expect(Number.isNaN(result)).toBe(false);
```

tests/unit/NumberUtilsTest.js
luacmartins: it should be a string, no? since we are generating a numeric string id

tests/unit/NumberUtilsTest.js
luacmartins: we declared `const id = NumberUtils.rand64();` above, so it should be id and yes the comparison should be `id >= 0 && id <= max64)`

stitesExpensify:
cc @tylerkaraszewski @roryabraham

tests/unit/NumberUtilsTest.js
```diff
@@ -1,14 +1,13 @@
...
-const max64 = 9223372036854775807;
-
 describe('libs/NumberUtils', () => {
     it('should generate a random 64-bit numeric string', () => {
         const id = NumberUtils.rand64();
-        expect(isNaN(result)).toBe(false);
-        expect(result > max64).toBe(false);
+        expect(typeof id).toBe('string');
+        // eslint-disable-next-line no-undef
+        expect(BigInt(id)).toBeLessThanOrEqual(BigInt(9223372036854775807));
```
roryabraham, line 8: Using BigInt here is safe because BigInt is defined in Node and we are using the node environment for jest?

tylerkaraszewski (approved):

tests/unit/NumberUtilsTest.js
flodnv: Can we not use string comparison?

roryabraham (approved):
LGTM, I'm not worried about using BigInt in tests. If it passes locally and passes in GitHub Actions, then we're good.

tests/unit/NumberUtilsTest.js
stitesExpensify: We could if we want to, this just _feels_ better IMO because it makes them numbers instead of JS black magic with strings

flodnv: Black magic? All programming languages (that we use at Expensify) do string comparisons 😕 🤷 NAB

luacmartins (approved):

ctkochan22:
hmm this was breaking when I was running the desktop locally. Could be me though?
```
App [main] :> npm run desktop

> new.expensify@1.1.89-2 desktop
> scripts/set-pusher-suffix.sh && node desktop/start.js

Using PUSHER_DEV_SUFFIX from ../Web-Expensify/_config.local.php
Found suffix: 69db2c8c04e54064a2d1ca48a515d180
Updating .env
[Renderer] ℹ ｢wds｣: Project is running at http://localhost:8080/
[Renderer] ℹ ｢wds｣: webpack output is served from /
[Renderer] ℹ ｢wds｣: Content not from webpack is served from /Users/kosuke/Expensidev/App/dist
[Renderer] ℹ ｢wds｣: 404s will fallback to /index.html
[Main] Hash: 93c0b04638346214ccf0
[Main] Version: webpack 4.46.0
[Main] Time: 530ms
[Main] Built at: 08/24/2022 4:04:49 PM
[Main]            Asset      Size         Chunks             Chunk Names
[Main] contextBridge.js   306 KiB  contextBridge  [emitted]  contextBridge
[Main]          main.js  2.61 MiB           main  [emitted]  main
[Main] Entrypoint main = main.js
[Main] Entrypoint contextBridge = contextBridge.js
[Main] [./desktop/ELECTRON_EVENTS.js] 356 bytes {main} {contextBridge} [built]
[Main] [./desktop/contextBridge.js] 2.39 KiB {contextBridge} [built]
[Main] [./desktop/main.js] 13.7 KiB {main} [built]
[Main] [./node_modules/webpack/buildin/module.js] (webpack)/buildin/module.js 497 bytes {main} [built]
[Main] [./src/CONFIG.js] 3.37 KiB {main} [built]
[Main] [./src/CONST.js] 416 bytes {main} [built] [failed] [1 error]
[Main] [./src/libs/Url.js] 321 bytes {main} [built]
[Main] [./src/libs/checkForUpdates.js] 630 bytes {main} [built]
[Main] [./src/libs/getPlatform/index.desktop.js] 111 bytes {main} [built]
[Main] [electron] external "electron" 42 bytes {main} {contextBridge} [built]
[Main] [electron-context-menu] external "electron-context-menu" 42 bytes {main} [built]
[Main] [electron-log] external "electron-log" 42 bytes {main} [built]
[Main] [electron-serve] external "electron-serve" 42 bytes {main} [built]
[Main] [electron-updater] external "electron-updater" 42 bytes {main} [built]
[Main]     + 457 hidden modules
[Main]
[Main] ERROR in ./src/CONST.js 793:26
[Main] Module parse failed: Identifier directly after number (793:26)
[Main] You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
[Main] |         LEAVE_ROOM: 'leaveRoom',
[Main] |     },
[Main] >     MAX_64BIT_LEFT_HALF: 9_223_372_036,
[Main] |     MAX_64BIT_RIGHT_HALF: 854_775_807,
[Main] |     IOS_KEYBOARD_SPACE_OFFSET: -30,
[Main]  @ ./src/CONFIG.js 6:0-28 9:53-58 23:20-25
[Main]  @ ./desktop/main.js
[Main] webpack --config config/webpack/webpack.desktop.js --config-name desktop-main --mode=development && wait-port localhost:8080 && electron desktop/dist/main.js exited with code 2
--> Sending SIGTERM to other processes..
^C^[[B^[[A[Renderer] webpack-dev-server --config config/webpack/webpack.dev.js --port 8080 --env.platform desktop exited with code 0
```

luacmartins:
@ctkochan22 there's a thread [here](https://expensify.slack.com/archives/C03TQ48KC/p1661393972628819) if you want to follow

PR #10444, opened 2022-08-18
Refactor WorkspaceInitialPage to no longer use withFullPolicy

yuwenmemon:
@arosiclair Please review

### Details
Refactor WorkspaceInitialPage to no longer use `withFullPolicy` and use the `FullPageNotFoundView` instead.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/218369

### Tests/QA
- Create a Workspace if you do not have one already
- Click on Settings -> [Workspace]
- Make sure you can interact with the page fully and without any errors:
https://user-images.githubusercontent.com/4741899/185338389-1b69cfd1-4bf9-424d-934d-06635a312134.mp4

---

- Using the URL bar above, change the workspace ID to something else.
- Make sure you see that the workspace is shown to be "Not found":
https://user-images.githubusercontent.com/4741899/185338667-4492d192-998d-4676-a362-42a92a39e2d4.mp4

---

- Using the JS console, switch the network tab to off
- Navigate to the workspace, make sure you are still able to and you load the page fully like before.

---

- [ ] For all of the above make sure there are no JS console errors while you do any of this.


src/pages/workspace/WorkspaceInitialPage.js
```diff
@@ -240,10 +247,12 @@ WorkspaceInitialPage.displayName = 'WorkspaceInitialPage';
 
 export default compose(
     withLocalize,
-    withFullPolicy,
```
arosiclair, line 243: Sorry for the confusion, but we're no longer looking to remove `withFullPolicy` completely. Instead, we'll just be updating it to not fetch policy data so you can keep it here

arosiclair (changes requested):
Aside from withFullPolicy removal, everything else looks good!

src/pages/workspace/WorkspaceInitialPage.js
yuwenmemon: Hmmm... Yes, I am confused 😅 . Do we _need_ `withFullPolicy` here - why keep it? Also, the `withFullPolicy` code is still loading the full policy - when are we removing that?

arosiclair: So `withFullPolicy` also tracks the last visited workspace/policy so we shouldn't remove that functionality.

We'll remove the policy loading with [this GH](https://github.com/Expensify/Expensify/issues/217650) but this is only safe to do once all the other refactors are done (if we do it before, these pages will have no data).

yuwenmemon: Okay great - makes sense thanks for clarifying!

yuwenmemon:
Updated and retested!

arosiclair (approved):

PR #10448, opened 2022-08-18
[NO QA]Ignore all `PropTypes.arrayOf(PropTypes.object)`

AndrewGable:
### Details

It seems like some other change upgraded lint to now check for `PropTypes.arrayOf(PropTypes.object)` as being `forbid-prop-types`. This ignores all of those prop types to make lint pass.

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/223483

### Tests

- [X] Verify lint test passes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
     extends: ['expensify', 'plugin:storybook/recommended'],
     parser: 'babel-eslint',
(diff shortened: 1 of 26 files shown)


AndrewGable:
I tried tweaking the `rules` object in `eslintrc` and was unable to get something that worked for this specific error, so this is the best solution I could come up with. Let me know if you can think of a better one.

luacmartins (approved):
LGTM thanks for fixing this!

luacmartins:
Merging to fix linting issues in App!

PR #10452, opened 2022-08-18
Migrate policy API to remove policy members

iwiznia:
~Hold on https://github.com/Expensify/Web-Expensify/pull/34618~

Needs:
- https://github.com/Expensify/Auth/pull/7032
- https://github.com/Expensify/Web-Expensify/pull/34618/

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/211560
$ https://github.com/Expensify/App/issues/11067

### Tests

No QA:
- Play around with the site, make sure all is working (create chat, send/edit/delete message, change profile settings, create/delete workspace, signout/in, change language, profile settings). This can be skipped in QA since it's covered by the regression testing
- Debug the activeClients code and check it is working properly
- Trigger an error by modifying the code, check the error is returned in all the appropriate rows and RBR is shown and the errors are dismissible

QA:
- Invite a user to a workspace
- Reload the page
- On another browser window in incognito mode sign in as that user
- On the window where you invited the user:
  - Remove the same user from the workspace
  - Check the user gets removed from the policy members list
  - Check the user gets removed from the members in the #announce channel of the workspace
  - Check the workspace chat for the expenses of the user is archived  (you need to find this by email or have it selected before)
- On the window of the user you removed:
  - Check the #announce channel of the policy is removed from the LHN and that if you were looking at it, you see the `You don't have access to this chat` message
  - Check the workspace chat for the expenses of the user is archived 
- Invite a new user (that does not exist in expensify) to the policy, check it works properly
- Navigate to a report url with an invalidID like http://localhost:8080/r/hello, check you see the `You don't have access to this chat` message
- Navigate to a report url with the ID of a report you don't have access to like http://localhost:8080/r/1, check you see the `You don't have access to this chat` message
image

    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop
    - [x] I verified any copy / text that was added to the app is correct English and approved by marketing by tagging the marketing team on the original GH to get the correct copy.
    - [x] Any functional components have the `displayName` property

<details>

</summary>

    - [ ] iOS / native
    - [ ] Android / native
    - [ ] iOS / Safari
    - [ ] Android / Chrome
    - [ ] MacOS / Chrome
    - [ ] MacOS / Desktop

</details>

### QA Steps

#### Web

https://user-images.githubusercontent.com/521248/192854882-8b02a5d7-0c84-41cb-abcb-343f1ed62a7b.mov


iwiznia:
This is on hold, but let's get it reviewed in the meantime

src/libs/ActiveClientManager/index.js
```diff
@@ -24,7 +24,9 @@ function isReady() {
 Onyx.connect({
     key: ONYXKEYS.ACTIVE_CLIENTS,
     callback: (val) => {
-        activeClients = !val ? [] : val;
+        if (val) {
+            activeClients = _.unique(activeClients.concat(val));
```
tgolen, line 28: Suggestion: Use `if (val && activeClients.indexOf(value) < 0)` instead of `_.unique()`. I think it would be better to do if you're trying to prevent the same value being put into the array again.

(sorry, I'm jaded against `_.unique()` after learning that it's not very performant).

src/libs/actions/Report.js
```diff
@@ -1263,6 +1263,7 @@ function editReportComment(reportID, originalReportAction, textForNewComment) {
                 isEdited: true,
                 html: htmlForNewComment,
                 text: textForNewComment,
+                type: originalReportAction.message[0].type,
```
tgolen, line 1266: Why is this change in this PR? It seems unrelated.

src/pages/workspace/WorkspaceMembersPage.js
```diff
@@ -283,14 +283,19 @@ class WorkspaceMembersPage extends React.Component {
...
-        const removableMembers = _.without(policyMemberList, this.props.session.email, this.props.policy.owner);
-        const data = _.chain(policyMemberList)
-            .map(email => this.props.personalDetails[email])
-            .filter()
-            .sortBy(person => person.displayName.toLowerCase())
-            .map(person => ({...person})) // TODO: here we will add the pendingAction and errors prop
-            .value();
+        const policyMemberList = lodashGet(this.props, 'policyMemberList', {});
+        const removableMembers = [];
+        let data = _.map(policyMemberList, (value, email) => {
```
tgolen, line 288: `value` is very ambiguious. Name suggestion: `policyMember`

tgolen (changes requested):

src/libs/actions/Report.js
iwiznia: It is not. Messages is an array and we were not sending the type in the optimistic action (neither in the action from the server), so when we replace the message fully now, this property was not set and was causing issues (we would render the comment as `fragment.text` [here](https://github.com/Expensify/App/blob/47a776bf229a5704d7c4a337af3b455884cf72f1/src/pages/home/report/ReportActionItemFragment.js#L170)

src/libs/ActiveClientManager/index.js
iwiznia: Not that it matters much here, since at most there will be 20 items, but updated

src/pages/workspace/WorkspaceMembersPage.js
```diff
@@ -283,14 +283,19 @@ class WorkspaceMembersPage extends React.Component {
...
-        const removableMembers = _.without(policyMemberList, this.props.session.email, this.props.policy.owner);
-        const data = _.chain(policyMemberList)
-            .map(email => this.props.personalDetails[email])
-            .filter()
-            .sortBy(person => person.displayName.toLowerCase())
-            .map(person => ({...person})) // TODO: here we will add the pendingAction and errors prop
-            .value();
+        const policyMemberList = lodashGet(this.props, 'policyMemberList', {});
+        const removableMembers = [];
+        let data = _.map(policyMemberList, (policyMember, email) => {
```
chiragsalian, line 288: This logic is a bit weird. You are using a map to build data but also using it to generate removableMembers values. I feel like this would have been easier with a `_.each` to do different things.

Anyway besides that, I feel like data would be a lot cleaner with the `chain`. Its weird that it's doing some things here and then in the next line it does sortBy. Also its not doing filter so it can end up with empty values which will fail when lowerCase is done,
image

I think data should be generated via chain and removableMembers should keep its operations separate. Or use a _.each to build both but that's extra work which I feel is not worth it just yet here.

chiragsalian (changes requested):

src/pages/workspace/WorkspaceMembersPage.js
iwiznia: Updated to use each

src/libs/ActiveClientManager/index.js
```diff
@@ -24,7 +24,9 @@ function isReady() {
 Onyx.connect({
     key: ONYXKEYS.ACTIVE_CLIENTS,
     callback: (val) => {
-        activeClients = !val ? [] : val;
+        if (val && activeClients.indexOf(val) < 0) {
```
chiragsalian, line 27: Whats this change for? When i test your PR the command `DeleteMembersFromWorkspace` never gets called in the network tab and I think its because of the code here.

chiragsalian (changes requested):
I found two issues. Besides those the rest LGTM

src/pages/workspace/WorkspaceMembersPage.js
```diff
@@ -283,14 +283,20 @@ class WorkspaceMembersPage extends React.Component {
...
-        const data = _.chain(policyMemberList)
-            .map(email => this.props.personalDetails[email])
-            .filter()
-            .sortBy(person => person.displayName.toLowerCase())
-            .map(person => ({...person})) // TODO: here we will add the pendingAction and errors prop
-            .value();
+        const policyMemberList = lodashGet(this.props, 'policyMemberList', {});
+        const removableMembers = [];
+        let data = [];
+        _.each(policyMemberList, (policyMember, email) => {
```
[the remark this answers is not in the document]
iwiznia: That was a hack that someone added because the app crashed and instead of fixing the real issue, they just removed the detail from the list, which is wrong. This works if you are using the latest version of auth, that returns personal details for all emails, even if they don't have explicitly set a name.

Well, at least I hope the crash you saw was due to outdated auth, sounds like it because it's the same I was getting before sending the auth fix.

src/libs/ActiveClientManager/index.js
iwiznia: I had to refactor this a bit because this was using the array merging feature of onyx that we changed. It is not called in the flow of that api command.

src/pages/workspace/WorkspaceMembersPage.js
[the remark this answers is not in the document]
iwiznia: Oh, you also need the web PR of course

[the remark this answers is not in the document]
chiragsalian: Odd, im sure i updates all repos and did a makeall yesterday. Maybe its buggy on an existing workspace, I'll try again on a fresh workspace.
If its buggy on existing data we still have to consider it so that users existing workspaces don't crash.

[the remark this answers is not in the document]
iwiznia: Should not be buggy on existing data. Maybe.you need to sign out and back in though, not sure.

chiragsalian:
Discussed 1:1 with ioni. The app still crashes and network requests are a bit broken. 
We found out the crash occurred if a non-existing user was invited.
We were able to confirm the activeClients change is causing the network requests problem.

I believe he mentioned he'll debug both on Monday now that he has more clues on why its breaking.

iwiznia:
This is fixed and added a test for inviting non existing users. I also discovered that my fix for the personal details list is incomplete and was not returning people in your policies without personal details. That makes it so that when that happens, it shows a blank avatar. Submitted an auth PR to fix that, but will not block this on it.

iwiznia:
Bump @nkuoch @tgolen

src/libs/ActiveClientManager/index.js
```diff
@@ -6,38 +12,61 @@ import * as ActiveClients from '../actions/ActiveClients';
...
+        let removed = false;
+        while (activeClients.length >= maxClients) {
+            activeClients.shift();
+            removed = true;
+        }
+
+        // Save the clients back to onyx, if they changed
+        if (removed || !addedSelf) {
+            ActiveClients.setActiveClients(activeClients).then(() => {
+                // If we just added ourselves, we need to resolve the promise so that isReady fires
```
tgolen, line 54: This comment doesn't really seem consistent with the logic below. It says that the promise needs resolved if we just added ourselves, but shouldn't that mean that the early return should be `if (!addedSelf)`?

src/libs/ActiveClientManager/index.js
```diff
@@ -6,38 +12,61 @@ import * as ActiveClients from '../actions/ActiveClients';
...
+    resolveSavedSelfPromise = resolve;
 });
 
 /**
+ * Determines when the client is ready. We need to wait both till we saved our ID in onyx AND the init method was called
  * @returns {Promise}
  */
 function isReady() {
-    return isReadyPromise;
+    return Promise.all([initedPromise, savedSelfPromise]);
```
tgolen, line 31: It seems to me like the `initedPromise` isn't really doing anything. Is it necessary? It feels like to me that `saveSelfPromise` is the only one that is needed (there is no way it can ever resolve without the lib initialized).

src/libs/actions/Policy.js
```diff
@@ -699,21 +691,14 @@ function updateLastAccessedWorkspace(policyID) {
...
-                });
-            }
-
+        Pusher.subscribe(pusherChannelName, 'policyEmployeeRemoved', ({removedEmails, defaultRoomChatIDs}) => {
             // Remove the default chats if we are one of the users getting removed
-            if (removedEmails.includes(sessionEmail) && !_.isEmpty(defaultRoomChatIDs)) {
-                _.each(defaultRoomChatIDs, (chatID) => {
-                    Onyx.set(`${ONYXKEYS.COLLECTION.REPORT}${chatID}`, null);
-                });
+            if (!removedEmails.includes(sessionEmail) || _.isEmpty(defaultRoomChatIDs)) {
```
tgolen, line 696: Is it possible to do this logic in PHP so that the logic on the client can be more simple? Also, why isn't PHP sending Onyx instructions that are passed to `Onyx.update()`?

tgolen (changes requested):

src/libs/ActiveClientManager/index.js
iwiznia: We don't want to say that it is ready if we have not been instructed to initialize. We call init after the UI is ready.

src/libs/ActiveClientManager/index.js
iwiznia: Stupid linter made me change this to early return and now is a bit more confusing, but no, both are correct. We only resolve the promise if we have not addedSelf yet, which is what the code is doing

src/libs/actions/Policy.js
iwiznia: Ah yes, we are doing this in PHP already, we can now remove all this

iwiznia: Updated

src/libs/ActiveClientManager/index.js
tgolen: Hm, so you're saying that `savedSelfPromise` can be resolved before the lib is initialized? But if that's true, then why does initialization matter? Initialization doesn't actually do anything, it just resolves a promise. I think the only thing that matters is that `savedSelfPromise` has been done.

iwiznia: It does not do anything in itself, it's indicating that everything is ready. If we did not have this, then it's possible that savedSelfPromise resolves first and starts returning isReady when the UI has not been initialized. 
I tried doing what you suggest and things did not work (in some cases, I think it was hard to reliably reproduce since the order of events is not ensured). What I saw happening is that the saved network request would not fire immediatelly, but instead the next time the sequential network code fired.

src/libs/ActiveClientManager/index.js
tgolen: OK, that still doesn't make a whole lot of sense to me. That would imply that this callback is called multiple times. Once with `addedSelf === true` and once with `addedSelf === false`, right? But since `ActiveClients.setActiveClients()` is wrapped in an `if (!addedSelf)`, how is it ever called with `addedSelf === true`?

Is the early return even necessary?

iwiznia: The callback is only called once, but `isReady` is called many times (each time the sequential queue runs).

> Is the early return even necessary?

hmmm probably not, because resolving a promise multiple times does nothing really... right?

src/libs/ActiveClientManager/index.js
tgolen: > it's indicating that everything is ready

Isn't this precisely what `savedSelfPromise` is for though?

> and starts returning isReady when the UI has not been initialized

I don't understand why `ActiveClients` cares if the UI has been initialized or not. Is something else relying on that? This kind of feels like something is being held wrong. Like, `ActiveClients` should not be the lib that is telling other things that the UI is initialized.

iwiznia: There's 2 things:
1. We need the list of activeClients to be updated before we start doing things (otherwise 2 tabs could send the same request)
2. We need to wait till the UI is ready before sending the any requests. Not sure exactly why, but this is what we did before, we did not do anything till init was called, even if the client was already saved.

> ActiveClients should not be the lib that is telling other things that the UI is initialized

It is not, it's the other way around, other code is telling ActiveClients that the UI is initialized, same as before.

src/libs/ActiveClientManager/index.js
tgolen: Yeah, resolving a promise twice does not trigger anything again (ie. any `.then()`s will not be called)

iwiznia: I think I could remove the early return then

src/libs/ActiveClientManager/index.js
iwiznia: > we already agreed that calling init() does NOTHING

No, we did not. We agreed it was signaling when the UI is ready, so that isReady does not return true till it is.

> 2 does not. I'd like to get to the bottom of that rather than just adding a second promise and assuming that what was there before needs to stay there, especially when it's not fully understood.

Fair. I did try what you said and observed what I said above (queue not firing immediately after load, like it should).
Do you want me to delve more into it and figure out why that was happening and why we added the init method in the first place?

iwiznia: One thing I want to clarify: this is broken right now, I have the suspicion is why we are seeing reportID collisions.
So, maybe we can leave as is and investigate all that as a follow up.

tgolen: Yes, could you delve into it more? I'd like to make sure we have a full understanding of this before changing it. When I look at the `init()` method in main:

```
function init() {
    ActiveClients.addClient(clientID)
        .then(resolveIsReadyPromise);
}
```

It's adding the current client, then resolving the ready promise. That tells me that the only thing that matters for the lib to be initialized is that the current client has been added. I don't think there is anything else beyond that.

I'm not sure about the reportID collisions. That's the first I've heard about it, so I don't know why that would imply that activeClients is broken.

iwiznia: > I'm not sure about the reportID collisions. That's the first I've heard about it, so I don't know why that would imply that activeClients is broken.

I am 99.9% sure activeClients is broken since we are calling merge on an array, which always replaces the first index of the array. Maybe it is not causing the [reportID collision](https://github.com/Expensify/Expensify/issues/228658), but it could.

> It's adding the current client, then resolving the ready promise. That tells me that the only thing that matters for the lib to be initialized is that the current client has been added. I don't think there is anything else beyond that.

If that's the case, why didn't we save the activeClient directly in this file? Why would we need to tell it to "init"?

iwiznia: BTW, I will delve into this tomorrow.
Going to cc @AndrewGable and @marcaaron here, because I see you both worked on this code and might have some context.

iwiznia: Looking at past commits, maybe it was because we wanted to [verify the authToken first](https://github.com/Expensify/App/commit/a06d2f2e9b2fbf2bd05a89c1fd23a4af5990d67c#diff-87897d9291cc82037a7d77b5c5196619a383acebf46cf631ae84192a0efa480aR34)? Though not sure if that exists anymore...

marcaaron: Trying to follow this conversation but a little lost so just going to write a bunch of stuff about how things work now...

>I am 99.9% sure activeClients is broken since we are calling merge on an array, which always replaces the first index of the array

This is not true for a pure array value like `activeClients`. We concat the value onto the existing array values:

https://github.com/Expensify/react-native-onyx/blob/98260f1deacbcca1a687441164cd45d7d915afb8/lib/Onyx.js#L653-L659

If we look at `main` the way this is intended to work is:

1. User gets the active clients in the initial collection (pruning the head of the list if necessary)
2. We then add the client to the end of the existing list with the `merge()` call
3. We then resolve the "is ready" promise so that anything that needs to check if the current client is the leader can do so with accuracy.

The reasoning behind 3 is if we call `merge()` and wait for it to finish we can also guarantee the `Onyx.connect()` callback has been called again with the updated value for `activeClients`.

As for whether there is some reason we are setting the client in `<Expensify/>` vs. somewhere else. I guess my first question is... where else should it be and what would make that a better place? Does it matter that we are waiting for that component to get constructed? We are going to "wait" until the active client is set at least once. We could probably do this in the `setup.js` code where `Onyx.init()` is called or anywhere early in the app start process. 

Noted that we also set the client again when the app state changes (which presumably happens when you switch to a new tab) and the listener for that is setup in `<Expensify/>` so my guess is that we just thought it was a logical place to stick this.

AFAIK, the only reason we care about the client being the leader is so we can avoid making duplicate requests  and some logic to prevent duplicate browser notifications.

>the only thing that matters for the lib to be initialized is that the current client has been added

Added _and_ the local variable for `activeClients` updated - which we can only guarantee by awaiting the `merge()`.

iwiznia: ok, updated and retested this. I indeed removed the initialize promise.

src/libs/ActiveClientManager/index.js
```diff
@@ -36,8 +49,9 @@ Onyx.connect({
  * Add our client ID to the list of active IDs
  */
 function init() {
-    ActiveClients.addClient(clientID)
-        .then(resolveIsReadyPromise);
+    activeClients = _.without(activeClients, clientID);
+    activeClients.push(clientID);
```
tgolen, line 53: I believe the intention of doing it like this is to ensure that the clientID is always added at the end of the array and its only added once. Would you mind updating the method description to talk about that? I know it's mentioned pretty well in the comment at the top of the file, but having the method doc comment talk about it is a little better IMO since it's closer to the code.

tgolen (changes requested):
Looks great now! Thanks for removing that extra promise 👍

tgolen (dismissed):
Perfect, thanks!

chiragsalian (dismissed):
LGTM and works well.

src/libs/actions/Policy.js
```diff
@@ -218,29 +216,21 @@ function removeMembers(members, policyID) {
...
+    const membersListKey = `${ONYXKEYS.COLLECTION.POLICY_MEMBER_LIST}${policyID}`;
+    const optimisticData = [{
+        onyxMethod: CONST.ONYX.METHOD.MERGE,
+        key: membersListKey,
+        value: _.object(members, Array(members.length).fill({pendingAction: CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE})),
+    }];
+    const failureData = [{
+        onyxMethod: CONST.ONYX.METHOD.MERGE,
+        key: membersListKey,
+        value: _.object(members, Array(members.length).fill({errors: {[DateUtils.getMicroseconds()]: Localize.translateLocal('workspace.people.error.genericRemove')}})),
```
chiragsalian, line 228: I have a question with this failureData. It looks good to me, but assume it fails when removing a user. Assume the  the failure happens in PHP/auth layer. Now your PHP code [here](https://github.com/Expensify/Web-Expensify/pull/34618/files#diff-31780863b86f2d567a1f2e3743719ba58bf64ed59edb6506e6d812ae897a58a4R839) will throw an ExpError with an `errorMessage`. 
Since the API is a failure, the failureData here will get triggered to insert all members including removed ones with the error. Additionally the workspace.people.error.genericRemove is also inserted. 
So if you removed 3 users, and the PHP layer failed on user 2.
1. Won't all 3 users get `workspace.people.error.genericRemove` added to them since the API is considered a failure? 
2. Won't user 2 show `workspace.people.error.genericRemove` and `errorMessage` i.e., two error messages which could be confusing?

Anyway my concern is mostly point 1, showing user1 and user3 back even though they were successfully removed, because push notification would run first to remove them and then failureData from the ExpError of user2 would add them back, right?

Let me know if that makes sense, if so I think maybe we should not have failureData over here.

chiragsalian: Also for any es device won't it be weird that the PHP layer is providing the error in English while the app adds `workspace.people.error.genericRemove` in Spanish?

iwiznia: > Also for any es device won't it be weird that the PHP layer is providing the ExpError in English while the app adds workspace.people.error.genericRemove in Spanish? and then the app would display both.

We don't care. We are not supporting multiple languages.

As for the other comment, you are incorrect on point 1, here's a test I did by adding a throw before calling sharePolicy on a specific user (I had removed 2 users):
image

As you can see, you are correct on point 2. I thought we would replace it, but that's not correct since we merge... this is a problem in many other APIs, since this is the pattern we are following in lots of them... I don't have a great solution for it right now, because:
- If we add the generic error, we get this behavior
- If we don't add the generic error, we risk not showing any error at all if what we got from the request was a failure with no data back

Any ideas? In any case, I don't think I would want to block on this, given this problem is endemic and must probably be fixed in many other places.

src/pages/workspace/withFullPolicy.js
```diff
@@ -73,9 +72,6 @@ const fullPolicyPropTypes = {
         */
         errorFields: PropTypes.objectOf(PropTypes.objectOf(PropTypes.string)),
     }),
-
-    /** The policy member list for the current route */
-    policyMemberList: PropTypes.objectOf(policyMemberPropType),
```
Santhosh-Sellavel, line 78: Why is this removed?

Santhosh-Sellavel:
> RBR is shown and the errors are dismissible
RBR means?

iwiznia:
The red dots leading you to where the error is (and the error itself)

src/pages/workspace/withFullPolicy.js
iwiznia: Damn, good question. If I have to be honest, I don't quite recall... 🤔

iwiznia: Oh I remember! Because the member list is now it's own key, not inside the policy anymore

Santhosh-Sellavel: They already separate key along with policy

Santhosh-Sellavel:
> Trigger an error by modifying the code, 

Can you share an example

src/libs/ActiveClientManager/index.js
marcaaron: I actually don't get why are we concerned with it only being added once. 

We only add more clients and take them away to keep the list of IDs from growing infinitely. The active client will always be the last one regardless of whether we clean the current user clientID.

iwiznia:
> Can you share an example

No, because I was referring to backend code. That's why those are under no QA section.

src/pages/workspace/withFullPolicy.js
Santhosh-Sellavel: @iwiznia

Santhosh-Sellavel:
> #### On the window of the user you removed:
> - Check the #announce channel of the policy is removed from the LHN
> - Check the workspace chat for the expenses of the user is archived

#announce channel is not removed

https://user-images.githubusercontent.com/85645967/191586236-87fba0b8-3309-431d-9503-b3d1e1d110be.mov

 workspace chat means what are you referring here?

src/pages/workspace/withFullPolicy.js
iwiznia: Yep, that's what I said too 😄

iwiznia:
> workspace chat means what are you referring here?

The workspace chat is the chat that gets created when you are invited to a policy, that says `Collaboration between USERNAME and WORKSPACENAME starts here! 🎉 This is the place to chat, request money and settle up.`

> #announce channel is not removed

Daaaamn, I wonder why, since both me and Chirag tested this... I will take a look tomorrow.

Santhosh-Sellavel:
> The workspace chat is the chat that gets created when you are invited to a policy, that says `Collaboration between USERNAME and WORKSPACENAME starts here! 🎉 This is the place to chat, request money and settle up.`

Are you saying referring to the welcome message in the #announce channel?

iwiznia:
> Are you saying referring to the welcome message in the #announce channel? If not whats the channel/report name?

No, that one says: `Collaboration between all WORKSPACENAME members starts here! 🎉 Use #announce to chat about anything WORKSPACENAME related.`.
The one I am referring to looks like this (my workspace name is TestDeletes):
image

Now that I am thinking about it, maybe you are not seeing this because you are not in the beta...

Santhosh-Sellavel:
Check the #announce channel of the policy is removed from the LHN
Check the workspace chat for the expenses of the user is archived

## After deleting a user from the workspace
Workspace chat got removed from LHN after refresh I found it appeared again it's archived as expected.
But the #announce channel still exists on LHN

iwiznia:
ok, found the error. You were totally right, not sure how it worked in my previous tests (probably fucked up the test).
Bad news is that it is a backend change, so you won't be able to test till it hits production.

iwiznia:
@Santhosh-Sellavel did all the other tests pass? If so, I might consider unblocking this anyway.

Santhosh-Sellavel:
> Invite a new user (that does not exist in expensify) to the policy, check it works properly

I just added a email account that was never created in Expensify to workspace. Invite & remove was successful

Quick Question after adding each user new chat is created for collaboration every time is this expected?
Screenshot 2022-09-23 at 11 18 48 PM

iwiznia:
Yes. The idea is that that chat room is where people upload their expenses to be paid by the company. We still have not built that functionality, but the chat rooms themselves exist. That's why it's on a beta, because it's missing the option to create expenses there.

Santhosh-Sellavel:
All looks good, #announce channel also gets hidden from LHN if the added user is not active on the report.

If the user is active #annouce an infinite loading appears on the screen & console errors.

https://user-images.githubusercontent.com/85645967/192363008-57d205f2-c353-4527-9089-a6aa8e873fd6.mp4

On Remove Chats disappear immediately, chat's not showing automatically is bit weird to test things.

iwiznia:
This was updated to show the `You don't have access to this chat` page. Please re-review.

iwiznia:
Something seems wrong with the linter... Posted [here](https://expensify.slack.com/archives/C03TQ48KC/p1664388638689549), but adding here too.

- The line marked does not have the class mentioned, so the line number seems wrong
- The class mentioned is in use in the file, so the error itself seems wrong

Santhosh-Sellavel:
<details>

</summary>

    - [ ] iOS / native
    - [ ] Android / native
    - [ ] iOS / Safari
    - [ ] Android / Chrome
    - [ ] MacOS / Chrome
    - [ ] MacOS / Desktop

</details>

Santhosh-Sellavel:
All looks good, one doubt. On adding a new user to the workspace, immediately workspace appears on LHN for the user but #announce chat is not, is this the expected behavior?

src/libs/ActiveClientManager/index.js
tgolen: I agree with Marc, but it's a NAB for me on this PR. Maybe a follow-up

tgolen (dismissed):

JmillsExpensify:
>On adding a new user to the workspace, immediately workspace appears on LHN for the user but #announce chat is not, is this the expected behavior?

This is not expected, but also not sure if that's a blocker. I think we have another issue for this, in fact.

marcaaron (dismissed):
LGTM just added some non blocking questions to improve my understanding of how things should work.

Santhosh-Sellavel (approved):
All good tests well!

Santhosh-Sellavel (dismissed):

src/pages/home/ReportScreen.js
```diff
@@ -237,71 +234,76 @@ class ReportScreen extends React.Component {
                 style={[styles.appContent, styles.flex1, {marginTop: this.state.viewportOffsetTop}]}
                 keyboardAvoidingViewBehavior={Platform.OS === 'android' ? '' : 'padding'}
             >
-                <OfflineWithFeedback
-                    pendingAction={addWorkspaceRoomPendingAction}
-                    errors={addWorkspaceRoomErrors}
-                    errorRowStyles={styles.dNone}
+                <FullPageNotFoundView
+                    shouldShow={!this.props.report.reportID}
```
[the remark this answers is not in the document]
iwiznia: When we receive the event from the server to set the report to `null` (when you lose access to the chat), if you were looking at the chat, then that's when we hit this flow. At first coded this to check if report was `null` and it was not working, when I debugged it, I saw this is what we receive right after the event is sent:
image
Not quite sure why those properties are there when that happens though

src/libs/actions/Policy.js
```diff
@@ -209,29 +205,21 @@ function removeMembers(members, policyID) {
...
```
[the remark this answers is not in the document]
iwiznia: Yes, agree

src/libs/actions/Policy.js
```diff
@@ -653,31 +641,6 @@ function updateLastAccessedWorkspace(policyID) {
...
- */
-function subscribeToPolicyEvents() {
-    _.each(allPolicies, (policy) => {
-        const pusherChannelName = `public-policyEditor-${policy.id}${CONFIG.PUSHER.SUFFIX}`;
-        Pusher.subscribe(pusherChannelName, 'policyEmployeeRemoved', ({removedEmails, policyExpenseChatIDs, defaultRoomChatIDs}) => {
-            // Refetch the policy expense chats to update their state and their actions to get the archive reason
-            if (!_.isEmpty(policyExpenseChatIDs)) {
-                Report.fetchChatReportsByIDs(policyExpenseChatIDs);
-                _.each(policyExpenseChatIDs, (reportID) => {
-                    Report.reconnect(reportID);
```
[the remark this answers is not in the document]
iwiznia: Policy expense chats should be sent via onyx when adding, removing or changing the role of users

src/libs/actions/Policy.js
```diff
@@ -653,31 +641,6 @@ function updateLastAccessedWorkspace(policyID) {
...
-                });
-            }
-
-            // Remove the default chats if we are one of the users getting removed
-            if (removedEmails.includes(sessionEmail) && !_.isEmpty(defaultRoomChatIDs)) {
-                _.each(defaultRoomChatIDs, (chatID) => {
-                    Onyx.set(`${ONYXKEYS.COLLECTION.REPORT}${chatID}`, null);
-                });
```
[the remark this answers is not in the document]
iwiznia: Yes. Added a screenshot of how it looks to the description

Santhosh-Sellavel:
### Desktop
https://user-images.githubusercontent.com/85645967/193297317-e1b19e25-db3b-46da-a563-8a6aebb58d3f.mov

### Web
https://user-images.githubusercontent.com/85645967/193298100-9f39edc6-c2dc-4adc-99b8-b960050bc30a.mov

### mWeb
https://user-images.githubusercontent.com/85645967/193297472-c651afca-6988-4f87-96a7-392adbaccc7b.mov

### iOS
https://user-images.githubusercontent.com/85645967/193297533-754efa1e-a22b-445b-aa51-8845d1e2f02f.mov

### Android
https://user-images.githubusercontent.com/85645967/193313299-8ad48e53-1ed4-41b2-8c35-528faea2d9e5.mov

Santhosh-Sellavel (changes requested):
Left a comment [here](https://github.com/Expensify/App/pull/10452?notification_referrer_id=NT_kwDOBRraj7M0MjIzOTAxODc2Ojg1NjQ1OTY3#issuecomment-1263773424) should fix them too!

iwiznia:
ok, updated to fix that.

tgolen (approved):

Santhosh-Sellavel:
Screenshot 2022-09-30 at 10 46 09 PM

Error message text is not centered here

Santhosh-Sellavel (approved):
Thanks, @iwiznia, all looks good now!

chiragsalian (approved):

iwiznia:
Not true Melvin, go home, you are drunk
image

src/pages/home/ReportScreen.js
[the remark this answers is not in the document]
mountiny: @iwiznia I am quite sure this is [causing a deploy blocker](https://github.com/Expensify/App/issues/11503). For some reason, the reportID is not accessible and then the `shouldShow` is true for a split of a second when creating a report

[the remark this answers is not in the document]
marcaaron: Late reply, but I think it could be better use something more reliable like `report.isAccessible: false` vs. a `report` not having a `reportID`. 

Perhaps there are other unexpected reasons why the report has no `reportID` and then instead of being intentional about when to show this "not found" view we will show this view randomly.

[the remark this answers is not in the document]
marcaaron: >Not quite sure why those properties are there when that happens though

Yeah, I think I would feel better about the change we made in this PR if we did understand why those properties are there and leave some kind of comment so that the intention is clear.

mananjadhav:
I am tagging this PR to highlight an issue fixed [here](https://github.com/Expensify/App/issues/16683). All conditions in ternary expressions or left-hand operands on conditional renders, should be boolean. This PR is one of the PRs that uses conditional render with string operands, hence I am tagging it here for the contributors to check. 

We've also updated the item in the checklist with this [PR](https://github.com/Expensify/App/pull/18069) to avoid this issue in the future.

src/pages/home/ReportScreen.js
```diff
@@ -239,81 +236,91 @@ class ReportScreen extends React.Component {
...
-                        onPress={this.chatWithAccountManager}
-                        shouldShowCloseButton
-                    />
-                )}
-                <View
-                    nativeID={CONST.REPORT.DROP_NATIVE_ID}
-                    style={[styles.flex1, styles.justifyContentEnd, styles.overflowHidden]}
-                    onLayout={event => this.setState({skeletonViewContainerHeight: event.nativeEvent.layout.height})}
+                <FullPageNotFoundView
+                    shouldShow={!this.props.report.reportID}
```
s77rt, line 240: If we are viewing a report and it gets deleted we will get the not found view, this caused https://github.com/Expensify/App/issues/28969. We should have taken into consideration if the report was previously accessible.

Issue #7948, opened 2022-03-01
[Performance] Refactor `SidebarLinks` to not use `getDerivedStateFromProps()` and other static methods


marcaaron:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

Coming from https://github.com/Expensify/App/pull/7907/files#r816376801

## Code cleanup

The file here seems to be following some practices that we should avoid because they are very uncommon in this codebase:

https://github.com/Expensify/App/blob/00805b2b495fac25f53a42231dd3e5ae78138a2a/src/pages/home/sidebar/SidebarLinks.js#L165-L174

There are some well documented reasons why we should use this and not here:

https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

And it _looks_ like we just want to perform a side effect where when various props change we update the reports we display in the sidebar. But for some reason have used derived state to do this instead of something like `componentDidUpdate()`.

Finding this component pretty hard to understand in general...

## Solution

1. Clean this up and make it use `componentDidUpdate()`
2. Put something in the style guide or eslint rule to forbid this practice
- [ ] Be sure to complete [this TODO](https://github.com/Expensify/App/pull/10619/files#diff-2ffec546a1daed561661d53400f598af96b06560c6d842886e2c100bbfb34654R239)

ctkochan22:
Taking up your offer on letting me steal this :)

parasharrajat:
There is one more component that uses it. `PopoverWithMeasuredContent`. 

But I think it is needed there. The popover render function is expensive for EmojiPicker and using it prevents render until we are fully ready. I don't remember why I used it there but there was a race condition that I was trying to overcome.

parasharrajat:
Ok. Great.

ctkochan22:
Punting

tgolen:
Work is coming along here nicely. Getting a lot of unit tests written

trjExpensify:
Added this App issue to the [WhatsApp Quality project](https://github.com/orgs/Expensify/projects/16/views/1) to keep track of it.

tgolen:
Got the first PR done which optimizes some of the current code which figures out the order reports should be displayed in. My second PR refactors the sidebar to render items more efficiently.

tgolen:
The last PR is in final stages of review and testing. It should merge this week!

[...]

PR #10489, opened 2022-08-23
Remove has draft collection

tgolen:
We recently discussed the removal of these collections that only hold boolean values.

While starting to work on https://github.com/Expensify/App/issues/7948 I wanted to start with fixing these drafts to help clean things up before getting into the meat of the issue.

### Details
This PR removes the old Onyx collection for `reportHasDraft_` and move the information to a `hasDraft` property on each report.

### Fixed Issues
Related to https://github.com/Expensify/App/issues/7948

### Tests
Covered by unit tests

### QA Steps
1. Open a chat
2. Start writing a comment
3. Verify the edit/pencil icon appears in the LHN for the active report
4. While still having a draft comment, switch to another report
5. Start writing a comment on the new report
6. Verify that both reports have the edit/pencil icon showing, since they have drafts

### DIFF
--- a/src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
+++ b/src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
@@ -7,7 +7,8 @@ export default function () {
     return new Promise((resolve) => {
         // Connect to the reportsWithDraft_ collection key in Onyx to get all the old data for
         // which reports have drafts.
-        // For each report that has a draft, set the hasDraft property in the report collection to true.
+        // For each report that has a draft, set the hasDraft property
+        // in the report collection to true.
         // Finally, remove the data from the old collection
         const connectionID = Onyx.connect({
             key: ONYXKEYS.COLLECTION.REPORTS_WITH_DRAFT,
--- a/src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
+++ b/src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
@@ -5,6 +5,46 @@ import ONYXKEYS from '../../ONYXKEYS';
 
 export default function () {
     return new Promise((resolve) => {
-        return resolve();
+        // Connect to the USER key in Onyx to get the value of expensifyNewsStatus
+        // then set that value as isSubscribedToNewsletter
+        // finally remove expensifyNewsStatus by setting the value to null
+        const connectionID = Onyx.connect({
+            key: ONYXKEYS.COLLECTION.REPORTS_WITH_DRAFT,
+            waitForCollectionCallback: true,
+            callback: (reportsWithDraftCollection) => {
+                Onyx.disconnect(connectionID);
+
+                if (!_.size(reportsWithDraftCollection)) {
+                    Log.info('[Migrate Onyx] No reports in the old draft collection');
+                    return resolve();
+                }
+
+                const reportIDsWithDrafts = [];
+                _.each(reportsWithDraftCollection, (reportHasDraft, onyxKey) => {
+                    if (reportHasDraft) {
+                        reportIDsWithDrafts.push(onyxKey.replace(ONYXKEYS.COLLECTION.REPORTS_WITH_DRAFT, ''));
+                    }
+                });
+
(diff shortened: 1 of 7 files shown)


src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
```diff
@@ -0,0 +1,68 @@
...
+                const mergeObject = _.reduce(reportIDsWithDrafts, (finalMergeObject, reportID) => ({
+                    [`${ONYXKEYS.COLLECTION.REPORT}${reportID}`]: {hasDraft: true},
+                    ...finalMergeObject,
+                }), {});
+
+                // eslint-disable-next-line rulesdir/prefer-actions-set-data
+                Onyx.mergeCollection(ONYXKEYS.COLLECTION.REPORT, mergeObject)
+                    .then(() => {
+                        // Resolve our main promise once all the reports have been removed from Onyx
+                        const done = _.after(allReportIDs.length, () => {
```
marcaaron, line 50: I don't quite get this. What is `_.after()` doing? Is it like a `setInterval()` watching the `allReportIDs`?

marcaaron, line 50: I don't quite get this. What is `_.after()` doing? Is it like a `setInterval()` watching the `allReportIDs`?

src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
```diff
@@ -0,0 +1,68 @@
+import Onyx from 'react-native-onyx';
+import _ from 'underscore';
+import Log from '../Log';
+import ONYXKEYS from '../../ONYXKEYS';
+
+export default function () {
+    return new Promise((resolve) => {
```
marcaaron, line 7: NAB, but not sure if the migration code is worth the complexity it adds. Is the worst case that we just start by assuming no reports have drafts and correctly track when a user edits the composer that has a draft in it? That seems pretty benign.

marcaaron (dismissed):
Nice changes! I had some questions about the migration code. Mostly not sure if it's needed, but NAB.

src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
jasperhuangg: ^^ just confirm my understanding, this would wipe out all existing drafts on all reports, right?

src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
jasperhuangg: so we're not going to run the anon callback function provided as the second argument into `after` until `done` has been called `allReportIDs.length` times, right?

jasperhuangg (dismissed):
Overall looks good! Just had a few clarifying questions mainly for my own understanding.

src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
tgolen: Ah, this is an old habit of mine and I think we have newer ways of doing this. It allows us to know when all the promises have completed and the triggers the final `resolve()`. I think we have something like `Promises.all()` now or something.

src/libs/migrations/MoveReportHasDraftCollectionToReportObject.js
tgolen: > Is the worst case that we just start by assuming no reports have drafts and correctly track when a user edits the composer that has a draft in it? That seems pretty benign.

Yeah, this is correct. It would also leave some stale data in Onyx that nothing is referencing. Also, not the end of the world. I am totally fine ditching the migration, but at least wanted to present what a true migration would be.

> this would wipe out all existing drafts on all reports, right?

If there was no migration, it doesn't delete the drafts, but it removes the edit/pencil icon in the LHN for any report with a draft.

tgolen:
all my commits are unsigned... great

tgolen:
Closing this PR in favor of https://github.com/Expensify/App/pull/10532 due to unsigned commits. I also removed the migration.

PR #10565, opened 2022-08-25
[NoQA] add generatePolicyID

marcochavezf:
cc @thienlnam 

### Details

Add `generatePolicyID()` to generate `PolicyID` on the client.

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/224679

### Tests

N/A This function is going to be used for https://github.com/Expensify/Expensify/issues/224690

### QA Steps

N/A

### Screenshots

N/A

### DIFF
--- a/src/libs/actions/Policy.js
+++ b/src/libs/actions/Policy.js
@@ -568,7 +568,7 @@ function hasPolicyMemberError(policyMemberList) {
 }
 
 /**
- * Returns a string representation of a 64-bit hexadecimal number
+ * Returns a client generated 16 character hexadecimal value for the policyID
  * @returns {String}
  */
 function generatePolicyID() {
@@ -596,4 +596,5 @@ export {
     clearDeleteMemberError,
     clearAddMemberError,
     hasPolicyMemberError,
+    generatePolicyID,
 };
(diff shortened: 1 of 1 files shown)


marcochavezf:
Hmm it's failing because of `no-unused-vars` eslint rule, I think it will be better to move this function to [NumberUtils.js](https://github.com/Expensify/App/blob/main/src/libs/NumberUtils.js) and import it into `Policy.js` when it's used. What do you think @thienlnam? In that way, I could also add a unit test [here](https://github.com/Expensify/App/blob/98f6a1b0a67d12cada9b9b50132518f6087e2fed/tests/unit/NumberUtilsTest.js)

thienlnam (dismissed):
Looking good! Just one nab about the comment

thienlnam:
> Hmm it's failing because of no-unused-vars eslint rule,

You should just be able to include the variable in the export below to fix this error

thienlnam (approved):
looks great, thanks for the ⚡ !

thienlnam:
Going to merge to push this forward - feel free to leave comments still if you notice something @MonilBhavsar

thienlnam:
Tests had passed

PR #10633, opened 2022-08-29
[No QA] Remove unused network loading key

tgolen:
### Details
This is no longer used since the reconnection spinner was removed from the avatar

### Tests
None


src/libs/NetworkConnection.js
```diff
@@ -1,7 +1,6 @@
 import _ from 'underscore';
 import NetInfo from '@react-native-community/netinfo';
 import AppStateMonitor from './AppStateMonitor';
-import promiseAllSettled from './promiseAllSettled';
```
luacmartins, line 4: `promiseAllSettled` seems to be unused in the code base now. I think we should remove it as well.

tgolen:
Updated

luacmartins (approved):
LGTM

luacmartins:
Can we add a link to an issue or comment so we can track where these changes are coming from?

PR #10692, opened 2022-08-30
Refactor of WorkspaceReimburseView to use new API

yuwenmemon:
@arosiclair please review
cc @ctkochan22 

### Details
Refactor WorkspaceReimburseView to use the new API

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/217856

### Tests/QA
1. Open the Reimburse Expenses page by going to Settings -> [Workspace] -> Reimburse expenses
2. Make sure that the Mileage rate displays the value you have in your policy
3. Make sure that when you open the selector it shows the mileage setting you have in your policy.
4. Edit the milage rate, make sure that the value saves and appears the same when you exit, refresh the page, and re-enter the Reimburse expenses view again
5. Do the same for the mi/km selector


arosiclair:
> Have we considered the offline behavior for this? Currently as set up if we're offline we just show a 0.000 rate

This is for when you have not previously viewed the custom unit while online correct? If so, that's expected since we don't have any rate saved in Onyx to display. If you did previously view the custom unit while online, then the same value should display after going offline.

yuwenmemon:
> This is for when you have not previously viewed the custom unit while online correct? If so, that's expected since we don't have any rate saved in Onyx to display. If you did previously view the custom unit while online, then the same value should display after going offline.

Yep, that's correct - and sounds good! This is ready to be reviewed

src/pages/workspace/reimburse/WorkspaceReimburseView.js
```diff
@@ -84,6 +88,35 @@ class WorkspaceReimburseView extends React.Component {
...
+
+            this.setState({
+                unitID: lodashGet(distanceCustomUnit, 'customUnitID', ''),
+                unitName: lodashGet(distanceCustomUnit, 'name', ''),
+                unitValue: lodashGet(distanceCustomUnit, 'attributes.unit', 'mi'),
+                rateID: lodashGet(distanceCustomUnit, 'rates[0].customUnitRateID', ''),
+                rateName: lodashGet(distanceCustomUnit, 'rates[0].name', ''),
+                rateValue: this.getRateDisplayValue(lodashGet(distanceCustomUnit, 'rates[0].rate', 0) / 100),
+            });
+        }
```
marcochavezf, line 110: NAB: Why don't we use the props directly? I think it will be a bit hard to keep props and state in sync if we add more props in the future, no?

marcochavezf:
I noticed that after refreshing the browser, the unit value changes quickly to the previous value (Kilometers -> Miles -> Kilometers), I wonder if it would be related to the props - state sync

Uploading Screen Recording 2022-08-31 at 12.47.39.mov…

src/pages/workspace/reimburse/WorkspaceReimburseView.js
yuwenmemon: Yeah, I'm trying to have as little impact on this aspect of the page as possible and simply support the new API command. It's been worked on by a number of contributors based on what I've seen. I agree it could use some improvement in how it's written but I feel like that would be the domain of a follow-up issue.

yuwenmemon:
Yeah, cc @arosiclair but I would guess that's by design since the doc states we use `componentDidMount` to make the call and load the rate. So it's most likely coming from there. Is this how we intended it?

arosiclair (changes requested):
It looks like the `componentDidUpdate` logic is conflicting with `getDerivedStateFromProps` since they're both trying to set the same state. I tested removing `getDerivedStateFromProps` and it fixed the kilometers-miles flickering on refresh so I think we should just remove it

yuwenmemon:
Sounds good, and added!

marcochavezf (approved):

ctkochan22 (approved):

arosiclair (approved):

mvtglobally:
Tester is facing a different issue now when re-testing. the error doesn't disappear after appearing. We are collecting more info to see if it needs to be logged 

https://user-images.githubusercontent.com/43995119/188957435-b5c84014-9480-4d90-8978-f77943a19385.mp4

PR #10784, opened 2022-09-02
Optimize report ordering in the LHN

tgolen:
This is the first step in optimizing the ordering logic of the sidebar. It does these things:

1. Removes all state data from the SidebarLinks component, which removes the need for `setDerivedStateFromProps` and any other static methods
2. It makes the rendering so efficient that it doesn't really matter how many times it's re-renderes or how many items it has
3. It optimizes all the stuff happening in `OptionsListUtils` so that it's very efficient and only doing the minimum amount of logic in the fastest way

### Fixed Issues
Part of https://github.com/Expensify/App/issues/7948

### Tests
Everything is covered by unit tests, but in general you need to test:
1. Chats in the LHN show as draft, pinned, unread, and have outstanding IOUs and that changes those states updates the LHN appropriately
2. Switching view mode changes the views as expected


src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -298,8 +180,8 @@ class SidebarLinks extends React.Component {
                         {paddingBottom: StyleUtils.getSafeAreaMargins(this.props.insets).marginBottom},
                     ]}
                     sections={sections}
-                    focusedIndex={_.findIndex(this.state.orderedReports, (
-                        option => option.reportID === activeReportID
+                    focusedIndex={_.findIndex(this.orderedReports, (
+                        option => option.reportID.toString() === this.props.currentlyViewedReportID.toString()
```
roryabraham, line 184: Why cast these to string?

src/pages/home/report/ReportActionCompose.js
```diff
@@ -57,7 +57,7 @@ const propTypes = {
     comment: PropTypes.string,
 
     /** The ID of the report actions will be created for */
-    reportID: PropTypes.number.isRequired,
+    reportID: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
tgolen: I guess we need to get clarification on that. I thought they were supposed to be strings everywhere so that when they are upgraded to 64bit integers, they would work out-of-the-box with no other changes.

src/libs/ReportUtils.js
```diff
@@ -246,8 +246,11 @@ function isArchivedRoom(report) {
  * @returns {String}
  */
 function getPolicyName(report, policies) {
-    const defaultValue = report.oldPolicyName || Localize.translateLocal('workspace.common.unavailable');
-    return lodashGet(policies, [`${ONYXKEYS.COLLECTION.POLICY}${report.policyID}`, 'name'], defaultValue);
+    const policyName = (policies[`${ONYXKEYS.COLLECTION.POLICY}${report.policyID}`] && policies[`${ONYXKEYS.COLLECTION.POLICY}${report.policyID}`].name) || false;
```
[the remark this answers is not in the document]
luacmartins: That seems to work as intended:

Screen Shot 2022-09-05 at 12 05 40 PM

 I'd change false to an empty string to keep `policyName` as a string
```suggestion
    const policyName = (policies[`${ONYXKEYS.COLLECTION.POLICY}${report.policyID}`] && policies[`${ONYXKEYS.COLLECTION.POLICY}${report.policyID}`].name) || '';
```

src/pages/home/report/ReportActionCompose.js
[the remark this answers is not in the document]
luacmartins: I think we should already use strings in App, since as Tim said we will be moving them to 64bit numeric strings later on.

src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -84,173 +89,50 @@ const defaultProps = {
...
- * @returns {Array}
- */
-function getUnreadReports(reportsObject) {
-    const reports = _.values(reportsObject);
-    if (reports.length === 0) {
-        return [];
+class SidebarLinks extends React.Component {
+    constructor(props) {
+        super(props);
+        this.getRecentReportsOptionListItems = memoizeOne(this.getRecentReportsOptionListItems.bind(this));
```
luacmartins, line 95: I'm not sure if we should keep `memoizeOne` since it just memoizes the last result, which seems kind of useless.

src/libs/OptionsListUtils.js
```diff
@@ -407,17 +465,18 @@ function getOptions(reports, personalDetails, activeReportID, {
         const isChatRoom = ReportUtils.isChatRoom(report);
         const isDefaultRoom = ReportUtils.isDefaultRoom(report);
         const isPolicyExpenseChat = ReportUtils.isPolicyExpenseChat(report);
-        const logins = lodashGet(report, ['participants'], []);
+        const logins = report.participants || [];
```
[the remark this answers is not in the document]
roryabraham: That was my first thought, but what if you have `[{ // some real object }, null, { // some real object }]` ? No idea if that can actually happen in this case though:

image

src/libs/ReportUtils.js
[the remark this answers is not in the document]
tgolen: Yeah, I think the logic still works, even with the inclusion of parenthesis. Here is a slightly more detailed example:

image

FYI, if you don't include the parenthesis, then ESLint throws a warning.

I like the idea of keeping it a string!

src/pages/home/sidebar/SidebarLinks.js
marcaaron: Why do you think it's useless?

src/libs/OptionsListUtils.js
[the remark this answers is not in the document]
tgolen: Yeah, there are a few things here... `_.each` doesn't protect against null members in a collection, so it's possible for `report` to be null.

This particular line of code (and several lines down below it) need to have null protection. There are two solutions around this typically...

1. Use `_.compact` for arrays to clear out falsy values or use `_.pluck` for objects
2. Add an early return if `report` is null

I think I'm going to do number `2` and then remove all the report null checks

src/libs/OptionsListUtils.js
```diff
@@ -143,6 +142,27 @@ function getParticipantNames(personalDetailList) {
...
+ */
+function uniqFast(items) {
+    const seenItems = {};
+    const result = [];
+    let j = 0;
+    for (let i = 0; i < items.length; i++) {
+        const item = items[i];
+        if (seenItems[item] !== 1) {
+            seenItems[item] = 1;
+            result[j++] = item;
```
marcaaron, line 160: Wait it is slower to push an item into an array than do this? 🤯

src/libs/OptionsListUtils.js
```diff
@@ -154,28 +174,29 @@ function getParticipantNames(personalDetailList) {
...
-        });
+        for (let i = 0; i < personalDetailList.length; i++) {
+            const personalDetail = personalDetailList[i];
+            searchTerms = searchTerms.concat([personalDetail.displayName, personalDetail.login.replace(/\./g, '')]);
+        }
     }
     if (report) {
-        searchTerms.push(...reportName);
-        searchTerms.push(..._.map(reportName.split(','), name => name.trim()));
+        Array.prototype.push.apply(searchTerms, reportName.split(''));
```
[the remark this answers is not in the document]
marcaaron: I'm curious about the explanation and +1

src/libs/OptionsListUtils.js
```diff
@@ -217,80 +238,117 @@ function createOption(logins, personalDetails, report, reportActions = {}, {
...
-    if (isChatRoom || isPolicyExpenseChat) {
-        alternateText = (showChatPreviewLine && !forcePolicyNamePreview && lastMessageText)
-            ? lastMessageText
-            : subtitle;
+    if (report) {
+        result.isChatRoom = ReportUtils.isChatRoom(report);
+        result.isDefaultRoom = ReportUtils.isDefaultRoom(report);
+        result.isArchivedRoom = ReportUtils.isArchivedRoom(report);
+        result.isPolicyExpenseChat = ReportUtils.isPolicyExpenseChat(report);
+        result.shouldShowSubscript = result.isPolicyExpenseChat && !report.isOwnPolicyExpenseChat && !result.isArchivedRoom;
```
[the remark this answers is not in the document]
tgolen: It looks like it could maybe be renamed to `shouldShowSubscriptAvatar`, which is a small avatar that is below the main line of text?

src/libs/OptionsListUtils.js
```diff
@@ -143,6 +142,27 @@ function getParticipantNames(personalDetailList) {
...
 }
 
+/**
+ * A very optimized method to remove unique items from an array.
+ * Taken from https://stackoverflow.com/a/9229821/9114791
+ *
+ * @param {Array} items
+ * @returns {Array}
+ */
+function uniqFast(items) {
```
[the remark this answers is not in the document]
tgolen: I thought of this, but in the end, I wasn't really keen on making a utils file that only contained one method called from one spot. I'm not 100% sure how this utility could be used and broken in other contexts, and I'm afraid that putting it into a utility file would signal that we should use it everywhere.

src/libs/OptionsListUtils.js
```diff
@@ -154,28 +174,29 @@ function getParticipantNames(personalDetailList) {
...
 function getSearchText(report, reportName, personalDetailList, isChatRoomOrPolicyExpenseChat) {
-    const searchTerms = [];
+    let searchTerms = [];
 
     if (!isChatRoomOrPolicyExpenseChat) {
-        _.each(personalDetailList, (personalDetail) => {
-            searchTerms.push(personalDetail.displayName);
-            searchTerms.push(personalDetail.login.replace(/\./g, ''));
-        });
+        for (let i = 0; i < personalDetailList.length; i++) {
```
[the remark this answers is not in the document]
tgolen: I'll add a comment about it in the method docs, because I think it should also explain why spread operators aren't being used.

src/libs/OptionsListUtils.js
```diff
@@ -154,28 +174,29 @@ function getParticipantNames(personalDetailList) {
...
+            Array.prototype.push.apply(searchTerms, chatRoomSubtitle.split(','));
         } else {
-            searchTerms.push(...report.participants);
+            searchTerms = searchTerms.concat(report.participants);
         }
     }
 
-    return _.unique(searchTerms).join(' ');
+    const finalSearchTerms = uniqFast(searchTerms).join(' ');
+    return finalSearchTerms;
```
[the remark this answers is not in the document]
tgolen: Ah yes, I was doing some debugging there and didn't clean it up.

src/libs/OptionsListUtils.js
```diff
@@ -217,80 +242,117 @@ function createOption(logins, personalDetails, report, reportActions = {}, {
     showChatPreviewLine = false,
     forcePolicyNamePreview = false,
 }) {
-    const isChatRoom = ReportUtils.isChatRoom(report);
-    const isPolicyExpenseChat = ReportUtils.isPolicyExpenseChat(report);
+    const result = {
```
marcaaron, line 244: It's hard for me to tell what is style and what is an improvement. Is there some benefit to declaring the object with all the properties it might have instead of assigning them later?

src/libs/OptionsListUtils.js
```diff
@@ -217,80 +238,117 @@ function createOption(logins, personalDetails, report, reportActions = {}, {
...
+        isChatRoom: false,
+        isArchivedRoom: false,
+        shouldShowSubscript: false,
+        isPolicyExpenseChat: false,
+    };
+
     const personalDetailMap = getPersonalDetailsForLogins(logins, personalDetails);
     const personalDetailList = _.values(personalDetailMap);
-    const isArchivedRoom = ReportUtils.isArchivedRoom(report);
-    const isDefaultRoom = ReportUtils.isDefaultRoom(report);
```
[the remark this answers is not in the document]
tgolen: Yes! This came from me solving a merge conflict. `isDefaultRoom` didn't exist when I started refactoring so it got kind of left out.

src/pages/home/sidebar/SidebarLinks.js
tgolen: Yeah, I hear ya. In [my second PR](https://github.com/Expensify/App/pull/10800) (which is based off this one) I have removed this in favor of no memoization.

src/libs/OptionsListUtils.js
```diff
@@ -351,7 +413,7 @@ function isCurrentUser(userDetails) {
  *
  * @param {Object} reports
  * @param {Object} personalDetails
- * @param {Number} activeReportID
+ * @param {String} activeReportID
```
marcaaron, line 416: When did reportID become strings and are all reportID now strings in JS?

src/libs/OptionsListUtils.js
```diff
@@ -404,20 +466,24 @@ function getOptions(reports, personalDetails, activeReportID, {
 
     const allReportOptions = [];
     _.each(orderedReports, (report) => {
+        if (!report) {
+            return null;
```
marcaaron, line 470:
```suggestion
            return;
```

Should work too or was there a reason to return `null` to move to the next iteration?

src/libs/OptionsListUtils.js
```diff
@@ -404,20 +466,24 @@ function getOptions(reports, personalDetails, activeReportID, {
...
         const isChatRoom = ReportUtils.isChatRoom(report);
         const isDefaultRoom = ReportUtils.isDefaultRoom(report);
         const isPolicyExpenseChat = ReportUtils.isPolicyExpenseChat(report);
-        const logins = lodashGet(report, ['participants'], []);
+        const logins = report.participants || [];
 
         // Report data can sometimes be incomplete. If we have no logins or reportID then we will skip this entry.
         const shouldFilterNoParticipants = _.isEmpty(logins) && !isChatRoom && !isDefaultRoom && !isPolicyExpenseChat;
-        if (!report || !report.reportID || shouldFilterNoParticipants) {
+        if (!report.reportID || shouldFilterNoParticipants) {
```
marcaaron, line 479: Should we do this sooner on line 469? 

```js
if (!report || !report.reportID) {
    return;
}
```

src/pages/home/report/ReportActionCompose.js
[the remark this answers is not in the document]
marcaaron: I am confused about whether reportID are all strings currently? Has that change happened already?

src/pages/home/sidebar/SidebarLinks.js
marcaaron: +1

src/pages/home/sidebar/SidebarLinks.js
tgolen: My understanding is that it caches the results once for _each set of unique arguments_.

```
foo(1) - not cached
foo(2) - not cached
foo(1) - cached
foo(1) - not cached
```

So, I don't think it's totally useless. It just allows each set of arguments to be cached once, so it cuts performance by up to 50%.

tests/unit/LHNOrderTest.js
```diff
@@ -158,18 +158,37 @@ Onyx.init({
 });
 
 function getDefaultRenderedSidebarLinks() {
+    class ErrorBoundary extends React.Component {
+        // Error boundaries have to implement this method. It's for providing a fallback UI, but
+        // we don't need that for unit testing, so this is basically a no-op.
```
marcaaron, line 163: **question:** Why do we need the error boundary?

marcaaron (commented):
Just some comments for now. Looks like great changes!

src/pages/home/sidebar/SidebarLinks.js
marcaaron: >So if of the arguments change, which seems likely especially for activeReportID and reportActions, we would compute the results again.

Yes the point is to prevent re-computing them when they **do not change** because the returned result would be the same if the arguments have not changed. If they change then we want to recompute them.

src/libs/OptionsListUtils.js
tgolen: I think the general answer is "yes, it's faster", but there are differences between JS engines which make the testing somewhat unreliable. Generally, `push()` does some extra bits of processing from the base JS object that arrays extend from, whereas using an index approach bypassing that little bit of overhead.

src/libs/OptionsListUtils.js
tgolen: This is being discussed in Slack to make sure we're getting to the right conclusion: https://expensify.slack.com/archives/C02HWMSMZEC/p1662382397049269

Regardless, we're not treating them consistently everywhere yet, but we want to!

src/libs/OptionsListUtils.js
tgolen: Ah, no. There was no reason for the `null`.

src/libs/OptionsListUtils.js
tgolen: We probably could. I'm honestly a bit perplexed by this and the comment. Since reports are indexed in Onyx by their ID, I don't think it should ever be possible for a report to not have an ID. If it is possible, I think it is a bug that we should track down and fix. So, I'd be more willing to either remove this check entirely, or at least add a log if we detect there is no reportID.

src/pages/home/ReportScreen.js
```diff
@@ -107,8 +107,7 @@ const defaultProps = {
  * @returns {Number}
  */
 function getReportID(route) {
-    const params = route.params;
-    return Number.parseInt(params.reportID, 10);
+    return route.params.reportID.toString();
```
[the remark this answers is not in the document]
tgolen: Oh, I love that we can parse at the navigation layer.

tests/unit/LHNOrderTest.js
tgolen: Is that a rhetorical question or something you would like more details on in the comment? I only ask because we talked about this face-to-face and I thought you understood about the error boundary being added and Jest somethingsomething swallowing up errors.

src/pages/home/report/ReportActionCompose.js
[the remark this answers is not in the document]
tgolen: # Conclusion from Slack
ReportIDs should be strings everywhere in JS.

I am creating a separate issue: https://github.com/Expensify/Expensify/issues/227425 which will clean this up from the source and make it a string everywhere consistently. For now, this PR is going to be a little bit janky with adding a few `toString()` casts where they are needed.

tests/unit/LHNOrderTest.js
tgolen: I've added a big code comment to explain the `why` behind why this is necessary.

tgolen (commented):
Updated and ready for another review!

marcaaron:
Tested on iOS and Desktop. Seems to be working very well.

src/libs/OptionsListUtils.js
marcaaron: >Since reports are indexed in Onyx by their ID, I don't think it should ever be possible for a report to not have an ID. If it is possible, I think it is a bug that we should track down and fix.

I'm feeling cautious about removing that since it looks like we use the "report without a reportID" as a kind of "hook" to fetch the report. Which was the expected behavior at one point. But also looks like there's maybe some further cleanup to do here:

https://github.com/Expensify/App/blame/c9b41bff8b728073db85f38782b591a0fd4a9772/src/libs/actions/Report.js#L1135-L1137

We could solve this by sending the report and comment when a new message is added.

marcaaron: oh wait scratch everything I just said haha that code looks for `report.reportID && !reportName` (but we should still clean that up)

marcaaron: Still feeling cautious and agree we should try to figure out if/why a report object can be merged with no `reportID` (and a log would help)

tests/unit/LHNOrderTest.js
marcaaron: No, it wasn't rhetorical. I was curious why it is needed (did not recall the previous conversation) and appreciate the code comment so that others with the same question will have a better idea.

**thought:** Feels like this could be moved into some kind of `utils` so we can re-use this in other tests. But NAB, since this is one of the few tests we have that renders a component.

src/libs/OptionsListUtils.js
marcaaron: That sounds intuitive to me. Mostly asking as I'm curious which of these changes had the biggest impact on performance (for my own understanding).

marcaaron (approved):
Changes are looking good to me. I'd be curious to see some before and after snapshots to get a better idea of which changes impacted performance the most, but that is NAB. 

I also think it would be good to distill any of the things we learned here and add them to the `PERFORMANCE.md` doc with some tips on when some of the strategies used in this PR would be appropriate to use.

tgolen:
Thanks, Marc! I'm trying to test on Android... and it's, totally unusable. I get the feeling though that it's not related to my branch and only related to the sheer number of reports. For example, I'm seeing errors like this:

```
 DEBUG  [info] [Onyx] Handled error: Error: Unexpected AsyncStorage error: too many SQL variables (code 1 SQLITE_ERROR): , while compiling: SELECT * FROM Storage WHERE `key` IN (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?) - ""
```

tgolen:
I also see these logs, which indicate that while it takes ~220ms on web to sort and filter the reports, it's taking almost 30 seconds on android.

```
Timing:expensify.cash.sidebar_links_filter_reports 31150
```

tgolen:
Hm, OK. Same thing is happening on `main`, so I don't think this branch is causing it (or fixing it unfortunately). I think maybe when I get the second PR done, that's when it might start working 🤞

src/libs/OptionsListUtils.js
tgolen: About the best I can offer in that case then is to add a log somewhere to detect if a report ID is missing. I don't think this code is the right place for it though. I would add it somewhere in Onyx.update() probably? Something like:

```
if (key starts with "report_" and value has no reportID) {
    Log to the server
    return early
}
```

tgolen: but that doesn't really work because of doing something like `Onyx.merge('report_123', {reportName: 'blah'})`

tgolen: I kind of just feel here like we need to trust that we would never have a report without a reportID. And if it happens, we will find out about it in buggy behavior and crashes.

tgolen:
As I was doing some performance testing this morning, to get you snapshots of main vs. this branch, it's pretty interesting.

This branch actually doesn't change much the rendering time of the LHN. BUT it achieves the same performance as main without having all of the memoized methods and the `serDerivedStateFromProps()`. 

Where we're really going to see the performance gains is in https://github.com/Expensify/App/pull/10800

luacmartins (approved):

tgolen:
I'm going to self merge since it's got a couple of reviews. This will let me move onto my next step of optimizing. The two failing tests are OK because there are no contributors involved with this one.

francoisl:
🚀 [Cherry-picked](https://github.com/Expensify/App/actions/runs/3055939937) to staging by @francoisl in version: 1.2.0-3 🚀

 platform | result 
 ---|--- 
🤖 android 🤖|success ✅ 
🖥 desktop 🖥|success ✅
🍎 iOS 🍎| success ✅ 
🕸 web 🕸|success ✅

@Expensify/applauseleads please QA this PR and check it off on the [deploy checklist](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3AStagingDeployCash) if it passes.


---

Note: we had a conflict on the `staging` branch which ended up needing to be manually fixed, so this PR doesn't automatically appear in the diff between production and staging - so I'm manually adding the note that it was CPed to staging here.

Issue #10799, opened 2022-09-02
Android - Chat - Deleted message is displayed for brief moment after back online


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Issue found when executing PR** https://github.com/Expensify/App/pull/10780

## Action Performed:
1. Launch the app
2. Log in with expensifail account
3. Select any user
4. Send message
5. .Go offline and delete that message
6. The message should not disappear but instead appear with a strike-through
7. Go online, the message should disappear

## Expected Result:
Deleted message should disappear after back online

## Actual Result:
Deleted message is displayed for brief moment after back online

## Workaround:
Unknown

## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.1.96.5

**Reproducible in staging?:** Yes

**Reproducible in production?:** No -new feature

**Email or phone of affected tester (no customers):** any

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/93399543/188166297-95d68c2c-091e-48f4-8c48-2806e03e7ba5.mp4

**Expensify/Expensify Issue URL:** 

**Issue reported by:** Applause - Internal Team

**Slack conversation:** 

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

PauloGasparSv:
Hey @tgolen, I think we can quickly fix this by using the reportAction's  `pendingAction` field inside the reportActionItem or reportActionItemFragment or maybe OfflineWithFeedback.

We can check if the reportAction is a comment and something like `pendingAction === styles.offlineFeedback.deleted && !props.network.isOffline` and render the text as an empty string in that case!

I can take this if that's ok or it can be an External issue too!

tgolen:
@PauloGasparSv Feel free to gab and work on it

PauloGasparSv:
Same issue from here: https://github.com/Expensify/App/issues/10753

Issue #10753, opened 2022-09-01
Web- Conversation - Attachment removal is not smooth, still appears after deleted


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Go to URL https://staging.new.expensify.com/
and login with any account
2. Go to any conversation
3. Add an attachment
4. Delete an attachment

Attachment used for testing:
![1f98b (1)](https://user-images.githubusercontent.com/93399543/187979457-2e0dec0c-2740-46b4-84c6-d35b63555acc.png)
![1f9d0 (1)](https://user-images.githubusercontent.com/93399543/187979459-9908fd4b-c7a5-480a-be40-c13e80024b48.png)

## Expected Result:
The attachment should be deleted immediately

## Actual Result:
 Attachment removal is not smooth, still appears after deleted

## Workaround:
Unknown

## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.1.96.0

**Reproducible in staging?:** Yes
 
**Reproducible in production?:** No

**Email or phone of affected tester (no customers):** any

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/93399543/187960816-47d9662f-0ef3-4a8f-9e26-e062c256275f.mp4

**Expensify/Expensify Issue URL:** 

**Issue reported by:** Applause - Internal Team

**Slack conversation:** 

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

kbecciv:
QA team is noticed the same behavior for deleting the comment, please attach video

https://user-images.githubusercontent.com/93399543/187982810-7e1036dc-64d9-489d-845d-e55531e14898.mp4

Julesssss:
Removing the label as this is not a critical enough issue to hold the release any further.

PR #10929, opened 2022-09-09
Automated tests for Unread Indicators feature and final polish

marcaaron:
### Details

There are several changes introduced in this PR, but most of them exist to enable UI testing for the entire app. 

- Several modules needed to be mocked
- Prevent editing or deleting of report actions that are `pendingAction === 'add'` (this is temporary but I noticed this feature is pretty broken so disabled it for now until we are totally migrated to using `reportActionID` for `reportAction_*` values keys)
- Report `sequenceNumber` were being set to `clientID` which made it impossible to mark a "pending" comment as unread. Instead we use a guessed `sequenceNumber` for this. I suspect this may need to change in the future or depend on `reportActionID` as well.
- When deleting a message we were not optimistically adding in the `type` field (Fixed)
- Strike-through for deleted messages also was not working correctly as the message was entirely unset and when going offline there was no comment to show with the strike-through. The message is entirely lost when we do this so I moved it to a `previousMessage` field that we can use as a fallback.

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/228076
$ https://github.com/Expensify/App/issues/10799
$ https://github.com/Expensify/App/issues/10753

### Tests

- Automated tests were added, but these testing flows laid out in the **Unread Indicators doc** should be performed.

**Testing Unread LHN Status, “New Messages” badge, and New Line Indicator**

1. Start a chat between User A and User B.
1. Ensure User A is not looking at the chat.
1. Send a message as User B
1. Verify the LHN row shows in bold for User A to indicate unread messages
1. Visit the chat as User A and verify the LHN row no longer shows in bold
1. Verify there is a “----- New” line above the message sent by User B
1. Scroll up and verify that the “New Messages” button appears floating above the content.
1. Verify that tapping this button will scroll you to the bottom of the chat but not clear the “---- New” line marker.
1. Mark a comment as “unread”
1. Verify the “---- New” line marker appears
1. Mark as unread comments above and below the existing new marker
1. Verify the marker appears in the correct position each time
1. While the new marker is active leave a comment as User A
1. Verify the new marker and new messages badge both disappear
1. Mark another comment as unread.
1. Navigate away from the chat
1. Navigate back to the chat
1. Verify the chat in the LHN shows as read
1. Verify the New Line marker and new messages badge is still present
1. Navigate away one more time and back
1. Verify the new marker and new messages badge both disappear
1. Mark another comment as unread.
1. Navigate away from the chat
1. Navigate back to the chat
1. Minimize the app and reopen it
1. Verify the new marker and new message badge both disappear
1. Switch to a mobile view (small screen) and mark a message as unread
1. Tap the header to reveal the LHN
1. Verify the chat in the LHN shows as “unread”
1. Tap the chat
1. Verify the new marker and badge are present
1. Tap the header to reveal the LHN
1. Verify the chat in the LHN shows as “read”
1. Tap the chat
1. Verify the new marker and badge are no longer present.
1. Scroll up in the chat so that last comment is out of view
1. Leave a comment as User B
1. Verify the “new messages badge” shows for User A
1. Scroll down and verify the “new line marker” shows for User A in the correct position.
1. Scroll back up and leave another comment as User B
1. Scroll down as User A
1. Verify the new marker is in the same position as it was before

**Testing Deleted Comment Edge Case**

1. Start a chat between User A and User B.
1. Ensure User A is not looking at the chat.
1. Send a message as User B.
1. Navigate to the chat as User A
1. Mark the most recent message as “unread”
1. As User B delete that message
1. Verify that the report no longer shows as “unread” for User A and that the new marker has disappeared.

    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop
    - [x] I verified any copy / text that was added to the app is correct English and approved by marketing by tagging the marketing team on the original GH to get the correct copy.
    - [x] Any functional components have the `displayName` property

<details>

</summary>

    - [ ] iOS / native
    - [ ] Android / native
    - [ ] iOS / Safari
    - [ ] Android / Chrome
    - [ ] MacOS / Chrome
    - [ ] MacOS / Desktop

</details>

### QA Steps

Same as Tests above


marcaaron:
This one is testing well for me locally, but going to wait until the Auth PR for `reportActionCount` is deployed before taking off draft (as I am testing against that branch locally).

jest/setup.js
```diff
@@ -1,15 +1,143 @@
...
-jest.mock('../src/libs/Notification/PushNotification', () => ({
-    // There is no need for a jest.fn() since we don't need to make assertions against it.
-    register: () => {},
-    deregister: () => {},
+jest.mock('react-native-blob-util', () => ({}));
+
+jest.mock('react-native-reanimated', () => {
+    const Reanimated = require('react-native-reanimated/mock');
+
+    // The mock for `call` immediately calls the callback which is incorrect
```
tgolen, line 16: Can you please explain why it's incorrect?

jest/setup.js
```diff
@@ -1,15 +1,143 @@
...
+
+// <App> uses <ErrorBoundary> and we need to mock the imported crashlytics module
+// due to an error that happens otherwise https://github.com/invertase/react-native-firebase/issues/2475
+jest.mock('@react-native-firebase/crashlytics', () => () => ({
+    log: jest.fn(),
+    recordError: jest.fn(),
 }));
 
-jest.mock('react-native-blob-util', () => ({}));
+jest.mock('../src/libs/BootSplash', () => ({
```
tgolen, line 29: Can you please add a comment for why this is mocked?

jest/setup.js
```diff
@@ -1,15 +1,143 @@
...
+    hide: jest.fn(),
+    getVisibilityStatus: jest.fn().mockResolvedValue('hidden'),
+}));
+
+jest.mock('../src/libs/Notification/LocalNotification', () => ({
```
tgolen, line 34: ditto

jest/setup.js
```diff
@@ -1,15 +1,143 @@
...
+    const imageFilenames = fs.readdirSync(path.resolve(__dirname, `../assets/${imagePath}/`));
+    // eslint-disable-next-line rulesdir/prefer-early-return
+    _.each(imageFilenames, (fileName) => {
+        if (/\.svg/.test(fileName)) {
+            jest.mock(`../assets/${imagePath}/${fileName}`, () => () => '');
+        }
+    });
+}
+
+// Mock all images so that Icons and other assets cannot break tests
```
tgolen, line 51: Can you please explain how assets break the tests?

jest/setup.js
```diff
@@ -1,15 +1,143 @@
...
+mockImages('images');
+mockImages('images/avatars');
+mockImages('images/bankicons');
+mockImages('images/product-illustrations');
+jest.mock('../src/components/Icon/Expensicons', () => ({
```
tgolen, line 56: Something that has never been clear to me... Why do we mock somethings in this file, but other things that get mocked live in the `__mock__` folder?

Is there a chance here for us to be consistent?

src/components/DisplayNames/index.native.js
```diff
@@ -4,7 +4,7 @@ import Text from '../Text';
 
 // As we don't have to show tooltips of the Native platform so we simply render the full display names list.
 const DisplayNames = props => (
-    <Text style={props.textStyles} numberOfLines={props.numberOfLines}>
+    <Text testID={props.testID} style={props.textStyles} numberOfLines={props.numberOfLines}>
```
tgolen, line 7: This highlights something that I think might already be on the "iffy" side. My concern is having a bunch of testIDs everywhere in the code that end up making a mess eventually.

In my tests with the LHN, I ended up using `accessabilityHint` as a way to query for some of the components I was looking for, and I felt like that it was at least making a small contribution toward accessibility.

Do you have any initial thoughts on how we can try to get a handle on this problem early (querying the rendered tree for specific components)?

src/libs/ReportActionsUtils.js
```diff
@@ -76,7 +97,55 @@ function isConsecutiveActionMadeByPreviousActor(reportActions, actionIndex) {
...
+ * @return {String}
+ */
+function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
+    const parser = new ExpensiMark();
+    const actions = _.toArray(lodashMerge({}, allReportActions[reportID], actionsToMerge));
+    const sortedActions = _.sortBy(actions, 'sequenceNumber');
+    const lastMessageIndex = _.findLastIndex(sortedActions, action => (
+        !isDeletedAction(action)
+    ));
+    const htmlText = lodashGet(actions, [lastMessageIndex, 'message', 0, 'html'], '');
```
tgolen, line 113: NAB, but I think this also works and is a little cleaner (not tested):
```suggestion
    const htmlText = lodashGet(actions, `[${lastMessageIndex}].message[0].html`, '');
```
From the docs:
image

src/libs/ReportActionsUtils.js
```diff
@@ -76,7 +97,55 @@ function isConsecutiveActionMadeByPreviousActor(reportActions, actionIndex) {
...
+ * Get the message text for the last action that was not deleted
+ * @param {Number} reportID
+ * @param {Object} [actionsToMerge]
+ * @return {String}
+ */
+function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
+    const parser = new ExpensiMark();
+    const actions = _.toArray(lodashMerge({}, allReportActions[reportID], actionsToMerge));
+    const sortedActions = _.sortBy(actions, 'sequenceNumber');
+    const lastMessageIndex = _.findLastIndex(sortedActions, action => (
```
tgolen, line 110: Would it ever be possible for this to be undefined? Just wondering if we should maybe do an early return and skip the parsing. But even if it is possible, it probably happens super rarely.

src/libs/ReportActionsUtils.js
```diff
@@ -76,7 +97,55 @@ function isConsecutiveActionMadeByPreviousActor(reportActions, actionIndex) {
...
+}
+
+/**
+ * @param {Number} reportID
+ * @param {Object} [actionsToMerge]
+ * @param {Number} deletedSequenceNumber
+ * @param {Number} lastReadSequenceNumber
+ * @return {String}
+ */
+function getNewLastReadSequenceNumberForDeletedAction(reportID, actionsToMerge = {}, deletedSequenceNumber, lastReadSequenceNumber) {
```
tgolen, line 125: What does "New" infer in the name of this message?

src/libs/ReportActionsUtils.js
```diff
@@ -76,7 +97,55 @@ function isConsecutiveActionMadeByPreviousActor(reportActions, actionIndex) {
...
+    // If the action we are deleting is unread then just return the current last read sequence number
+    if (deletedSequenceNumber > lastReadSequenceNumber) {
+        return lastReadSequenceNumber;
+    }
+
+    // Otherwise, we must find the first previous index of an action that is not deleted and less than the lastReadSequenceNumber
+    const actions = _.toArray(lodashMerge({}, allReportActions[reportID], actionsToMerge));
```
tgolen, line 132: NAB, we usually use `_.chain()` for these. It makes it more readable, but at the cost of debugging.

src/libs/ReportActionsUtils.js
```diff
@@ -76,7 +97,55 @@ function isConsecutiveActionMadeByPreviousActor(reportActions, actionIndex) {
...
+    const lastMessageIndex = _.findLastIndex(sortedActions, action => (
+        !isDeletedAction(action) && action.sequenceNumber <= lastReadSequenceNumber
+    ));
+
+    // It's possible we won't find any and in that case the last read should be reset
+    if (lastMessageIndex < 0) {
+        return 0;
+    }
+
+    return actions[lastMessageIndex].sequenceNumber;
```
tgolen, line 143: At this point, is the code sure that `actions[lastMessageIndex]` will always exist? It kind of looks like that case is covered in the `return 0`. Is that right?

src/pages/home/report/ReportActionsList.js
```diff
@@ -106,14 +106,14 @@ class ReportActionsList extends React.Component {
     }
 
     /**
-     * Create a unique key for Each Action in the FlatList.
-     * We use the reportActionId that is a string representation of a random 64-bit int, which should be
-     * random enought to avoid colisions
+     * Create a unique key for each action in the FlatList.
+     * We use the reportActionID that is a string representation of a random 64-bit int, which should be
```
tgolen, line 110: This comment is only giving a partial explanation. Why is it also using clientID and sequenceNumber? Why isn't the reportActionID enough?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+import fontWeightBold from '../../src/styles/fontWeight/bold';
+import * as AppActions from '../../src/libs/actions/App';
+import ReportHeaderViewBackButton from '../../src/pages/home/ReportHeaderViewBackButton';
+import ReportActionsView from '../../src/pages/home/report/ReportActionsView';
+import * as NumberUtils from '../../src/libs/NumberUtils';
+import LocalNotification from '../../src/libs/Notification/LocalNotification';
+import * as Report from '../../src/libs/actions/Report';
+
+beforeAll(() => {
+    global.fetch = TestHelper.getGlobalFetchMock();
```
tgolen, line 23: Please add a comment to explain why this is being mocked. Should this go with all our other mocks in the setup file?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+
+    // We need a bit more time for this test in some places
```
tgolen, line 25: Can you please enhance the comment to explain which tests? Why not just increase the timeout in the tests that are necessary?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+    return badge.props.style.transform[0].translateY === 10;
+}
+
+/**
+ * @param {RenderAPI} renderedApp
+ * @return {Promise}
+ */
+async function navigateToSidebar(renderedApp) {
+    const reportHeader = renderedApp.getByTestId('report-header');
+    const reportHeaderBackButton = reportHeader.findByType(ReportHeaderViewBackButton);
```
tgolen, line 73: A lot of the reading I did gave me the impression that querying by type was a bad thing and it was deprecated. Did you ever come across any of that? Also, why use `findBy*` here instead of `getBy`?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+/**
+ * @param {RenderAPI} renderedApp
+ * @return {Boolean}
+ */
+function isDrawerOpen(renderedApp) {
+    const reportScreen = renderedApp.getByTestId('report-screen');
+    return reportScreen.findByType(ReportActionsView).props.isDrawerOpen;
+}
+
+test('Unread LHN Status, “New Messages” badge, and New Line Indicator', async () => {
```
tgolen, line 98: Using `describe()` and `it()` will make it consistent with our other tests

tgolen: Do we have a policy written down which says it's OK to use async/await in unit tests?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+    const USER_C_ACCOUNT_ID = 3;
+    const USER_C_EMAIL = 'user_c@test.com';
+
+    // Render the App and sign in as a test user.
+    const renderedApp = render(<App />);
+
+    // Note: act() is necessary since react-navigation's NavigationContainer has an internal state update that will throw some
+    // warnings related to async code. See: https://callstack.github.io/react-native-testing-library/docs/understanding-act/#asynchronous-act
+    await act(async () => {
+        await waitForPromisesToResolve();
```
tgolen, line 113: Why are there two of these?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+        await waitForPromisesToResolve();
+    });
+
+    const loginForm = renderedApp.getByTestId('login-form');
+    expect(loginForm).toBeTruthy();
```
tgolen, line 118: This is a weird test, since if there is no `login-form`, an error will get thrown. That basically means this assertion is not doing anything. I find it's better to use `queryBy*` and then do an assertion for the number of items returned.

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+
+    TestHelper.signInWithTestUser(USER_A_ACCOUNT_ID, USER_A_EMAIL, undefined, undefined, 'A');
```
tgolen, line 120: `TestHelper.signInWithTestUser` already returns a promise, so why not just add the `await` here and drop the `waitForPromisesToResolve();` on the next line?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+
+    const MOMENT_TEN_MINUTES_AGO = moment().subtract(10, 'minutes');
+
+    // Simulate setting an unread report and personal details
+    Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${REPORT_ID}`, {
+        reportID: REPORT_ID,
+        reportName: 'Chat Report',
+        maxSequenceNumber: 9,
+        lastReadSequenceNumber: 1,
+        lastMessageTimestamp: MOMENT_TEN_MINUTES_AGO.utc(), // Ten minutes ago
```
tgolen, line 131:
```suggestion
        lastMessageTimestamp: MOMENT_TEN_MINUTES_AGO.utc(),
```

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+        5: TestHelper.buildTestReportComment(USER_B_EMAIL, 5, MOMENT_TEN_MINUTES_AGO.add(50, 'seconds').unix()),
+        6: TestHelper.buildTestReportComment(USER_B_EMAIL, 6, MOMENT_TEN_MINUTES_AGO.add(60, 'seconds').unix()),
+        7: TestHelper.buildTestReportComment(USER_B_EMAIL, 7, MOMENT_TEN_MINUTES_AGO.add(70, 'seconds').unix()),
+        8: TestHelper.buildTestReportComment(USER_B_EMAIL, 8, MOMENT_TEN_MINUTES_AGO.add(80, 'seconds').unix()),
+        9: TestHelper.buildTestReportComment(USER_B_EMAIL, 9, MOMENT_TEN_MINUTES_AGO.add(90, 'seconds').unix()),
+    });
+    Onyx.merge(ONYXKEYS.PERSONAL_DETAILS, {
+        [USER_B_EMAIL]: TestHelper.buildPersonalDetails(USER_B_EMAIL, USER_B_ACCOUNT_ID, 'B'),
+    });
+    await waitForPromisesToResolve();
```
tgolen, line 156: Again here, adding `await` to each of the `merge()` methods should allow this `waitForPromisesToResolve()` to be removed.

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+
+    // Verify no notifications are created for these older messages
+    expect(LocalNotification.showCommentNotification.mock.calls.length).toBe(0);
+
+    // Verify the sidebar links are rendered
+    const sidebarLinks = renderedApp.getByTestId('sidebar-links');
```
tgolen, line 162: Similar comment as above

tests/ui/UnreadIndicatorsTest.js
tgolen: OK, after seeing how long the test is below, I can see why this might be necessary. What I would rather see though is have the test broken up into smaller testable chunks. It's basically a huge monolithic test right now and I think that splitting it up into smaller chunks will make it easier to reason about what the test is doing.

tgolen (changes requested):

src/pages/home/report/ReportActionsList.js
```diff
@@ -106,14 +106,14 @@ class ReportActionsList extends React.Component {
...
+     * random enough to avoid collisions
      * @param {Object} item
```
marcochavezf, line 112: NAB:
```suggestion
     * @param {Object} item
     * @param {Object} item.action
```

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+function isNewMessagesBadgeVisible(renderedApp) {
+    const badge = renderedApp.getByTestId('new-messages-badge');
+    return badge.props.style.transform[0].translateY === 10;
+}
+
+/**
+ * @param {RenderAPI} renderedApp
+ * @return {Promise}
+ */
+async function navigateToSidebar(renderedApp) {
```
marcochavezf, line 71: `await` is not used inside this function so I think we can remove the `async` here

```suggestion
function navigateToSidebar(renderedApp) {
```

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * @param {RenderAPI} renderedApp
+ * @param {Number} index
+ * @return {Promise}
+ */
+async function navigateToSidebarOption(renderedApp, index) {
+    const optionRows = renderedApp.getAllByTestId('option-row');
+    fireEvent(optionRows[index], 'press');
+    await waitForPromisesToResolve();
```
marcochavezf, line 86:
```suggestion
function navigateToSidebarOption(renderedApp, index) {
    const optionRows = renderedApp.getAllByTestId('option-row');
    fireEvent(optionRows[index], 'press');
    return waitForPromisesToResolve();
```

tests/utils/TestHelper.js
```diff
@@ -5,18 +5,43 @@ import HttpUtils from '../../src/libs/HttpUtils';
 import ONYXKEYS from '../../src/ONYXKEYS';
 import waitForPromisesToResolve from './waitForPromisesToResolve';
 import * as ReportUtils from '../../src/libs/ReportUtils';
+import * as NumberUtils from '../../src/libs/NumberUtils';
+
+/**
+ * @param {String} login
+ * @param {Number} accountID
+ * @param {String} [firstName]
+ * @returns {Promise}
```
marcochavezf, line 14: We're returning an object instead of a promise here
```suggestion
 * @returns {Object}
```

marcochavezf (dismissed):
Nice! 🔥 Just a few comments

tests/ui/UnreadIndicatorsTest.js
marcaaron: I am still working on my expertise with this library, but looks like you're correct that `getByType()` has been deprecated. AFAICT `findBy` queries are not. I didn't come across anything other than that the [usage is discouraged](https://callstack.github.io/react-native-testing-library/docs/migration-v2/#removed-functions).

Unsure if there's any other implied issue with using `findBy*` or querying by type - let me know if there's something specific you read! 

https://callstack.github.io/react-native-testing-library/docs/api-queries/#findby

Then again, maybe this ties into your concern about consistency about the kinds of selectors we should use. IIRC I ended up going with "type" here because of some issues triggering the "press" event, but maybe I can clean this up with the `accessibilityHint` trick. Gonna give that a try.

tests/ui/UnreadIndicatorsTest.js
marcaaron: >Do we have a policy written down which says it's OK to use async/await in unit tests?

I can add something. 99% sure these these tests will not work without it and if they do will be much less readable 😄

tests/ui/UnreadIndicatorsTest.js
marcaaron: 🤦 because two is better than one? :trollface: (it's not needed)

tests/ui/UnreadIndicatorsTest.js
marcaaron: >since if there is no login-form, an error will get thrown

I was thinking that "throwing an error and breaking the test" would be the "anything" that we are doing. 😄  

But I like your idea much better.

tests/ui/UnreadIndicatorsTest.js
marcaaron: Ah in this case, I think we are not just waiting for `Onyx.merge()` but also some fun stuff that "eventually happens" after the merge.

jest/setup.js
marcaaron: Left an improved comment here.

jest/setup.js
marcaaron: Ah ha! This is actually copy/pasta [from the docs](https://reactnavigation.org/docs/testing/#mocking-native-modules). So I'd be inclined to either remove this or add a `- thus speaketh the JS Gods` here.

tests/ui/UnreadIndicatorsTest.js
marcaaron: Created a follow up for this here: https://github.com/Expensify/Expensify/issues/228440

jest/setup.js
marcaaron: Touched on in this follow-up issue: https://github.com/Expensify/Expensify/issues/228440 so we can discuss there and make this more standardized.

src/components/DisplayNames/index.native.js
marcaaron: Managed to replace 99% of these.

tests/ui/UnreadIndicatorsTest.js
marcaaron: I got rid of this usage. There's one other spot where I left in the access by type, but I think I can get rid of it as well. Since the thing I want to know is whether the "drawer is open" I _think_ we could give an `accessibilityElementsHidden` property to the sidebar and check that.

marcaaron: It worked!

jest/setup.js
marcaaron: Just removed this and linked to the docs. I don't have a satisfying explanation other than "react navigation made me do it". 🧙

src/components/DisplayNames/index.native.js
tgolen: Cool, thanks! I'll give it a look.

src/libs/ReportActionsUtils.js
tgolen: Ah, if that's your preference, then I think it's fine. Both is acceptable!

src/pages/home/report/ReportActionsList.js
tgolen: Haha, OK. I'm not sure I followed all of that, but I can appreciate if this is in an intermediary state until the full refactoring is complete.

tests/ui/UnreadIndicatorsTest.js
tgolen: Yeah, it's weird that `findByType` or `queryByType` wouldn't also be unsafe. Very strange. You found the same stuff that I did, but I don't think it registered to me that find and query by type were OK.

Great that the accessibility stuff worked out!

tests/ui/UnreadIndicatorsTest.js
tgolen: Hm, maybe this is the cause to some of my test failures that I'm seeing on my new sidebarlinks PR. I need to explore it further.

tests/ui/UnreadIndicatorsTest.js
marcaaron: Haha touché and very understandable. I'm also seeing a different kind of value in just following the doc examples and using the testing library as it was intended to be used. Most if not all examples I've come across while learning the testing library use async/await. So while I understand there is value in consistency I don't completely love that we will chose to not follow those examples. It feels counter-intuitive to learn something one way and then take an extra step to decompile it all into something else.

tests/ui/UnreadIndicatorsTest.js
marcaaron: Broke it up into separate independent tests and you were right! The extra time for the tests was able to be removed.

tests/ui/UnreadIndicatorsTest.js
marcaaron: After looking at the tests for a bit I was able to make this 99% async/await free 🎉 

Unfortunately, there is still a single function that is using async/await, but I moved it out of the test file and put some disclaimers around it. From what I can gather, there is some magic happening with the [`act()` method](https://reactjs.org/docs/test-utils.html#act) that waits for state updates to occur in React components, but for some reason I cannot get this one warning to go away without it.

>Warning: You called act(async () => ...) without await. This could lead to unexpected testing behaviour, interleaving multiple act calls and mixing their scopes. You should - await act(async () => ...);

It _only_ goes away if I use async/await. Curious if you have a better understanding than me about how the `act()` method works and whether you'd be comfortable with us using async/await for this one case?

marcaaron:
Reworked the tests here a bunch based on @tgolen's feedback. Feeling pretty good about the changes. The one thing I'm not 100% sure on is the usage of `act()`. When I tried to refactor it to not use async/await it stopped working and I saw errors that seemed to suggest we _had_ to use async/await. Here's what I've got for an alternative...

https://github.com/Expensify/App/blob/7b76284a011aa08a4ad24aab7b867eaed725b395/tests/utils/waitForPromisesToResolveWithAct.js#L1-L21

PauloGasparSv (dismissed):
My dev env is really buggy (not sending notifications correctly, duplicating messages, only working correctly while debugging) this past week so for now I'll approve the code for the changes I've read and I'll leave the tests for the other reviewers.

src/libs/actions/Report.js
```diff
@@ -1159,26 +1169,32 @@ Onyx.connect({
...
+    const deletedMessage = [{
+        type: 'COMMENT',
+        html: '',
+        text: '',
+        isEdited: true,
+    }];
     const optimisticReportActions = {
         [sequenceNumber]: {
             pendingAction: CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE,
+            previousMessage: reportAction.message,
```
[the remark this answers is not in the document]
marcaaron: Mostly the change to actually delete the actions optimistically instead of leave them in (as described [here](https://github.com/Expensify/App/blob/b4b12addac16a3ef3753273e25c3e58379715a76/src/libs/actions/Report.js#L1163-L1164)) makes more sense to my brain. 

I didn't love that when we are optimistically deleting an action it wasn't _actually_ reflecting the "optimistic response" and instead we would leave it untouched and have the `OfflineWithFeedback` component hide the children based on the `pendingAction`. 

Mainly to the rest of the app the action looks like it is "not deleted", but I think for most purposes it should appear to as "deleted" e.g. when calculating what the last message of a report or `lastReadSequenceNumber` should be.

So, to clarify the statement in the description... you are right it seems like it works on `main`, but I noticed this wasn't working locally after I made the decision to unset the message optimistically on delete instead of leaving it in place.

Since you worked on this feature before, do you agree with the changes I've made or see any issues that I might not have caught?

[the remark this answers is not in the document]
PauloGasparSv: Totally agree with the changes and the code looks cleaner :D

neil-marcellini (commented):
I've only looked at this a bit so far, but it's really cool! For now, I have a few questions:

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+const USER_B_EMAIL = 'user_b@test.com';
+const USER_C_ACCOUNT_ID = 3;
+const USER_C_EMAIL = 'user_c@test.com';
+
+/**
+ * Sets up a test with a logged in user that has one unread chat from another user. Returns the <App/> test instance.
+ *
+ * @returns {RenderAPI}
+ */
+function signInAndGetAppWithUnreadChat() {
```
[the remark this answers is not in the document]
marcaaron: Great point. I think there's also a chance that a lot of UI tests could not need this since they might test more isolated individual components and not flows that require navigation. I think this would be good to add in a future PR though once there is a second test that needs this behavior.

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+                expect(optionRows.length).toBe(1);
+
+                // And that the text is bold
+                const displayNameText = renderedApp.getByA11yLabel('Chat user display names');
+                expect(lodashGet(displayNameText, ['props', 'style', 0, 'fontWeight'])).toBe(fontWeightBold);
+
+                return navigateToSidebarOption(renderedApp, 0);
+            })
+            .then(() => {
+                // Verify that the report screen is rendered and the drawer is closed
```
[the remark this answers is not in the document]
marcaaron: Good question. They are mocked here as `withWindowDimensions` is using `ReactNative.Dimensions`:

https://github.com/Expensify/App/blob/b3ab1ba3c7c0a99d9604f1df0562f7684f468222/__mocks__/react-native.js#L56-L59

__mocks__/@react-native-firebase/crashlytics.js
```diff
@@ -0,0 +1 @@
+export default {};
```
tgolen, line 1: Would it be OK if we also make a requirement that any mock file also should have an explanation for why it's mocked? I think that would be super helpful.

__mocks__/react-native.js
```diff
@@ -0,0 +1,64 @@
...
+                    }
+
+                    return ReactNative.AppState.addEventListener(type, listener);
+                },
+            },
+            Dimensions: {
+                ...ReactNative.Dimensions,
+                addEventListener: jest.fn(),
+                get: () => ({
+                    // Tests will run with the app in a typical small screen size
```
tgolen, line 57: Is there a reason or purpose for being a small screen? Is this something that maybe a test would be able to specify or not (since it can change the behavior of components)?

jest/setup.js
```diff
@@ -1,15 +1,71 @@
...
+// <App> uses <ErrorBoundary> and we need to mock the imported crashlytics module
+// due to an error that happens otherwise https://github.com/invertase/react-native-firebase/issues/2475
+jest.mock('@react-native-firebase/crashlytics', () => () => ({
+    log: jest.fn(),
+    recordError: jest.fn(),
 }));
 
-jest.mock('react-native-blob-util', () => ({}));
+// The main app uses a NativeModule called BootSplash to show/hide a splash screen. Since we can't use this in the node environment
+// where tests run we simulate a behavior where the splash screen is always hidden (similar to web which has no splash screen at all).
```
tgolen, line 27: ❤️ I love this comment. It makes it so clear why the mock is necessary!

src/libs/ReportActionsUtils.js
```diff
@@ -76,7 +97,59 @@ function isConsecutiveActionMadeByPreviousActor(reportActions, actionIndex) {
...
+    const messageText = parser.htmlToText(htmlText);
+    return ReportUtils.formatReportLastMessageText(messageText);
+}
+
+/**
+ * @param {Number} reportID
+ * @param {Object} [actionsToMerge]
+ * @param {Number} deletedSequenceNumber
+ * @param {Number} lastReadSequenceNumber
+ * @return {String}
```
tgolen, line 127:
```suggestion
 * @return {Number}
```

tests/ui/UnreadIndicatorsTest.js
tgolen: Great! Mind removing the comment too? 😁

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+                expect(isDrawerOpen(renderedApp)).toBe(false);
+
+                // That the report actions are visible along with the created action
+                const createdAction = renderedApp.getByA11yLabel('Chat welcome message');
+                expect(createdAction).toBeTruthy();
+                const reportComments = renderedApp.getAllByA11yLabel('Chat message');
```
tgolen, line 199: Was there a specific reason to keep these as `getBy` instead of switching to `queryBy`?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+                const sequenceNumber = lodashGet(unreadIndicator, [0, 'props', 'data-sequence-number']);
+                expect(sequenceNumber).toBe(2);
+
+                // Scroll up and verify that the "New messages" badge appears
+                scrollUpToRevealNewMessagesBadge(renderedApp);
+                expect(isNewMessagesBadgeVisible(renderedApp)).toBe(true);
+            });
+    });
+
+    it('Clear the new line indicator and bold when we navigate away from a chat that is now read', () => {
```
tgolen, line 215: I usually try to write these so that can be read from the CLI easier like a sentence combining both the group description and the test description. So, for this, the group description is "Unread Indicators". If you then add the test description to that, you get "Unread Indicators Clear the new line indicator and bold when we navigate away from a chat that is now read" which reads very weird. 

My suggested edit would be:

```suggestion
    it('are removed when a chat is read and navigated away from', () => {
```

That way, you end up with "Unread Indicators are removed when a chat is read and navigated away from" which is very readable.

This is just a NAB, and something to consider.

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+            .then(() => {
+                // Verify the unread indicator is not present
+                const unreadIndicator = renderedApp.queryAllByA11yLabel('New message line indicator');
+                expect(unreadIndicator.length).toBe(0);
+                expect(isDrawerOpen(renderedApp)).toBe(false);
+
+                // Scroll and verify that the new messages badge is also hidden
+                scrollUpToRevealNewMessagesBadge(renderedApp);
+                expect(isNewMessagesBadgeVisible(renderedApp)).toBe(false);
+            });
```
tgolen, line 250: This looks great with the promises!!

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+                        sequenceNumber: 1,
+                        timestamp: NEW_REPORT_CREATED_MOMENT.add(5, 'seconds').unix(),
+                        message: [{type: 'COMMENT', html: 'Comment 1', text: 'Comment 1'}],
+                        reportActionID: NumberUtils.rand64(),
+                    },
+                });
+                Onyx.merge(ONYXKEYS.PERSONAL_DETAILS, {
+                    [USER_C_EMAIL]: TestHelper.buildPersonalDetails(USER_C_EMAIL, USER_C_ACCOUNT_ID, 'C'),
+                });
+                return waitForPromisesToResolve();
```
tgolen, line 292: Something else I did in the LHN tests was to use `return Onyx.multiSet()` which would prevent you from needing to deal with three separate promises here and can remove another `waitForPromisesToResolve`.

But in that case, I think we also need to wrap them in `act()` to ensure all the component tree is updated properly before making assertions. I have yet to test that out fully.

That also assume that using `set()` is OK (which I think it usually is for tests).

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+            .then(() => {
+                // Verify notification was created as the new message that has arrived is very recent
+                expect(LocalNotification.showCommentNotification.mock.calls.length).toBe(1);
+
+                // // Navigate back to the sidebar
+                return navigateToSidebar(renderedApp);
+            })
+            .then(() => {
+                // // Verify the new report option appears in the LHN
+                const optionRows = renderedApp.getAllByA11yHint('Navigates to a chat');
```
tgolen, line 303: `queryAllBy`?

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -0,0 +1,521 @@
...
+                expect(optionRows.length).toBe(2);
+
+                // Verify the text for both chats are bold indicating that nothing has not yet been read
+                const displayNameTexts = renderedApp.queryAllByA11yLabel('Chat user display names');
+                expect(displayNameTexts.length).toBe(2);
```
tgolen, line 308: NAB: Here is a pretty cool assertion I came across:

```suggestion
                expect(displayNameTexts).toHaveLength(2);
```

I like it because it makes the intention of the assertion to be really clear.

tests/utils/waitForPromisesToResolveWithAct.js
```diff
@@ -0,0 +1,21 @@
+import {act} from '@testing-library/react-native';
+import waitForPromisesToResolve from './waitForPromisesToResolve';
+
+/**
```
tgolen, line 4: I love this comment, but what's not clear to me is when it should be used. Is there some guidance that can be added for that? Like... "Use when..." and "Do not use when..."

tgolen (changes requested):
Wow, those tests are looking great now!

__mocks__/react-native.js
marcaaron: Great question. I'm not sure yet. I chose to hardcode in the "small screen" since I wanted to primary make sure we are covering the native behavior and because the react-native test renderer is always pulling from `index.native.js` source.

I was thinking a good future improvement would be to allow some way to do something like:

```js
Dimensions.set(TestHelper.SCREEN_SIZE.SMALL);
```

tests/ui/UnreadIndicatorsTest.js
marcaaron: Doh!

tests/ui/UnreadIndicatorsTest.js
marcaaron: Nope! I think we are good to update them all.

tests/ui/UnreadIndicatorsTest.js
marcaaron: Makes sense. I guess some of these are testing a few different things so there's not always a single subject.

tests/ui/UnreadIndicatorsTest.js
marcaaron: Gonna leave this one - mainly because having a few different merges happen seems closer to what would happen in the UI.

__mocks__/react-native.js
marcaaron: I went ahead and did this. Even though no one is using it yet we'll have the option in the future to test on a non small screen.

__mocks__/@react-native-firebase/crashlytics.js
marcaaron: I think that sounds great. The answer won't always be the same so it is good context to add for sure and also will help people understand all the reasons why something could or should be mocked.

marcaaron: But also would say we can do that in a cleanup step.

tests/ui/UnreadIndicatorsTest.js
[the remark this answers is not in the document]
marcaaron: Improved this with a constant and also made it customizable.

marcaaron:
Updated. Thanks for the additional comments @tgolen and @neil-marcellini

tests/ui/UnreadIndicatorsTest.js
tgolen: In my LHN tests, I even used nested `describe()` to help with that. Maybe there is a good way to do this in your tests.

image

It's easy to see how tests are grouped together and why (different description for each subject kinda)

tgolen (approved):
Looks great!

marcaaron:
Oh shoot are we supposed to add a reviewer checklist in a comment or edit the description? I’m not sure 🥲

This was not an emergency. Just the contributor checklist check.

PR #11343, opened 2022-09-27
[WIP / Experiment] Onyx state tracking w/ global state proof of concept

marcaaron:
cc @roryabraham 

### Details

Proof of Concept for switching `withOnyx()` calls to get their values from a global state provider + consumers which get updated "magically" when the properties of the state they are interested in change.

Not optimized and haven't benchmarked. It seems a bit slower or equivalent than what we have now so I think needs tweaking. I am also sure there are things about this `react-tracked` library that I don't quite understand. 

But the concept is pretty interesting -> https://react-tracked.js.org/docs/introduction/

This could be further improved on by having some kind of more specific selector e.g. if a component only needs one field it could in theory do:

```js
withTrackedState((state, props) => ({
    report: state[`${ONYXKEYS.COLLECTION.REPORT}${props.reportID}`].notificationPreference,
}))(NotificationPreferenceComponent);
```

There are also other libraries in the same realm like [`valtio`](https://github.com/pmndrs/valtio) that look interesting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,6 +81,7 @@
         "react-native-webview": "^11.17.2",
         "react-pdf": "5.7.2",
         "react-plaid-link": "3.3.2",
+        "react-tracked": "^1.7.10",
         "react-web-config": "^1.0.0",
         "save": "^2.4.0",
         "shim-keyboard-event-key": "^1.0.3",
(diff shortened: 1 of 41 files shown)


marcaaron:
This was a super crazy experiment and didn't seem to have any immediate improvements on performance so I'm going to close it for now.

PR #11362, opened 2022-09-27
[No QA] Change reportID type from number to string

tgolen:
### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/227425
- [x] HOLD on Web-E PR https://github.com/Expensify/Web-Expensify/pull/34976

### Tests
1. Select a chat in the LHN
2. Verify that the row in the LHN shows that the chat is selected (it has a slightly darker background than the rest)

There are a few propType warnings that happen regarding the IOU chat reportIDs, but that API is not fully migrated over to an OfflineFirst world, so those will be fixed later when it's migrated.

### QA Steps
1. Select a chat in the LHN
2. Verify that the row in the LHN shows that the chat is selected (it has a slightly darker background than the rest)

### DIFF
--- a/src/pages/reportPropTypes.js
+++ b/src/pages/reportPropTypes.js
@@ -31,6 +31,9 @@ export default PropTypes.shape({
     /** Flag to check if the report actions data are loading */
     isLoadingReportActions: PropTypes.bool,
 
+    /** Whether the user is not an admin of policyExpenseChat chat */
+    isOwnPolicyExpenseChat: PropTypes.bool,
+
     chatType: PropTypes.string,
     ownerEmail: PropTypes.string,
     isPinned: PropTypes.bool,
--- a/src/pages/reportPropTypes.js
+++ b/src/pages/reportPropTypes.js
@@ -1,0 +1,15 @@
+import PropTypes from 'prop-types';
+
+export default PropTypes.shape({
+    /** Name of the report */
+    reportName: PropTypes.string,
+
+    /** List of primarylogins of participants of the report */
+    participants: PropTypes.arrayOf(PropTypes.string),
+
+    /** List of icons for report participants */
+    icons: PropTypes.arrayOf(PropTypes.string),
+
+    /** ID of the report */
+    reportID: PropTypes.string,
+});
--- a/src/pages/reportPropTypes.js
+++ b/src/pages/reportPropTypes.js
@@ -11,5 +11,14 @@ export default PropTypes.shape({
     icons: PropTypes.arrayOf(PropTypes.string),
 
     /** ID of the report */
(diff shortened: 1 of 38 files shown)


src/pages/reportPropTypes.js
```diff
@@ -0,0 +1,76 @@
...
+    participants: PropTypes.arrayOf(PropTypes.string),
+
+    /** Linked policy's ID */
+    policyID: PropTypes.string,
+
+    /** Name of the report */
+    reportName: PropTypes.string,
+
+    /** ID of the report */
+    reportID: PropTypes.string.isRequired,
```
danieldoglas, line 69: Question: since reportID is required here, won't it break the usages where reportID was not required (or even existed) in the props where it was not necessary? files?

danieldoglas (commented):
Tests are failing. Overall changes look good to me, but I added a question that might be mostly because I don't fully understand props on ReactNative.

src/pages/reportPropTypes.js
tgolen: Yeah, so I ended up removing the `.isRequired` here, however, I not sure what you mean by "files". A report cannot exist in Onyx without a reportID since the collection is keyed by reportID (but not all reports going to the props comes straight from Onyx either).

danieldoglas: I mean other JS files like [src/pages/home/report/ReportActionsList.js](https://github.com/Expensify/App/pull/11362/files#diff-9ac76d645ad53156c86195c1a8f5adb3bb9b71f78351e43c4e1b6d6896921c4e), that was not requiring reportID. 

But I think removing solves it! Thanks!

danieldoglas (approved):
LGTM

danieldoglas:
not emergency

aldo-expensify:
I think this PR may have missed these places:

- Shouldn't the client generated reportIDs be strings too?
https://github.com/Expensify/App/blob/87921cad84020318a165559cc1984c2876a43f57/src/libs/ReportUtils.js#L577-L579

https://github.com/Expensify/App/blob/87921cad84020318a165559cc1984c2876a43f57/src/pages/home/report/ReportActionItemCreated.js#L18

There are some `iouReportID` and `chatReportID` as numbers, but from the PR description I'm getting that these were intentionally skipped, is that correct?

tgolen:
Ah, I did totally miss `gernerateReportID()`. Good catch. Do you want to do a PR on that?

I did intentionally skip the others, yes, since it looked like those APIs had not been migrated over to NewDot APIs yet in PHP. I didn't want to make any changes on the old API endpoints.

aldo-expensify:
Ok! I'll do it

PR #11455, opened 2022-09-29
Add isDrawerReady method

alex-mechler:
### Details
The drawer navigator is conditionally rendered, only after reports are fetched from the server. We need to wait on transition links for this to load if the link is to a drawer component, like a report.

### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/230725

### Tests
1. Pulled https://github.com/Expensify/Web-Expensify/pull/34988 locally
2. Fetched a report ID from newdot
3. Signed out of newdot
4. Navigated to `expensify.com.dev/newdotreport?reportID=<reportID>`
5. _Verified_ I was logged into newdot, and dropped into the correct report

1. Created a free plan workspace from Olddot
2. _Verified_ the transition worked as expected 

### QA Steps
1. Create a free plan workspace from Old Dot
2. _Verify_ the transition works as expected

### DIFF
--- a/src/libs/Navigation/AppNavigator/BaseDrawerNavigator.js
+++ b/src/libs/Navigation/AppNavigator/BaseDrawerNavigator.js
@@ -49,6 +49,12 @@ class BaseDrawerNavigator extends Component {
         };
     }
 
+    componentDidMount() {
+        // We need to resolve the isDrawerReady promise so that any pending drawer actions, like direct navigation from OldDot to
+        // a NewDot report, can happen.
+        Navigation.setIsDrawerReady();
+    }
+
     componentDidUpdate(prevProps) {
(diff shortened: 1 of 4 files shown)


marcaaron (approved):
LGTM since this is fairly urgent. I can't really think of another way to do this without triggering a refactor of the the drawer navigator component so that it always loads the `ReportScreen`.

I think that a refactor could be the better solution, but not sure exactly what would go into that.

alex-mechler:
Agreed, I think that a refactor is a better long term solution. This is blocking a daily 121 issue, which is why I went this route

jasperhuangg (approved):

jasperhuangg:
Tested and works great!

PR #11909, opened 2022-10-17
Revert "Update handling so that Concierge link redirects to chat"

sketchydroide:
Reverts Expensify/App#11527

Internal QA for this did not pass and we though it was better to revert, and then let @thienlnam take a look tomrrow
https://expensify.slack.com/archives/C07J32337/p1666020529387209


iwiznia (approved):

luacmartins (approved):

sketchydroide:
This is an emergency, and the rever should not need to go through the check lists.

Issue #11807, opened 2022-10-13
[HOLD for payment] [$1000] Bug: Mobile app still shows workspace after removed from Web reported by @aimane-chnaif 


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Create workspace if not exists
2. Refresh web and mobile apps and confirm that workspace shows
3. Close mobile app
4. Delete workspace on web
5. Open mobile app

## Expected Result:
Workspace should be deleted on mobile app as well

## Actual Result:
Deleted workspace still remaining on mobile app

## Workaround:
unknown

## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.2.14
**Reproducible in staging?:** y
**Reproducible in production?:** y
**Email or phone of affected tester (no customers):** 
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 

https://user-images.githubusercontent.com/43996225/195644451-d78c48fd-823b-413f-ab46-b282928eb102.mp4

https://user-images.githubusercontent.com/43996225/195644586-577c054a-4007-4adb-9310-3f2e13ef1e1e.mp4

**Expensify/Expensify Issue URL:** 
**Issue reported by:** @aimane-chnaif 
**Slack conversation:** https://expensify.slack.com/archives/C01GTK53T8Q/p1665065560778309

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

## Upwork Automation - Do Not Edit
- Upwork Job URL: https://www.upwork.com/jobs/~0174f3f0c9ecb3ddeb
- Upwork Job ID: 1594940834258837504

zanyrenney:
Ugh yeah this is weird, I don't know why the disconnect between mobile and web here - will get some eng eyes!

madmax330:
Yeah they should be synced, not sure why mobile isn't receiving the updates. Maybe we're not sending the updates for this action via Pusher? 
I think this should be internal

MonilBhavsar:
cc @aldo-expensify If you have any idea about this issue or can take a look

aldo-expensify:
[...]

PR #11130, opened 2022-09-20
Allow retryable optimistic workspace creation

jasperhuangg:
### Details
This creates workspaces with `API.write` so workspace creation can be retried if created offline. It also excludes fetching workspaces that were created offline when reconnecting the app, since that'll effectively clear out any optimistic values that were initially created locally (see comment in `App.js` diff for additional context).

https://user-images.githubusercontent.com/31285285/191351237-8bc8fd2d-48bd-4725-9e3c-ae835ba7fcbe.mov

### Fixed Issues

$ https://github.com/Expensify/App/issues/10897

### Tests and QA Steps
(See above screen recording)

#### Creation via FAB
1. Create and sign into a fresh account.
2. Turn off your internet connection.
3. Use the FAB menu > New Workspace.
4. Verify you are navigated to the workspace.
5. Update the workspace name while offline.
6. Turn your internet connection back on.
7. Verify that the workspace stays open and no Page Not Found message appears.

#### Creation via another workspace (perform above FAB steps first)
1. Navigate to the workspace you just created, then go offline.
2. Click the vertical ellipses menu > New Workspace.
3. Verify you are navigated to the new workspace.
4. Update the workspace name while offline.
5. Turn your internet connection back on.
6. Verify that the workspace stays open and no Page Not Found message appears.

- [ ] Verify that no errors appear in the JS console

#### Contributor (PR Author) Checklist
- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
[...]

PR #12129, opened 2022-10-25
Fix removal of deleted workspaces from the client

aldo-expensify:
### Details

To clear a workspace that no longer exists from the client we send the policy ids that we have in the client in the requests `ReconnectApp` and `OpenApp`. The backend will send onyx updates to remove the workspace that no longer exists.

We were sending the `ReconnectApp` and `OpenApp` before reading the policy data from Onyx, resulting in always sending an empty array of policies ids, resulting in not cleaning deleted workspaces.

### Fixed Issues

$ https://github.com/Expensify/App/issues/11807

and this https://github.com/Expensify/App/pull/11130/files#r1004909597

#### `OpenApp` request test

1. Have an account with a workspace created
2. Login in any two different sessions (mobile + web, web + web, etc)
3. Close the app:
    - if in mobile / desktop: just close the app
    - if in web: close the tab, but leave the window (session) open, so later you can reopen the url in a new tab and you are still logged in
5. With the session that is online, delete the workspace
6. Wait 10 seconds
7. In the other session, reopen the App or the tab
9. Verify that the workspace gets deleted

#### `ReconnectApp` request test

Same steps as above, but instead of closing the tab/app, you have to turn off the internet connection (make it offline), and instead of reopening the tab/app, you restore the internet connection

#### OpenApp request test

1. Have an account with a workspace created
2. Login in any two different sessions (mobile + web, web + web, etc)
3. Close the app:
    - if in mobile / desktop: just close the app
    - if in web: close the tab, but leave the window (session) open, so later you can reopen the url in a new tab and you are still logged in
5. With the session that is online, delete the workspace
6. Wait 10 seconds
7. In the other session, enable the internet connection again (or open the mobile app if you closed it)
8. Verify that the workspace gets deleted

#### ReconnectApp request test

Same steps as above, but instead of closing the tab/app, you have to turn off the internet connection (make it offline), and instead of reopening the tab/app, you restore the internet connection

#### PR Author Checklist
    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop

<details>

The reviewer will copy/paste it into a new comment and complete it after the author checklist is completed
</summary>

    - [ ] iOS / native
    - [ ] Android / native
    - [ ] iOS / Safari
    - [ ] Android / Chrome
    - [ ] MacOS / Chrome
    - [ ] MacOS / Desktop

</details>

#### Web

`OpenApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200705961-a96a1a86-9e63-4269-8a7a-017723bb772f.mov

`ReconnectApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200706298-c8f0cc22-97b2-431e-b719-a30678fc8a06.mov

#### Mobile Web - Chrome

`OpenApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200708321-ebe7a88b-6906-4fcf-b7a8-205d46144da4.mov

`ReconnectApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200708357-c74c513e-da87-42ec-b46a-26427dc76e6e.mov

#### Mobile Web - Safari

`OpenApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200714233-ae918831-90c5-4a50-8fe4-19934edfc725.mov

`ReconnectApp` clears correctly the deleted workspace:

I have no idea how to disable the wifi only from the IOS emulator without disabling it in my server/web browser. This could be easier if I had a real IOS device.

#### Desktop

https://user-images.githubusercontent.com/87341702/199351320-1e640553-3183-4990-b73c-1b79eee438fe.mov

#### iOS

`OpenApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200714117-1f165a30-e221-454c-a277-2840da16c353.mov

`ReconnectApp` clears correctly the deleted workspace:

I have no idea how to disable the wifi only from the IOS emulator without disabling it in my server/web browser. This could be easier if I had a real IOS device.

#### Android

`OpenApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200707567-67f0f65f-a971-4d46-bfee-933d7e4a8601.mov

`ReconnectApp` clears correctly the deleted workspace:

https://user-images.githubusercontent.com/87341702/200707625-a0df48ac-c73e-4c01-b190-3964cc2cd928.mov


src/libs/actions/App.js
```diff
@@ -45,28 +45,22 @@ Onyx.connect({
...
-    waitForCollectionCallback: true,
-    callback: (policies) => {
-        policyIDList = _.compact(_.pluck(policies, 'id'));
-    },
-});
-
 // When we reconnect the app, we don't want to fetch workspaces created optimistically while offline since they don't exist yet in the back-end.
 // If we fetched them then they'd return as empty objects which would clear out the optimistic policy values initially created locally.
 // Once the re-queued call to CreateWorkspace returns, the full contents of the workspace excluded here should be correctly saved into Onyx.
 let policyIDListExcludingWorkspacesCreatedOffline = [];
```
luacmartins, line 51: NAB I know that this was introduced in a different PR, but I think this is part of a bigger issue (see this [thread](https://expensify.slack.com/archives/C03TQ48KC/p1666371883290389)). 

The original intent of the `policyIDList` param was to clean up locally stored policies in case the user missed the delete policy pusher event. I think once we solve the issue in the thread that I linked, we should clean this up and send all policy IDs like we were doing before.

luacmartins (commented):
I'm still getting an empty policyIDList in OpenApp

Screen Shot 2022-10-25 at 2 20 38 PM

src/libs/actions/App.js
```diff
@@ -45,28 +45,22 @@ Onyx.connect({
...
+const policyIDListExcludingWorkspacesCreatedOfflineLoaded = new Promise((resolve) => {
+    Onyx.connect({
+        key: ONYXKEYS.COLLECTION.POLICY,
+        waitForCollectionCallback: true,
+        callback: (policies) => {
+            const policiesExcludingWorkspacesCreatedOffline = _.reject(policies,
+                policy => lodashGet(policy, 'pendingAction', null) === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD
+                && lodashGet(policy, 'type', null) === CONST.POLICY.TYPE.FREE);
+            policyIDListExcludingWorkspacesCreatedOffline = _.compact(_.pluck(policiesExcludingWorkspacesCreatedOffline, 'id'));
+            resolve();
```
aldo-expensify, line 61: uh, I have to check that calling multiple times `resolve` wont cause a bug

aldo-expensify:
> I'm still getting an empty policyIDList in OpenApp

Strange, for me it worked. Can you put a console log right before `API.read('OpenApp'` logging `policyIDListExcludingWorkspacesCreatedOffline` and also console log the same variable right before calling `resolve()`? this is to double check that `API.read('OpenApp'` is running after the variable is set

luacmartins:
This is what I got:
Screen Shot 2022-10-25 at 4 12 08 PM
Screen Shot 2022-10-25 at 4 11 56 PM

luacmartins:
It seemed to work when I refreshed the page, but not on sign in. I guess we don't care about the sign in case, since Onyx would be empty anyways?

aldo-expensify:
> It seemed to work when I refreshed the page, but not on sign in. I guess we don't care about the sign in case, since Onyx would be empty anyways?

ohh, yes, this won't work for that case (sign in), but as you say it shouldn't matter.

src/libs/actions/App.js
aldo-expensify: Ty for linking that, I read it and https://github.com/Expensify/Expensify/issues/229696. I understood that the problem is that read command can potentially be executed before queued offline writes. This caused flickering because there is optimistic stuff removed from onyx that gets re-added back later. I feel like this code here should be fixing that by omitting the optimistic policies, I'm not sure why it remains a problem.

I noticed that @jasperhuangg passed the policy ids with an incorrect parameter name, so the backend is completely ignoring the list: https://github.com/Expensify/App/pull/11130/files#r1004909597, could that be why it remains a problem?

src/libs/actions/App.js
arosiclair: Seems to be fine: https://stackoverflow.com/a/29491617/18466468

src/libs/actions/App.js
```diff
@@ -116,45 +110,49 @@ AppState.addEventListener('change', (nextAppState) => {
...
-            key: ONYXKEYS.IS_LOADING_REPORT_DATA,
-            value: false,
-        }],
-        failureData: [{
-            onyxMethod: CONST.ONYX.METHOD.MERGE,
-            key: ONYXKEYS.IS_LOADING_REPORT_DATA,
-            value: false,
-        }],
+    policyIDListExcludingWorkspacesCreatedOfflineLoaded.then(() => {
+        API.write('ReconnectApp', {policyIDList: policyIDListExcludingWorkspacesCreatedOffline}, {
```
arosiclair, line 139: Instead of waiting on a promise and then accessing the value separately, I think the promise should return the value itself. So ideally we would have something like the following
```javascript
getPolicyIdList()
    .then((policyIdList) =>  API.write('ReconnectApp', {policyIDList}, ...)
```

To do this I think you'd have to change the promise to `Onyx.connect` to get the value and `Onyx.disconnect` to clean up.

arosiclair (changes requested):

src/libs/actions/App.js
aldo-expensify: Hmm, I agree with that it looks nicer, but this may not be so straight forward

- if the code stayed like it is and we return the value in the `resolve` and if `ReconnectApp` gets called a second time, it could see an outdated value because the subsequent resolves don't do anything.
- if we add a `getPolicyIdList()` that creates a promise each time, I don't see how you can reliably make sure that the promise will get resolved. Currently the promise only gets resolved when Onyx connect callback is called, but this won't happen if there is no change in the collection.

aldo-expensify: If you want to push this path, could you help me with developing more your suggestion? I really don't see how this could be done that way at the moment... thanks!

src/libs/actions/App.js
aldo-expensify: thanks for checking, I forgot about it hehe

src/libs/actions/App.js
arosiclair: > Currently the promise only gets resolved when Onyx connect callback is called, but this won't happen if there is no change in the collection.

I had this concern as well. It's not straight forward to just get a value from Onyx I think the only way is to subscribe a callback with Onyx.connect. However, when you subscribe a new callback shouldn't the callback be called with the current value? If so, to get a value I think you should be able write a function like this:

```javascript
function getPolicyIDList () {
        const connectionID = Onyx.connect({ // temporary connection
            key: ONYXKEYS.COLLECTION.POLICY,
            waitForCollectionCallback: true,
            callback: (policies) => {
                const policyIDList = _.map(policies, policy => policy.id);
                Onyx.disconnect(connectionID); // cleanup connection
                resolve(policyIDList); // resolve with value
            },
        });
    });
}
```

It's a little messy with the temp connection (Ideally we could just `Onyx.get()` or something) but it should work the way you'd normally expect a promise to work. Do you mind trying if that works?

aldo-expensify: Thanks @arosiclair for the complete example... sounds like it could work. I looked around our code and found the pattern you are proposing in a couple of places:

- https://github.com/Expensify/App/blob/main/src/libs/migrations/AddEncryptedAuthToken.js#L14-L22
- https://github.com/Expensify/App/blob/main/src/libs/E2E/actions/e2eLogin.js#L16-L50
- https://github.com/Expensify/App/blob/main/src/libs/migrations/RenamePriorityModeKey.js
- etc...

I have a concern about performance though, is it fast to read data from Onyx like that? If it is slow, I don't think it will be a problem if `getPolicyIDList()` is used only for cases like `OpenApp` or `ReconnectApp`, but if it is called from a render method, it could impact the app performance.

@roryabraham @marcaaron can I get your opinion here? 🙏

arosiclair: Yeah I agree this probably should never be called from a render method. To prevent that we could just not export the function

src/libs/actions/App.js
```diff
@@ -116,45 +110,49 @@ AppState.addEventListener('change', (nextAppState) => {
...
-            onyxMethod: CONST.ONYX.METHOD.MERGE,
-            key: ONYXKEYS.IS_LOADING_REPORT_DATA,
-            value: false,
-        }],
-        failureData: [{
-            onyxMethod: CONST.ONYX.METHOD.MERGE,
-            key: ONYXKEYS.IS_LOADING_REPORT_DATA,
-            value: false,
-        }],
+    policyIDListExcludingWorkspacesCreatedOfflineLoaded.then(() => {
```
marcaaron, line 113: If we want to guarantee some values are read from storage in order for the App's init logic to run correctly this is certainly one way to do it. This could be a good opportunity to think a bit more about the "cold cache" problem that [we've run into before](https://github.com/Expensify/App/blob/e5281fcb343b9f9825711b871efb8909ea0df308/src/libs/Network/SequentialQueue.js#L62-L73) and see if we can't solve it another way.

The pattern we are introducing here looks a bit like what we were trying to get away from with the new API stuff in general (promise chaining) and is kind of a thinly veiled `Onyx.get()` (which we have avoided so far)..

I think if we are going to "wait for data from the cache" then we should either:

- accept there is a valid use case for waiting for data values to be read and expose `Onyx.get()` (not sure if there really is yet)
- implement some kind of "initialKeysToCache" array in Onyx so that the data we need when the app starts are already in the cache (would that solve this problem? or could `OpenApp` still run before a call to `Onyx.connect()` initializes the value of the policies).
- some other solution?

cc @tgolen

tgolen: Would `initialKeysToCache` be different than the existing param for `initialKeyStates`?

However, I'm not really sure that this would solve the problem that this PR is intending to solve, or if it does, I don't see how it solve the problem in this PR (since you can't know the policy IDs to seed the cache with).

I definitely agree with @marcaaron that this solution opens up a very slippery slope with chaining.

It seems to me like the policies will definitely be read from Onyx when `reconnectApp()` is called. Is that a correct assumption? So, I don't think we need a promise there.

That leaves us with needing some kind of solution for `openApp()` and I think I would be mostly in favor of adding a single `Onyx.connect()` inside of `openApp()` which isn't exposed to anything else.

marcaaron: >Would initialKeysToCache be different than the existing param for initialKeyStates?

I think they're different as `initialKeyStates` is used to set a default for a key and not "pre-fill" the cache with already stored values.

>However, I'm not really sure that this would solve the problem that this PR is intending to solve, or if it does, I don't see how it solve the problem in this PR

I'm not sure either. Just a random idea that came to mind.

>It seems to me like the policies will definitely be read from Onyx when reconnectApp() is called. Is that a correct assumption?

Agree. I'm not seeing why we need this for `ReconnectApp`.

aldo-expensify: > > It seems to me like the policies will definitely be read from Onyx when reconnectApp() is called. Is that a correct assumption?
> 
> Agree. I'm not seeing why we need this for `ReconnectApp`.

Oh, yes, you are right, the variables not having been read from Onyx in time is only a problem for `OpenApp`

src/libs/actions/App.js
aldo-expensify: Thanks, I didn't know about the caches.

> maybe you can help me understand what we are trying to do and why it's important to know how fast the data will be read?

I just had a concern that adding such function could eventually end up in someone using it in rendering code hindering performance, but, thinking again about my concern, I think it wasn't on point because the `getPolicyIDList()` would really be async (return a promise), so I don't think the "performance" would a problem. The only "problem" in such case would be a new pattern for getting policies leaking into our components logic.

This last "problem" is something that can be assessed by leaving a comment and not exporting it.

aldo-expensify: So.. considering the conversation https://github.com/Expensify/App/pull/12129/files#r1015744356, which made me realize that this really only needed for OpenApp, and not ReconnectApp, it sounds less useful to make a function `getPolicyIDList` if it is only going to be used in one place

src/libs/actions/App.js
aldo-expensify: > That leaves us with needing some kind of solution for openApp() and I think I would be mostly in favor of adding a single Onyx.connect() inside of openApp() which isn't exposed to anything else.

This pattern is in a few places, it is almost like a `Onyx.get`, but async, right? I wonder if we should have a `Onyx.connectOneTime` or an extra parameter to `Onyx.connect` to handle the disconnection.

tgolen: I personally don't think we need any wrapper or functionality around it. I would probably think differently if I saw it misused or abused in several locations.

aldo-expensify:
> @aldo-expensify sorry, i don't have the bandwidth to review this PR. can you please assign another C+ if needed, thanks!

Ok! no problem. I still have to update the changes to address the comments.

aldo-expensify:
This is ready for another round of reviews.

arosiclair (dismissed):
Appreciate the code changes they look good to me!

src/libs/actions/App.js
```diff
@@ -116,51 +116,59 @@ AppState.addEventListener('change', (nextAppState) => {
...
-                key: ONYXKEYS.IS_LOADING_INITIAL_APP_DATA,
-                value: false,
-            },
-        ],
+    // We need a fresh connection/callback here to make sure that we have loaded the policies from Onyx before sending the OpenApp request
+    const connectionID = Onyx.connect({
+        key: ONYXKEYS.COLLECTION.POLICY,
+        waitForCollectionCallback: true,
+        callback: (policies) => {
+            Onyx.disconnect(connectionID);
```
parasharrajat, line 124: **Not related to this PR** It feels like a pattern that we are following for some time. Why don't we create a new method `Onyx.once()`?

marcaaron: It was discussed above but not really common enough of a thing for us to implement in Onyx.

aldo-expensify:
Update: resolved conflicts

@parasharrajat Friendly bump for review

src/libs/actions/App.js
```diff
@@ -116,36 +116,44 @@ AppState.addEventListener('change', (nextAppState) => {
...
-            },
-        ],
-        failureData: [
-            {
-                onyxMethod: CONST.ONYX.METHOD.MERGE,
-                key: ONYXKEYS.IS_LOADING_REPORT_DATA,
-                value: false,
-            },
-        ],
+    // We need a fresh connection/callback here to make sure that we have loaded the policies from Onyx before sending the OpenApp request
```
tgolen, line 119:
```suggestion
    // We need a fresh connection/callback here to make sure that the list of policyIDs that is sent to OpenApp is the most updated list from Onyx
```

src/libs/actions/App.js
```diff
@@ -45,29 +45,29 @@ Onyx.connect({
...
 
-let policyIDList = [];
+let allPolicies = [];
 Onyx.connect({
     key: ONYXKEYS.COLLECTION.POLICY,
     waitForCollectionCallback: true,
     callback: (policies) => {
-        policyIDList = _.compact(_.pluck(policies, 'id'));
+        allPolicies = policies;
     },
```
tgolen, line 54:
```suggestion
callback: policies => allPolicies = policies,
```

src/libs/actions/App.js
```diff
@@ -45,29 +45,29 @@ Onyx.connect({
...
-        const policiesExcludingWorkspacesCreatedOffline = _.reject(policies,
-            policy => lodashGet(policy, 'pendingAction', null) === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD
-            && lodashGet(policy, 'type', null) === CONST.POLICY.TYPE.FREE);
-        policyIDListExcludingWorkspacesCreatedOffline = _.compact(_.pluck(policiesExcludingWorkspacesCreatedOffline, 'id'));
-    },
-});
+/**
+ * When we reconnect the app, we don't want to fetch workspaces created optimistically while offline since they don't exist yet in the back-end.
+ * If we fetched them then they'd return as empty objects which would clear out the optimistic policy values initially created locally.
+ * Once the re-queued call to CreateWorkspace returns, the full contents of the workspace excluded here should be correctly saved into Onyx.
```
tgolen, line 60: I'm not sure that this description makes sense to me. When I read this:

> If we fetched them then they'd return as empty objects

The first thing I think is... why don't we just update the API to not return them as empty objects?

> Once the re-queued call to CreateWorkspace returns

I'm not really sure what this means. What does "re-queue" mean in this context and why does it happen for CreateWorkspace? What does "returns" mean (return is a statement at the end of a method)?

A lot of this comment is assuming a lot about network requests, and other things happening in Onyx, yet the code inside the method has nothing to do with any of that. The code inside the method is just filtering policies. I would suggest simply renaming this method to `getNonOptimisticPolicyIDs` and drop the method description entirely. With a name like that, it's clear what the method does.

src/libs/actions/App.js
```diff
@@ -45,29 +45,29 @@ Onyx.connect({
...
+ *
+ * @param {Array} policies
+ * @return {Array<String>} array of policy ids
+*/
+function getPolicyIDListExcludingWorkspacesCreatedOffline(policies) {
+    const policiesExcludingWorkspacesCreatedOffline = _.reject(policies,
+        policy => lodashGet(policy, 'pendingAction', null) === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD
+        && lodashGet(policy, 'type', null) === CONST.POLICY.TYPE.FREE);
```
tgolen, line 68: Why does it matter if they are free or not? The type of the policy has nothing to do with if the policy was created while it was offline or not (taking the name of the function at face value). I suggest removing this piece of logic.

tgolen (changes requested):

src/libs/actions/App.js
parasharrajat: I do not know much about the policy types but I agree with removing the free policy check,

aldo-expensify: Agreed, looks like an unnecessary check

src/libs/actions/App.js
```diff
@@ -45,29 +45,29 @@ Onyx.connect({
...
-});
+/**
+ * When we reconnect the app, we don't want to fetch workspaces created optimistically while offline since they don't exist yet in the back-end.
+ * If we fetched them then they'd return as empty objects which would clear out the optimistic policy values initially created locally.
+ * Once the re-queued call to CreateWorkspace returns, the full contents of the workspace excluded here should be correctly saved into Onyx.
+ *
+ * @param {Array} policies
+ * @return {Array<String>} array of policy ids
+*/
+function getPolicyIDListExcludingWorkspacesCreatedOffline(policies) {
```
parasharrajat, line 65: We should change it to a short name. `getNonOptimisticPolicyIDs` sounds good to me.

src/libs/actions/App.js
aldo-expensify: > I'm not sure that this description makes sense to me. When I read this:
> 
> > If we fetched them then they'd return as empty objects
> 
> The first thing I think is... why don't we just update the API to not return them as empty objects?

Yeah, the comment seems to be wrong about this, the API doesn't return them as "empty object" it creates Onyx updates setting them to `null`.  This code: https://github.com/Expensify/Web-Expensify/blob/383a1bff0220bf2b795715f371ad443dc634283c/lib/AppInit.php#L87-L103

> 
> > Once the re-queued call to CreateWorkspace returns
> 
> I'm not really sure what this means. What does "re-queue" mean in this context and why does it happen for CreateWorkspace? What does "returns" mean (return is a statement at the end of a method)?
> 
> A lot of this comment is assuming a lot about network requests, and other things happening in Onyx, yet the code inside the method has nothing to do with any of that. The code inside the method is just filtering policies. I would suggest simply renaming this method to `getNonOptimisticPolicyIDs` and drop the method description entirely. With a name like that, it's clear what the method does.

Agreed!

src/libs/actions/App.js
aldo-expensify: Changed to `getNonOptimisticPolicyIDs`

src/libs/actions/App.js
aldo-expensify: Updated, also refactored the code a bit into something that looks more readable to me.

src/libs/actions/App.js
aldo-expensify: Dropped the long description

tgolen (approved):
Thanks! Looks good

arosiclair (approved):

parasharrajat:
Reviewing now.

parasharrajat (approved):
## Reviewer Checklist

    - [x] I verified the steps for expected offline behavior are in the `Offline steps` section
    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop

### Screenshots

#### :black_square_button: MacOS / Desktop + iOS / native



https://user-images.githubusercontent.com/24370807/202530442-bc4077e0-5a82-4603-8d63-7f510f59762f.mp4


#### :black_square_button: MacOS / Desktop +  iOS / Safari
https://user-images.githubusercontent.com/24370807/202530453-6f849fc9-4fb4-4b8c-9f7e-407e948daf3c.mp4


#### :black_square_button:  Chrome + Chrome


https://user-images.githubusercontent.com/24370807/202530414-26153e51-6a4e-4a39-aa5b-e91dead35ecb.mp4

#### :black_square_button:  Chrome + Android / native


https://user-images.githubusercontent.com/24370807/202530428-4abbc7af-b5a2-43b5-910d-0eba368ae37a.mp4


:ribbon: :eyes: :ribbon:  C+ reviewed

aldo-expensify:
@parasharrajat I just learned that for the check `PR Reviewer Checklist / checklist (pull_request_review)` to pass, [your comment](https://github.com/Expensify/App/pull/12129#pullrequestreview-1184862880) **must** start with `## Reviewer Checklist` (see [here](https://github.com/Expensify/App/blob/f82f72c94440db707ab9065092fba5065c935e60/.github/actions/javascript/reviewerChecklist/reviewerChecklist.js#L7))

aldo-expensify:
I'll merge it since C+ reviewed and I consider that we have enough reviews from internal too.

aldo-expensify:
uh, I thought I saw all checks passed. Now I see that there is a bunch of repeated "checklist" checks failed, all related to the failed reviewer checklist check.

image

Anyway, not an emergency!

parasharrajat:
I was scratching my head for 10 mins about what was wrong. Thanks, @aldo-expensify.

Issue #8503, opened 2022-04-06
[HOLD for payment 2024-04-25] [Tracking] [Performance] Enable the new Fabric architecture


mvtglobally:
GitHub Project: https://github.com/orgs/Expensify/projects/35

Main New Arch PR: https://github.com/Expensify/App/pull/13767

## Context 
New render engine: https://reactnative.dev/architecture/fabric-renderer
Summary of findings 
1) There are 4 main components of RN: React, JavaScript, a series of elements collectively known as the “bridge”, and the native elements/modules. Each of these components are getting complimentary upgrades.
2) The notable React upgrades discussed are concurrent rendering, synchronous event callbacks, and priority rendering queues. Together those improvements basically allow for the UI thread to perform heavy rendering work in the background without compromising quick response times to user events such as scrolling and navigation gestures.
3) In the existing RN architecture, the JS and native sides are essentially unaware of each other. All communications are serialized JSON messages sent asynchronously over the bridge.
4) When JavaScript communicates with a web browser, the browser objects hold direct references to native input elements which are controlled internally by the browser. So an HTML checkbox in a browser actually internally holds a reference to a native macOS or Windows checkbox that’s managed by the browser. Similarly, JSI (JavaScript Interface) is a new piece of RN that enables direct communication between JS and the native side, because JSI internally manages direct references to C++ host objects (similar to the HTMLInput element on a browser). This means that messages sent between JS and native modules no longer need to be JSON-serialized or sent over the single bottleneck communication channel. It also means messages can be sent either synchronously or asynchronously.
6) At a high level, the traditional RN bridge manages two tasks: computing layout via a component tree called the “shadow tree” (that shadows the React tree), and managing native modules such as the camera. In the new architecture, these are broken into two distinct pieces which facebook is calling Fabric and TurboModules, respectively.
6) Fabric, the new UI manager, leverages JSI to create the shadow tree in C++ directly, cutting out the middle man and dramatically improving performance. It also exposes UI operations on both sides, so the native side and JS side can both directly manipulate the shadow tree without having to send serialized messages back and forth across the bridge.
[...]

PR #12167, opened 2022-10-26
[WIP] Enable Fabric, Turbo Modules, and New Architecture 

AndrewGable:
### Details

Enable Fabric, Turbo Modules, and New Architecture and fixing many bugs found while enabling it.

### Fixed Issues
$ https://github.com/Expensify/App/issues/8503


hannojg:
The flashes are gone, great! Just wondering:

- Are we fine with just completely removing `Freeze`
- Or do you want us to further investigate and fix the issue with `react-freeze` and expensify?

roryabraham:
@marcaaron Thoughts on @hannojg's last question?

marcaaron:
It was added to prevent an issue where the previous chat would be displayed before displaying a new chat. So if we are OK with that change or if that issue does not exist with `react-freeze` disabled then I have no issues with removing it.

marcaaron:
Not sure if react-freeze was really the best solution to the issue. It's just one I stumbled on. So, if we want to investigate that separately or other alternatives I am 👍

AndrewGable:
One thing I've noticed in addition, it seems like the bold font is broken

`main`|`andrew-fabric`
-------|---------------
Screenshot 2022-10-28 at 12 16 35 PM | Screenshot 2022-10-28 at 11 50 13 AM

marcaaron:
The rendering of a new-ish chat with Welcome message looks a little jumpy. Caught this series of frames:



Also if a chat has not yet been loaded then the content loader also flashes like so

https://user-images.githubusercontent.com/32969087/198706405-6fc74d9a-86e2-4ec2-b6ca-749ecf4c368d.mp4

But looks less bad on a second visit.

Some things that are weird (but not sure if introduced by the current changes):

1. The compose input is hidden when a report first opens.
2. The welcome message flashes on a new-ish chat
3. The content loader flash when we have not yet loaded a chat from the server

marcaaron:
@AndrewGable testing on `main` with `react-freeze` disabled and confirmed that these rendering issues are new to this PR. However, I do see that the previous chat displays before the next loads (in this PR it happens so fast that we barely see it but you can look at the [first frame of the image in this comment](https://github.com/Expensify/App/pull/12167#issuecomment-1295323653) to see that it happens).

Here's an example of my tests on `main`:

https://user-images.githubusercontent.com/32969087/198714950-23404887-7e6c-4102-8109-9abc8a616c07.mp4

marcaaron:
Not too sure but seems like the slow rendering of the compose input is dependent on other activity e.g. if we disable the `SidebarLinks` (expensive to re-render) while the drawer is open you will see the composer input does not hang as bad.
 
**With sidebar links re-rendering**

https://user-images.githubusercontent.com/32969087/198723267-bbac0641-7823-49cd-b792-e63ab81b7ca5.mp4

**With sidebar links disabled**

https://user-images.githubusercontent.com/32969087/198723340-1cb43420-62e2-4b85-b3df-1d4e93d2bffe.mp4

Now the fun part... why does it happen? 😄 Is it something about the fabric architecture? Sorry I am a n00b at understanding how it all works. It kind of looks like things are "hanging" the same amount of time (check my previous comment's video), but some things are revealed in a staggered fashion (whereas before stuff would just hang until the entire view is ready to display).

Looks like there are still some just general improvements to avoiding re-renders of the `SidebarLinks`, `LHNOptionsList`, etc. Gonna look into them as a tangent to this update and maybe it will help resolve things 🤷‍♂️

hannojg:
I think a lot of it could be related to the concurrent rendering mode. I would suggest that we disable that mode first, and then take the time to enable it and properly update all the expensify code to work well with it?

hannojg:
Okay, I just checked it and turned off concurrent mode for testing. Unfortunately, the named issues didn't dissolve magically 😄 

I think it will take more time/effort to enable fabric, by going through the app, identifying all issues and then providing fixes for them.
We can work on those for sure, just wondering whether its something you want us to spend time on (we'd love to get fabric working for Expensify ofc) 😊 ?

roryabraham:
Going to close this in favor of https://github.com/Expensify/App/pull/13767

Issue #13088, opened 2022-11-28
Opening a chat that was not opened recently while being offline seems to take forever to load


techievivek:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Login to newDot.
2. Navigate to one of the chats and from the network tab select `offline` mode.
3. Now click on any other chat and it will load forever.
<img width="1703" alt="Screenshot 2022-11-28 at 7 45 55 PM" src="https://user-images.githubusercontent.com/35863227/204326541-8566ee86-6a5f-495d-bdd5-43a94d522b07.png">

## Expected Result:
Users should alteast be allowed to take any optimistic actions though the chat is not loaded locally in ONYX.
<img width="658" alt="image" src="https://user-images.githubusercontent.com/35863227/205002792-55a734a7-e5d2-453a-be68-7164a5a51098.png">

## Actual Result:
The chat seems to take forever to load.
## Workaround:
No workaround at least when we are offline. When you are online you can manually go to each chat report to make sure ONYX has the data saved for all the chats and now if you go offline it would read stale data from ONYX.
## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 
**Reproducible in staging?:**  Y
**Reproducible in production?:** Y
**Email or phone of affected tester (no customers):** 
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** Any additional supporting documentation
**Expensify/Expensify Issue URL:** 
**Issue reported by:** 
**Slack conversation:**  https://expensify.slack.com/archives/C01GTK53T8Q/p1669644904805489

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

## Upwork Automation - Do Not Edit
- Upwork Job URL: https://www.upwork.com/jobs/~012b395e85cfcc6b49
- Upwork Job ID: 1598232191671402496

s77rt:
I don't think the expected results make sense.
What if you had a chat with that user but you are just in another device?

I don't think this qualifies as a bug

techievivek:
[...]

Issue #13074, opened 2022-11-28
[HOLD for payment 2022-12-27] White screen appears when loading chat, after being redirected from IOU modal


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Have an account with some good number of chats between accounts A and B
2. From account A, go to any other chat except B
3. Signout from A
4. Sign in into A again
5. From + button, Request Money from B
6. In the chat view with B, white screen appears while chats are loading

## Expected Result:
Skeleton view should appear

## Actual Result:
White screen appears

## Workaround:
unknown

## Platform:

Where is this issue occurring?

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.2.32-1
**Reproducible in staging?:** y
**Reproducible in production?:** n
**Email or phone of affected tester (no customers):** 
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 

https://user-images.githubusercontent.com/43996225/204194300-ebbb7960-7a92-4936-b15d-fdfda3bc74ad.mov

https://user-images.githubusercontent.com/43996225/204194309-f65f746a-d4bb-4de8-850e-a86e91747eb6.mp4

**Expensify/Expensify Issue URL:** 
**Issue reported by:** @MonilBhavsar 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1669293337450379

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

maddylewis:
Managing a BUG issue

From  this  SO: https://stackoverflow.com/c/expensify/questions/14418
_______
- [ ] Confirm the "bug" is actually a  bug
- [ ] Confirm the  bug is  not  a  duplicate  report of  an existing  GH (close  the  GH  and  add  any  relevant  details  to  the  original  GH ) 
- [ ] The  bug  is  reproducible

**If you’re unable  to  reproduce  the  bug:**
-  [ ] Add  the `Needs reproduction`  label
-  [ ] Comment  on the  issue  outlining the steps  you  took  to reproduce the bug, and your  results
[...]

PR #13233, opened 2022-12-01
Show static skeleton UI if no reportAction exists for chat report in ONYX when the client is offline

techievivek:
### Details

Opening a chat that was not opened recently while being offline seems to take forever to load and just displays the animating skeleton UI. I have updated the code to now show a non-animating skeleton UI along with the compose box so optimistic action can be taken while the user is offline. I have also added some general fixes to not load the report if we are offline which was needed as part of this code change.

### Fixed Issues

$ https://github.com/Expensify/App/issues/13088 
PROPOSAL: GH_LINK_ISSUE(COMMENT)

### Tests

1. Login to newDot with an account that contains several chats with other users.
2. Now go to `settings -> preferences` and toggle the `offline` mode using `Force offline` switch.
3. Once you are offline, navigate to a chat report that was not open recently.
4. Make sure you see a non-animating skeleton UI and compose box to add new messages to the chat report.
5. Send a few messages and go back to online mode.
6. Once you are online it will start showing the skeleton UI in an animating state and chat history will get loaded.

### Offline tests

1. If you are offline and you opened a chat that was not opened recently you are expected to see a non-animating skeleton UI along with compose box input that will allow you to add new messages even though the chat history doesn't exist in the ONYX.
2. Once you come online you should an animating skeleton UI and the chat history along with the newly added messages will get loaded.

### QA Steps

Same as the testing step above.

### PR Author Checklist

    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/35863227/208663976-98a294c9-c112-424b-8090-0fb090de96c5.mov

</details>

<details>

https://user-images.githubusercontent.com/35863227/208672570-c3851e24-3dbf-41e0-bcce-0c16b39ef474.mov

</details>

<details>

https://user-images.githubusercontent.com/35863227/208667798-7368345b-08f2-465e-8358-5ecea1fb0a5e.mov

</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/35863227/208663308-d39655f0-0464-4083-b184-b631bd4e3958.mov

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/35863227/208666555-219d9cda-72fa-4673-8b06-3ba9f76bdb7c.mov

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/35863227/208672334-c97a5a9e-51f7-4115-ac5c-214faf305f36.mov

</details>


techievivek:
This is a DRAFT PR for now because I wanted to discuss something here.
So there are two ways to tackle this PR.
1. Since report actions are empty and we are offline I can falsely include a report action with the action `CREATED` which will get passed [here](https://github.com/Expensify/App/blob/820e876d5e4a8cbffe272220c1182c8942ed93cf/src/pages/home/ReportScreen.js#L289) which calls ReportActionsList [here](https://github.com/Expensify/App/blob/820e876d5e4a8cbffe272220c1182c8942ed93cf/src/pages/home/report/ReportActionsView.js#L354) and it will automatically render the `ReportActionItemCreated` component. You can look through the flow to understand it better.
2. Second way is what I have done in this PR, which is to return early in `ReportActionsView` with `ReportActionItemCreated` component. This does seem to work but it gets messed up(styling) in the Native Android and iOS apps.

Screenshot 2022-12-01 at 7 02 12 PM

src/pages/home/ReportScreen.js
```diff
@@ -215,7 +215,10 @@ class ReportScreen extends React.Component {
         const screenWrapperStyle = [styles.appContent, styles.flex1, {marginTop: this.state.viewportOffsetTop}];
 
         // There are no reportActions at all to display and we are still in the process of loading the next set of actions.
-        const isLoadingInitialReportActions = _.isEmpty(this.props.reportActions) && this.props.report.isLoadingReportActions;
+        const isLoadingInitialReportActions = (_.isEmpty(this.props.reportActions) && !this.props.network.isOffline) && this.props.report.isLoadingReportActions;
```
marcaaron, line 218: Why do we need this change?

src/pages/home/ReportScreen.js
```diff
@@ -215,7 +215,10 @@ class ReportScreen extends React.Component {
...
+
+        // Check if for the current chat report there are no report actions available while we are offline
+        const isChatReportEmptyWhileOffline = _.isEmpty(this.props.reportActions) && this.props.network.isOffline;
```
marcaaron, line 221: I think since only the `ReportActionsView` cares about this we could move the check into that component instead of here.

src/pages/home/ReportScreen.js
```diff
@@ -293,6 +296,7 @@ class ReportScreen extends React.Component {
                                         isComposerFullSize={this.props.isComposerFullSize}
                                         isDrawerOpen={this.props.isDrawerOpen}
                                         parentViewHeight={this.state.skeletonViewContainerHeight}
+                                        isChatReportEmptyWhileOffline={isChatReportEmptyWhileOffline}
```
marcaaron, line 299: Continuing the previous thought.. we are creating this new variable and prop - but nothing else uses `isChatReportEmptyWhileOffline` so it seems like the responsibility of the `ReportActionsView` to:

1. Check the `reportActions`
2. See if we're offline
3. If there are no `reportActions` and we are offline then incorporate the fallback action

src/pages/home/report/ReportActionsView.js
```diff
@@ -338,6 +342,11 @@ class ReportActionsView extends React.Component {
     }
 
     render() {
+        // If the `isChatReportEmptyWhileOffline` prop is true then let's render ReportActionItemCreated component
+        if (this.props.isChatReportEmptyWhileOffline) {
+            return <ReportActionItemCreated reportID={this.props.report.reportID} />;
+        }
```
marcaaron, line 348: This works, but I think is going to be a bit weird UX. As soon as someone adds an optimistic action while offline this will disappear since `reportActions.length > 0`.

marcaaron (changes requested):
>I can falsely include a report action with the action CREATED

I think by "falsely include" you mean that we will add a fake "created" action into the `reportActions`? I am not sure if it's a good idea to add it to the data in Onyx - I think not as it will soon be migrated to an object keyed by `reportActionID` and would be very confusing to have to remove it somehow later when we are "online".

Though we could possibly develop a solution around modifying the final array that gets passed here: https://github.com/Expensify/App/blob/3342c39668b88daa352e72a52fe423a4739b5e26/src/pages/home/report/ReportActionsList.js#L151

I think this would be a better way to do it as then we would not need to explicitly return early with the created action. Though - this could have other unintended side effects... 

>return early in ReportActionsView with ReportActionItemCreated component. This does seem to work but it gets messed up (styling) in the Native Android and iOS apps which I can look into but the button doesn't seem to response so I am thinking I am missing something here.

We have to look closer at the `ReportActionsList` styles to understand why this doesn't look good and isn't working.

This is a good start - but I think we should bring this one to Slack to get more ideas.

src/pages/home/ReportScreen.js
marcaaron: I guess the thing that confuses me here is why `isLoadingReportActions` would be true if we are offline. I would maybe expect it to be `false` or ask why we are triggering a loader state when we know we can't fetch these actions because we are offline.

techievivek: That was my first question as well. I will look into this part and probably clean it.

src/pages/home/report/ReportActionsView.js
techievivek: But do we have any other way out? Because it is bound to happen, right? Can we show some additional message to let user know we don't have data cached?

marcaaron: Yeah, very tricky. So to summarize:

- We can tell a report has not loaded yet if it has `0` `reportActions.length`
- If we are offline then we can show this new special "created" action 
- As soon as they add a comment it will look like the report has loaded

So the question becomes...

**How can we tell that we need to keep showing the "created" action if we are depending on `reportActions.length` and adding optimistic actions will increase this length?**

First solution that comes to mind is something like...

```js
const hasAtLeastOneNonPendingAction = reportActions.some(action => !action.pendingAction);
```

Which would be the same as "empty or empty with only pending actions". What do you think?

techievivek: I am not sure if I am thinking in the right direction or not but the problem with `some` will be that even if a chat report is loaded and we added an optimistic action to the report the above will evaluate to `true` right? So it all boils down to all the report actions must be in a pending state or the list should be empty inorder for us to show the `created` action, no?

I also went through the other [issue](https://github.com/Expensify/App/issues/13074) which is very similar to this one and I really liked the idea of showing `[broken cloud] Let's sync up when you get back online` rather than a `created` action because both seem to be aligned with our optimistic approach and the latter one defines the situation better here.

marcaaron: >the problem with some will be that even if a chat report is loaded and we added an optimistic action to the report the above will evaluate to true right

My assumption is that if the report has truly loaded it will have a "created" action that is not pending.

>all report actions must be in a pending state or the list should be empty in order for us to show the created action, no

Yep, I think that makes sense.

techievivek: Yeah, I agree we can show the skeleton UI without animation rather than a created action.

techievivek:
@yuwenmemon I am not sure if branching off issues will be a good thing to do here. Because I am experiencing the issue that you will be working on. Also, the report actions seem to be shuffling here.


https://user-images.githubusercontent.com/35863227/206539997-65546153-283d-4780-9ca1-192cee72bf6b.mov

CC @marcaaron

techievivek:
Keeping this on HOLD until the online behaviour gets fixed [here](https://github.com/Expensify/App/issues/13074).

techievivek:
This is ready for review. @marcaaron @yuwenmemon 
I am seeing a bug where messages sent offline change positions once the user comes online, but this is already being discussed [here](https://github.com/Expensify/App/issues/13310) and is not a priority for now.

src/pages/home/report/ReportActionsList.js
```diff
@@ -145,6 +145,10 @@ class ReportActionsList extends React.Component {
     }
 
     render() {
+        // Non-animating skeleton UI represents that chat history exists but is not loaded locally in Onyx.
+        // While offline if there is at least one non-pending action or if this is a new chat then we don't need to show the non-animating skeleton UI.
+        const shouldShowNonAnimatingSkeleton = !(_.some(this.props.sortedReportActions, action => !action.pendingAction || action.actionName === CONST.REPORT.ACTIONS.TYPE.CREATED));
```
yuwenmemon, line 150: Parens are unneccessary

src/pages/home/report/ReportActionsList.js
```diff
@@ -163,7 +167,7 @@ class ReportActionsList extends React.Component {
                     keyExtractor={this.keyExtractor}
                     initialRowHeight={32}
                     initialNumToRender={this.calculateInitialNumToRender()}
-                    onEndReached={this.props.loadMoreChats}
+                    onEndReached={() => !this.props.network.isOffline && this.props.loadMoreChats()}
```
yuwenmemon, line 170: Could we not move the isOffline check into the `loadMoreChats` function itself?

src/pages/home/report/ReportActionsList.js
yuwenmemon: How is this different from `this.props.report.isLoadingReportActions`?

yuwenmemon: `shouldShowNonAnimatingSkeleton` is not a great variable name either. It should ideally reflect what it represents - so in this case ... whether or not we still have report actions to load? (Hence why I have my question above)

yuwenmemon (changes requested):

src/pages/home/report/ReportActionsList.js
techievivek: Agree, that would be cleaner.

src/pages/home/report/ReportActionsList.js
techievivek: `this.props.isLoadingReportActions` should be `false` when we are offline(more [here](https://github.com/Expensify/App/pull/13233#discussion_r1037488570)), so I have updated some parts of code, so we don't call `openReport` when we are offline.

So we need another way to know whether we want to show the non-animating skeleton UI. (i.e., when we don't have any chat history in onyx and all we are left with are optimistic actions that were added by the user while being offline). So we will show the non-animating skeleton UI only if every action is optimistic and it is not a new chat report.

techievivek: I have updated the condition to be more clearer.
```js
// Non-animating skeleton UI represents that chat history exists but is not loaded locally in Onyx.
// While offline if there is at least one non-pending action or if this is a new chat then we don't need to show the non-animating skeleton UI.
const shouldShowNonAnimatingSkeleton = !this.props.report.isOptimisticReport && _.every(this.props.sortedReportActions, action => action.pendingAction);
```

techievivek:
Gentle bump for reminding everyone to review this PR. Thanks

src/pages/home/ReportScreen.js
```diff
@@ -165,7 +165,8 @@ class ReportScreen extends React.Component {
         // It possible that we may not have the report object yet in Onyx yet e.g. we navigated to a URL for an accessible report that
         // is not stored locally yet. If props.report.reportID exists, then the report has been stored locally and nothing more needs to be done.
         // If it doesn't exist, then we fetch the report from the API.
-        if (this.props.report.reportID) {
+        // If the client is offline, we cannot fetch the report, so let's just return.
```
marcaaron, line 168: This comment explains what we are doing but not "why".

Looking at this - and I don't understand why being offline should prevent us from making this request. Here's my understanding of how our networking layer works:

1. `OpenReport` is a "write" request because we update the "last read" while fetching the report
2. If we are offline we will queue any "write" requests

https://github.com/Expensify/App/blob/0ed26362d4201ed548525cca7460459d7db6a678/src/libs/API.js#L47-L48

https://github.com/Expensify/App/blob/0ed26362d4201ed548525cca7460459d7db6a678/src/libs/Network/SequentialQueue.js#L90-L96

3. When we come back from offline -> online those queued "write" requests will run

https://github.com/Expensify/App/blob/0ed26362d4201ed548525cca7460459d7db6a678/src/libs/Network/SequentialQueue.js#L83-L84

In this case, we are offline - so shouldn't it be fine to make the request? And when we come back online the report will be fetched?

Is there some problem with the way this is working now that needs to be addressed? Because I want to say it's an anti-pattern to prevent a "write" from happening if we are offline (as that's what the queue itself was designed to do) but maybe I am missing something.

src/pages/home/report/ReportActionsList.js
```diff
@@ -145,6 +141,10 @@ class ReportActionsList extends React.Component {
     }
 
     render() {
+        // Non-animating skeleton UI represents that chat history exists but is not loaded locally in Onyx.
```
marcaaron, line 144: >Non-animating skeleton UI represents that

I'd rephrase this in this format

>When [x] happens a non-animating skeleton UI is shown because...

But honestly, I'm a bit confused about how to finish that sentence. 

I'm having trouble understanding the idea that "chat history exists but is not loaded locally in Onyx" so I think we need to improve this comment more generally. 

Does it mean that the data exists in storage, but not yet read into memory? 

Or does it mean that we know some data exists on the server, but we haven't tried to load it yet? 

Let's please try to be very specific here so that people don't have to work too hard to understand why we are doing something.

src/pages/home/report/ReportActionsList.js
```diff
@@ -145,6 +141,10 @@ class ReportActionsList extends React.Component {
...
+        // While offline if there is at least one non-pending action or if this is a new chat then we don't need to show the non-animating skeleton UI.
+        const shouldShowNonAnimatingSkeleton = !this.props.report.isOptimisticReport && _.every(this.props.sortedReportActions, action => action.pendingAction);
```
marcaaron, line 146:
```suggestion
        const shouldShowStaticSkeleton = !this.props.report.isOptimisticReport && _.every(this.props.sortedReportActions, action => action.pendingAction);
```

We generally prefer variables that describe what something **is** vs. what it **is not**.

src/pages/home/report/ReportActionsList.js
```diff
@@ -177,12 +177,13 @@ class ReportActionsList extends React.Component {
                         // Make sure the oldest report action loaded is not the first. This is so we do not show the
                         // skeleton view above the created action in a newly generated optimistic chat or one with not
                         // that many comments.
+                        // We will show non-animating skeleton UI if we are offline and chat history is not loaded in ONYX.
```
marcaaron, line 180: Comment does not explain "why" only the "what". Also let's try to be more consistent about capitalization. In other comments we have "Onyx" and this one "ONYX".

```suggestion
                        // We will show a static skeleton UI if we are offline and chat history is not loaded in Onyx because...
```

src/pages/home/report/ReportActionsList.js
marcaaron: Alternatively, change this to `shouldAnimateSkeleton` and then we don't have to invert later `!shouldShowNonAnimatingSkeleton`. I think I like that suggestion better.

marcaaron: Seems unrelated to the changes here. But I'm kind of confused about why and how `isOptimisticReport` got introduced. I'd think that an optimistic `report` would have a `pendingAction` and not a separate flag to denote it as "optimistic".

src/pages/home/report/ReportActionsList.js
```diff
@@ -145,6 +141,10 @@ class ReportActionsList extends React.Component {
     }
 
     render() {
+        // Non-animating skeleton UI represents that chat history exists but is not loaded locally in Onyx.
+        // While offline if there is at least one non-pending action or if this is a new chat then we don't need to show the non-animating skeleton UI.
```
marcaaron, line 145: How does being offline come into play here? This component does not seem aware of the network status anymore since we are removing `withNetwork`.

src/pages/home/report/ReportActionsView.js
```diff
@@ -270,7 +270,7 @@ class ReportActionsView extends React.Component {
 
     // If the report is optimistic (AKA not yet created) we don't need to call openReport again
     openReportIfNecessary() {
-        if (this.props.report.isOptimisticReport) {
+        if (this.props.report.isOptimisticReport || this.props.network.isOffline) {
```
marcaaron, line 273: The first comment applies here too.

src/pages/home/report/ReportActionsView.js
```diff
@@ -283,7 +283,7 @@ class ReportActionsView extends React.Component {
      */
     loadMoreChats() {
         // Only fetch more if we are not already fetching so that we don't initiate duplicate requests.
-        if (this.props.report.isLoadingMoreReportActions) {
+        if (this.props.report.isLoadingMoreReportActions || this.props.network.isOffline) {
```
marcaaron, line 286: Why are we preventing this request from happening?

src/pages/home/report/ReportActionsView.js
```diff
@@ -352,7 +352,7 @@ class ReportActionsView extends React.Component {
 
     render() {
         // Comments have not loaded at all yet do nothing
-        if (!_.size(this.props.reportActions)) {
+        if (!_.size(this.props.reportActions) && !this.props.network.isOffline) {
```
marcaaron, line 355: I think this needs a comment. It's not obvious why we should choose to return `null` when there are no report actions _and_ we are online.

marcaaron: It also feels like maybe we should not be rendering the `ReportActionsView` at all if we are offline and have no actions.

marcaaron (changes requested):
I'm having some trouble wrapping my head around these changes, but don't want to be a blocker on this PR. I have a couple of ideas on how to proceed: 

1. Discuss more widely what it is we are trying to solve and the best way to do it. It feels like we are experimenting in the PR a bit (which is OK, but might be less effective than leading a conversation to get everyone on the same page before making changes).

2. Bring someone else in on the review who is also familiar with the `ReportScreen` flows to weigh in on some of the comments I've left.

src/pages/home/ReportScreen.js
techievivek: The reason I decided to do so was otherwise in the `openReport` method call 
https://github.com/Expensify/App/blob/0bde3f839aa45245df09d9b558eef8b59c066cf1/src/libs/actions/Report.js#L309-L314
it would set `isLoadingReportActions: true` and because of that in the ReportScreen.js 
https://github.com/Expensify/App/blob/0bde3f839aa45245df09d9b558eef8b59c066cf1/src/pages/home/ReportScreen.js#L317-L322
we will keep showing the animating full-screen skeletonUI. This change will prevent that from happening.

But with the recent code change, I can see we are showing the compose box along with the skeletonUI on the report screen page so I can clean all of this.

src/pages/home/report/ReportActionsList.js
techievivek: > Or does it mean that we know some data exists on the server, but we haven't tried to load it yet?
This is the case here, I will update it.

src/pages/home/report/ReportActionsList.js
techievivek: We have decided to show the static skeleton UI for the chats that we know exist on the server but aren't loaded locally in ONYX. 

Now if the chat is not loaded locally and we are offline then we can still take optimistic action but we will need to keep showing the static skeleton UI. But in the case of a new chat report, we don't have to show the static skeleton UI because we know this chat report doesn't exist on our server. But if I just check by `pendingAction` then static skeleton UI would show up for new chats as well so to filter them out I have added the condition to see if it is an optimistic report(which I think denotes a new chat created while offline).

src/pages/home/report/ReportActionsList.js
techievivek: Aaah, I was just trying to explain in comments and used that word. I will get rid of it.

src/pages/home/report/ReportActionsView.js
techievivek: Discussed above, I will get rid of it.

src/pages/home/report/ReportActionsView.js
techievivek: Going to relook into it because we now show the composer box while the chats are loading, so I think I can get rid of most of it.

src/pages/home/ReportScreen.js
techievivek: This change and one [below](https://github.com/Expensify/App/pull/13233#discussion_r1059352019) in ReportActionsView page are mainly to prevent setting value for a report prop. Do you have any recommendations on how to handle this better?

marcaaron: Let's stop here and consult the team on the best way to handle this. 

The main question in my mind is why we should ever set any loading state before the API request actually begins? In the current design, it seems like we have a hybrid where loading states are set to `true` even though we are offline and not actually "loading" anything.

Taking a look at our usages elsewhere - and this `isLoading` is mostly only being used in a few "blocking read requests" (stuff related to payments and bank accounts) that I don't _think_ will ever happen unless we are online. So it feels like there's an inconsistency or exception being made for the report page - and I'm not sure it's valid.

src/pages/home/report/ReportActionsList.js
marcaaron: Awesome thanks! And to make sure I understand - we know this is the case because we have some report data already stored, the report is "not optimistic", and no report actions?

src/pages/home/report/ReportActionsView.js
marcaaron: Ok I think this is also related to the weird issue where we are setting "loading" states when no "loading" is actually happening and we should pause until there is agreement about how to proceed with that.

src/pages/home/report/ReportActionsView.js
marcaaron: >I guess this whole flow is a bit messed up.

I agree and think it's a possible sign that we are gonna make it even more messed up than it already is. 

I would propose we stop exploring code changes for a moment and instead try to pull things back into the realm of problem/solution. 

**Next steps:**

1. Figure out what problem we need to solve (I think we have mostly done this on the issue)
2. Look at how things work today. Does the existing code easily support it or not? Does an obvious solution present itself or not? What solution should we ultimately go with?
3. Summarize our best idea and take it to a few people to see what they think (either `#engineering-chat` or `#expensify-open-source` seems good)
4. Agree on a plan forward + make sure everyone understands the general thought process behind the changes
5. Make the changes

yuwenmemon: @aimane-chnaif No, @techievivek has been OOO - so you're not missing out on anything I believe 🙂

techievivek: Catching up, I will open a thread to discuss this issue in the open-source channel by tomorrow.

src/pages/home/report/ReportActionsList.js
techievivek: Aaah, my bad. I see your question now. But to be honest I don't have any explanation for this one(will have to check the blame). I used it because it was right in the report object but this could have been easily computed as well.

src/pages/home/report/ReportActionsList.js
techievivek: Yep, that's my understanding so far 😄 (This is my first PR in like 4-5 months).

src/pages/home/report/ReportActionsView.js
techievivek: I have opened a discussion [here](https://expensify.slack.com/archives/C01GTK53T8Q/p1673246681648109) feel free to jump to add your thoughts.

marcaaron:
I think we are [still discussing these changes](https://expensify.slack.com/archives/C01GTK53T8Q/p1673986403012729?thread_ts=1673246681.648109&cid=C01GTK53T8Q). @techievivek do you think it might be helpful to just close out this PR for now until there's more of a consensus on what an agreed Problem and Solution? I'm trying to help move this one forward, and if feels like there's been so much conversation that it would be best to just start over at this point with a fresh Problem/Solution statement.

yuwenmemon:
Based off of https://github.com/Expensify/App/issues/13088#issuecomment-1386157984 I do think we should close this out. Feel free to reopen if you disagree!

PR #12648, opened 2022-11-10
[Image][Performance] Fast Image with Auth Token Header Caching

thomas-coldwell:
### Details

This PR combines the draft PR from @Szymon20000 https://github.com/Expensify/App/pull/11009 for adding fast image and the PoC auth token PR from @Beamanator https://github.com/Expensify/App/pull/12146 to properly cache images using fast image with the auth token being passed in the headers rather than the URL.

### Fixed Issues

https://github.com/Expensify/App/issues/10792

$ GH_LINK   
PROPOSAL: GH_LINK_ISSUE(COMMENT)

### Tests

1. Checkout this branch
2. Run the app on both iOS and Android
3. Open a chat with some image attachments
4. They will load in as normal the first time, but reload the app and then you will see the cached images loaded instantly in the same chat

- [x] Verify that no errors appear in the JS console

### QA Steps

- [ ] Open a report chat and add a new image attachment and send so the image appears in the chat (this should display a loading indicator the first time it is requested)
- [ ] Reload the app and you should see the cached image displayed (with no loading indicator)
- [ ] Click on an image in the chat to open the attachment modal view, close it and reopen it to see the cached image displayed without a loading state 
- [ ] Verify that any attachments still download as normal without any errors (including image attachments and PDFs)

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
[...]

Issue #10792, opened 2022-09-02
[Image][Performance] Better caching for iOS images


Szymon20000:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## What performance issue do we need to solve?
Better caching for iOS images

## What is the impact of this on end-users?
Previously seen avatars/assets will show up faster with suggested change.

## Proposed solution (if any)
Use FastImage for Avatars and images that are par of the messages. 
`Avatar.js` uses Image from react-native
react-native-render-html - that as far as I know is used for rendering attachment (among the other things) uses Image from 'react-native' if the change with avatars will give us good results it's worth to test patching that library as well. So that it uses fast-image.

## Platform:
Where is this issue occurring?

- [ ] Web
- [x] iOS
- [ ] Android
- [ ] Desktop App
- [ ] Mobile Web

**Version Number:**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** Any additional supporting documentation
**Upwork URL:** https://www.upwork.com/jobs/~013a8690a2c460efd2

[View all open jobs on Upwork](https://www.upwork.com/ab/jobs/search/?q=Expensify%20React%20Native&sort=recency&user_location_match=2)

**Reported by:** @parasharrajat originally on [this issue](https://github.com/Expensify/App/issues/6527). 
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~019ac837ee43fe141c</li>
        <li>Upwork Job ID: 1620162574495588352</li>
        <li>Last Price Increase: 2023-01-30</li>
    </ul>
</details>

roryabraham:
Switching this to weekly

roryabraham:
@Szymon20000 I saw you put up a draft PR but you weren't sure about the change because right now FastImage isn't compatible with Fabric. Overall I think we should prioritize Fabric compatibility over using FastImage.

Looking into FastImage some more, it seems like one of the main things it handles is persistent image caching. However, that functionality feels potentially in-scope for our persistent storage library Onyx. Going to add this issue to the image tracking issue [here](https://github.com/Expensify/App/issues/10894)

roryabraham:
Any new thoughts here @Szymon20000?

[...]

Issue #12893, opened 2022-11-21
[$2000] Large image doesn't zoom well in Android and gets blurry


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Go to any chat and upload a very long/large image. [Example image](https://user-images.githubusercontent.com/44479856/132889916-e46ac3dc-19ad-4cca-a033-ec13b8ac79ad.jpg)
2. In the Attachment preview, tap the image to zoom
3. Keep zooming until you'll see that the image is blurry

## Expected Result: 
Image should zoom to show a clear image instead of a blurry one

## Actual Result: 
Image content is not visible, and it is blurry.

## Workaround:
unknown

## Platform:

Where is this issue occurring?

- Android

**Version Number:** 1.2.29-6
**Reproducible in staging?:** Y
**Reproducible in production?:** Y
**Email or phone of affected tester (no customers):** 
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 

https://user-images.githubusercontent.com/43996225/203113972-1f4bb867-186f-4d5c-89e7-cd2761fa2c6b.mp4

https://user-images.githubusercontent.com/43996225/203113166-fb40d7b0-6bb3-418c-ad80-6bb22280e9ba.mov

**Expensify/Expensify Issue URL:** 
**Issue reported by:** @mananjadhav 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1668928921788569

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0100609c253c2769f4</li>
        <li>Upwork Job ID: 1597637165026803712</li>
        <li>Last Price Increase: 2022-12-08</li>
    </ul>
</details>

aldo-expensify:
I remember this same images was causing this a long time ago: https://github.com/Expensify/App/issues/5193

I wonder if it ever got fixed and if this is a regression.

mananjadhav:
[...]

PR #13304, opened 2022-12-03
[Image][Performance] Fast Image with Auth Token Header Caching (revert follow-up)

thomas-coldwell:
### Details

This PR is the follow up the the original fast image PR that was reverted [here](https://github.com/Expensify/App/pull/12648). The key improvements in this PR are:
- [x] Better naming convention for the Image component
- [x] Getting the auth token and headers are now fully encapsulated by the `Image` component itself (+ reactive with the `withOnyx` HoC rather than `Onyx.connect()`)
- [x] Resolves the revert issue https://github.com/Expensify/App/issues/13002 where `react-native-fast-image` would not load if the style width / height were set to zero.

### Fixed Issues

Original image caching issue - https://github.com/Expensify/App/issues/10792
Android PDF thumbnail not displayed (triggered initial revert) - https://github.com/Expensify/App/issues/13002

### Tests

1. Checkout this branch & `npm i`
2. Run the app on both iOS and Android
3. Open a chat with some image attachments & PDF thumbnails
4. They will load in as normal the first time, but reload the app and then you will see the cached images loaded instantly in the same chat

### Offline tests

1. Open a chat and wait for all chat attachment thumbnail images to finish loading
2. Go offline and reload the app and open the same chat - all chat attachment thumbnails should be displayed
3. Go back online and open the attachment view of an image attachment in the chat and let it load
4. Go offline, reload the app and open the same chat and then open the same attachment - this should very briefly show a loading spinner before then displaying the cached image

NOTE: The thumbnail images should be cached for any attachment type, however, when opening the attachment view for an attachment only image attachments will be cached - for example, PDFs will have their preview thumbnail image cached, but the PDF file itself is not cached and thus it is expected behaviour that you will see "Failed to load PDF" when performing steps 3&4 for PDF attachment.

### QA Steps

- [ ] Open a report chat and add a new image attachment and send so the image appears in the chat (this should display a loading indicator the first time it is requested)
- [ ] Reload the app and you should see the cached image displayed (with no loading indicator)
- [ ] Click on an image in the chat to open the attachment modal view, close it and reopen it to see the cached image displayed without a loading state 
- [ ] Verify that any attachments still download as normal without any errors (including image attachments and PDFs)
- [ ] Upload a PDF and the thumbnail should render correctly in chat

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

image

</details>

<details>

Screenshot 2023-01-03 at 3 47 32 PM

</details>
<details>

Screenshot 2023-01-03 at 3 45 09 PM

</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/31568400/210455348-0b372b0a-5ca2-472c-a200-971a05656baa.mov

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/31568400/210451867-a91623f3-1352-49a1-9190-b3a40fa2e3bf.mp4

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/31568400/210453901-cda6f127-e509-4de6-8909-eb106cfaa236.mov

</details>


thomas-coldwell:
I've just done a quick test with the fix that was mentioned [here](https://expensify.slack.com/archives/C01GTK53T8Q/p1669307872628459?thread_ts=1669305509.853359&cid=C01GTK53T8Q) in slack. It does indeed load the images and the `onLoad` event is now fired, however as suggested by other comments in the OG GitHub issue, this applies no compression to the actual image on Android and thus large images cause a crash - I tested with a 10,000 x 5,000 px image (only 1.1mb on disk) and was able to reproduce this.

Test image: https://user-images.githubusercontent.com/31568400/205428950-cd7119fb-cc70-4e76-9518-8027ead43f6f.jpg

I will do some more digging and see if a better fix can be found as we definitely cannot apply this directly if it is going to cause crashes. Also maybe raises a question if we want to ever impose a max resolution for images and videos uploaded as chat attachments cc @Beamanator. Will keep you posted 🙌

thomas-coldwell:
Looks like Glide (the image loading module used in Fast Image on Android) requires a valid, non-zero view size before it can begin image loading as this comment from one of the collaborators suggests:
- https://github.com/bumptech/glide/issues/618#issuecomment-275277165
- https://github.com/bumptech/glide/blob/v4.7.1/library/src/main/java/com/bumptech/glide/request/target/ViewTarget.java#L516
- https://bumptech.github.io/glide/doc/targets.html#view-targets

With that in mind do we still want to class the `onLoad` not firing for zero width/height images on Android as a bug or just how it should work on Android? I know you strive to have the same functionality of components cross-platform

I think the options are from here:
- Rework Glide so it can somehow provide the width/height of the image source and fire the `onLoad` event regardless of view size
- Rework Fast Image to have either a `onImageSizeAvailable` prop at the component level or a static `FastImage.getSizeWithHeaders()` function that can somehow access the natural width and height of the image with Glide, likely through `Targets.SIZE_ORIGINAL`
- Ignore the fact it doesn't load if the view size is zero and then always set the expectation that the view should have a non-zero width/height (which in this specific case will be fixed by the backend change documented here by @Beamanator https://github.com/Expensify/App/pull/12648#issuecomment-1335180031)

Let me know your thoughts @Beamanator @kidroca 🙌

Beamanator:
Hey @thomas-coldwell thanks so much for the in depth investigation!! I believe I'd prefer we go with options 1 or 2 ^ since they're fixing bugs instead of implementing workarounds... It looks like https://github.com/bumptech/glide/ is a pretty active repo so maybe we could get the fix in there, though that issue you linked was opened (and closed) a long time ago 😅  Maybe you can see if there's other related issues that are more recent / still open? And if not, maybe we can start a discussion there to see if it's something they'd consider "fixing"?

thomas-coldwell:
Actually, this seems to be the recommended way of getting the natural dimensions of the image with Glide and ensure that it is only making a single request or decoding this from the cached image https://github.com/bumptech/glide/issues/781#issuecomment-160953996.

One question I have is do we want to always report the natural image dimensions in the `onLoad` event (which would be the case with this fix)? Glide currently resizes images so they have a smaller memory footprint and then reports the resized dimensions of the image like so (this is for a tiny 50x50 dp view):

```
 LOG  RN Image (onLoad) = {
  "height": 1800,
  "width": 1800,
}
 LOG  Fast Image (onLoad) = {
  "width": 132,
  "height": 132
}
```

I'm assuming we do want the natural dimensions so the API is identical with RN core image component? I'm conscious that the image displayed by glide will not actually have these dimensions but I don't think this will have any knock on effects as the width and height are just used to compute image aspect ratio and not directly for other view layouts.

Beamanator:
@thomas-coldwell since there's no current issues / PRs on Glide about this issue, do you think it makes sense to open an issue there to discuss this situation? It's possible those maintainers would recommend we take this issue to the `react-native-fast-image` lib, but I'm inclined to want to start there since that's higher up the chain, what do you think?

Also I am not sure who we'd contact at `@pieter-pot/react-native-fast-image` to get a PR merged... I wonder if we would have to go through the main `react-native-fast-image` lib first and get that passed down. Though, as I'm comparing the two, there's very very very small differences between the fork we use & the main library - honestly it's hard to even see what the differences are, do you see any? I wonder if it's a better idea to see if we can use the main repo instead of the fork?

About the natural image dimensions vs resized dimensions, this is getting a bit far above my understanding, sorry 😅  So far I like your thinking but honestly I don't quite understand the benefits / drawbacks very well  - if you think you're onto something I thinkkkk we should move forward with that and see what library maintainers think and move forward that way?

Also FYI I'm going to start a discussion in Slack (soon) to see if people think we should move forward with a little workaround **while** this conversation / implementation is happening in other repos, then once the fix is implemented & is usable by the New Expensify app, we can remove the workaround. Coming soon :D

thomas-coldwell:
Hey @Beamanator 👋 I don't think it makes sense to open an issue with Glide, glide works as it should loading an image into a target - usually a drawable. This is how Glide is used in Fast Image.

The fix that I've found, and implemented in a separate sandbox app, is to setup an additional request for glide which decodes the image to get the size and then emits this as the `onLoad` event back to JS.

I agree there seems to be [literally no difference](https://github.com/DylanVann/react-native-fast-image/compare/main...pieter-pot:react-native-fast-image:main) between the fork we are currently using and the main `react-native-fast-image` repo so I'd opt that we make a fork of `react-native-fast-image` for Expensify and then we can hopefully PR these changes upstream (though the maintainer of Fast Image is very inactive so the changes might have to stay on our fork, but will try to get it merged upstream nonetheless). Is this possible to get a fork on Expensify or should I just fork it myself? 😃

The natural dimensions are that of the source image we load from the backend e.g. you might have a 1000x1000px image. With Glide and a lot of other image loading libraries they will resize this down appropriately for the target e.g. if its being loaded into a tiny 50x50dp view (like a profile picture) then it will resize it down to maybe 100x100px to preserve memory. I think it makes more sense to do it the same as the RN core Image API and report the natural dimensions so will stick with that 👍

Beamanator:
> The fix that I've found, and implemented in a separate sandbox app, is to setup an additional request for glide which decodes the image to get the size and then emits this as the `onLoad` event back to JS.

That sounnnds great! Do you think we should try to test it with our App before moving forward with that solution? Or was your test solid enough? :D

About the forking question - we reeeeally try not to make our own Forks whenever possible, so what we'll most likely do is try to get the conversation & PR going on the main `react-native-fast-image` repo, and if it goes too long w/ no response we will consider making the Fork. In the meantime I may implement the backend workaround just to get this unstuck

thomas-coldwell:
Perfect, cheers @Beamanator !! I'll create a fork and apply my changes there, open up a PR to the OG repo and then I'll test using my PR branch in the Expensify app as well to make sure there are no obvious issues with this approach 🙌

Beamanator:
That sounds fantastic, thanks so much @thomas-coldwell ! 👍

thomas-coldwell:
I've fixed any merge conflicts there were (as `main` contained the revert changes for the original PR) so this should be good to test with the backend workaround if you wanted to @Beamanator :)

I'm now creating a PR for `react-native-fast-image` with the changes I made to return the natural dimensions from the `onLoad` event. There was [one similar PR](https://github.com/DylanVann/react-native-fast-image/pull/446) already opened there, but they decode the whole bitmap to get the size and use the `.override(Target.SIZE_ORIGINAL)` method for Glide which does not perform any compression and causes crashes on Android (I've verified this is the case in the Expensify app too with a 10,000px image) - their PR seemed to go stale due to these issues being pointed out. My PR is following the recommended way to achieve this as mentioned here https://github.com/bumptech/glide/issues/781#issuecomment-160953996 

I will link this here when it's up 🙂

Beamanator:
Thanks @thomas-coldwell ! I think I'll wait till some resolution on [this convo](https://expensify.slack.com/archives/C01GTK53T8Q/p1671112210394559?thread_ts=1670868002.066289&cid=C01GTK53T8Q) (feel free to add your opinion) before testing w/ my workaround PR :D

thomas-coldwell:
Opened up the PR on the `react-native-fast-image` repo https://github.com/DylanVann/react-native-fast-image/pull/953 🙌

Beamanator:
Reached out to maintainer to see if we can get the upstream PR moving - if no response this week we will have to think about forking

thomas-coldwell:
Established with @puneetlath that a patch of the [upstream PR](https://github.com/DylanVann/react-native-fast-image/pull/953) will work best for now and works perfectly fine with the native code changes. The upstream PR will remain open and hopefully the maintainer will have some time to eventually look at it - if it does get merged upstream this patch can then be removed and upgraded to the latest version

thomas-coldwell:
@kidroca might be good to get your eyes on this as you had comments on the previous PR :)

sketchydroide:
This is a rather big one :D 
Shouldn't this have a C+?

src/components/Image/imagePropTypes.js
```diff
@@ -0,0 +1,56 @@
+import PropTypes from 'prop-types';
+import stylePropTypes from '../../styles/stylePropTypes';
+import RESIZE_MODES from './resizeModes';
+
+const propTypes = {
+
```
mountiny, line 6: Styling
```suggestion
```

src/components/ImageView/index.native.js
```diff
@@ -1,35 +1,45 @@
...
 import withWindowDimensions, {windowDimensionsPropTypes} from '../withWindowDimensions';
 import FullscreenLoadingIndicator from '../FullscreenLoadingIndicator';
+import Image from '../Image';
 
 /**
  * On the native layer, we use a image library to handle zoom functionality
  */
 const propTypes = {
+
+    /** Do the urls require an authToken? */
```
mountiny, line 19:
```suggestion
    /** Whether the urls require an authToken */
```

src/components/AttachmentView.js
```diff
@@ -12,8 +12,13 @@ import compose from '../libs/compose';
 import Text from './Text';
 import Tooltip from './Tooltip';
 import themeColors from '../styles/themes/default';
+import addEncryptedAuthTokenToURL from '../libs/addEncryptedAuthTokenToURL';
 
 const propTypes = {
+
+    /** Do the urls require an authToken? */
```
mountiny, line 19: NAB: phrasing
```suggestion
    /** Whether the urls require an authToken */
```

src/components/ImageWithSizeCalculation.js
```diff
@@ -16,11 +17,15 @@ const propTypes = {
 
     /** Callback fired when the image has been measured. */
     onMeasure: PropTypes.func,
+
+    /** Does the image require an authToken? */
```
mountiny, line 21:
```suggestion
    /** Whether the image require an authToken */
```

src/components/ThumbnailImage.js
```diff
@@ -16,7 +15,7 @@ const propTypes = {
     // eslint-disable-next-line react/forbid-prop-types
     style: PropTypes.any,
 
-    /** Do the urls require an authToken? */
+    /** Does the image require an authToken? */
```
mountiny, line 18: NAB
```suggestion
    /** Whether the image require an authToken */
```

mountiny (commented):
@thomas-coldwell Code looks good to me, had couple of comment NABs.

The PR author checklist seems to be failing, can you please use the latest PR author checklist and fill it in? I will try to test tomorrow.

mananjadhav:
Going through the comments and the code. (also from the previous PR before getting into the testing). Will have the review and testing completed by tomorrow.

src/CONST.js
```diff
@@ -874,6 +874,8 @@ const CONST = {
     },
 
     TFA_CODE_LENGTH: 6,
+
```
Beamanator, line 877: NAB - probably don't need this extra space

```suggestion
```

src/components/AttachmentModal.js
```diff
@@ -115,7 +115,7 @@ class AttachmentModal extends PureComponent {
      * @param {String} sourceURL
      */
     downloadAttachment(sourceURL) {
-        fileDownload(sourceURL, this.props.originalFileName);
+        fileDownload(this.props.isAuthTokenRequired ? addEncryptedAuthTokenToURL(sourceURL) : sourceURL, this.props.originalFileName);
```
Beamanator, line 118: Just noting that @kidroca add a comment about also being able to let `fileDownload` work with headers on the old PR: https://github.com/Expensify/App/pull/12648#discussion_r1032172931

Maybe we can add this in a follow-up PR?

src/components/Image/imagePropTypes.js
```diff
@@ -0,0 +1,55 @@
...
+import stylePropTypes from '../../styles/stylePropTypes';
+import RESIZE_MODES from './resizeModes';
+
+const propTypes = {
+    /** Styles for the Image */
+    style: stylePropTypes,
+
+    /** The static asset or URI source of the image */
+    source: PropTypes.oneOfType([
+        PropTypes.number.isRequired,
```
Beamanator, line 11: Just wondering, is `isRequired` here? I don't see it after the next `PropTypes.shape` so I am guess it's either needed on both or neither?

src/components/Image/imagePropTypes.js
```diff
@@ -0,0 +1,55 @@
...
+const defaultProps = {
+    session: {
+        authToken: null,
+    },
+    isAuthTokenRequired: false,
+    source: null,
+    resizeMode: RESIZE_MODES.cover,
+    onLoadStart: () => {},
+    onLoadEnd: () => {},
+    onLoad: () => {},
```
Beamanator, line 52: Do we need `style: []` here? It's not required in `stylePropTypes`

src/components/Image/index.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    componentDidUpdate(prevProps) {
+        if (prevProps.source === this.props.source) {
+            return;
+        }
+        this.configureImageSource();
+    }
+
+    configureImageSource() {
```
Beamanator, line 29: Can you add some comments explaining what this does?

src/components/Image/index.js
```diff
@@ -0,0 +1,70 @@
...
+            imageSource = {uri: `${source.uri}?encryptedAuthToken=${encodeURIComponent(authToken)}`};
+        }
+        this.setState({imageSource});
+
+        // If an onLoad callback was specified then manually call it and pass
+        // the natural image dimensions to match the native API
+        if (this.props.onLoad == null) {
+            return;
+        }
+        const uri = typeof imageSource === 'number'
```
Beamanator, line 44: I see two times we check if `imageSource === 'number'` - maybe we can me this a variable like `imageSourceIsNumber`?

src/components/Image/index.js
```diff
@@ -0,0 +1,70 @@
...
+Image.propTypes = propTypes;
+Image.defaultProps = defaultProps;
+
+const ImageWithOnyx = withOnyx({
+    session: {
+        key: ONYXKEYS.SESSION,
+    },
+})(Image);
+ImageWithOnyx.resizeMode = RESIZE_MODES;
+export default ImageWithOnyx;
```
Beamanator, line 70: Just wondering, why can't we do:

```suggestion
Image.resizeMode = RESIZE_MODES;
export default withOnyx({
    session: {
        key: ONYXKEYS.SESSION,
    },
})(Image);
```

src/components/Image/index.native.js
```diff
@@ -0,0 +1,64 @@
...
+    componentDidUpdate(prevProps) {
+        if (prevProps.source === this.props.source) {
+            return;
+        }
+        this.configureImageSource();
+    }
+
+    configureImageSource() {
+        const source = this.props.source;
+        const isAuthTokenRequired = this.props.isAuthTokenRequired;
```
Beamanator, line 32: NAB - not needed since we only use `this.props.isAuthTokenRequired` once

src/components/Image/index.js
```diff
@@ -0,0 +1,70 @@
...
+    componentDidUpdate(prevProps) {
+        if (prevProps.source === this.props.source) {
+            return;
+        }
+        this.configureImageSource();
+    }
+
+    configureImageSource() {
+        const source = this.props.source;
+        const isAuthTokenRequired = this.props.isAuthTokenRequired;
```
Beamanator, line 31: Similarly, can we just use `this.props.isAuthTokenRequired` below instead of making this new const?

src/components/Image/index.js
```diff
@@ -0,0 +1,70 @@
...
+        let imageSource = source;
+        if (typeof source !== 'number' && isAuthTokenRequired) {
+            const authToken = lodashGet(this.props, 'session.encryptedAuthToken', null);
+            imageSource = {uri: `${source.uri}?encryptedAuthToken=${encodeURIComponent(authToken)}`};
+        }
```
Beamanator, line 36: Can we add a comment here saying right now there's a `react-native-web` bug prevent us from adding headers to images for web / desktop? And that's why we do this here, but we add `headers` in `index.native.js`?

src/components/ImageView/index.js
```diff
@@ -1,20 +1,27 @@
...
+import {View, Pressable} from 'react-native';
+import Image from '../Image';
 import styles from '../../styles/styles';
 import * as StyleUtils from '../../styles/StyleUtils';
 import canUseTouchScreen from '../../libs/canUseTouchscreen';
 import withWindowDimensions, {windowDimensionsPropTypes} from '../withWindowDimensions';
 import FullscreenLoadingIndicator from '../FullscreenLoadingIndicator';
 
 const propTypes = {
+
```
Beamanator, line 12: NAB

```suggestion
```

src/components/ImageView/index.native.js
```diff
@@ -107,44 +82,38 @@ class ImageView extends PureComponent {
         return false;
     }
 
-    imageLoadingStart() {
-        this.setState({isLoading: true});
+    imageLoad({nativeEvent}) {
```
Beamanator, line 85: Can you add a comment explaining what this function does / its purpose?

src/components/ImageView/index.native.js
```diff
@@ -1,35 +1,44 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {
-    View, InteractionManager, PanResponder,
+    View, PanResponder, InteractionManager,
```
Beamanator, line 4: NAB - this change isn't needed

```suggestion
    View, InteractionManager, PanResponder,
```

src/components/ImageView/index.native.js
```diff
@@ -107,44 +82,38 @@ class ImageView extends PureComponent {
         return false;
     }
 
-    imageLoadingStart() {
-        this.setState({isLoading: true});
+    imageLoad({nativeEvent}) {
+        // Wait till animations are over to prevent stutter in navigation animation
+        this.state.interactionPromise = InteractionManager.runAfterInteractions(() => {
```
Beamanator, line 87: I wonder if we should just create a class variable instead of attaching this to state? Any reason one way is better than the other?

mountiny:
Having some issues with sending attachments locally

src/components/ThumbnailImage.js
```diff
@@ -16,7 +15,7 @@ const propTypes = {
     // eslint-disable-next-line react/forbid-prop-types
     style: PropTypes.any,
 
-    /** Do the urls require an authToken? */
+    /** Whether the image require an authToken */
```
Beamanator, line 18:
```suggestion
    /** Whether the image requires an authToken */
```

src/components/ImageWithSizeCalculation.js
```diff
@@ -16,11 +17,15 @@ const propTypes = {
 
     /** Callback fired when the image has been measured. */
     onMeasure: PropTypes.func,
+
+    /** Whether the image require an authToken */
```
Beamanator, line 21:
```suggestion
    /** Whether the image requires an authToken */
```

Beamanator (changes requested):
Lots of super tiny comments, overall looking super rad!

Also just wondering, how did you generate the `patches` file, @thomas-coldwell ? Just curious cuz it looks complicated :D

src/components/ThumbnailImage.js
```diff
@@ -87,9 +86,7 @@ class ThumbnailImage extends PureComponent {
     }
 
     render() {
-        const url = this.props.isAuthTokenRequired
-            ? addEncryptedAuthTokenToURL(this.props.previewSourceURL)
-            : this.props.previewSourceURL;
+        const url = this.props.previewSourceURL;
```
mananjadhav, line 89: This is only used at one place, I think we can just use `this.props.previewSourceURL`.

mananjadhav:
I see many of the styling comments raised already hence I've added only one that wasn't added in the previous reviews. Still reviewing. Might need 1 more day to review and test.

src/components/AttachmentModal.js
thomas-coldwell: Yes I think a follow up PR is the right place for this so its not blocking the rest of this :) Thanks for keeping a note of that 🙌

src/components/Image/imagePropTypes.js
thomas-coldwell: Yes good catch - its required to either be an object with a URI or a `require()` (a number) so I've set it for both now :)

src/components/Image/index.js
thomas-coldwell: In the suggestion, due to the HoC, the static `resizeModes` is lost from the top level export so you'd get undefined if you tried to access `Image.resizeModes` :)

mountiny (commented):
@thomas-coldwell Got this error when testing on iOS simulator when I have attempted to upload an image
image


Actually it is this photo which does this:
image

The other ones worked fine too

PDF worked fine 🎉

src/components/ImageView/index.native.js
thomas-coldwell: I think that state was there before this PR, but it should be reactive to changes such as [here](https://github.com/margelo/expensify-app-fork/blob/f97273b6700252e1431250d7c7a71f91e5d65fe9/src/components/ImageView/index.native.js#L178) where its used to keep the image hidden while layout calculations are performed

thomas-coldwell:
Chatted with @mountiny on slack [here](https://expensify.slack.com/archives/C035J5C9FAP/p1672927588499199?thread_ts=1672790743.169889&cid=C035J5C9FAP) and looks like this above error comes from the attachment picker process and not the changes in this PR - will do another PR to investigate this issue (or check if anyone has opened an issue for it already) 🙌

src/components/Image/index.js
thomas-coldwell: I've removed this as its not required for web - its only on native that metro resolves this to a number and not on web https://facebook.github.io/metro/docs/getting-started#assets. I've updated the propTypes to reflect this change too :)

src/components/ImageView/index.native.js
Beamanator: Aah ok cool sounds good to me! Let's keep it :D

src/components/Image/index.js
Beamanator: Woah interesting! Good to know!

Beamanator (dismissed):
Tested on Android & iOS and looks great!

Android:

https://user-images.githubusercontent.com/3885503/210824748-e22ab3bf-58a1-4c0b-922b-d7de74f696a9.mov

iOS:


https://user-images.githubusercontent.com/3885503/210827529-fa7a5419-77c1-40e4-a119-8523a0d66a7b.mov

(I won't merge or anything till more reviews are done)

mananjadhav:
I am close to 70% done, will start testing in an hour or so. Should have all the screencasts in 2 hours.

mananjadhav (dismissed):

mananjadhav:
@thomas-coldwell Is this an acceptable behavior when I open the image it shows a loader everytime? for iOS and Android?


https://user-images.githubusercontent.com/3069065/210849653-305858ca-d81a-42e5-974e-03849e3f5963.mov


https://user-images.githubusercontent.com/3069065/210849756-6a8fe0a4-c4d7-4ede-b4fc-90ee29341d1b.mov

thomas-coldwell:
Hey @mananjadhav there should be a very small delay where it is showing the loading indicator as it has to to perform some layout calculations after the image has already loaded as this is required by the parent `ImageZoom` component.

thomas-coldwell:
But on iOS some of those recordings seem to take a fairly long time so I will take a look and confirm there it is definitely getting the image from cache in this situation

thomas-coldwell:
I've made that change - was a lot more simple than I thought. The `onProgress` is only called if it is downloading an image and not from cache so now the loading state is only set if it actually is downloading and not getting it from the cache:

https://user-images.githubusercontent.com/31568400/210869498-40edbaef-7fc5-4794-9ad0-521059f2588e.mp4

thomas-coldwell:
Also just checked the other case where the loading indicator is shown when you have uploaded an image, it is displayed in the chat and then you got onto the attachment view and it appears to reload. The cache it still working correctly here its simply that the thumbnail URL differs with a `.1024.jpg` append and the image in the attachment view does not have this in the URL and thus they are separate images so it will have to re-download it and then subsequent visits to this image in the attachment view will show the cached copy.

Just wanted to document this in case there was any questions around this other loading state 🙌

puneetlath (dismissed):

Beamanator:
Thanks for the updates & explanations @thomas-coldwell 👍 👍  I'll do another review shortly 🙏

kidroca (commented):
Left a few minor notes

I see the PR is already approved so I want to say that I don't put much weight on my notes, but it would be nice to address them now

ios/Podfile.lock
```diff
@@ -722,7 +722,7 @@ DEPENDENCIES:
   - "RNCClipboard (from `../node_modules/@react-native-community/clipboard`)"
   - "RNCPicker (from `../node_modules/@react-native-picker/picker`)"
   - "RNDateTimePicker (from `../node_modules/@react-native-community/datetimepicker`)"
-  - "RNFastImage (from `../node_modules/@pieter-pot/react-native-fast-image`)"
+  - RNFastImage (from `../node_modules/react-native-fast-image`)
```
mountiny, line 725: Out of curiosity, why are some of these wrapped in "" and others are not? No difference?

src/components/AttachmentView.js
```diff
@@ -12,8 +12,12 @@ import compose from '../libs/compose';
 import Text from './Text';
 import Tooltip from './Tooltip';
 import themeColors from '../styles/themes/default';
+import addEncryptedAuthTokenToURL from '../libs/addEncryptedAuthTokenToURL';
 
 const propTypes = {
+    /** Whether the urls require an authToken */
```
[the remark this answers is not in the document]
mountiny: This comment is really controversial, changed probably 3 times already 😄

src/components/Image/index.native.js
```diff
@@ -0,0 +1,63 @@
...
+        let imageSource = source;
+        if (typeof source !== 'number' && this.props.isAuthTokenRequired) {
+            const authToken = lodashGet(this.props, 'session.encryptedAuthToken', null);
+            imageSource = {
+                ...source,
+                headers: authToken ? {
+                    [CONST.CHAT_ATTACHMENT_TOKEN_KEY]: authToken,
+                } : null,
+            };
+        }
```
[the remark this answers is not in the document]
mountiny: That indeed looks more compact and I would not say it is harder to understand/read

src/components/ImageView/index.native.js
```diff
@@ -107,44 +82,40 @@ class ImageView extends PureComponent {
...
-    imageLoadingStart() {
+    imageProgress() {
         this.setState({isLoading: true});
     }
 
-    imageLoadingEnd() {
-        this.setState({isLoading: false});
+    // Handles the `onLoad` event when the image loads providing the natural
+    // image dimensions required for layout calculations
+    imageLoad({nativeEvent}) {
```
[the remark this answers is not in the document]
mountiny: +1

mountiny (commented):
@thomas-coldwell Great updates yesterday, and I agree with Kidroca's comments so not approving yet.

src/components/Image/imagePropTypes.js
```diff
@@ -0,0 +1,59 @@
...
+
+    /** The static asset or URI source of the image */
+    source: PropTypes.oneOfType([
+        PropTypes.number.isRequired,
+        PropTypes.shape({
+            uri: PropTypes.string.isRequired,
+            // eslint-disable-next-line react/forbid-prop-types
+            headers: PropTypes.object,
+        }).isRequired,
+    ]),
```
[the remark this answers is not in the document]
thomas-coldwell: Agreed - I've removed it as an optional prop with a default of `null` and updated the prop types.

src/components/Image/index.native.js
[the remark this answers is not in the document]
thomas-coldwell: Sounds good - changing it to a FC now. I think this was originally a FC, but had some side effects that needed to be run (in the web side component)

src/components/Image/index.native.js
```diff
@@ -0,0 +1,63 @@
...
+        this.setState({imageSource});
+    }
+
+    render() {
+        // eslint-disable-next-line
+        const { source, ...rest } = this.props;
+
+        // eslint-disable-next-line
+        return <RNFastImage {...rest} source={this.state.imageSource} />;
```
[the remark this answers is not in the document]
thomas-coldwell: Solved by the above change to FC :)

ios/Podfile.lock
thomas-coldwell: Yeah there shouldn't be any difference at all there - I think its simply a difference between XCode versions or something like that, but doesn't actually effect the pod in anyway :)

puneetlath (dismissed):

mananjadhav (dismissed):

mananjadhav:
@thomas-coldwell I've two comments on the testing:

1. When I upload and view smaller image, then it resizes on the attachment modal. This is an issue with iOS and Android.

https://user-images.githubusercontent.com/3069065/211095672-fe05df89-8508-4680-86e6-f1cf67b927c9.mov

2. Not sure if this is an issue, but on Android when I open the image in the attachment modal, I see a delay in the loading indicator. Are you seeing the same behavior?

https://user-images.githubusercontent.com/3069065/211096039-47ecaf5d-41bb-46fa-8c17-37018ed4a30e.mov

--------

@Beamanator @puneetlath I've reviewed and tested this for the following:

1. Small, regular and Large Images
3. Offline loading state
4. PDF and Password Protected PDF
5. Other attachment types
6. File and Image Downloads

Do we have any regression suite that can validate if we need to cover something

thomas-coldwell:
Thanks @mananjadhav! I believe the second issues is caused by changing `isLoading` logic. Looks like I need to add a check in the `onLoadStart` to see if the image is coming from cache or not and then only set `isLoading` to true if Glide / SDWebImage is downloading it as otherwise there will _always_ be a loading indicator. The first issue is related to this so will make a change and resolve both 🙌

thomas-coldwell:
Hey @mananjadhav this should be ready for a review again and you should see those two issues from the last review resolved now, cheers 🙌

mananjadhav:
Thanks @thomas-coldwell. I’ll be reviewing it today.

mananjadhav:
Hi @thomas-coldwell,
The earlier issue of image getting resized on view is resolved.

But now a flicker is introduced with the last change. You'll notice a small flicker, where actually the image is previewed, then a loader is shown and then the image loads when it is already cached. This issue is only with iOS and not with Android. . Are you seeing this? Attaching the screencast below.

https://user-images.githubusercontent.com/3069065/211210885-eeefb512-c08b-4d56-babe-907395d2d301.mov

----

Though on Android I have a question, will it always a show a loader even when the image is cached? It shows a loader very quick, where I am assuming `RNFastImage` calculates the image size?

https://user-images.githubusercontent.com/3069065/211210972-e24802d9-d684-4383-ab18-f8da8e5573b6.mov

mananjadhav:
@thomas-coldwell Quick bump on the comments above.

thomas-coldwell:
Thanks @mananjadhav - yeah I think this problem is being caused by the `ImageZoom` component as that requires the width/height of the image, but this is only available once the image has loaded as thus there will always be some small flicker. At the minute the image is hidden until the image dimensions are set for the `ImageZoom` which is what causes the flicker - ideally we would need to replace the `ImageZoom` component so that it does not require knowledge of the image dimensions OR we just always show some loading state (whether that's showing the indicator or just hiding the image) while its performing this layout calculation. What are your thoughts @Beamanator ? I think this is probably linked to https://github.com/Expensify/App/issues/12893

Beamanator:
@thomas-coldwell Yeah ideally we won't have the flicker where it goes from image to loader to image (which I think is currently only happening [on iOS](https://github.com/Expensify/App/pull/13304#issuecomment-1374890318)). So I agree it would be best to show the loader while layout calculation is happening - is this something easily done in this PR?

thomas-coldwell:
Alright, I've updated this so it always shows the loading spinner for actual image load + layout calculation. 

In a follow-up PR (or the PR for https://github.com/Expensify/App/issues/12893) this can ideally be changed so it only shows a loading state when the image is loading and more specifically only when its having to download the image the first time to cache. As a result I've left the code that supplies the `cachePath` in the `onLoad` event for any follow up work 🙂

thomas-coldwell:
Also just to double check, I've made the initial width/height zero in the `ImageView` on native - I had left a comment before that this should use large values to avoid the `ImageZoom` component causing lower resolutions images to be displayed but I've tested with some high resolution images and it doesn't seem to cause any of these issues on either platform.

Let me know if you have any concerns about this or if you know it will likely cause any kind of regression here that I'm not aware of :)

mountiny:
@thomas-coldwell thanks for the updates on this!

@mananjadhav Will you be able to retest this based on the updated behaviour?

mananjadhav:
@mountiny I saw the update last night, but I wasn't sure if we agreed with the behavior. Can you or @Beamanator confirm if the approach is okay then I can test today and close it.

src/components/Image/index.js
```diff
@@ -0,0 +1,82 @@
...
+import {defaultProps, imagePropTypes} from './imagePropTypes';
+import RESIZE_MODES from './resizeModes';
+
+const propTypes = {
+    ...imagePropTypes,
+
+    /** The URI source of the image */
+    source: PropTypes.shape({
+        uri: PropTypes.string.isRequired,
+    }).isRequired,
```
Beamanator, line 16: I believe this is duplicated in the `imagePropTypes` file so we can remove this?

Or are you overwriting it since we should only ever pass a source with `uri` here, where on mobile we can pass a source with `uri` & `headers` OR a source that's just a `number`?

src/components/Image/index.js
```diff
@@ -0,0 +1,82 @@
...
+        if (prevProps.source === this.props.source) {
+            return;
+        }
+        this.configureImageSource();
+    }
+
+    // Check if the image source is a URL - if so the `encryptedAuthToken` is appended
+    // to the source. The natural image dimensions can then be retrieved using this source
+    // and as a result the `onLoad` event needs to be maunually invoked to return these dimensions
+    configureImageSource() {
```
Beamanator, line 42: This should also be a "method doc" type of a comment

```suggestion
    /**
     * Check if the image source is a URL - if so the `encryptedAuthToken` is appended
     * to the source. The natural image dimensions can then be retrieved using this source
     * and as a result the `onLoad` event needs to be maunually invoked to return these dimensions
     */
    configureImageSource() {
```

Beamanator (commented):
Hmm I'm seeing this (in Android only) when I view a PDF attachment online, then go offline (wait for the simulator to realize it's offline), then try to view the same PDF attachment - did nobody else see the same results? 😅 

Screenshot 2023-01-12 at 1 13 11 PM

mananjadhav:
I didn't reach the offline tests for Android/iOS. Will test and update.

src/components/Image/index.js
thomas-coldwell: Yes this was just a duplicate - removed it :)

thomas-coldwell:
@Beamanator I believe that is the expected behaviour for a PDF when offline as PDFs are not cached - only image attachments. Though I believe that would certainly be a good follow-up PR to add full chat attachment cache support

Beamanator:
Aah ok @thomas-coldwell so that would probably be the same for all non-image attachments i guess? It would be great to fix that in a folow up :D

thomas-coldwell:
Yeah all caching is handled internally by Fast Image so we'd need a separate mechanism for this non-image cache 👍

Beamanator:
Aah right good point @thomas-coldwell - that's out of scope for now 😅

mananjadhav:
## Reviewer Checklist

-  [x] I checked that screenshots or videos are included for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/3069065/212072746-70b081c6-3fad-41ca-a637-f54ce1f009bd.mp4

</details>

<details>

https://user-images.githubusercontent.com/3069065/212071367-30e58556-f80a-4dbd-9f02-c7f7513dc103.mov

</details>

<details>

https://user-images.githubusercontent.com/3069065/212071516-e51d1188-46ce-4e5c-80e0-8e87ad40a452.mov

</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/3069065/212071776-19f42769-60e8-4a3b-8825-8c6ef23d214a.mov

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/3069065/212071991-da5cb57d-215c-4343-bf78-dfcf0d647157.mov

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/3069065/212072304-45ebe8cb-3197-465f-a908-9d4f6bb3a702.mp4

</details>


--- 

@Beamanator Based on comments with @thomas-coldwell, I am ignoring offline PDF.  `Force offline` doesn't work as expected it seems and hence the PDF loaded for me. I turned off the Wifi on the emulator to reproduce the failed offline behavior. Also, even if I turn off Mac Wifi, the iOS Simulator still loaded the PDF (not sure how).

I also think we should add this as a Tests so that QA is on the same page.

mananjadhav (dismissed):

Beamanator (dismissed):
This looks great to me 👍  Though I'd like if @kidroca could also give it one last look if he has a chance today, then I'd be even more confident about this :D

Beamanator:
@thomas-coldwell do you mind updating the Offline test a bit in your OP? @mananjadhav makes a good point that we want QA to be 100% on the same page as what we've decided, and right now I don't think the offline test case is very clear what we expect to happen

thomas-coldwell:
Thanks @Beamanator - I've updated the OP to reflect this - very good points for them to be aware of!

Beamanator:
Looks fantastic now, thanks so much @thomas-coldwell 👍

mountiny (dismissed):
Really great job on this one @thomas-coldwell and @mananjadhav with the thorough testing 🙇 

Leaving this for @kidroca to also give it a look!

puneetlath (dismissed):

mananjadhav:
@Beamanator @thomas-coldwell @puneetlath Anything pending here?

thomas-coldwell:
I think we were waiting to get one last review from @kidroca - I'll ping them on slack

src/components/ImageWithSizeCalculation.js
```diff
@@ -39,68 +44,29 @@ class ImageWithSizeCalculation extends PureComponent {
...
-            });
-    }
-
-    imageLoadingStart() {
-        this.setState({isLoading: true});
-    }
-
-    imageLoadingEnd() {
-        this.setState({isLoading: false});
+        this.props.onMeasure({width: event.nativeEvent.width, height: event.nativeEvent.height});
```
kidroca, line 69: This check is contradicting:

1. `imageLoadSuccessfuly` is called from `onLoad` 
2. `onLoad` would be triggered only on success 

If that's not the case, we should address the problem in `onLoad` instead of doing a check here

Simply said - `onLoad` shouldn't be triggered when the image didn't load properly 

---

Seems this check is carried over from the original code
And it was introduced a long time ago here: 091abdc44b73bc9dc5880fbabac1855b8d925575

Like I said - if there's ever a case where `onLoad` (or `getSize`) was called without `width` and `height` - it is a bug in the underlying library and not something that should be handled in App 

_(For what I know for RN web - there won't be a case where `getSize` would not return `width` or `height`
@thomas-coldwell do you know whether `FastImage` needs this check?)_

@Beamanator @mountiny @tgolen what do you think guys?
- drop the check
- handle in a follow up
- do nothing / not big deal

---

I think this might be easier to understand:

```js
const {width, height} = lodashGet(event, 'nativeEvent' {});
if (!width || !height) {
// Image didn't load properly
return;

this.props.onMeasure({ width, height });
```

Or, if we don't want to destruct values
```js
const nativeEvent = lodashGet(event, 'nativeEvent' {});
if (!nativeEvent.width || !nativeEvent.height) {
...
```

kidroca (dismissed):
All good, but something I saw in `ImageWithSizeCalculation` - left one big comment there

src/components/ImageWithSizeCalculation.js
```diff
@@ -39,68 +44,29 @@ class ImageWithSizeCalculation extends PureComponent {
...
-                reject(error);
-            });
-        });
+    imageLoadingEnd() {
+        this.setState({isLoading: false});
     }
 
-    calculateImageSize() {
-        if (!this.props.url) {
+    imageLoadedSuccessfuly(event) {
```
kidroca, line 63: There's a typo:
- Successfuly
- Successful**l**y

```suggestion
    imageLoadedSuccessfully(event) {
```

kidroca:
I've tested (rebased) these changes with my PR for Web Images With Headers and they are working fine

thomas-coldwell:
Hey @kidroca thanks for the testing with your RNW PR! Agree on the comment on `ImageWithSizeCalculation` - the `onLoad` event for Fast Image will **only** be called if the image is successfully downloaded or retrieved from cache with a valid width/height, else it will call its `onError` event. 

With that in mind I don't think we actually need the check here and it can just be removed 👍

mountiny (approved):
Great @thomas-coldwell and @kidroca I think we are now all happy and this can be merged as we had 5 approvals before.

Is there anything else @kidroca which you would like to address?

Beamanator (approved):
Thanks for the quick response & fix @thomas-coldwell 👍 

I think we should do one more round of testing before we merge 🙏

mananjadhav:
Yeah I agree. I am going to test this in a while @Beamanator. Give me like 30-45 mins.

Beamanator:
Amazing, thanks so much @mananjadhav 👍

mountiny:
Amazing @mananjadhav

mananjadhav:
@thomas-coldwell , @Beamanator 

Quick update, I just tested Desktop and Web. The caching seems to be working fine. I won't be able to do offline tests as I can't disconnect my laptop from Wifi right now. While I am testing other platforms, I found one issue.

I missed to check this earlier. When I download the images and PDF, the file name of the downloaded file is correct. But when I download a word document, the file name of the downloaded doc, doesn't match the original file name. Attaching the video.

https://user-images.githubusercontent.com/3069065/212859216-36a86db4-3db8-46ac-b45b-3c24f97d894a.mov

mananjadhav:
@thomas-coldwell @Beamanator So I did all the testing for all the platforms and the caching works fine.

I've got two concerns/issues:

1. File Download name isn't working as expected so we'll have to fix that (which I posted in the previous comment).

2. If I am offline, and I open a new image, what's the expected output for the Attachment Preview? Android has just a loader, and Web has a blank screen. PDF shows `Failed to load PDF file.`.  And I think that's how it is in the current prod/main version too.

offline-web-new-image


https://user-images.githubusercontent.com/3069065/212867354-77212604-7f55-4d53-90a7-f7f4dbc2739e.mov

Beamanator:
Thanks for the tests and notes @mananjadhav !

In my opinion:
1. File Download - I would be completely fine opening a follow-up issue to investigate this, I don't see how this PR could have broken that, especially if it's only for word documents
2. As for chat attachment images in the Attachment Preview - I think we'd expect to be able to see the images, right @thomas-coldwell ?
    - Note: For PDFs, we don't expect to see the PDF (see notes starting [here](https://github.com/Expensify/App/pull/13304#pullrequestreview-1245421643))

mananjadhav:
Okay for 1.

for 2, what I meant is opening the image for the first time in the offline mode. I am inclined to solve this also as a follow up Pr because the behavior is same as prod, but just wanted to highlight it here.

Beamanator:
@mananjadhav Aah because the image hasn't been viewed yet so it probably hasn't been cached either? I'd at least expect a consistent thing to happen across apps - but yeah as @thomas-coldwell said if this is consistent with production, let's discuss outside this PR - probably in Slack to get some more eyes 👍

mananjadhav:
Thanks for confirming @thomas-coldwell.

Agreed to keep both of them outside the scope of this PR. All good here then.

Beamanator:
FYI it looks like the failing test is not related to this PR, and it was noted in Slack here: https://expensify.slack.com/archives/C01GTK53T8Q/p1673916481148719

So I'm going to merge 👍

Beamanator:
Mentioned above that 1 test wasn't passing, but it's not related to this PR and is being investigated in [Slack](https://expensify.slack.com/archives/C01GTK53T8Q/p1673916481148719) and has this issue for investigating: https://github.com/Expensify/App/issues/14352

PR #13419, opened 2022-12-07
Rebuild unread reports without sequence numbers

puneetlath:
On HOLD until https://github.com/Expensify/Web-Expensify/pull/35907 is live. 

### Details

This PR rebuilds the unread functionality to use reporAction IDs and timestamps instead of sequence numbers. At a high-level: 
1. Every report action has a unique `reportActionID` (non-sequential) 
2. Every report action has a `created` field with millisecond precision 
3. Every report has a `lastReadTime` indicating the last time that report was read
4. Every report was a `lastActionCreated` indicating the `reportAction.created` of the last visible report action on the report
5. A report is unread if `lastReadTimes` < `lastActionCreated`
6. To mark a given action as unread, we set the `lastReadTime` to 1 millisecond before the `reportAction.created` of that action 
7. We display the "new" message indicator above the oldest action whose `reportAction.created` is after the `lastReadTimestamp`

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/240916
$ https://github.com/Expensify/App/issues/13472
$ https://github.com/Expensify/App/issues/13684

### Tests

1. Create two accounts, A and B, that have never chatted before.
1. Send a DM from B to A. The chat should be:
    - Unread for A
    - Read for B
1. User A opens the chat. The chat should become Read
1. User A sends a message in the chat. It should remain Read
1. When user A has the chat fully visible and scrolled to the bottom, user B sends a message to user A. The chat should remain Read for both users.
1. User A sends a bunch of messages, such that they can scroll back in the chat and the bottom messages aren't visible. User A then scrolls back up in the chat.
1. User B sends a message to user A. The chat should become unread for user A.
1. User B sends two more messages to user. The chat should remain unread.
1. User A scrolls back down to the bottom of the chat. It should stay unread for now, and a new message indicator should appear above the first message that user B sent while user A was scrolled up.
1. (narrow screens) User A goes to the LHN, _opens a different chat_,  then goes back to the LHN. The chat with user B should now show as read.
1. (narrow screens) User A returns to the DM with user B and starts uploading a large attachment.
1. (narrow screens) _Before the upload finishes,_ User A goes back to the LHN.
1. (narrow screens) Wait for a bit until the upload completes. Verify that the chat remains read and does not become unread.
    - _Note:_ The past three steps cover [this issue](https://github.com/Expensify/App/issues/13684)
1. User A navigates to a different chat.
1. User B sends a message to User A. The chat should become unread for User A.
1. User A opens the chat from user B, then navigates to a different chat, then comes back. It should now be read.
1. User A scrolls up, marks a given message as Unread. The chat should become unread and an unread indicator should appear above the message that was marked as unread.
1. User A leaves the chat, opens a different one, then comes back. The unread marker should be where it was in the previous step, and the chat should become read.
1. User A opens a different chat, then comes back. The unread marker should be gone.
1. User A scrolls back, then marks one of their own messages in the middle of the chat as unread. The chat should become unread and a new message indicator should appear above that message.
1. User A deletes the message they marked as unread. 
1. User A opens a different chat, then comes back. The new message indicator should be above the message after the deleted message, and the chat should become read.
1. User A opens a different chat, then comes back. The new message indicator should be gone.
1. User A sends a message in the chat, then marks it (i.e: the newest message in the chat) as unread. The chat should become unread and the new message indicator should appear above that message.
1. User A then deletes the message they marked as unread. The chat should become read and the new message indicator should disappear.
1. User A sends a message in the chat, then marks it (i.e: the newest message in the chat) as unread. The chat should become unread (bolded) and the new message indicator should appear above that message.
1. User A navigates to a different chat, closes the app completely (by closing the tab (web), pressing the home button (mobile), or exiting electron (desktop)) and reopens it. The chat/message from the previous step should still be unread.
1. User A opens the chat, opens a different chat, then comes back, and the message should be read.
1. User A closes the app completely and reopens it. The chat should remain read.

### Offline tests

1. User A goes offline.
1. User B sends a DM to user A. User A is offline so their chat should remain unchanged.
1. User A sends an (offline) DM to User B. User A should see the chat as read.
1. User A comes back online. Since their message was sent most recently, the chat should appear as read.
    - _Potential improvement_: As described [here](https://docs.google.com/document/d/1xYZfYX3HoNRF9iZOQJH4Tt7GvPgdS4eTMSRX27tQpl4/edit#bookmark=id.o6ckuehr9mu), we could change the unread feature such that the chat appears unread for User A in this scenario, but that's out-of-scope for now.

### QA Steps

Run through any existing regression tests that deal with unread chat functionality. 

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/1371996/209873984-f890afb0-be50-405f-ae30-0c7c57b50291.mov

</details>

<details>

https://user-images.githubusercontent.com/1371996/210107822-5ee60cd3-23c3-4b5e-995c-a192f7b1bb09.mov

</details>

<details>

https://user-images.githubusercontent.com/1371996/209993346-e02fd823-4870-4ea0-8d86-cf8761e2aa6e.mov

</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/1371996/209875445-a77b1395-25f6-43d1-8698-d8fbbbe41dc8.mov

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/1371996/209991593-40239a09-0e1c-4e16-a38f-c027cb040747.mov

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/1371996/210106859-3131c142-a4f3-41ef-9c7c-f0a23140e833.mov

</details>


src/pages/home/report/ReportActionsView.js
```diff
@@ -102,28 +100,28 @@ class ReportActionsView extends React.Component {
 
         // This callback is triggered when a new action arrives via Pusher and the event is emitted from Report.js. This allows us to maintain
         // a single source of truth for the "new action" event instead of trying to derive that a new action has appeared from looking at props.
-        this.unsubscribeFromNewActionEvent = Report.subscribeToNewActionEvent(this.props.report.reportID, (isFromCurrentUser, currentLastSequenceNumber) => {
+        this.unsubscribeFromNewActionEvent = Report.subscribeToNewActionEvent(this.props.report.reportID, (isFromCurrentUser, currentLastReportActionID) => {
```
puneetlath, line 103: I'm not exactly sure how to hit each one of these conditions in order to test them. @rushatgabhane do you know?

puneetlath:
@rushatgabhane FYI you won't be able to test this yet as it requires some back-end changes that you don't yet have access to.

src/libs/actions/Report.js
```diff
@@ -1237,19 +1234,22 @@ function subscribeToNewActionEvent(reportID, callback) {
...
     const isFromCurrentUser = action.actorAccountID === currentUserAccountID;
     const updatedReportObject = {};
 
-    // When handling an action from the current user we can assume that their last read actionID has been updated in the server,
-    // but not necessarily reflected locally so we will update the lastReadSequenceNumber to mark the report as read.
-    updatedReportObject.maxSequenceNumber = action.sequenceNumber;
+    // When handling an action from the current user we can assume that their lastReadMessage has been updated in the server,
+    // but not necessarily reflected locally so we will update the lastMessageTimestamp to mark the report as read.
+    if (report.lastMessageTimestamp < action.reportActionTimestamp) {
+        updatedReportObject.lastMessageTimestamp = action.reportActionTimestamp;
```
puneetlath, line 1244: The comment makes me think that this should actually be looking at the `lastReadTimestamp` instead of the `lastMessageTimestamp`.

src/pages/home/report/ReportActionsView.js
puneetlath: Or @roryabraham

src/libs/ReportActionsUtils.js
```diff
@@ -154,37 +154,22 @@ function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
...
-    if (deletedSequenceNumber > lastReadSequenceNumber) {
-        return lastReadSequenceNumber;
-    }
-
-    // Otherwise, we must find the first previous index of an action that is not deleted and less than the lastReadSequenceNumber
-    const actions = _.toArray(lodashMerge({}, allReportActions[reportID], actionsToMerge));
-    const sortedActions = getSortedReportActions(actions);
-    const lastMessageIndex = _.findLastIndex(sortedActions, action => (
-        !isDeletedAction(action) && action.sequenceNumber <= lastReadSequenceNumber
+function getLastMessageTimestampAfterDeletingAction(reportID, deletedReportActionID) {
```
roryabraham, line 162: Personally I don't think we need the `AfterDeletingAction` suffix here. I think it's pretty intuitive that a function called `getLastMessageTimestampAfterDeletingAction` would exclude deleted actions.

src/libs/ReportActionsUtils.js
```diff
@@ -154,37 +154,22 @@ function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
...
+    const actionsWithoutDeleted = _.filter(allReportActions[reportID], action => (
```
roryabraham, line 163: I would recommend implementing this slightly differently:

```js
const actionsWithoutDeleted = _.filter(allReportActions[reportID], action => !isDeletedAction(action));
```

roryabraham: thus removing the need for `deletedReportActionID`

roryabraham (changes requested):

src/libs/ReportActionsUtils.js
```diff
@@ -154,37 +154,22 @@ function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
 }
 
 /**
+ * Get the last comment on a report by its reportActionTimestamp
```
roryabraham, line 157: NAB
```suggestion
 * Get the last comment on a report by its reportActionTimestamp
 *
```

src/pages/home/report/ReportActionsView.js
```diff
@@ -65,23 +64,22 @@ class ReportActionsView extends React.Component {
...
         this.currentScrollOffset = 0;
-        this.sortedReportActions = this.getSortedReportActionsForDisplay(props.reportActions);
         this.mostRecentIOUReportActionID = ReportActionsUtils.getMostRecentIOUReportActionID(props.reportActions);
         this.trackScroll = this.trackScroll.bind(this);
         this.toggleFloatingMessageCounter = this.toggleFloatingMessageCounter.bind(this);
         this.loadMoreChats = this.loadMoreChats.bind(this);
         this.recordTimeToMeasureItemLayout = this.recordTimeToMeasureItemLayout.bind(this);
         this.scrollToBottomAndMarkReportAsRead = this.scrollToBottomAndMarkReportAsRead.bind(this);
         this.openReportIfNecessary = this.openReportIfNecessary.bind(this);
+        this.computeNewMarkerReportActionID = this.computeNewMarkerReportActionID.bind(this);
```
roryabraham, line 82: This does not need to be bound to `this` because it's not passed to a child component and it's always executed in the context in which it's declared (the class in this case)

roryabraham (dismissed):

src/pages/home/report/ReportActionsView.js
```diff
@@ -277,6 +271,22 @@ class ReportActionsView extends React.Component {
...
 
+    /**
+     * @returns {String|null}
+    */
+    computeNewMarkerReportActionID() {
+        let newMarkerReportActionID = null;
+        if (ReportUtils.isUnread(this.props.report)) {
+            const newMarkerIndex = _.findLastIndex(this.sortedReportActions, reportAction => (
+                (reportAction.reportActionTimestamp || 0) > this.props.report.lastReadTimestamp
+            ));
```
marcaaron, line 282: I am a little confused about this logic and whether it works. `findLastIndex()` should iterate in reverse until we find an action that has a `timestamp` greater than the `lastReadTimestamp`. If there were 3 unread messages wouldn't it hit the last one first? But we want the 1 unread? Or does it work because `this.sortedReportActions` is backwards?

src/pages/home/report/ReportActionsView.js
```diff
@@ -277,6 +271,22 @@ class ReportActionsView extends React.Component {
...
+            newMarkerReportActionID = _.has(this.sortedReportActions[newMarkerIndex], 'reportActionID')
+                ? this.sortedReportActions[newMarkerIndex].reportActionID
+                : null;
```
marcaaron, line 285: What are we guarding against in this case? Why would any action not have a `reportActionID`?

marcaaron (commented):
Lots of questions  - a lot of it is just trying to get up to speed since I don't recall the exact details from the doc.

src/pages/home/report/ReportActionsView.js
puneetlath: Right you are! Thanks.

src/libs/actions/Report.js
```diff
@@ -689,29 +682,25 @@ function readNewestAction(reportID, createdDate) {
...
  *
  * @param {String} reportID
- * @param {String} createdDate
- * @param {Number} sequenceNumber
+ * @param {String} created
  */
-function markCommentAsUnread(reportID, createdDate, sequenceNumber) {
-    const newLastReadSequenceNumber = sequenceNumber - 1;
+function markCommentAsUnread(reportID, created) {
+    // We subtract 1 millisecond so that the lastRead is updated to just before this reportAction's created date
```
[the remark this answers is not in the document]
puneetlath: How about something like: 

```
// We subtract 1 millisecond so that the lastRead is updated to just before this reportAction's created date
// For example, if we marked a report action with timestamp 1000 unread, we set the lastRead to 999
// This makes it so that the report action with timestamp 1000 is the first one shown as unread
```

src/pages/home/report/ReportActionsView.js
```diff
@@ -102,28 +100,28 @@ class ReportActionsView extends React.Component {
...
                     Report.readNewestAction(this.props.report.reportID, _.last(_.toArray(this.props.reportActions)).created);
-                    this.setState({newMarkerSequenceNumber: 0});
-                } else if (this.state.newMarkerSequenceNumber === 0) {
-                    this.setState({newMarkerSequenceNumber: currentLastSequenceNumber});
+                    this.setState({newMarkerReportActionID: null});
+                } else if (!this.state.newMarkerReportActionID) {
+                    this.setState({newMarkerReportActionID: currentLastReportActionID});
                 }
-            } else if (this.state.newMarkerSequenceNumber === 0) {
+            } else if (!this.state.newMarkerReportActionID) {
```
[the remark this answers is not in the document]
puneetlath: I like that.

src/pages/home/report/ReportActionsView.js
```diff
@@ -143,19 +141,15 @@ class ReportActionsView extends React.Component {
             return true;
         }
 
-        if (nextProps.report.isLoadingReportActions !== this.props.report.isLoadingReportActions) {
-            return true;
-        }
```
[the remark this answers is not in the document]
puneetlath: Weird, I have no memory of this. Putting it back!

src/pages/home/report/ReportActionsView.js
puneetlath: Exactly. `this.sortedReportActions` is sorted in reverse order.

src/pages/home/report/ReportActionsView.js
```diff
@@ -277,6 +271,22 @@ class ReportActionsView extends React.Component {
...
+        if (ReportUtils.isUnread(this.props.report)) {
+            const newMarkerIndex = _.findLastIndex(this.sortedReportActions, reportAction => (
+                (reportAction.reportActionTimestamp || 0) > this.props.report.lastReadTimestamp
+            ));
+            newMarkerReportActionID = _.has(this.sortedReportActions[newMarkerIndex], 'reportActionID')
+                ? this.sortedReportActions[newMarkerIndex].reportActionID
+                : null;
+        }
+        return newMarkerReportActionID;
+    }
```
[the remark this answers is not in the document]
puneetlath: Nice improvement. I like it.

src/libs/actions/Report.js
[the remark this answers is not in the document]
marcaaron: Yeah I think that helps visualize it really well.

src/libs/ReportUtils.js
```diff
@@ -1013,9 +1013,9 @@ function buildOptimisticWorkspaceChats(policyID, policyName) {
  * @returns {Boolean}
  */
 function isUnread(report) {
-    const lastReadSequenceNumber = report.lastReadSequenceNumber || 0;
-    const maxSequenceNumber = report.maxSequenceNumber || 0;
-    return lastReadSequenceNumber < maxSequenceNumber;
+    const lastMessageTimestamp = report.lastMessageTimestamp || 0;
+    const lastReadTimestamp = report.lastReadTimestamp || 0;
```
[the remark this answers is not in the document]
puneetlath: In theory, `report.lastMessageTimestamp` should always be the last visible message.

src/libs/ReportActionsUtils.js
puneetlath: I like `getLastVisibleMessageTimestamp`

src/libs/ReportActionsUtils.js
puneetlath: I think I find filter + max more intuitive. I'll updated it to use the `isDeletedAction` function though.

src/libs/actions/Report.js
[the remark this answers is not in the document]
puneetlath: Cool.

src/libs/actions/Report.js
```diff
@@ -689,29 +682,25 @@ function readNewestAction(reportID, createdDate) {
...
+    const lastReadTimestamp = Date.parse(`${created} UTC`) - 1;
```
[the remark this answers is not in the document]
puneetlath: Oh wow, good digging. I didn't consider multi-browser support. 

Hm, maybe it's better to use `moment` in that case. There's a combination of both being used in various places. I think I'll just standardize on `moment`.

src/libs/actions/Report.js
```diff
@@ -1235,23 +1219,25 @@ function subscribeToNewActionEvent(reportID, callback) {
...
  */
 function viewNewReportAction(reportID, action) {
+    const report = allReports[reportID];
     const isFromCurrentUser = action.actorAccountID === currentUserAccountID;
     const updatedReportObject = {};
 
-    // When handling an action from the current user we can assume that their last read actionID has been updated in the server,
-    // but not necessarily reflected locally so we will update the lastReadSequenceNumber to mark the report as read.
-    updatedReportObject.maxSequenceNumber = action.sequenceNumber;
+    updatedReportObject.lastMessageTimestamp = report.lastMessageTimestamp < action.reportActionTimestamp ? action.reportActionTimestamp : report.lastMessageTimestamp;
```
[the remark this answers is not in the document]
puneetlath: Yes, that's a good point actually. The ideal would be not to have to have special logic here. And for the server to just tell the client the updated `lastMessageTimestamp` whenever it is telling the client about new messages.

src/libs/ReportActionsUtils.js
roryabraham: Me too, good suggestion 👍🏼

src/libs/actions/Report.js
```diff
@@ -417,6 +409,7 @@ function addActions(reportID, text = '', file) {
         lastMessageText: ReportUtils.formatReportLastMessageText(lastAction.message[0].text),
         lastActorEmail: currentUserEmail,
         lastReadSequenceNumber: newSequenceNumber,
+        lastReadTimestamp: Date.now(),
```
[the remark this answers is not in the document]
roryabraham: Good eye. I think ms is what we expect here

src/libs/actions/Report.js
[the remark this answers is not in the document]
roryabraham: > Just wanted to double check that the created argument we are passing here is using that format?

For clarity – it's not. `created` is using a similar but different format: `YYYY-MM-DD HH:mm:ss.sss`, modeled after the timestamps we typically use in Auth

src/pages/home/report/ReportActionsView.js
```diff
@@ -200,9 +194,9 @@ class ReportActionsView extends React.Component {
         const didReportBecomeVisible = isReportFullyVisible && (didSidebarClose || didScreenSizeIncrease);
         if (didReportBecomeVisible) {
             this.setState({
-                newMarkerSequenceNumber: !ReportUtils.isUnread(this.props.report)
-                    ? 0
-                    : this.props.report.lastReadSequenceNumber + 1,
+                newMarkerReportActionID: !ReportUtils.isUnread(this.props.report)
+                    ? null
+                    : this.computeNewMarkerReportActionID(),
```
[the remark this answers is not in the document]
roryabraham: Another alternative would be to just sort the reportActions again in `computeNewMarkerReportActionID`, but it would come at a performance cost because we would be re-sorting an already-sorted array.

I think I'd prefer to just do your suggestion above and make it more clear that the function currently called `computeNewMarkerReportActionID` needs a sorted array of reportActions

src/libs/actions/Report.js
```diff
@@ -689,29 +682,27 @@ function readNewestAction(reportID, createdDate) {
...
+            createdDate: DateUtils.getDBTime(lastReadTimestamp),
         },
         {
             optimisticData: [{
                 onyxMethod: CONST.ONYX.METHOD.MERGE,
                 key: `${ONYXKEYS.COLLECTION.REPORT}${reportID}`,
                 value: {
-                    lastReadSequenceNumber: newLastReadSequenceNumber,
-                    lastReadTimestamp: Date.now(),
+                    lastReadTimestamp,
```
mountiny, line 705: Not sure if if I am missing something here but the comment talks about subtracting 1 millisecond, but we dont do it here.

src/pages/home/report/ReportActionsView.js
```diff
@@ -65,23 +64,22 @@ class ReportActionsView extends React.Component {
...
 
         this.unsubscribeVisibilityListener = null;
+        this.sortedReportActions = this.getSortedReportActionsForDisplay(props.reportActions);
 
         this.state = {
             isFloatingMessageCounterVisible: false,
-            newMarkerSequenceNumber: ReportUtils.isUnread(props.report)
-                ? props.report.lastReadSequenceNumber + 1
-                : 0,
+            newMarkerReportActionID: this.computeNewMarkerReportActionID(),
```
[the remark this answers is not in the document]
mountiny: Great points

mountiny (commented):
Great job on this one Puneet and already some very thorough reviews in!

src/pages/home/report/ReportActionsView.js
```diff
@@ -33,7 +33,7 @@ const propTypes = {
     report: reportPropTypes.isRequired,
```
MonilBhavsar, line 33: Just noticed, this file has permissions being changed

src/libs/actions/Report.js
puneetlath: Whoops, that seems to have gotten lost in the last commit when I switched to using `moment`. Thanks for catching!

src/pages/home/report/ReportActionsView.js
MonilBhavsar: This PR also proposes to change file permissions. I think you'll need to reset it.

Screenshot 2022-12-19 at 11 00 15 PM

mountiny: Wow, never seen this, never would notice 😂

src/pages/home/report/ReportActionsView.js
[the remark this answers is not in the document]
puneetlath: Makes sense to me too. Updating!

src/libs/actions/Report.js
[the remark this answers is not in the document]
puneetlath: I think it depends on whether we can count on these to come in order. Asked here: https://expensify.slack.com/archives/C04DC6LU2UB/p1671477740898959

src/libs/actions/Report.js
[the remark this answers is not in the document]
puneetlath: Correct! Going to go ahead and resolve.

src/libs/actions/Report.js
```diff
@@ -1235,23 +1219,25 @@ function subscribeToNewActionEvent(reportID, callback) {
...
  * @param {String} reportID
  * @param {Object} action
  */
 function viewNewReportAction(reportID, action) {
+    const report = allReports[reportID];
     const isFromCurrentUser = action.actorAccountID === currentUserAccountID;
     const updatedReportObject = {};
 
-    // When handling an action from the current user we can assume that their last read actionID has been updated in the server,
-    // but not necessarily reflected locally so we will update the lastReadSequenceNumber to mark the report as read.
```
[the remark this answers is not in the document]
puneetlath: Ended up removing this logic altogether as discussed in Slack.

src/libs/actions/Report.js
```diff
@@ -1235,23 +1219,25 @@ function subscribeToNewActionEvent(reportID, callback) {
 }
 
 /**
+ * This is called whenever a new report action is inserted in Onyx.
+ * Typically, this will be called with the latest report action on the report. However, because we cannot
+ * guarantee the order that messages are sent from the server, it is not safe to assume that it is the latest report action.
```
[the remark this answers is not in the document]
puneetlath: Makes sense. Thanks for the feedback!

src/libs/actions/Report.js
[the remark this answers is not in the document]
puneetlath: Thanks for the discussion in those threads. Went ahead and removed that logic.

src/pages/home/report/ReportActionsView.js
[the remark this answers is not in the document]
puneetlath: Moved to `ReportUtils` as discussed.

src/pages/home/report/ReportActionsView.js
puneetlath: Wow, I've also never seen that! Thanks for noticing.

src/pages/home/report/ReportActionsView.js
```diff
@@ -33,7 +33,7 @@ const propTypes = {
...
 
     /** Array of report actions for this report */
-    reportActions: PropTypes.objectOf(PropTypes.shape(reportActionPropTypes)),
+    reportActions: PropTypes.objectOf(PropTypes.shape(reportActionPropTypes)).isRequired,
```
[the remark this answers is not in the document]
puneetlath: Actually you're right. The doc said to do this and I thought it was necessary for how we're computing the `reportActions` so that they wouldn't be empty but that actually gets handled just fine.

src/libs/actions/Report.js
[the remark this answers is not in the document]
marcaaron: Oh, ok this is actually not the logic I thought we would be removing. But I think makes sense (depending on what the backend does). 

Are we certain that each time a new action is created we will set the "last read" and "last action" for the report?

src/libs/actions/Report.js
```diff
@@ -1235,27 +1220,18 @@ function subscribeToNewActionEvent(reportID, callback) {
 }
 
 /**
+ * Handles callbacks and notifications for new report actions.
+ *
  * @param {String} reportID
  * @param {Object} action
  */
-function viewNewReportAction(reportID, action) {
+function handleNewReportAction(reportID, action) {
```
marcaaron, line 1228: https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md#naming-conventions

src/libs/actions/Report.js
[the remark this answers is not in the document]
puneetlath: Oh ok. Which logic did you think we would be removing? 

>Are we certain that each time a new action is created we will set the "last read" and "last action" for the report?

There are some scenarios where it doesn't. Have discussed with @roryabraham and going to open a Web PR after Tyler's to address those.

src/libs/ReportActionsUtils.js
```diff
@@ -154,37 +154,22 @@ function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
 }
 
 /**
+ * Get the reportActionTimestamp of the last visible (non-deleted) comment on a report
```
marcaaron, line 157: See: https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md#jsdocs

src/pages/home/report/ReportActionsView.js
puneetlath: If, for example, the sortedReportActions is an empty array for some reason (e.g. we are offline and don't have the report actions).

src/libs/ReportUtils.js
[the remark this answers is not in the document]
marcaaron: I'm not sure based on your response, but it sounds like maybe you're saying that if someone deletes a message that we will update a participant's report object with the `lastMessageTimestamp` of the last _visible_ message? Where does that happen exactly?

I appreciate that we are assuming this will work in theory, but we can add a test case for this in the QA steps and see whether we got it right.

src/libs/actions/Report.js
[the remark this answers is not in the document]
marcaaron: Ok so this will be on HOLD til then? If so I can stop reviewing for now.

src/libs/ReportActionsUtils.js
puneetlath: Thanks.

src/libs/actions/Report.js
puneetlath: Thanks. Updated to `sendNewReportActionNotification`.

roryabraham:
One regression I noticed testing this on iOS:

1. User A is scrolled up.
1. User B sends a message to user A
1. The chat becomes unread for user A
1. User A does not scroll to the bottom, then opens a different chat.
1. User A goes back to the chat, and sees the new message marker.
1. User A goes back to the LHN, and the chat is still unread. 💥 
1. User A opens a different chat _again_, then goes back to the other one. The new message marker is still there (even though we've seen it, gone to a different report, then come back) 💥 
1. User A goes back to the LHN. The chat now appears as read.
1. User A goes back into the same chat, and the unread marker is still there 💥 
1. User A opens a different chat then comes back. Now the unread marker is finally gone and the chat is fully unread.

Compared to on main:

1. User A is scrolled up
1. User B sends a message to user A
1. The chat becomes unread for user A
1. User A does not scroll to the bottom, then opens a different chat
1. User A goes back to the chat, and sees the new message marker.
1. User A goes back to the LHN, and the chat is read ✅ 
1. User A goes back to the chat, and even though the chat is read the unread marker is still there 💥 
1. User A opens a different chat _again_, then goes back to the other one. The new message marker is gone now 💥

roryabraham:
🐛 

10. (narrow screens) User A scrolls to the bottom of the chat and then goes back to the LHN/Home screen but does not open a different report.
11. (narrow screens) User B sends a message in the chat
12. (narrow screens) The report becomes unread for User A. ❌ 
    - Instead, the report remains read

roryabraham:
🐛 

21. User A scrolls back, then marks one of their own reportActions in the middle of the report as unread. The report should become unread and a new message indicator should appear above that reportAction.
22. User A deletes the reportAction they marked as unread. The report should remain unread and the unread action indicator should appear above the next-newest reportAction. ❌ 
    - Instead, the action fails to delete:

mountiny:
Just confirming, is this still WIP as the title say or should we review/test?

AndrewGable:
Can we please add testing attachments for unread message? Let's make sure this bug does not exist: https://github.com/Expensify/App/issues/13472 after this PR.

puneetlath:
@mountiny I need to update the `UnreadIndicators` test and then address the bugs that have been found. Feel free to review now if you like or if you want to wait until after I do that, that's cool too.

roryabraham:
_Note:_ I added test steps for: https://github.com/Expensify/App/issues/13472 and https://github.com/Expensify/App/issues/13684 so included them in `Fixed Issues` in the PR description, but it remains to be seen if this PR fixes them or not.

roryabraham:
> Can we please add testing attachments for unread message? Let's make sure this bug does not exist: https://github.com/Expensify/App/issues/13472 after this PR.

Done. Also added testing steps for https://github.com/Expensify/App/issues/13684

puneetlath:
Ok, for the issue with needing to navigate to a chat twice for it to get marked as read, I think that's because we don't update the `lastReadCreated` here: https://github.com/Expensify/Web-Expensify/blob/596df9162173157ede9e1e385b28f09240f5ad1c/lib/ReportAPI.php#L4205

So we optimistically set a `lastRead` of `Date.now()` when we open the report. But the `openReport` command from the API returns the old last read, so the report gets re-marked as unread.

puneetlath:
For the narrow screens issue, it seems that for some reason `isDrawerOpen` is not returning true even though it is true for the parent `ReportScreen` component (convo [here](https://expensify.slack.com/archives/C04DC6LU2UB/p1671657937698209)).

puneetlath:
For the delete message issue, I'm going to wait on Tyler's PR as that seems like it's most likely a back-end issue also.

neil-marcellini:
I found some issues with tests 10-12 here https://github.com/Expensify/Web-Expensify/pull/35907#pullrequestreview-1232736570.

neil-marcellini (commented):
I ran through the steps with this PR and Chirag's. I found that steps 25, 31, 36 didn't work. 38 worked except I didn't see a new message bubble but the green unread line was in the right place.

On main for Web-E and App I found that steps 25, 26, 30, 31, 36, 38 fail. Again at 30 and 38 I didn't see a new message bubble but the green unread line was in the right place.

src/libs/ReportUtils.js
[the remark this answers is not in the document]
puneetlath: Ok, so based on my testing the lastMessageTime is always the last visible message.

src/libs/actions/Report.js
```diff
@@ -830,16 +820,10 @@ function deleteReportComment(reportID, reportAction) {
...
-        optimisticReportActions,
-        reportAction.sequenceNumber,
-        getLastReadSequenceNumber(reportID),
-    );
+    const lastMessageTimestamp = ReportActionsUtils.getLastMessageTimestampAfterDeletingAction(reportID, reportActionID);
     const optimisticReport = {
         lastMessageText,
-        lastReadSequenceNumber,
-        maxSequenceNumber: lastReadSequenceNumber,
+        lastMessageTimestamp,
```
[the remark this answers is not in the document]
puneetlath: >What happens when we refresh the page? Does the API command that returns updated report data return a report.lastMessageTimestamp that takes deleted actions into consideration?

Yes! 

>What happens when someone else deletes a message? Are we handling that event and updating the lastMessageTimestamp in realtime?

Yes!

>Should we rename this to lastVisibleMessageTimestamp? I think it's slightly less ambiguous than lastMessageTimestamp (which makes me wonder if the message could be deleted).

We are planning to do this in a follow-up: https://github.com/Expensify/Expensify/issues/251662

puneetlath:
Thanks for the testing @neil-marcellini. I just ran through the full list of tests on all devices and I didn't find any issues that weren't already also on main. And all the ones I did find that aren't fixed in this PR have follow-up issues linked.

src/libs/ReportUtils.js
```diff
@@ -1018,9 +1018,9 @@ function isUnread(report) {
         return false;
     }
 
-    const lastReadSequenceNumber = report.lastReadSequenceNumber || 0;
-    const maxSequenceNumber = report.maxSequenceNumber || 0;
-    return lastReadSequenceNumber < maxSequenceNumber;
+    const lastActionCreated = report.lastActionCreated || '';
+    const lastReadTime = report.lastReadTime || '';
+    return lastReadTime < lastActionCreated;
```
marcaaron, line 1023: NAB, sorry, I just keep forgetting.. what format and/or type are `lastReadTime` and `lastActionCreated`? They can be compared with a less than symbol? This might be a good thing to add a very brief comment.

src/libs/ReportUtils.js
```diff
@@ -1185,6 +1185,26 @@ function chatIncludesChronos(report) {
...
+ * @returns {String|null}
+ */
+function getNewMarkerReportActionID(report, sortedAndFilteredReportActions) {
+    if (!isUnread(report)) {
+        return null;
+    }
+
+    const newMarkerIndex = _.findLastIndex(sortedAndFilteredReportActions, reportAction => (
+        (reportAction.created || '') > report.lastReadTime
+    ));
```
marcaaron, line 1201: I'm not sure about this code tbh. Doesn't the dependency on having a list of `sortedAndFilteredReportActions` mean that we can only figure out _where_ to show the new marker if we have the latest report actions stored locally?

IIRC `ReportActionsView` can load with stale or no actions, but it must be initialized with the correct new marker position. In the current design the only thing required to figure out the position is a `report`. And there's a bunch of logic to update it later for various reasons - I think we also lean on the assumption that we had the right value on the first `render()` of the `ReportActionsView`.

src/libs/actions/Report.js
```diff
@@ -471,52 +465,45 @@ function openPaymentDetailsPage(chatReportID, iouReportID) {
...
+    // Since the report action with ID 100 will be the first with a timestamp above '2014-04-01 16:07:02.998', it's the first one that will be shown as unread
+    const lastReadTime = DateUtils.oneMillisecondBefore(reportActionCreated);
     API.write('MarkAsUnread',
         {
             reportID,
-
-            // We subtract 1 millisecond so that the lastRead is updated to just before this reportAction's created date
-            createdDate: DateUtils.getDBTime(moment.utc(createdDate).valueOf() - 1),
-            sequenceNumber,
+            isoTime: lastReadTime,
```
marcaaron, line 499: Why is this sent as `isoTime` and not just `lastReadTime`? `isoTime` describes the type of timestamp but not what the variable represents. So I think it will be a little unclear from the API side?

src/libs/actions/Report.js
```diff
@@ -1073,27 +1054,18 @@ function subscribeToNewActionEvent(reportID, callback) {
 }
 
 /**
+ * Sends callbacks and notifications for new report actions.
```
marcaaron, line 1057: nab, I tend to think of callbacks as functions that are called and not things that are "sent" somewhere

so I'd change this to:

>Calls callbacks and presents notifications for new report actions arriving in realtime

src/libs/actions/Report.js
```diff
@@ -1073,27 +1054,18 @@ function subscribeToNewActionEvent(reportID, callback) {
...
+ *
  * @param {String} reportID
  * @param {Object} action
  */
-function viewNewReportAction(reportID, action) {
+function sendNewReportActionNotification(reportID, action) {
```
marcaaron, line 1062: I would change this to `presentReportActionNotification()` if we are struggling to get the function name to make sense because of the callback then it might be a good idea to just remove the callback logic out of this and have the subscriber callback get called here instead:

```js
            presentReportActionNotification(reportID, action);

            // Notify the ReportActionsView that a new comment has arrived
            if (reportID === newActionSubscriber.reportID) {
                newActionSubscriber.callback(isFromCurrentUser, action.reportActionID);
            }
            handledReportActions[reportID] = handledReportActions[reportID] || {};
            handledReportActions[reportID][action.sequenceNumber] = true;
        });
```

src/pages/home/report/ReportActionsList.js
```diff
@@ -126,10 +127,9 @@ class ReportActionsList extends React.Component {
         item: reportAction,
         index,
     }) {
-        // When the new indicator should not be displayed we explicitly set it to 0. The marker should never be shown above the
-        // created action (which will have sequenceNumber of 0) so we use 0 to indicate "hidden".
```
marcaaron, line 130: NAB, but did we check to make sure that the new marker can't be displayed above the created action? The previous comment here suggests something about that - but I'm not too sure what we were worried about tbh.

src/pages/home/report/ReportActionsView.js
```diff
@@ -63,18 +63,17 @@ class ReportActionsView extends React.Component {
 
         this.didLayout = false;
         this.didSubscribeToReportTypingEvents = false;
-
         this.unsubscribeVisibilityListener = null;
 
+        // We need this.sortedAndFilteredReportActions to be set before this.state is initialized because the function to calculate the newMarkerReportActionID uses the sorted report actions
+        this.sortedAndFilteredReportActions = this.getSortedReportActionsForDisplay(props.reportActions);
```
marcaaron, line 69: Commented above, but this component is also responsible for loading the report actions - so how do we know that `props.reportActions` are up to date at this point?

marcaaron (changes requested):
Left a few more comments. I haven't tested this yet. I'm a little concerned still about the `newMarkerReportActionID` changes. That said, we also know there are a few things broken about this feature at the moment.

src/libs/ReportUtils.js
marcaaron: This might be a NAB if everyone else agrees that the whole design is kinda fundamentally borked and we'd rather do some larger refactor later. FWIW I feel like it will be very hard to solve this problem in this PR.

src/libs/ReportUtils.js
```diff
@@ -109,7 +109,7 @@ function sortReportsByLastRead(reports) {
     return _.chain(reports)
         .toArray()
         .filter(report => report && report.reportID && !isIOUReport(report))
-        .sortBy('lastReadTimestamp')
+        .sortBy('lastReadTime')
```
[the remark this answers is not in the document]
puneetlath: Yes, this goes with [this web PR](https://github.com/Expensify/Web-Expensify/pull/35907) where we updated the field name from `lastReadTimestamp` to `lastReadTime`.

src/libs/ReportActionsUtils.js
```diff
@@ -161,35 +160,18 @@ function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
...
+ * @returns {String}
  */
-function getOptimisticLastReadSequenceNumberForDeletedAction(reportID, actionsToMerge = {}, deletedSequenceNumber, lastReadSequenceNumber) {
-    // If the action we are deleting is unread then just return the current last read sequence number
-    if (deletedSequenceNumber > lastReadSequenceNumber) {
-        return lastReadSequenceNumber;
-    }
-
-    // Otherwise, we must find the first previous index of an action that is not deleted and less than the lastReadSequenceNumber
+function getLastVisibleActionCreated(reportID, actionsToMerge = {}) {
```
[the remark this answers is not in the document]
puneetlath: There is a field on every report action called `created` which stores the datetime that it was created at. This will return the `created` field for the last visible action. 

image

Is there a different name you think would be clearer? Maybe `getCreatedForLastVisibleAction`?

src/libs/actions/Report.js
puneetlath: That is what the API call expects ([here](https://github.com/Expensify/Web-Expensify/pull/35907/files#diff-8367f109728fcc54db479ccc765a52d2115e52d35eba9c660e665be1a34411acR778)). However @chiragsalian I like Marc's suggestion of calling the API param `lastReadTime` better. What do you think.

puneetlath: That is what the API call expects ([here](https://github.com/Expensify/Web-Expensify/pull/35907/files#diff-8367f109728fcc54db479ccc765a52d2115e52d35eba9c660e665be1a34411acR778)). However @chiragsalian I like Marc's suggestion of calling the API param `lastReadTime` better. What do you think.

src/libs/actions/Report.js
puneetlath: Ah I like that. I'll do it that way.

src/libs/DateUtils.js
```diff
@@ -175,6 +175,17 @@ function getDBTime(timestamp = '') {
         .replace('Z', '');
 }
 
+/**
+ * Returns the datetime string one millisecond before a given datetime string
+ *
+ * @param {String} dateTimeString in the format expected by the database
```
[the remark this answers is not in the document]
puneetlath: The challenging part is that we don't use a standard format which has a name (I don't think). Our format is this `2022-12-30 20:49:11.234`. ISO format would be `2022-12-30T20:49:11.234Z`. I think that's why we've been describing it as DB time. Because that's the format our DB uses.

src/libs/DateUtils.js
```diff
@@ -175,6 +175,17 @@ function getDBTime(timestamp = '') {
...
+ * @returns {String} in the format expected by the database
+ */
+function oneMillisecondBefore(dateTimeString) {
```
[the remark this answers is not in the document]
puneetlath: Cool, makes sense. Good feedback. Updated!

src/libs/ReportActionsUtils.js
```diff
@@ -161,35 +160,18 @@ function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
...
-        !isDeletedAction(action) && action.sequenceNumber <= lastReadSequenceNumber
-    ));
-
-    // It's possible we won't find any and in that case the last read should be reset
-    if (lastMessageIndex < 0) {
-        return 0;
-    }
-
-    return sortedActions[lastMessageIndex].sequenceNumber;
+    const actionsWithoutDeleted = _.filter(actions, action => (!isDeletedAction(action)));
```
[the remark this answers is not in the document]
puneetlath: I like that better too. Updated.

src/libs/DateUtils.js
```diff
@@ -175,6 +175,17 @@ function getDBTime(timestamp = '') {
         .replace('Z', '');
 }
 
+/**
+ * Returns the datetime string one millisecond before a given datetime string
+ *
+ * @param {String} dateTimeString in the format expected by the database
+ * @returns {String} in the format expected by the database
```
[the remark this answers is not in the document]
puneetlath: Same response as above.

src/libs/ReportUtils.js
puneetlath: They are datetime strings, so they can be compared with comparison operators. I added a comment.

src/pages/home/report/ReportActionsList.js
puneetlath: Yes, it will never show above the created action because if an action is a created action we render it via `ReportActionItemCreated` which doesn't even get passed the `shouldDisplayNewIndicator` prop whereas regular report actions are rendered via ReportActionItem. 

So actually looking at the code, we don't actually even need the null check. And because this component only ever gets passed visible actions, I don't think we need the `isDeletedAction` check either. I think we could actually just simplify this to: 

```const shouldDisplayNewIndicator = reportAction.reportActionID === this.props.newMarkerReportActionID;```

File: src/libs/ReportUtils.js (hunk 28 above)
Comment (puneetlath) [reply src/libs/ReportUtils.js:1200]: Yeah I agree it could be improved but what we have here brings us up to par with what is currently on main. I think further improvements can be implemented via this convo: https://expensify.slack.com/archives/C01GTK53T8Q/p1671704053216749

File: src/pages/home/report/ReportActionsView.js (hunk 33 above)
Comment (puneetlath) [reply src/pages/home/report/ReportActionsView.js:69]: I suppose we don't. It will load the marker based on the data it has and then the callback gets called when new report actions come into onyx. But I think further improvements can be handled via this convo: https://expensify.slack.com/archives/C01GTK53T8Q/p1671704053216749

Comment (marcaaron) [reply src/pages/home/report/ReportActionsView.js:69]: Yeah, I was doing some testing on this last week and think the design is broken in the same way on staging/production today anyway.

File: src/pages/home/report/ReportActionsList.js (hunk 32 above)
Comment (marcaaron) [reply src/pages/home/report/ReportActionsList.js:130]: Oh nice, that would be pretty sweet. Could we also set `newMarkerReportActionID` to an empty string instead of `null`?

File: src/libs/ReportActionsUtils.js (hunk 35 above)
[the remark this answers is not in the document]
Comment (marcaaron) [reply src/libs/ReportActionsUtils.js:165]: idk maybe I am overthinking it - I can't think of any improvements on this other than `getLastVisibleActionCreatedTime()` - which doesn't feel like much of an improvement. Naming things is hard 😅

File: src/libs/DateUtils.js (hunk 36 above)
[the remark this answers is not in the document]
Comment (marcaaron) [reply src/libs/DateUtils.js:181]: Oh hmm.. maybe we can add to the comment what the format is as a reminder? I think my concern is that people will see this and ask, "what format is expected by the database?"

File: src/pages/home/report/ReportActionsList.js (hunk 32 above)
Comment (puneetlath) [reply src/pages/home/report/ReportActionsList.js:130]: Yeah I don't see why not. Why is that better than null?

File: src/libs/actions/Report.js (hunk 29 above)
Comment (chiragsalian) [reply src/libs/actions/Report.js:499]: I don't mind either. I do like marc's suggestion. The variable is isoTime in PHP layer and in auth, updating it would be a bit tricky. I'll update web-e for now.

File: src/libs/DateUtils.js (hunk 36 above)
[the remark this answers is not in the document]
Comment (puneetlath) [reply src/libs/DateUtils.js]: Actually, I think we can just leave that detail out since it can in theory accept any format that `moment` supports. Updated.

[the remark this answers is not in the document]
Comment (puneetlath) [reply src/libs/DateUtils.js]: Actually, I think we can just leave that detail out since it can in theory accept any format that `moment` supports. Updated.

File: src/libs/actions/Report.js (hunk 29 above)
Comment (chiragsalian) [reply src/libs/actions/Report.js:499]: PR here - https://github.com/Expensify/Web-Expensify/pull/35959

File: src/libs/ReportActionsUtils.js (hunk 35 above)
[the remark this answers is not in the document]
Comment (roryabraham) [reply src/libs/ReportActionsUtils.js:165]: Seems like we could DRY up some logic between this function and `getLastVisibleMessageText` by adding a function called `getLastVisibleAction`:

```js
/**
 * @param {String} reportID
 * @param {Object} [actionsToMerge]
 * @returns {Object}
 */
function getLastVisibleAction(reportID, actionsToMerge = {}) {
    const actions = _.toArray(lodashMerge({}, allReportActions[reportID], actionsToMerge));
    const visibleActions = _.filter(actions, action => (!isDeletedAction(action)));
    return _.max(visibleActions, action => moment.utc(action.created).valueOf());
}

function getLastVisibleMessageText(reportID, actionsToMerge = {}) {
    const lastVisibleAction = getLastVisibleAction(reportID, actionsToMerge);
    const htmlText = lodashGet(lastVisibleAction, 'message[0].html', '');
    if (!htmlText) {
        return '';
    }

    const parser = new ExpensiMark();
    const messageText = parser.htmlToText(htmlText);
    return ReportUtils.formatReportLastMessageText(messageText);
}

function getLastVisibleActionCreated(reportID, actionsToMerge = {}) {
    const lastVisibleAction = getLastVisibleAction(reportID, actionsToMerge);
    return _.has(lastVisibleAction, 'created') ? lastVisibleAction.created : DateUtils.currentDBTime();
}
```

src/pages/home/report/ReportActionsList.js
puneetlath: Ok, updated!

src/libs/ReportActionsUtils.js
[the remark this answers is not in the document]
puneetlath: That's a good idea. I'll do that.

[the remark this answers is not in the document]
puneetlath: Ok. I decided to do it like this: 

1. Added `getLastVisibleAction`
2. Removed  `getLastVisibleActionCreated`
3. Have `getLastVisibleMessageText` call `getLastVisibleAction` 
4. Have `reportActionsView` call `getLastVisibleAction` and use the `created` field from the returned object. 

That feels the cleanest to me and seems to solve feedback from both of you. Let me know what you think!

src/libs/ReportUtils.js
```diff
@@ -1185,6 +1184,26 @@ function chatIncludesChronos(report) {
                 && _.contains(report.participants, CONST.EMAIL.CHRONOS);
 }
 
+/**
+ * @param {Object} report
```
chiragsalian, line 1188: i think you should have an annotation for report.lastReadTime since you are using it below.

src/libs/actions/Report.js
```diff
@@ -1076,38 +1057,21 @@ function subscribeToNewActionEvent(reportID, callback) {
...
-    }
-
-    if (reportID === newActionSubscriber.reportID) {
-        newActionSubscriber.callback(isFromCurrentUser, updatedReportObject.maxSequenceNumber);
-    }
-
-    Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${reportID}`, updatedReportObject);
-
-    const notificationPreference = lodashGet(allReports, [reportID, 'notificationPreference'], CONST.REPORT.NOTIFICATION_PREFERENCE.ALWAYS);
+function presentReportActionNotification(reportID, action) {
```
chiragsalian, line 1060: NAB: Rename method to `showActionNotification` or `notifyAction`? We are already in `Report.js` so having Report in this method name is redundant IMO.

tests/actions/ReportTest.js
```diff
@@ -207,6 +208,7 @@ describe('actions/Report', () => {
         const REPORT_ID = 1;
         let report;
         let reportActionCreatedDate;
+        let currentTime = DateUtils.getDBTime();
```
chiragsalian, line 211: i don't think there is a reason to set this value here right? So we can just have it as,
```
let currentTime;
```

It looks like its immediately overridden below.

chiragsalian (dismissed):
code LGTM, i left some minor NABs

tests/actions/ReportTest.js
puneetlath: Ah, looks like you're right. Updating.

src/pages/home/report/ReportActionsView.js
puneetlath: Ok cool. Gonna resolve then.

src/libs/ReportUtils.js
puneetlath: Going to resolve based on the discussion in the other comment.

puneetlath: Going to resolve based on the discussion in the other comment.

src/libs/actions/Report.js
puneetlath: I like "show" instead of "present" but I think keeping it as `reportAction` is clearer, since that's what the onyx collection/object is called.

src/libs/ReportUtils.js
puneetlath: Ah makes sense. Added.

puneetlath: Ah makes sense. Added.

puneetlath: Ah makes sense. Added.

marcaaron: plane WiFi 😂

marcaaron:
Testing in progress...

Quick note before I forget - I think if we have a guaranteed test failure on Step 12-13 then we should remove it from the test steps as QA is going to be confused by this, but maybe it's good to add as a "note" somewhere at the bottom as a "known issue".

Step 14 is also a little ambiguous because it says: 

>(narrow screens) User A opens a different report then goes back to the LHN. The report with user B should now be read.

In the previous step they had navigated to a report. So, are they going back to the LHN to open a different report? Or opening it via a deep link?

>Potential improvement: User A should not need to open a different report for the report to become read. Just going back to the LHN and then back to the report should cause the report to become read.

This is the current expected behavior as I understand it and more of a regression vs. an improvement. 

I tested this out though and what I saw is that the report is marked as read, but the new marker isn’t clearing. Then tested on staging and saw the same. So it's a regression, but not from this PR.

marcaaron:
>Potential improvement: Auto-scroll back to where the unread indicator is.

Let's get these ideas moved to Slack or issues. They are not really doing much for us in the QA test steps as we get closer to merging this.

marcaaron:
I think we should re-write these test steps so that there is less ambiguity about what we're looking for. I generally refer to any indication that a report is unread as an "unread indicator" but I think maybe you are using it to refer to the "new marker line" e.g. in step 22

>User A leaves the report, opens a different one, then comes back. The unread marker should be where it was in the previous step, and the report should become read.

If I understand this step correctly it is not the expected behavior. If you navigate away from a report that had the "new marker line" set then it should be cleared when you navigate away from the report. It shouldn't matter if you open a different report or not.

marcaaron:
Step 25 - Again, we should move failing tests out of the tests or test the expected behavior (even if we expect it to fail). I'd probably say we should avoid testing behaviors we know will fail because sometimes QA uses these tests in the future to assume a correct behavior.

marcaaron:
I think step 26 is also hard to test because it's dependent on 25. 

>User A opens a different report, then comes back. The new message indicator should be where is was before, and the chat should become read.

What state should we be in for this to work? I think it's right after we mark a message as unread? But in the tests it is after we delete a report action.

marcaaron:
Step 31 is a little ambiguous too

>User A closes the app completely and reopens it. The message from the previous step should still be unread.

Close the app how? Can we be more specific and if there are platform specific things we are doing?

marcaaron (approved):
Code LGTM! Ran into some trouble following the test steps. I get the sense that whatever is broken about the unread indicators at this point is consistent with staging or some possible recent regressions. I haven't looked into the source of those regressions, but certainly doesn't seem like this PR is going to make things any worse.

puneetlath:
Thanks @marcaaron! The tests have been updated. The steps that are in the PR now should bring us up to parity with main and should all be passing.

marcaaron:
## Reviewer Checklist

    - [ ] Android / native
    - [ ] Android / Chrome
    - [ ] iOS / native
    - [ ] iOS / Safari
    - [ ] MacOS / Chrome / Safari
    - [ ] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>



</details>

<details>



</details>

<details>



</details>

<details>
<summary>Desktop</summary>



</details>

<details>
<summary>iOS</summary>



</details>

<details>
<summary>Android</summary>



</details>

marcaaron:
>User A opens the report, navigates to a different one, then comes back. It should now be read.

should say

User A opens the report, navigates to LHN?, selects a different report, navigates back to the LHN. Verify the chat in the LHN is read

?

marcaaron:
Found this while testing on iOS (maybe it is a known issue)

1. Mark chat message as unread
2. Navigate to LHN
3. Close app
4. Reopen app
5. Unread status of report flashes from "read" to "unread" then eventually "read" 🐛  
6. Navigate to chat
7. No "new marker" present 🐛 

https://user-images.githubusercontent.com/32969087/210287046-2f560d71-154d-4ea3-bf3c-ffcf6b3901a5.mp4

chiragsalian (approved):

marcaaron:
Sorry this one is pretty wonky, but on mobile web (Safari) I performed these steps:

1. Visit a report where there was just one new comment
2. New marker shows in the correct spot ✅ 
3. Scroll up in the chat
4. Pull down to refresh
5. LHN will show (report is now "read") ✅ 
6. Tap back into chat and scroll up
7. Chat messages disappear briefly 🤔 
8. When they return the new marker is in the wrong spot 💥 

https://user-images.githubusercontent.com/32969087/210287611-488812f9-6893-4aab-bba6-9e574d7047a1.mp4

Didn't see this on any other platforms so far...

marcaaron:
Minor issue while testing mobile Safari

1. Mark last comment as read
2. Close Safari
3. Reopen Safari
4. New marker is not shown but appears a short time later 🐛 

https://user-images.githubusercontent.com/32969087/210287840-53a31fcb-23df-4e16-8cf6-d286e26390d3.mp4

marcaaron:
I don't think it was introduced here but for transparency there is a propType warning popping up when you long press on a comment (found while testing Android)

marcaaron:
Great work @puneetlath! 

Completed testing on various platforms. Turned up a few issues (minor I think). I'm also not certain if those are on staging or not (sorry testing took several hours). My personal opinion on this is that the things turned up are not urgent. There are a few open issues related to the unread indicators floating around and I don't think having this PR open any longer will help us solve those issues. So seeing as nothing major is broken I am 👍

That said, it does seem like something has recently regressed related to setting the "new marker" and unread status in the LHN. This stuff was working (definitely not perfectly, but I think better than it is working now) and we should commit to exploring improvements in the ReportScreen 2.0 conversation started by @roryabraham [here](https://github.com/Expensify/App/issues/13793).

puneetlath:
Thanks for the super thorough testing! 

> Found this while testing on iOS (maybe it is a known issue)
> 
> Mark chat message as unread
> Navigate to LHN
> Close app
> Reopen app
> Unread status of report flashes from "read" to "unread" then eventually "read" 🐛
> Navigate to chat
> No "new marker" present 🐛

This one is known and exists on main too. This is the same issue as this: https://github.com/Expensify/App/issues/13799

> Sorry this one is pretty wonky, but on mobile web (Safari) I performed these steps:
> 
> Visit a report where there was just one new comment
> New marker shows in the correct spot ✅
> Scroll up in the chat
> Pull down to refresh
> LHN will show (report is now "read") ✅
> Tap back into chat and scroll up
> Chat messages disappear briefly 🤔
> When they return the new marker is in the wrong spot 💥

I wasn't able to reproduce this one on main. I'll open an issue for it. It feels sufficiently niche to not block over. 

> Minor issue while testing mobile Safari (Edit: also happens on Android)
> 
> Mark last comment as read
> Close Safari
> Reopen Safari
> New marker is not shown but appears a short time later 🐛

This one also exists on main. I'll open an issue. 

>I don't think it was introduced here but for transparency there is a propType warning popping up when you long press on a comment (found while testing Android)

Confirmed this is also happening on main. I'll open an issue.

chiragsalian:
Discussed on slack, i'm proceeding to merge [this](https://expensify.slack.com/archives/C04DC6LU2UB/p1672788288616379?thread_ts=1672432986.026959&cid=C04DC6LU2UB).

src/libs/ReportUtils.js
```diff
@@ -1018,9 +1016,10 @@ function isUnread(report) {
         return false;
     }
 
-    const lastReadSequenceNumber = report.lastReadSequenceNumber || 0;
-    const maxSequenceNumber = report.maxSequenceNumber || 0;
-    return lastReadSequenceNumber < maxSequenceNumber;
+    // lastActionCreated and lastReadTime are both datetime strings and can be compared directly
+    const lastActionCreated = report.lastActionCreated || '';
+    const lastReadTime = report.lastReadTime || '';
+    return lastReadTime < lastActionCreated;
```
s77rt, line 1022: Edge case: If a chat report is empty (has no report actions except the `CREATED` action) then `lastReadTime` wouldn't be set and we fallback to an empty string and `'' < lastActionCreated` will be `true` making an empty chat unread (https://github.com/Expensify/App/issues/38368).

PR #13952, opened 2023-01-03
Update CODEOWNERS so that we don't accidentally deploy a new version of react-native-onyx

tgolen:
# Problem
`react-native-onyx` has had [this PR](https://github.com/Expensify/react-native-onyx/pull/211) merged to `main` which adds SQLite support, but we are not ready to deploy it to App yet.

# Solution
I will review all PRs with changes to `package.json` to ensure it doesn't get deployed.

### Tests
None

### Offline tests
None

### QA Steps
Internal QA:
1. Create a PR that modifies `package.json`
2. Verify that I am assigned to it for review

### PR Author Checklist
Checklist doesn't apply to any of the changes in this PR

### Screenshots/Videos
Screenshots don't apply to this PR

PR #13970, opened 2023-01-04
[No QA] Revert "Update CODEOWNERS so that we don't accidentally deploy a new version of react-native-onyx"

tgolen:
Reverts Expensify/App#13952

We are rolling out the SQLite upgrade today, so it's no longer necessary for me to review these PRs.

No testing or checklists required since this change doesn't affect the App code


puneetlath (approved):

puneetlath:
This is a simple revert that doesn't affect product-facing code, so the checklist didn't make sense in this scenario.

Issue #12268, opened 2022-10-28
[HOLD for payment 2023-02-08] [$250] Strip workspace name of html or handle validation php side reported by @kerupuksambel


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Open [staging.new.expensify.com](http://staging.new.expensify.com/)
2. Create new workspace
3. Change the name of the workspace with `<b> </b>` and save it

## Expected Result:
 Either the workspace name changed to <b> </b> after HTML-escaping (something like &lt;b&gt; &lt;/b&gt; ), or returned error  after the HTML tag got stripped and only the space remained

## Actual Result:
The workspace name changed to no name

## Workaround:
unknown

## Platform:

Where is this issue occurring?

- Web

**Version Number:** 1.2.21-4
**Reproducible in staging?:** y
**Reproducible in production?:** y
**Email or phone of affected tester (no customers):** 
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 

https://user-images.githubusercontent.com/43996225/198730117-135e716c-ada7-4069-b153-df931cee5fc1.mp4

https://user-images.githubusercontent.com/43996225/198730150-69afdab6-b896-461c-945c-97c8914b4e39.mp4

**Expensify/Expensify Issue URL:** 
**Issue reported by:** @kerupuksambel
**Slack conversation:** https://expensify.slack.com/archives/C01GTK53T8Q/p1666949886613589

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

NicMendonca:
I am having trouble reproducing this. Asked in slack [here](https://expensify.slack.com/archives/C01GTK53T8Q/p1666949886613589).

NicMendonca:
Okay, I got it. Am able to reproduce with using HTML spacing: 

![image](https://user-images.githubusercontent.com/24947345/199086995-f44e1b6f-fde3-4b93-bd3c-e21674277819.png)

ctkochan22:
Looks related to https://github.com/Expensify/App/issues/12000

trjExpensify:
[...]

PR #13978, opened 2023-01-04
Add workspace name validation in App

ctkochan22:
cc @youssef-lr 

### Details
We added back-end validation on the PHP side here: https://github.com/Expensify/Web-Expensify/pull/35548

We should now add front end validation as per convenstion

### Fixed Issues
$ https://github.com/Expensify/App/issues/12268   

### Tests
Same as QA

### QA Steps
1. Create or have a workspace ready
2. Go to Workspace Settings -> General Settings. And get ready to try various workspace name choices
3. With each name type, you may need to click away so that we test validation onBlur
    - [ ] Remove all text in name field, so that its completely empty. Verify that you see the inline error: 
        image
    - [ ] Just add spaces `   `. Verify that you see same error as above
        image
    - [ ] Now try a regular string `asdf`. Verify inline error goes away
    - [ ] Now add a html tag to that string. Verify that you an inline error pointing out that html tags are not allowed
        image

### PR Author Checklist

    - [ ] Android / native
    - [ ] Android / Chrome
    - [ ] iOS / native
    - [ ] iOS / Safari
    - [ ] MacOS / Chrome / Safari
    - [ ] MacOS / Desktop
    - [ ] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`)
- [ ] I verified that if a function's arguments changed that all usages have also been updated correctly

### Screenshots/Videos
<details>
<summary>Web</summary>
https://recordit.co/6z6ycgPqva

</details>

<details>
https://recordit.co/ypC4QXXLOI

</details>

<details>
https://recordit.co/cWD61wZmh9

</details>

<details>
<summary>Desktop</summary>
https://recordit.co/gkZdDVTQDo

</details>

<details>
<summary>iOS</summary>
https://recordit.co/DSwbQO3IiC

</details>

<details>
<summary>Android</summary>

</details>


src/pages/workspace/WorkspaceSettingsPage.js
```diff
@@ -55,20 +56,25 @@ class WorkspaceSettingsPage extends React.Component {
...
-        if (!values.name || !values.name.trim().length) {
+        const name = this.trimName(values.name);
+        if (!name || !name.length) {
             errors.name = this.props.translate('workspace.editor.nameIsRequiredError');
         }
         return errors;
     }
 
+    trimName(name) {
+        return name.replace(/<(.|\n)*?>/g, '').trim();
```
flodnv, line 75: Why do we have this replace() call?

ctkochan22: Whats wrong with it?

flodnv: Nothing, I am curious why we have it 😄

src/pages/workspace/WorkspaceSettingsPage.js
```diff
@@ -55,20 +56,31 @@ class WorkspaceSettingsPage extends React.Component {
...
         Keyboard.dismiss();
     }
 
     validate(values) {
         const errors = {};
-        if (!values.name || !values.name.trim().length) {
+        const name = values.name.trim();
+
+        if (this.hasHtml(name)) {
+            errors.name = 'HTML tags are not allowed in workspace names.';
```
youssef-lr, line 69: I think we we should use `props.translate` here? I'd ask for the Spanish version in https://expensify.slack.com/archives/C21FRDWCV

flodnv, line 69: Does the API enforce this too? 🤔

src/pages/workspace/WorkspaceSettingsPage.js
```diff
@@ -55,20 +56,31 @@ class WorkspaceSettingsPage extends React.Component {
...
+        }
+
+        if (!name || !name.length) {
             errors.name = this.props.translate('workspace.editor.nameIsRequiredError');
         }
+
         return errors;
     }
 
+    hasHtml(name) {
```
flodnv, line 79: This function is not really necessary given it's a 1 liner called in 1 place

mananjadhav: Are html tags the only restriction? Do we have any constraints on special characters?

src/pages/workspace/WorkspaceSettingsPage.js
```diff
@@ -55,20 +56,31 @@ class WorkspaceSettingsPage extends React.Component {
...
-        const name = values.name.trim();
         const outputCurrency = values.currency;
-        Policy.updateGeneralSettings(this.props.policy.id, name, outputCurrency);
+        Policy.updateGeneralSettings(this.props.policy.id, values.name, outputCurrency);
         Keyboard.dismiss();
     }
 
     validate(values) {
         const errors = {};
-        if (!values.name || !values.name.trim().length) {
```
mananjadhav, line 66: Will it cause any issue if we don't check `!values.name`? If for any reason `values.name` is not string, then `.trim()` will throw an error.

youssef-lr: I think that check is unnecessary as `values.name` is always guaranteed to be string from the `Form` component.

mananjadhav: Thanks for confirming @youssef-lr. Was just highligting.

src/pages/workspace/WorkspaceSettingsPage.js
ctkochan22: So currently the WAF input rules strip the html from the variable, as its under `FREETEXT`

ctkochan22: To allow html through, it looks like we need to downgrade the type to `ANYTHING`. 

I'd say, since html doesn't make it to php, lets just filter it out here. If it does get through somehow, it will be stripped, and whats left is saved as a workspace name.

src/pages/workspace/WorkspaceSettingsPage.js
ctkochan22: Testing the WAF input rules, it seems special characters are cool. We do aggressively go after anything that looks like a tag

src/pages/workspace/WorkspaceSettingsPage.js
ctkochan22: Oh Whoops

ctkochan22:
Updated

src/pages/workspace/WorkspaceSettingsPage.js
flodnv: Yeah that's fine

src/languages/en.js
```diff
@@ -944,7 +944,8 @@ export default {
         editor: {
             nameInputLabel: 'Name',
             nameInputHelpText: 'This is the name you will see on your workspace.',
-            nameIsRequiredError: 'You need to define a name for your workspace',
+            nameIsRequiredError: 'You need to define a name for your workspace.',
+            nameHasHtml: 'HTML tags are not allowed in workspace names.',
```
flodnv, line 948: I wonder how many of these errors we're going to have... most of our inputs are not allowed to have HTML 😬

src/languages/es.js
```diff
@@ -947,6 +947,7 @@ export default {
             nameInputLabel: 'Nombre',
             nameInputHelpText: 'Este es el nombre que verás en tu espacio de trabajo.',
             nameIsRequiredError: 'Debes definir un nombre para tu espacio de trabajo.',
+            nameHasHtml: 'Las etiquetas HTML no están permitidas en los nombres de espacios de trabajo.',
```
flodnv, line 950: Should you make Carlos's suggestion change here?

src/pages/workspace/WorkspaceSettingsPage.js
flodnv: This function is not really necessary given it's a 1 liner called in 1 place

flodnv (changes requested):

src/languages/en.js
flodnv: Maybe we need something more generic where we prevent HTML in all inputs? 😅

ctkochan22: I'm confused. Are you wanting to check for html in more fields? 

Or do you want to make this error message more generic so that it can be used in more places?

flodnv: I am saying this bug exists in more fields, eg first and last name, but not only : it exists in most fields in our app.

What I am suggesting is that instead of waiting for the next 42 bug reports to say the same thing, we should fix this in all text inputs that don't accept HTML (which is probably 99.999% of them).

mananjadhav: qq, @flodnv are you suggesting it to be a follow up PR or a part of this issue?

flodnv: I think either is fine, but why not this PR, because this bug doesn't really matter much. What matters more is to fix it holistically instead of 42 times.

ctkochan22: I think thats fine. I'm worried this will just get dragged on the larger the scope is. Rather than just doing it for this now, and the next time this is an issue. 

The author DRY's this logic up so it can be used in two places. Then each subsequent time, people can add that method. 

But we can drop it in the issue

ctkochan22: Discussed here: https://github.com/Expensify/App/issues/12268#issuecomment-1405266063

We will tackle the validation and testing/QA in this GH: https://github.com/Expensify/App/issues/14610

ctkochan22:
@mananjadhav I think we are ready to move forward. Can you test this on all platforms please? 🙇

flodnv (approved):

mananjadhav (approved):

flodnv:
>With this change we're not allowing <  any number of characters event with space  >. I hope this is fine?

Great point, I did realize the same but did not mention it. I think it's ok, but perhaps the more holistic solution should fix that. What do you think @ctkochan22 ?

ctkochan22:
Yeah, I thought it was fine because our WAF input rules does the same thing. Anything within `< >` are filtered out. So this applies for fields across the board, unless specifically designated as `_ANYTHING`

ctkochan22:
So without this validation or this PR. `<beta> <beta>` would still be stripped into an empty space

PR #14502, opened 2023-01-24
Do not select participant in IOU confirmation list

luacmartins:
### Details
Prevents the `focusedIndex` being set to 0 when the `IOUConfimationList` component updates

### Fixed Issues
$ https://github.com/Expensify/App/issues/14441

### Tests
1. Open the App in 2 sessions logged in to the same account
2. In session A, navigate to `+ > Request money/Send money > Enter Amount > Select participant > Confirmation page`
3. In session B, navigate to `Settings > Preferences` and change the selected language
4. In session A, verify that the selected participant is not highlighted

- [x] Verify that no errors appear in the JS console

### Offline tests
N/A

### QA Steps
Same as tests.

- [ ] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
[...]

PR #14788, opened 2023-02-03
Revert "Do not select participant in IOU confirmation list"

roryabraham:
Reverts Expensify/App#14502

Straight revert


roryabraham:
Was just testing, does not seem like https://github.com/Expensify/App/pull/14502 was the cause of this regression.

Issue #14787, opened 2023-02-03
Arrow key navigation stopped working


roryabraham:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

There is a regression on staging. All the testing / reproduction steps can be found [here](https://github.com/Expensify/Expensify/issues/217359#issue-1293575797):

Log in to New Expensify.

##### All Platforms

1. Click/tap on the magnifying glass icon to open the search page.
2. Click/tap on the x icon to close the search page.
3. Click/tap on the magnifying glass to open the search page.
4. Ensure you can scroll up and down in the results.
5. Enter a search query – the list should update accordingly.
6. Click/tap on a result: the search page should close and that chat should open.

##### Web/Desktop

1. Press CMD+K to open the search page.
1. Press Esc to close the search page.
1. Press CMD+K to reopen the search page.
1. Type in a search query.
1. Use the arrow keys to go through the search results. Make sure that the list "wraps", so:
    - If you are focused on the last result and press ArrowDown you are taken to the first result
    - If you are focused on the first result and press ArrowUp you are taken to the last result
1. With the focus somewhere in the middle of the list, update the search input such that there are no results
1. Press the arrow keys a few times – nothing should happen.
1 Undo the change from the previous step by clearing the search input (press `ALT+Delete`). The focus should be on the first element of the list, and you should be able to use the arrows to navigate focus (as before)
1. Enter a new search term. Verify that the first result in the list is highlighted, and that you can navigate up and down from there using the arrow keys.
1. With a chat highlighted on the search page, press Enter. The search page should close and that chat should open.
1. Use the mouse to click on the magnifying glass icon and reopen the search page.
1. Press Esc to close the search page.
1. Click on the magnifying glass to reopen the search page.
1. Use the arrow keys and the enter key to select and open a different chat.
1. Press `CMD+K` to reopen the search page.
1. Use the arrows keys to highlight a search result, then press `CMD+Enter`. The search page should close and your selected chat should open.

##### All Platforms

1. Click/tap on the green + icon on the home screen
2. Click/tap on New Chat.
3. Ensure that you can scroll the list of results.
4. Enter a search query and the results should update accordingly.
5. click/tap on a search result
6. The New Chat Page should close and a chat with your selected result should open.

##### Web/Desktop

1. Click on the green + icon on the home screen
[...]

PR #14790, opened 2023-02-03
Revert "fix: Timezone changes to some 1st option in the list while pressing the enter key"

roryabraham:
Reverts Expensify/App#14466

Straight revert


roryabraham:
Just testing, this does not seem to resolve [this regression](https://github.com/Expensify/App/issues/14787)

PR #15146, opened 2023-02-14
Demote warning to info

iwiznia:
Context https://expensify.slack.com/archives/C03TQ48KC/p1676402115548569

### Tests
No tests, just a log line change

### QA Steps
- Send a message to chronos
- Check in the network tab that we are not calling the Log API command dozens of times (one or two is fine)


roryabraham (approved):
Approving because it's just a log change.

iwiznia:
Merging without checklists as this is only a log line change

Issue #15087, opened 2023-02-13
[HOLD for payment 2023-03-28] Make sure attachments are fetched from appropriate servers


mountiny:
Coming from [here](https://github.com/Expensify/App/issues/14227#issuecomment-1426112286)

### Problem

Chat attachments are loaded from API not matching the client environment. So you can be on Staging app but the chat attachment is loaded from production

### Why is it important

Environment inconsistency can make it harder to find regressions and can introduce some unexpected ones in one deployment cycle

### Solution

Make sure the production App loads the chat attachments from production api and staging from appropriate api server. cc @kidroca

MelvinBot:
Triggered auto assignment to @zanyrenney (`Bug`), see https://stackoverflow.com/c/expensify/questions/14418 for more details.

MelvinBot:
### Bug0 Triage Checklist ([Main S/O](https://stackoverflowteams.com/c/expensify/questions/14418))

- [ ] This "bug" occurs on a supported platform (ensure `Platforms` in OP are ✅)
- [ ] This bug is not a duplicate report (check [E/App issues](https://github.com/Expensify/App/issues) and [#expensify-bugs](https://expensify.slack.com/archives/C049HHMV9SM))
    - If it is, comment with a link to the original report, close the issue and add any novel details to the original issue instead
- [ ] This bug is reproducible using the reproduction steps in the OP. [S/O](https://stackoverflowteams.com/c/expensify/questions/14418#:~:text=The%20bug%20is%20reproducible)
    - If the reproduction steps are clear and you're unable to reproduce the bug, check with the reporter and QA first, then close the issue.
    - If the reproduction steps aren't clear and you determine the correct steps, please update the OP.
- [ ] This issue is filled out as thoroughly and clearly as possible
    - Pay special attention to the title, results, platforms where the bug occurs, and if the bug happens on staging/production.
- [ ] I have reviewed and subscribed to the linked Slack conversation to ensure Slack/Github stay in sync

mountiny:
@kidroca Do you want to get assigned to this and work hourly for this issue?

kidroca:
> @kidroca Do you want to get assigned to this and work hourly for this issue?

Yes!

zanyrenney:
Chatted to @mountiny in DM - I can unassign as kidroca has their own contract with expensify for bug fixes!

kidroca:
Found another discrepancy here: 

[...]

Issue #14227, opened 2023-01-11
[HOLD for payment 2023-02-08] [$1000] Backend - Attachment links capture the environment they are from


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Use https://staging.new.expensify.com/ to chat with somebody
2. Send an attachment
3. Use https://new.expensify.com/ to open the same conversation
4. Inspect the URL for the attachment
5. It will say it's loaded from https://staging.expensify.com/chat-attachments/ even though you are on new.expensify

## Expected Result:
all attachment URLs to be resolved relative to current environment
E.g. when I’m on staging.new I expect all attachments to be resolved from https://staging.new.expensify.com/chat-attachments/…
Or when Im on prod I expect all attachments to be resolved from https://new.expensify.com/chat-attachments/…

## Actual Result:
some attachment urls are resolved from [staging.expensify.com](http://staging.expensify.com/) others from [new.expensify.com](http://new.expensify.com/) 

## Workaround:
unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?

- [ ] MacOS / Chrome / Safari

**Version Number:** 1.2.52-4
**Reproducible in staging?:** y
**Reproducible in production?:** y
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 
![image](https://user-images.githubusercontent.com/43996225/211927253-2ca18c54-f8a4-4a35-abfc-feef9441c85c.png)
<img width="960" alt="Untitled" src="https://user-images.githubusercontent.com/43996225/211927271-7b5a895f-19fb-4bf6-a99b-c019abe602ba.png">

**Expensify/Expensify Issue URL:**
**Issue reported by:** @kidroca
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1673289214225929

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
[...]

PR #14944, opened 2023-02-08
Use app environment to determine olddot URL and update staging env

MonilBhavsar:
cc @AndrewGable 

### Details

- Update staging env file to point to staging olddot
- Use Environment to determine olddot domain/URL

### Fixed Issues

$ https://github.com/Expensify/App/issues/14820   
PROPOSAL: https://expensify.slack.com/archives/C03TQ48KC/p1675664867491769

#### Use case 1

1. Go to settings > Workspace.
2. Go to Reimburse Expense > Click on "View all receipts" and confirm you're redirected to appropriate olddot (if using staging then staging, if using production, then production)
3. Go to Pay Bills > Click on "View all bills" and confirm you're redirected to appropriate olddot (if using staging then staging, if using production, then production)
4. Go to Send Invoices > Click on "Send Invoices" and "View all invoices" and confirm you're redirected to appropriate olddot (if using staging then staging, if using production, then production)

#### Use case 2

Have 2FA disabled on olddot.

1. Go to Workspace > Connect bank account and start the process of setting up bank account in VERIFYING state
2. Go till Verifying state and confirm the 2FA prompt
4. Click on "Secure Account" and confirm you're redirected to appropriate olddot (if using staging config then staging, if using production config, then production)

Now, updated .env file and added `ENVIRONMENT=staging`
Tested both use cases above, ensured it redirects to staging olddot URL

Similarly, updated .env file and added `ENVIRONMENT=production`
Tested both use cases above, ensured it redirects to production olddot URL

For mobile native, updated `getEnvironment()` in `src/libs/Environment/getEnvironment/index.native.js` to return early and simulate staging env

```diff
function getEnvironment() {
    return new Promise((resolve) => {
        // If we've already set the environment, use the current value
+        return resolve(CONST.ENVIRONMENT.STAGING);

        if (environment) {
```        

#### Use case 3  - building Staging app

1. Followed this [SO](https://stackoverflowteams.com/c/expensify/questions/8638) to point to staging app. Updated env file similar to .env.staging file to point to staging API. 
```
Copy the contents of the .env.staging file to the .env file.

Run npm run desktop
```
2. In preferences >  disabled "Use staging server", so that API points to production
3. Go to Reimburse Expense > Click on "View all receipts" and confirm you're redirected to appropriate olddot (if using staging then staging, if using production, then production)

**_Note_**:
- (For internal engineer) For testing on mobile web, update `DEV_EXPENSIFY_URL` in CONST.js to your ngrok URL

[...]

PR #15178, opened 2023-02-15
Refactor: Create libs/ApiUtils

kidroca:
### Details

- Introduce `ApiUtils` and `ApiUtils.getApiRoot` 

#### Why?
https://expensify.slack.com/archives/C01GTK53T8Q/p1676320010253979
We build native apps using the PROD env config
The API is resolved dynamically at run time
But we also have the constant `URL_API_ROOT` which cannot always point to the correct API (it points to PROD)
Some ENVs (staging) allow for toggling between different APIs and a function like `getApiRoot` encapsulates
the checks we have to make when we want to use the root url 

The staging toggle now works correctly on DEV and it can switch between the PRIMARY and STAGING APIs

### Fixed Issues

$ https://github.com/Expensify/App/issues/15087
PROPOSAL: N/A

### Tests

- [ ] Verify staging and prod urls are resolved from API ROOT
    - Follow the steps from the issue description https://github.com/Expensify/App/issues/14227#issue-1529764776 in order to create attachment links from staging and prod. 
    - Inspect the Network tab or the HTML source (Chrome Dev tools -> Elements tab) to verify any attachment url (staging or prod) is resolved against the API ROOT of the current environment
- [ ] Verify absolute urls are resolved from API ROOT
    - At the moment the backend is not yet saving attachments using the absolute path, so we need to simulate the result
    - We need to inspect and modify Onyx data
    - Find an attachment action and modify the `html` of the message so that the `` part becomes ``
        <details>
        <summary>Example</summary>
        In Chrome dev console:
       
        ```js
        Onyx.connect({ key: 'reportActions_71955477', callback: actions => {
            const action = actions[932];
            action.message[0].html = action.message[0].html.replace('https://staging.expensify.com/', '/');
            Onyx.merge('reportActions_71955477', { 932: action });
        }});
        ```
        </details>
    - Now refresh the browser. The (modified) attachment should resolve and load successfully from API ROOT
- [ ] Verify all attachment usages are covered. All these places should resolve the attachment from API ROOT
    - attachment thumbnail in chat message comments 
    - attachment preview modal
    - attachment download button
    - use different attachment types - image formats / document formats

- [ ] Settings / Preferences / Test Preferences - staging toggle works correctly on DEV and STAGING
    - It should switch between PRIMARY (PROD) and STAGING urls
    - Keep in mind that on WEB using the STAGING API would case CORS errors, it needs further changes to the proxy server covered in the follow-up for https://github.com/Expensify/App/pull/15517
- [ ] Using the "Staging toggle" changes the attachment URLs 
    - switching to the staging api and viewing attachment requests on the network tab should show `staging.expensify.com/chat-attachments/...` (or whatever STAGING_EXPENSIFY_URL address is configured in .env)
    - You can use Desktop to inspect this

- [ ] Pusher functionality should be unaffected
    - there was a rename that touched Pusher code, but nothing was altered. Pusher should still subscribe and be able to receive new messages live

- [ ] `AddressSearch` component functionality was refactored, verify `GooglePlacesAutocomplete` has no regression regarding the `requestUrl` parameter. The fields below should show autocomplete suggestions
  - http://localhost:8080/settings/payments/add-debit-card (Billing Address)
  - http://localhost:8080/settings/profile/personal-details/address (Address Line 1)

### Offline tests

Expected Offline behavior
- Images that were seen recently should be visible and render instantly (from cache)
- Images that haven't been already loaded would be blank

Expected slow internet behavior
- Images that were seen recently should be visible and render instantly (from cache)
- Other images would display a spinner / loader animation for a while until they load

#### Testing on STAGING

1. Go to PRODUCTION and upload attachments
2. Go to STAGING and view the uploaded attachments

- [ ] Attachments should load successfully
- [ ] Requests on the network tab should be directed to `staging.expensify.com/chat-attachments/...`
- [ ] Upload different attachment types and verify they work
- [ ] Upload attachments from STAGING - uploaded and viewing them should still work
- [ ] Verify all attachment usages are covered. All these places should load and render successfully
    - attachment thumbnail in chat message comments 
    - attachment preview modal
    - attachment download button
    - use different attachment types - image formats / document formats
- [ ] Settings / Preferences / Test Preferences - Staging toggle is rendered and we can switch between PROD and STAGING API
- [ ] Using the "Staging toggle" changes the attachment URLs 
    - switching to the prod api and viewing attachment requests on the network tab should show `www.expensify.com/chat-attachments/...`
- [ ] `AddressSearch` component functionality was refactored, verify address can still be autocomplete 
    - e.g. Address Line 1 here: https://staging.new.expensify.com/settings/profile/personal-details/address
    - another place to check is "Add Debit Card" (settings / Payments), for the Billing address here: https://staging.new.expensify.com/settings/payments

#### Testing on PRODUCTION

_After PR is merged to PROD_

1. Go to STAGING and upload attachments
2. Go to PRODUCTION and view the uploaded attachments

- [ ] Attachments should load successfully
- [ ] Requests on the network tab should be directed to `www.expensify.com/chat-attachments/...`
- [ ] Upload different attachment types and verify they work
- [ ] Upload attachments from PRODUCTION - uploaded and viewing them should still work
- [ ] Verify all attachment usages are covered. All these places should load and render successfully
    - attachment thumbnail in chat message comments 
    - attachment preview modal
    - attachment download button
    - use different attachment types - image formats / document formats
- [ ] Settings / Preferences / Test Preferences - staging toggle is not rendered on PROD
- [ ] `AddressSearch` component functionality was refactored, verify address can still be autocomplete 
    - e.g. Address Line 1 here: https://new.expensify.com/settings/profile/personal-details/address
    - another place to check is "Add Debit Card" (settings / Payments), for the Billing address here: https://new.expensify.com/settings/payments

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

Screenshot 2023-02-17 at 9 54 47

</details>

<details>

image

</details>

<details>

Screenshot 2023-02-17 at 9 57 37

</details>

<details>
<summary>Desktop</summary>

Screenshot 2023-02-17 at 10 00 07

</details>

<details>
<summary>iOS</summary>

Screenshot 2023-03-10 at 23 09 17
Screenshot 2023-03-10 at 23 12 11

</details>

<details>
<summary>Android</summary>

image
image
image

</details>


src/CONFIG.js
```diff
@@ -49,11 +49,11 @@ export default {
...
         EXPENSIFY_URL: expensifyURL,
-        SECURE_EXPENSIFY_URL: secureURLRoot,
         NEW_EXPENSIFY_URL: newExpensifyURL,
-        URL_API_ROOT: expensifyURLRoot,
-        STAGING_EXPENSIFY_URL: stagingExpensifyURL,
-        STAGING_SECURE_EXPENSIFY_URL: stagingSecureExpensifyUrl,
+        PRODUCTION_API_ROOT: expensifyURLRoot,
+        PRODUCTION_SECURE_API_ROOT: secureURLRoot,
+        STAGING_API_ROOT: stagingExpensifyURL,
+        STAGING_SECURE_API_ROOT: stagingSecureExpensifyUrl,
```
kidroca, line 56: The reason for renaming `URL_API_ROOT`  to `PRODUCTION_API_ROOT` is because the underlying value of `URL_API_ROOT` is always the production api, we want to make that clear and avoid confusion

The API (on some envs) can be changed dynamically. `URL_API_ROOT` variable can be mistaken to always point to the correct API for the current ENV, so instead we make it clear by renaming it to `PRODUCTION_API_ROOT` and introduce `ApiUtils.getApiRoot()`

src/components/AddressSearch.js
```diff
@@ -184,7 +184,7 @@ const AddressSearch = (props) => {
                     query={query}
                     requestUrl={{
                         useOnPlatform: 'all',
-                        url: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=Proxy_GooglePlaces&proxyUrl=`,
+                        url: `${CONFIG.EXPENSIFY.PRODUCTION_API_ROOT}api?command=Proxy_GooglePlaces&proxyUrl=`,
```
kidroca, line 187: Should we use `ApiUtils.getRootApi` here or is this intentionally meant to always use the PROD api? 
_(in that case the rename makes it clear)_

src/components/TestToolMenu.js
```diff
@@ -31,7 +29,7 @@ const propTypes = {
 
 const defaultProps = {
     user: {
-        shouldUseStagingServer: false,
+        shouldUseStagingServer: undefined,
```
kidroca, line 32: Providing a default value interferes with the `loadasGet` call, where we provide a default value 

If there is a default boolean value, then `lodashGet` would never use the fallback we provide

Maybe this got unnoticed because of how `withOnyx` works, but IMO the plan is to use the Onyx value
And fallback to something (specific) when Onyx does not have a value

The specific thing we want to fallback to is whether we use the staging api right now or not
Unfortunately this cannot be evaluated at the time we declare `defaultProps`

src/components/TestToolMenu.js
```diff
@@ -43,12 +41,18 @@ const TestToolMenu = props => (
...
+                <TestToolRow title="Use Staging Server">
+                    <Switch
+                        isOn={lodashGet(props, 'user.shouldUseStagingServer', ApiUtils.isUsingStagingApi())}
+                        onToggle={() => User.setShouldUseStagingServer(
+                            !lodashGet(props, 'user.shouldUseStagingServer', ApiUtils.isUsingStagingApi()),
+                        )}
+                    />
+                </TestToolRow>
+            )
+        }
```
kidroca, line 55: Show the toggle only on STAGING, because the toggle doesn't work in DEV: https://github.com/Expensify/App/issues/15087#issuecomment-1431364109

Default the `isOn` value (when Onyx data is missing) to whether we currently use the staging api or not

src/libs/HttpUtils.js
```diff
@@ -1,20 +1,9 @@
...
-import shouldUseStagingServer from './shouldUseStagingServer';
-import getPlatform from './getPlatform';
-
-// Desktop and web use staging config too so we we should default to staging API endpoint if on those platforms
-const shouldDefaultToStaging = _.contains([CONST.PLATFORM.WEB, CONST.PLATFORM.DESKTOP], getPlatform());
-let stagingServerToggleState = false;
-Onyx.connect({
-    key: ONYXKEYS.USER,
-    callback: val => stagingServerToggleState = lodashGet(val, 'shouldUseStagingServer', shouldDefaultToStaging),
-});
```
kidroca, line 17: This logic was pretty much moved/replaced by ApiUtils

src/libs/HttpUtils.js
```diff
@@ -107,13 +96,8 @@ function xhr(command, data, type = CONST.NETWORK.METHOD.POST, shouldUseSecure =
         formData.append(key, val);
     });
 
-    let apiRoot = shouldUseSecure ? CONFIG.EXPENSIFY.SECURE_EXPENSIFY_URL : CONFIG.EXPENSIFY.URL_API_ROOT;
-
-    if (shouldUseStagingServer(stagingServerToggleState)) {
-        apiRoot = shouldUseSecure ? CONFIG.EXPENSIFY.STAGING_SECURE_EXPENSIFY_URL : CONFIG.EXPENSIFY.STAGING_EXPENSIFY_URL;
-    }
```
kidroca, line 114: This logic was pretty much moved/replaced by ApiUtils

src/libs/HttpUtils.js
```diff
@@ -107,13 +96,8 @@ function xhr(command, data, type = CONST.NETWORK.METHOD.POST, shouldUseSecure =
...
-
-    return processHTTPRequest(`${apiRoot}api?command=${command}`, type, formData, data.canCancel);
+    const url = ApiUtils.getCommandUrl({shouldUseSecure, command});
+    return processHTTPRequest(url, type, formData, data.canCancel);
```
kidroca, line 100: Perhaps we want to stick doing `${apiRoot}api?command=${command}` here in HttpUtils.js instead of having `getCommandUrl` 

I saw another place that uses `{root}api?command=...` so it might be handy to have something reusable like `getCommandUrl`

src/libs/Navigation/AppNavigator/AuthScreens.js
```diff
@@ -97,7 +97,7 @@ class AuthScreens extends React.Component {
         Pusher.init({
             appKey: CONFIG.PUSHER.APP_KEY,
             cluster: CONFIG.PUSHER.CLUSTER,
-            authEndpoint: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=AuthenticatePusher`,
+            authEndpoint: `${CONFIG.EXPENSIFY.PRODUCTION_API_ROOT}api?command=AuthenticatePusher`,
```
kidroca, line 100: Should we use `ApiUtils.getRootApi` here or is this intentionally meant to always use the PROD api? 
_(in that case the rename makes it clear)_

src/libs/tryResolveUrlFromApiRoot.js
```diff
@@ -1,9 +1,15 @@
...
 
 // Absolute URLs (`/` or `//`) should be resolved from API ROOT
 // Legacy attachments can come from either staging or prod, depending on the env they were uploaded by
 // Both should be replaced and loaded from API ROOT of the current environment
-const ORIGINS_TO_REPLACE = ['/+', Config.EXPENSIFY.EXPENSIFY_URL, Config.EXPENSIFY.STAGING_EXPENSIFY_URL];
+const ORIGINS_TO_REPLACE = [
+    '/+',
+    Config.EXPENSIFY.EXPENSIFY_URL,
+    Config.EXPENSIFY.STAGING_API_ROOT,
+    Config.EXPENSIFY.PRODUCTION_API_ROOT,
```
kidroca, line 11: Added the PRODUCTION_API_ROOT here to complete the list and make it clear that it is handled

I think for staging and production envs EXPENSIFY_URL == PRODUCTION_API_ROOT, but in any case now it's clearer that we handle both staging and prod cases here

tests/actions/ReportTest.js
```diff
@@ -42,7 +42,7 @@ describe('actions/Report', () => {
         Pusher.init({
             appKey: CONFIG.PUSHER.APP_KEY,
             cluster: CONFIG.PUSHER.CLUSTER,
-            authEndpoint: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=AuthenticatePusher`,
+            authEndpoint: `${CONFIG.EXPENSIFY.PRODUCTION_API_ROOT}api?command=AuthenticatePusher`,
```
kidroca, line 45: Should we use `ApiUtils.getRootApi` here or is this intentionally meant to always use the PROD api? 
_(in that case the rename makes it clear)_

tests/ui/UnreadIndicatorsTest.js
```diff
@@ -46,7 +46,7 @@ beforeAll(() => {
     Pusher.init({
         appKey: CONFIG.PUSHER.APP_KEY,
         cluster: CONFIG.PUSHER.CLUSTER,
-        authEndpoint: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=AuthenticatePusher`,
+        authEndpoint: `${CONFIG.EXPENSIFY.PRODUCTION_API_ROOT}api?command=AuthenticatePusher`,
```
kidroca, line 49: Should we use `ApiUtils.getRootApi` here or is this intentionally meant to always use the PROD api? 
_(in that case the rename makes it clear)_

kidroca (commented):
I've pointed out some usages of `URL_API_ROOT` that might be confusing or be set by mistake to the production API root, because `URL_API_ROOT` always points there

mountiny:
cc @tgolen @marcaaron if you would want to review this too 🙌

kidroca:
Added description, tests and QA tests and screenshots 

✅  Ready for Review

_We still need to decide what to do about the remaining direct usages of `PRODUCTION_API_ROOT`_

mountiny:
Just found there is this PR as well which is a bit stepping on our toes here https://github.com/Expensify/App/pull/14944/files

ctkochan22:
@mountiny Are we still reviewing this PR? Or wait until yours is merged?

Santhosh-Sellavel:
@kidroca Can you please update the test steps here, instead of links?

kidroca:
@Santhosh-Sellavel 
> @kidroca Can you please update the test steps here, instead of links?

Done, but I think we should wait for the resolution of the PR here (which might call for additional changes on my PR):
-  #14944

Santhosh-Sellavel:
Cool, then can you update the title HOLD For the PR?

mountiny:
Put this on hold, @ctkochan22 sorry was ooo on monday and tuesday

ctkochan22:
@mountiny Looks like you got a handle on this review. I'm going to unassign and defer to you!

src/CONFIG.js
```diff
@@ -49,11 +49,13 @@ export default {
...
         NEW_EXPENSIFY_URL: newExpensifyURL,
-        URL_API_ROOT: expensifyURLRoot,
-        STAGING_EXPENSIFY_URL: stagingExpensifyURL,
-        STAGING_SECURE_EXPENSIFY_URL: stagingSecureExpensifyUrl,
+
+        // The main API is usually PRODUCTION, but there are cases where it points to internal DEV server
+        PRIMARY_API_ROOT: expensifyURLRoot,
+        PRIMARY_SECURE_API_ROOT: secureURLRoot,
+        STAGING_API_ROOT: stagingExpensifyURL,
+        STAGING_SECURE_API_ROOT: stagingSecureExpensifyUrl,
```
kidroca, line 58: The reason for renaming `URL_API_ROOT`  to `PRIMARY_API_ROOT` is because the underlying value of `URL_API_ROOT` is (almost) always the production api, we want to make that clear and avoid confusion
(Though it's also possible point to an internal DEV url)

The API (on some envs) can be changed dynamically. `URL_API_ROOT` variable can be mistaken to always point to the correct API for the current ENV, so instead we make it clear by renaming it to `PRIMARY_API_ROOT` and introduce `ApiUtils.getApiRoot()` 

We have a "Staging toggle" that can switch between primary and staging 
It should be more clear that it switches between the two, and that it's not always between staging and prod

src/libs/Navigation/AppNavigator/AuthScreens.js
```diff
@@ -98,7 +98,7 @@ class AuthScreens extends React.Component {
         Pusher.init({
             appKey: CONFIG.PUSHER.APP_KEY,
             cluster: CONFIG.PUSHER.CLUSTER,
-            authEndpoint: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=AuthenticatePusher`,
+            authEndpoint: `${CONFIG.EXPENSIFY.PRIMARY_API_ROOT}api?command=AuthenticatePusher`,
```
kidroca, line 101: Should we use `ApiUtils.getRootApi` here or is this intentionally meant to always use the PRIMARY api? 

It was pointed out in slack that it's probably best to leave it out for the time being 
- Don't try to change pusher connection when API is switched
- Always use the primary api for the current environment for the pusher connection

src/libs/tryResolveUrlFromApiRoot.js
```diff
@@ -1,9 +1,15 @@
...
 // Absolute URLs (`/` or `//`) should be resolved from API ROOT
 // Legacy attachments can come from either staging or prod, depending on the env they were uploaded by
 // Both should be replaced and loaded from API ROOT of the current environment
-const ORIGINS_TO_REPLACE = ['/+', Config.EXPENSIFY.EXPENSIFY_URL, Config.EXPENSIFY.STAGING_EXPENSIFY_URL];
+const ORIGINS_TO_REPLACE = [
+    '/+',
+    Config.EXPENSIFY.EXPENSIFY_URL,
+    Config.EXPENSIFY.STAGING_API_ROOT,
+    Config.EXPENSIFY.PRIMARY_API_ROOT,
+];
```
kidroca, line 12: Added the PRIMARY_API_ROOT here to complete the list and make it clear that it is handled

I think for staging and production envs EXPENSIFY_URL == PRIMARY_API_ROOT, but in any case now it's clearer that we handle both staging and prod cases here

kidroca (commented):
This PR is ready for review and testing ✅ 

Made some changes to be compatible with other PRs touching the same functionality
It was possible to remove the platform specific implementations for `ApiUtils` sub functionality

We are no longer blocked by:
- https://github.com/Expensify/App/pull/14944 - no overlapping changes
- https://github.com/Expensify/App/pull/15517 - agreed to move with my PR first [on slack](https://expensify.slack.com/archives/C01GTK53T8Q/p1678119621723989?thread_ts=1678113020.610549&cid=C01GTK53T8Q)

---

Updated PR **description** and **test** sections to indicated the "Staging toggle" can be used in DEV
It would cause CORS issues (only on DEV), because the proxy server changes are not there yet. They will be handled in the follow of https://github.com/Expensify/App/pull/15517

src/CONFIG.js
marcaaron: >the underlying value of URL_API_ROOT is (almost) always the production api, we want to make that clear and avoid confusion

The url is based on the environment variables so maybe we can be more explicit to say `ENVIRONMENT_API_ROOT`? Something about `PRIMARY` makes me wonder if there are "other additional APIs" or something to worry about.

src/components/AddressSearch.js
```diff
@@ -184,7 +184,7 @@ const AddressSearch = (props) => {
                     query={query}
                     requestUrl={{
                         useOnPlatform: 'all',
-                        url: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=Proxy_GooglePlaces&proxyUrl=`,
+                        url: `${CONFIG.EXPENSIFY.PRIMARY_API_ROOT}api?command=Proxy_GooglePlaces&proxyUrl=`,
```
marcaaron, line 187: Should this use `ApiUtils.getApiRoot()`?

src/components/TestToolMenu.js
marcaaron: >Unfortunately this cannot be evaluated at the time we declare defaultProps

Meaning we can't make the defaultProps use `ApiUtils.isUsingStagingApi()`? That seems fine. 

**alternate idea (non blocking)**: Possibly could be solved by passing a default value for `user.shouldUseStagingServer` to Onyx during initialization and making it a required prop here. Then we could eliminate the `lodashGet` calls as well.

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,78 @@
...
+
+let stagingServerToggleState;
+Onyx.connect({
+    key: ONYXKEYS.USER,
+    callback: (val) => {
+        stagingServerToggleState = lodashGet(val, 'shouldUseStagingServer');
+    },
+});
+
+function shouldUseStagingServer() {
```
marcaaron, line 24: JSDoc with `@returns {Boolean}` for this? 🙇

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,78 @@
...
+    // Choosing between staging and prod/dev is only available on DEV and STAGING
+    if (ENV_NAME === CONST.ENVIRONMENT.PRODUCTION) {
+        return false;
+    }
+
+    if (typeof stagingServerToggleState !== 'boolean') {
```
marcaaron, line 30:
```suggestion
    if (!_.isBoolean(stagingServerToggleState)) {
```

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,78 @@
...
+        return ENV_NAME === CONST.ENVIRONMENT.STAGING;
```
marcaaron, line 31: Could we make the default value for `stagingServerToggleState` be this comparison? Then simplify this method to:

```js
function shouldUseStagingServer() {
    return ENV_NAME === CONST.ENVIRONMENT.PRODUCTION 
        ? false 
        : stagingServerToggleState;
}
```

src/libs/ApiUtils.js
marcaaron: This function has the same name as the property of the Onyx key. Can we rename it to something like `canUseStagingServer()`?

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,78 @@
...
+        : CONFIG.EXPENSIFY.PRIMARY_API_ROOT;
+}
+
+/**
+ * Get the command url for the given request
+ *
+ * @param {Object} request
+ * @param {Boolean} request.shouldUseSecure
+ * @param {String} request.command - the name of the API command
+ * @returns {string}
```
marcaaron, line 63:
```suggestion
 * @returns {String}
```

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+function getCommandUrl(request) {
+    return `${getApiRoot(request)}api?command=${request.command}`;
+}
+
+function isUsingStagingApi() {
```
marcaaron, line 69: `@returns {Boolean}`

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,78 @@
...
+    return getApiRoot({shouldUseSecure: false}) === CONFIG.EXPENSIFY.STAGING_API_ROOT;
```
marcaaron, line 70: **small suggestion:** Make passing the `request` optional to `getApiRoot()` by default it will return the non-secure url. Otherwise, if there's a request it can look at that. Passing the "fake request" here can be removed.

src/libs/Navigation/AppNavigator/AuthScreens.js
marcaaron: I am 👍 on leaving it out. There is maybe a very small chance that we would want to test the staging Pusher auth endpoint under any circumstance. The workaround would be to change the environment variables which is not hard to do on dev (but impossible on production).

src/libs/HttpUtils.js
marcaaron: Either seem fine to me. Could go "rule of 3" on this one.

marcaaron (commented):
Looks pretty good. Just left some small ideas for improvements.

src/CONFIG.js
kidroca: > Something about PRIMARY makes me wonder if there are "other additional APIs" or something to worry about.

That's what I'm trying to do here - prevent people from blindly using it

Problem: "Developers don't know the API root can change dynamically and that the change would not be reflected in `URL_API_ROOT`" 
Instead they should use `getApiRoot()`

> The url is based on the environment variables so maybe we can be more explicit to say ENVIRONMENT_API_ROOT?

That sounds good, but it's a bit ambiguous to me - I don't make the connection that it's about env variables, but I think it's about the environment name e.g. staging | prod | dev  
So I would think that on staging ENVIRONMENT_API_ROOT is `staging.exepsnify.com`

And this leads to Problem 2: "Developers think `URL_API_ROOT` always refers to the build environment"
I initially renamed `URL_API_ROOT` to `PROD_API_ROOT`, to make that clear
- on DEV (external contributors) it would referer to `www.expensify`)
- on PROD it refer to `www.expensify`
- on STAGING **it would again refer to `www.expensify`** (see .env.staging) 

Some developers (me included) used to think that, on staging `URL_API_ROOT` would be `staging.expensify` when it's actually `www.expensify` 

This is an inconsistency problem which remained hidden for a very long time, because it didn't result in noticeable issues - when we're on staging we're allowed use the prod api and some requests (like attachments and images) would use it

The reason this happens is because some links are configured to use `URL_API_ROOT` which always points to PROD (unless overridden by internal developer)

This PR aims to expose such usages by renaming the constant

I selected the word PRIMARY, because of the "Staging toggle" in settings - it switches between PRIMARY and STAGING, where PRIMARY
- on PROD is prod
- on STAGING is **prod**
- on DEV it's either prod or env value (internal dev) 

---

This is not the main goal of the PR, I just wanted to make the rename and draw attention to usages, so maybe we should revert the renames and do a separate PR for the problems outlined here 

It might be best to declare all URLs - staging / dev / prod and sort out the logic in `ApiUtils`, so we don't have a constant like PRIMARY that might be PROD or internal DEV url

src/components/AddressSearch.js
kidroca: I had the same question, I guess I forgot to post it 

I think since we're letting the "staging toggle" change the API for attachments we should follow the same here

src/components/TestToolMenu.js
kidroca: > Meaning we can't make the defaultProps use `ApiUtils.isUsingStagingApi()`? That seems fine.

Yes

kidroca: > **alternate idea (non blocking)**: Possibly could be solved by passing a default value for `user.shouldUseStagingServer` to Onyx during initialization and making it a required prop here. Then we could eliminate the `lodashGet` calls as well.

I think this would be a bit complicated

The default value `shouldUseStagingServer` is environment specific 
We need to first get the ENV and then call `Onyx.init` with default value for `user` - `{ shouldUseStagingServer: ENV == staging }` 

I can expose a `ApiUtils.getDefaultApi`, but I can't guarantee it would have the correct value at the time `Onyx.init` is called (similar issues as to why default props value was removed)

src/libs/ApiUtils.js
kidroca: Yes, though see this
https://github.com/Expensify/App/blob/bb0b7883210a9aba510adf682363004d536e29b4/src/libs/ApiUtils.js#L10-L16

we want the default value for staging to be `true`, but we can't simply do this
```js
let stagingServerToggleState = ENV_NAME === CONST.ENVIRONMENT.STAGING;
```

because the real ENV_NAME is resolved a few moments later

We can only do this:

```js
let stagingServerToggleState = false;
Onyx.connect({
    key: ONYXKEYS.USER,
    callback: (val) => {
        const defaultValue = ENV_NAME === CONST.ENVIRONMENT.STAGING;
        stagingServerToggleState = lodashGet(val, 'shouldUseStagingServer', defaultValue);
    },
});
```

It's probably fine, but ENV_NAME is obtained async and it might not be updated in the Onyx.connect callback 
IMO the safest is what we have right now
What do you think?

src/libs/Navigation/AppNavigator/AuthScreens.js
kidroca: There's seems to be one thing that keeps evading us, perhaps it was more clear when I renamed `URL_API_ROOT` to `PROD_API_ROOT`:

`authEndpoint` is always given the PROD address 
- staging native apps don't use `.env.staging` but `.env.production` so they pass `www.expensify` here
- web/desktop use `.env.staging` but they also pass `www.expensify` because that's what's specified in `.env.staging` 

So perhaps it doesn't matter, or we're always supposed to pass the PROD _(AKA main, primary, non-staging)_ url here

kidroca (commented):
Added a few detailed explanations, sorry if it's a bit too long

I can address most of the [suggestions](https://github.com/Expensify/App/pull/15178#pullrequestreview-1327165691) right now 
But for a few I'll need a response from @marcaaron

src/CONFIG.js
```diff
@@ -49,11 +49,17 @@ export default {
...
+
+        // The PRIMARY API is the API used by most environments, but staging, where we use STAGING (bellow)
+        // The "staging toggle" in settings toggles between PRIMARY and STAGING APIs
+        // We can think of PRIMARY as PROD, because on both STAGING and PROD it points to the production address
+        // On DEV this can be configured through ENV settings and can be a proxy or ngrok address (defaults to PROD)
+        // Usually you don't need to use this URL directly - prefer `ApiUtils.getApiRoot()`
+        PRIMARY_API_ROOT: expensifyURLRoot,
+        PRIMARY_SECURE_API_ROOT: secureURLRoot,
+        STAGING_API_ROOT: stagingExpensifyURL,
+        STAGING_SECURE_API_ROOT: stagingSecureExpensifyUrl,
```
kidroca, line 62: We have this thread, that's hidden due to the updated code comment: https://github.com/Expensify/App/pull/15178#discussion_r1126937268

kidroca:
Heads up - for some reason, requesting a new review from Marc, removed @mountiny and @Santhosh-Sellavel from the reviewers

mountiny:
Sorry havent got around to this one yet

src/components/TestToolMenu.js
marcaaron: Maybe I missed it. But I thought the `user.shouldUseStagingServer` value is straight from Onyx and represents the toggle state? My suggestion is to just enforce that it be false by default at the Onyx level.

src/libs/ApiUtils.js
marcaaron: ya in my mind it's a style guide thing and a code convention we use everywhere (not really something I want to debate in this PR so would recommend we just fix it).

https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md#boolean-variables-and-props

tl;dr boolean variables have the "should" convention so a function is usually "can" instead of "should"

src/libs/ApiUtils.js
marcaaron: >ENV_NAME is obtained async and it might not be updated in the Onyx.connect callback

Does it lead to any clear problem though? We can test to see if it's fine or not?

src/CONFIG.js
marcaaron: It feels like we are fundamentally worried that people will use this when they shouldn't - that sounds great. I'm mostly trying to avoid the idea that there are different levels of API or implying importance. 

I get the feedback about environment though.. so any of these alternatives could work equally well...

`STATIC_API_ROOT`
`FIXED_API_ROOT`
`DEFAULT_API_ROOT`
`NON_DYNAMIC_API_ROOT`

?

I think the main thing here is that the URL we are calling PRIMARY - is the one that doesn't change.

src/components/AddressSearch.js
marcaaron: Agree

src/CONFIG.js
kidroca: I'll update to `DEFAULT_API_ROOT` 

One confusing part that remains is `DEFAULT_API_ROOT` on staging would be `www.expensify`, so by default staging should not use `DEFAULT_API_ROOT` but `STAGING_API_ROOT` 
(The new code comment explains why that is so) 

What I'd like to be able to change in the future is to make the configuration in such a way that 
DEFAULT_API, always returns the default for the current environment (so it would return `staging.expensify` on staging)

src/libs/Navigation/AppNavigator/AuthScreens.js
kidroca: > > authEndpoint is always given the PROD address
> 
> yeah I think it's wrong. It should use the default API as we need to be able to test the Pusher auth endpoint locally on dev at least.

This is so. It uses either PROD or what you have in your .env configuration

What I mostly wanted to point out is that right now staging uses the prod URL 
#### Shouldn't the api here be `staging.exepensify.com`  when we're on staging? 
image

kidroca: I can create a `ApiUtils.getDefaultApi()` that always returns the default endpoint for the given environment and update the code here

So that `authEndpoint` gets `staging.expensify` here (on staging env)

src/CONFIG.js
```diff
@@ -49,11 +49,17 @@ export default {
     EXPENSIFY: {
         // Note: This will be EXACTLY what is set for EXPENSIFY_URL whether the proxy is enabled or not.
         EXPENSIFY_URL: expensifyURL,
-        SECURE_EXPENSIFY_URL: secureURLRoot,
         NEW_EXPENSIFY_URL: newExpensifyURL,
-        URL_API_ROOT: expensifyURLRoot,
-        STAGING_EXPENSIFY_URL: stagingExpensifyURL,
-        STAGING_SECURE_EXPENSIFY_URL: stagingSecureExpensifyUrl,
+
+        // The DEFAULT API is the API used by most environments, but staging, where we use STAGING (bellow)
```
mountiny, line 54:
```suggestion
        // The DEFAULT API is the API used by most environments, except staging, where we use STAGING (below)
```

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+function canUseStagingServer() {
+    return ENV_NAME === CONST.ENVIRONMENT.PRODUCTION
+        ? false
+        : stagingServerToggleState;
+}
+
+/**
+ * Get the currently used API endpoint
+ * (Some environments allow for dynamically switching the API)
```
mountiny, line 39:
```suggestion
 * (Non-production environments allow for dynamically switching the API)
```

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,85 @@
...
+    },
+});
+
+/**
+ * Helper method used to decide which API endpoint to call
+ * Non PROD environments allow API switching via the {@link stagingServerToggleState}
+ *
+ * @returns {Boolean}
+ */
+function canUseStagingServer() {
```
mountiny, line 31: Given that both dev and staging **_can_** use staging server, I think we should name this method shouldUseStagingServer 
```suggestion
function shouldUseStagingServer() {
```

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Get the command url for the given request
+ *
+ * @param {Object} request
+ * @param {String} request.command - the name of the API command
+ * @param {Boolean} [request.shouldUseSecure]
+ * @returns {String}
+ */
+function getCommandUrl(request) {
```
mountiny, line 67: I believe based on the naming guidelines we should write such shortcuts uppercase
```suggestion
function getCommandURL(request) {
```

src/CONFIG.js
```diff
@@ -49,11 +49,17 @@ export default {
...
+        // The "staging toggle" in settings toggles between DEFAULT and STAGING APIs
+        // We can think of DEFAULT as PROD, because on both STAGING and PROD it points to the production address
+        // On DEV this can be configured through ENV settings and can be a proxy or ngrok address (defaults to PROD)
+        // Usually you don't need to use this URL directly - prefer `ApiUtils.getApiRoot()`
+        DEFAULT_API_ROOT: expensifyURLRoot,
+        DEFAULT_SECURE_API_ROOT: secureURLRoot,
+        STAGING_API_ROOT: stagingExpensifyURL,
+        STAGING_SECURE_API_ROOT: stagingSecureExpensifyUrl,
```
kidroca, line 62: We have this thread, that's hidden due to the updated code comment: https://github.com/Expensify/App/pull/15178#discussion_r1126937268

src/libs/ApiUtils.js
kidroca: It was initially called `shouldUseStagingServer` but an update was requested: https://github.com/Expensify/App/pull/15178#discussion_r1128487808

src/libs/Navigation/AppNavigator/AuthScreens.js
```diff
@@ -98,7 +98,7 @@ class AuthScreens extends React.Component {
         Pusher.init({
             appKey: CONFIG.PUSHER.APP_KEY,
             cluster: CONFIG.PUSHER.CLUSTER,
-            authEndpoint: `${CONFIG.EXPENSIFY.URL_API_ROOT}api?command=AuthenticatePusher`,
+            authEndpoint: `${CONFIG.EXPENSIFY.DEFAULT_API_ROOT}api?command=AuthenticatePusher`,
```
kidroca, line 101: A "hidden" thread about this: https://github.com/Expensify/App/pull/15178#discussion_r1126939927

TL;DR 

staging is using DEFAULT_API_ROOT, which on staging is the prod API - www.expensify.com and not the staging api - staging.expensify.com

src/CONFIG.js
```diff
@@ -49,11 +49,17 @@ export default {
...
+
+        // The DEFAULT API is the API used by most environments, except staging, where we use STAGING (below)
+        // The "staging toggle" in settings toggles between DEFAULT and STAGING APIs
+        // We can think of DEFAULT as PROD, because on both STAGING and PROD it points to the production address
+        // On DEV this can be configured through ENV settings and can be a proxy or ngrok address (defaults to PROD)
+        // Usually you don't need to use this URL directly - prefer `ApiUtils.getApiRoot()`
+        DEFAULT_API_ROOT: expensifyURLRoot,
+        DEFAULT_SECURE_API_ROOT: secureURLRoot,
+        STAGING_API_ROOT: stagingExpensifyURL,
+        STAGING_SECURE_API_ROOT: stagingSecureExpensifyUrl,
```
kidroca, line 62: We have this thread, that's hidden due to updates: https://github.com/Expensify/App/pull/15178#discussion_r1126937268

src/components/TestToolMenu.js
marcaaron: I think it would be fine to do a follow-up. I must admit, I find some of the changes a bit confusing. I don't mean to criticize, as I understand the goal is to make things better. However, I want to make sure that the added complexity is worth it. When you explained it on Slack, it seemed simpler to me.

src/libs/Navigation/AppNavigator/AuthScreens.js
marcaaron: I think we are both agreeing we don't need a change here. The default URL on staging is always production. 👍 It hits the production API by default. 👍 

This endpoint should be whatever is defined in the environment (not always production). Locally we can change it to be whatever we want.There's no need to be able to switch it on prod or staging versions of the app.

src/libs/Navigation/AppNavigator/AuthScreens.js
marcaaron: I think we can just leave this so gonna resolve this one.

mountiny:
@kidroca I think we have merged code in these files recently https://github.com/Expensify/App/pull/14944 there is probably not lots of overlap, but can you merge main and check if no changes are required

src/components/TestToolMenu.js
kidroca: There are 3 options here 
- keep this change, because it's technically correct - when `user` prop is not passed the `shouldUseStagingServer` value should not default to `false`, but `undefined` so that `ApiUtils.isUsingStagingApi()` can be used as fallback 
- discard the change - currently we always have a `user` object passed when this component renders, the object lacks a `shouldUseStagingServer` so it's same as if it's `undefined` (when we haven't touched the toggle, like after fresh login)
- remove `defaultProps` - a `user` is always passed. Avoids the confusion around `defaultProps` with `undefined` value

kidroca:
> @kidroca I think we have merged code in these files recently #14944 there is probably not lots of overlap, but can you merge main and check if no changes are required

Updated with latest `main` (69af7413)
There were no overlapping changes

src/CONFIG.js
```diff
@@ -49,11 +49,17 @@ export default {
...
         EXPENSIFY_URL: expensifyURL,
-        SECURE_EXPENSIFY_URL: secureURLRoot,
         NEW_EXPENSIFY_URL: newExpensifyURL,
-        URL_API_ROOT: expensifyURLRoot,
-        STAGING_EXPENSIFY_URL: stagingExpensifyURL,
-        STAGING_SECURE_EXPENSIFY_URL: stagingSecureExpensifyUrl,
+
+        // The DEFAULT API is the API used by most environments, except staging, where we use STAGING (below)
+        // The "staging toggle" in settings toggles between DEFAULT and STAGING APIs
+        // We can think of DEFAULT as PROD, because on both STAGING and PROD it points to the production address
```
marcaaron, line 56:
```suggestion
On both STAGING and PROD it points to the production address
```

I think this is enough. Saying "we can think of DEFAULT as PROD" is unclear advice and we immediately contradict it on the next line by saying it could be other things. It's not necessarily pointing to the production API. It can be several things.

src/components/TestToolMenu.js
marcaaron: Let's start by leaving a comment explaining why `shouldUseStagingServer` needs to be `undefined`. If after reading that comment everyone decides it is reasonable then we can skip discussing whether it feels too complex.

src/libs/ApiUtils.js
kidroca: > > ENV_NAME is obtained async and it might not be updated in the Onyx.connect callback
> 
> Does it lead to any clear problem though?

A risk STAGING might use the PROD API

I think it's a known risk when we [depend on 2 async callbacks where order matters](https://github.com/Expensify/App/blob/698ef4a97ee93a5db4014aca829a7fde4cca7cc7/src/libs/ApiUtils.js#L10-L35) 
- Onyx callback
- Env callack 

"Onyx callback" uses env value and depends that "Env callback" has already ran

The logic can only work correctly if "Onyx callback" runs after "Env Callback", but nothing ensures such order

Is it OK to declare like:
```js
// To avoid rebuilding native apps, native apps use production config for both staging and prod
// We use the async environment check because it works on all platforms
let ENV_NAME = CONST.ENVIRONMENT.PRODUCTION;
let stagingServerToggleState = false;

Environment.getEnvironment()
    .then((envName) => {
        ENV_NAME = envName;

        Onyx.connect({
          key: ONYXKEYS.USER,
          callback: (val) => {
              const defaultValue = ENV_NAME === CONST.ENVIRONMENT.STAGING;
              stagingServerToggleState = lodashGet(val, 'shouldUseStagingServer', defaultValue);
          },
      });
    });
```

src/components/TestToolMenu.js
kidroca: Done

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,86 @@
...
+
+        // We connect here, so we have the updated ENV_NAME when Onyx callback runs
+        Onyx.connect({
+            key: ONYXKEYS.USER,
+            callback: (val) => {
+                const defaultToggleState = ENV_NAME === CONST.ENVIRONMENT.STAGING;
+                stagingServerToggleState = lodashGet(val, 'shouldUseStagingServer', defaultToggleState);
+            },
+        });
+    });
```
kidroca, line 24: Is it OK to run `Onyx.connect` in a callback instead of a top level call?
I've browsed existing code and found some similar usages

src/libs/ApiUtils.js
kidroca: I've updated the code to subscribe to Onyx after `getEnvironment` resolves (like above sample) to avoid the risk 
Let's tie this thread and continue here if we need to: https://github.com/Expensify/App/pull/15178#discussion_r1132173422

src/libs/ApiUtils.js
mountiny: I think if we are using it successfully in other places it should be fine.

src/components/TestToolMenu.js
mountiny: The comment makes sense to me ✅

mountiny (dismissed):
This is looking good to me, tested locally with dev environment and local VM and seems to work fine as well as the attachments 
image

But I have not tested all the other environments.

kidroca:
@mountiny 
(Testing iOS) I just found a place where the rename of `URL_API_ROOT` wasn't applied 
I'll push an update shortly

s77rt (changes requested):

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,86 @@
...
+function getCommandURL(request) {
+    return `${getApiRoot(request)}api?command=${request.command}`;
+}
+
+/**
+ * Check if we're currently using the staging API root
+ *
+ * @returns {Boolean}
+ */
+function isUsingStagingApi() {
```
s77rt, line 77: Isn't `isUsingStagingApi` the same as `canUseStagingServer`? It seems redundant.

s77rt (changes requested):

s77rt:
@kidroca Is this test "Verify absolute urls are resolved from API ROOT" really required? The attachments urls will always be in a different (and correct) format due to the use of `tryResolveUrlFromApiRoot`.

s77rt:
1. Offline tests fail; going Offline/Slow Connection will always load attachmnets from url and never from cache, this is also the current behaviour on staging. So this can be ignoreed?
2. Using staging switch will cause CORS (https://github.com/Expensify/App/pull/15517 this PR has been revered). I guess this also have nothing to do with this PR?

Besides that, tests well!, checklist in the way...

s77rt:
## Reviewer Checklist

- [X] I have verified the author checklist is complete (all boxes are checked off).
- [X] I verified the correct issue is linked in the `### Fixed Issues` section above
- [X] I verified testing steps are clear and they cover the changes made in this PR
    - [X] I verified the steps for local testing are in the `Tests` section
    - [X] I verified the steps for Staging and/or Production testing are in the `QA steps` section
    - [X] I verified the steps cover any possible failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
- [X] I checked that screenshots or videos are included for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [X] I verified tests pass on **all platforms** & I tested again on:
    - [X] Android / native
    - [X] Android / Chrome
    - [X] iOS / native
    - [X] iOS / Safari
    - [X] MacOS / Chrome / Safari
    - [X] MacOS / Desktop
- [X] If there are any errors in the console that are unrelated to this PR, I either fixed them (preferred) or linked to where I reported them in Slack
- [X] I verified proper code patterns were followed (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [X] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`).
    - [X] I verified any copy / text that was added to the app is correct English and approved by marketing by adding the `Waiting for Copy` label for a copy review on the original GH to get the correct copy.
- [X] I verified that this PR follows the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [X] I verified other components that can be impacted by these changes have been tested, and I retested again (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` have been tested & I retested again)
    - [X] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/styles/StyleUtils.js) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(themeColors.componentBG`)
- [X] I have checked off every checkbox in the PR reviewer checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/16493223/224480524-c2fe6c80-2c63-43f4-8060-819ee6438107.mp4

</details>

<details>

mweb-chrome

</details>

<details>

mweb-safari

</details>

<details>
<summary>Desktop</summary>


https://user-images.githubusercontent.com/16493223/224481407-f1b3091e-0da3-410e-8ef9-35a99636c586.mp4


</details>

<details>
<summary>iOS</summary>

ios

</details>

<details>
<summary>Android</summary>

android

</details>

src/libs/ApiUtils.js
kidroca: Yeah, this was mainly created for external usages like the `TestToolMenu.js` where an export like `isUsing` - it's a better name compared to `shouldUse` or `canUse` (those are more internal as they disclose how stuff works)

Instead of 

```js
return getApiRoot() === CONFIG.EXPENSIFY.STAGING_API_ROOT;
```

we can 

```js
return canUseStagingServer();
```

But the function would just look weird

```js
function isUsingStagingApi() {
  return canUseStagingServer();
}
```

I don't want to explain why `isUsingStagingApi` `==` `canUseStagingServer`, because `is` is not always `==` `can` 
So I prefer to put a check that is inline with what the function does (and is still a one liner)

If you it's a problem I can remove the function altogether it's all about `is` vs `can` stuff and how it is exported for external usage

kidroca:
> @kidroca Is this test "Verify absolute urls are resolved from API ROOT" really required? The attachments urls will always be in a different (and correct) format due to the use of `tryResolveUrlFromApiRoot`.

I'm not sure / I don't know - `tryResolveUrlFromApiRoot` was from my previous PR and this behavior was tested, but then we realized it didn't respect the "staging toggle" 

This PR if a followup to make`tryResolveUrlFromApiRoot` respect the "staging toggle" and I added the same steps 

> 1. [Web Only] Offline tests fail; going Offline/Slow Connection will always load attachmnets from url and never from cache, this is also the current behaviour on staging. So this can be ignoreed?

I'm not sure how did you test that
I do this 
1. load a chat with attachments
2. see thumbnails load (thumbnails get cached)
3. open one of them (it gets cached)
4. go to a different chat
5. go offline
6. go to chat with attachments again
7. see thumbnails load
8. open one of them - it loads 
9. open another one - it doesn't load (full image fails to load as it's not cached)

https://user-images.githubusercontent.com/12156624/224540391-3f26f35c-17ab-4035-883e-eacb39760277.mov


> 2\. [Web Only] Using staging switch will cause CORS ([[Staging API] To Use Staging Server or not to use staging server #15517](https://github.com/Expensify/App/pull/15517) this PR has been revered). I guess this also have nothing to do with this PR?

That's correct, the followup to fix this is waiting on the current PR

src/libs/ApiUtils.js
s77rt: That's clear, thanks @kidroca but I think the name `canUseStagingServer` is misleading, it should be reset to `shouldUseStagingServer` as that's what the function really do.

Reading `canUseStagingServer` I expected a simple `return ENV_NAME !== CONST.ENVIRONMENT.PRODUCTION` which is not the case.

s77rt:
> I'm not sure how did you test that

I think it's related to how the browser handle requests while offline, check the video for Web from the checklist above, I went offline differently than how you did. I think that's outside the scope of this PR since same behaviour is on staging and it may be actually a browser thing so nothing to worry about here. Just to be sure can you do the steps I have and confirm the same results

kidroca:
> I think it's related to how the browser handle requests while offline, check the video for Web from the checklist above, I went offline differently than how you did.

You have dev tools open and cache disabled - that's why cache won't be used
You can untoggle the checkmark here, and see a request won't be made again
Screenshot 2023-03-13 at 14 35 54

src/libs/ApiUtils.js
kidroca: I can rename `canUseStagingServer` to `isUsingStagingApi` and export it, so we can delete the function here, is that alright? 

The logic in `getRootApi` becomes 
- if: `isUsingStagingApi` -> use staging api
- otherwise: use default api

s77rt:
@kidroca Yeah, sorry my bad, I tend to disable cache for testing. So everything is actually as expected, as can be seen on the Desktop video.

src/libs/ApiUtils.js
s77rt: I think it would be better to do it this was:
1. Remove the function `canUseStagingServer` given it's unlikely to be used anywhere besides `getApiRoot`
2. On `getApiRoot` define `const shouldUseStagingServer = ENV_NAME === CONST.ENVIRONMENT.PRODUCTION ? false : stagingServerToggleState;`

___

Question: Why `isUsingStagingApi` only checks if the env is staging and not secure-staging?

kidroca: Regarding 1 and 2 - I like to keep conditions simple and easy to understand like `if (shouldUseStaging()` vs the expression you suggest 
Though I can rename `let stagingServerToggleState` and add all the logic in the Onyx callback so we just use value

> Question: Why `isUsingStagingApi` only checks if the env is staging and not secure-staging?

Because it's not possible to use prod api but secure staging api - you always use the APIs together - staging with secure staging

s77rt: But if you are using secure-staging the `getApiRoot` will return `CONFIG.EXPENSIFY.STAGING_SECURE_API_ROOT` and `isUsingStagingApi()` will return false.

kidroca: > But if you are using secure-staging the `getApiRoot` will return `CONFIG.EXPENSIFY.STAGING_SECURE_API_ROOT` and `isUsingStagingApi()` will return false.

Sorry, this doesn't work this way
Whether is a secure request or not you are either using the staging api or the default api 
Depending on the request you'll get either the secure path or the non secure path - but in both cases when the toggle is toggled to staging, it's a staging api url

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,80 @@
...
+                if (ENV_NAME === CONST.ENVIRONMENT.PRODUCTION) {
+                    shouldUseStagingServer = false;
+                    return;
+                }
+
+                const defaultToggleState = ENV_NAME === CONST.ENVIRONMENT.STAGING;
+                shouldUseStagingServer = lodashGet(val, 'shouldUseStagingServer', defaultToggleState);
+            },
+        });
+    });
```
kidroca, line 30: Updated the code and moved all the `shouldUseStaging` logic here

A follow-up PR can move this logic as a default Onyx value for the key `user. shouldUseStagingServer` so we don't have any logic here at all

src/libs/ApiUtils.js
s77rt: I think you didn't understand my question, let me rephrase, If my env is staging and I'm using the secure path, I'd be using the staging api, right?

If the answer to that question is yes, then why `isUsingStagingApi()` says otherwise (returns false)

src/libs/ApiUtils.js
```diff
@@ -0,0 +1,80 @@
...
+let shouldUseStagingServer = false;
+Environment.getEnvironment()
+    .then((envName) => {
+        ENV_NAME = envName;
+
+        // We connect here, so we have the updated ENV_NAME when Onyx callback runs
+        Onyx.connect({
+            key: ONYXKEYS.USER,
+            callback: (val) => {
+                // Toggling between APIs is not allowed on staging
```
s77rt, line 20:
```suggestion
                // Toggling between APIs is not allowed on production
```

s77rt (changes requested):

src/libs/ApiUtils.js
kidroca: I don't think we have to discuss this anymore, the logic was changed, is the new logic ok?

s77rt: Oh sorry you fixed that on the last commit too, I didn't notice since those comments covered most of the page bottom and I though it's the end of the changes :sweat_smile:. Looking good now. Thanks :+1:

s77rt (dismissed):
LGTM! :rocket:

src/components/TestToolMenu.js
```diff
@@ -45,8 +45,10 @@ const TestToolMenu = props => (
         This enables QA and internal testers to take advantage of sandbox environments for 3rd party services like Plaid and Onfido. */}
```
mountiny, line 45: The comment above here is not correct @kidroca, can you remove the "secure" part from the line above as this is changing not only secure api but any api. Thanks!

mountiny (commented):
@kidroca one comment, thank I will merge this, thanks!

mountiny (approved):
Lets do this

s77rt:
@mountiny Do we track payment for this in the linked issue or in another separated issue?

Issue #15506, opened 2023-02-27
[HOLD for payment 2023-03-23] [$1000] Inconsistency while selecting dropdown in mWeb and Android


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Open the app in Android
2. Navigate to Settings > Workspaces > Select any workspace
3. Reimburse expenses
4. Click on Unit Input
5. Do the exact same steps for mWeb
6. Notice that the dropdown input's border gets green when focused in mWeb but in Android this only happens when we click on the input

## Expected Result:
Behavior should be consistent.

## Actual Result:
The behavior of the field isn't consistent

## Workaround:
unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [x] Android / Chrome
- [ ] iOS / native
- [ ] iOS / Safari
- [ ] MacOS / Chrome / Safari
- [ ] MacOS / Desktop

**Version Number:** 1.2.76-7
**Reproducible in staging?:** y
**Reproducible in production?:** y
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 

https://user-images.githubusercontent.com/43996225/221579483-aab7a1f8-0181-42b6-bf90-6465848868e0.mp4

https://user-images.githubusercontent.com/43996225/221579538-8a836cfc-a657-46ec-9789-d6896fb08e91.mp4

**Expensify/Expensify Issue URL:**
**Issue reported by:** @daraksha-dk 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1677253047142869

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~010c6b32476cf961bc</li>
        <li>Upwork Job ID: 1631088899411161088</li>
        <li>Last Price Increase: 2023-03-02</li>
    </ul>
</details>

MelvinBot:
[...]

PR #15673, opened 2023-03-06
Add additional events for controlling picker indicator on web

redstar504:
### Details

I had to work through this a bit more to make the picker indicator behave consistently all platforms. As proposed, I added two new event handlers specific to the web version of the app that update the `isOpen` state, but they needed to be adapted a bit from what I suggested.

The original proposal worked to solve the reported issue, but it was causing the picker to remain highlighted on desktop web after selecting an item. To solve this, I switched from using `onClick` to `onMouseDown`. This solved it simply because of the order the event handlers are called in.

In addition, I included a new `onChange` event handler whose presence overrides the `onValueChange` handler for web. This was necessary because my original solution of setting `isOpen` to `false` in the `onValueChange` event was resulting in the  indicator changing back to the default setting whenever a new item was scrolled to in the iOS picker wheel.

### Fixed Issues

$ https://github.com/Expensify/App/issues/15506
PROPOSAL: https://github.com/Expensify/App/issues/15506#issuecomment-1451139715

### Tests

1. Login to the app and navigate to any page containing a Picker component.
2. In this issue, we used the `Settings > Workspace > Reimburse Expenses` picker for selecting mileage units.
3. Open the picker and confirm that the green indicator beneath the picker is bright green while the picker is open.
4. Select an item.  Confirm that after selecting an item the indicator returns to it's default dark green color.

### QA Steps

1. Login to the staging environment and navigate to any page containing a Picker component.
2. In this issue, we used the `Settings > Workspace > Reimburse Expenses` picker for selecting mileage units.
3. Open the picker and confirm that the green indicator beneath the picker is bright green while the picker is open.
4. Select an item.  Confirm that after selecting an item the indicator returns to it's default dark green color.

### PR Author Checklist

    - [X] Android / native
    - [X] Android / Chrome
    - [X] iOS / native
    - [X] iOS / Safari
    - [X] MacOS / Chrome / Safari
    - [X] MacOS / Desktop
    - [X] I verified any copy / text that was added to the app is correct English and approved by marketing by adding the `Waiting for Copy` label for a copy review on the original GH to get the correct copy.
    - [X] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/styles/StyleUtils.js) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(themeColors.componentBG`)
- [X] I have checked off every checkbox in the PR author checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>
<summary>Web</summary>

#### Safari

https://user-images.githubusercontent.com/1311325/222996647-5df944a0-fcb7-4a4a-8d27-025e10f745e5.mp4

#### Chrome

Having an issue with OBS not displaying `<select>` elements during a windowed screen recording (seems like `select` elements are considered a separate window), but still demonstrates the functionality:

https://user-images.githubusercontent.com/1311325/223631008-f2da367a-34e0-45b7-b7b3-fcf28453efe6.mp4

</details>

<details>

https://user-images.githubusercontent.com/1311325/223631057-a5b5df0b-87d9-4d1a-af0e-742be584be31.mp4

</details>

<details>

https://user-images.githubusercontent.com/1311325/222996270-192ab927-379d-46f2-bdf0-c379928f2032.mp4

</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/1311325/223631123-6fec3bae-890c-4142-96fe-8ad0eb67c042.mp4

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/1311325/223631143-ce7c083a-e397-49d3-a465-4a5b17b71c3c.mp4

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/1311325/223631164-b97b1458-d075-463b-be23-01dc04c2722a.mp4

</details>


src/components/Picker.js
```diff
@@ -193,6 +193,21 @@ class Picker extends PureComponent {
                                 this.setState({isOpen: false});
                                 this.props.onBlur();
                             },
+
+                            // The following 2 handlers are specific to web (onChange overrides onValueChange on web)
+                            // They are used to make the picker indicator behave the same way as it does on native
```
marcaaron, line 198: >To solve this, I switched from using onClick to onMouseDown. This solved it simply because of the order the event handlers are called in.

Can we also briefly mention something about this in this code comment? Feels possible someone might make the mistake of switching back to an `onClick` behavior without this.

marcaaron: Also, how does `onChange` override `onValueChange` on web?

src/components/Picker.js
```diff
@@ -193,6 +193,21 @@ class Picker extends PureComponent {
...
+                                this.setState({isOpen: true});
+                            },
+                            onChange: (e) => {
+                                if (e.target.selectedIndex === undefined) {
+                                    return;
+                                }
+                                const index = e.target.selectedIndex;
+                                const value = e.target.options[index].value;
+                                this.onInputChange(value, index);
+                                this.setState({isOpen: false});
```
marcaaron, line 209: It feels like we are sneaking in some platform dependent code here, but it's not clear which is the web or native implementation. Normally, [how we deal with that](https://github.com/expensify/app#platform-specific-file-extensions) is create a "base" component and then an `index.js` and `index.native.js` components. In a file like `Picker.js` we'd then just import the Picker but the platform dependent changes would be contained in the respective versions of the picker. Would this approach work here?

marcaaron:
@thesahindia I don't feel super strongly about solving that case in this ticket. We didn't spot it until now and it's not clear what the solution would be - so I say we continue here and then file it as a separate bug. Unless @redstar504 has some idea on how to tackle it and is willing to look into it.

redstar504:
@marcaaron @thesahindia Thanks for the input guys. I have been exploring ways to split up the platform-specific code as requested. It appears that our goal is to share the Picker component across all platforms while providing a means to specify any platform-specific event handlers through their respective index files.

Would it make sense to refactor `Picker.js` into `Picker/BasePicker.js`, containing all of the platform independent handlers, and then import `BasePicker` into the platform specific index files? We could pass any platform specific event handlers to `BasePicker` as props to enable customization for each platform.

redstar504:
My latest commit includes platform specific event handling for the Picker component. The implementation involves wrapping the refactored BasePicker with platform specific entrypoint files, and wrapping the entrypoints with a higher order component that supplies default events for both platforms. Web specific events are encapsulated in the index.js file. Native events are supplied as defaults through the HOC since they supplement the web-based events.

Testing on all platforms was successful, and I have included new recordings in my checklist.

@marcaaron @thesahindia

src/components/Picker.js
redstar504: I added some comments regarding this in the `index.js` file of my latest commit.

Let me know if you have any more questions.

src/components/withPickerEvents.js
```diff
@@ -0,0 +1,120 @@
+import React, {Component} from 'react';
+import {pickerPropTypes, pickerDefaultProps} from './Picker/pickerPropTypes';
+
+export default function (WrappedComponent) {
+    class WithSharedEvents extends Component {
```
marcaaron, line 5: I haven't delved deeply into the reasons for needing it, but is an HOC necessary? Feels like the "WrappedComponent" is always going to be the `Picker`.

Additionally, I'm having trouble following some of the conventions used here e.g. the `callbackConsumer()` (stuck out to me as unusual for our codebase and my instincts are telling me it could confuse people).

I'd have to do a deeper dive to come up with more concrete suggestions, but just wanted to toss that observation out there that maybe we could simplify this.

redstar504: The HOC wraps the Picker in both entrypoints (index.js and index.native.js), so that we can share all of the code contained in it between them. This also gives us a way to manipulate the events passed down to BasePicker using  an entrypoint file. I originally had placed the HOC in the Picker directory because it would only ever be used for the Picker, but I noticed all of the others were contained in the base `src` directory so I moved it there.

An HOC would seem like overkill but I am kind of stuck between a dead simple solution (the one I originally provided), and an over-engineered solution.  I have been having trouble finding something in the middle ground that won't introduce a lot of code duplication.  Another option might be class inheritance.

We basically need a way to share all of the event handlers between both entrypoints but give the web entry point a way to extend/override them before passing them down to BasePicker.  If you have any alternative ideas for that please let me know. If we stick with the HOC I can definitely rename `callbackConsumer` back to `onInputChange`.

marcaaron: Ok I see the issue now - sorry for the distraction here. I think where this is at is a bit hard to reason about for not much gain.

Here's what I'm thinking as an alternative...

1. Start by reverting what you have here back to https://github.com/Expensify/App/pull/15673/commits/15fb9c7db99ac31f24b6b78926115a1a4cca5e8b
1. Create some prop like `shouldUseAdditionalPickerEvents` on the Picker
1. Override the methods that way inside the Picker

It might seem vain, but we can get around the cross platform file extension convention this way and it will allow web/native to switch on/off the functionality we need without specifically referencing the platform anywhere.

redstar504: @marcaaron Thank you for the advice.  I just need a bit more clarification:

> It might seem vain, but we can get around the cross platform file extension convention this way and it will allow web/native to switch on/off the functionality we need without specifically referencing the platform anywhere.

Are you suggesting not having platform-specific entry points in this case?  I am just curious what is going to toggle `shouldUseAdditionalPickerEvents` if we don't, and if we didn't, where do we define the web-specific events?

redstar504: @marcaaron Ok I will give that a shot.  Since it's just a conditional in the entry points, is this implying that we keep the web events defined in the Picker itself?

marcaaron: >implying that we keep the web events defined in the Picker itself

That was my meaning yea.

But I thought of something that is slightly better and might result in fewer changes while avoiding a reference to the web specific `e.target`...

- Create `getAdditionalPickerEvents()`
    - have native return `{}`
    - have web return something like

```js
// getAdditionalPickerEvents/index.js
function getAdditionalPickerEvents(onMouseDown, onChange) {
    return {
        onMouseDown,
        onChange: (e) = {
            if (e.target.selectedIndex === undefined) {
                return;
            }
            const index = e.target.selectedIndex;
            const value = e.target.options[index].value;
            onChange(value, index);
        },
    };
}
```

Then pass them like:
```js
pickerProps={{
    ...getAdditionalPickerEvents(this.setIsOpen, this.updateAndClose),
```
  
This feels slightly better to me as it manages the web event outside of the cross platform Picker code. Sorry for all the back and forth. Our style guide sometimes paints us into a corner 😄

redstar504: @marcaaron I like the looks of that.  I have been trying to find a way to pass down the events from the web entry, but also to be able to pass a function back that can enable/disable the highlight without having to reference them from inside the shared Picker.  This looks like it could work.

Also, no worries. I am happy to work towards something we can all be satisfied with.

redstar504:
That concept worked perfectly @marcaaron, nice job.

The way I implemented it is by passing the `additionalPickerEvents` containing the web events through props from the `index.js` entrypoint.  In the Picker, I declared `additionalPickerEvents` as a `defaultProp` with an empty function, so we don’t have to pass anything from the native entrypoint.

I refactored all of the `setState({isOpen: bool})` calls into setters like in your example.  I did not introduce an `updateAndClose()` function, as the native implementation does not handle both operations in a single event, so it would have only been used for web.

Again, checks out on all platforms.

src/components/Picker/Picker.js
```diff
@@ -151,6 +155,18 @@ class Picker extends PureComponent {
...
+        this.setState({
+            isOpen: true,
+        });
+    }
+
+    setIsClosed() {
+        this.setState({
+            isOpen: false,
+        });
+    }
```
marcaaron, line 168: Can we bind these in the constructor?

src/components/Picker/Picker.js
```diff
@@ -184,15 +200,22 @@ class Picker extends PureComponent {
                         Icon={() => this.props.icon(this.props.size)}
                         disabled={this.props.isDisabled}
                         fixAndroidTouchableBug
-                        onOpen={() => this.setState({isOpen: true})}
-                        onClose={() => this.setState({isOpen: false})}
+                        onOpen={() => this.setIsOpen()}
+                        onClose={() => this.setIsClosed()}
```
marcaaron, line 204:
```suggestion
                        onOpen={this.setIsOpen}
                        onClose={this.setIsClosed}
```

src/components/Picker/Picker.js
```diff
@@ -184,15 +200,22 @@ class Picker extends PureComponent {
...
                         textInputProps={{allowFontScaling: false}}
                         pickerProps={{
-                            onFocus: () => this.setState({isOpen: true}),
+                            onFocus: () => this.setIsOpen(),
```
marcaaron, line 207:
```suggestion
                            onFocus: this.setIsOpen,
```

src/components/Picker/Picker.js
```diff
@@ -184,15 +200,22 @@ class Picker extends PureComponent {
...
                             onBlur: () => {
-                                this.setState({isOpen: false});
+                                this.setIsClosed();
                                 this.props.onBlur();
                             },
+                            ...this.props.additionalPickerEvents(
+                                () => this.setIsOpen(),
```
marcaaron, line 213:
```suggestion
                                this.setIsOpen,
```

marcaaron (changes requested):
Looks great to me. Just a couple of places where I think we could use a bound function instead of an anonymous arrow function.

redstar504:
@marcaaron Requested changes pushed!

marcaaron (dismissed):
Didn't test but code LGTM so approving. All yours @thesahindia.

src/components/Picker/Picker.js
```diff
@@ -151,6 +157,18 @@ class Picker extends PureComponent {
...
+        this.setState({
+            isOpen: true,
+        });
+    }
+
+    setIsClosed() {
+        this.setState({
+            isOpen: false,
+        });
+    }
```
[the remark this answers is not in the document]
redstar504: You are right, I think we should clarify what that state variable actually controls.  Since it is only responsible for enabling and disabling the highlight, we can make it more understandable by using an `isHighlighted` state variable, and calling `enableHighlight` and `disableHighlight` to control it.

src/components/Picker/Picker.js
```diff
@@ -3,13 +3,13 @@ import React, {PureComponent} from 'react';
...
-import {ScrollContext} from './ScrollViewWithContext';
+import Icon from '../Icon';
+import * as Expensicons from '../Icon/Expensicons';
+import FormHelpMessage from '../FormHelpMessage';
+import Text from '../Text';
+import styles from '../../styles/styles';
+import themeColors from '../../styles/themes/default';
+import {ScrollContext} from '../ScrollViewWithContext';
 
 const propTypes = {
```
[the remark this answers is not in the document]
redstar504: Since no props are specifically referenced in the platform-specific wrappers, it might be unnecessary to perform validation there before passing them to the BasePicker. However, if this is an established pattern in the codebase, it would be a good idea to follow that with our changes as well.

redstar504:
Thanks for the review @thesahindia.  I have added your requested changes to the latest commit.

redstar504:
@thesahindia

The error is due to the prop validation we've added in the entrypoints.  Adding `propTypes` directly to the component created by `forwardRef` does not validate as expected. It seems that we would have to wrap the BasePicker, do the prop validation and then wrap it again using `forwardRef`.  This would be unnecessary, since all of the props are already being validated in the `BasePicker`. There is no need to validate them in the entrypoints as well.

For example, this works but does not provide any benefit:
```JavaScript
// index.js
import {defaultProps, propTypes} from './pickerPropTypes';

// eslint-disable-next-line react/jsx-props-no-spreading
const Picker = props => <BasePicker {...props} additionalPickerEvents={additionalPickerEvents} />;

Picker.propTypes = propTypes;
Picker.defaultProps = defaultProps;

// eslint-disable-next-line react/jsx-props-no-spreading
export default forwardRef((props, ref) => <Picker {...props} innerRef={ref} />);
```

Using shared `propTypes` looks to be common when there are multiple entrypoints that do not wrap an underlying component.  It's purpose would be to give them a common interface to conform to. 

I suggest that we rollback this change as it is not necessary for our scenario.

redstar504:
@thesahindia I reverted shared prop types due to the reasons described in my last comment.  Props are still validated as expected through the BasePicker component, and there are no console errors.  So we should be good.

thesahindia (approved):
~Looks good and tests well!~

~@marcaaron, I have filled the checklist. It's all yours!~

thesahindia (changes requested):
I take back what I said. It didn't test well. I was checking other pages to confirm that all the functionalities are working as expected. It is not working correctly. Not getting the error message for the picker. @redstar504, can you look into it?

I am on a slow internet so I couldn't upload the video. You can try reproducing it at "Add debit card page" or any other page where we are validating.

redstar504:
@thesahindia Great job catching that.  We needed to pass the `ref` all the way down to the base component by that name to continue forwarding it, and only then it's passed to the RNPickerSelect as `innerRef`.  It should be fixed now.

thesahindia (approved):
Thanks! It works well.

cc: @marcaaron

src/components/Picker/Picker.js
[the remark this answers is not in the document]
marcaaron: I think I don't feel too passionate about this one. Or at least I can't think of a situation where it would help because the props are just getting passed straight to the BasePicker. If there was some prop validation warning we would see it either way.

marcaaron (approved):
Nice job here overall and LGTM. Thanks @thesahindia for the comments and testing 🙇

PR #16323, opened 2023-03-20
Update Style Guide on Hooks

marcaaron:
### Details

- Updating the style guide to remove the "hooks" ban and remove references to class components more generally.
- Also including an eslint plugin for hooks
- And refactoring a small component to get us started

### Fixed Issues

$ https://github.com/Expensify/App/issues/16325

### Tests

1. Create a workspace
2. Navigate to the page for that workspace
3. Delete the workspace
4. Verify there are no issues
5. Go offline
6. Create a workspace.
7. Verify that the workspace is shown as pending.
8. Come back online.
9. Verify that the workspace is shown as non-pending.

### Offline tests
N/A

### QA Steps
Same as tests

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

</details>

<details>

</details>

<details>

</details>

<details>
<summary>Desktop</summary>

</details>

<details>
<summary>iOS</summary>

</details>

<details>
<summary>Android</summary>

</details>


contributingGuides/STYLE.md
```diff
@@ -413,63 +411,7 @@ const propTypes = {
 };
 ```
 
-* Avoid public methods on components and calling them via refs
```
marcaaron, line 416: This defensive comment is largely irrelevant if we are removing class components. The way to do this with hooks would be [`useImperativeHandle()`](https://react.dev/reference/react/useImperativeHandle#useimperativehandle), but it's kind of a rare edge case that we would reach for it at all. And generally much harder to abuse this without knowing that it's an option.

contributingGuides/STYLE.md
```diff
@@ -540,96 +482,48 @@ const propTypes = {
 }
 ```
 
-## Binding methods
```
marcaaron, line 543: This is class specific and totally unnecessary in function components.

contributingGuides/STYLE.md
```diff
@@ -684,125 +574,9 @@ When writing a stateless component you must ALWAYS add a `displayName` property
 
 ## Stateless components vs Pure Components vs Class based components vs Render Props - When to use what?
 
-*_1. Stateless components: Used when you don't need to maintain state or use lifecycle methods._*
-
-In many cases, we create components that do not need to have a state, lifecycle hooks or any internal variables. In other words just a simple component that takes props and renders something presentational. But often times we write them as class based components which come with a lot of cruft (for e.g., it has a state, lifecycle hooks and it is a javascript class which means that React creates instances of it) that is unnecessary in many cases.
```
marcaaron, line 689: This section is irrelevant because function components now can have "state" so they are not "stateless"

contributingGuides/STYLE.md
```diff
@@ -684,125 +574,9 @@ When writing a stateless component you must ALWAYS add a `displayName` property
...
- - Here is an [example](https://github.com/Expensify/JS-Libs/blob/e3b7ee4b111a3a370a1427ad904485df4e65a472/lib/components/StepProgressBar.js#L20) from our codebase of a stateless component.
-```js
-const StepProgressBar = ({steps, currentStep}) => {
-    const currentStepIndex = Math.max(0, _.findIndex(steps, step => step.id === currentStep));
-    return (
-        <div id="js_steps_progress" className="progress-wrapper">
-...
-
-```
-*_2. Pure components: Use to improve performance where a component does not need to be rendered too often._*
```
marcaaron, line 705: This whole section is a bit out of date though I suppose we can mention something about performance. `React.memo()` is basically the replacement for any `PureComponent` or `shouldComponentUpdate()`.

contributingGuides/STYLE.md
```diff
@@ -684,125 +574,9 @@ When writing a stateless component you must ALWAYS add a `displayName` property
...
-        return <Input ref={el => this.inputRef = el} />;
-    }
-}
-```
-
-*_4. Render Props: Use for Cross-Cutting Concerns and Code Reuse_*
-
-Let's say you have two separate components that render two different things, but you want them both to be based on the same state. You can pass a `render` property to a component or pass a function as a child to accomplish this. You can read more about how this works and when to use it in the [React Docs](https://reactjs.org/docs/render-props.html).
-
-## Do not use inheritance for other React components
```
marcaaron, line 755: There isn't any concern here since we won't have any class components. Hooks are designed to facilitate code reuse so while there may still be HOCs render props are going to be a less attractive pattern than hooks.

contributingGuides/STYLE.md
```diff
@@ -684,125 +574,9 @@ When writing a stateless component you must ALWAYS add a `displayName` property
...
-                    {this.state.whatever}
-                </Text>
-            </AComponent>
-        )
-    }
-    ...
-}
-```
-
-## isMounted is an Antipattern
```
marcaaron, line 785: The cleanup method of `useEffect()` should solve any issues with this section.

src/components/PlaidLink/index.native.js
```diff
@@ -1,10 +1,13 @@
...
+import {useEffect, useRef} from 'react';
 import {openLink, useDeepLinkRedirector, usePlaidEmitter} from 'react-native-plaid-link-sdk';
 import Log from '../../libs/Log';
 import CONST from '../../CONST';
 import {plaidLinkPropTypes, plaidLinkDefaultProps} from './plaidLinkPropTypes';
 
 const PlaidLink = (props) => {
+    // We are stashing initial props in a ref since we don't want a new token to trigger the link to open again
+    // and just want openLink() to be called once
+    const propsRef = useRef(props.token);
```
roryabraham, line 10: Did this come up with the new ESLint rule?

roryabraham (commented):
Overall nothing here looks concerning, the changes looks pretty good. It's exciting how much we can remove from our style guide. Seems consistent with our hope that switching to hooks will reduce "gotchas" for NewDot contributors.

src/pages/workspace/WorkspaceInitialPage.js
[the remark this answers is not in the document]
marcaaron: Yep pretty much

src/components/PlaidLink/index.js
```diff
@@ -5,13 +5,15 @@ import Log from '../../libs/Log';
...
 const PlaidLink = (props) => {
     const [isPlaidLoaded, setIsPlaidLoaded] = useState(false);
-    const onSuccess = useCallback((publicToken, metadata) => {
-        props.onSuccess({publicToken, metadata});
-    }, []);
+    const onSuccess = props.onSuccess;
+    const onError = props.onError;
+    const successCallback = useCallback((publicToken, metadata) => {
+        onSuccess({publicToken, metadata});
+    }, [onSuccess]);
```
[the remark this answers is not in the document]
marcaaron: Ah so, I was following this lint rule here:

>React Hook useCallback has a missing dependency: 'props'. Either include it or remove the dependency array. However, 'props' will change when *any* prop changes, so the preferred fix is to destructure the 'props' object outside of the useCallback call and refer to those specific props inside useCallback.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

What do you think?

src/components/PlaidLink/index.native.js
marcaaron: Yeah, this one is interesting. We could solve it a couple ways... 

Basically, there is a `props.token` which we would using only once when the component mounts. If we pass `[]` as a dependency array then the exhaustive deps rule will warn that there's a missing dependency. But if we put it in a ref then React knows to ignore it.

We don't want to put the token as a dependency since then it will be "watched" and we will re-trigger `openLink()` if the token changes... I'm not really sure if the token will change, but it would introduce a different behavior.

Another option would be to ignore the rule in this case.

src/components/PlaidLink/index.js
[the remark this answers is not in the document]
roryabraham: Yeah, that makes sense. It's a bit of a GOTCHA and makes me wonder if we should just start destructuring props as a more general rule? It seems to be the approach that React is promoting.

This is tangentially related, but according to [this tweet](https://twitter.com/hswolff/status/1133759319571345408) `defaultProps` will eventually be deprecated. Why? Because React wants you to do this instead:

```js
const MyComponent = ({
   propA,
   propB = 'propBDefaultValue',
}) => {
   // Component stuff in here
}
```

src/components/PlaidLink/index.native.js
roryabraham: Hmmmm, yeah I think I would opt to suppress the exhaustive deps rule with a comment explaining why. NAB though

src/components/PlaidLink/index.js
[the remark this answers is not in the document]
marcaaron: Yeahhh, I think probably our rule about de-structuring is a bit outdated - but maybe a separate concern for now, or do you think we should reopen that conversation first? At the end of the day - declaring a variable or using destructuring assignment achieve the same thing.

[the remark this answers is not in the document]
roryabraham: Chatted about this in person, but yeah I think I'm going to just resolve this for now. Agree it's not the most critical thing and we don't want to rock the boat _toooo_ much

src/components/PlaidLink/index.native.js
roryabraham: Maybe I just need to get more used to `useRef`. Going to resolve this one too

roryabraham (commented):
I noticed there are a couple items on the PR checklist:

> - [ ] For Class Components, any internal methods passed to components event handlers are bound to this properly so there are no scoping issues (i.e. for onClick={this.submit} the method this.submit should be bound to this in the constructor)
> - [ ] Any internal methods bound to this are necessary to be bound (i.e. avoid this.submit = this.submit.bind(this); if this.submit is never passed to a component event handler like onClick)

Should we remove these? Or not yet?

**Edit:** Maybe we shouldn't remove those until we don't have any more class components.

src/pages/workspace/WorkspaceInitialPage.js
```diff
@@ -41,230 +41,207 @@ const defaultProps = {
...
-        ];
-
-        return (
-            <ScreenWrapper includeSafeAreaPaddingBottom={false}>
-                {({safeAreaPaddingBottomStyle}) => (
-                    <FullPageNotFoundView
-                        shouldShow={_.isEmpty(this.props.policy)}
+    }, [props.reports, props.policy]);
+
+    const policy = props.policy;
```
roryabraham, line 74: Since you're creating a local variable for `policy`, let's move it up and use it on line 72 where we're currently using `props.policy`?

roryabraham:
Just noting that I found a bug but was able to reproduce it on production. Reported [here](https://expensify.slack.com/archives/C049HHMV9SM/p1680092907036909)

contributingGuides/STYLE.md
marcaaron: I think I would come back and update this later as part of some evolving best practices. My instinct is to say something like...

If you aren't going to memoize the function with `useCallback()` then make it a pure function and move it outside the component if you can. If you need to create a closure around the props or state or something else then either allow it to be redeclared or just `useCallback()`. Probably the community has some idea about what would be best.

roryabraham (approved):

danieldoglas (approved):

src/components/PlaidLink/index.native.js
```diff
@@ -15,18 +18,17 @@ const PlaidLink = (props) => {
...
-                props.onSuccess({publicToken, metadata});
+                propsRef.current.onSuccess({publicToken, metadata});
             },
             onExit: (exitError, metadata) => {
                 Log.info('[PlaidLink] Exit: ', false, {exitError, metadata});
-                props.onExit();
+                propsRef.current.onExit();
             },
         });
     }, []);
```
situchan, line 31: We don't need to use `useRef` here.
useEffect(() => {}, `[]`) - empty array dependency - this is equivalent to `componentDidMount`.
So `openLink` is called only once regardless of `props.token` change.

marcaaron: interesting. but does it cause any harm? if we use the `props.token` directly then the `exhaustive-deps` linter would complain about a missing dependency.

src/components/PlaidLink/index.native.js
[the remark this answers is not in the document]
marcaaron: Oh, yes, I see my mistake now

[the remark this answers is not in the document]
marcaaron: Thank you

src/components/PlaidLink/index.native.js
gedu: if it complains, we can disable with a comment instead of using ref: https://expensify.slack.com/archives/C01GTK53T8Q/p1680256754764159

Issue #10632, opened 2022-08-29
[$2000] When you long-press on a message in iOS mobile, the message moves out of the screen frame and is hidden


mvtglobally:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Navigate to any chat with multiple messages 
2. Long press on the last message you sent

## Expected Result:
When long pressing on the last message, we should adjust the scroll so that the message you just selected is visible. 

## Actual Result:
When you long-press on a message in iOS mobile, the message moves out of the screen frame and is hidden.

## Workaround:
None 

## Platform:

Where is this issue occurring?

- iOS

**Version Number:** 1.1.92-0
**Reproducible in staging?:** Y
**Reproducible in production?:** Y
**Email or phone of affected tester (no customers):** 
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/43995119/187123528-d0427883-8364-40be-899d-4ba63c33d98d.MP4

**Expensify/Expensify Issue URL:** 
**Issue reported by:** @roryabraham 
**Slack conversation:** https://expensify.slack.com/archives/C01GTK53T8Q/p1661612352759039

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

marcaaron:
>User should be able to see which messages is being selected

What does this mean exactly... ? Looking at the video and I'm not even sure what the bug is.

marcaaron:
>When you long-press on a message in iOS mobile, the message moves out of the screen frame, and it’s not obvious which message you’ve selected

Also don't understand what is being proposed here.

roryabraham:
The problem is that once the context menu pops up and the modal blur is applied, it's not clear which message you've long-pressed on. What I was imagining is that the message you selected would have a white background / be "immune" from the modal blur effect.

But I think a mockup would be very helpful here. @marcaaron suggested that we could use a "shimmer effect" as an alternative

michelle-thompson:
Hmm, what about when you hold down on a message the background highlights as grey like so?
[...]

PR #16356, opened 2023-03-21
When you long-press on a message in iOS mobile, the message moves out of the screen frame and is hidden (iOS only)

terrysahaidak:
### Important notice

> [!IMPORTANT]  
> This PR relies on `react-native-keyboard-controller` (`useKeyboardHandler`) hook and `KeyboardAvoidingView` component).

### How it works 🤔 

This PR is **iOS only** for now. Other platforms are not affected. Also, in this explanation, we are talking about iOS only since on Android, iOS web and Android Web keyboard handling works differently.

> [!NOTE]
> We're talking about default platform behavior. If we start to use `react-native-keyboard-controller` on Android - it'll disable all default keyboard handling by OS and will delegate the keyboard handling to us - so we will have identical behavior across both iOS/Android platforms.
>
> But it may introduce additional issues (because we need to enter edge-to-edge mode - that would be good to do on entire app level, but it may bring some issues with `StatusBar` paddings, so everything has to be verified very carefully). That's why I think it would be better to handle in separate PR.

When we long press on a message, we want to be able to see it. It's easy when we have the keyboard closed. But gets tricky when the keyboard is open.

But first of all let's focus on the case when the keyboard is closed. User long press on the message, we render the bottom sheet (action sheet) with some options. If the message is higher than the height of bottom sheet - the message is visible. When the message is the last one (first from the bottom), it can be fully or partially covered by the bottom sheet.

<details>
  <summary>Press to see videos</summary>
  <table>
    <tr>
      <th>Message above bottom sheet</th>
      <th>Message and bottom sheet overlap</th>
    </tr>
    <tr>
      <td><video src="https://github.com/Expensify/App/assets/22820318/9dc8be61-799d-470b-997f-ee5c5995fe19"></td>
      <td><video src="https://github.com/Expensify/App/assets/22820318/0d0570cd-a85f-4cf5-8a91-6a3d5ad72c58"></td>
    </tr>
  </table>
</details>

Also everything is tricky when we have keyboard open 😓 

<details>
  <summary>Press to see videos with keyboard</summary>
  <table>
    <tr>
      <th>emoji picker instead of the keyboard (the messages should not jump)</th>
      <th>showing a message which is covered by the keyboard and the bottom sheet</th>
    </tr>
    <tr>
      <td><video src="https://github.com/Expensify/App/assets/22820318/cbe3b610-fa85-4071-a3dd-bbc1a9d56e05"></td>
      <td><video src="https://github.com/Expensify/App/assets/22820318/2e5ba833-c0fa-4fad-b383-b3854e5fa112"></td>
    </tr>
  </table>
</details>

To achieve this goal we had to:

- use our own scroll component on iOS for `FlatList` of `ReportListView` (it uses `reanimated` to move the view by the offset)
- use `KeyboardAvoiding` so we're in sync with keyboard-avoiding style updates
- implement animated style handling using a state machine
- create `ActionSheetAwareScrollViewContext` to pass the transition function, so we can transition from one state to another by action triggered from different parts of the application

Below you can find more use cases that involves keyboard and other UI elements interactions:

<details>
  <summary>Interactions with other UI elements</summary>
  <table>
    <tr>
      <th>Add attachment</th>
      <th>Call popover</th>
    </tr>
    <tr>
      <td><video src="https://github.com/Expensify/App/assets/22820318/04991f52-a41b-4d3e-abaf-12712b94f1aa"></td>
      <td><video src="https://github.com/Expensify/App/assets/22820318/91b076a9-4c7a-4fbb-bc50-1e8194763e9d"></td>
    </tr>
  </table>
</details>

> [!NOTE]
> **Call popover**, **bottom sheet** and **emoji picker** are implemented as **Modal** windows and on iOS keyboard instantly gets hidden when modal window is shown. But "**Attachment**" popup is implemented not as modal, so for this type of transition we are using `progress`-based animations. When we press on **plus** -> popup is already shown under keyboard and we dismiss the keyboard. Overall the mechanism is similar to what we've used in the past, except the moment where we substitute `translationY` value - in case when keyboard disappear/appears we interpolate it by `progress` variable to achieve smooth transitions.

### Technical details ⚒️ 

So in order to calculate the offset for the message, we:

1. Measure the message item in the window using `ref.current.measureInWindow`, so we can have `height` and `y` coordinate of the image
2. Measure the height of the bottom sheet using `onLayout` callback
3. Get the window height from `Dimensions`
4. Get `safeArea.top` since `y` coordinate doesn't get into account safe area
5. Calculate offset as `elementOffset = (y + safeArea.top + itemHeight) - (windowHeight - popoverHeight)`
In order to animate item when they keyboard was open, we need to also know its height which is provided by useAnimatedKeyboard hook from react-native-reanimated, but we should always subtract safeArea.bottom from it since it's not part of the calculation for the offset, but only keyboardHeight.

Overall, the keyboard is the tricky part. On iOS, the keyboard doesn't resize the viewport. So for the content to not be covered by the keyboard, `KeyboardAvoidingView` is used. In React Native, `KeyboardAvoidingView` is implemented as a view that changes its height or resizes the children's view by the keyboard height. It's done using RN's Layout Animations. This results in the animation for the keyboard being applied using the native event emitter callback which uses the bridge. So open/close events of the keyboard will always be delayed that resulting in the delayed keyboard avoiding animation and, which is more important, scheduling layout animations.

In order to be consistent 100% of the time with animation and timings, we had to:

- use `KeyboardAvoidingView` from `react-native-keyboard-controller` for iOS. It reacts on keyboard height/state change on UI thread in the same frame, so we can schedule an update for styles in the next frame
- use just` useDerivedValue` for all the logic, since introducing `useAnimatedReaction` or having multiple shared values will delay updates for 1 frame, but we want to be 100% in sync with keyboard updates
- use `paddingTop: translateY` (because `ScrollView` is inverted).

Another important aspect of the implementation is the usage of the state machine for the animation. State machine allows us to specify the chain of events and how to handle them, and more importantly what actions to ignore, for example:

1. Initial state is idle. it can react to `KEYBOARD_OPEN` action
2. we open emoji picker. it sends `EMOJI_PICKER_OPEN` action
3. there is no handling for this action in idle state so we do nothing
4. we close emoji picker and it sends `EMOJI_PICKER_CLOSE` action which again does nothing
5. we open keyboard. it sends `KEYBOARD_OPEN` action. idle can react to this action by transitioning into keyboardOpen state
6. our state is keyboardOpen. it can react to` KEYBOARD_CLOSE`,` EMOJI_PICKER_OPEN` actions
7. we open emoji picker again. it sends `EMOJI_PICKER_OPEN` action which transitions our state into emojiPickerOpen state. now we react only to `EMOJI_PICKER_CLOSE` action
8. before rendering the emoji picker, the app hides the keyboard. it sends `KEYBOARD_CLOSE` action. but we ignore it since our `emojiPickerOpen` state can only handle `EMOJI_PICKER_CLOSE` action. so we write the logic for handling hiding the keyboard but maintaining the offset based on the keyboard state shared value
9. we close the picker and send `EMOJI_PICKER_CLOSE` action which transitions us back into `keyboardOpen` state.

### Fixed Issues

$ https://github.com/Expensify/App/issues/10632   
PROPOSAL: GH_LINK_ISSUE(COMMENT)

### Tests

These steps were performed on all the platforms. The new behavior is iOS only for now.
1. Open chat
2. Long press on the first message - message should be pushed up to be displayed above the bottom sheet
3. Long press on the top message - if the message is not covered by the bottom sheet, it should stay in place
4. Long press on any attachment (document, message) should result in the same behavior
5. Pressing on reaction icon on the message should push the message above it too.
6. Long press on the message and then pressing on the reactions icon should open the reactions sheet and if the message is covered by the emoji picker it should animate the message so it's above the picker too

Before each next we make sure the keyboard is open:
1. Long press on the message should either animate the message above the bottom sheet or keep it in place
2. Messages should not "jump" because of keyboard hide/open when showing bottom sheet, emoji picker
3. Press on emoji picker in the composer, messages should stay in place even though the keyboard was closer - emoji picker is rendered instead of keyboard
4. Long press on the mesage, click on add reactions - close emoji reactions

### Offline tests

Should not affect the offline state.

### QA Steps

Same as Tests

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

</details>

<details>

</details>

<details>

</details>

<details>
<summary>Desktop</summary>

</details>

<details>
<summary>iOS</summary>

</details>

<details>
<summary>Android</summary>

</details>


terrysahaidak:
recheck

hannojg (commented):
Hey @terrysahaidak the changes look good to me (I haven't tested them actually). Just added a few comments about formatting for the way Expensify likes to have it 😊

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,447 @@
...
+function renderScrollComponent(props) {
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    return <ActionSheetAwareScrollView {...props} />;
+}
+
+export {
+    renderScrollComponent,
+    ActionSheetAwareScrollViewProvider,
+    ActionSheetAwareScrollViewContext,
+    withActionSheetAwareScrollViewContext,
```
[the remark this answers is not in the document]
terrysahaidak: Thanks, done

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,447 @@
...
+    const transitionWorklet = useWorkletCallback((action) => {
+        if (!action) {
+            throw new Error('state machine action is required');
+        }
+
+        const state = currentState.value;
+
+        // uncomment for debugging
+        // console.log('Current STATE: ', state.current.state);
+        // console.log('Next ACTION: ', action.type, action.payload);
```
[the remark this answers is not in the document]
terrysahaidak: Thanks, done

src/components/ActionSheetAwareScrollView/index.js
```diff
@@ -0,0 +1,84 @@
...
+export default ActionSheetAwareScrollView;
+
+// mock
+const renderScrollComponent = undefined;
+
+export {
+    renderScrollComponent,
+    ActionSheetAwareScrollViewProvider,
+    ActionSheetAwareScrollViewContext,
+    withActionSheetAwareScrollViewContext,
```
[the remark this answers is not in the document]
terrysahaidak: Thanks, done

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,527 @@
...
+    const transition = useCallback((action) => {
+        runOnUI(transitionWorklet)(action);
+    }, []);
+
+    return {
+        currentState,
+        transitionWorklet,
+        transition,
+        reset,
+        resetWorklet,
```
roryabraham, line 146: Looks like `resetWorklet` is not used anywhere outside this function.

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,527 @@
...
+import {makeRemote} from 'react-native-reanimated/src/reanimated2/core';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+import getComponentDisplayName from '../../libs/getComponentDisplayName';
+import styles from '../../styles/styles';
+import Log from '../../libs/Log';
+
+// When you need to debug state machine change this to true
+const DEBUG_MODE = false;
+
+function log(message, params) {
```
roryabraham, line 25: Can we please:

1. Extract this to a separate library like `src/libs/WorkletLog`?
1. Implement the other log levels as well:
    - `info`
    - `alert`
    - `warn`
    - `hmmm`
    - `client`

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        const nextState = nextMachine[action.type];
+
+        if (!nextState) {
+            log(`No next state found for action: ${action.type}`);
+            return;
+        }
+
+        // eslint-disable-next-line no-nested-ternary
```
roryabraham, line 111: Let's respect this eslint rule please and avoid using nested ternaries

src/components/KeyboardAvoidingView/index.ios.js
```diff
@@ -1,13 +1,74 @@
...
-);
+const KeyboardAvoidingView = (props) => {
+    const behavior = props.behavior;
+
+    const keyboard = useAnimatedKeyboard();
+    const insets = useSafeAreaInsets();
+
+    // makeRemote makes an object that can be used on the UI thread
+    // similar to using `global` in worklet but it's just a local object
+    const ctx = useMemo(() => makeRemote({}), []);
```
roryabraham, line 21: I feel a bit suspicious that we need to use `makeRemote`. It's not included anywhere in their public documentation, and is not included in the main exports of the library.

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,527 @@
...
+        reset,
+        resetWorklet,
+    };
+}
+
+const ActionSheetAwareScrollViewContext = createContext();
+
+const stateMachine = {
+    idle: {
+        POPOVER_OPEN: 'popoverOpen',
```
roryabraham, line 154: I think you should extract a separate const called `STATES`:

```js
// All the valid states of the "Action Sheet"
const STATES = {
    IDLE: 'idle',
    POPOVER_OPEN: 'popoverOpen',
    KEYBOARD_OPEN: 'keyboardOpen',
    // etc...
};
```

Then create a second const called `ACTIONS`:

```
// A set of actions that can affect the state of the "Action Sheet"
const ACTIONS = {
    OPEN_POPOVER: 'openPopover',
    OPEN_KEYBOARD: 'openKeyboard',
    CLOSE_KEYBOARD: 'closeKeyboard',
    // etc...
};
```

Then create a third object called `STATE_MACHINE`:

```js
// A state machine which represents the valid actions that can be taken from each state,
// and the ending state after those actions are taken
const STATE_MACHINE = {
    [STATES.IDLE]: {
        [ACTIONS.OPEN_POPOVER]: STATES.POPOVER_OPEN,
        [ACTIONS.OPEN_KEYBOARD]: STATES.KEYBOARD_OPEN,
        // etc...
    },
    [STATES.KEYBOARD_OPEN]: {
        [ACTIONS.CLOSE_KEYBOARD]: STATES.IDLE,
        // etc...
    },
    // etc...
};
```

roryabraham (changes requested):

src/components/ActionSheetAwareScrollView/index.ios.js
terrysahaidak: I don't think it's worth it to even do it since you should never do "logging" in workelts except when debugging them. Since calling things from the JS thread on the UI thread is overhead you don't want to have in animations at all.

I left this since it's useful when adding new stuff. But it basically does nothing when "DEBUG_MODE" is off.

src/components/ActionSheetAwareScrollView/index.ios.js
terrysahaidak: Thanks for pointing it out. I totally forgot about this one.

src/components/KeyboardAvoidingView/index.ios.js
```diff
@@ -1,13 +1,74 @@
 /*
  * The KeyboardAvoidingView is only used on ios
  */
-import React from 'react';
-import {KeyboardAvoidingView as KeyboardAvoidingViewComponent} from 'react-native';
+import React, {useMemo} from 'react';
```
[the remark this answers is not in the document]
terrysahaidak: If they plan to build it following the same API and behavior React Native has - it will be identical.

src/components/ActionSheetAwareScrollView/index.ios.js
roryabraham: It's either:

- worth it, in which case we should move it
- not worth it, in which case we should remove it

but we should not leave it here as-is

terrysahaidak: I can move it inside the state machine hook since it's used only in there and is useful only there.
There is a plan to add this same support for Android, and removing it completely feels wrong.

roryabraham: > What's the best place to put library-like hooks?

You can create a new directory at `src/hooks` for custom hooks.

terrysahaidak: Moved this hook into its own directory/file

src/components/ActionSheetAwareScrollView/index.ios.js
terrysahaidak: Fixed.

src/components/ActionSheetAwareScrollView/index.ios.js
terrysahaidak: Yeah, it was my goal, but i was easier to go without it. But now it's used.

src/components/ActionSheetAwareScrollView/index.ios.js
terrysahaidak: It's now own separated hook so keeping this method as "exported" for it makes more sense.

src/components/ActionSheetAwareScrollView/index.ios.js
```diff
@@ -0,0 +1,527 @@
...
+ * Context will include:
+ * - currentActionSheetState - the current state of the state machine
+ * - transitionActionSheetState - a function to transition the state machine
+ * - transitionActionSheetStateWorklet - a worklet function to transition the state machine
+ * - resetStateMachine - a function to reset the state machine to the initial state
+ *
+ * @param {React.Component} WrappedComponent
+ * @returns {React.Component} A wrapped component that has the ActionSheetAwareScrollViewContext
+ */
+function withActionSheetAwareScrollViewContext(WrappedComponent) {
```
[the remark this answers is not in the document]
terrysahaidak: In my recent changes i use `useContext` for functional components and `.contextType` for everything else - let me know if it's better.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
```diff
@@ -0,0 +1,237 @@
...
+    duration: 350,
+    easing: Easing.bezier(0.33, 0.01, 0, 1),
+};
+
+function ActionSheetKeyboardSpace(props) {
+    const safeArea = useSafeAreaInsets();
+    const keyboard = useAnimatedKeyboard();
+
+    // similar to using `global` in worklet but it's just a local object
+    const ctx = useMemo(() => ({}), []);
```
roryabraham, line 26: Sorry, I'm confused why this is needed. Why can't we just do:

```js
const ctx = {};
```

?

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
```diff
@@ -0,0 +1,237 @@
...
+    const windowHeight = Dimensions.get('screen').height;
```
roryabraham, line 27: Should this be:

```js
const {height} = useWindowDimensions();
```

?

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
```diff
@@ -0,0 +1,237 @@
...
+            transitionActionSheetState({
+                type: Actions.ON_KEYBOARD_CLOSE,
+            });
+        };
+        const removeDidHide = Keyboard.addListener('keyboardDidHide', onDidHide);
+
+        return () => {
+            resetStateMachine();
+
+            // We try to use the new API first, but it doesn't work all the time
```
roryabraham, line 64: This seems like a workaround, can we try to fix the root cause instead?

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
```diff
@@ -0,0 +1,237 @@
...
+    const ctx = useMemo(() => ({}), []);
+    const windowHeight = Dimensions.get('screen').height;
+    const {
+        currentActionSheetState, transitionActionSheetStateWorklet: transition, transitionActionSheetState, resetStateMachine,
+    } = useContext(ActionSheetAwareScrollViewContext);
+
+    useEffect(() => {
+        // sometimes it might trigger multiple times the same actions
+        let lastAction = null;
+        const onDidShow = () => {
```
roryabraham, line 35: Minor detail, but one of our review guidelines is that "functions should be named for what they do, not what they handle". So I would recommend just declaring this inline in the call to `Keyboard.addListener`. I think that's a bit easier to understand why we have these nested functions too.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
```diff
@@ -0,0 +1,237 @@
...
+                removeDidShow();
+                removeDidHide();
+            } catch (err) {
+                Keyboard.removeSubscription(onDidShow);
+                Keyboard.removeSubscription(onDidHide);
+            }
+        };
+    }, []);
+
+    // We need this because of the bug in useAnimatedKeyboard.
```
roryabraham, line 76: Cool, again let's get the fix merged upstream (if it's not already merged), update Reanimated in a separate PR, then remove the workaround here.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
roryabraham: Is it because we want `ctx` to be preserved between renders?

roryabraham: Also, I think it would be clearer to track `shouldRunAnimation` and `keyboardValue` separate v.s: putting them in the same object.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
```diff
@@ -0,0 +1,237 @@
...
+    Easing,
+    KeyboardState,
+    useAnimatedKeyboard,
+    useAnimatedStyle, useDerivedValue, withTiming,
+} from 'react-native-reanimated';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+import styles from '../../styles/styles';
+import {Actions, States, ActionSheetAwareScrollViewContext} from './ActionSheetAwareScrollViewContext';
+
+const config = {
```
roryabraham, line 16: What's the story with this fancy bezier timing?

src/hooks/useWorkletStateMachine.js
roryabraham, line 1: Does this seem like something that might be useful in Reanimated itself? Or in a separate lib entirely? If it's using a separate lib, should it depend on Reanimated or something a bit more generic like https://github.com/chrfalch/react-native-worklets ?

src/components/ActionSheetAwareScrollView/index.ios.js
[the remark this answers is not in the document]
roryabraham: Yep, that's better.

src/hooks/useWorkletStateMachine.js
terrysahaidak: It uses worklets, so it can be used with both Reanimated and rn-worklets.
I don't think it should be its own library since this hook doesn't really implement a really simple state machine. There are way better libraries for such things. But since we're in worklet context - both Reanimated and rn-worklets don't support npm modules. 

Our case is pretty unique. It could be way simpler if we did not depend so much on frame synchronization. We need 100% of cases be in sync with the keyboard. Even 1 frame delay causes visible flickers. Because of that we can't really use other useful Reanimated things like `useAnimatedReaction` and separate things into other shared values, since all of these things will introduce that 1-frame delay.

But I'll ask guys from Reanimated if it's something they are interested in. Since adding it might add more work, but be used only by us.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
terrysahaidak: This bezier timing represents something similar to how the keyboard works on iOS and Android. It's kinda in between these two animations. On iOS keyboard uses native spring which we can't really implement in React Native.

But this config is subject to change after I get more feedback from people testing this PR on their devices.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
terrysahaidak: I totally forgot about the second value, i'll put them directly into the same object instead and will use a better name than `ctx`, something like `localState`.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
roryabraham: Can't we use `useAnimatedKeyboard` to hook into the actual position of the native keyboard on a frame-by-frame basis?

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
roryabraham: Cool, well even if it takes extra time I'd rather wait for the upstream fix v.s: implement a workaround

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
tomekzaw: @roryabraham @terrysahaidak I think I've already mentioned it but the current implementation of `useAnimatedKeyboard` has a design flaw that introduces one frame of delay which is noticeable to some users and may cause UI flickers. The good news is that we have an idea how this should work and we are refactoring the native implementation of `useAnimatedKeyboard` which should hopefully eliminate the delay.

src/components/KeyboardAvoidingView/index.ios.js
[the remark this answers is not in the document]
tomekzaw: It's really nice to see how `useAnimatedKeyboard` can be used in cases just like this one. Since the hook itself resolves one of major pain points of the React Native, it makes sense to go one step forward and include the `KeyboardAvoidingView` component in the library as well, so that the whole community can benefit from it, instead of just embedding it into the source code of the E/App.

@terrysahaidak Would you be open to submitting this part of the code as a PR to react-native-reanimated? We'll be more than happy to review and test it or even help with the implementation. Ideally, it should expose the same API as in the one from React Native so it can be used as a drop-in replacement, but if that's a limitation, we can think of a better API, just let me know if you have any ideas.

roryabraham:
@terrysahaidak btw, you've got merge conflicts here

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
terrysahaidak: @tomekzaw I created an issue in the Reanimated repo that shows the problem I have. I tried with both 3.0.2 and the latest commit as of today:
https://github.com/software-mansion/react-native-reanimated/issues/4324

TLDR: the keyboard height is updating and the `useAnimatedStyle` hook is triggered for each value change, but the styles won't update until I click somewhere on the screen.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
terrysahaidak: I think it happens only sometimes randomly for me on iphone 11. 
I have more problems with modals when it has a constant 1-frame delay.
Basically, when you open a modal, the keyboard is closed without animation. And when you open the modal - it's back again. And now even though the state is correct the height has 1 frame delay, but i just use the previous height so it's not a big of an issue.
But that random flickering is.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
terrysahaidak: Removed them since we're not gonna use it anymore

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
terrysahaidak: I'll look into that again, but it seems like it no longer happens for me, will play around it again

terrysahaidak:
Created a PR to Reanimated repo implementing drop-in-replacement for React Native Keyboard Avoiding View:
https://github.com/software-mansion/react-native-reanimated/pull/4339

In the future, we can add more things to this implementation so improve the developer experience using the component.

terrysahaidak:
@roryabraham, reanimated team fixed some of the bugs with `useAnimatedKeyobard` but @tomekzaw mentioned that the refactoring of that component is still in progress. I created a PR for KeyboardAvoidingView and it also reproduces one bug @tomekzaw is aware of. I think this PR should be on Hold until we have these fixes.

stitesExpensify:
Hi @terrysahaidak now that the reacti-native-reanimated PR is merged, are we ready to move forward here? This issue is blocking emoji Search for Mobile which we would really like to get out ASAP!

terrysahaidak:
> Hi @terrysahaidak now that the reacti-native-reanimated PR is merged, are we ready to move forward here? This issue is blocking emoji Search for Mobile which we would really like to get out ASAP!

hey, there was a release, but not everything is fixed, there is a tiny flicker when opening a modal. 

@tomekzaw is currently investigating this issue, should be the last one.

roryabraham:
> there is a tiny flicker when opening a modal

@terrysahaidak @tomekzaw can one of you link the issue here so I can subscribe?

Also looks like we'll probably want to try to get https://github.com/software-mansion/react-native-reanimated/pull/4339 merged before proceeding with this.

terrysahaidak:
@roryabraham yes, that PR is basically a playground for all the bugs `useAnimatedKeyboard` might have, so once it's ready - we can safely use `useAnimatedKeyboard` and KeyboardAvodingView in our code. 

The flicker I mentioned is kind of random, related to this issue https://github.com/software-mansion/react-native-reanimated/issues/4351, i saw it in Reanimated 3.0.2, so could be non-related to other fixes Reanimated team already implemented.

It will affect the logic in this PR as well as https://github.com/software-mansion/react-native-reanimated/pull/4339

image

terrysahaidak:
@stitesExpensify I'll work with @perunt to see what we can do to unblock the emoji search while we're waiting for this PR.

stitesExpensify:
Sounds great, thanks!

perunt:
I checked it out again and figured how to get it done without Terry's changes. Basically, I'll use them later to make it even better.

terrysahaidak:
The reanimated upgrade did not go well, the app is crashing, looking into what causes that, but it looks like it's a common issue:
https://github.com/Expensify/App/pull/18073#issuecomment-1531136904

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.js
roryabraham: This seems to have been solved upstream, resolving this thread

roryabraham: Actually, we need to remove the workaround from here in this PR

roryabraham:
Still waiting to see how https://github.com/software-mansion/react-native-reanimated/pull/4339 plays out

chrispader:
Hey @kirillzyusko !

Thanks for taking a look at this PR. Once you've done your investigations, could you give an update here if we should look into using `react-native-keyboard-controller` instead and if so, how?

Thanks 🙌🏼

tomekzaw:
I believe the answer is in the description of the PR that @terrysahaidak submitted to Reanimated 😄 

Zrzut ekranu 2023-11-3 o 18 14 29

chrispader:
> I believe the answer is in the description of the PR that @terrysahaidak submitted to Reanimated 😄 The PR is being reviewed at the moment so please let us know if we should hurry or not 😄

@kirillzyusko i'll let you take this PR over and decide on this! Hit me up if there's anything i can help with...

kirillzyusko:
Just a quick update from my side - I've experimented with new reanimated version and I saw again a problem of incorrect keyboard frame for one frame. When I close popover (modal), then I have next values:
- 0 (keyboard closed);
- 291 (keyboard opened)
- 0 (why?)
- 336 (keyboard is fully opened)

The thing with intermediate keyboard value (291) was introduced in iOS 17 - it seems like the OS now distinguish events when keyboard is shown and autosuggestions are added. Nonetheless that on UI we'll see only one frame (keyboard + autosuggestion = 336) we'll receive two events instead of one (but with REA hook I was getting 3 - one with empty frame 🤷‍♂️ ).

To quickly verify I decided to switch to `react-naive-keyboard-controller` (to see whether such problem happens with this library or not + in slack we've also discussed a possibility to use this library for interactive keyboard dismissal). With RNKC I don't see that issue and I'm getting correct order of frames: 0, 291, 336. However in rare cases I was also getting one incorrect frame (that was due to the fact that in `KeyboardAvoidingView` we are using `paddingBottom` and in `ActionSheetKeyboardSpace` we were using `translateY` - I switched to `paddingBottom` in both cases and now it works kind of okay). This is still WIP, but here is results for several transitions:

|Keyboard + Popover|Keyboard -> emoji picker|Bottom sheet (no keyboard)|
|--------------------|-------------------------|----------------------------|
|<video src="https://github.com/Expensify/App/assets/22820318/f6728610-5da3-42e3-a304-6c1c402e6660">|<video src="https://github.com/Expensify/App/assets/22820318/7e4ce35d-c87b-43c8-8585-f047d61f6cf7">|<video src="https://github.com/Expensify/App/assets/22820318/7297bdc0-8c00-4478-bb8c-3ad7b7b38772">|

For safety reasons I decided to make a work in separate branch: https://github.com/margelo/expensify-app-fork/tree/fix/10632-when-long-press-on-message-add-space-for-image-keyboard-controller

If I can make it finally working, then I'll merge these changes into this branch 👀

hannojg:
These results look really promising @kirillzyusko 😻

kirillzyusko:
Just sharing an update 🙂 

I went further with `react-native-keyboard-controller` library and it proves to be working fine. Basic interactions works fine and don't conflict with each other (previous demo-videos were captured separately). However I'm more than sure that current codebase will have some issues on some edge cases - right now I'm actively testing the implementation and I'm trying to discover all these issues. My target it to provide a testable build by 26.01.24 (if it works okay I'll merge my branch into this PR).

Below I attached a demo how it works right now (captured on iPhone 11, iOS 17.2 - I recorded everything in one go, but divided video in order not to exceed GitHub file uploading limit (10 Mb)).

|Part 1|Part 2|
|------|------|
|<video src="https://github.com/Expensify/App/assets/22820318/b64f33e9-4afb-428c-8c49-4f2c808d23de">|<video src="https://github.com/Expensify/App/assets/22820318/0e161141-5577-425b-98f5-6f7119b4e527">|

src/App.js
```diff
@@ -69,14 +71,17 @@ function App() {
...
+                    ActionSheetAwareScrollView.ActionSheetAwareScrollViewProvider,
                 ]}
             >
-                <CustomStatusBarAndBackground />
-                <ErrorBoundary errorMessage="NewExpensify crash caught by error boundary">
-                    <ColorSchemeWrapper>
-                        <Expensify />
-                    </ColorSchemeWrapper>
-                </ErrorBoundary>
+                <KeyboardProvider enabled={Platform.OS === "ios"}>
```
roryabraham, line 77: I notice you're enabling this only on iOS, but this issue occurs the same on Android as it does on iOS. So I think we'd like to see a consistent approach to keyboard avoidance on all platforms.

Currently, I think we use an unintuitive mix of native and JS-layer keyboard avoidance, and it's inconsistent between platforms.

roryabraham: Also, should this be inside the error boundary?

src/components/KeyboardAvoidingView/index.tsx
```diff
@@ -13,6 +13,4 @@ function KeyboardAvoidingView(props: KeyboardAvoidingViewProps) {
     );
 }
 
-KeyboardAvoidingView.displayName = 'KeyboardAvoidingView';
```
roryabraham, line 16: This change looks like it should be undone

src/hooks/useWorkletStateMachine.ts
```diff
@@ -0,0 +1,166 @@
+import {useCallback} from 'react';
+import {runOnJS, runOnUI, useSharedValue} from 'react-native-reanimated';
+import Log from '@libs/Log';
+
+// When you need to debug state machine change this to true
+const DEBUG_MODE = true;
```
roryabraham, line 6: You shouldn't really need this flag I don't think. Just use `Log.client` (I don't think we want these logs sent to server).

We currently have Logs enabled in production, but I don't think there's really a performance concern with logs that's in-scope for this issue.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
```diff
@@ -0,0 +1,411 @@
...
+    damping: 500,
+};
+
+function ActionSheetKeyboardSpace(props: ViewProps) {
+    const styles = useThemeStyles();
+    const safeArea = useSafeAreaPaddings();
+    const keyboard = useAnimatedKeyboard();
+
+    // similar to using `global` in worklet but it's just a local object
+    const [syncLocalWorkletState] = useState({
```
roryabraham, line 109: Why would we need to use `useState` if we're never modifying the state?

roryabraham: Seems like this should use `useRef` instead?

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
```diff
@@ -0,0 +1,411 @@
...
+    useEffect(() => () => resetStateMachine(), [resetStateMachine]);
+
+    useAnimatedReaction(
+        () => keyboard.state.value,
+        (lastState) => {
+            if (lastState === syncLocalWorkletState.lastState) {
+                return;
+            }
+
+            syncLocalWorkletState.lastState = lastState;
```
roryabraham, line 125: Don't typically see this done for state – is this meant to re-render the component by potentially changing the state?

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
```diff
@@ -0,0 +1,411 @@
...
+                /* console.log("TRANSITION #14 -> ", lastKeyboardHeight);
+                return lastKeyboardHeight; */
+                console.log("TRANSITION #13 ->", 0);
+                return 0;
+            }
+
+            case States.KEYBOARD_POPOVER_CLOSED: {
+                // transition({type: Actions.END_TRANSITION});
+                if (keyboard.heightWhenOpened.value === keyboard.height.value) {
+                    console.log("TRANSITION #15 (1) -> ", 0, {keyboard: keyboard.height.value, lastKeyboardHeight, heightWhenOpened: keyboard.heightWhenOpened.value});
```
roryabraham, line 302: If you want to keep these logs, we should use the `Log` lib

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
kirillzyusko: @roryabraham originally it was created by @terrysahaidak so I think he can provide more output.

But as I can understand such construction serves as an analog of `useRef` in reanimated world. If we re-assign a shared value, then corresponding `useDerivedValue`/`useAnimatedReaction` effects will be run. And in order to save a value somewhere and don't re-run reanimated lifecycles we're using state variable with direct assigning (on next reanimated hooks execution they will capture a new variable into the closure).

Such trick will not work with `useRef` I guess, because `ref` will always point to the same object (it has only `current` property that can be modified).

@terrysahaidak also don't hesitate to jump into conversation if my explanation is not correct 👀

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
kirillzyusko: @roryabraham see my answer above - I'll resolve this thread to keep a communication in single place if you don't mind (because anyway we're talking about the same thing).

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
kirillzyusko: @roryabraham all these logs (as well as commented out code) intended to be removed before a merge. But thanks for the hint - next time I'll use `Log` directly.

src/hooks/useWorkletStateMachine.ts
kirillzyusko: Again this code from @terrysahaidak 

@terrysahaidak what do you think?

src/App.js
kirillzyusko: > Also, should this be inside the error boundary?

I added it to the level of all providers - the reason why it's a separate JSX element because I specify `props` (enabled) and right now `ComposeProviders` doesn't have an ability to specify additional props. And as I can see all providers are not under `ErrorBoundary`.

Would you mind to give a brief explanation why do you think it's worth to put this element under `ErrorBoundary`?

> Currently, I think we use an unintuitive mix of native and JS-layer keyboard avoidance, and it's inconsistent between platforms.

Yes, because an initial PR supported only iOS (from PR description: _This PR is iOS only for now. Other platforms are not affected. Also, in this explanation, we are talking about iOS only since on Android, iOS web and Android Web keyboard handling works differently_.)

But I agree with you that it would be good to have cross-platform code that handles keyboard interactions. And we actually wanted to suggest to do that, but our vision was to merge this PR first (because it's quite old and contains a lot of changes) and then add Android support in next PRs. What do you think about it?

src/hooks/useWorkletStateMachine.ts
terrysahaidak: Having these logs is only helpful when you debug these state machines, in other cases there will be just a bunch of noise in the logs every time you interact with the keyboard and other things that interact with the state machine.

Another reason for having this flag was to avoid unnecessary calls from the UI thread where we run all the worklets logic to the JS thread where we log stuff, which can affect performance.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
terrysahaidak: I think it can also be something like `const syncLocalWorkletState = useRef({}).current`. It's still a bit unclear for me if this gonna work with useRef the same way it would work with useState which should preserve its state in React's Concurrent mode. useState was the saver bet.

src/components/ActionSheetAwareScrollView/ActionSheetKeyboardSpace.tsx
terrysahaidak: we should not keep any logs in this code for sure. any animation-related code should be log-free so we don't drop any frames.

src/App.js
hannojg: @kirillzyusko lets create a new `KeyboardHandlerProvider` in components, that returns the <KeyboardProvider with the `enabled` prop set.
Then we can pass it to `ComposeProviders` so it's in line with all the other providers

src/components/ConfirmContent.tsx
```diff
@@ -85,15 +86,27 @@ function ConfirmContent({
     contentStyles,
     iconAdditionalStyles,
 }: ConfirmContentProps) {
+    const actionSheetAwareScrollViewContext = useContext(ActionSheetAwareScrollViewContext);
     const styles = useThemeStyles();
     const {translate} = useLocalize();
     const theme = useTheme();
     const {isOffline} = useNetwork();
 
+    const onLayout = (event) => {
```
hannojg, line 95: useCallback

src/components/ShowContextMenuContext.ts
```diff
@@ -55,7 +57,7 @@ function showContextMenuForReport(
         action?.reportActionID,
         ReportUtils.getOriginalReportID(reportID, action),
         undefined,
-        checkIfContextMenuActive,
+        undefined,
```
hannojg, line 60: Is that change correct? Seems like we removed existing code? (if we don't need it anymore its probably good to cleanup the usage of `checkIfContextMenuActive`)

src/hooks/useWorkletStateMachine.ts
```diff
@@ -0,0 +1,166 @@
...
+    const currentState = useSharedValue(initialState);
+
+    const log = useCallback((message: string, params?: Payload | null) => {
+        'worklet';
+
+        if (!DEBUG_MODE) {
+            return;
+        }
+
+        runOnJS(Log.info)(`[StateMachine] ${message}`, false, params);
```
hannojg, line 85: Should be changed to `Log.client` (no need to send to the server)

src/hooks/useWorkletStateMachine.ts
hannojg: I think we should change it to `Log.client` but keep that flag (but set it to `false`), because it would really bloat the logs if we set it to true. And it's good to keep it in case you ever need to debug it.

Although we could consider just covering it with a unit test, and then as this is quite isolated code it shouldn't really cause any bugs.

src/pages/home/report/ReportActionCompose/AttachmentPickerWithMenuItems.js
```diff
@@ -130,10 +132,11 @@ function AttachmentPickerWithMenuItems({
     isFocused,
     raiseIsScrollLikelyLayoutTriggered,
 }) {
+    const actionSheetAwareScrollViewContext = useContext(ActionSheetAwareScrollView.ActionSheetAwareScrollViewContext);
     const theme = useTheme();
     const styles = useThemeStyles();
     const {translate} = useLocalize();
-    const {windowHeight} = useWindowDimensions();
+    const {windowWidth} = useWindowDimensions();
```
hannojg, line 139: just double checking, its correct we changed from height to width here?

src/pages/home/report/ReportActionItem.js
```diff
@@ -255,7 +259,50 @@ function ReportActionItem(props) {
...
         setIsContextMenuActive(ReportActionContextMenu.isActiveReportAction(props.action.reportActionID));
-    }, [props.action.reportActionID]);
+
+        actionSheetAwareScrollViewContext.transitionActionSheetState({
+            type: ActionSheetAwareScrollView.Actions.CLOSE_POPOVER,
+        });
+    }, [actionSheetAwareScrollViewContext, props.action.reportActionID]);
+
+    const handlePressOpenPicker = useCallback(
+        (openPicker) => {
```
hannojg, line 269: param should be typed

src/pages/home/report/ReportActionItem.js
```diff
@@ -255,7 +259,50 @@ function ReportActionItem(props) {
...
+                        type: ActionSheetAwareScrollView.Actions.CLOSE_EMOJI_PICKER_POPOVER,
+                    });
+                });
+            });
+        },
+        [actionSheetAwareScrollViewContext],
+    );
+
+    const handleShowContextMenu = useCallback(
+        (callback) => {
```
hannojg, line 290: param should be typed

src/pages/home/report/ReportActionItem.js
kirillzyusko: @hannojg in js file? 🤔

src/pages/home/report/ReportActionItem.js
kirillzyusko: @hannojg in js file? 🤔

src/pages/home/report/ReportActionCompose/AttachmentPickerWithMenuItems.js
kirillzyusko: @hannojg I think it's correct, but actually @chrispader made these changes 🤔

kirillzyusko:
I fetched latest `main` and I can see flickering again when I use a combination of keyboard + long press on message. I'll fix it once I finish task with sound.

src/pages/home/report/ReportActionItem.js
hannojg: oh whoops, my oversight, all good 😄

src/pages/home/report/ReportActionItem.js
hannojg: oh whoops, my oversight, all good 😄

kirillzyusko:
@roryabraham I think I addressed all comments - would you mind to have a look on code again and test it, please?

twisterdotcom:
Assigning a C+ here as well

roryabraham:
@shubham1206agra I think before you dive in here, we should wait for consensus on https://github.com/Expensify/App/issues/35270

jakub-trzebiatowski (changes requested):

cubuspl42:
I understand that this is conflict-prone. So I guess we'll need to parallelize review and testing with conflict resolution.

@kirillzyusko Would you handle minor comments I left and ensure to keep all commits signed? We'll need at least one force-push to fix the commit signing thing.

kirillzyusko:
@cubuspl42 @roryabraham I couldn't sign unsigned commits 😔 All ways come to rebase - and it's close to unreal to rebase 137 commits over 50000 other commits (we literally have incompatible changes even in first commit, because from 5 modified files 4 were removed (migrated to TS I guess) - so it's unbelievably difficult to rebase these changes to latest branch.

I suggest to keep this PR in the current shape and move all work in a new PR with squashed commit - https://github.com/Expensify/App/pull/38232

Such approach has several benefits:
- In case if we need to have an exact history of changes we always can go to this PR and see why a particular commit was added
- we'll not send an unbelievable amount of time for rebasing and further post fixing
- I opened a PR, so I have all access to PR description etc, so we don't need to involve more people to change PR description, check/uncheck mark etc.

Let me know what do you think 😊

kirillzyusko:
> @kirillzyusko Would you handle minor comments I left and ensure to keep all commits signed? We'll need at least one force-push to fix the commit signing thing.

@cubuspl42 yeah, sure, let's come to conclusion what should we od with unsigned commits and then I'll address your comments 👍

kirillzyusko:
> Expensify has a policy of accepting only signed commits to the main branch, and I don't think this situation is special enough to make an exception.

Yeah, I know - and I agree that it's not exceptional situation, that's why I proposed squashing and new PR (to keep history of changes) 😅 

Waiting for @roryabraham feedback 👀

roryabraham:
closing this in favor of https://github.com/Expensify/App/pull/38232

PR #17400, opened 2023-04-13
`useOnNetworkReconnect()`

marcaaron:
### Details

See linked issue and discussion here:  https://expensify.slack.com/archives/C01GTK53T8Q/p1681117333835879

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/276050

### Tests

1. Navigate to the request money page
2. Verify that the `OpenIOUModalPage` command runs
3. Go offline
4. Go online
5. Verify that the `OpenIOUModalPage` command runs again because we reconnected (check the Network logs e.g. Chrome Dev Tools)

### Offline tests

N/A

### QA Steps

Web QA only (same as tests - it is harder to test on mobile)

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

</details>

<details>

</details>

<details>

</details>

<details>
<summary>Desktop</summary>

</details>

<details>
<summary>iOS</summary>

</details>

<details>
<summary>Android</summary>

</details>


src/components/hooks/useOnNetworkReconnect.js
```diff
@@ -0,0 +1,21 @@
...
+    const {isOffline} = useContext(NetworkContext);
+    const prevOfflineStatusRef = useRef(isOffline);
+    useEffect(() => {
+        // If we were offline before and now we are not offline then we just reconnected
+        const didReconnect = prevOfflineStatusRef.current && !isOffline;
+        if (!didReconnect) {
+            return;
+        }
+
+        callback();
```
techievivek, line 14: Can we not call the callback with the `didReconnect` so we can do further optimization like @kidroca mentioned in the [thread](https://expensify.slack.com/archives/C01GTK53T8Q/p1681400895336669?thread_ts=1681117333.835879&cid=C01GTK53T8Q)?

techievivek: It looks like the custom hook here is meant only if the network reconnects but maybe we just need `useNetworkState` or something? And then callback can have the updated state of the network and decide what they need to do next.

src/components/Avatar.js
```diff
@@ -62,22 +58,13 @@ const defaultProps = {
 
 function Avatar(props) {
     const [imageError, setImageError] = useState(false);
-    const prevNetworkStatusRef = useRef(props.network.isOffline);
 
-    useEffect(() => {
-        const isReconnecting = prevNetworkStatusRef.current && !props.network.isOffline;
-        if (!imageError || !isReconnecting) {
+    useOnNetworkReconnect(() => {
+        if (!imageError) {
```
techievivek, line 63: If we are sticking with custom hook for reconnect then we can just do 
```js
useOnNetworkReconnect(() => {
    setImageError(false);
}
```

src/components/hooks/useOnNetworkReconnect.js
kidroca: I think this is fine, the idea is to cleanup and simplify, so we have 1 `if` statement here instead of many `if` statements everywhere we need to run such checks 

In terms of usage we achieve a declarative approach
1. on reconnect
2. do something 

No ifs

kidroca: A minor problem here is that `callback` would trigger the hook to re-run on-rerender, when it is passed inline 
For this particular hook this won't cause anything unexpected, just triggering the effect more often than necessary 

#### Problem
We have a custom effect hook that triggers an `action` (callback)
We don't want the `action` itself to be a reason to run/re-run the custom effect
The custom effect should only run the action when a list of dependencies change
But the action itself shouldn't be a dependency

#### Proposal

_Marc, I think you know my way of dealing with that:_

Exclude the action from dependencies by capturing it to a ref (internal hook logic)
This way we only run the effect when network conditions change
But we still use the latest value of the action when we do need to run it

```js
// prevent the callback param from re-triggering the hook if it changes
// we want to only trigger the hook on reconnect, and then use the latest value for the 
// `callback` so we keep a ref updated with the latest value
const callbackRef = useRef(callback);
callbackRef.current = callback;

useEffect(() => {
  // ...
  callbackRef.current();
}, [isOffline])
```

marcaaron: Ah so, one curiosity I have in response is if it matters that the effect re-runs when the callback changes? Since we will only do the callback if the conditions of the props are correct i.e. the `isOffline` changed in the right way. It would almost be a no-op - but maybe not entirely true that it does _nothing_ (still has to do the comparison).

I like the suggestion to use the ref anyway so I think we can do it! But mostly just curious if there was some worse consequence to having the effect run more than necessary.

src/components/Avatar.js
marcaaron: Yeah I don't really get why this optimization existed before really. It is the reconnect we care about. No update should happen I think if the value is the same?

From React docs...
>If the new value you provide is identical to the current state, as determined by an [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will skip re-rendering the component and its children.

src/components/hooks/useOnNetworkReconnect.js
kidroca: I have these 2 concerns. Though I can't measure their possible impact right now

**Performacnewise** triggering the effect would happen on each render unless the caller uses a `useCallback` wrapping for the callback, or it's a static callback (like in the Modal file change)
If we only have a few usages/instances of the hook I wouldn't think it's a big deal, but since it might be used on a frequently rendered component like Avatar it multiplies the potential issue 
It probably won't be noticeable, but little things add up 
_(the actual call to the function is the overhead even if exists immediately - it has to be orchestrated stacked/destaced, at least that's what I think, I don't have a quote to reference)_ 

By using ref for the callback we don't have to convey to developers there might be performance implications so that they should use `useCallback` when passing params to the hook

**Code pattern wise**  someone is bound to need to write a new hook and might use this one as inspiration, not realizing that adding the `callback` to dependencies here does not affect the effect, but it might be a problem for their use case. They'll probably find out soon that there's a problem (the problem stated above) and look for a solution (similar to the proposed solution) - so by doing the ref pattern here we've saved them, but they'll never know 😆

marcaaron: I think they're great points all around. Thanks for your guidance here 🙇

marcaaron:
Working on testing for this one and fixed conflicts will take out of draft soon.

src/components/hooks/useOnNetworkReconnect.js
```diff
@@ -0,0 +1,27 @@
+import {useRef, useContext, useEffect} from 'react';
+import {NetworkContext} from '../OnyxProvider';
+
+/**
+ * @param {Function} onNetworkReconnect
+ */
+export default function useOnNetworkReconnect(onNetworkReconnect) {
+    const callback = useRef(onNetworkReconnect);
+    callback.current = onNetworkReconnect;
```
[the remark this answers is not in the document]
marcaaron: Yes because the function itself could change e.g. if it was passed as a `prop`. Not saying it would be the _best_ idea, but I think we rather support that than confuse someone down the line.

thienlnam:
@marcaaron Are you planning on using all existing instances or should we take care of it on a case by case basis when components are being refactored to functional ones?

marcaaron:
yeah just trying to keep the scope small for this one so there are not too many flows to test at once.

marcaaron:
Oh also sorry did not see there was an issue for this 😓

stitesExpensify (approved):
Makes sense to me!

fedirjh (approved):
LGTM

Issue #17391, opened 2023-04-13
[HOLD for payment 2023-07-26] [$2000] Desktop - App is show endless spinner when tap popup Open app in "New Expensify


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Issue found when executing PR** https://github.com/Expensify/App/pull/16869 

## Action Performed:
1. Open staging.expensify.com, login with account A.
2. Open staging.new.expensify.com and login with same account.
3. Download and open the staging Desktop App, if you haven't already.
4. Click on the concierge button at bottom right on staging.expensify.com/inbox
5. Tap on the "Open app in "New Expensify" popup.

## Expected Result:
Desktop App should be open and user login.

## Actual Result:
Desktop App is show endless spinner.

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [ ] Android / Chrome
- [ ] iOS / native
- [ ] iOS / Safari
- [ ] MacOS / Chrome / Safari
- [x] MacOS / Desktop

**Version Number:** 1.2.99.4

**Reproducible in staging?:** Yes

**Reproducible in production?:** Yes

**If this was caught during regression testing, add the test name, ID and link from TestRail:**

**Email or phone of affected tester (no customers):**

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/93399543/231794818-9e4fb26c-ecfb-4945-9586-dd6cf5cbf65b.mp4

**Expensify/Expensify Issue URL:**

**Issue reported by:** Applause - Internal Team

**Slack conversation:**

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0151b55ece236c38ac</li>
        <li>Upwork Job ID: 1646650899463368704</li>
        <li>Last Price Increase: 2023-04-24</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

[...]

PR #16043, opened 2023-03-16
Pass user session information when opening the electron app from web platform

ntdiary:
### Fixed Issues

$ https://github.com/Expensify/App/issues/15651
PROPOSAL: https://github.com/Expensify/App/issues/15651#issuecomment-1467954112

### Tests

1. Login with account A in the desktop app.
2. Login with account B in the web app.
3. Navigate to the user profile page in the web app.
4. Refresh the browser.
5. A prompt should appear asking to 'Open the app in Electron'.
6. Click on it and wait for the desktop app to open.
7. Verify the user profile page for account B is displayed in the desktop app.

- [x] Verify that no errors appear in the JS console

### Offline tests

1. Login with account A in the desktop app.
2. Login with account B in the web app.
3. Navigate to the user profile page in the web app.
4. Refresh the browser.
5. A prompt should appear asking to 'Open the app in Electron'.
6. Turn off the network.
7. Click on 'Open the app in Electron' and wait for the desktop app to open.
8. Verify the login page is displayed in the desktop app.

### QA Steps

1. Login with account A in the desktop app.
2. Login with account B in the web app.
3. Navigate to the user profile page in the web app.
4. Refresh the browser.
5. A prompt should appear asking to 'Open the app in Electron'.
6. Click on it and wait for the desktop app to open.
7. Verify the user profile page for account B is displayed in the desktop app.

- [ ] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
[...]

Issue #16083, opened 2023-03-17
[HOLD for payment 2023-04-20] [$1000]  When clicking link to chat with Concierge on expensify.com, the pop up option to open the desktop app is not surfaced.


kavimuru:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:

1. Open [expensify.com](http://expensify.com/) on web and log in
2. Click the concierge icon in bottom right corner

## Expected Result:
New tab to be opened and, at the top, for there to be modal that says Open New Expensify?

## Actual Result:
No option surfaced

## Workaround:
unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [ ] Android / Chrome
- [ ] iOS / native
- [ ] iOS / Safari
- [x] MacOS / Chrome / Safari
- [ ] MacOS / Desktop

**Version Number:** 1.2.87-0
**Reproducible in staging?:** y
**Reproducible in production?:** y
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** 

https://user-images.githubusercontent.com/43996225/226037199-03b30b03-6d05-4a1f-970a-6d101eedc5b1.mp4

**Expensify/Expensify Issue URL:**
**Issue reported by:** @mallenexpensify 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1679001987306399

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0172780395328fee9c</li>
        <li>Upwork Job ID: 1638582294280368128</li>
        <li>Last Price Increase: 2023-03-22</li>
    </ul>
</details>

MelvinBot:
Triggered auto assignment to @strepanier03 (`Bug`), see https://stackoverflow.com/c/expensify/questions/14418 for more details.

MelvinBot:
[...]

PR #17452, opened 2023-04-14
clean up the deep link code

ntdiary:
### Details

This is a followup PR for #16043.
After we remove the redirection ui in the deep link component, We no longer need to update the `appInstallationCheckStatus` variable. 
We can also remove the call to `Network.post()`.
We need to implement the transition process: oldDot --> newDot --> Desktop app 

### Fixed Issues

$ https://github.com/Expensify/App/issues/17391

### Tests

Note: If we can't start OldDot locally, we can open the developer tools and add a breakpoint in the source panel to get the `transition` path, just like the image above.

1. Open the local newDot and login with account A (tab 1).
2. Open the local oldDot and login with the same account (tab 2).
3. Open the local Desktop App and stay logged out.
4. Click on the concierge button at bottom right on oldDot/inbox (tab 2).
5. (Optional) Execute `console.dir(t.replace('https://staging.new.expensify.com', ''))` in console to get the `/transition...` path part, create a new tab and type `localhost:8080` and append the transition path part and then open this url (tab 3).
6. Tap on the 'Open Electron.app' button in the tab 3.
7. Verify that account A is also logged in the desktop app.

### Offline tests

N/A
### QA Steps

1. Open staging.new.expensify.com and login with account A (tab 1).
2. Open staging.expensify.com and login with the same account (tab 2).
3. Open the staging Desktop App and stay logged out.
4. Click on the concierge button at bottom right on staging.expensify.com/inbox (tab 2).
5. Tap on the 'Open New Expensify' button in the tab 3.
6. Verify that account A is also logged in the desktop app.

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/8579651/234503847-c4790472-ecaf-44ca-936b-9b3d93a3172a.mov

</details>

<details>
N/A

</details>

<details>
N/A

</details>

<details>
<summary>Desktop</summary>
N/A

</details>

<details>
<summary>iOS</summary>
N/A

</details>

<details>
<summary>Android</summary>
N/A

</details>


src/libs/actions/Session/index.js
```diff
@@ -642,6 +621,23 @@ function authenticatePusher(socketID, channelName, callback) {
     });
 }
 
+function getShortLivedAuthToken() {
+    // TODO: If we have any way to read the shortLivedAuthToken and save it, we can also no longer call makeRequestWithSideEffects.
```
tgolen, line 625: I'm not sure what "and save it" implies. We are saving it in Onyx, but that isn't enough? Also, TODOs are not allowed in our code. Instead, you should open a GH issue and point to this line of code and then detail what needs to be done.

src/pages/LogInWithShortLivedAuthTokenPage.js
```diff
@@ -62,7 +74,12 @@ class LogInWithShortLivedAuthTokenPage extends Component {
     }
 
     render() {
-        if (this.props.isTokenValid) {
+        // TODO: Not sure if we should redirect the user to the login page or just show the expiration screen or offline message
```
tgolen, line 77: Same comment as above about TODOs (unless you are meaning to clean up this TODO in this PR)

tgolen (changes requested):
This looks 💯 times better! Thank you

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -96,7 +101,7 @@ class DeeplinkWrapper extends PureComponent {
                     return;
                 }
                 iframe.parentNode.removeChild(iframe);
-            }, 100);
+            }, 0);
```
marcaaron, line 104: Why do actually need this `setTimeout()`? The explanation above is not very satisfying...

>we need to give this iframe some time for it to do what it needs to do

What does it mean?

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -48,6 +48,10 @@ class DeeplinkWrapper extends PureComponent {
         if (!this.isMacOSWeb() || CONFIG.ENVIRONMENT === CONST.ENVIRONMENT.DEV) {
             return;
         }
+
+        // We need to clear the old short-lived auth token if it exists.
```
marcaaron, line 52: Why do we need to do this?

src/pages/LogInWithShortLivedAuthTokenPage.js
```diff
@@ -74,7 +74,7 @@ class LogInWithShortLivedAuthTokenPage extends Component {
     }
 
     render() {
-        // TODO: Not sure if we should redirect the user to the login page or just show the expiration screen or offline message
+        // redirect the user to the login page if there is a sudden disconnection
         if (this.props.network.isOffline) {
             Navigation.navigate();
```
marcaaron, line 79: Should we be passing something more explicit here? Calling `navigate()` with no arguments seems strange to me.

src/pages/LogInWithShortLivedAuthTokenPage.js
```diff
@@ -74,7 +74,7 @@ class LogInWithShortLivedAuthTokenPage extends Component {
     }
 
     render() {
-        // TODO: Not sure if we should redirect the user to the login page or just show the expiration screen or offline message
+        // redirect the user to the login page if there is a sudden disconnection
```
marcaaron, line 77: I'm unsure what the exact edge case we're trying to handle is based on this comment. The user hits this page and in between downloading the JS bundle and executing it they somehow become offline? Can you expand on what exactly we are preventing and why we are doing this?

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -59,8 +63,9 @@ class DeeplinkWrapper extends PureComponent {
         if (prevProps.session.shortLivedAuthToken || !this.props.session.shortLivedAuthToken) {
```
marcaaron, line 63: Are we not handling the errors we were handling before anymore?

marcaaron (changes requested):
@ntdiary I am working on a clean up for this in another PR... it would have been nice to have a heads up that you are going to tackle improving this code so that we are not duplicating any effort and more generally agree on a plan for how to do it.

src/libs/actions/Session/index.js
```diff
@@ -642,6 +621,23 @@ function authenticatePusher(socketID, channelName, callback) {
...
+    // eslint-disable-next-line rulesdir/no-api-side-effects-method
+    API.makeRequestWithSideEffects(
+        'OpenOldDotLink', {shouldRetry: false}, {},
+    ).then((response) => {
+        const {shortLivedAuthToken} = response;
+        Onyx.merge(ONYXKEYS.SESSION, {shortLivedAuthToken});
```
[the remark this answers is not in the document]
marcaaron: **original context:** https://github.com/Expensify/Expensify/issues/218742

src/pages/LogInWithShortLivedAuthTokenPage.js
```diff
@@ -107,6 +124,7 @@ LogInWithShortLivedAuthTokenPage.defaultProps = defaultProps;
 export default compose(
     withLocalize,
     withOnyx({
-        isTokenValid: {key: ONYXKEYS.IS_TOKEN_VALID},
```
marcaaron, line 110: Just curious, but what was it doing before and why are we removing it now?

src/components/DeeplinkWrapper/index.website.js
tgolen: Please update the code comment to explain the "why" and not just "what" the code is doing. Why do we need to clear out the previous one before fetching a new one?

tgolen: I think that's a perfect explanation 👍 Here is how I would write it:

> Since there is no way to know if a previous short-lived authToken is still valid, any previous short-lived authToken must be cleared out and a new one must be fetched so that the popup window will only open when we know the short-lived authToken is valid.

src/pages/LogInWithShortLivedAuthTokenPage.js
marcaaron: I'm sorry I don't think I fully understood the previous comments. I don't think the desktop app needs to download the js bundle because it's an electron app and should be bundled with the app (IIRC).

>if the network is disconnected at this time

If they are disconnected then how can they access the page at all 🤔 

>Without this condition, the session expired ui will be displayed, if we think this page is also ok, then we can remove this condition.

I think it's fine but the message about the session being expired seems wrong?

tgolen:
Yeah, I'd like to get this merged. Can you please complete the PR description and checklist? It'd also be nice to have a C+ review/test this.

tgolen (dismissed):

tgolen:
bump @arosiclair for review and checklist https://github.com/Expensify/App/pull/17452#issuecomment-1508218684

tgolen:
@arosiclair sorry about the bump, I'm still confused on if we are wanting to merge this or not. I'm going to place a HOLD on it until we decide.

src/libs/actions/Session/index.js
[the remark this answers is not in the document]
arosiclair: Yeah it looks like this should just be an `API.read` now. We're just listening for the short lived token to change in Onyx instead of blocking on the request.

marcaaron:
What's the status of this PR? I'm thinking maybe we can create an issue for this to go over the general the plan then treat it like a normal "Bug" issue so that it doesn't fall off our radar? Does that sound OK to you @tgolen ?

tgolen:
I thought the plan was to update this PR with all the necessary fixes. There was some discussion about it over here: https://github.com/Expensify/App/issues/17391#issuecomment-1520268938

That's a good reminder that this can be taken off HOLD (but it still needs updated before it can be reviewed).

marcaaron:
Ah hmm, first time seeing that comment. If there's a plan let's put it [here](https://github.com/Expensify/App/issues/17987) to get on the same page?

src/ROUTES.js
```diff
@@ -103,7 +103,7 @@ export default {
     getReportDetailsRoute: reportID => `r/${reportID}/details`,
     REPORT_SETTINGS: 'r/:reportID/settings',
     getReportSettingsRoute: reportID => `r/${reportID}/settings`,
-    TRANSITION_FROM_OLD_DOT: 'transition',
+    TRANSITION: 'transition',
```
tgolen, line 106: I think this was renamed because this route is used for two things now:
1. OldDot --> NewDot
2. NewDot (web) --> NewDot (desktop)

so the name "transition from olddot" didn't make sense anymore. I think `TRANSITION` is a bit ambiguous and could mean many different things which doesn't make it very intuitive.

How about calling it `TRANSITION_BETWEEN_APPS`?

tgolen (changes requested):

tgolen (dismissed):

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -21,68 +21,76 @@ const propTypes = {
...
 class DeeplinkWrapper extends PureComponent {
     constructor(props) {
         super(props);
-
-        this.state = {
-            appInstallationCheckStatus: (this.isMacOSWeb() && CONFIG.ENVIRONMENT !== CONST.ENVIRONMENT.DEV)
-                ? CONST.DESKTOP_DEEPLINK_APP_STATE.CHECKING : CONST.DESKTOP_DEEPLINK_APP_STATE.NOT_INSTALLED,
-        };
-        this.focused = true;
+        this.hasPopupBeenOpenedBefore = !this.isMacOSWeb() || CONFIG.ENVIRONMENT === CONST.ENVIRONMENT.DEV;
```
marcaaron, line 41: I'm not quite making the connection between being on mac os web or dev and whether the popup has been opened before.

Could we add a brief comment here to explain why these platform and environment checks are necessary?

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -21,68 +21,76 @@ const propTypes = {
...
     }
 
     componentDidMount() {
-        if (!this.isMacOSWeb() || CONFIG.ENVIRONMENT === CONST.ENVIRONMENT.DEV) {
+        if (this.hasPopupBeenOpenedBefore) {
```
marcaaron, line 45: Once this is set to `true` then the `DeepLinkWrapper` popup won't ever happen again even when signing out?

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -21,68 +21,76 @@ const propTypes = {
...
             return;
         }
 
-        window.addEventListener('blur', () => {
-            this.focused = false;
-        });
+        // Since there is no way to know if a previous short-lived authToken is still valid,
+        // any previous short-lived authToken must be cleared out and a new one must be fetched
+        // so that the popup window will only open when we know the short-lived authToken is valid.
+        Session.removeShortLivedAuthToken();
```
marcaaron, line 52: I don't really get this. Why don't we just not use it when the component mounts? Why does it need to be "cleared out"? It will get replaced with a new token and we will only ever check it once because of the `this.hasPopupBeenOpenedBefore` flag.

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -21,68 +21,76 @@ const propTypes = {
...
 
-        const expensifyUrl = new URL(CONFIG.EXPENSIFY.NEW_EXPENSIFY_URL);
-        const params = new URLSearchParams();
-        params.set('exitTo', `${window.location.pathname}${window.location.search}${window.location.hash}`);
+        // If the current link is transition from oldDot, we should wait for it to finish,
+        // because we should pass the final path to the desktop app instead of the transition path.
```
marcaaron, line 55: Can we update this comment to explain what it means for us to "wait for it to finish"? It isn't obvious to me what that means or why we need to do that.

>we should pass the final path to the desktop app instead of the transition path

I'm pretty confused by this statement too. I don't see anything getting 'passed' here.

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -21,68 +21,76 @@ const propTypes = {
...
+        if (Str.startsWith(window.location.pathname, Str.normalizeUrl(ROUTES.TRANSITION_BETWEEN_APPS))) {
+            Session.getShortLivedAuthTokenAfterTransition();
+            return;
+        }
         if (!this.props.session.authToken) {
-            const expensifyDeeplinkUrl = `${CONST.DEEPLINK_BASE_URL}${expensifyUrl.host}/transition?${params.toString()}`;
-            this.openRouteInDesktopApp(expensifyDeeplinkUrl);
+            this.openRouteInDesktopApp();
```
marcaaron, line 61: Should this be higher up in the order of operations? 

I'm confused because it looks like we early return here if we don't have an `authToken`. Which makes sense because we "get one" on the next line. But the previous line is trying to "get" a short lived token as well.

src/components/DeeplinkWrapper/index.website.js
marcaaron: Can you remind me what errors we were handling and why they are not necessary anymore? I looked at the linked PR but I'm not seeing anything about errors or error handling (maybe I missed it).

src/components/DeeplinkWrapper/index.website.js
```diff
@@ -21,68 +21,76 @@ const propTypes = {
...
 
-    openRouteInDesktopApp(expensifyDeeplinkUrl) {
-        this.updateAppInstallationCheckStatus();
+    openRouteInDesktopApp() {
+        if (this.hasPopupBeenOpenedBefore) {
+            return;
+        }
+
+        // Once the popup has appeared after the page is loaded, it does not need to be displayed again.
+        // e.g. When a user refresh one tab, the popup should not appear again in other tabs.
```
marcaaron, line 82: I don't understand this comment. Why are tabs relevant in this context? Why would refreshing a tab make popups appear in other tabs?

src/libs/Navigation/Navigation.js
```diff
@@ -313,6 +318,14 @@ function drawerGoBack(backRoute) {
     navigate(backRoute);
 }
 
+function isTransitionEnd() {
```
marcaaron, line 321: needs a JS doc

src/libs/Navigation/Navigation.js
```diff
@@ -313,6 +318,14 @@ function drawerGoBack(backRoute) {
...
+    return transitionIsEndPromise;
+}
+
+function setIsTransitionEnd() {
```
marcaaron, line 325: also needs a JS doc

src/libs/actions/Session/index.js
```diff
@@ -266,29 +261,13 @@ function signInWithShortLivedAuthToken(email, authToken) {
...
-        {authToken, oldPartnerUserID},
-        {optimisticData, successData, failureData},
-        null,
-    ).then((response) => {
-        if (response) {
-            return;
-        }
-        Navigation.navigate();
-    });
+    API.read('SignInWithShortLivedAuthToken', {authToken, oldPartnerUserID, shouldRetry: false}, {optimisticData, successData, failureData});
```
marcaaron, line 270: I don't think we need the `shouldRetry` here do we? Read requests are not retryable by default IIRC and we have an open issue to remove this param entirely.

marcaaron (dismissed):
I appreciate the efforts to improve this. But have to admit I'm pretty unsatisfied still with the general readability of this code. It feels like it's below the standards of what we should be shooting for.

That said, I'm struggling to make any great suggestions as I find a lot of it just confusing in general and unclear. 

Sounds like maybe we will have another opportunity to improve it further in the Apple/Google SSO projects. So maybe we can improve some of the documentation then.

src/ROUTES.js
tgolen: @marcaaron Not sure if you saw this or not, but you commented about the same thing and this was my reasoning for the suggestion.

marcaaron: Thanks, maybe you can help me understand this because I'm really struggling to see it.

How does this route allow us to go from NewDot web to NewDot desktop?

And if we are using it for that purpose - maybe we should separate these concerns a bit and using some more specific "components" or "routes" that have a more specific purpose?

tgolen: It happens in this:

```
        if (Str.startsWith(window.location.pathname, Str.normalizeUrl(ROUTES.TRANSITION_BETWEEN_APPS))) {
            App.beginDeepLinkRedirectAfterTransition();
            return;
        }
```

Then `App.beginDeepLinkRedirectAfterTransition();` opens up desktop with a deeplink. My understanding is that there are places where OldDot deeplinks directly to the Desktop app (it seems like this might have started in https://github.com/Expensify/App/issues/16083 or earlier), and the only possible way for that to happen is to use NewDot web as a middle-man.

There is a fairly long discussion about this in the issue: https://github.com/Expensify/App/issues/17391#issuecomment-1515999087

tgolen: > maybe we should separate these concerns a bit and using some more specific "components" or "routes" that have a more specific purpose?

I believe this was also mentioned in the discussion a little bit, but there wasn't a strong push for it. I honestly wouldn't mind separating them out either, but it could require changing some things on OldDot (not too big a deal though probably).

src/pages/LogInWithShortLivedAuthTokenPage.js
```diff
@@ -35,12 +37,23 @@ const propTypes = {
...
-    /** Whether the short lived auth token is valid */
-    isTokenValid: PropTypes.bool,
+    /** The details about the account that the user is signing in with */
+    account: PropTypes.shape({
+        /** Whether a sign is loading */
+        isLoading: PropTypes.bool,
+    }),
+
+    /** Props to detect online status */
+    network: networkPropTypes,
```
tgolen, line 47: It looks like everywhere else in the app it uses `.isRequired` for this prop. Let's do that here.

src/libs/Browser/index.web.js
```diff
@@ -50,8 +51,42 @@ function isMobileSafari() {
     return /iP(ad|od|hone)/i.test(userAgent) && /WebKit/i.test(userAgent) && !(/(CriOS|FxiOS|OPiOS|mercury)/i.test(userAgent));
 }
 
+function openRouteInDesktopApp(shortLivedAuthToken = '', email = '') {
+    const params = new URLSearchParams();
```
tgolen, line 55: Can you please add a code comment here explaining _why_ the params have to be passed this way?

tgolen (changes requested):
OK, taking another look at this since Marc was pulled onto other things. I think there are still a few things which can be cleaned up here to make it a little easier to understand what's going on. Thanks for your continued work on this!

src/libs/Browser/index.web.js
tgolen: Oh, sorry... I am just thinking of a comment like this:

> The session information needs to be passed to the Desktop app, and the only way to do that is by using query params. There is no other way to transfer the data.

src/libs/Browser/index.web.js
```diff
@@ -50,8 +51,42 @@ function isMobileSafari() {
...
+    if (email && shortLivedAuthToken) {
+        params.set('email', email);
+        params.set('shortLivedAuthToken', shortLivedAuthToken);
+    }
+    const expensifyUrl = new URL(CONFIG.EXPENSIFY.NEW_EXPENSIFY_URL);
+    const expensifyDeeplinkUrl = `${CONST.DEEPLINK_BASE_URL}${expensifyUrl.host}/transition?${params.toString()}`;
+
+    // This check is necessary for Safari, otherwise, if the user
+    // does NOT have the Expensify desktop app installed, it's gonna
+    // show an error in the page saying that the address is invalid
```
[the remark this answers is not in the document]
tgolen: OK, interesting. Like, what lead you to try an iframe in the first place?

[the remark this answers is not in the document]
tgolen: It seems like someone had a similar problem here and there was a slightly different way of solving it. Could we try it out? https://stackoverflow.com/questions/18986200/suppress-safari-cant-open-the-page-because-the-address-is-invalid-custom-app-l

Also, do you know if the link URL matches this scheme? https://github.com/Expensify/App/blob/0b35d7c010157841617867ac3924699168cd9d27/ios/NewExpensify/Info.plist#L28

It looks like there are a few more tips in https://csiandal.medium.com/safari-cannot-open-the-page-because-the-address-is-invalid-2147d8963f89

src/libs/actions/App.js
```diff
@@ -184,6 +185,23 @@ function reconnectApp() {
     });
 }
 
+/**
+ * This promise is used so that deeplink component know when a transition is end.
+ * This is necessary because we want to begin deeplink redirection after the transition is end.
+ */
+let resolveTransitionIsEndPromise;
```
tgolen, line 192: I like seeing it here in `App.js` a lot better, yeah. I think that makes more sense. My last suggestion would be to once again rename this (and the other related methods and variables) to be more a little more intuitive.

What do you think about `waitForSignonTransitionToFinish()` and `endSignonTransition()`

src/libs/actions/Policy.js
```diff
@@ -952,7 +952,7 @@ function createWorkspace(ownerEmail = '', makeMeAdmin = false, policyName = '',
         ],
     });
 
-    Navigation.isNavigationReady()
+    return Navigation.isNavigationReady()
```
tgolen, line 955: It's not good to return a promise from an action method (our docs state this and it can lead to bad code chaining if someone else sees the pattern and copies it).

I was thinking of how we could stop doing this, and here is what I think:
1. Create a new method in `App.js` called `CreateWorkspaceAndNavigateToIt()`
2. Generate the optimistic policyID in that new method and pass it to `createWorkspace()` as a parameter
3. Move this navigation logic into that new method and only have the API call here

It feels to me like that would clean up a lot of this awkwardness that we're trying to work around and it keeps all the navigation stuff in `App.js` where it should probably belong (or I could even see using a `PolicyCreationUtils.js` file at some point)

tgolen (changes requested):
Looks like there is a conflict in `App.js`. 

Thanks for your dedicated work on this! I think we are getting closer, and I had a few more suggestions today as I was looking at the new solution.

src/libs/actions/Policy.js
tgolen: > Eh, There are two other places that also need navigation logic.

Well, I was thinking those would just be updated to call `CreateWorkspaceAndNavigateToIt()`. Doesn't that still work like it's supposed to?

I think your alternate approach works too, but I still don't really like seeing the navigation happen inside of `createWorkspace()`. IMO, the only thing that method should be doing is creating a workspace. It should not be doing any navigation, because that is business logic that it is not concerned with.

src/libs/actions/App.js
```diff
@@ -195,6 +196,43 @@ function reconnectApp() {
...
+function waitForSignOnTransitionToFinish() {
+    return signOnTransitionToFinishPromise;
+}
+
+function endSignOnTransition() {
+    return resolveSignOnTransitionToFinishPromise();
+}
+
+/**
+ * create a new workspace and navigate to it
```
tgolen, line 217:
```suggestion
 * Create a new workspace and navigate to it
```

tgolen (changes requested):
Thanks for those changes. I think this is ready to merge, but I noticed this one last nit-pick. 

I think we can leave the iframe stuff in there, but I'd love to get rid of it at some point if it's not really necessary. Seems like just a bad hack, not your fault though!

arosiclair:
Sounds like this is ready for the review checklist?

tgolen (dismissed):

tgolen (dismissed):

tgolen:
Yep, last thing here is to get the C+ review and checklist. Thanks for doing this one @0xmiroslav

tgolen (dismissed):

tgolen:
I'm looking into that error on the server side. I see it happening for many people

tgolen:
The reason the error happens is because the `authToken` that's used to get the short-live-auth-token has expired. So... something must not be refreshing the `authToken` properly. We're still looking into it

cc @yuwenmemon @robertjchen @NikkiWines @johnmlee101

tgolen:
There are multiple layers of our code which the error comes from:

- Auth (our database layer) checks the authToken and throws a 407
- PHP (the web API layer) catches the 407 and returns a 666 to the frontend

I'm still trying to get guidance from @NikkiWines and @johnmlee101 about this and bumping them today (including this bump 😁 )

tgolen:
OK, so the current running theory is this:

- The `authToken` being sent to the API command 'OpenOldDotLink' is the authToken that's expired (I think this is kind of a "no duh" situation)

I think we need to figure out why that authToken is expired and where it's coming from

tgolen:
I cannot reproduce the 666 error either. So, it looks like there are a few things left to do here yet:

1. Fix the lint error that's failing
2. Complete the changes mentioned in https://github.com/Expensify/App/pull/17452/files/d9a3cbc67f61d65823bd6384976e090fa97a0404#r1183808816 which were never fully implemented

tgolen (approved):
> Do you prefer to have CreateWorkspaceAndNavigateToIt return a Promise? 😂

No, it shouldn't return a promise.

1. Action methods shouldn't ever return promises (there are SUPER rare exceptions)
2. The promise isn't used anywhere
3. It just adds something to the code which the next engineer will see and think it's OK to return a promise in the method they are creating

src/libs/actions/App.js
```diff
@@ -213,6 +214,45 @@ function reconnectApp() {
...
+ * @param {String} [ownerEmail] Optional, the email of the account to make the owner of the policy
+ * @param {Boolean} [makeMeAdmin] Optional, leave the calling account as an admin on the policy
+ * @param {String} [policyName] Optional, custom policy name we will use for created workspace
+ * @param {Boolean} [transitionFromOldDot] Optional, if the user is transitioning from old dot
+ * @returns {Promise}
+ */
+function createWorkspaceAndNavigateToIt(ownerEmail = '', makeMeAdmin = false, policyName = '', transitionFromOldDot = false) {
+    const policyID = Policy.generatePolicyID();
+    Policy.createWorkspace(ownerEmail, makeMeAdmin, policyName, policyID);
+    return Navigation.isNavigationReady()
```
tgolen, line 246:
```suggestion
 */
function createWorkspaceAndNavigateToIt(ownerEmail = '', makeMeAdmin = false, policyName = '', transitionFromOldDot = false) {
    const policyID = Policy.generatePolicyID();
    Policy.createWorkspace(ownerEmail, makeMeAdmin, policyName, policyID);
    Navigation.isNavigationReady()
```

tgolen (changes requested):

tgolen (dismissed):

tgolen (dismissed):

src/pages/signin/LoginForm.js
```diff
@@ -75,6 +78,9 @@ class LoginForm extends React.Component {
             login: '',
         };
 
+        if (!this.props.isFocused) {
```
tgolen, line 81: Please add a code comment here for why this logic is here.

tgolen (changes requested):
For skipping the reauthentication logic, I think you can specify `shouldRetry: false` in the data.

https://github.com/Expensify/App/blob/cd32eccc8768262a18beafb5ec7c0e1be2249111/src/libs/Middleware/Reauthentication.js#L58-L61

tgolen:
Of all the alternatives, I think adding a `skipReauthentication` param is probably the best idea. Just be very verbose with code comments about why it's being added. This solution is better than:

- if the command of the request is SignInWithShortLivedAuthToken, return early. (because `skipReauthentication` can be used on any API request)
- Is it possible to consider adding a shouldReauth? (this is kinda the same thing but a slight different name)
- should we temporarily revert to makeRequestWithSideEffects (because this is going backwards and we want to avoid this)
- the server can return other error codes instead of 407. (because we want to keep the server errors consistent)

tgolen (dismissed):

s77rt:
I will take it

s77rt:
Looks we may already have another solution for https://github.com/Expensify/App/issues/17724 ([comment](https://github.com/Expensify/App/issues/17724#issuecomment-1569469641)).
___

@ntdiary Can't we add the "show loader if route is X" logic on top of the current PR code?

s77rt:
@ntdiary I guess the plan is to wait for https://github.com/Expensify/App/pull/19819 and revert https://github.com/Expensify/App/pull/19125. Is that correct?

Also can you please confirm that the issue https://github.com/Expensify/App/issues/17724 will no longer be reproducible (create a new account)

s77rt:
Yeah I got assigned by accident :sweat_smile:. @0xmiroslav is the primary C+ here, I'm just here for "extra" review. Looks like the PR is almost done anyway.

pecanoro:
@0xmiroslav Let us know when you had a chance to take a look!

pecanoro:
@0xmiroslav Friendly bump to review the last changes asap!

0xmiroslav (approved):

src/libs/Browser/index.web.js
```diff
@@ -60,4 +61,43 @@ function isMobileChrome() {
...
-export {getBrowser, isMobile, isMobileSafari, isMobileChrome};
+/**
+ * The session information needs to be passed to the Desktop app, and the only way to do that is by using query params. There is no other way to transfer the data.
+ * @param {String} shortLivedAuthToken
+ * @param {String} email
+ */
+function openRouteInDesktopApp(shortLivedAuthToken = '', email = '') {
+    const params = new URLSearchParams();
+    params.set('exitTo', `${window.location.pathname}${window.location.search}${window.location.hash}`);
+    if (email && shortLivedAuthToken) {
```
pecanoro, line 72: So this means that either both of them are passed or none of them are?

pecanoro:
@marcaaron @arosiclair In general, it looks good to me but I am not great with the transitioning logic so it would be great if you guys can give it a review as well!

arosiclair (dismissed):
LGTM as well. All yours @marcaaron

pecanoro:
Merge conflicts!

marcaaron:
Sorry, I gotta pass on the review for this one as I am scrambling to make up other work and really don't want this to be blocked on my review. I tried to unassign myself a while back, but looks like it's not possible after you've left a comment. 🙃 Anyways, I trust you guys got this.

arosiclair (approved):

jasperhuangg:
Tests well! Gonna check this off :)

src/libs/actions/App.js
```diff
@@ -214,23 +261,25 @@ function setUpPoliciesAndNavigate(session) {
...
-            Navigation.goBack();
-            Navigation.navigate(exitTo);
-        });
+        Navigation.isNavigationReady()
+            .then(() => {
+                // We must call goBack() to remove the /transition route from history
+                Navigation.goBack();
+                Navigation.navigate(exitTo);
+            })
+            .then(endSignOnTransition);
```
Pujan92, line 282: We also need to call `endSignOnTransition` when exitTo param isn't available to avoid [issue](https://github.com/Expensify/App/issues/52450)

Issue #20433, opened 2023-06-08
Cache node_modules to speed up CI/CD


roryabraham:
### Problem

Our `setupNode` action runs a lot and often takes 3-5 minutes, which is pretty slow.

### Solution

While the `actions/setup-node` action caches npm / node itself, it doesn't cache the actual node_modules. This means that we're doing a clean install every time, on slow CI machines, and it takes a while. Instead, let's cache the node_modules directory and only run `npm ci` if needed.

roryabraham:
I'm going to handle this one myself @zanyrenney so I'm unassigning you.

roryabraham:
I've been testing this out over [here](https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/workflows/test.yml)

roryabraham:
Got PR ready for review: https://github.com/Expensify/App/pull/20436

roryabraham:
This is live and seems to work well

PR #20436, opened 2023-06-08
[No QA] Cache node modules in GitHub Actions

roryabraham:
### Details
This PR caches node_modules in our GitHub Actions CI according to the following logic:

- if `package-lock.json` hasn't changed, and a cache is restored, skip installing `node_modules` because it's not necessary.
- if `package-lock.json` has changed, or a cache is not restored, run `npm ci`
- If a workflow has been manually retried and is not in its first attempt, no caching will occur.

### Fixed Issues
$ https://github.com/Expensify/App/issues/20433

### Tests
This workflow was [tested in Public-Test-Repo](https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/workflows/test.yml):

- cache miss: https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5206794080/jobs/9393712554
- cache hit (first run): https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5206799767/jobs/9393722481
- caching skipped (second run): https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5206799767/jobs/9393730879
- package-lock.json updated, cache miss: https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5206813491/jobs/9393747848

### Offline tests
n/a

### QA Steps
None.

### PR Author Checklist
    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

</details>

<details>

</details>

<details>

</details>

<details>
<summary>Desktop</summary>

</details>

<details>
<summary>iOS</summary>

</details>

<details>
<summary>Android</summary>

</details>


roryabraham:
Opting to skip C+ review for this PR

.github/actions/composite/setupNode/action.yml
```diff
@@ -9,7 +9,15 @@ runs:
...
 
+    - id: cache-node-modules
+      if: github.run_attempt == '1'
+      uses: actions/cache@v3
+      with:
+        path: node_modules
+        key: ${{ runner.os }}-node-modules-${{ hashFiles('package-lock.json') }}
+
     - name: Install node packages
+      if: github.run_attempt != '1' || steps.cache-node-modules.outputs.cache-hit != 'true'
```
roryabraham, line 20: Note: I tried using `fromJSON` here, but it doesn't work because `steps.cache-node-modules.outputs.cache-hit` is either `true` or `null`, not `true` or `false` as the documentation suggests.

.github/actions/composite/setupNode/action.yml
```diff
@@ -9,7 +9,15 @@ runs:
         node-version-file: '.nvmrc'
         cache: npm
 
+    - id: cache-node-modules
+      if: github.run_attempt == '1'
+      uses: actions/cache@v3
+      with:
+        path: node_modules
+        key: ${{ runner.os }}-node-modules-${{ hashFiles('package-lock.json') }}
```
roryabraham, line 17: Note: no partial matches are allowed, because `npm ci` will delete `node_modules` if it's present. It's either a full match and `npm ci` is skipped entirely, or it's a cache miss and we run `npm ci`. We do not do `npm i` because it can change `package-lock.json` and is considered unsafe for CI environments

.github/actions/composite/setupNode/action.yml
amyevans: 👍 I see this matches the example in the docs [here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action)

amyevans (dismissed):

AndrewGable:
Reading the docs: https://github.com/actions/setup-node#caching-global-packages-data

Makes me think it should be working by default, why is it not? We could maybe cache even more using [`cache-dependency-path`](https://github.com/actions/setup-node#caching-global-packages-data) but interested to hear why the built in isn't working

.github/actions/composite/setupNode/action.yml
```diff
@@ -9,7 +9,15 @@ runs:
         node-version-file: '.nvmrc'
         cache: npm
```
AndrewGable, line 10: Maybe we should add:

```
    cache-dependency-path: |
      package-lock.json
      desktop/package-lock.json
```

roryabraham:
Yeah, the docs are confusing for sure. They say this:

> The action has a built-in functionality for caching and restoring dependencies

But then soon after says this:

> Note: The action does not cache node_modules

---

So I dug into it and discovered what they mean by `setup-node` having built-in caching... Check out the verbose logs from [this workflow run](https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5213976292/jobs/9409675988). Basically `setup-node` caches the dependencies of `npm` and `node` itself, so it's not like you have to completely install whatever node version you're using from scratch every time you install node on a runner. A lot of the files are actually C files and binaries, and the cache is in `.npm`. It does not cache any the dependencies you define in your `package.json`. Those live in `node_modules` and have to be cached separately.

roryabraham:
Also, fwiw `npm ci` will first delete your node_modules before running an install of all your dependencies according to their exact version in the package-lock.json. So the examples shown in that documentation definitely do not cache `node_modules`.

roryabraham:
> Caches are only stored per workflow run? (Not even sure this is the right term) Basically there is no way that two PRs can share the cache of mains node_modules?

Where do you see that? Here's a workflow run I triggered about 20 minutes ago: https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5213976292

And without changing the package-lock.json, I just triggered another: https://github.com/Andrew-Test-Org/Public-Test-Repo/actions/runs/5214138966/jobs/9410020020 and you can see that the cache was restores and we did not run `npm ci`.

.github/actions/composite/setupNode/action.yml
```diff
@@ -9,7 +9,15 @@ runs:
...
 
+    - id: cache-node-modules
+      if: github.run_attempt == '1'
```
AndrewGable, line 13: [Question 1 and 2 here](https://github.com/Expensify/App/pull/20436#issuecomment-1583074039) is me wondering if we can just remove this line and the `github.run_attempt != '1'` below.

roryabraham:
> Basically there is no way that two PRs can share the cache of mains node_modules?

Two PRs can share the cache of mains node_modules unless they updated `package-lock.json`. If a PR does update `package-lock.json`, then a new cache will be computed. If there are future commits pushed to that same PR without changing `package-lock.json` further, then the new cache computed when that PR was created will be used.

> If yes, does cache-ing take much time? I would think we'd want to cache even if it's not the first run if there are changes to package.json later in the PR?

The only time the `run_attempt` will come into play is if a workflow is manually retried. Separate commits pushed to a PR will trigger separate workflow runs, where the `run_attempt` will start at `1` for each.

.github/actions/composite/setupNode/action.yml
roryabraham: The purpose of this code is to make it so that we will not use the cache if a workflow is ever _manually retried_.

My reasoning is because if we're manually retrying a workflow, then something must have gone wrong, and doing a clean install is a good idea because it removes the cache from the equation. I could be convinced otherwise but it seems like a good way to make it so deployers have to think about caches less

roryabraham:
> Can we cache node_modules in desktop?

Good question. I can look into this

.github/actions/composite/setupNode/action.yml
AndrewGable: Ah I see, I think personally I'd rather have it always on and then just know that we can delete them via https://github.com/Expensify/App/actions/caches

(This is how it works in Travis too)

However, I do not feel strongly enough to block. If we do keep this code, let's add a comment explaining this is how it works.

roryabraham:
Added caching for desktop node_modules. Also made it so that we run `patch-package` no matter what – new or changed patch files wouldn't have been caught by hashing `package-lock.json`. `patch-package` is generally really fast and we currently don't have a ton of patches so running it every time shouldn't be a problem.

.github/actions/composite/setupNode/action.yml
roryabraham: Added this as well

.github/actions/composite/setupNode/action.yml
AndrewGable: Thoughts on this one @roryabraham ?

roryabraham: Ok, I removed the `run_attempt` logic from here and from the pod caching as well

AndrewGable (approved):

roryabraham:
CPing to test the cocoapods caching change

roryabraham:
Some hiccups here, but was able to fix them in https://github.com/Expensify/App/pull/20488 and we seem to be good-to-go here.

Issue #20623, opened 2023-06-12
Define TypeScript guidelines


hayata-suenaga:
Part of [TS migration](https://github.com/Expensify/App/issues/16665) issues.

- Create the first version of the guideline in `app/contributingGuides/TYPESCRIPT_STYLE.md`.
- A rough overview of the guideline is in [the design doc](https://docs.google.com/document/d/1LNcgYeYyovQ88gNjLIgBqBReLtbLAbxeAbna9QKxeAQ/edit#bookmark=id.pz75s9ij30u2).
- Add succinct code examples as needed.
- Create sub-files if necessary to divide the guideline into sections.
- While writing the guideline, if you encounter any unclear points that should be discussed, post them on the #typescript-new-expensify channel.

hayata-suenaga:
The new guideline is developed in [this google doc](https://docs.google.com/document/d/1CYivl5lGGJ1TgwPAmshJ-pIAClwszIVBYg7hpQCS6to/edit#heading=h.ffi45lc01b3a)

fabioh8010:
@hayata-suenaga The [PR](https://github.com/Expensify/App/pull/21050) is ready to review!

hayata-suenaga:
@fabioh8010 thank you so much for this PR 🎉 I'll take a look at this tomorrow

hayata-suenaga:
the PR is still being worked on and review is still going on

hayata-suenaga:
The PR is actively reviewed and worked on 👍

hayata-suenaga:
https://expensify.slack.com/archives/C03TQ48KC/p1689011336554509

hayata-suenaga:
The PR has been merged. closing this

PR #20179, opened 2023-06-05
Setup App for TypeScript

blazejkustra:
### Details

- TSConfig, Eslint, webpack, babel configured
- Github actions for type checking
- Installed all typescript dependencies
- Removed unused “save” dependency
- Install dependencies that are used but weren’t installed before

### Related Issue

$ [#20626](https://github.com/Expensify/App/issues/20626)

### Tests

This touches the project globally, there are no files to reasonably test individually. We should run basic smoke tests here to see that the app runs and works as expected. 

Check that all scripts listed in `package.json` works as expected.

- [x] Verify that no errors appear in the JS console

### Offline tests

n/a

### QA Steps

None. This should be covered by regression testing.

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
[...]

PR #21050, opened 2023-06-19
TypeScript Guidelines

fabioh8010:
### Details

This PR aims to establish the TypeScript guidelines that are going to be used in the project.

### Fixed Issues

$ https://github.com/Expensify/App/issues/20623
PROPOSAL: -

### Tests

Not applicable.

### Offline tests

Not applicable.

### QA Steps

Not applicable.

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos

Not applicable.


contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+function processData(inputData: string) {}
+class GoodExample {
+  doWork() {}
+  workDone: boolean;
+}
+import * as someModule from "some-module";
+
+```
+
+Use **CONSTANT_CASE** for global constant values to distinguish them from other variables and emphasize their fixed state.
```
neil-marcellini, line 171: NAB: Aka screaming snake case 🤣...jk

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+
+// Good
+import { ValueOf } from 'type-fest';
+
+const COLORS = {
+    Red: 'red',
+    Green: 'green',
+    Blue: 'blue',
+} as const;
+type Colors = ValueOf<typeof COLORS]>
```
neil-marcellini, line 326: That square bracket looks lonely or like it shouldn't be there 🤔 `<typeof COLORS]>`

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+```
+
+If using enum is necessary, utilize string enum. Do not use numeric and heterogenous enums. This can help prevent potential bugs at runtime.
```
neil-marcellini, line 329: > This can help prevent potential bugs at runtime.

How, why? I'm curious.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+enum Colors {
+  Red = 'red',
+  Green = 'green',
+  Blue = 'blue',
+}
+```
+
+## Usage of type `any` and `unknown`
+
+The `any` type allows assignment to all types and dereference of any property, which is undesirable and should be avoided. Instead, in most cases use the `unknown` type which expresses a similar concept and is much safer as it requires narrowing the type before using it.
```
neil-marcellini, line 354: I think this reads a little better. `Instead, and in most cases, use...`

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+
+// Good
+// As we are just going to persist the response in Onyx and
+// it would be very hard to type all possible data structures,
+// we just leave `response` as `any`.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function processAPIResponse (response: any) {
+    // your code
+}
+```
```
neil-marcellini, line 390: I feel like this section should go right under the "## Usage of type `any` and `unknown`" heading, because it's the most important thing to know in this section.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+
+## TypeScript Annotations
+
+### Usage of `@ts-expect-error`
+
+During migration, it is expected to find TS errors revealing underlying JS issues. Treat significant errors as separate issues linked to the TS migration issue, unless the fix is minor and verifiable (document it in the PR). Use `@ts-expect-error` for such cases.
+
+### Usage of `@ts-ignore`
+
+`@ts-ignore` should not be used. If the usage of `@ts-ignore` cannot be avoided for rare situations, explain the reason in a comment.
```
neil-marcellini, line 400: For the Typescript un-initiated, it would be nice to include a one sentence description of what each of these annotations does. Yes we could go look at the docs but a description make it easier to read this document, especially since most of use are new to Typescript.

neil-marcellini: Also bookmark 🔖 , this is where I have stopped reviewing for today.

neil-marcellini (changes requested):
Looking good so far. I added some comments and I'll review the rest tomorrow.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+if (isPolicyRoomReport(report)) {
+    // I'm guaranteed to have access to PolicyRoomReport's properties e.g. "roomId".
+    // report.roomId
+} else {
+    // If this is not a PolicyRoomReport, then I'm guaranteed that this is a PolicyExpenseChatReport, so I'll have access to it's properties e.g. "expenseChatId"
+    // report.expenseChatId
+}
+```
+
+In this example, we have the `ChatReport` type which can be two possible types of object, `PolicyRoomReport` or `PolicyExpenseChatReport`. The `isPolicyRoomReport` function will accept `report` as a parameter and will narrow the type of `report` to `PolicyRoomReport` if the condition inside it is true, and to PolicyExpenseChatReport if the condition is false.
```
thiagobrez, line 559: Super nitpick: add backticks to the last usage of PolicyExpenseChatReport to display it as code block

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+    <View>
+      <ChildComponent ref={viewRef} {...someViewProps} />
+    </View>
+  );
+};
+```
+
+### Children and render props
+
+To type children props, use `React.ReactNode`.
```
thiagobrez, line 823: Question: did we decide against `PropsWithChildren`, or should we add an example here?

I generally prefer it over `ReactNode` because it makes the type easier to read (one less property), although I'm not sure how well it works with render props

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+Remove any words that are already clear from the variable's type declaration.
+
+```ts
+// Bad
+let nameString: string;
+let holidayDateList: Array<Date>;
+
+// Good
+let name: string;
+let holidays: Array<Date>;
```
mrousavy, line 208: Is there any reason we are using `Array<Date>` instead of `Date[]`? 

* In my opinion `Date[]` makes more sense and is more widely understood as arrays are kinda like a primary type whereas `Array` could be misunderstood as a custom generic type.
* Also `Date[]` is less letters / faster to type
* `Date[5]` could be used for constant sized arrays

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+    // properties
+}
+
+// Good
+type Report = {
+    // properties
+}
+```
+
+TypeScript supports both type and interface expressions. Types can be used more broadly: to name primitives, unions, objects and any other type. As these two forms are practically equivalent, it's best to select one over the other to maintain consistency and reduce variation.
```
mrousavy, line 243: Makes sense 👍

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+function processAPIResponse(response: any) {
+    // your code
+}
+
+// Good
+// As we are just going to persist the response in Onyx and
+// it would be very hard to type all possible data structures,
+// we just leave `response` as `any`.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function processAPIResponse (response: any) {
```
mrousavy, line 387: I think this could still be unknown or at least some basic type like `Record<string, unknown>`?

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+
+// Bad
+const getUsersSummary = (users: User[], minAge: number, isActive: boolean) => {
+  return users
+    .filter(user => user.age >= minAge && user.active === isActive)
+    .map(user => `${user.name} (${user.age})`);
+};
+
+// Good
+const getUsersSummary = (users: User[], minAge: number, isActive: boolean): string[] => {
```
mrousavy, line 468: Here we have `string[]` instead of `Array<string>`. I think `string[]` is better.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+```ts
+// Bad
+const CONST = {
+    // ...
+};
+
+// Good
+const CONST = {
+    // ...
+} as const;
```
mrousavy, line 490:
```suggestion
const Something = {
    // ...
};

// Good
const Something = {
    // ...
} as const;
```

mrousavy (commented):
looks good overall!

contributingGuides/TYPESCRIPT_STYLE.md
fabioh8010: Some examples:

```ts
enum NumberEnum {
  A, // when not assigning anything, the value will be 0.
  B, // and will increment, 1.
  C, // 2...
}

enum NumberEnum {
  A = 10, // 10
  B, // 11
  C, // 12
  D = 200, // 200
  E, // now increments to 201, 202, 203...
}

enum HeterogenousEnum {
    VALUE_1 = 'string value', // "string value"
    VALUE_2 = 0, // 0
    VALUE_3, // 1
    VALUE_4 = 200, // 200
    VALUE_5, // 201
    VALUE_6 = 'string value',
    VALUE_7 = 0, // 0
    VALUE_8 // 1
}
```

Also, when using numeric enums we would have reverse mapping, for example:

```ts
enum NumberEnum {
  A = 10, // 10.
  B, // 11.
  C, // 12...
}

console.log(NumberEnum.A); // outputs 10.
console.log(NumberEnum[10]); // outputs "A".
```

There are just some examples of how dangerous using numeric/heterogenous enum can be.

contributingGuides/TYPESCRIPT_STYLE.md
fabioh8010: Yes, we discussed about this in the document and we agreed about using  T[] for most cases (literals and easier types), and leaving Array<T> for more complex cases.

I will create a section about this.

contributingGuides/TYPESCRIPT_STYLE.md
fabioh8010: I can add another example about using `Record<string, unknown>`, but I would like to leave the `any` in case someone face such exceptional case and we will have this covered in the doc.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+function Component() {
+  return (
+    <TextInput onChangeText={(newTextValue: string) => setTextValue(newTextValue)} />
+  );
+}
+
+// Good
+function Component() {
+  return (
+    <TextInput onChangeText={(newTextValue) => setTextValue(newTextValue)} />
```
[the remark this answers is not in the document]
fabioh8010: Yeah it's just to explain show the usage of parameter typing 😄

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: I don't have a strong opinion on this. Both `PropsWithChildren` and typing children with `ReactNode` are good approaches. I would leave it as is it written in the guidelines just because it is easier for non-Typescript users as you don't need to use generics in that case.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,857 @@
...
+}
+
+// Your code
+declare module 'library/path' {
+  interface LibraryComponentProps {
+    // Add or modify the typings
+    additionalProp: string;
+  }
+}
+```
```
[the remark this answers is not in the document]
blazejkustra: There is a file in Typescript setup [PR](https://github.com/Expensify/App/pull/20179/files#diff-80ab2d8524b0603b482312f0d97b5c3c829ce1843880dbe0f7bcc983971faafb) (declarations.d.ts). We could use this file when extending or correcting types @hayata-suenaga @fabioh8010. We could also rename it to `global.d.ts`.

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: It's hard to find a real case where `any` is the only option... In most cases we will be able to use other types such as unknown, Record etc.

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: Use CONST_CASE here @fabioh8010

fabioh8010:
Comments addressed!

contributingGuides/TYPESCRIPT_STYLE.md
neil-marcellini: Awesome thanks. Could you please link to some examples like these or add a brief description for why numeric and heterogenous enums are dangerous? Something like "numeric enums are dangerous because they might not increment as expected, and heterogenous enums are dangerous because the type of a given value is unclear."?

contributingGuides/TYPESCRIPT_STYLE.md
neil-marcellini: Thanks for the update, I'm continuing to review.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,886 @@
...
+### Default props
+
+As **defaultProps** are not allowed anymore in the project, use object destructuring in your `Props` in order to set default values to them. The props that have default value must also be marked as optional in your `Props` type.
+
+```tsx
+// Before
+const propTypes = {
+    requiredProp: PropTypes.string.isRequired,
+    optionalPropWithDefaultValue: PropTypes.number,
+    optionalProp: PropTypes.bool,
```
neil-marcellini, line 748: I think we don't currently allow optional props without default values, according to our style guide. Maybe I'm wrong, but why should we allow optional props without default values?

neil-marcellini (changes requested):
Looking good, I've reviewed all of it now. I have a few small changes I would like to see and a handful of questions. I'm really glad we have you guys working on this, I've learned so much already!

hayata-suenaga:
still reviewing...

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,886 @@
...
+// As we are just going to persist the response in Onyx and
+// it would be very hard to type all possible data structures,
+// we just leave `response` as `any`.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function processAPIResponse (response: any) {
+    // your code
+}
+```
+
+The `any` type allows assignment to all types and dereference of any property, which is undesirable and should be avoided. Instead, and in most cases, use the `unknown` type which expresses a similar concept and is much safer as it requires narrowing the type before using it. When you know that the type structure is a object but you don't have context about the content, use `Record<string, unknown>`.
```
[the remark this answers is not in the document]
fabioh8010: I didn't know about this one until now to be honest 😅 . I guess we have the explanation [here](https://github.com/microsoft/TypeScript/pull/23592) and [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#support-number-and-symbol-named-properties-with-keyof-and-mapped-types).

contributingGuides/TYPESCRIPT_STYLE.md
fabioh8010: Hi @neil-marcellini , could you point me out where do we have this specification in our style guides? I didn't find it.

About the optional props, the reason is because we may have a situation where the prop of that component is optional and don't need to or can't have a default value set. For example:

```tsx
import React from 'react';
import {StyleProp, Text, TextStyle} from 'react-native';

type Props = {
  // The developer can apply a custom styling to this component
  // but it doesn't make sense to have a default value, so when using
  // the prop it will be `undefined` unless set.
  style?: StyleProp<TextStyle>;

  // Required prop.
  label: string;

  // The developer can set a custom count, but we need a value
  // anyway so we are going to set `0` as a default value.
  count?: number;
};

function LabelWithBadge({style, label, count = 0}: Props) {
  return (
    <Text style={[{color: 'red'}, style]}>{`${label} - Count: ${count}`}</Text>
  );
}

function App() {
  return <LabelWithBadge label="Notifications" />;
}
```

fabioh8010:
Addressed more comments!

contributingGuides/TYPESCRIPT_STYLE.md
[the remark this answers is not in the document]
neil-marcellini: Thanks for finding it in the docs! NAB: Maybe add a sentence to clarify that the Record key can also be a number or a symbol when it's written like this, and link to the TS docs.

contributingGuides/TYPESCRIPT_STYLE.md
neil-marcellini: Ok got it thank you for the example. Yeah it's not mentioned in our App style guide. It is mentioned in the Airbnb javascript/react style guide, which was are supposed to follow, in the [props section](https://github.com/airbnb/javascript/tree/master/react#props).

> Always define explicit defaultProps for all non-required props.

In the example you showed above we would provide a default of `null` for the optional prop `style`. I think that's helpful, but it's not critical. Maybe we can get some more opinions. @hayata-suenaga @roryabraham

neil-marcellini (dismissed):
I'm happy with it! Thanks.

contributingGuides/TYPESCRIPT_STYLE.md
hayata-suenaga: @neil-marcellini

> I think we don't currently allow optional props without default values

thank you for reminding us of this 👍 

Are we discussing the necessity of providing default values for optional properties?

```ts
type MyCompProps = { style?: StyleProp<TextStyle>; };

function MyComp({ style }: MyCompProps) {...}
// vs.
function MyComp({ style = null }: MyCompProps) {...}
```

IMO, it depends. We don't have to always require default values. If a value's default value is `undefined` (or `null`), it might be redundant to explicitly assign `undefined` to it.

Or if it's easy to handle an optional prop having `undefined` value inside a component, it's also not necessary to come up with default values for these.

```tsx
type MyCompProps = { person?: { name: string } };

function MyComp({ person }: MyCompProps} {
   ...<Text>{person?.name ?? 'anonymous'}</Text>...
}
```

hayata-suenaga:
was reviwng TS declaration PR for `expensify-common`

still reviewing the guideline

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: > We don't have to always require default values. If a value's default value is undefined (or null), it might be redundant to explicitly assign undefined to it.

Totally agree 👍 

My take on this is that with Typescript we don't need to always provide default props, since we can rely on type checking to identify any missing props. However, it's important to remember that having default values can be helpful in some cases. This way, we don't need to check if a prop exists every time it's used. The goal is to find a good balance between readable code, simplicity and safety.

contributingGuides/TYPESCRIPT_STYLE.md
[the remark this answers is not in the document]
blazejkustra: Added clarification to this section @neil-marcellini @fabioh8010. 

> Maybe add a sentence to clarify that the Record key can also be a number or a symbol

Turns out that `Record<string, T>` allows numeric keys but not symbol keys: [TS playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZiEAuGAlApqATgEwDzRYCWYA5gDQwCuYA1mCAO5gB8MAvDAN4C+A3ACgBoSLAgBPALYcYAZSkAjEABsAFAEpBAhCADaAJgC6MgAyCdugERR00S8c4BGc4isn7M59tcTJDmPp8QA)

hayata-suenaga:
still reviewing... I have several items I think we should add to the guideline, but I don't have time to get down on them right now
I'll finish review this weekend 🙇

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+};
+
+type Helpers = {
+    doSomething: () => void;
+};
+
+export type { ComponentProps, Helpers };
+```
+
+In case there is no default implementation, you have to create a `index.d.ts` declaration file. When importing the module from other files, TypeScript will automatically pick up the type definitions from `index.d.ts`. Be careful when defining `index.d.ts` as declaration files aren't checked by the TypeScript compiler (with `skipLibCheck: true` - [source](https://www.typescriptlang.org/tsconfig#skipLibCheck)).
```
roryabraham, line 48: Apologies if this has been asked-and-answered, but I wonder if, instead of creating two patterns where one of them causes us to lose type-safety, we should instead enforce that every module has a default implementation (i.e: if you have `index.desktop.js`, `index.android.js`, and `index.ios.js`, then we would have `index.js` and know that it must be for web)

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+
+- `index.native.ts` - Native (iOS and Android)
+- `index.android.ts` - Android
+- `index.ios.ts` - iOS
+- `index.website.ts` - Web
+- `index.desktop.ts` - Desktop
+
+Note that `index.ts` should contain the default implementation, and only platform-specific implementations should be done in their respective files. i.e: If you have mobile-specific implementation in `index.native.ts`, then the desktop/web implementation can be contained in a shared `index.ts`.
+
+For each platform-specific module, create shared type definitions in a separate `types.ts` file and place it in the same folder. `types.ts` has to export shared types which are **compatible with all platform-specific implementations**. Declare component props, return types, and other common types in this file.
```
roryabraham, line 33: Will `tsc` validate that each platform-specific implementation is compatible with the shared types? i.e: it needs to check types not only for `index.ts`, but also `index.native.ts`. Will `tsc` handle that for us?

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+
+// index.website.ts
+export default function getPlatform(): Platform {
+  return "website";
+}
+```
+
+Refer to the React Native documentation for more information about [Platform Specific Code](https://reactnative.dev/docs/platform-specific-code).
+
+## Naming Conventions
```
roryabraham, line 127: It seems like most of this might be possible to enforce via https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md, maybe we can implement that in https://github.com/Expensify/App/pull/20179 and then remove this section. In general I don't think the guidelines would be needed if they were enforced by lint rules.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+Usage of `any` is forbidden unless there is a very good reason for it. The `any` type bypasses the TypeScript type-checking system, thus making the code less safe and more prone to bugs. Any usage of `any` type must be followed with a comment explaining why you are using it.
+
+```ts
+// Bad
+function processAPIResponse(response: any) {
+    // your code
+}
+
+// Good
+// As we are just going to persist the response in Onyx and
```
roryabraham, line 372: I think this is a bad example because we absolutely should be typing everything we put in Onyx imo

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+// we just leave `response` as `any`.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function processAPIResponse (response: any) {
+    // your code
+}
+```
+
+The `any` type allows assignment to all types and dereference of any property, which is undesirable and should be avoided. Instead, and in most cases, use the `unknown` type which expresses a similar concept and is much safer as it requires narrowing the type before using it. 
+
+When you know that the type structure is a object but you don't have context about the content, use `Record<string, unknown>`. Note that numeric keys are allowed with the `Record<string, unknown>` type. Numeric keys are implicitly converted to strings during property assignment and access. To read more about this, see [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/2/objects.html#dynamically-adding-properties).
```
roryabraham, line 383:
```suggestion
When you know that the type structure is an object, but you don't have context about the content, use `Record<string, unknown>`. Note that numeric keys are allowed with the `Record<string, unknown>` type. Numeric keys are implicitly converted to strings during property assignment and access. To read more about this, see [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/2/objects.html#dynamically-adding-properties).
```

roryabraham: I also think it might be good to include an example showing `Record<string, unknown>`, because that's not included in the example that follows.

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+const name = 'name';
+const [counter, setCounter] = useState(0);
+
+// In this case I need to specify the type because `hint` state can be either a `string` or `undefined`.
+const [hint, setHint] = useState<string | undefined>(undefined);
+```
+
+### Parameter and return types
+
+For function parameter types, depending on the clarity of the function use inferred types. Provide explicit typing when it enhances code readability.
```
roryabraham, line 454:
```suggestion
For function parameter types, depending on the clarity of the function, use inferred types. Provide explicit typing when it enhances code readability.
```

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+const items: Array<Partial<Item>> = [];
+const ids: Array<string | number> = ['some_id', 42];
+const colors: Array<ValueOf<typeof COLORS>> = [];
+```
+
+## Usage of “defaultProps”
+
+To achieve better and safer typing of components with default prop values, all usages `defaultProps` shall be removed and replaced by prop destructuring. Please head to **Typing components** section to understand how you can convert your implementation.
+
+## JSDoc annotations
```
roryabraham, line 605: Can we also enforce this with a lint rule?

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+}
+```
+
+## Typing Components
+
+### Transitioning from PropTypes to TypeScript
+
+TypeScript provides a more efficient and comprehensive solution for managing prop types in our codebase. When developing new components or updating existing ones to TypeScript, do not use PropTypes. Instead, use `types` to define the structure of your component's props.
+
+When converting a `propTypes` object to a `type`, use the following table to help and guide you in the process:
```
roryabraham, line 639: This is awesome, going to be really helpful 👏🏼

contributingGuides/TYPESCRIPT_STYLE.md
roryabraham: I agree that we don't need to always have a default since TypeScript will warn us if a prop is possibly undefined 👍🏼

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,897 @@
...
+
+### File Extensions
+
+For TypeScript files containing JSX code, use the `.tsx` file extension. For other TypeScript code use the `.ts` file extension.
+
+Files containing only type declarations should **not** use the `.d.ts` extension. The only exception is when defining a common interface for platform-specific implementations (see below).
+
+### Platform-Specific Variants
+
+In most cases, the code written for this repo should be platform-independent. In such cases, each module should have a single file, `index.ts`, which defines the module's exports. There are, however, some cases in which a feature is intrinsically tied to the underlying platform. In such cases, the following file extensions can be used to export platform-specific code from a module:
```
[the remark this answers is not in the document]
blazejkustra: Done 👍

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: Here is a long [discussion](https://docs.google.com/document/d/1YczjOMTEOdbWhsn0nEC2MzA31ME7OsNw-DYhIEJ4Hk8/edit?disco=AAAAzBdnj5M) in TS config doc regarding this topic.

TL;DR: 
We extracted two types of platform specific variants:
- With default implementation, only when it makes sense to have it (usually applies to more than one platform)
- No default implementation, each platform is different (like in `getPlatform` utility - [example](https://github.com/blazejkustra/App/commit/b936507771d97d9c71bd640d123fd7d0477b8742)). 

@fabioh8010 comment on platform specific variants without default implementation 👇 

> I think we don't have better options than using `.d.ts` files for this kind of situation. The `getPlatform` is a perfect example why I need this type of file because we don't have `index.ts` for it. An possible alternative would be to choose one of the platform-specific files and rename it to `index.ts`, but this is terrible in my opinion. Using `.d.ts` will be very beneficial for us.

blazejkustra: > instead of creating two patterns where one of them causes us to lose type-safety, we should instead enforce that every module has a default implementation

We don't really lose type-safety with declaration files as long as the name of the function is the same and types are shared from `types.ts` file and used in all implementations.

> if you have index.desktop.js, index.android.js, and index.ios.js, then we would have index.js and know that it must be for web

I have two problems with this approach:
1. It doesn't really eliminate type-safety problem that you mentioned with declaration files. What if each platform has a return type of literal strings: 'desktop' for desktop, 'ios' for iOS, 'android' for Android and 'web' for web. Rename `index.web.ts` to `index.ts`, now the return type is 'web'. When importing, type 'web' is inferred, but it should be either `string` or `'ios' | 'android' | 'desktop' | 'web'` in reality. That's why `types.ts` is needed and it resolves all these issues.
2. It is less readable what is the purpose of default implementation, I would have to check all remaining files to know that

blazejkustra: If I were to standardize one approach, I would opt for always requiring implementations for each of the four platforms and including an `index.d.ts` file with a common interface + `types.ts` with shared types.

The approach described in the doc is also fine with me 👍

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: > Will tsc validate that each platform-specific implementation is compatible with the shared types? 

Shared types (types defined in `types.ts`) will be used in each variant. Each TS file is checked individually, meaning that if a shared type isn't compatible with implementation we will get an error (error will be showed in platform specific implementations, not in `types.ts` file).

> Will tsc validate that each platform-specific implementation is compatible with other platform specific implementations?

No, from TS perspective all these files are separate modules that aren't connected, that means there might be a case where `index.ios.ts` exports function named 'x' and `index.android.ts` exports function named 'xx' and tsc won't complain.

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: I made the example more generic, let me know what do you think! It's hard to come with a good example where any is justifiable 😆

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: Changed the text and added an example.

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: I found this rule to disallow types on `@returns` and `@param` tags - [link](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/no-types.md#readme). I did not find a rule to completely ban params and returns 😒

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: Changed!

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: I think this eslint rule is a nice addition to this section. Even with the rule I would leave this part in the guideline.

> It seems like most of this might be possible to enforce via [link](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md)

I don't think it is possible to enforce CONSTANT_CASE to only work for global-level constant values. Other should be fine I believe 👍

blazejkustra: I never used this rule before. I'm not sure if we need it as I haven't seen any naming convention discrepancies in the codebase but it is nice addition and I can add it to the Typescript setup PR!

@hayata-suenaga @fabioh8010 thoughts?

contributingGuides/TYPESCRIPT_STYLE.md
hayata-suenaga: > No, from TS perspective all these files are separate modules that aren't connected, that means there might be a case where index.ios.ts exports function named 'x' and index.android.ts exports function named 'xx' and tsc won't complain.

@roryabraham This is the reason we have to define shared types in `types.ts` and make sure you export them in each platform-specific file.

contributingGuides/TYPESCRIPT_STYLE.md
hayata-suenaga: @roryabraham, great find! @blazejkustra do you think you can figure out options for this rule to make the rule conform  to our naming standard?

I think it's worth writing down even short sentences in the guideline even lint rules enforce some aspects of the guideline.

otherwise, we also don't need section for `Enum` or `any` or the usage of `@ts-ignore`

contributingGuides/TYPESCRIPT_STYLE.md
hayata-suenaga: @blazejkustra thank you for coming up with examples for `any` usage. I know coming with examples is very hard 👍 

I don't want to ruin your effort, but I was wondering do we even need to put down an example where `any` is justifiable?

If I remember correctly, the approach we specified in the design doc is that whenever an engineer feels a need to go against a lint rule (`any` usage by forbidden in our ESLint rule configuration), they need to post a comment in `#expensify-open-source` and check if the situation they encounter really warrants an exception.

so instead of prescribing exceptions in the guideline, we can just say "Don't use `any`. If you think you need to, post explanation in the slack channel`

this is much simper IMO.

generally, if we have a lint rule that errors when the rule is broken, we can just say "Don't do it" in the guideline.
`ex. Don't use Enum.`

contributingGuides/TYPESCRIPT_STYLE.md
hayata-suenaga: @blazejkustra thank you for the research 🙌 

If I remember correctly, ESLint is stopping support for lining JSDoc 😨 . We can add it for now, but we should be careful not to rely on them overly.

hayata-suenaga:
I reviewed the guideline over the weekend. I like the guideline and great job @fabioh8010 🍾 

I just wanted to organize the guideline more and wanted to make it way more concise. So I created [another PR](https://github.com/Expensify/App/pull/21549) based off this guideline PR. Please review that one when you have time this week. (I'm still finalizing details and still doing proofreading for grammatical and spelling mistakes).

In the new PR, I split this guideline into two files (one contains rules that must be followed and the other one contains guides on how to accomplish certain things in TypeScript (ex. typing `forwardRef`). and of course, I added a dedicated file just for the `propTypes` to TS conversion table that @fabioh8010 created for us

Most of the stuff is just copy and past from this PP, but I drastically cut off guidelines that are already covered by ESLint rules. For those rules that errors if not followed by ESLint, I made the wording more strict ex. "Do not use `any`".

Also, I added instructions on what to do when rules in guidelines cannot be followed for certain situation (posting comments on `#expensify-open-source` Slack channel as discussed on the design doc.

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: Done! cc @fabioh8010

contributingGuides/TYPESCRIPT_STYLE.md
```diff
@@ -0,0 +1,902 @@
...
+}
+
+// index.desktop.ts
+export default function getPlatform(): Platform {
+  return "desktop";
+}
+
+// index.website.ts
+export default function getPlatform(): Platform {
+  return "website";
```
blazejkustra, line 113: Change to 'web'

contributingGuides/TYPESCRIPT_STYLE.md
fabioh8010: I agree @hayata-suenaga , it's ok for me removing the example about `any` usage as it has been exceptionally hard to come up, and not including an example kinda help discouraging its usage even more imo. cc @blazejkustra

contributingGuides/TYPESCRIPT_STYLE.md
fabioh8010: @blazejkustra I think you last commit about the eslint rule solve this discussion as well, isn't?

blazejkustra: I tried and came up with this rule (added it to the Typescript setup PR):
```ts
'@typescript-eslint/naming-convention': [
  'error',
  {
      selector: 'variable',
      format: ['camelCase', 'UPPER_CASE', 'PascalCase'],
  },
  {
      selector: ['property', 'function'],
      format: ['camelCase', 'PascalCase'],
  },
  {
      selector: ['enumMember', 'typeLike', 'typeParameter'],
      format: ['PascalCase'],
  },
  {
      selector: ['parameter', 'method'],
      format: ['camelCase'],
  },
],
```

Explanations:
- `variable`: `camelCase` for basic variables, `UPPER_CASE` for global-level constants, `PascalCase` for React Components (we need it when using `forwardRef`)
- `property`, `function`: 'camelCase' for default functions and object properties. `PascalCase` for react components and Enum like object properties.
- `enumMember`, `typeLike`, `typeParameter`: `PascalCase` for all types.
- `parameter`, `method`: `camelCase` for class methods and function parameters

This rule doesn't fully conform with our naming standards: 
- Global-level constants can be written with `camelCase` or `PascalCase`.
- Object properties can be written in `PascalCase`.
- Functions (not components) can be written in `PascalCase`.

There are more inconsistencies like this due to the fact that some selectors are used in multiple contexts. It doesn't fully conform with our naming standards, it's still helpful and eliminates some of the mistakes that could happen.

Thoughts @fabioh8010 @hayata-suenaga @roryabraham?

fabioh8010: LGTM! In this case, we should keep the conventions in the guideline as it's not possible to cover all cases with the rule.

contributingGuides/TYPESCRIPT_STYLE.md
blazejkustra: > But if we really want to standardize it, I'll opt for the approach @gedu mentioned: always use index.ts for only export purposes.

What if there is no default implementation? Do we need to implement `Button.d.ts` file then? I think files structure is less readable with this approach.

> This approach is what I see in component libraries and is what I believe is the standard.

Could you provide an example? @hayata-suenaga

hayata-suenaga: `index.ts` is usually used to export stuff in component libraries

- [`mui` Button component](https://github.com/mui/material-ui/blob/master/packages/mui-base/src/Button/index.ts)
- [`fluent-react-native` Button component (Microsoft)](https://github.com/microsoft/fluentui-react-native/blob/main/packages/components/Button/src/index.ts)

Those libraries are natively typed and also have `Button.types.ts` for `Button` component, for example.

However...

> What if there is no default implementation? Do we need to implement Button.d.ts file then?

I tested this, and no, it didn't work without default implementation (without `Button.ts`) 😢 and creating `Buttno.d.ts` and `index.ts` seems too much.

So let's go with our original plan 👍

contributingGuides/TYPESCRIPT_STYLE.md
hayata-suenaga: We should put non-typescript naming conventions inside our [existing JavaScript style guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md).

I know linting of the naming convention was made possible because we introduced TS eslint plugin, but let's keep JavaScript stuff in `STYLE.md` instead of TypeScript guideline

hayata-suenaga:
Closing this one in favor of [the new PR](https://github.com/Expensify/App/pull/21549) 🙇 

We can keep existing unresolved conversations happening here

PR #21406, opened 2023-06-23
[Fix #21022] Improve performance by reducing re-renders SidebarLinks

### LINKED ISSUE #21022: SidebarLinks unnecessary re-renders
### Problem:
When we open a report, we end up re-rendering parts of the UI that are not visible, specifically the SidebarLinks. This re-rendering of SidebarLinks is costly and negatively affects performance. Additionally, the re-rendering is unnecessary because it doesn't change anything in the UI. It only occurs because there are changes in the lastRead timestamp and the lastMentionedTime fields of the report that we just opened.

### Importance:
When we measure performance, we observe a significant improvement when we optimize SidebarLinks to avoid unnecessary re-rendering (see screenshots).

### Solution:
Certain parts of the UI require these specific fields, such as the LHNOptionRow, to be rendered as unread. However, it is not necessary to pass these fields to the entire list. Instead, we can link the OptionRows to the relevant section of the Onyx state to retrieve that information. (We need to measure if this approach of connecting each OptionRow to Onyx introduces any new performance issues.)

hannojg:
### Details

Improve the performance by reducing unnecessary re-renders of the `SidebarLinks` component.

### Fixed Issues

$ https://github.com/Expensify/app/issues/21022
PROPOSAL: https://github.com/expensify/app/issues/21022#issuecomment-1600631095

### Tests

- [ ] Open a report, verify that the highlight of the items still work
- [ ] Send a message from user A to user B and verify that the reports are resorting accordingly

### Offline tests

n/a

### QA Steps

- [ ] Open a report, verify that the highlight of the items still work
- [ ] Send a message from user A to user B and verify that the reports are resorting accordingly

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

browser

</details>

<details>

mWeb-Chrome

</details>

<details>

ios_safari

</details>

<details>
<summary>Desktop</summary>

desktop

</details>

<details>
<summary>iOS</summary>

ios

</details>

<details>
<summary>Android</summary>

android

</details>


src/libs/ReportUtils.js
```diff
@@ -1891,14 +1891,13 @@ function canSeeDefaultRoom(report, policies, betas) {
  * filter out the majority of reports before filtering out very specific minority of reports.
  *
  * @param {Object} report
- * @param {String} currentReportId
```
venture1981, line 1894: this `currentReportId` is unnecessary variable in the code?

venture1981 (commented):
@hannojg 
How about css optimization?

src/components/LHNOptionsList/OptionRowLHN.js
```diff
@@ -254,4 +261,34 @@ OptionRowLHN.propTypes = propTypes;
...
+            {..._.omit(props, 'currentReportId')}
+            isFocused={isFocused}
+        />
+    );
+}
+OptionRowIsFocusedSupport.propTypes = propTypes;
+OptionRowIsFocusedSupport.defaultProps = defaultProps;
+OptionRowIsFocusedSupport.displayName = 'OptionRowIsFocusedSupport';
+
+// TODO: Note on mobile we could skip this HOC
```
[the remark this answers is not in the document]
hannojg: No, sorry, I wasn't pushing the latest changes. On mobile we can't skip this HOC, so never mind

src/libs/SidebarUtils.js
```diff
@@ -20,6 +20,7 @@ import * as UserUtils from './UserUtils';
 // Session also can remain stale because the only way for the current user to change is to sign out and sign in, which would clear out all the Onyx
 // data anyway and cause SidebarLinks to rerender.
 
+// TODO: get rid of this shit
```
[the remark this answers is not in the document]
hannojg: that shouldn't be here, sorry

src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -274,8 +281,6 @@ const chatReportSelector = (report) =>
         errorFields: {
             addWorkspaceRoom: report.errorFields && report.errorFields.addWorkspaceRoom,
         },
-        lastReadTime: report.lastReadTime,
-        lastMentionedTime: report.lastMentionedTime,
```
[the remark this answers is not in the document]
hannojg: Yes, this is the main performance optimisation as described in the proposal

hannojg:
> How about css optimization?

Sorry, what do you mean by that @venture1981

hannojg:
@0xmiroslav I think I did it correctly, no?

Screenshot 2023-06-26 at 10 31 28

Also thx for the review, I wasn't pushing the latest changes 😅 now everything looks correct to me

src/libs/ReportUtils.js
```diff
@@ -1916,13 +1915,6 @@ function shouldReportBeInOptionList(report, currentReportId, isInGSDMode, iouRep
         return false;
     }
 
-    // Include the currently viewed report. If we excluded the currently viewed report, then there
```
WoLewicki, line 1919: Are you sure we can just remove this condition and it will work correctly?

mountiny:
@hannojg could we get back to this one soon? thanks 🙇

src/libs/ReportUtils.js
hannojg: Haha okay, I may spoke too fast. I see that some tests are failing, indicating the need for this code. Will double and triple check 😊

mountiny:
@hannojg is this ready for review?

hannojg:
No not yet, sorry, the tests are still failing. Working on it right now

hannojg:
Okay, this needed a bit of a larger rework, as we do still need the `currentReportID`. Without it we'd cause a regression in GSD mode (which I was overlooking previously).
I am however very happy with the rework, as I can already feel how the code gets cleaner, easier to understand and more efficient.

Will share details when finishing this tomorrow ✌️

hannojg:
Okay, this is good for review again. Let me explain the changes I made on a high level to help with the code review:

## Splitting OptionRowLHN and SidebarLinks into UI and data components

There were two new components introduced:

- `<SidebarLinksData />`
- `<OptionRowLHNData />`

The corresponding none-data-components are now memoized, and are only expecting to receive derived values that are necessary for displaying the UI (and nothing more).

So, for example, `SidebarLinksData` calculates the list of report IDs we want to display given all the onyx data. Only when the calculated array really changed in its content (e.g. item order as in one report a new unread message arrived) it will pass the updated value to `SidebarLinks` causing an update.

The same goes for `OptionRowLHN` and its `optionItem` it wants to render. 

## Passing data from the components to onyx

Before, we were loading all the data needed for calculating the derived values for the components manually with `Onyx.connect` in e.g. `SidebarUtils`.
We also still had to connect the component to the onyx keys, to cause a re-render, which would then call the function to calculate the derived value (see e.g. `SidebarUtils.getOrderedReportIDs`).
This structure has its flaws, for example we were often missing to connect the right keys to the component, which would cause us to show stale data (See for example [this issue](https://github.com/Expensify/App/issues/21497))

We now connect all data needed to the component, and the component then calls the function with the data from the props. This is a transparent process that is easy to understand and reason about.
As this happens in the data component, we don't need to worry about unnecessary UI updates, as the underlaying UI component won't re-render unless the derived data really changed and we'd need to re-render.

### Future Considerations

The end game would be to not call the calculation function for derived data from the component using props, but from an onyx selector. In onyx we currently lack the functionality to have "dependent selectors".

For getting the user's personal data, we first need to load the `session` from onyx to get the user's `accountID` and then look into `personalDetailsList` (for this example, we currently have the `withCurrentUserPersonalDetails` hoc, which loads the `session` and `personalDetailsList` separately.)

Imagine we'd have the following in onyx:

```js
withOnyx({
   currentUserPersonalDetails: {
     keys: [ONYXKEYS.SESSION, ONYXKEYS.PERSONAL_DETAILS_LIST],
     selector: (session, personalDetails) => personalDetails[session.accountID]
   }
})
```

or another API could look like this:

```js
const currentUserPersonalDetailsSelector = createSelector(
   selectSessionAccountID,
   selectPersonalDetailsList,
   (accountID, personalDetails) => personalDetails[session.accountID]
)

withOnyx({
   currentUserPersonalDetails: currentUserPersonalDetailsSelector
})
```

The last example would even be better, as we don't rerun the selector when anything on the session changes, but just when the account ID changes.
I will try to make a proposal for this soon and then revisit.

## Utility `onyxSubscribe`

We want to avoid re-rendering our components for getting a new prop value, that is not needed for the UI output.
Earlier, we were connecting the `SidebarLinks` component to the `modal` key to get `willAlertModalBecomeVisible` which we needed for some keyboard shortcuts. 

`onyxSubscribe` is a convenience wrapper around `Onyx.connect` so we can read from onyx and receives updates, but not in the react state to avoid re-renders.

src/libs/ReportUtils.js
hannojg: Yeah, we can't remove that, otherwise the sorting in GSD/focus mode would break. I added it back but changed the general structure a bit to still be optimised.

mountiny:
@fedirjh This is another one which could be higher on your list.

thanks @hannojg for pushing this forwards and great write up ❤️

tests/unit/SidebarTest.js
```diff
@@ -31,7 +32,10 @@ describe('Sidebar', () => {
     );
 
     // Initialize the network key for OfflineWithFeedback
```
WoLewicki, line 34: Can you extend the comment to mention all things we do here?

src/libs/SidebarUtils.js
```diff
@@ -13,12 +13,16 @@ import * as CollectionUtils from './CollectionUtils';
 import * as LocalePhoneNumber from './LocalePhoneNumber';
 import * as UserUtils from './UserUtils';
 
-// Note: It is very important that the keys subscribed to here are the same
-// keys that are connected to SidebarLinks withOnyx(). If there was a key missing from SidebarLinks and it's data was updated
-// for that key, then there would be no re-render and the options wouldn't reflect the new data because SidebarUtils.getOrderedReportIDs() wouldn't be triggered.
-// There are a couple of keys here which are OK to have stale data. Having redundant subscriptions causes more re-renders which should be avoided.
-// Session also can remain stale because the only way for the current user to change is to sign out and sign in, which would clear out all the Onyx
-// data anyway and cause SidebarLinks to rerender.
+// Note: Earlier SidebarUtils.getOrderedReportIDs() used to have to parameters. All the needed data was loaded here directly
```
WoLewicki, line 16: `used to have to parameters` what were you trying to say here?

WoLewicki (commented):
Added some comments. The PR looks good and I believe it implements the solutions we had in mind too 🚀 Good job! cc @j-piasecki did we have any other things that could be made better?

src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -253,13 +217,12 @@ class SidebarLinks extends React.Component {
                         )}
                     </PressableWithoutFeedback>
                 </View>
-                {isLoading ? (
+                {this.props.isLoading ? (
                     skeletonPlaceholder
```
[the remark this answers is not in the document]
hannojg: You are right, thx, no clue why it has been defined outside

hannojg:
I incorporated the latest feedback and resolved conflicts. Eagerly waiting for your reviews 😊

fedirjh (commented):
Code looks solid. We should use props distrusting for functional components.

src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -136,11 +91,19 @@ class SidebarLinks extends React.Component {
         SidebarUtils.setIsSidebarLoadedReady();
         this.isSidebarLoaded = true;
 
+        let modal = {};
+        this.unsubscribeOnyxModal = onyxSubscribe({
+            key: ONYXKEYS.MODAL,
+            callback: (modalArg) => {
+                modal = modalArg;
+            },
+        });
```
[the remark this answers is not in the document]
hannojg: This is one of the performance improvements. The modal state is needed for a logic, that has nothing to do with the render/UI output. Thus, the component shouldn't re-render (because it's a really expensive to render one) just because we open a modal.

Long term I want to remove the modal state from onyx and move it e.g. to a react context, as I think it doesn't belong in the onyx store.

hannojg:
Thanks for the review @fedirjh will address all points/bugs shortly

src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -275,122 +236,5 @@ class SidebarLinks extends React.Component {
...
-        },
-        preferredLocale: {
-            key: ONYXKEYS.NVP_PREFERRED_LOCALE,
-        },
-        modal: {
-            key: ONYXKEYS.MODAL,
-        },
-    }),
-)(SidebarLinks);
+export default compose(withLocalize, withCurrentUserPersonalDetails, withNavigationFocus, withWindowDimensions, withNavigation)(SidebarLinks);
```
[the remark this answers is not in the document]
hannojg: It is used here:

https://github.com/margelo/expensify-app-fork/blob/5c35629da75b5fdca327c445c1f88aa392faff90/src/pages/home/sidebar/SidebarLinks.js#L154

and here

https://github.com/margelo/expensify-app-fork/blob/5c35629da75b5fdca327c445c1f88aa392faff90/src/pages/home/sidebar/SidebarLinks.js#L84-86

Although I think (in a follow up PR), we should be able to clean this up as well).

hannojg:
@fedirjh I addressed your comments and fixed all named bugs! Good for testing again

src/pages/home/sidebar/SidebarLinksData.js
```diff
@@ -0,0 +1,177 @@
...
+    priorityMode: CONST.PRIORITY_MODE.DEFAULT,
+    isPersonalDetailsLoading: true,
+    betas: [],
+    policies: [],
+};
+
+function SidebarLinksData({allReportActions, betas, chatReports, currentReportID, insets, isPersonalDetailsLoading, isSmallScreenWidth, onLayout, onLinkClick, policies, priorityMode}) {
+    const reportIDsRef = useRef([]);
+    const optionListItems = useMemo(() => {
+        const reportIDs = SidebarUtils.getOrderedReportIDs(currentReportID, chatReports, betas, policies, priorityMode, allReportActions);
```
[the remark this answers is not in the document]
mountiny: @hannojg was this one addressed?

fedirjh (commented):
@hannojg bugs were fixed 👍🏼. I left some comments that should be resolved.

src/pages/home/sidebar/SidebarLinksData.js
```diff
@@ -0,0 +1,177 @@
...
+        if (deepEqual(reportIDsRef.current, reportIDs)) {
+            return reportIDsRef.current;
+        }
+        reportIDsRef.current = reportIDs;
```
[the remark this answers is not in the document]
hannojg: That's because even when we put the same values to `SidebarUtils.getOrderedReportIDs` it will always return a new array reference. The same goes for every selector. So it can happen that one of the prop updates, but the real content of it stayed the same. In this case the `useMemo` will run and re-calculate the list of report IDs. The list will be exactly the same, but it's a new ref.If we'd return the new ref it would cause a re-render of `SidebarLinks`.
And we always must return from the `useMemo` hook. That's why we store the results in a ref, so in case its the very same result we can return the same reference.

src/pages/home/sidebar/SidebarLinksData.js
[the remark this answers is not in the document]
hannojg: Whoops, sorry, missed to comment here.
Check the comment I wrote here: https://github.com/Expensify/App/pull/21406#discussion_r1265226360

That won't work. The `useMemo` only compares values by reference, not by its real content.

src/components/LHNOptionsList/OptionRowLHNData.js
```diff
@@ -0,0 +1,132 @@
...
+    ...baseDefaultProps,
+};
+
+/*
+ * This component gets the data from onyx for the actual
+ * OptionRowLHN component.
+ * The OptionRowLHN component is memoized, so it will only
+ * re-render if the data really changed.
+ */
+function OptionRowLHNData(props) {
```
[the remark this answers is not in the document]
hannojg: Done 😊

hannojg:
Hey, I added one other performance improvement. Instead of checking for the navigation focus to disable accessibility elements in `SidebarLinks`, I am doing that now in `SidebarLinksData`. With this change we prevent re-rendering the whole `FlatList` just because the `isFocused` prop changed (the propagation will stop at `SidebarLinks` and only `SidebarLinksData` will re-render)

mountiny:
Thanks Hanno! 

@fedirjh seems like this is ready for another look 🚀

WoLewicki:
@hannojg I think we can remove `accessibilityElementsHidden` prop since it was needed when we did not have `stack` based navigation and now it is used only for tests (so we'd have to use something else there), but I am not 100% sure.  Code in stack for it: https://github.com/react-navigation/react-navigation/blob/24c03924397a6e59aba9f6b74a9c5cb4b939d9e1/packages/stack/src/views/Stack/CardContainer.tsx#L242.

hannojg:
Oh, would be even better to get rid of it completely @WoLewicki! Can you somehow confirm whether it can be removed or not?

@fedirjh You're the best tester, thank you so much for catching these! Will fix it shortly!

hannojg:
@fedirjh The `@expensify.sms` seems to be a general bug, I also have it on staging 👀 

Screenshot 2023-07-17 at 18 14 20

Once refreshed its gone

src/pages/home/sidebar/SidebarLinks.js
```diff
@@ -275,122 +231,5 @@ class SidebarLinks extends React.Component {
...
-        },
-        preferredLocale: {
-            key: ONYXKEYS.NVP_PREFERRED_LOCALE,
-        },
-        modal: {
-            key: ONYXKEYS.MODAL,
-        },
-    }),
-)(SidebarLinks);
+export default compose(withLocalize, withCurrentUserPersonalDetails, withWindowDimensions, withNavigation)(SidebarLinks);
```
WoLewicki, line 234: We are currently investigating another issue and it seems that `withNavigation` is no longer used in that component. Or are we missing something?

hannojg: @WoLewicki i think you are right, good catch. There is the usage of `Navigation.js` which uses the global navigation ref. I am not sure if `useNavigation` is there to make sure the navigationRef is ready, but either way i think it can get removed

hannojg:
@fedirjh Would appreciate another review 🙏 (Sorry if there are still bugs, there are many conditions to the LHN and I really hope we catch them all! In my testings everything looks good, curious about yours)

src/pages/home/sidebar/SidebarLinks.js
hannojg: removed it 🥷

hannojg:
I might be wrong but i think localhost is using the development code version, which is less optimized and slower, while staging already uses an optimized build, that is closer to the final release build.
But it might be that it caused a performance regression in that sense, will look into it today!

hannojg:
@fedirjh Hey, this is a bit hard to reproduce for me. However, I have a theory of why this might be happening.

I added a patch for onyx, from this PR:

- https://github.com/Expensify/react-native-onyx/pull/280

it removes the blocking behaviour from `withOnyx`.

Can you please pull, run `npc patch-package`, make sure the onyx patch got applied, and try testing for performance again?
In addition it'd be cool if you could run `npm run build-staging` and then test with the release build.

Thank you so much in advance!

hannojg:
You can, or i think you can run the build command and then use something like [npx serve](https://www.npmjs.com/package/serve) to try out the build

hannojg:
Okay, i reverted the patch!

hannojg:
I pushed two little last changes. I think we should be good now

fedirjh (approved):
Looks good to me and tests well

src/components/LHNOptionsList/OptionRowLHNData.js
```diff
@@ -0,0 +1,140 @@
...
+                firstName: personalData.firstName,
+                avatar: UserUtils.getAvatar(personalData.avatar, personalData.accountID),
+            };
+            return finalPersonalDetails;
+        },
+        {},
+    );
+
+// This component is rendered in a list.
+// On scroll we want to avoid that a item re-renders
```
mountiny, line 113:
```suggestion
// On scroll we want to avoid item re-renders
```

src/components/LHNOptionsList/OptionRowLHNData.js
```diff
@@ -0,0 +1,140 @@
...
+                displayName: personalData.displayName,
+                firstName: personalData.firstName,
+                avatar: UserUtils.getAvatar(personalData.avatar, personalData.accountID),
+            };
+            return finalPersonalDetails;
+        },
+        {},
+    );
+
+// This component is rendered in a list.
```
mountiny, line 112: Could you also use the `/ **` style for this, please?

src/libs/ReportActionsUtils.js
```diff
@@ -97,13 +97,14 @@ function hasCommentThread(reportAction) {
  * Returns the parentReportAction if the given report is a thread/task.
  *
  * @param {Object} report
+ * @param {Object} [allReportActionsParam]
  * @returns {Object}
  */
-function getParentReportAction(report) {
+function getParentReportAction(report, allReportActionsParam) {
```
mountiny, line 103: If its optional should we give it an empty object as a default value?
```suggestion
function getParentReportAction(report, allReportActionsParam = {}) {
```

mountiny (dismissed):
Great job on this one @hannojg and @fedirjh just couple of comments

WoLewicki (approved):
LGTM 🚀

0xmiroslav (commented):
Some issues after logout and re-login:
- LHN items order changed (2nd and 3rd are swapped in video)
- go to any random report after login, instead of last opened report

https://github.com/Expensify/App/assets/97473779/f7a743c3-ff48-475c-859d-a63dd5d41035

mountiny (approved):
@hannojg  @fedirjh Lets be on a look out for deploy blockers tomorrow or after the next deploy. I know you have tested well but with such PRs its always tricky and high chance of something breaking would be great to have you around to follow up with a fix as reverting would be messy

mountiny:
Thanks was not sure if you are reviewing, sorry

hannojg:
@0xmiroslav looking into it now!

hannojg:
@0xmiroslav Oh, I have seen this as well and thought I'd be a bug! but its on staging as well. Pinned rooms are only sorted by their name. When two workspaces have the same name their order will always be random (often times its by luck that its the same, but its really random) 😅

hannojg:
https://github.com/Expensify/App/assets/16821682/91462cf2-c974-4616-9816-d5cabe798c44

Proof that I am no liar 😄

hannojg:
@mountiny is it possible that something broke during merge?

@0xmiroslav are you on the latest version of the perf branch?

mountiny:
@hannojg it is possible

jasperhuangg:
This PR has caused a number of regressions:

https://github.com/Expensify/App/issues/23189
https://github.com/Expensify/App/issues/23195
https://github.com/Expensify/App/issues/23202

I'm going to revert it

src/libs/ReportActionsUtils.js
```diff
@@ -97,13 +97,14 @@ function hasCommentThread(reportAction) {
  * Returns the parentReportAction if the given report is a thread/task.
  *
  * @param {Object} report
+ * @param {Object} [allReportActionsParam]
  * @returns {Object}
  */
-function getParentReportAction(report) {
+function getParentReportAction(report, allReportActionsParam = {}) {
```
aldo-expensify, line 103: This default is causing the deploy blocker here: https://github.com/Expensify/App/issues/23202

If `allReportActionsParam` is not passed, it has value `{}` which is truthy, so we never use `allReportActions`. The `allReportActionsParam || allReportActions` doesn't work as intended.

hannojg: thanks you for quickly fixing this!!

mountiny: Thanks for the fix

src/pages/home/sidebar/SidebarLinksData.js
```diff
@@ -0,0 +1,200 @@
...
+        avatar: policy.avatar,
+    };
+
+export default compose(
+    withCurrentReportID,
+    withNavigationFocus,
+    withOnyx({
+        chatReports: {
+            key: ONYXKEYS.COLLECTION.REPORT,
+            selector: chatReportSelector,
```
s77rt, line 179: This caused a "regression"? Without the use of the selector we used to pass the reports objects as is in Onyx. Then we mutated those objects here https://github.com/Expensify/App/blob/bba0eb2868fa95d483da1b380551bc46fe0df427/src/libs/SidebarUtils.js#L109-L110 by adding a `displayName` to each report. If you try to access `report.displayName` from anywhere in the App you'd be able to (because we modified the onyx data).

Now, we are using a selector and we are returning a copy of Onyx data, we are mutating a clone and not the onyx data. If we try to access `report.displayName` from anywhere outside the Sidebar we won't get a result which in turn caused this issue https://github.com/Expensify/App/issues/23348.

hannojg: Wow great catch! But at the same time I have to say that how it was working before is very dirty and we should find a different solution for it 😅

hannojg: I see that the proposal proposes a clean way, good job 💪

Santhosh-Sellavel:
This PR Causes [this](https://github.com/Expensify/App/issues/23676) regression more details [here](https://github.com/Expensify/App/issues/23676#issuecomment-1652400735)

tests/unit/SidebarFilterTest.js
```diff
@@ -253,6 +258,7 @@ describe('Sidebar', () => {
             );
         });
 
+        // NOTE: This is also for #focus mode, should we move this test block?
```
marcaaron, line 261: I am not sure what the original intention here was. Maybe @tgolen can remember. This comment stuck out to me as pretty confusing and something we should fix.

**context:** Working on a PR and encountered a failing test with this title:

>Sidebar › in default (most recent) mode › all combinations of isArchived, isUserCreatedPolicyRoom, hasAddWorkspaceError, isUnread, isPinned, hasDraft › the booleans true,false,false,true,false,false

But when I look at the lines of code it's referencing I see that we are not testing "most recent" but "focus" mode.

marcaaron: This test also confuses me if the intention is that we should be testing "Most recent" mode. 

If we are trying to see that a report is in the "Most recent" list then most all reports will be in the list.

Or at least doesn't seem dependent on the arguments we are passing.

>isArchived, isUserCreatedPolicyRoom, hasAddWorkspaceError, isUnread, isPinned, hasDraft

All of those will would show in "Most recent" (all other things about the report being "normal")

marcaaron: Ok after reviewing the two blocks - I have determined they are identical and we should be able to remove this...

I don't think this was intended to be put here and it's not logical to test.

PR #25963, opened 2023-08-25
[NoQA] Revert "Fixing the race condition on `history.go` due to buggy queue async callbacks"

mountiny:
Straight revert of Expensify/App#25866


thienlnam (approved):

thienlnam:
Merged without passing checklists since this is a straight revert

PR #26781, opened 2023-09-05
[26737] Made sure that the sections from the request money are scroll…

DanutGavrus:
As this is a draft PR, I'm adding here a video with before(latest staging as of now) and after(this PR changes) so we may decide if we want(or not) to proceed with these changes:

https://github.com/Expensify/App/assets/56603839/48a8a160-5d0d-4a31-8f49-dfcc3271dcc3

https://github.com/Expensify/App/assets/56603839/78b0267e-f839-459a-a84e-49cb6d25dd55

https://github.com/Expensify/App/assets/56603839/0ccec0f6-ac5f-4bc5-a404-a55c1a389280


DanutGavrus:
The comparison was made against the latest staging, when on `main`, we have this PR: https://github.com/Expensify/App/pull/26728 which added the `<ScrollView contentContainerStyle={styles.flex1}>` in the distance request. The only difference on `main` is that we may now scroll down to the button, but we lose the bottom margin when we scroll as seen here(I can't scroll anymore):
image

And we still have 1 problem left, we have functionality in code which want to show 4 locations before making the list scrollable, and to also show only half of the fifth and up location as to indicate that the list is scrollable as seen here:  

https://github.com/Expensify/App/blob/b9c48328903b583ebfeb1f02d1e2110d25c42a2e/src/components/DistanceRequest.js#L42  

https://github.com/Expensify/App/blob/b9c48328903b583ebfeb1f02d1e2110d25c42a2e/src/components/DistanceRequest.js#L133-L135

DanutGavrus:
I'll close this based on this discussion: https://expensify.slack.com/archives/C049HHMV9SM/p1693986149607419. We've fixed the mentioned issues on the `main` branch by replacing `flex1` with `flexGrow1` in the distance page. This PR's only remaining benefit would be that it would remove all scroll views from sections and place only 1 upper in the hierarchy, making any existing & future section scrollable. We don't know for sure that it won't introduce new problem and It doesn't seem required for now.

Issue #27585, opened 2023-09-16
[$500] Chat - Link in end of line displays tooltip over text and not on link


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Open the app
2. Open any report
3. Type in some text and at the end of line, add link with long text eg: [Test test test test test test test test test test]([testing.com](http://testing.com/))
4. Send the text, text should have text and link in first line and link should be continued in second line too
5. Hover on link and observe that tooltip is displayed over text

## Expected Result:
App should display tooltip over link on hover

## Actual Result:
 App displays tooltip over text on link hover when link is multiline and occupies the end of line with text in start

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [ ] Android / Chrome
- [ ] iOS / native
- [ ] iOS / Safari
- [x] MacOS / Chrome / Safari
- [ ] MacOS / Desktop

**Version Number:** 1.3.70-5

**Reproducible in staging?:** Yes

**Reproducible in production?:** Yes

**If this was caught during regression testing, add the test name, ID and link from TestRail:**

**Email or phone of affected tester (no customers):**

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://github.com/Expensify/App/assets/78819774/93edea85-b4a3-4757-a14b-8d2ca5676956

https://github.com/Expensify/App/assets/78819774/8aae4252-8177-4a90-8a3d-ae482334dd10

**Expensify/Expensify Issue URL:**

**Issue reported by:** @Dhanashree-Sawant

**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1694604089052599

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~01aa77f02fd5409216</li>
        <li>Upwork Job ID: 1703100827931217920</li>
        <li>Last Price Increase: 2023-09-16</li>
<li>Automatic offers: </li>
<ul>
[...]

PR #27817, opened 2023-09-19
Fix: Chat - Link in end of line displays tooltip over text and not on link

ewanmellor:
### Details

Add a new prop to `Tooltip` called `shouldUseMultilinePositioning`. This enables a new algorithm for finding the bounding box target for the tooltip. In the case where the target has wrapped onto multiple lines (e.g. it is a link in a chat window) then the link has multiple bounding boxes, and we want to show the tooltip against the one that the user is hovering over.

As part of this, extend `Hoverable` to pass the Event to its `onHoverIn` / `onHoverOut` callbacks.

Enable this new algorithm from `BaseAnchorForCommentsOnly`, which is the base class for links that show up in chat.

### Fixed Issues

$ https://github.com/Expensify/App/issues/27585
PROPOSAL: https://github.com/Expensify/App/issues/27585#issuecomment-1724380334

### Tests

Action Performed:

On a platform with a mouse (i.e. not mobile):

1. Open the app
2. Open any report
3. Type in some text and at the end of line, add link with long text eg: Test test test test test test test test test test
4. Send the text, text should have text and link in first line and link should be continued in second line too. Adjust the window size to force the text to wrap if necessary.
5. Hover on link and observe that tooltip is displayed over text

Expected Result:
App should display tooltip over link on hover. The tooltip should appear over the same section of the link that the user is hovering over.

### Offline tests

N/A.  There is no network interaction in this change.

### QA Steps

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

Screenshot 2023-09-19 at 3 38 28 PM
Screenshot 2023-09-19 at 3 38 23 PM

</details>

<details>

</details>

<details>

</details>

<details>
<summary>Desktop</summary>

</details>

<details>
<summary>iOS</summary>

</details>

<details>
<summary>Android</summary>

</details>

### DIFF
--- a/src/components/Hoverable/index.js
+++ b/src/components/Hoverable/index.js
@@ -76,9 +76,8 @@ class Hoverable extends Component {
      * Sets the hover state of this component to true and execute the onHoverIn callback.
      *
      * @param {Boolean} isHovered - Whether or not this component is hovered.
-     * @param {Event} ev - The event that triggered this state change.
      */
-    setIsHovered(isHovered, ev) {
+    setIsHovered(isHovered) {
         if (this.props.disabled) {
             return;
         }
@@ -95,7 +94,7 @@ class Hoverable extends Component {
         if (this.isScrollingRef && this.props.shouldHandleScroll && !this.state.isHovered) return;
 
         if (isHovered !== this.state.isHovered) {
-            this.setState({isHovered}, () => (isHovered ? this.props.onHoverIn : this.props.onHoverOut)(ev));
+            this.setState({isHovered}, isHovered ? this.props.onHoverIn : this.props.onHoverOut);
         }
     }
 
@@ -114,18 +113,15 @@ class Hoverable extends Component {
             return;
         }
 
-        this.setIsHovered(false, e);
+        this.setIsHovered(false);
     }
 
-    /**
-     * @param {Event} ev - The visibility-change event object.
-     */
-    handleVisibilityChange(ev) {
+    handleVisibilityChange() {
         if (document.visibilityState !== 'hidden') {
             return;
         }
 
-        this.setIsHovered(false, ev);
+        this.setIsHovered(false);
     }
 
     render() {
@@ -158,24 +154,30 @@ class Hoverable extends Component {
                 }
             },
             onMouseEnter: (el) => {
-                this.setIsHovered(true, el);
+                this.setIsHovered(true);
 
                 if (_.isFunction(child.props.onMouseEnter)) {
                     child.props.onMouseEnter(el);
                 }
+                if (_.isFunction(this.props.onMouseEnter)) {
+                    this.props.onMouseEnter(el);
+                }
             },
             onMouseLeave: (el) => {
-                this.setIsHovered(false, el);
+                this.setIsHovered(false);
 
                 if (_.isFunction(child.props.onMouseLeave)) {
                     child.props.onMouseLeave(el);
+                }
+                if (_.isFunction(this.props.onMouseLeave)) {
+                    this.props.onMouseLeave(el);
                 }
             },
             onBlur: (el) => {
                 // Check if the blur event occurred due to clicking outside the element
                 // and the wrapperView contains the element that caused the blur and reset isHovered
                 if (!this.wrapperView.contains(el.target) && !this.wrapperView.contains(el.relatedTarget)) {
-                    this.setIsHovered(false, el);
+                    this.setIsHovered(false);
                 }
 
                 if (_.isFunction(child.props.onBlur)) {
--- a/src/components/Hoverable/index.js
+++ b/src/components/Hoverable/index.js
@@ -76,8 +76,9 @@ class Hoverable extends Component {
      * Sets the hover state of this component to true and execute the onHoverIn callback.
      *
      * @param {Boolean} isHovered - Whether or not this component is hovered.
+     * @param {Event} ev - The event that triggered this state change.
      */
-    setIsHovered(isHovered) {
+    setIsHovered(isHovered, ev) {
         if (this.props.disabled) {
             return;
         }
@@ -94,7 +95,7 @@ class Hoverable extends Component {
         if (this.isScrollingRef && this.props.shouldHandleScroll && !this.state.isHovered) return;
 
         if (isHovered !== this.state.isHovered) {
-            this.setState({isHovered}, isHovered ? this.props.onHoverIn : this.props.onHoverOut);
+            this.setState({isHovered}, () => (isHovered ? this.props.onHoverIn : this.props.onHoverOut)(ev));
         }
     }
 
@@ -113,15 +114,18 @@ class Hoverable extends Component {
             return;
         }
 
-        this.setIsHovered(false);
+        this.setIsHovered(false, e);
     }
 
-    handleVisibilityChange() {
+    /**
+     * @param {Event} ev - The visibility-change event object.
+     */
+    handleVisibilityChange(ev) {
         if (document.visibilityState !== 'hidden') {
             return;
         }
 
-        this.setIsHovered(false);
+        this.setIsHovered(false, ev);
     }
 
     render() {
@@ -154,14 +158,14 @@ class Hoverable extends Component {
                 }
             },
             onMouseEnter: (el) => {
-                this.setIsHovered(true);
+                this.setIsHovered(true, el);
 
                 if (_.isFunction(child.props.onMouseEnter)) {
                     child.props.onMouseEnter(el);
                 }
             },
             onMouseLeave: (el) => {
-                this.setIsHovered(false);
+                this.setIsHovered(false, el);
 
                 if (_.isFunction(child.props.onMouseLeave)) {
                     child.props.onMouseLeave(el);
@@ -171,7 +175,7 @@ class Hoverable extends Component {
                 // Check if the blur event occurred due to clicking outside the element
                 // and the wrapperView contains the element that caused the blur and reset isHovered
                 if (!this.wrapperView.contains(el.target) && !this.wrapperView.contains(el.relatedTarget)) {
-                    this.setIsHovered(false);
+                    this.setIsHovered(false, el);
                 }
 
                 if (_.isFunction(child.props.onBlur)) {
--- a/src/components/Hoverable/index.js
+++ b/src/components/Hoverable/index.js
@@ -156,21 +156,21 @@ class Hoverable extends Component {
             onMouseEnter: (el) => {
                 this.setIsHovered(true);
 
-                if (_.isFunction(child.props.onMouseEnter)) {
-                    child.props.onMouseEnter(el);
-                }
                 if (_.isFunction(this.props.onMouseEnter)) {
                     this.props.onMouseEnter(el);
+                }
+                if (_.isFunction(child.props.onMouseEnter)) {
+                    child.props.onMouseEnter(el);
                 }
             },
             onMouseLeave: (el) => {
                 this.setIsHovered(false);
 
-                if (_.isFunction(child.props.onMouseLeave)) {
-                    child.props.onMouseLeave(el);
-                }
                 if (_.isFunction(this.props.onMouseLeave)) {
                     this.props.onMouseLeave(el);
+                }
+                if (_.isFunction(child.props.onMouseLeave)) {
+                    child.props.onMouseLeave(el);
                 }
             },
             onBlur: (el) => {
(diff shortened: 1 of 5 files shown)


ewanmellor:
I have read the CLA Document and I hereby sign the CLA

ewanmellor:
recheck

ewanmellor:
recheck

src/components/Tooltip/tooltipPropTypes.js
```diff
@@ -42,6 +59,7 @@ const defaultProps = {
     renderTooltipContent: undefined,
     renderTooltipContentKey: [],
     shouldHandleScroll: false,
+    shouldUseMultilinePositioning: false,
```
s77rt, line 62: Let's remove this prop. I think we want to have this feature enabled for all tooltips

src/components/Hoverable/index.js
```diff
@@ -94,7 +95,7 @@ class Hoverable extends Component {
         if (this.isScrollingRef && this.props.shouldHandleScroll && !this.state.isHovered) return;
 
         if (isHovered !== this.state.isHovered) {
-            this.setState({isHovered}, isHovered ? this.props.onHoverIn : this.props.onHoverOut);
+            this.setState({isHovered}, () => (isHovered ? this.props.onHoverIn : this.props.onHoverOut)(ev));
```
s77rt, line 98: Given that `onHoverIn` may be called without events it's better to use `onMouseEnter`. Add onMouseEnter and onMouseLeave props to Hoverable and use onMouseEnter to set the target and mouse position.

In Tooltip:
```js
    const target = useRef(null);
    const initialMousePosition = useRef({x: 0, y: 0});

    const updateTargetAndMousePosition = useCallback((e) => {
        target.current = e.target;
        initialMousePosition.current = {x: e.clientX, y: e.clientY};
    }, []);

    <Hoverable onMouseEnter={updateTargetAndMousePosition) ... />

```

src/components/Tooltip/index.js
```diff
@@ -44,33 +79,59 @@ function Tooltip(props) {
...
+     */
+    const showTooltip = useCallback(
+        (ev) => {
+            if (shouldUseMultilinePositioning) {
+                if (ev) {
+                    const {clientX, clientY, target} = ev;
+                    const bb = chooseBoundingBox(target, clientX, clientY);
+                    updateBounds(bb);
+                }
+            }
```
s77rt, line 107: Undo this. `chooseBoundingBox` should be called from `updateBounds`


```js
    const updateBounds = (bounds) => {
        if (bounds.width === 0) {
            setIsRendered(false);
        }
        const betterBounds = chooseBoundingBox(target.current, initialMousePosition.current.x, initialMousePosition.current.y);
        setWrapperWidth(betterBounds.width);
        setWrapperHeight(betterBounds.height);
        setXOffset(betterBounds.x);
        setYOffset(betterBounds.y);
    };

```

src/components/Tooltip/index.js
```diff
@@ -13,14 +13,49 @@ import useWindowDimensions from '../../hooks/useWindowDimensions';
...
+ * @param {number} clientX The X position from the MouseEvent.
+ * @param {number} clientY The Y position from the MouseEvent.
+ * @param {number} slop An allowed slop factor when searching for the bounding
+ * box. If the user is moving the mouse quickly we can end up getting a
+ * hover event with the position outside any of our bounding boxes. We retry
+ * with a small slop factor in that case, so if we have a bounding box close
+ * enough then we go with that.
+ * @return {DOMRect} The chosen bounding box.
+ */
+function chooseBoundingBox(target, clientX, clientY, slop = 0) {
```
s77rt, line 34: I think we'd need a safer algo without the recursive call.

src/components/Tooltip/index.js
```diff
@@ -13,14 +13,49 @@ import useWindowDimensions from '../../hooks/useWindowDimensions';
...
+            return bb;
+        }
+    }
+    if (slop === 0) {
+        // Retry with a slop factor, in case the user is moving the mouse quickly.
+        return chooseBoundingBox(target, clientX, clientY, 5);
+    }
+    // Fall back to the full bounding box if we failed to find a matching one
+    // (shouldn't happen).
+    return target.getBoundingClientRect();
```
s77rt, line 48: Don't write code that can't be reached.

src/components/Tooltip/index.js
```diff
@@ -13,14 +13,49 @@ import useWindowDimensions from '../../hooks/useWindowDimensions';
...
+ * @param {number} clientY The Y position from the MouseEvent.
+ * @param {number} slop An allowed slop factor when searching for the bounding
+ * box. If the user is moving the mouse quickly we can end up getting a
+ * hover event with the position outside any of our bounding boxes. We retry
+ * with a small slop factor in that case, so if we have a bounding box close
+ * enough then we go with that.
+ * @return {DOMRect} The chosen bounding box.
+ */
+function chooseBoundingBox(target, clientX, clientY, slop = 0) {
+    const bbs = target.getClientRects();
```
s77rt, line 35: minor optimisation, if bbs contains only one element then just use it (that element would be same as getBoundingClientRect)

s77rt (changes requested):

src/components/Tooltip/index.js
ewanmellor: @s77rt This doesn't make sense. If the user hovers over one part of the link and then moves to the other part, you need to move the tooltip. `updateBounds` is only called when the element moves, but in that case the element itself has not moved, so we will miss the necessary update. That's why it needs to be in the hover event handler.

s77rt: I'm not sure if this is a valid case. if the user moves to the other part then we'd have two possible outcomes:
1. User is still hovering the target, in this case the tooltip does not need to be updated as this case is only possible if the mouse stays at the same row (x position)
2. User lost hover, in this case onMouseEnter will be called once the user hover the next part and the tooltip will be placed correctly

ewanmellor: @s77rt Right, but your proposed `updateTargetAndMousePosition` doesn't set the tooltip position, it just saves the mouse position at that time. So when the `onMouseEnter` is called, you remember the location but don't do anything with that. And you won't get a call to `updateBounds` because the element hasn't moved.

s77rt: Yes, that's the intended use. updateBounds will be called once the tooltip is visible or on bounds change

ewanmellor: @s77rt I think we're talking at cross-purposes.

Earlier you said "Undo this. chooseBoundingBox should be called from updateBounds". I took this to mean "don't call chooseBoundingBox from showTooltip, only call it from updateBounds". So how does "updateBounds will be called once the tooltip is visible" happen in your mind?

s77rt: updateBounds will be called by BoundsObserver

ewanmellor: @s77rt "updateBounds will be called by BoundsObserver". This only happens if the element moves. If the user is moving the mouse between two parts of the same link, the element does not move and updateBounds is not called.

s77rt: That's correct. But the user can't move the mouse between two parts without firing the onMouseLeave and then onMouseEnter again on the second part

ewanmellor: @s77rt Yes, but again, if the user does onMouseLeave and onMouseEnter, then you have nothing calling `updateBounds`. Your proposed `updateTargetAndMousePosition` does not call `updateBounds`. So the tooltip does not move. What am I missing from what you are saying?

s77rt: - onMouseLeave -> onHoverOut -> hideTooltip -> setIsVisible(false)
- onMouseEnter (1st callback) -> updateTargetAndMousePosition
- onMouseEnter (2nd callback) -> onHoverIn -> showTooltip -> setIsVisible(true)

Besides setting the target and mouse position, we have a change in the isVisible state which is passed to the enabled prop of BoundsObserver and per the docs https://socket.dev/npm/package/@react-ng/bounds-observer
> Additionally, this callback will be invoked on enabling, i.e. when the enabled property changes to true

ewanmellor: @s77rt OK, I finally see what you are saying. I have updated this PR with all the changes mentioned.

ewanmellor:
recheck

src/components/Hoverable/index.js
```diff
@@ -159,13 +159,19 @@ class Hoverable extends Component {
                 if (_.isFunction(child.props.onMouseEnter)) {
                     child.props.onMouseEnter(el);
                 }
+                if (_.isFunction(this.props.onMouseEnter)) {
+                    this.props.onMouseEnter(el);
+                }
```
s77rt, line 164: Move this to be the first callback (same for onMouseLeave)

src/components/Tooltip/index.js
```diff
@@ -91,10 +156,15 @@ function Tooltip(props) {
...
             setIsRendered(false);
         }
-        setWrapperWidth(bounds.width);
-        setWrapperHeight(bounds.height);
-        setXOffset(bounds.x);
-        setYOffset(bounds.y);
+        const t = target.current;
+        if (!t) {
+            return;
+        }
```
s77rt, line 162: Is there a case where target is not defined?

src/components/Tooltip/index.js
```diff
@@ -91,10 +156,15 @@ function Tooltip(props) {
...
+        const betterBounds = chooseBoundingBox(t, initialMousePosition.current.x, initialMousePosition.current.y);
```
s77rt, line 163: Please add a comment here on why we need better bounding box

s77rt (changes requested):

src/components/Hoverable/index.js
ewanmellor: Done.

src/components/Tooltip/index.js
ewanmellor: Done.

src/components/Tooltip/index.js
ewanmellor: No, but the type system would not be able to see that, because it cannot tell that `updateBounds` will never be called before a hover event makes the tooltip visible.

Regardless, this file is not TypeScript, so I've removed it.

ewanmellor:
recheck

src/components/Hoverable/index.js
s77rt: This is still not the first, it needs to goes before setIsHovered

src/components/Tooltip/index.js
```diff
@@ -13,6 +13,63 @@ import useWindowDimensions from '../../hooks/useWindowDimensions';
...
+ * This is a helper function for chooseBoundingBox below.
+ *
+ * @param {Element} bbs The bounding boxes of DOM element being hovered over.
+ * @param {number} clientX The X position from the MouseEvent.
+ * @param {number} clientY The Y position from the MouseEvent.
+ * @param {number} slop An allowed slop factor when searching for the bounding
+ * box. If the user is moving the mouse quickly we can end up getting a
+ * hover event with the position outside any of our bounding boxes. We retry
+ * with a small slop factor in that case, so if we have a bounding box close
+ * enough then we go with that.
```
s77rt, line 27: I have been checking and I think the need for the tolerance is not related to moving the mouse quickly, I can reproduce the bug constantly and this seems due to the returned bbs values corresponding to the text fragments only and not the block. Let's have the tolerance always used and integrate this function inside `chooseBoundingBox`

src/components/Tooltip/index.js
```diff
@@ -13,6 +13,63 @@ import useWindowDimensions from '../../hooks/useWindowDimensions';
...
+    }
+    // Retry with a slop factor, in case the user is moving the mouse quickly.
+    bb = chooseBoundingBoxWithSlop(bbs, clientX, clientY, 5);
+    if (bb) {
+        return bb;
+    }
+    // Fall back to the full bounding box if we failed to find a matching one.
+    // This could only happen if the user is moving the mouse very quickly
+    // and they got it outside our slop above.
+    return target.getBoundingClientRect();
```
s77rt, line 70:
```suggestion
    return bbs[0];
```

s77rt (changes requested):

PR #28395, opened 2023-09-28
Update _routes.yml

bfitzexpensify:
Editing this to add the article in https://github.com/Expensify/Expensify/issues/309819 via the instructions from https://stackoverflowteams.com/c/expensify/questions/15135


dangrous:
Okay so I think (I'm not sure) that you don't have to do this part anymore. Do you mind checking in Slack to confirm? If so we should update those instructions. I might be working off of a faulty memory though!

PR #28615, opened 2023-10-02
Improve memoization of LHN by handling the isFocused prop better

kacper-mikolajczak:
## Intro

Transition between chats (reports), seems to be quite laggy. In this document, we try to investigate the issue and come up with possible solution. 

## Steps to reproduce

1. Load two chats
2. Navigate from one to another

## Recording

https://github.com/Expensify/App/assets/62747088/85bb45e8-e394-4f61-bc8a-5fe57361ef77

## Investigations

It looks like there is a big overhead in form of rendering LHN’s list of `OptionRowLHNData` components.
BeforeFixes

In the profiling trace we see that every item in the list is re-rendered. Each re-render takes up to **5ms** to complete, which with current list size can significantly throttle rendering.

The re-rendering is caused by change in `currentReportID` state, which is provided by `withCurrentReportID` HOC. `currentReportID` prop is used in  `OptionRowLHNData` in one place, to find out wether component should be focused or not:

```tsx
const isFocused = !shouldDisableFocusOptions && currentReportID === reportID;
```

The `isFocused` is then directly passed down into child.

The change in `currentReportID` breaks memoization, causing heavy re-render of each item.

### Wrapper component

As `currentReportID` is only used **********once********** and `isFocused` is a boolean flag, we can extract this logic into separate **************wrapper************** component and refactor `OptionRowLHNData` to receive primitive `isFocused` prop, which will not break memoization. 

AfterFixes copy

By applying these changes, we were able to shrink down the rendering time of List component to **~100ms**, which will be even more significant as number of list items grows. The rest of rendering overhead seems to be related to different components, `Report` itself etc.

The rendering time should be now only merely related to list size, as it only triggers one heavy re-render.

Wrapper component would look more or less like this:

```tsx
function OptionRowLHNDataWrapper({currentReportID, shouldDisableFocusOptions, ...props}) {
    const isFocused = !shouldDisableFocusOptions && currentReportID === props.reportID;
    return (
        <OptionRowLHNDataComposed
            {...props}
            isFocused={isFocused}
        />
    );
}
```

This solution does not add anything beyond a simple wrapper so it is quite easy to do and should not be error-prone.

### Use `withCurrentReportID` in list component (Alternative)

AfterFixesDetailedAlternatives copy

Alternatively, we can use `withCurrentReportID` once at the level of list component instead of in each children. Then we will directly pass `isFocused` prop into children, so it won’t break memoization of list items.

The solution is more elegant, as it uses `currentReportID` provider once and does not expose such internals into children components. On the flipside, we need to re-render list itself, which might add some overhead.

In this scenario, it takes about **********~120ms********** to render the list, which is about **********15ms********** slower than previous solution.

## Conclusions

While we managed to reduce rendering times, we still need to think on which solution should be implemented. Furthermore, it might be a good idea to investigate rendering of list items, as it is quite simple component but still takes ************~5ms************ to render.

### Fixed Issues

$ https://github.com/Expensify/App/issues/28631
PROPOSAL: https://github.com/Expensify/App/issues/28631 (it contains a proposal)

### Tests

- [x] Verify that no errors/warnings appear in the JS console.

1. Open app
2. See if can easily switch between multiple chats
3. See if focus (opacity) changes according to selected chats

### Offline tests

N/A
### QA Steps

1. Open app
2. See if can easily switch between multiple chats
3. See if focus (opacity) changes according to selected chats

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

https://github.com/Expensify/App/assets/62747088/54d241b3-210a-4f0a-ab46-69b4cb0775a8

</details>

<details>

N/A

</details>

<details>

N/A

</details>

<details>
<summary>Desktop</summary>

https://github.com/Expensify/App/assets/62747088/e4320146-5aa4-410c-b4d7-cc14a749f96f

</details>

<details>
<summary>iOS</summary>

Mobile recording for reference

https://github.com/Expensify/App/assets/62747088/9f24267c-0a2a-421e-9cd8-a7dad7134b31

</details>

<details>
<summary>Android</summary>

N/A

</details>


kacper-mikolajczak:
Hi @muttmuure, here is a draft of partial improvements that I've done recently. I wanted to share it on Slack but it is quite hard to put such document into Slack message :D Let me know what you think about results and overall shape of such document. In the meantime I will work on further improvements

mountiny:
@kacper-mikolajczak Thanks! I think this looks great!

I agree that the second solution leads to more readable code, adding that one wrapper feels a bit odd to me.

@kacper-mikolajczak could you summarize these two options and post it to #newdot-performance room? I think it would be great to get opinions of other agency experts and engineers on this and then we can implement this. Thanks!

kacper-mikolajczak:
@mountiny  Yeah definitely! Thanks for the feedback 👍

src/components/LHNOptionsList/OptionRowLHNData.js
```diff
@@ -20,8 +19,8 @@ import CONST from '../../CONST';
 import reportActionPropTypes from '../../pages/home/report/reportActionPropTypes';
 
 const propTypes = {
-    /** If true will disable ever setting the OptionRowLHN to focused */
-    shouldDisableFocusOptions: PropTypes.bool,
+    /** Wether row should be focused */
```
hurali97, line 22:
```suggestion
    /** Whether row should be focused */
```

src/components/LHNOptionsList/OptionRowLHNData.js
```diff
@@ -91,7 +86,6 @@ function OptionRowLHNData({
     const reportID = propsToForward.reportID;
     // We only want to pass a boolean to the memoized component,
     // instead of a changing number (so we prevent unnecessary re-renders).
-    const isFocused = !shouldDisableFocusOptions && currentReportID === reportID;
```
hurali97, line 94: We can move this comment to `OptionRowLHNDataWithFocus`, since we moved the `isFocused` there as well.

src/components/LHNOptionsList/OptionRowLHNDataWithFocus.js
```diff
@@ -0,0 +1,32 @@
...
+    ...withCurrentReportIDPropTypes,
+    shouldDisableFocusOptions: PropTypes.bool,
+};
+
+const defaultProps = {
+    ...withCurrentReportIDDefaultProps,
+    shouldDisableFocusOptions: false,
+};
+
+function OptionRowLHNDataWithFocus({currentReportID, shouldDisableFocusOptions, ...props}) {
```
hurali97, line 16: Can we add some comments stating what this `OptionRowLHNDataWithFocus ` does? It would be cool in future to have an understanding of what it does because not everyone will be coming back to the PR to understand.

hurali97 (changes requested):
LGTM ! Just a couple comments.

src/components/LHNOptionsList/OptionRowLHNData.js
kacper-mikolajczak: Done ✅

src/components/LHNOptionsList/OptionRowLHNDataWithFocus.js
kacper-mikolajczak: Done ✅

kacper-mikolajczak:
BTW, on iPhone 13 simulator the difference is huge

mountiny:
Damn, that looks great!

kacper-mikolajczak:
@mountiny Yeah that does look promising! Wonder what would be a perceived perf gain, as the list is out of the view when the render happens (it is covered by report screen). But hey, at least we reduced the resources consumption a lil bit :D

kacper-mikolajczak:
Sure @mountiny, testing it right now!

kacper-mikolajczak:
Here are the results @mountiny.

### Mobile (iOS)

From testing on iOS it seems that the app performance is highly coupled to the amount of chats present. It is no surprise that it is influencing performance but the app should not slow down almost exponentially when amount of chats grows.

My first impression is that there are some heavy calculations based on entire data set going on in each item. I am going to investigate it further in some future, as it is out of the scope of the current RP.

Main branch:
Screenshot 2023-10-05 at 11 39 15
Current branch:
Screenshot 2023-10-05 at 11 38 40

### Web
For web, I was not able to load the app at all - no errors or anything particularly wrong. It after inspecting devtools it seems that there are a lot of calculations (probably related to what Matt and Monil discussed on Slack).

mountiny:
Lets gooo

mountiny:
@allroundexperts Thank you! Do you want to raise that here? https://expensify.slack.com/archives/C05LX9D6E07/p1696354046851689 it might be quicker to discuss and more people ot bound the ideas off

kacper-mikolajczak:
Hi @allroundexperts, thanks for your feedback! 👍 
I agree with your concern about growing complexity and while I was thinking about the solution you've suggested, I came to the conclusion that it would eventually lead us to adding more abstraction that we might need at the moment.

In the current approach, we simply couple the OptionRowLHNData with the tiny wrapper in the form of OptionRowLHNDataWithFocus (naming itself is another topic - I am not sure it's correct). That's the whole abstraction we introduce.

In the suggested approach, we would first need to create withCurrentReportFocused HOC, but then we would still need to bound it to the OptionRowLHNData, which will lead us to creating OptionRowLHNDataWithFocus for this purpose. We just simply moved part of the logic (as shouldDisableFocusOptions still needs to be handled inside the wrapper component).

I agree that tightly coupling should be less preferred than composition, but the question is how much do we need to have withCurrentReportFocused abstraction as it is potentially just one line of extracted logic that seems kind of specific. Plus, it looks like it's going to be only used in one place and by sticking to YAGNI rule, we can think about leaving it as-is, wdyt?

Having that said, I am not as familiar with the codebase as you are, please let me know if I understood it right. Thanks a lot for the discussion :smile:

allroundexperts (approved):
As per the [Slack discussion](https://expensify.slack.com/archives/C05LX9D6E07/p1696354046851689), there was an agreement that we should continue using this pattern. As such, the changes are testing good!

mountiny (approved):

PR #26307, opened 2023-08-30
Add DraggableList component

kosmydel:
### Details

Implement a `DraggableList` for waypoints.

### Fixed Issues
$ https://github.com/Expensify/App/issues/22716
PROPOSAL: N/A

### Tests

- [x] Check that draggable list works on all platforms
  - Reorder the waypoints
  - Add more waypoints
  - Change window size
- [x] Verify that no errors appear in the JS console

### Offline tests
N/A

### QA Steps
1. Click on '+' button
2. Click on Request money
3. Choose 'Distance' option
4. Add Start and Finish waypoints
5. Add more waypoints
6. Reorder them with long press
7. Changes should be displayed on the map
8. Save the distance money request.
9. Edit the distance request.

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
[...]

PR #28898, opened 2023-10-05
[HOLD] Revert "Add DraggableList component"

hayata-suenaga:
Reverts Expensify/App#26307

Just in case we need to revert the PR that has the performance regression as a deploy blocker.


hayata-suenaga:
no need for this revert PR anymore

PR #26219, opened 2023-08-29
Allow detaching receipts from transactions

Beamanator:
To be tested with https://github.com/Expensify/Web-Expensify/pull/39018

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/295255

### Tests
Currently for testing, you need this Auth PR: https://github.com/Expensify/Auth/pull/8620

1. Make sure you're on the `scanReceipts` beta (or you have 'all' betas enabled)
2. Log into Newdot, open a chat and click + -> Request Money -> Scan and upload a receipt. It should create a "smartscanning" request
3. Open the Money Requeste and click on the receipt. The Receipt preview modal should show. Click on the `...` in the upper-right corner. There should be no "Remove" option:
Screenshot 2023-09-28 at 2 57 49 PM

4. Close the receipt preview and stop smartscanning, either by completing it in `expensiworks` or manually adding an amount.
5. After that, open the receipt preview again. You should now see the remove option:
Screenshot 2023-09-28 at 2 59 00 PM

6. Click on it. You should see a confirm modal. Click `Delete`:
Screenshot 2023-09-28 at 4 45 52 PM

7. Confirm the receipt is removed from everywhere:
Screenshot 2023-09-28 at 3 00 19 PM
Screenshot 2023-09-28 at 3 00 28 PM
Screenshot 2023-09-28 at 3 00 38 PM

8. Open the transaction again and click on the `...`. Confirm you can now add a receipt:
Screenshot 2023-09-28 at 3 01 56 PM

9. Add it and confirm it all works correctly. Remove it again and confirm it behaves as expected

- [x] Verify that no errors appear in the JS console

### Offline tests
Make sure we optimistically, correctly show that the receipt has been removed

### QA Steps
Same as above

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
[...]

PR #29512, opened 2023-10-12
Revert "Allow detaching receipts from transactions"

jasperhuangg:
Reverts Expensify/App#26219


luacmartins (changes requested):
That's a large PR, what's the blocker it's causing?

PR #30142, opened 2023-10-22
Revert "DRY up the code for the backTo param in routes"

thienlnam:
Reverts Expensify/App#29813


thienlnam:
Straight revert

luacmartins (changes requested):
I think we should go with [this fix](https://github.com/Expensify/App/pull/30144) instead of reverting

Julesssss:
> I think we should go with https://github.com/Expensify/App/pull/30144 instead of reverting

Seems like we did that. Closing

Issue #27939, opened 2023-09-21
[$500] Task-  [Delete message] shows only for the creator after cancel the task


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Login 2 accounts on 2 browsers
2. Both accounts open chat together
3. User A create a task
4. User A click header to back chat report
4. User B open and cancel task
5. From user A Observe task preview disappears but user B still show task report on chat report

## Expected Result:
Deleted task] should show with both users after canceling the task

## Actual Result:
Only the user, who cancels the task see [Delete message]. 
Task preview disappears for users who are not the one who canceled the task.

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [ ] Android / Chrome
- [x] iOS / native
- [x] iOS / Safari
- [x] MacOS / Chrome / Safari
- [x] MacOS / Desktop

**Version Number:** 1.3.72.6
**Reproducible in staging?:** y
**Reproducible in production?:** y
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** Any additional supporting documentation

https://github.com/Expensify/App/assets/93399543/97c758a6-4475-44ad-8f51-08ccf365fc12

https://github.com/Expensify/App/assets/93399543/84a947b2-2cd3-4d15-9294-72d9f181100b

**Expensify/Expensify Issue URL:**
**Issue reported by:** @namhihi237
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1695181722185619

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0132a403c700f46782</li>
        <li>Upwork Job ID: 1704836851861000192</li>
        <li>Last Price Increase: 2023-09-28</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

[...]

PR #30236, opened 2023-10-24
fix canceled task issues

rojiphil:
@abdulrahuman5196 @flodnv 

### Details
This PR makes the following changes to the code:
1) Adds a utility function `shouldHideCanceledTaskReport` here. This function will return true if the given task is canceled and has no visible actions.
2) In `cancelTask`, set the `pendingAction` for the task preview action in Chat Report to `Delete` so that the default behavior is to hide the task preview until the API for task cancelation sends back a response. Since we do not want to hide the task preview for canceled tasks with visible actions, we prevent hiding the task preview by setting `shouldHideOnDelete` accordingly. We will use the utility function `shouldHideCanceledTaskReport` to decide on this.

Note: The extended fix in LHN for deleted tasks was agreed to be fixed later as and when the team raises it. Reference [here](https://github.com/Expensify/App/pull/30236#issuecomment-1822431764)

### Fixed Issues
$ https://github.com/Expensify/App/issues/27939
PROPOSAL: https://github.com/Expensify/App/issues/27939#issuecomment-1729683466

### Tests

1. Add a task in a Chat Report
2. In the Task Report, click on 3dots menu and cancel the task
3. Navigate immediately to the Chat Report
4. Verify the expected behaviour for the below-mentioned problem

Verify Problem:
Actual Behaviour: [Deleted task] message gets displayed in Chat Report Screen
Expected Behaviour: [Deleted task] message should not be displayed in Chat Report Screen

### Offline tests

Same as the Steps for Tests Section.

### QA Steps

Please note that the expected behavior for the issue differs from what is mentioned in the GH issue. 
For QA, the steps mentioned within the Tests Section have to be followed.
The actual behavior is also mentioned in the test description for clarity.
For further clarity, a test video demonstrating the problems is also added here  
<details>
<summary>Video demonstration of the problems</summary>

https://github.com/Expensify/App/assets/3004659/e59599d2-b8fb-4845-856d-f98767436fe3

</details>

### PR Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos

<details>
<summary>iOS</summary>

https://github.com/Expensify/App/assets/3004659/f48c27f4-b17d-4452-b86a-0215910ae244

</details>

<details>
<summary>Android</summary>

https://github.com/Expensify/App/assets/3004659/6d010120-c73d-4456-b412-26d25d033ccb

</details>

<details>

https://github.com/Expensify/App/assets/3004659/e62940ba-bf68-452e-9afd-50606b134ee8

</details>

<details>

https://github.com/Expensify/App/assets/3004659/846aa7bb-c90e-413f-b03d-8d3f3189b410

</details>

<details>
<summary>Web</summary>

https://github.com/Expensify/App/assets/3004659/45438285-2ecb-4519-8c01-3515f640989b

</details>

<details>
<summary>Desktop</summary>

https://github.com/Expensify/App/assets/3004659/78520e39-d73d-4805-99a6-1a0c07b4fdce

</details>

<details>
<summary>Offline-Desktop</summary>

https://github.com/Expensify/App/assets/3004659/e1bfe403-2be6-4609-99c1-1f4f36165de2

</details>


rojiphil:
@abdulrahuman5196
PR checklist added and works well too
Over to you for review

rojiphil:
@abdulrahuman5196 
Gentle bump on review here.

abdulrahuman5196:
Will review in my morning

abdulrahuman5196:
Reviewing now

src/libs/ReportUtils.js
```diff
@@ -213,6 +213,18 @@ function isCanceledTaskReport(report = {}, parentReportAction = {}) {
...
+/**
+ * Checks if a canceled task should be hidden.
+ * When a canceled task does not have any visible actions, it can be hidden in report screen and LHN.
+ *
+ * @param {Object} report - Represents the parent Chat Report of the canceled task.
+ * @param {Object} parentReportAction - Represents the task preview action of the canceled task in Chat Report
+ * @returns {Boolean}
+ */
+function shouldHideCanceledTaskReport(report = {}, parentReportAction = {}) {
+    return isCanceledTaskReport(report, parentReportAction) && !ReportActionsUtils.isDeletedParentAction(parentReportAction);
```
abdulrahuman5196, line 225: @rojiphil How are we finding if the thread has children or not?

rojiphil: @abdulrahuman5196 
The utility function `isDeletedParentAction` internally checks if `childVisibleActionCount` is > 0. That is how it understands that it has children. I think the name `isDeletedParentAction` is misleading. However, the utility function is used in many places and is also in use for a long time. Maybe, the utility function can be renamed whenever a cleanup activity is taken up in the future.

abdulrahuman5196:
Starting to review again

abdulrahuman5196:
@rojiphil Cancelled tasks are not removed from LHN in other devices, unless they are specifically visited.


https://github.com/Expensify/App/assets/46707890/8c6276be-96ea-44ff-a2dd-2aecf2046b95

rojiphil:
> Cancelled tasks are not removed from LHN in other devices, unless they are specifically visited.

@abdulrahuman5196 
That looks like a BE issue. The BE is not sending out pusher updates for `message` due to which `isDeletedParentAction` is not updated. Only after we manually open the task report, the `openreport` API will fetch `message`. Then, it works well.  
Please have a look at the following test video that supports the above mentioned issue.

https://github.com/Expensify/App/assets/3004659/8a78143e-2131-458c-beaf-e40a81990809

abdulrahuman5196:
@rojiphil But do we need to use the isDeleteParentAction, if we just check for cancelledTask and childCount, we should be good here right?

rojiphil:
> But do we need to use the isDeleteParentAction, if we just check for cancelledTask and childCount, we should be good here right?

@abdulrahuman5196 
That’s an interesting thought. I did check on this now. 
I think, while using canceled task and childVisibleActionCount alone may solve the problem, we would still need the BE to send `isDeletedParentAction` as the LHN report name has to change to `[Deleted task]`. That is another issue I observed which will also be fixed. We can see this issue of the report name in LHN not getting updated in the test video that I sent today. 
Further, `isDeletedParentAction` flag is an integral check of canceled task as we can see in the existing function of `isCanceledTaskReport` [here](https://github.com/Expensify/App/blob/a31933a049b9a9aa4222a4d3edab19b54a4555b7/src/libs/ReportUtils.js#L205). So, I doubt if we can do away with `isDeletedParentAction` here.
I think, it is best to keep our code as it is. And, if everything else is fine, we can go ahead with the merge and also raise an issue with BE to fix this.
What are your thoughts?

abdulrahuman5196:
@thienlnam / @flodnv throughs on the above? Seems we might actually need backend fix after all if required to fix LHN

flodnv:
>we would still need the BE to send isDeletedParentAction as the LHN report name has to change to [Deleted task]
>we can go ahead with the merge and also raise an issue with BE to fix this.
>Seems we might actually need backend fix after all if required to fix LHN

Sounds good to me, how about you @thienlnam ?

thienlnam:
We already do send a pusher update to the task report participants for the new task cancelled reportAction 

> we would still need the BE to send isDeletedParentAction as the LHN report name has to change to [Deleted task]

What is this in reference to - what report is missing this parameter?

rojiphil:
>> we would still need the BE to send isDeletedParentAction as the LHN report name has to change to [Deleted task]
>
>What is this in reference to - what report is missing this parameter?

@thienlnam 

**Problem 1**:
Steps:
1) User A creates a Task for User B. (Verify that task is received in User B FE)
2) User A cancels the Task.

**Actual Result**: User B’s Report Screen keeps showing the task even though it was canceled by User A
**Expected Result**: User B’s Report Screen should hide the task as it was canceled by User A.
The test video to this problem is attached to this comment below.

**Problem 2**: In this issue’s context, the problem is shown in the test video in [comment here](https://github.com/Expensify/App/pull/30236/#issuecomment-1787039958). 

Both have the same root cause

**Root Cause**:
The FE depends on `isDeletedParentAction`. We can see this in the implementation of `isCanceledTaskReport` [here](https://github.com/Expensify/App/blob/a31933a049b9a9aa4222a4d3edab19b54a4555b7/src/libs/ReportUtils.js#L205). As the updated `isDeletedParentAction` is not received from BE, this prevents the User B’s FE to reflect the canceled task (i.e. due to User A canceling the task). However, if we trigger the `OpenReport` for the chat report in User B’s FE, the chat report will behave as expected. 

**Solution**:
I think the problem will get solved if BE also sends pusher updates for the updated `message` in the task’s parent action. Since this will contain `isDeletedParentAction` with `true` value, it will update the UI accordingly. 

Hope this helps.

https://github.com/Expensify/App/assets/3004659/f14d06e7-f01a-4ca9-adf2-52552eb52e88

rojiphil:
> Seems we might actually need backend fix after all if required to fix LHN

@abdulrahuman5196  @flodnv @thienlnam 
As it is obvious, the identified problem has been there before. It is just that we identified this problem here. Also, as detailed in the previous comment, this problem has the same root cause as the one we are facing here. Now although I think it may be simple to fix this on the BE side, a good analysis is required if we want to explore solving this from the FE side. Either way, I think it may be better to take this up separately. Sharing my thoughts here if it helps to take this forward.

thienlnam:
Oh you're talking about the ADD_COMMENT reportAction from the parent chat. There was another scenario from having access to a task but not the parentReportAction that created the task. This happens when you assign someone a task, but they don't have access to the report where the task was shared. So to handle that case, we add this field to the report so we can check if the task is deleted. That might also work here, checking if the isDeletedParentAction is on the actual task report. 

I agree we'll also need to do a BE fix to update the main parent reportAction to update the key. But because we don't update all the possible linked reportActions, there might be a couple other scenarios where this doesn't work with different assignees

rojiphil:
> I agree we'll also need to do a BE fix to update the main parent reportAction to update the key. But because we don't update all the possible linked reportActions, there might be a couple other scenarios where this doesn't work with different assignees.

Thanks @thienlnam for bringing clarity on where the fix is required. 

With this clarity in mind, how do we take this PR forward? 
1) Wait for the BE fix and then continue on this PR or
2) Consider the BE fix as separate. And, go ahead with this PR and get this merged if there are no other problems. 

@thienlnam @flodnv   
Please advise

thienlnam:
Let's just hold on the BE PR, and then we can see if there are other scenarios that need to be fixed in this App PR

rojiphil:
> Let's just hold on the BE PR, and then we can see if there are other scenarios that need to be fixed in this App PR

@thienlnam 
Gentle bump on this. 
When do you think the BE PR can be ready for us to take this further here?

thienlnam:
Currently in review - should be live EOW or early next week

rojiphil:
> Currently in review - should be live EOW or early next week

@thienlnam 
Is this still under review? Or is it live already?

thienlnam:
We've had to revert the BE PR due to some other unintended issues that were caused. We've also had a discussion of possibly removing the key. I can spin up another PR later, but this is low priority since the bug is pretty minor

rojiphil:
> We've had to revert the BE PR due to some other unintended issues that were caused. We've also had a discussion of possibly removing the key. I can spin up another PR later, but this is low priority since the bug is pretty minor

@thienlnam 
Thanks for the update. 
Curious to know the next steps with the FE PR as the BE PR's priority is low. 
Also, since this bug is pretty minor, can we go ahead and merge the FE PR if there are no other issues? or wait for the BE PR?
Please let us know

thienlnam:
> Also, since this bug is pretty minor, can we go ahead and merge the FE PR if there are no other issues? or wait for the BE PR?

If I'm misunderstanding but from this [comment](https://github.com/Expensify/App/pull/30236#issuecomment-1788508965), the root cause is in the BE and so we might not even need to move forward with this PR right? 

I looked over the changes and they look fine, but if it's not necessary to fix here then I think we should just close until the BE is resolved later

abdulrahuman5196:
Maybe this point out during C+ approval can provide more information - https://github.com/Expensify/App/issues/27939#issuecomment-1762703334

I think the core fix on the chat page alone can still work with BE changes, only the extended fix for the LHN need BE changes.

Correct me if wrong @rojiphil

thienlnam:
Gotcha, yeah if that's the case then we can just move forward with this

rojiphil:
> I think the core fix on the chat page alone can still work with BE changes, only the extended fix for the LHN need BE changes.

**Core Fix**:
Regarding core fix, I too agree that only FE changes are required. No BE fix is required here.

**Extended Fix**:
Regarding the extended fix, I think FE changes are required. With the current FE changes, the extended fix works smoothly in both local and remote devices except for one specific scenario as described below. 

**Extended Fix - Specific Scenario**: 
This specific scenario was identified in [comment here](https://github.com/Expensify/App/pull/30236/#issuecomment-1786908335)

> Cancelled tasks are not removed from LHN in other devices, unless they are specifically visited.

This scenario happens only on the remote device and is temporary in nature until the report is specifically visited or the report is updated via `OpenReport` API. I think there is no inconsistency in BE. Just that a pusher update would have helped the remote user to experience real-time change. This is where the need for BE fix came up. 

> If I'm misunderstanding but from this [comment](https://github.com/Expensify/App/pull/30236#issuecomment-1788508965), the root cause is in the BE 

For the specific scenario mentioned above, the root cause is in the BE as mentioned. So, while the FE changes are required for core fix and extended fix, BE changes will also address this specific scenario of extended fix. As I understood from [comment here](https://github.com/Expensify/App/pull/30236#issuecomment-1817928424), this specific scenario is a minor bug (low priority) and can be fixed at some later convenient time in BE. 

**Summary**:
My suggestion [here](https://github.com/Expensify/App/pull/30236#issuecomment-1819591308) was to go ahead with the FE PR changes for core fix and extended fix if there are no other issues. 
And, for the specific scenario of extended fix, BE change can be independently taken up.

@thienlnam @abdulrahuman5196 
Does this make sense?

abdulrahuman5196:
@rojiphil I don't think doing the extended fix without BE changes makes sense. It would cause a behavioral difference between current device and remote device which might cause user confusion. 

**I would suggest to go ahead with just the core fix as part of this PR** and the team can create a different issue for the LHN and the team can take on the new issue when it becomes priority.

rojiphil:
> **I would suggest to go ahead with just the core fix as part of this PR** and the team can create a different issue for the LHN and the team can take on the new issue when it becomes priority.

@abdulrahuman5196 
As per your suggestion, I have reverted the extension fix. 
Also, tested the core fix again and works well.
All yours for review

src/libs/ReportUtils.js
abdulrahuman5196: https://github.com/Expensify/App/assets/46707890/8ca22c11-a07b-423f-9501-153f077e46b5



@rojiphil I think even the core fix is affected by BE changes for remote devices on chat page also. Correct me if wrong.

abdulrahuman5196: Am I grasping the core fix correctly? Or am I gauging more fixes not covered as part of core fix?

rojiphil: @abdulrahuman5196 
Just tested the remote scenario for the core fix. Here too, the root cause for the problem at the remote device is due to missing pusher updates. Here is an attached test video for the same.

This problem is there even in the test video of the OP. There, the testing team refreshed the page. I am so, wondering if QA considers this an issue at all. May be, this is why BE team also considers this as a pretty minor bug

https://github.com/Expensify/App/assets/3004659/bfb15401-fb90-461f-9079-d3defdd10902

rojiphil:
Looks like we need more discussion on this.

> I don't think doing the extended fix without BE changes makes sense. It would cause a behavioral difference between current device and remote device which might cause user confusion.

Based on the [analysis here](https://github.com/Expensify/App/pull/30236/files#r1404334398), we have the remote device scenario even for the core issue. Actually, the test video in OP also have this problem. The testing team, however, refreshes the page. 

@abdulrahuman5196  @thienlnam 

1) Can we agree on whether the behavioral difference between the current device and the remote device is serious enough to be fixed now? Or is it minor enough to be fixed later? 
I think the decision to make BE changes now or later depends on the outcome of the above question.

Meanwhile, I also noticed a similar issue raised two days ago [here](https://github.com/Expensify/App/issues/31710#issuecomment-1823014639)

2) Further, as I understand, we need FE changes for the Core Fix and Extended Fix to make it work. And, to test on the remote device, we do have a workaround by either revisiting the report or by refresh. 
So, is it not better to go ahead with the FE changes here? And, create another issue for the BE changes to handle behavioral changes between current and remote device? What am I missing here?

rojiphil:
@abdulrahuman5196 

Since this PR is blocked on the BE issue [here](https://github.com/Expensify/App/pull/30236#issuecomment-1788508965), I was thinking if there is a way to unblock this from the FE. 

Since the root cause is insufficient information being pushed by BE via pusher updates, wondering if a refetch of report via `reconnect` ([here](https://github.com/Expensify/App/blob/af69ba7176d695d40e7c74cc2fe067348ea0ea20/src/libs/actions/Report.ts#L747)) would help when FE receives insufficient information via pusher updates [here](https://github.com/Expensify/App/blob/af69ba7176d695d40e7c74cc2fe067348ea0ea20/src/libs/actions/OnyxUpdates.ts#L51C1-L63C2) as shown in the attached image below.

What do you think?

rojiphil:
@abdulrahuman5196 
Gentle bump for your feedback to the [comment here](https://github.com/Expensify/App/pull/30236#issuecomment-1873174158).

@thienlnam 
Also curious to know your thoughts too to the [comment here](https://github.com/Expensify/App/pull/30236#issuecomment-1873174158)

thienlnam:
Yeah agreed, we shouldn't do a workaround when we know the root cause - that's why we do the RCA in the first place. Unfortunately this is just in a bad position with regards to other priorities right now.

rojiphil:
Closing PR as this is stale.

Issue #31084, opened 2023-11-08
Update MoneyRequestView.js


cead22:
- [Design Doc](https://docs.google.com/document/d/1iGvpCoQ6Wt-i2wcpvaMKsWZRNs-oAsWoaS_nLue7_E0/edit)
- Let's update MoneyRequestView.js
   - The following should only apply to users in the violations beta
   - Let's add a connection to the transactionViolations and policyCategories Onyx collections as shown in the code snippet below, which will go in the `withOnyx` section of the code at the bottom
   - Let's add the new onyx key for the transaction violation collection
   - Add a new transactionViolation prop
   - Pass the transactionViolation to the MoneyRequestView function 
   - For each field that can have a violation, Inside OfflineWithFeedback and below MenuItemWithTopDescription let's add a View that wraps a Text component to show the violation text, as shown in the snippet below
       - The fields are merchant, amount, category, date, tag, comment, billable, receipt, and tax
   - Fields with violations that go into a new page, like a list page for the cate of category and description for instance, should show a RBR circle to the right, in front of the caret `>`. Screenshots below

```js
withOnyx({
    …
    policyCategories: {
        key: ({report}) => `${ONYXKEYS.COLLECTION.POLICY_CATEGORIES}${report.policyID}`,
    },
    transactionViolation: {
        key: ({report}) => {
            const parentReportAction = ReportActionsUtils.getParentReportAction(report);
            const transactionID = lodashGet(parentReportAction, ['originalMessage', 'IOUTransactionID'], 0);
            return `${ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS}${transactionID}`;
        },
    },
...
```
```js
{Boolean(ViolationUtils.getViolationForField(transactionViolation, 'amount')) && (
    <View>
        <Text style={[styles.ph5, styles.textLabelError]}>{ViolationUtils.getViolationForField(transactionViolation, 'amount')}</Text>
    </View>
)}
```

![image](https://github.com/Expensify/App/assets/165133/3c9a024a-b926-4279-b828-81b90773eb2e)
![image](https://github.com/Expensify/App/assets/165133/ad577ccd-b681-4916-b5f8-a7433fe702df)

cead22:
@lindboe @cdanwards can you please comment here so I can assign you?

(@NicMendonca sorry about that ping, that was a mistake)

cdanwards:
[...]

Issue #27262, opened 2023-09-12
Remove ReportActionsUtils.getParentReportAction(), ReportUtils.getParentReport(), TransactionUtils.getLinkedTransaction(), ReportUtils.getPolicy(), and TransactionUtils.getTransaction()


tgolen:
# Problem
These methods are anti-patterns because they are most always used for loading data into a component without using `withOnyx()`. This breaks the data flow of a react application. (data is coming from somewhere that is not props or state and cannot be debugged in react dev tools).

# Solution
Switch all references to properly use `withOnyx()` for components and `connect()` for libs.

### Components
- [x] MoneyRequestHeader @tgolen https://github.com/Expensify/App/pull/28645
- [x] OptionRowLHNData @tgolen https://github.com/Expensify/App/pull/35975
- [x] EditRequestPage @tgolen 
- [x] ~SplitBillDetailsPage~
- [x] extractAttachmentsFromReport @tgolen https://github.com/Expensify/App/pull/30866
- [x] #32309
- [x] #32443
- [x] FlagCommentPage @tgolen https://github.com/Expensify/App/pull/33676
- [x] ReportScreen @tgolen https://github.com/Expensify/App/pull/33970
- [x] ReportActionItem @tgolen https://github.com/Expensify/App/pull/34057
- [x] ~ReportActionList~ No longer references deprecated methods
- [x] withReportAndReportActionOrNotFound @tgolen https://github.com/Expensify/App/pull/34113
- [x] ~ComposerWithSuggestions @tgolen~ already removed\
- [x] `MoneyRequestView` @tgolen  https://github.com/Expensify/App/pull/37703
- [x] `HeaderView` @tgolen https://github.com/Expensify/App/pull/37913
- [x] `ReportActionsListItemRenderer` @tgolen https://github.com/Expensify/App/pull/37449
- [x] FlagCommentPage `getParentReport()` @tgolen https://github.com/Expensify/App/pull/38356
- [x] #38355
- [x] #39091
- [x] #40316
### Libs
- [x] ~IOU~ already removed
- [x] Task @tgolen https://github.com/Expensify/App/pull/37340
- [x] TransactionUtils @tgolen https://github.com/Expensify/App/pull/38258
- [ ] ReportUtils
- [x] #38959
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0156f7187832f56740</li>
[...]

PR #32594, opened 2023-12-06
Update MoneyRequestView.js to show Violations

trevor-coleman:
### Details

This adds Violations warnings and RBR indicators to fields in the `MoneyRequestView`.

#### Notes on Implementation:

1. **Created FieldViolationMessages component** -- The violation messages display is repeated 8 times on
   the `MoneyRequestView` page, which is enough that it should be encapsulated in a component. It takes a single prop --
   the list of violations for that field -- which it then maps to a list of `<Text>` components

2. **`CONST.VIOLATIONS`** -- The list of violation names is defined in the CONST file, and other types are derived from
   there. This way there is a single place to add/remove violations and the type system would provide helpful errors if
   other changes to be made.

3. **Onyx Keys** -- added `ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS` with the value `transactionViolation_` -- this
   seemed to follow the pattern

4. **Prop Types** -- created `propTypes` objects for a `transactionViolation` and array of `transactionViolations` --
   put them in `@libs/Violations` in case some other component needs to use them later.

5. **Clean-up** -- cleaned up some imports I wrote in the ViolationUtils PR to match the repo style. Also added some
   JSDoc comments for things

### Fixed Issues

$ https://github.com/Expensify/App/issues/31084

### Tests

Unfortunately the Violations API is not yet working on the back end, so testing this requires changing the code to add
some test data.

#### Adding Test Violations

In `@src/hooks/useViolations.ts` add the following snippet anywhere below the definition of `violationFields`:

```ts
function getTestViolations({violations, fields}: { violations?: ViolationName[]; fields?: ViolationField[] } = {}) {
    const allViolations = Object.keys(violationFields) as ViolationName[];
    return allViolations
        .filter((key) => violations?.includes(key) ?? true)
        .filter((key) => fields?.includes(violationFields[key]) ?? true)
        .map((key) => ({name: key, type: `violation.${key}.type`, userMessage: `violation.${key}.userData`} as TransactionViolation));
}
```

Then, define a `testViolations` variable and use it in place of `violations` in the `violationsByField` factory
function. For instance this will add all violations on the `amount` field:

```ts
 const violationsByField = useMemo((): ViolationsMap => {
    const violationGroups = new Map<ViolationField, TransactionViolation[]>();
    
    const testViolations = getTestViolations({field: 'amount'});      // <--- ADD THIS (change the field to the one you want to test)

    // for (const violation of violations ?? []) {                     // <--- REPLACE THIS LINE
    for (const violation of testViolations) {                         // <---WITH THIS LINE
        const field = violationFields[violation.name];
        const existingViolations = violationGroups.get(field) ?? [];
        violationGroups.set(field, [...existingViolations, violation]);
    }

    return violationGroups ?? new Map();
}, [violations]);
```

This will allow you to select which violations are shown by redefining `testViolations` in the test.

As no native code has changed you will not need to rebuild the app between tests. Simply adjust the value of
testViolations and it should hot-reload.

#### Billable, Category, and Tags Violations

To test the violations on the Billable, Category and Tag violations, we need to be in a report with a policy -- make
sure the policy has Billable Expenses, Categories, and Tags enabled.

Alternatively, if you have access to the code, you can edit the `MoneyRequestView` component and set the values
of `shouldShowBillable`, `shouldShowCategory`, and `shouldShowTags` in the to `true` to test these violations.

#### Tax Violations

The MoneyRequestView component does not currently support Tax violations, so there is no way to test them.

#### Testing violations are displayed correctly for each field:

1. Set up the test data to include all violations on the `amount` Field:
    ```ts 
       const testViolations = getTestViolations('amount');
    ```
2. Create a new Money Request without a Receipt
  a. Open the app
  b. Click the `+` button
  c. Touch `Create New Money Request`
  d. Select the "Manual" tab
  e. Enter a random amount
  f. Touch the `Next` button
  g. Select a policy to request money from
  h. Enter a description
  i. Touch the `Request ${amount}` button
  j. When you are taken to the chat, touch the Report Preview to open the report
  k. Touch the money request you just created to view it
4. For the `amount` field, check that:
   a. the correct violations are displayed in red text under the associated field
   b. that the RBR indicator is displayed next to the chevron (when applicable)
   c. that the field can be edited as normal
   d. that the screen matches the screenshot below for the `amount` field
5. Repeat for each additional field in the table below. You can do this

   | Field | Violations | RBR Indicator |
   |-----------|-----------------------------------|---------------|
   | tag | dummy.violations.allTagLevelsRequired<br>dummy.violations.missingTag<br>dummy.violations.someTagLevelsRequired<br>dummy.violations.tagOutOfPolicy | :white_check_mark: |
   | merchant | dummy.violations.autoReportedRejectedExpense<br>dummy.violations.customUnitOutOfPolicy<br>dummy.violations.duplicatedTransaction<br>dummy.violations.fieldRequired<br>dummy.violations.nonExpensiworksExpense<br>dummy.violations.rter | :white_check_mark: |
   | billable | dummy.violations.billableExpense | :x:           |
   | receipt | dummy.violations.cashExpenseWithNoReceipt<br>dummy.violations.receiptNotSmartScanned<br>dummy.violations.receiptRequired<br>dummy.violations.smartscanFailed | :x: |
   | category | dummy.violations.categoryOutOfPolicy<br>dummy.violations.missingCategory | :white_check_mark: |
   | amount | dummy.violations.conversionSurcharge<br>dummy.violations.invoiceMarkup<br>dummy.violations.modifiedAmount<br>dummy.violations.overAutoApprovalLimit<br>dummy.violations.overCategoryLimit<br>dummy.violations.overLimit<br>dummy.violations.overLimitAttendee<br>dummy.violations.perDayLimit | :white_check_mark: |
   | date | dummy.violations.futureDate<br>dummy.violations.maxAge<br>dummy.violations.modifiedDate | :white_check_mark: |
   | comment | dummy.violations.missingComment | :white_check_mark: |

#### Test that NO violations are displayed when there are no violations

1. Create a new Money Request with a Receipt
  a. Open the app
  b. Click the `+` button
  c. Touch `Create New Money Request`
  d. Scan or select a receipt image
  e. Select a policy to request money from
  f. Complete each field on the request with any valid input
  g. Touch the `Request` button
  h. Touch the Money Request to navigate to the Money Request View
3. Set the value of `testViolations` to an empty array:
    ```ts 
       const testViolations = [];
    ```
4. For each field in the table above, check that no violations are displayed.

### Offline tests

This PR does not have any offline behaviour, as it only specifies where/how error messages should appear in the display.

### QA Steps

1. Create a new Money Request with a Receipt
  a. Open the app
  b. Click the `+` button
  c. Touch `Create New Money Request`
  d. Scan or select a receipt image
  e. Select a policy to request money from
  f. Complete each field on the request with any valid input
  g. Touch the `Request` button
  h. Touch the Money Request to navigate to the Money Request View
3. View the money request and ensure that it looks the same as usual. The violations should not be displayed as the back
   end is not implemented yet.
4. Touch the fields on the money request to edit them, this behaviour should be the same as usual.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
- [x] If the PR modifies the form input styles:
    - [x] I added `Design` label so the design team can review the changes.

### Screenshots/Videos

<details>

Amount
android-amount

BIllable
android-billable

Category
android-category

Date
android-date

Merchant
android-merchant

Receipt
android-receipt

Tag
android-tag

</details>

<details>
Amount
web-android-amount

Billable
web-android-billable

Comment
web-android-comment

Date
web-android-date

Merchant
web-android-merchant

Receipt
web-android-receipt

Tag
web-android-tag

</details>

<details>

Amount
ios-amount

Billable
ios-billable

Category
ios-category

Comment
ios-comment

Date
ios-date

Merchant
ios-merchant

Receipt
ios-receipt

Tag
ios-tag
</details>

<details>
Amount
ios-web-amount

Billable
ios-web-billable

Category
ios-web-category

Comment
ios-web-comment

Date
ios-web-date

Merchant
ios-web-merchant

Receipt
ios-web-receipt

Tag
ios-web-tag
</details>

<details>
**Chrome**

Amount
web-desktop-amount

Billable
web-desktop-billable

Category
web-desktop-category

Comment
web-desktop-comment

Date
web-desktop-date

Merchant
![web-desktop-merchant]
(https://github.com/Expensify/App/assets/22041394/ed300c94-4ba4-4a51-a974-ad068df7ee9d)

Receipt
web-desktop-receipt

Tag
web-desktop-tag

</details>

<details>

MacOS Desktop is currently crashing on
launch [due to an unrelated issue](https://expensify.slack.com/archives/C01GTK53T8Q/p1701891308625069), so I have been
unable to test it. Given the scope of the changes I am highly confident that there will not be any errors specific to
that platform.

</details>


trevor-coleman:
@cead22 -- this one is ready to go. I couldn't test on the desktop app because it's currently crashing on launch [due to an unrelated issue](https://expensify.slack.com/archives/C01GTK53T8Q/p1701891308625069). But I can add screenshots whenever that gets resolved.

cead22:
Thanks for the detailed Tests section. If `Create a new Money Request with a Receipt` and `Create a new Money Request without a Receipt` are setup steps, can you make it clearer that there aren't any steps after that to verify anything, or add a section for steps to follow that end up in a verification so it's clear what the tests are?

It looks like step 8. under the first one is missing something: `Touch the Money Request to navig`

> 2\. **`CONST.VIOLATIONS`** -- The list of violation names is defined in the CONST file, and other types are derived from
>     there. This way there is a single place to add/remove violations and the type system would provide helpful errors if
>     other changes to be made

Admittedly I liked where this was before, since CONST aren't used to defined types (even though sting literals define types in TS). Would it make sense to move it back, or is there anything about the type system providing helpful errors that we'd lose if we do that?

> 4\. **Prop Types** -- created `propTypes` objects for a `transactionViolation` and array of `transactionViolations` --
>     put them in `@libs/Violations` in case some other component needs to use them later.

Right now there's only one component using this right? Do you anticipate having another component that will use it? If not, can we move the prop type definition to to `src/components/ReportActionItem/MoneyRequestView.js`? I understand the reason for extracting this, but we typically only do that if and when a second place in the code uses the repeated code

```js
const violationNames = Object.values(CONST.VIOLATIONS) as ViolationName[];
...

transactionViolations: PropTypes.shape({
    type: PropTypes.string.isRequired,
    name: PropTypes.oneOf(violationNames).isRequired,
    userMessage: PropTypes.string.isRequired,
    data: PropTypes.objectOf(PropTypes.string),
});
```

src/libs/Violations/propTypes.ts
```diff
@@ -0,0 +1,15 @@
+import PropTypes from 'prop-types';
+import CONST from '@src/CONST';
+import {ViolationName} from '@src/types/onyx';
+
+const violationNames = Object.values(CONST.VIOLATIONS) as ViolationName[];
+
+const transactionViolationPropType = PropTypes.shape({
+    type: PropTypes.string.isRequired,
+    name: PropTypes.oneOf(violationNames).isRequired,
+    userMessage: PropTypes.string.isRequired,
```
cead22, line 10:
```suggestion
```
We don't need this one

src/components/FieldViolationMessages.tsx
cead22, line 1: NAB: Should we name this `src/components/ViolationMessages.tsx` without Field? The billable _field_ will be a toggle/switch, and the receipt _field_ is more of a container

src/components/FieldViolationMessages.tsx
```diff
@@ -0,0 +1,26 @@
+import React, {useMemo} from 'react';
+import {View} from 'react-native';
+import useLocalize from '@hooks/useLocalize';
+import ViolationsUtils from '@libs/Violations/ViolationsUtils';
+import useThemeStyles from '@styles/useThemeStyles';
+import {TransactionViolation} from '@src/types/onyx';
+import Text from './Text';
+
+export default function FieldViolationMessages({violations}: {violations: TransactionViolation[]}) {
```
cead22, line 9:
```suggestion
export default function ViolationMessages({violations}: {violations: TransactionViolation[]}) {
```

src/components/FieldViolationMessages.tsx
```diff
@@ -0,0 +1,26 @@
...
+    const styles = useThemeStyles();
+
+    const localize = useLocalize();
+    const translate = localize.translate;
+
+    const violationMessages = useMemo(() => violations.map((violation) => [violation.name, ViolationsUtils.getViolationTranslation(violation, translate)]), [translate, violations]);
+
```
cead22, line 16:
```suggestion
    const localize = useLocalize();
    const translate = localize.translate;
    const violationMessages = useMemo(() => violations.map((violation) => [violation.name, ViolationsUtils.getViolationTranslation(violation, translate)]), [translate, violations]);
```

src/components/FieldViolationMessages.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        <>
+            {violationMessages.map(([name, message]) => (
+                <View key={`violationMessages.${name}`}>
```
cead22, line 20: Out of curiosity, can we use the `message` in the key instead of the `name`? If so, let's simplify this and only pass `name`, and return an array of names in the useMemo function, instead of an array of `name, message`

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -131,6 +138,9 @@ function MoneyRequestView({report, parentReport, parentReportActions, policyCate
     const shouldShowTag = isPolicyExpenseChat && (transactionTag || OptionsListUtils.hasEnabledOptions(lodashValues(policyTagsList)));
     const shouldShowBillable = isPolicyExpenseChat && (transactionBillable || !lodashGet(policy, 'disabledFields.defaultBillable', true));
 
+    const {getViolationsForField} = useViolations(transactionViolations);
+    const hasViolations = useCallback((field) => canUseViolations && Boolean(getViolationsForField(field).length > 0), [canUseViolations, getViolationsForField]);
```
cead22, line 142:
```suggestion
    const hasViolations = useCallback((field) => canUseViolations() && getViolationsForField(field).length > 0, [canUseViolations, getViolationsForField]);
```

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -190,12 +200,13 @@ function MoneyRequestView({report, parentReport, parentReportActions, policyCate
...
                     </OfflineWithFeedback>
                 )}
-                {!hasReceipt && canEdit && !isSettled && canUseViolations && (
+                {!hasReceipt && canEdit && !isSettled && (
                     <ReceiptEmptyState
                         hasError={hasErrors}
                         onPress={() => Navigation.navigate(ROUTES.EDIT_REQUEST.getRoute(report.reportID, CONST.EDIT_REQUEST_FIELD.RECEIPT))}
                     />
                 )}
+                {canUseViolations && <FieldViolationMessages violations={getViolationsForField('receipt')} />}
```
cead22, line 209: All the occurrences of `canUseViolations` should be function calls `canUseViolations()` right?
```suggestion
                {canUseViolations() && <FieldViolationMessages violations={getViolationsForField('receipt')} />}
```

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -206,9 +217,10 @@ function MoneyRequestView({report, parentReport, parentReportActions, policyCate
                         interactive={canEditAmount}
                         shouldShowRightIcon={canEditAmount}
                         onPress={() => Navigation.navigate(ROUTES.EDIT_REQUEST.getRoute(report.reportID, CONST.EDIT_REQUEST_FIELD.AMOUNT))}
-                        brickRoadIndicator={hasErrors && transactionAmount === 0 ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : ''}
+                        brickRoadIndicator={hasViolations('amount') || (hasErrors && transactionAmount === 0) ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : ''}
                         error={hasErrors && transactionAmount === 0 ? translate('common.error.enterAmount') : ''}
                     />
+                    {canUseViolations && <FieldViolationMessages violations={getViolationsForField('amount')} />}
```
cead22, line 223: NAB: Do we have const/types we can use instead of the string literals for fields like `amount`?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+     * The issue is with the type of {@link PhraseParameters} which is defined as
+     *   `type PhraseParameters<T> = T extends (...args: infer A) => string ? A : never[];`
+     *
+     *  When a union type is passed that _may_ include a key that returns  a `string` or a key that contains a
+     * `function`, the type returns `never`.
+     *
+     *  If you have only switch cases with params, and use a default case to catch the strings, it will throw an error.
+     *
+     * @param violation
+     * @param translate
```
cead22, line 99:
```suggestion
```

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+            key: `${ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS}${transaction.transactionID}`,
+            value: newTransactionViolations,
+        };
+    },
+    /**
+     * Gets the translated message for each violation type. Necessary because `translate` throws a type error if you
+     * attempt to pass it template strings, or members of a string literal union that contains both function and string
+     * members.
+     *
+     * The issue is with the type of {@link PhraseParameters} which is defined as
```
cead22, line 90: Does this `{@link` thing make the editor render this doc nicely, and does it do the same for words wrapped in underscores and backticks? If not, should we remove that since our docs and comments are normally in plain text and not markdown?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+     *   `type PhraseParameters<T> = T extends (...args: infer A) => string ? A : never[];`
+     *
+     *  When a union type is passed that _may_ include a key that returns  a `string` or a key that contains a
```
cead22, line 93: Can you add an example of this to make it super clear?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+     * `function`, the type returns `never`.
+     *
+     *  If you have only switch cases with params, and use a default case to catch the strings, it will throw an error.
+     *
+     * @param violation
+     * @param translate
+     */
+    getViolationTranslation(violation: TransactionViolation, translate: <TKey extends TranslationPaths>(phraseKey: TKey, ...phraseParameters: PhraseParameters<Phrase<TKey>>) => string) {
```
cead22, line 101: Is it possible to define the return type of this function as `string`?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+                return translate('violations.overCategoryLimit', {categoryLimit: violation.data?.categoryLimit ?? ''});
+            case 'perDayLimit':
+                return translate('violations.perDayLimit', {limit: violation.data?.limit ?? ''});
+            default:
+                // The interpreter should never get here because the switch cases should be exhaustive.
+                //
+                // If typescript is showing an error on the assertion below it means the switch statement is out of
+                // sync with the `ViolationNames` type, and one or the other needs to be updated.
+                //
+                // The 'unnecessary' type assertion below is the guarantee of type safety for these translations.
```
cead22, line 187: Why is that? Just curious

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+            case 'overCategoryLimit':
+                return translate('violations.overCategoryLimit', {categoryLimit: violation.data?.categoryLimit ?? ''});
+            case 'perDayLimit':
+                return translate('violations.perDayLimit', {limit: violation.data?.limit ?? ''});
+            default:
+                // The interpreter should never get here because the switch cases should be exhaustive.
+                //
+                // If typescript is showing an error on the assertion below it means the switch statement is out of
+                // sync with the `ViolationNames` type, and one or the other needs to be updated.
+                //
```
cead22, line 186:
```suggestion
                // If typescript is showing an error on the assertion below it means the switch statement is out of
                // sync with the `ViolationNames` type, and one or the other needs to be updated.
```

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+                // The 'unnecessary' type assertion below is the guarantee of type safety for these translations.
+                //
```
cead22, line 188:
```suggestion
```

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+                // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+                return violation.name as never;
```
cead22, line 190: Would it be better to return empty string here?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+     * `function`, the type returns `never`.
+     *
+     *  If you have only switch cases with params, and use a default case to catch the strings, it will throw an error.
+     *
+     * @param violation
+     * @param translate
+     */
+    getViolationTranslation(violation: TransactionViolation, translate: <TKey extends TranslationPaths>(phraseKey: TKey, ...phraseParameters: PhraseParameters<Phrase<TKey>>) => string) {
+        switch (violation.name) {
+            case 'missingCategory':
```
cead22, line 103: The type checker ensures these string literals cannot be something random like 'carlos' right?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+                return translate('violations.missingCategory');
+            case 'fieldRequired':
```
cead22, line 105: Should we order these alphabetically?

src/libs/Violations/ViolationsUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+            case 'rter':
+                return translate('violations.rter', {
+                    brokenBankConnection: Boolean(violation.data?.brokenBankConnection),
+                    isAdmin: Boolean(violation.data?.isAdmin),
+                    email: violation.data?.email,
+                    isTransactionOlderThan7Days: Boolean(violation.data?.isTransactionOlderThan7Days),
+                    member: violation.data?.member,
+                });
+            case 'maxAge':
+                return translate('violations.maxAge', {maxAge: violation.data?.maxAge ?? 0});
```
cead22, line 120: Above we do `Boolean(violation.data?.brokenBankConnection),`, but here we use `??` to fallback to a default value. Can we use the same pattern for all cases, eg, `violation.data?.brokenBankConnection ?? false` or `Number(violation.data?.maxAge)

src/languages/en.ts
```diff
@@ -1990,38 +2004,52 @@ export default {
...
+        nonExpensiworksExpense: 'Non-Expensiworks expense',
+        overAutoApprovalLimit: ({formattedLimitAmount}: ViolationsOverAutoApprovalLimitParams) => `Expense exceeds auto approval limit of ${formattedLimitAmount}`,
+        overCategoryLimit: ({categoryLimit}: ViolationsOverCategoryLimitParams) => `Amount over ${categoryLimit}/person category limit`,
+        overLimit: ({amount}: ViolationsOverLimitParams) => `Amount over ${amount}/person limit`,
+        overLimitAttendee: ({amount}: ViolationsOverLimitParams) => `Amount over ${amount}/person limit`,
+        perDayLimit: ({limit}: ViolationsPerDayLimitParams) => `Amount over daily ${limit}/person category limit`,
+        receiptNotSmartScanned: 'Receipt not verified. Please confirm accuracy.',
+        receiptRequired: ({amount, category}: ViolationsReceiptRequiredParams) => `Receipt required over ${amount} ${category ? ' category limit' : ''}`,
+        rter: ({brokenBankConnection, email, isAdmin, isTransactionOlderThan7Days, member}: ViolationsRterParams) => {
+            if (brokenBankConnection && isAdmin) {
```
cead22, line 2033: I made a mistake in what I sent you, and the message below is for when `$isAdmin` is false. With this fix, the copy below will be for when `brokenBankConnection && !isAdmin` which is right cause the submitter (not the approver/admin) is the one that needs to fix the bank connection, and the copy under `if (brokenBankConnection) {` is also right because in that case `$isAdmin` is true, and the message reflects that the employee, which is the submitter, needs to fix the bank connection
```suggestion
            if (brokenBankConnection && !isAdmin) {
```

src/languages/es.ts
```diff
@@ -2475,38 +2489,56 @@ export default {
...
+        modifiedAmount: 'Importe superior al del recibo escaneado',
+        modifiedDate: 'Fecha difiere del recibo escaneado',
+        nonExpensiworksExpense: 'Gasto no es de Expensiworks',
+        overAutoApprovalLimit: ({formattedLimitAmount}: ViolationsOverAutoApprovalLimitParams) => `Importe supera el límite de aprobación automática de ${formattedLimitAmount}`,
+        overCategoryLimit: ({categoryLimit}: ViolationsOverCategoryLimitParams) => `Importe supera el límite para la categoría de ${categoryLimit}/persona`,
+        overLimit: ({amount}: ViolationsOverLimitParams) => `Importe supera el límite de ${amount}/persona`,
+        overLimitAttendee: ({amount}: ViolationsOverLimitParams) => `Importe supera el límite de ${amount}/persona`,
+        perDayLimit: ({limit}: ViolationsPerDayLimitParams) => `Importe supera el límite diario de la categoría de ${limit}/persona`,
+        receiptNotSmartScanned: 'Recibo no verificado. Por favor, confirma su exactitud',
+        receiptRequired: ({amount, category}: ViolationsReceiptRequiredParams) => `Recibo obligatorio para importes sobre ${amount} ${category ? ' el limite de la categoría de' : ''}`,
```
cead22, line 2516:
```suggestion
        receiptRequired: ({amount, category}: ViolationsReceiptRequiredParams) => `Recibo obligatorio para importes sobre ${category ? 'el limite de la categoría de ' : ''}${amount}`,
```

src/languages/es.ts
```diff
@@ -2475,38 +2489,56 @@ export default {
...
+        rter: ({brokenBankConnection, isAdmin, email, isTransactionOlderThan7Days, member}: ViolationsRterParams) => {
+            if (brokenBankConnection && isAdmin) {
+                return `No se puede adjuntar recibo debido a una conexión con su banco que necesita arreglar`;
```
cead22, line 2519:
```suggestion
            if (brokenBankConnection && !isAdmin) {
                return 'No se puede adjuntar recibo debido a una conexión con su banco que necesitas arreglar';
```

src/languages/es.ts
```diff
@@ -2475,38 +2489,56 @@ export default {
...
+            }
+
+            if (brokenBankConnection) {
+                return `No se puede adjuntar recibo debido a una conexión con su banco que ${email} necesita arreglar`;
+            }
+
+            if (isAdmin && !isTransactionOlderThan7Days) {
+                return `Pídele a ${member} que marque la transacción como efectivo o espere 7 días e intente de nuevo`;
```
cead22, line 2527:
```suggestion
                return `Pídele a ${member} que marque la transacción como efectivo o espera 7 días e intenta de nuevo`;
```

src/languages/es.ts
```diff
@@ -2475,38 +2489,56 @@ export default {
...
+            }
+
+            if (!isTransactionOlderThan7Days) {
+                return `Esperando adjuntar automáticamente con transacción de tarjeta de crédito`;
```
cead22, line 2531:
```suggestion
                return 'Esperando adjuntar automáticamente a transacción de tarjeta de crédito';
```

src/languages/es.ts
```diff
@@ -2475,38 +2489,56 @@ export default {
...
+            }
+
+            return ``;
```
cead22, line 2534:
```suggestion
            return '';
```

cead22 (changes requested):

src/types/onyx/TransactionViolation.ts
```diff
@@ -1,46 +1,33 @@
+import CONST from '@src/CONST';
+
 /**
- * Names of transaction violations
+ * Names of Fields where violations can occur. Derived from `CONST.VIOLATIONS` to maintain a single source of truth.
```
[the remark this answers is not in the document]
trevor-coleman: No this is correct -- I i Fer the type from the contents of the array.

We can do that because CONST is declared 'as const'

src/components/FieldViolationMessages.tsx
trevor-coleman: No we can't use the message as the key. This is the translation key that we use to get the message.

src/components/ReportActionItem/MoneyRequestView.js
trevor-coleman: We can't use types at runtime. Best practice in Typescript is to use string literals because they are type safe and autocomplete will suggest the right answer. They are essentially the same as an enum in both function and DX.

We discussed this in the previous thread about sting literals.

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: When you mouse over a function with JSDoc, the links function as actual links. So you can browse to the docs for those functions through the poppers. I'll provide some examples in the morning.

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: The types are narrowed by each case.

The typescript compiler traces the control flow of the value and modifies the type in branches where the set of possible values is reduced.

https://www.typescriptlang.org/docs/handbook/2/narrowing.html

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: No because this is the guarantee of exhaustiveness.

We want it to be clear that if the string passed in doesn't not match  one of the other cases that something has gone wrong. (And we want that to be a compile time error, not a runtime error.)

I originally omitted the default case, but that also causes a linter error, and I think it's more dangerous overall.

I felt like this was a more important guard because if you weren't paying attention you might see the require-default linter error, and remove it and add a default case.

This would break the guarantee that you will get back the right translation key *and* an object of the right shape. But there would be no errors when you did it.

This is at least conspicuous enough to be noticed, and if someone does something wrong (I.e. Modifies the list of violations without adding a message to be displayed) they compiler will flag this line as an error and they'll come and find the comment and add an entry to the switch statement.

The other way someone would say "Oh this doesn't have a default case, I'll add one" and there would not be any errors at all for any future changes to the list of violations.

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: Yeah -- they are type safe. Think of them as a more efficient enum that you don't need to import or call.

src/hooks/useViolations.ts
```diff
@@ -47,6 +47,12 @@ const violationFields: Record<ViolationName, ViolationField> = {
 
 type ViolationsMap = Map<ViolationField, TransactionViolation[]>;
 
+/**
+ * Hook to access violations for a transaction. Returns `getViolationsForField()`
+ * @example const {getViolationsForField} = useViolations(transactionViolations);
+ * @param violations - Array of {@link TransactionViolation}s
+ * @returns - Object with `getViolationsForField()` callback
```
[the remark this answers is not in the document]
trevor-coleman: I believe JSDoc is required by the style guide.

These comments appear when a developer hovers over the name of the function, and they are useful for giving someone a general idea of what a function does so they don't need to click into the function and read it.

It also will show rich-formatted text, and even include hyperlinks that can be followed.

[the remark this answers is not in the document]
cead22: Cool 👍 on internal repos we don't use formatting, and typically avoid descriptions, @param, @return, etc in JS and PHP if they don't add anything

src/components/ReportActionItem/MoneyRequestView.js
cead22: Awesome, thanks for that info, I'm learning quite a bit about TS best practices 🙇

src/libs/Violations/ViolationsUtils.ts
cead22: 🤩 thanks for the explanation, I love it, particularly this _And we want that to be a compile time error, not a runtime error_

src/types/onyx/TransactionViolation.ts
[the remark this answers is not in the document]
trevor-coleman: image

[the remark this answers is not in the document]
trevor-coleman: Whoops yep.

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: Yeah, it's the best thing about typescript. When your codebase is properly typed, then when you make a change you can just follow the type errors through the code and the compiler shows you what needs to change.


There's obviously a bit of pain as you transition -- like trying to type the translate params, but once it gets to a certain critical mass it makes things so much easier.

src/components/FieldViolationMessages.tsx
trevor-coleman: The `name` is passed as the key prop for the message views, so it needs to be unique.

We know that each violation will only appear once, but there are some violations that could present the same message to the user:

```ts
overLimit: ({amount}: ViolationsOverLimitParams) => `Amount over ${amount}/person limit`,
overLimitAttendee: ({amount}: ViolationsOverLimitParams) => `Amount over ${amount}/person limit`,
```

So we need something to distinguish them to ensure uniqueness.

trevor-coleman: I can probably get away with using the array index to distinguish them, since we're confident they won't be changing frequently.

trevor-coleman: Ah no, they are one step ahead of us:
CleanShot 2023-12-12 at 11 58 10@2x

src/components/ReportActionItem/MoneyRequestView.js
trevor-coleman: I took a look and apparently not. Use permissions returns a map from keys to boolean values:

```
export default function usePermissions(): UsePermissions {
    const betas = useContext(BetasContext);
    return useMemo(() => {
        const permissions: UsePermissions = {};

        for (permissionKey in Permissions) {
            if (betas) {
                const checkerFunction = Permissions[permissionKey];

                permissions[permissionKey] = checkerFunction(betas);
            }
        }

        return permissions;
    }, [betas]);
}
```

I think `canUseViolations()` is one of the checker functions, so the `usePermissions` hook calls it and stores the result in that map with a key that has the same name as the function.

If you highlight `canUsePermissions` and choose 'go to definition' it takes you to the function, but that's not the actual value being returned there.

src/components/ReportActionItem/MoneyRequestView.js
trevor-coleman: Here's an example of the type of error you get if you try to use a value outside of the union:

CleanShot 2023-12-12 at 12 11 28@2x

trevor-coleman: And autocomplete only suggests values compatible with the type.

CleanShot 2023-12-12 at 12 12 44@2x

src/languages/en.ts
trevor-coleman: Ah then we can group the cases:

```ts
       if (brokenBankConnection) {
                return isAdmin
                    ? `Can't auto-match receipt due to broken bank connection which ${email} needs to fix`
                    : "Can't auto-match receipt due to broken bank connection which you need to fix";
        }
 ```

Comment (trevor-coleman) [reply src/languages/en.ts:2033]: And we can do the same below...

```
if (!isTransactionOlderThan7Days) {
                return isAdmin ? `Ask ${member} to mark as a cash or wait 7 days and try again` : 'Awaiting merge with card transaction.';
            }
```

Comment (trevor-coleman) [reply src/languages/en.ts:2033]: And in spanish:

```
rter: ({brokenBankConnection, isAdmin, email, isTransactionOlderThan7Days, member}: ViolationsRterParams) => {
            if (brokenBankConnection) {
                return isAdmin
                    ? `No se puede adjuntar recibo debido a una conexión con su banco que ${email} necesita arreglar`
                    : `No se puede adjuntar recibo debido a una conexión con su banco que necesita arreglar`;
            }
            if (!isTransactionOlderThan7Days) {
                return isAdmin
                    ? `Pídele a ${member} que marque la transacción como efectivo o espere 7 días e intente de nuevo`
                    : `Esperando adjuntar automáticamente con transacción de tarjeta de crédito`;
            }
            return ``;
        },
        ```

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: CleanShot 2023-12-12 at 12 38 35

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: The two cases are different in a few ways:

1. Boolean values don't really need a fallback, since `undefined  === false` returns `true` in Javascript. So undefined is equal to the base case -- there's nothing to fall back from.
2. `Number(violation.data?.maxAge) will be `NaN` when `violation.data?.maxage` is `undefined` so we can't cast it in the same way. 
3. Boolean values will never be rendered into a string visible by the user, while other values will appear in the rendered output. For instance:
```ts
maxAge: ({maxAge}: ViolationsMaxAgeParams) => `Date older than ${maxAge} days`
```
would return `Date older than undefined days` once it goes through the stringification process. 

This is part of why I would much rather set these defaults in the translation file, so they are not divorced from the context of the string they are appearing in. The problem with that is that undefined then becomes an acceptable value, and we lose the type hints if someone reuses these strings elsewhere. 

None of this is ideal, but the type issue with `translate` is forcing us into some compromises.

trevor-coleman: I can add the boolean fallback for clarity though rather than use the cast.

src/libs/Violations/ViolationsUtils.ts
trevor-coleman: Here's a quick example:

```
type Foo = "Foo"
type Bar = "Bar"
type FooBar = Foo | Bar

function doFooBarBaz(fooBarBaz: FooBar) {
    if(fooBar === "foo") {
       const foo:Foo = fooBar  // this is ok!
       const bar:Bar = fooBar  // this will have a type error, because fooBar cannot possibly be 'bar'
       return [foo, bar];
   }

   const foo: Foo =  fooBar    // this will throw an error because if `fooBar === 'foo'` we've already returned above
   const bar:Bar = fooBar   // this will be fine
}
```

Typescript is a lot different than other languages in that the compiler follows the flow of logic through the system and sort of evaluates your code without running it to give really helpful errors.

src/components/ReportActionItem/MoneyRequestView.js
cead22: Nice, thanks for the clarification

trevor-coleman:
> >2. CONST.VIOLATIONS -- The list of violation names is defined in the CONST file, and other types are derived from
there. This way there is a single place to add/remove violations and the type system would provide helpful errors if
other changes to be made
>
> Admittedly I liked where this was before, since CONST aren't used to defined types (even though sting literals define types in TS). Would it make sense to move it back, or is there anything about the type system providing helpful errors that we'd lose if we do that?

I moved this because I needed to access the list of violation names _both_ at runtime, _and_ as part of the type system, but I didn't want to have the same list defined twice. 

Where it was in the OnyxTypes package, everything in that package is only types. There is no runtime information there.  I would need to modify the `index.js` and change the shape of the exports (because it's currently all exported with `export type`) and that felt like too big of a break from the pattern.

If I defined them in the component, then the onyx types would need to import them from the component, and then the component would need to import the onyx types creating a circular dependency.

So then I started looking for a nearest common ancestor that they could both import from, and CONST.VIOLATIONS seemed best.

As for the types themselves, it's important to note: while we can't create runtime values from types  -- because type information is all removed at compilation -- we CAN ask the compiler to infer types from our code at compile time. 

If we have a list of values that we *know* will not change, then we can use `as const` to tell the compiler that, and then we can infer a type from that value. For an object each property is defined not just as the type that it is, but the actual value that is there.

```ts
// If we have a const object declared `as const`:
const foo = {bar: "one", baz: "two"} as const

//then typescript can use that as a type using the `typeof` operator
type  Foo = typeof foo
// this is fine!
const x: Foo = {bar: "one", baz: "two"} 
// this will error
x.bar = "hello"   // S2322: Type string is not assignable to type 'one'
// so will this
const y: Foo = {bar: "one", baz: "two"} // S2322: Type string is not assignable to type 'one'
```
Once we have that type, then we can make other inferences:
```ts
// We can get a type that is the keys of the object:
type FooKey = keyof Foo  // 'bar' | 'baz'
const fooKey: FooKey = "banana" // TS2322: Type "banana" is not assignable to type "bar" | "baz"

// or the type of one of the properties
type Bar = Foo['bar']   // The type Bar is the string literal type 'one'
const bar:Bar  = "hello" // TS2322: Type string is not assignable to type "one"
```

So to get the list of violations, we take the type of the `as const` object, and then indext it with its own keys, and they type we get back is a union of all the types that are properties of the object. 

Since they're all strings, it's a string literal union of all the violation names.

```ts
type ViolationName = (typeof CONST.VIOLATIONS)[keyof typeof CONST.VIOLATIONS];
```

We could also have defined `CONST.VIOLATIONS` as an array, and used `(typeof CONST.VIOLATIONS)[number]` but the pattern in the CONST file seems to be nested objects so I stuck with that.

trevor-coleman:
> > 4. Prop Types -- created propTypes objects for a transactionViolation and array of transactionViolations --
put them in @libs/Violations in case some other component needs to use them later.
>
>Right now there's only one component using this right? Do you anticipate having another component that will use it? If not, can we move the prop type definition to to src/components/ReportActionItem/MoneyRequestView.js? I understand the reason for extracting this, but we typically only do that if and when a second place in the code uses the repeated code

Make sense -- I moved it into the component and deleted the references.

trevor-coleman:
@cead22 -- a couple comments above, but otherwise the changes are all in!

trevor-coleman:
For the testing steps, I just integrated them into the instructions -- they didn't repeat as often as I expected.

cead22:
Thank you for the detailed explanation in [this comment](https://github.com/Expensify/App/pull/32594#issuecomment-1852873230)

src/ONYXKEYS.ts
```diff
@@ -270,10 +270,10 @@ const ONYXKEYS = {
         REPORT_USER_IS_TYPING: 'reportUserIsTyping_',
         REPORT_USER_IS_LEAVING_ROOM: 'reportUserIsLeavingRoom_',
         SECURITY_GROUP: 'securityGroup_',
-        TRANSACTION: 'transactions_',
         TRANSACTION_VIOLATIONS: 'transactionViolations_',
 
         // Holds temporary transactions used during the creation and edit flow
+        TRANSACTION: 'transactions_',
```
cead22, line 276: I think this should go back up with the rest of the collections

src/components/ViolationMessages.tsx
```diff
@@ -0,0 +1,23 @@
...
+import useThemeStyles from '@styles/useThemeStyles';
+import {TransactionViolation} from '@src/types/onyx';
+import Text from './Text';
+
+export default function ViolationMessages({violations}: {violations: TransactionViolation[]}) {
+    const styles = useThemeStyles();
+    const {translate} = useLocalize();
+    const violationMessages = useMemo(() => violations.map((violation) => [violation.name, ViolationsUtils.getViolationTranslation(violation, translate)]), [translate, violations]);
+
+    return (
```
cead22, line 14: NAB, but should we return `''` when there are no violationMessages, so that we don't render empty fragments `<> </>`?
```suggestion
    return violations.length === 0 ? '' : (
```

src/languages/en.ts
```diff
@@ -2001,38 +2015,50 @@ export default {
...
+            if (brokenBankConnection) {
+                return isAdmin
+                    ? `Can't auto-match receipt due to broken bank connection which ${email} needs to fix`
+                    : "Can't auto-match receipt due to broken bank connection which you need to fix";
+            }
+
+            if (!isTransactionOlderThan7Days) {
+                return isAdmin ? `Ask ${member} to mark as a cash or wait 7 days and try again` : 'Awaiting merge with card transaction.';
+            }
+
```
cead22, line 2053: Avoid adding blank lines that aren't preceded by comments
```suggestion
            if (!isTransactionOlderThan7Days) {
                return isAdmin ? `Ask ${member} to mark as a cash or wait 7 days and try again` : 'Awaiting merge with card transaction.';
            }
```

src/libs/ViolationsUtils.ts
```diff
@@ -80,6 +82,134 @@ const ViolationsUtils = {
...
+     *  // this too
+     *  const overLimitsTranslation = translate(`violations.${overLimit.key}`, overLimit.params)
+     *
+     * // but this will throw an error on the params arg  because the type checker can't sufficiently resolve the type.
+     * // Even though the type _should_ be sufficiently narrowed by the time it gets there.
+     *  [missingCategory, overLimit].map({key, params}} => params
+     *        ? translate(`violations.${key}`, params)
+     *        : translate(`violations.$key}`)
+     * ```
+     */ getViolationTranslation(
```
cead22, line 122: I know I asked for an example, so I'm sorry if that was misleading -- I was mainly looking for a line or two with an example that would return `never`. I think this comment is a bit too long, and I would rather seeing those cases being enforced (and sorta automatically documented) in automated tests (not necessarily as part of this PR)

I think we can keep the comment up until the paragraph `Necessary because `translate` throws...`, but I won't block on this, so if you prefer to keep the whole thing, let's fix the format

```suggestion
     *
     * The issue is with the type of {@link PhraseParameters} which is defined as
     *   `type PhraseParameters<T> = T extends (...args: infer A) => string ? A : never[];`
     *
     * This says that if T is a function from some args to a string, then T is the type of the arguments. If T is
     * anything else, then T should not be assigned a value.
     *
     * When a templated key is passed whose final value could be either, the type checker pessimistically assigns it
     * never. There seems to be some block somewhere in the type system that prevents it from fully resolving it.
     *
     *  For instance:
     *   `translate('violations.missingCategory')` resolves to the  string 'Missing category',
     *   `translate('violations.overLimit') resolves to the function `({amount}: ViolationsOverLimitParams) => `Amount
     * over ${amount}/person limit`,
     *
     *  But for some reason, typescript can't correctly infer the types when there is a union of multiple strings:
     *  ```
     *  const missingCategory = {key: 'missingCategory' as ViolationName , params:undefined};
     *  const overLimit = {key: }'overLimit' as ViolationName, params: {amount: 1};
     *
     *  // This works fine:
     *  const missingCategoryTranslation = translate(`violations.${missingCategory.key}`)
     * 
     *  // This too
     *  const overLimitsTranslation = translate(`violations.${overLimit.key}`, overLimit.params)
     *
     * // But this will throw an error on the params arg  because the type checker can't sufficiently resolve the type,
     * // even though the type _should_ be sufficiently narrowed by the time it gets there.
     *  [missingCategory, overLimit].map({key, params}} => params
     *        ? translate(`violations.${key}`, params)
     *        : translate(`violations.$key}`)
     * ```
     */ 
     getViolationTranslation(
```

src/types/onyx/TransactionViolation.ts
```diff
@@ -1,46 +1,34 @@
+import CONST from '@src/CONST';
+
 /**
- * Names of transaction violations
+ * Names of violations.
+ * Derived from `CONST.VIOLATIONS` to maintain a single source of truth.
```
cead22, line 5: 💚

cead22 (changes requested):

src/components/ViolationMessages.tsx
trevor-coleman: I would say no -- a fragment isn't actually a component, so an empty fragment won't create a DOM node or anything. It'll just disappear.

I'm not even sure the fragment is necessary. We can just return the array of nodes.

trevor-coleman:
```tsx
return violationMessages.map(([name, message]) => (
        <View key={`violationMessages.${name}`}>
            <Text style={[styles.ph5, styles.textLabelError]}>{message}</Text>
        </View>
    ));
    ```

**Comment by trevor-coleman** [comment]
@cead22 -- changes are in!

File: src/libs/ViolationsUtils.ts [hunk 32]
```diff
@@ -80,6 +82,103 @@ const ViolationsUtils = {
             value: newTransactionViolations,
         };
     },
+    /**
+     * Gets the translated message for each violation type.
+     *
+     * Necessary because `translate` throws a type error if you attempt to pass it a template strings, when the
+     * possible values could be either translation keys that resolve to  strings or translation keys that resolve to
+     * functions.
+     */ getViolationTranslation(
```
cead22, line 91:
```suggestion
     */ 
     getViolationTranslation(
```

cead22 (changes requested):
Last change 🎉

trevor-coleman:
@cead22 - done

cead22 (dismissed):
All yours @aimane-chnaif

trevor-coleman:
@aimane-chnaif Conflict should be resovled now.

cead22 (approved):

src/components/ViolationMessages.tsx
```diff
@@ -0,0 +1,19 @@
+import React, {useMemo} from 'react';
+import {View} from 'react-native';
+import useLocalize from '@hooks/useLocalize';
+import ViolationsUtils from '@libs/ViolationsUtils';
+import useThemeStyles from '@styles/useThemeStyles';
```
cead22, line 5:
```suggestion
import useThemeStyles from '@hooks/useThemeStyles';
```

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -14,9 +14,11 @@ import Switch from '@components/Switch';
 import tagPropTypes from '@components/tagPropTypes';
 import Text from '@components/Text';
 import transactionPropTypes from '@components/transactionPropTypes';
+import ViolationMessages from '@components/ViolationMessages';
 import withCurrentUserPersonalDetails, {withCurrentUserPersonalDetailsPropTypes} from '@components/withCurrentUserPersonalDetails';
 import useLocalize from '@hooks/useLocalize';
 import usePermissions from '@hooks/usePermissions';
+import useViolations from '@hooks/useViolations';
```
cead22, line 21:
```suggestion
```

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -14,9 +14,11 @@ import Switch from '@components/Switch';
...
 import useStyleUtils from '@hooks/useStyleUtils';
 import useTheme from '@hooks/useTheme';
 import useThemeStyles from '@hooks/useThemeStyles';
```
cead22, line 24:
```suggestion
import useThemeStyles from '@hooks/useThemeStyles';
import useViolations from '@hooks/useViolations';
```

cead22 (changes requested):
Prettier and linter fixes

trevor-coleman:
Fixed.

src/components/ReportActionItem/MoneyRequestView.js
lindboe: This looks like it's still open, but the preview is wrong if you compare with the current file

lindboe:
TSC/prettier fixes pushed

aimane-chnaif (changes requested):

dubielzyk-expensify:
To me it mostly looks good. I'd add `8-12px` or so padding between the last push input and the divider line cause it looks a bit tight at the moment:
CleanShot 2023-12-18 at 08 51 23@2x

It also looks like some of the error messages has more padding than others?
CleanShot 2023-12-18 at 08 53 59@2x

cc @Expensify/design as they'll know this more closely. I could be wrong here

shawnborton:
Definitely agree with Jon to tighten up the spacing here so that the violation feels more closely tied to the input/push row above. Let's make sure that's a global change everywhere too and not just this page.

src/components/ViolationMessages.tsx
```diff
@@ -0,0 +1,19 @@
+import React, {useMemo} from 'react';
```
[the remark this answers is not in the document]
hayata-suenaga: confirming [here](https://expensify.slack.com/archives/C0593T50UHG/p1702937059950879)

trevor-coleman:
@dubielzyk-expensify or @shawnborton  -- can you provide the changes for the styles? 

If they arrive quickly I can get them in here otherwise I'll get @cead22 to create a follow-up issue.

shawnborton:
I think the idea is to make sure the error message is closer to the push input above it. Perhaps just 4px margin below the push row?

dubielzyk-expensify:
Yep agree. 4px should be enough

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -61,6 +91,9 @@ const propTypes = {
     /** The transaction associated with the transactionThread */
     transaction: transactionPropTypes,
 
+    /** Violations detected in this transaction */
+    transactionViolations: transactionViolationsPropTypes,
```
hayata-suenaga, line 95: can you remove `transactionViolationsPropTypes` constant declaration and directly put the following line here?

```
PropTypes.arrayOf(transactionViolationPropType)
```

I think it makes more clear that this prop is an array of objects

trevor-coleman:
How does this look? @shawnborton  @dubielzyk-expensify 



A couple notes:
1.  Not using the built-in `error` prop
2. Negative margin
3. New style: `style.mtn2`

### 1. Not using the built in `error` field

So the MenuItem component has an error field, but currently only accepts a single string, so we can't pass the array of violations into it.

I took a swing at refactoring it to accept an array of components, and it actually looks ok, except that the chevron at the side gets vertically centered over the whole object (including all of the error messages) which stretches it out like this:



Making the change to have that align `flex-start` is probably pretty simple, however this means changing the behaviour of every instance of that component throughout the app.

What's giving me pause is that this would be a pretty significant refactor to a very widely used component. It would require a lot of regression testing to make sure we didn't break anything. 

If it's necessary, I can do it (with @cead22's blessing) but I found a bit of a workaround which will get us the same place with a lot less risk, which brings me to:


### Workaround: Negative Margin

The reason the spacing is off to begin with, is that currently the field is a self contained component, and the violations are displayed below it.

Overriding the bottom margin of the input didn't seem to work -- I imagine there is some minimum height that is set somewhere.

So instead I reduced the space between them by using a negative margin on the top of the violations view.

This is a much less invasive approach, and far less risky than refactoring the component that handles all form inputs everywhere.

### New Style `style.mtn2`

To do this I had to introduce a new style in `spacing.ts`:

```ts
// src/styles/utils/spacing.ts
export default {
  // ...
  mtn2: {
    marginTop: -8,
  },
}
```

It seemed like there was a pattern:

```ts
mt0 = 0
mt1 = 4
mt2 = 8
//...
mt6 = 24

//...

mb0 = 0
mb1 = 4
mb2 = 8
// ...
mb6 = 24
```

and I also saw: 

```ts
mtn6 = -24
```

So i inferred `mtn2 = -8`

This is my first contact with the styling system, so if there is another existing style I should use let me know.


Anyway, would like to hear your thoughts on whether this approach seems ok to you! Thanks!

src/components/ViolationMessages.tsx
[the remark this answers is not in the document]
hayata-suenaga: so the discussion was that we should not ban new components from being written in TypeScript when we can do so. (we're also not enforcing all new components to be written in TypeScript). So this is optional right now.

Let's leave this new component in TypeScript 👍

trevor-coleman:
Resolved merge conflict.

shawnborton:
Visually this looks pretty good to me. Can you show me what the hover style looks like when you hover over a row with an error message? I'm a little worried about the negative margin approach. In Figma, we have it so that the error message just sits underneath the push component, then that whole thing might be wrapped with the hoverable styles:

CleanShot 2023-12-20 at 16 56 07@2x

CleanShot 2023-12-20 at 16 56 13@2x

Let me know if that makes sense!

lindboe:
> In Figma, we have it so that the error message just sits underneath the push component, then that whole thing might be wrapped with the hoverable styles

@shawnborton sorry, this is ambiguous to me. Are you saying you do or do not want the violations message to be included within the hoverable styles?

shawnborton:
We do want them in the hoverable area.

lindboe:
Okay, that's sounding a lot like we should be using the existing `error` field and expanding its behavior as necessary, which wasn't part of the original spec. @cead22 do you agree? Is this PR in a mergeable state, considering the beta flag?

cead22:
Yes, I think we can handle the style issues in a follow up PR. @aimane-chnaif are you available to re-review this PR?

cead22:
Awesome, thank you. Will work on resolving conflicts

cead22:
@aimane-chnaif conflicts are resolved

aimane-chnaif (approved):

trevor-coleman:
Resolved conflicts.

src/languages/en.ts
```diff
@@ -2031,38 +2045,49 @@ export default {
...
+        categoryOutOfPolicy: 'Category no longer valid',
+        conversionSurcharge: ({surcharge}: ViolationsConversionSurchargeParams) => `Applied ${surcharge}% conversion surcharge`,
+        customUnitOutOfPolicy: 'Unit no longer valid',
+        duplicatedTransaction: 'Potential duplicate',
+        fieldRequired: 'Report fields are required',
+        futureDate: 'Future date not allowed',
+        invoiceMarkup: ({invoiceMarkup}: ViolationsInvoiceMarkupParams) => `Marked up by ${invoiceMarkup}%`,
+        maxAge: ({maxAge}: ViolationsMaxAgeParams) => `Date older than ${maxAge} days`,
+        missingCategory: 'Missing category',
+        missingComment: 'Missing comment',
```
cead22, line 2061: I missed this one when I shared the definitive list with you
```suggestion
        missingComment: 'Description required for selected category',
```

cead22 (changes requested):

src/languages/es.ts
```diff
@@ -2518,38 +2532,50 @@ export default {
...
+        categoryOutOfPolicy: 'La categoría ya no es válida',
+        conversionSurcharge: ({surcharge}: ViolationsConversionSurchargeParams) => `${surcharge}% de recargo aplicado`,
+        customUnitOutOfPolicy: 'Unidad ya no es válida',
+        duplicatedTransaction: 'Potencial duplicado',
+        fieldRequired: 'Los campos del informe son obligatorios',
+        futureDate: 'Fecha futura no permitida',
+        invoiceMarkup: ({invoiceMarkup}: ViolationsInvoiceMarkupParams) => `Incrementado un ${invoiceMarkup}%`,
+        maxAge: ({maxAge}: ViolationsMaxAgeParams) => `Fecha de más de ${maxAge} días`,
+        missingCategory: 'Falta categoría',
+        missingComment: 'Falta comentario',
```
cead22, line 2548: Descripción obligatoria para la categoría seleccionada

cead22 (changes requested):

src/languages/es.ts
cead22, line 2548:
```suggestion
        missingComment: 'Descripción obligatoria para categoría seleccionada',
```

cead22 (changes requested):

src/languages/es.ts
trevor-coleman: done!

src/languages/en.ts
trevor-coleman: done!

cead22 (approved):

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -349,5 +403,15 @@ export default compose(
                 return `${ONYXKEYS.COLLECTION.TRANSACTION}${transactionID}`;
             },
         },
+        transactionViolation: {
```
youssef-lr, line 406: there's a typo here which I'm fixing [here](https://github.com/Expensify/App/pull/34479/files) if you can review please @trevor-coleman . I was able to see the violations once I fixed it.

src/languages/en.ts
```diff
@@ -2035,38 +2049,49 @@ export default {
...
+            }
+            if (!isTransactionOlderThan7Days) {
+                return isAdmin ? `Ask ${member} to mark as a cash or wait 7 days and try again` : 'Awaiting merge with card transaction.';
+            }
+
+            return '';
+        },
+        smartscanFailed: 'Receipt scanning failed. Enter details manually.',
+        someTagLevelsRequired: 'Missing tag',
+        tagOutOfPolicy: ({tagName}: ViolationsTagOutOfPolicyParams) => `${tagName ?? ''} no longer valid`,
```
s77rt, line 2091: If `tagName` is not available we should have fallen back to `Tag` instead of an empty string for consistency with `categoryOutOfPolicy` and also becausse `no longer valid` message on its own is not clear

cead22: Same for tax

src/components/ReportActionItem/MoneyRequestView.js
```diff
@@ -349,5 +403,15 @@ export default compose(
...
+            key: ({report}) => {
+                const parentReportAction = ReportActionsUtils.getParentReportAction(report);
```
tgolen, line 408: @trevor-coleman `ReportActionsUtils.getParentReportAction` is a deprecated method. I'm trying to remove these in https://github.com/Expensify/App/issues/27262. Can you please create a PR to clean this up and remove the usage of the deprecated method?

trevor-coleman: Hi, I was a paid contractor and I'm no longer on this project.

cead22: @tgolen can you do this as part of one of your PRs removing the method?

tgolen: I'll add it to the list.

Thanks for responding @trevor-coleman and I hope you come back at some point! 👋

cead22: Thanks!

Issue #35344, opened 2024-01-30
[HOLD for payment 2024-02-07] [HOLD for payment 2024-02-07] [$500] Android - Wallet - App crashes on tapping enable wallet.


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 1.4.33
**Reproducible in staging?:** Y
**Reproducible in production?:** N
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** Applause - Internal Team
**Slack conversation:**

## Action Performed:
1. Launch app
2. Tap profile icon
3. Tap Wallet
4. Tap Enable Wallet

## Expected Result:
App must not crash on tapping enable wallet

## Actual Result:
App crashes on tapping enable wallet

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [x] Android: Native
- [ ] Android: mWeb Chrome
- [ ] iOS: Native
- [ ] iOS: mWeb Safari
- [ ] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

Add any screenshot/video evidence

</details> 

https://github.com/Expensify/App/assets/78819774/b41b112e-4839-4017-be33-2a547664470d

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0165d2a9876eb54d92</li>
        <li>Upwork Job ID: 1752165534277464064</li>
        <li>Last Price Increase: 2024-01-30</li>
    </ul>
</details>

marcaaron:
@lanitochka17 can you confirm that this only happens on Android?

marcaaron:
Hmm seems like possibly some issue with Plaid. This is the crash:

[...]

PR #35346, opened 2024-01-30
Revert "Don't persist error after reconnecting with plaid"

marcaaron:
Reverts Expensify/App#35117

Just testing something for https://github.com/Expensify/App/issues/35344 as this is the only PR I can find related to Plaid


marcaaron:
still reproducible confirming that [this issue](https://github.com/Expensify/App/issues/35344) is not related to the PR being reverted and tested here.

PR #36322, opened 2024-02-12
[Audit][Implementation] getOrderedRaportsIDs

kacper-mikolajczak:
## Actions Points

- [x]  Remove caching
- [x]  Short-circuit `nonArchivedReports` sorting
- [x]  Remove `iouReportAmount` computation
- [ ]  Streamline`getReportName` implementation to avoid complex name retrieval
- [ ]  `ReportUtils.shouldReportBeInOptionList` TODO (not sure what it is supposed to do)
- [ ]  `getDisplayNameForParticipant` optimisations


kacper-mikolajczak:
- [ ] From which commit shall we start? It would be the best to find greatest common ancestor.

kacper-mikolajczak:
> Remove iouReportAmount computation

I am concerned of what is the purpose of the `iouReportAmount` and whether we are 💯 sure that it is not used

PR #34080, opened 2024-01-07
Simplify native attachment gallery paging context and improve code

chrispader:
maybe @tgolen, @pecanoro

### Details

Based on https://github.com/Expensify/App/pull/33756. Check [this](https://github.com/margelo/expensify-app-fork/compare/%40chrispader/restructure-multi-gesture-canvas...margelo:expensify-app-fork:%40chrispader/simplify-native-attachment-carousel) for actual diff.

This PR removes unnecessary prop-drilling and improves how we use the `AttachmentCarouselPagerContext`.

It also improves many of the components used in the (native) attachment gallery and simplifies some rendering logic.

The `MultiGestureCanvas` component is becoming more generic and can therefore be used in other cases more easily. The `Lightbox` component takes control of the adapted behaviour while in an attachment carousel.

### Fixed Issues

$ https://github.com/Expensify/App/issues/36034
PROPOSAL:

### Tests

- [x] Verify that no errors appear in the JS console

1. Go to a report with attachments
2. Open an (image) attachment
3. Check that the attachment carousel is working
4. Check that image gestures/transformations are working (pinching, panning, double tap, swiping)

### Offline tests

None needed.

### QA Steps

- [ ] Verify that no errors appear in the JS console

Same as in Tests.

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
[...]

PR #36524, opened 2024-02-14
Revert "Simplify native attachment gallery paging context and improve code"

tgolen:
Reverts Expensify/App#34080

I accidentally merged the original PR without the reviewer checklist being completed and without all the platforms tested.

There is no need to test this, since it's a straight revert and we want it reverted before it gets pushed to staging.


pecanoro (approved):

pecanoro:
Tim explained in the description why we are reverting this, not an emergency

Issue #35234, opened 2024-01-26
[HOLD for payment 2024-04-15] [Tracking] Metric - App Startup (Performance audit by Callstack)


hurali97:
### Description:

As a part of the https://github.com/Expensify/App/issues/33070, the issue tracks and summarises phases for App Startup Metric. Here we will share the Measurements, Analysis and Implementation details as we progress through the Audit.

hurali97:
Metric: **App Startup**
Phase: **Measurement**
Commit hash: f268c39c331305c49dbb80a0d0e59984289ecb25
Commit hash with setup for audit: 452595f48ac6bce07a69c7f750b0083e8a799171

### Flow:

1. Open the app
2. Splash will be hidden after a few secs
3. Login to the app
4. Then wait for the data to show
5. Close the app and force stop it from the settings
6. Open the app again
7. Splash will be hidden after a long period
8. Data will be visible when the splash hides

For each subsequent measurement, steps 1-3 are not followed. Which means we loop over from step 5-8 for the measurement cycle.

This way we get the app startup of the usual flow that a user faces after the fresh install, uses the app for a while and closes it. And then each time he'll launch the app he follows the same flow as we are following here.

### Markers:

There are already a bunch of markers in place in the Performance class but it's missing some other valuable markers that might come in handy when analysing the measurements. Also, the ending TTI marker wasn't placed precisely as it was being executed even before the splash was hidden. Ideally, we want to end the marker when the splash hides.

- We are only interested in the following markers:
    - `nativeLaunch`
    - `nativeLaunchEnd_To_appCreationStart`
    - `appCreation`
    - `appCreationEnd_To_contentAppeared`
    - `contentAppeared_To_screenTTI`

    And
    - `TTI`

- If we combine the markers without `TTI` we should get the value equivalent to `TTI`. This gives us a chance to analyse each phase in app startup and pin point the root cause for potential fixes.

### Measurements:

<img width="1086" alt="Screenshot 2024-01-26 at 4 46 01 PM" src="https://github.com/Expensify/App/assets/47336142/cc3a9949-9e45-47b3-8e1c-14cd82ae7bf9">


<hr />

We see that `TTI` takes about 50 seconds, which means for that period the user can't interact with the app. This reflects a use case for a really heavy account and can be considered as a worst case scenario. Each improvement considering this scenario will benefit other users as well.

mountiny:
Thanks!

trjExpensify:
[...]

Issue #33070, opened 2023-12-14
[Tracking] Callstack performance audit of the App 


mountiny:
Tracking issue for @adhorodyski and Callstack team performing performance audit of the app.

This is a performance audit of the Expensify app done by Callstack, following the DMAIC methodology. It will consist of 5 phases and result in a report summarising the whole audit.

In the first phase we have defined the following metrics:

- APK size in MB (App Size)
- JS bundle size in MB (App Size)
- App start time in MS (TTI)
- Report screen load time in MS (TTI)
- Sending a message in MS (TTI)
- Average FPS on LHN (FPS)
- Average FPS on Report screen load (FPS, CPU usage)
- Search screen load time in MS (TTI)

adhorodyski:
Adam from Callstack here, I'd like to work on this.

adhorodyski:
This is a performance audit of the Expensify app done by Callstack, following the DMAIC methodology. It will consist of 5 phases and result in a report summarising the whole report.

In the first phase we have defined the following metrics:

- APK size in MB (App Size)
- JS bundle size in MB (App Size)
- App start time in MS (TTI)
- Report screen load time in MS (TTI)
- Sending a message in MS (TTI)
- Average FPS on LHN (FPS)
- Average FPS on Report screen load (FPS, CPU usage)
- Search screen load time in MS (TTI)

adhorodyski:
Current status:
- @adhorodyski is working on the **JS bundle size in MB** metric
- @hurali97 is working on the **APK size in MB** metric

We've conducted the measurements (will post shortly) and are finishing up analysing them, you can expect separate comments with the findings.

adhorodyski:
Metric: **JS bundle size in MB**
Phase: **Analysis**
Commit hash: **c09a3fead81512a4b01ccafff4c736301e0dbe89**

Below is the breakdown of our findings from the analysis process. To sum this up, if we take the associated actions the JS bundle size can be reduced by at least ~11%.

| Name | Platforms | Before | After | % Impact | Notes | Actions |
| --- | --- | --- | --- | --- | --- | --- |
| date-fns locale imports | All | 915 KB | 15 KB | 8% | Locales are being imported from the `date-fns/locale` module using named imports. This causes unused ones to end up in the bundle. | Importing only 'en-GB' and 'es' separately from their respective paths, eg. `date-fns/locale/es` to only include the ones in use. |
[...]

PR #37200, opened 2024-02-26
[Audit][Implementation] - Reduce Redundant Operations

hurali97:
### Details

Part of [Callstack Performance Audit](https://github.com/Expensify/App/issues/33070).

This PR removes the redundant operations like static translation that happens on each loop iteration and this PR also checks beforehand whether the number to parse is a valid phone number and if it is, only then the number is parsed by `parsePhoneNumber`. The reason is that call to `parsePhoneNumber` is costly and should only be executed for the numbers that are valid.

### Fixed Issues

$ https://github.com/Expensify/App/issues/37275
PROPOSAL: https://github.com/Expensify/App/issues/35234#issuecomment-1926911643

### Tests

### QA Steps

### PR Author Checklist

    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
    - [X] MacOS: Desktop
    - [X] I verified that the left part of a conditional rendering a React component is a boolean and NOT a string, e.g. `myBool && <MyComponent />`.
    - [X] I verified any copy / text that was added to the app is grammatically correct in English. It adheres to proper capitalization guidelines (note: only the first word of header/labels should be capitalized), and is approved by marketing by adding the `Waiting for Copy` label for a copy review on the original GH to get the correct copy.
- [X] If the PR modifies the form input styles:
    - [X] I added `Design` label so the design team can review the changes.

### Screenshots/Videos
<details>

https://github.com/Expensify/App/assets/47336142/f984b6a7-f37c-45ed-8453-665a3c831600

</details>

<details>

https://github.com/Expensify/App/assets/47336142/3bf70fe1-7c3d-479b-bea1-7be9f2fea83b

</details>

<details>

https://github.com/Expensify/App/assets/47336142/758dbcf2-0260-4281-8e22-f56cbdefeabc

</details>

<details>

https://github.com/Expensify/App/assets/47336142/4c82bb26-b9ba-477b-98d4-ec5011dc6ab9

</details>

<details>

https://github.com/Expensify/App/assets/47336142/a39eaed6-1428-4795-8f31-daa33fedff45

</details>

<details>

https://github.com/Expensify/App/assets/47336142/9f6a4787-1c39-44dd-86ea-633a434631d9

</details>


src/libs/PersonalDetailsUtils.ts
```diff
@@ -24,9 +24,14 @@ Onyx.connect({
...
 });
 
+const hiddenText = Localize.translateLocal('common.hidden');
+const substringStartIndex = 8;
 function getDisplayNameOrDefault(passedPersonalDetails?: Partial<PersonalDetails> | null, defaultValue = '', shouldFallbackToHidden = true): string {
-    const displayName = passedPersonalDetails?.displayName ? passedPersonalDetails.displayName.replace(CONST.REGEX.MERGED_ACCOUNT_PREFIX, '') : '';
-    const fallbackValue = shouldFallbackToHidden ? Localize.translateLocal('common.hidden') : '';
+    let displayName = passedPersonalDetails?.displayName ?? '';
+    if (displayName.startsWith(CONST.MERGED_ACCOUNT_PREFIX)) {
+        displayName = displayName.substring(substringStartIndex);
```
adhorodyski, line 32: if I understand we want everything past `MERGED_`, doesn't this return slightly incorrectly without the 1st character right after the `_`?

src/libs/PersonalDetailsUtils.ts
```diff
@@ -24,9 +24,14 @@ Onyx.connect({
...
+    }
+    const fallbackValue = shouldFallbackToHidden ? hiddenText : '';
```
adhorodyski, line 34: this function is getting called a lot of times, maybe since we default to the `displayName` being returned (I assume that's the hot path) we can check whether to return it first and only then declare the additional `fallbackValue` constant we fallback to? Might save us some memory being allocated.

src/libs/PersonalDetailsUtils.ts
hurali97: Correct. I missed that some how. Fixed now 👍

src/libs/PersonalDetailsUtils.ts
hurali97: I agree and nice spotting 👍

src/libs/PersonalDetailsUtils.ts
adhorodyski: uff good we spotted that hahah

adhorodyski: Might be worth documenting the `substringStartIndex` constant with this little explanation! It'd be valuable to have this information in a comment - what it helps to strip and why.

src/libs/PersonalDetailsUtils.ts
```diff
@@ -24,10 +24,23 @@ Onyx.connect({
     },
 });
 
+const hiddenText = Localize.translateLocal('common.hidden');
+const substringStartIndex = CONST.MERGED_ACCOUNT_PREFIX.length;
```
adhorodyski, line 28: looks like we're missing a new line in between just for the nice formatting

src/libs/LocalePhoneNumber.ts
```diff
@@ -18,6 +19,10 @@ function formatPhoneNumber(number: string): string {
         return '';
     }
 
+    // do not parse the string, if it doesn't contain the SMS domain and it's not a phone number
+    if (number.indexOf(CONST.SMS.DOMAIN) === -1 && !CONST.REGEX.DIGITS_AND_PLUS.test(number)) {
```
roryabraham, line 23: I wonder if we could improve this at compile time by creating separate types for `EmailLogin` and `SMSLogin`, and making this function accept `SMSLogin` only. Then remove the check in this low-level function? thoughts?

src/libs/PersonalDetailsUtils.ts
```diff
@@ -24,10 +25,28 @@ Onyx.connect({
     },
 });
 
+/**
```
roryabraham, line 28: can you please reserve `/** */` JSDoc-style comments for:

- function signatures
- propType declarations (needed here for use in storybook)

Otherwise, stick to `//` for one-or-two line comments like this one, or `/* */` for larger block comments that aren't JSDocs?

src/libs/PersonalDetailsUtils.ts
```diff
@@ -24,10 +25,28 @@ Onyx.connect({
...
+
 function getDisplayNameOrDefault(passedPersonalDetails?: Partial<PersonalDetails> | null, defaultValue = '', shouldFallbackToHidden = true): string {
-    const displayName = passedPersonalDetails?.displayName ? passedPersonalDetails.displayName.replace(CONST.REGEX.MERGED_ACCOUNT_PREFIX, '') : '';
-    const fallbackValue = shouldFallbackToHidden ? Localize.translateLocal('common.hidden') : '';
-    return displayName || defaultValue || fallbackValue;
+    let displayName = passedPersonalDetails?.displayName ?? '';
+    /**
+     * If the displayName starts with the merged account prefix, remove it.
+     */
+    if (displayName.startsWith(CONST.MERGED_ACCOUNT_PREFIX)) {
```
roryabraham, line 38: Is this really much faster than using `passedPersonalDetails.displayName.replace(CONST.REGEX.MERGED_ACCOUNT_PREFIX, '')`?

I heard recently from @tomekzaw that regexes are known to be much slower in Hermes than in other engines. maybe we can try to address that at the source? I'd love to not have to tip-toe around using super basic regexes in our JS code...

src/libs/ReportUtils.ts
```diff
@@ -773,11 +774,12 @@ function isAnnounceRoom(report: OnyxEntry<Report>): boolean {
     return getChatType(report) === CONST.REPORT.CHAT_TYPE.POLICY_ANNOUNCE;
 }
 
+const chatTypes = [CONST.REPORT.CHAT_TYPE.POLICY_ADMINS, CONST.REPORT.CHAT_TYPE.POLICY_ANNOUNCE, CONST.REPORT.CHAT_TYPE.DOMAIN_ALL];
```
roryabraham, line 777: move this to CONST as `DEFAULT_POLICY_ROOM_CHAT_TYPES`

roryabraham (changes requested):
Since this is a performance-related PR, can you please update the PR description to include benchmarks?

src/libs/ReportUtils.ts
hurali97: Only reason to not move this to `CONST` was that we will now have same string in two places and if we update it in one place, we will have to do it in other place as well.

For eg, below is what the `CONST` will look like if I add that:

```js
const CONST = {
   // This doesn't work !!
    DEFAULT_POLICY_ROOM_CHAT_TYPES: [CONST.REPORT.CHAT_TYPE.POLICY_ADMINS, CONST.REPORT.CHAT_TYPE.POLICY_ANNOUNCE, CONST.REPORT.CHAT_TYPE.DOMAIN_ALL],
   ...otherProperties,
   REPORT: {
        CHAT_TYPE: {
            POLICY_ANNOUNCE: 'policyAnnounce',
            POLICY_ADMINS: 'policyAdmins',
            DOMAIN_ALL: 'domainAll',
            POLICY_ROOM: 'policyRoom',
            POLICY_EXPENSE_CHAT: 'policyExpenseChat',
        },
   }
```

Since the above doesn't work, we will have to add magic strings in `DEFAULT_POLICY_ROOM_CHAT_TYPES`:

```js
const CONST = {
    DEFAULT_POLICY_ROOM_CHAT_TYPES: ['policyAdmins', 'policyAnnounce', 'domainAll'],
   ...otherProperties,
   REPORT: {
        CHAT_TYPE: {
            POLICY_ANNOUNCE: 'policyAnnounce',
            POLICY_ADMINS: 'policyAdmins',
            DOMAIN_ALL: 'domainAll',
            POLICY_ROOM: 'policyRoom',
            POLICY_EXPENSE_CHAT: 'policyExpenseChat',
        },
   }
```

Now, we have same string in two places and each time if we update any, we'll have to update other one as well. Thoughts?

src/libs/LocalePhoneNumber.ts
hurali97: I am not that much familiar with business logic, so I might fancy some insights. From what I understand, we only have one property that holds either Email or Sms/number. How can we differentiate between them? 

I don't know whether we get this property from backend Or we compile the data on front end. In case of latter, we can check whether the property is an email or sms using the logic we have in this PR and add a new property like `isEmail`. Then in `formatPhoneNumber` we can just check if it's email using the new property, we will early return.

But the issue is we will need to pass the `isEmail` property to the function wherever it's called. And just looking in VScode search, it's ~40 places so a lot of refactor. I guess we might be safe with what we have in this PR because it doesn't introduce such huge refactor.

Did I understand correctly?

src/libs/CommonTranslationUtils.ts
```diff
@@ -0,0 +1,49 @@
+import Onyx from 'react-native-onyx';
+import ONYXKEYS from '@src/ONYXKEYS';
+import * as Localize from './Localize';
+
+/**
+ * This file contains common translations that are used in multiple places in the app.
```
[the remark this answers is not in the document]
roryabraham: > The approach that is visible as of now, is to introduce a Map which will hold the translated text for each key and each time we invoke Localize.translateLocal(someKey), we will first check whether it exists in the Map and if it doesn't we translate it and store it

Yes, thanks this sounds like a better approach than essentially doing that same thing in a different lib. There are a number of terms for this approach that I think could be used interchangeably in this context:

- caching
- memoization
- dynamic programming
- probably 42 other terms

but it sounds like we're on the same page 👍🏼

src/libs/LocalePhoneNumber.ts
roryabraham: > From what I understand, we only have one property that holds either Email or Sms/number. How can we differentiate between them?

We could check _once_ at a higher level if a login is an SMS login and narrow the type accordingly.

Less passionate about changing anything here - I think the most important change I'd like to see done in this PR is to move the memoization into the existing localization lib rather than adding "CommonTranslations"

src/libs/PersonalDetailsUtils.ts
roryabraham: > However, using the applied solution startsWith only appears to be slightly faster, so it's a matter of discussion whether we should use it or not, since it's only slightly faster

benchmarks would make it easier to discuss this. How much is "slightly faster"?

> I very much like this approach but it's not straight forward to fix performance issues in Hermes. We can take it as an action item from this PR and then follow up on that and also remove the startsWith from this PR. Wdyt?

Sure. I'm sure it would be difficult to improve regex performance in Hermes, but let's look into it. He might hate me for suggesting this, but it seems like it could be a good task for @fabioh8010, since he's worked on regexes in Hermes in the past.

src/libs/ReportUtils.ts
roryabraham: you could also just add separate const for the duplicated strings above the `const CONST = {...}`, right? Or does that break the const-ness in TS?

roryabraham:
marking this as WIP at least until the caching/memoization/dynamicProgramming has been moved to the existing translation lib

src/libs/PersonalDetailsUtils.ts
fabioh8010: Hi @roryabraham @hurali97 It looks like an interesting challenge 😅 And definitely more difficult than the indices flag one.

I would suggest creating a follow-up GH issue so we can take a look at it in the future, when I have more capacity to do so. Also, would be great if we could put all evidence we have that Regexes in Hermes are very slow compared to other engines.

src/libs/Localize/index.ts
```diff
@@ -93,11 +93,46 @@ function translate<TKey extends TranslationPaths>(desiredLanguage: 'en' | 'es' |
...
+
+    // If the phrase is already translated, return the translated value
+    if (valueFromCache) {
+        return valueFromCache;
+    }
+    const translatedText = translate(preferredLocale, phrase, ...variables);
+
+    // We don't want to store translated values for phrases with variables
+    if (isVariablesEmpty) {
+        translatedValues.set(key, translatedText);
```
roryabraham, line 133: NAB but I wonder if it would make sense to do the cache set operation async so that it doesn't delay the return of `translatedText`:

```suggestion
        setImmediate(() => translatedValues.set(key, translatedText));
```

src/libs/Localize/index.ts
```diff
@@ -93,11 +93,46 @@ function translate<TKey extends TranslationPaths>(desiredLanguage: 'en' | 'es' |
...
+ * {
+ *  "name_en": "Name",
+ *  "name_es": "Nombre",
+ * }
+ *
+ * Note: We are not storing any translated values for phrases with variables,
+ * as they have higher chance of being unique, so we'll end up wasting space
+ * in our cache.
+ */
+const translatedValues = new Map<string, string>();
```
roryabraham, line 111: It's probably more efficient to have a separate cache for each locale.

src/libs/Localize/index.ts
hurali97: I think let's not introduce fixes for anything which has not performed worse as of yet. There's a chance that this might introduce any regressions as well 🤔

src/libs/PersonalDetailsUtils.ts
hurali97: @fabioh8010 Thank you 👍 

I will create an issue shortly and update it all of the relevant information. Just one thing, I will still keep the following check in the codebase because even if we fix the RegEx performance, we are still trying to replace a substring in a string which doesn't contain it, hence the check for `startsWith` is in place.

```ts
    // If the displayName starts with the merged account prefix, remove it.
    if (displayName.startsWith(CONST.MERGED_ACCOUNT_PREFIX)) {
        displayName = displayName.substring(substringStartIndex);
    }
```

We should only be using replace on a string which contains that substring.

src/CONST.ts
```diff
@@ -41,6 +41,7 @@ const KEYBOARD_SHORTCUT_NAVIGATION_TYPE = 'NAVIGATION_SHORTCUT';
 const cardActiveStates: number[] = [2, 3, 4, 7];
 
 const CONST = {
+    MERGED_ACCOUNT_PREFIX: 'MERGED_',
```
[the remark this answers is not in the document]
hurali97: We need it as the other one we have includes RegExp and we only need the substring `MERGED_`.

hurali97:
@youssef-lr Can you proceed with the review ? 🙂

src/libs/PersonalDetailsUtils.ts
youssef-lr: Not sure why we need the extra variable, why not just use  `displayName = displayName.substring(CONST.MERGED_ACCOUNT_PREFIX.length);`?

youssef-lr: bump on my question above @hurali97 please

youssef-lr (dismissed):
LGTM! Left a minor comment [here](https://github.com/Expensify/App/pull/37200/files#r1516044437). All yours @roryabraham

src/libs/Localize/index.ts
```diff
@@ -93,11 +94,69 @@ function translate<TKey extends TranslationPaths>(desiredLanguage: 'en' | 'es' |
...
+    }
+}
+
 /**
  * Uses the locale in this file updated by the Onyx subscriber.
  */
 function translateLocal<TKey extends TranslationPaths>(phrase: TKey, ...variables: PhraseParameters<Phrase<TKey>>) {
-    return translate(BaseLocaleListener.getPreferredLocale(), phrase, ...variables);
+    const preferredLocale = BaseLocaleListener.getPreferredLocale();
+    const key = `${phrase}`;
```
roryabraham, line 138: Just seeing this line, I think it's weird how we're casting a more specific `TranslationPaths` variable to a plain string. I recommend refactoring this like so:

```diff
diff --git a/src/libs/Localize/index.ts b/src/libs/Localize/index.ts
index 72bb4dd89b..9d31bf27bc 100644
--- a/src/libs/Localize/index.ts
+++ b/src/libs/Localize/index.ts
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import * as RNLocalize from 'react-native-localize';
 import Onyx from 'react-native-onyx';
+import type {ValueOf} from 'type-fest';
 import Log from '@libs/Log';
 import type {MessageElementBase, MessageTextElement} from '@libs/MessageElement';
 import Config from '@src/CONFIG';
@@ -101,49 +102,35 @@ function translate<TKey extends TranslationPaths>(desiredLanguage: 'en' | 'es' |
  * The data is stored in the following format:
  *
  * {
- *  "name": "Name",
+ *   "en": {
+ *     "name": "Name",
+ *   }
  * }
  *
  * Note: We are not storing any translated values for phrases with variables,
  * as they have higher chance of being unique, so we'll end up wasting space
  * in our cache.
  */
-const TRANSLATED_VALUES_EN = new Map<string, string>();
-const TRANSLATED_VALUES_ES = new Map<string, string>();
-const TRANSLATED_VALUES_ES_ES = new Map<string, string>();
-const TRANSLATED_VALUES_ES_ONFIDO = new Map<string, string>();
-
-/**
- * Returns the map for the given locale.
- */
-function getTranslatedValuesMap(locale: BaseLocale) {
-    switch (locale) {
-        case CONST.LOCALES.ES_ES:
-            return TRANSLATED_VALUES_ES_ES;
-        case CONST.LOCALES.ES_ES_ONFIDO:
-            return TRANSLATED_VALUES_ES_ONFIDO;
-        case CONST.LOCALES.ES:
-            return TRANSLATED_VALUES_ES;
-        case CONST.LOCALES.DEFAULT:
-        default:
-            return TRANSLATED_VALUES_EN;
-    }
-}
+const translationCache = new Map<ValueOf<typeof CONST.LOCALES>, Map<TranslationPaths, string>>(
+    Object.values(CONST.LOCALES).reduce((cache, locale) => {
+        cache.push([locale, new Map()]);
+        return cache;
+    }, [] as Array<[ValueOf<typeof CONST.LOCALES>, Map<TranslationPaths, string>]>),
+);
 
 /**
  * Uses the locale in this file updated by the Onyx subscriber.
  */
 function translateLocal<TKey extends TranslationPaths>(phrase: TKey, ...variables: PhraseParameters<Phrase<TKey>>) {
     const preferredLocale = BaseLocaleListener.getPreferredLocale();
-    const key = `${phrase}`;
     const isVariablesEmpty = variables.length === 0;
 
-    // Get the map for the preferred locale
-    const map = getTranslatedValuesMap(preferredLocale);
+    // Get the cache for the preferred locale
+    const cacheForLocale = translationCache.get(preferredLocale);
 
     // Directly access and assign the translated value from the cache, instead of
     // going through map.has() and map.get() to avoid multiple lookups.
-    const valueFromCache = map.get(key);
+    const valueFromCache = cacheForLocale?.get(phrase);
 
     // If the phrase is already translated, return the translated value
     if (valueFromCache) {
@@ -154,7 +141,7 @@ function translateLocal<TKey extends TranslationPaths>(phrase: TKey, ...variable
     // We don't want to store translated values for phrases with variables
     if (isVariablesEmpty) {
         // We set the translated value in the cache
-        map.set(key, translatedText);
+        cacheForLocale?.set(phrase, translatedText);
     }
     return translatedText;
 }

```

src/libs/PersonalDetailsUtils.ts
```diff
@@ -24,15 +24,27 @@ Onyx.connect({
...
+    }
+
     if (shouldAddCurrentUserPostfix && !!displayName) {
         displayName = `${displayName} (${Localize.translateLocal('common.you').toLowerCase()})`;
     }
 
-    const fallbackValue = shouldFallbackToHidden ? Localize.translateLocal('common.hidden') : '';
-
-    return displayName || defaultValue || fallbackValue;
+    // If displayName exists, return it early so we don't have to allocate
```
roryabraham, line 42: NAB but I don't think this comment makes much sense without the context of the code that was here before, so I think it would be fine to remove it.

roryabraham (changes requested):
Otherwise, LGTM. though I do want to wait for @fedirjh to review, test, and complete the checklist before we merge this

src/libs/PersonalDetailsUtils.ts
hurali97: @youssef-lr I added the extra variable just for better readability. I can remove the variable and add the comment instead 👍

src/libs/Localize/index.ts
hurali97: @roryabraham Thank you for simplifying it 🙏

roryabraham:
Cool, now just waiting for @fedirjh to complete the checklist, which [he said he will do soon](https://expensify.slack.com/archives/C05LX9D6E07/p1710171738756239?thread_ts=1710170978.313699&cid=C05LX9D6E07).

src/libs/Localize/index.ts
```diff
@@ -93,11 +94,54 @@ function translate<TKey extends TranslationPaths>(desiredLanguage: 'en' | 'es' |
...
+    Object.values(CONST.LOCALES).reduce((cache, locale) => {
+        cache.push([locale, new Map()]);
+        return cache;
+    }, [] as Array<[ValueOf<typeof CONST.LOCALES>, Map<TranslationPaths, string>]>),
+);
+
 /**
  * Uses the locale in this file updated by the Onyx subscriber.
  */
 function translateLocal<TKey extends TranslationPaths>(phrase: TKey, ...variables: PhraseParameters<Phrase<TKey>>) {
```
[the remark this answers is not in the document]
roryabraham: good catch @fedirjh, 100% agree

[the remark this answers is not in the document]
hurali97: @fedirjh Whoops, nice one 👍 I will look into this today !!

hurali97:
@fedirjh I have addressed your feedback regarding moving cache to `translate` function. Can you re-review the implementation?

Also, below we can see that we save around ~1344 times translating some phrases that's already translated. Again, thanks for highlighting the right place for the cache 💯 


Screenshot 2024-03-12 at 2 27 12 PM

src/libs/Localize/index.ts
```diff
@@ -58,16 +81,43 @@ type Phrase<TKey extends TranslationPaths> = TranslationFlatObject[TKey] extends
...
+    // Directly access and assign the translated value from the cache, instead of
+    // going through map.has() and map.get() to avoid multiple lookups.
+    const valueFromCache = cacheForLocale?.get(phraseKey);
+
+    // If the phrase is already translated, return the translated value
+    if (valueFromCache) {
+        return valueFromCache;
+    }
+
     let translatedPhrase = translations?.[language]?.[phraseKey] as Phrase<TKey>;
```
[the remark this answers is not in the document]
roryabraham: > If that's the case, why don't we modify the structure of the translations to return a map instead of a plain object in this file?

That makes sense to me. I'm also a bit confused by why `translate` is slow, since it should just be a direct lookup in an object?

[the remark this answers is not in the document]
hurali97: @fedirjh I will look into this today and see whether changing translation structure to Map performs better than what we have now 👍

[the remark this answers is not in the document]
hurali97: @fedirjh @roryabraham  Here is an update on how things looks like with:

- translations with object structure ( what we have in main )
- translations with Map structure ( the proposed change by @fedirjh  )
- translations with object structure with cache applied ( what we have right now in this PR )

Below is a video attached which shows the Hermes traces on a heavy account ( Release + Android real device ). For the same section of the trace, we observe that:

- translations with object structure ( before translations Tab in video ) has 11 calls to translate method
- translations with map structure ( after translations with Map Tab in video ) has 5 calls to translate method
- translations with object structure with cache applied ( after translations with cache Tab in video ) has 3 calls to translate method

https://github.com/Expensify/App/assets/47336142/9cfa68cb-7d61-4ce4-be02-f61883331f09


One thing to note here is that "I only created a Map of `en` translations with only 300 entries, whereas we have around 2000 entries in object structure." to test the map structure but since it had more calls than cache implemented, it's safe to say that it's not worth to change object to map.

Below is another video, where we can see the cache implemented and see that how continuously we are accessing the huge translations object to get the desired value for the key. I just want to emphasise that since the cache Map is marginally small as compared to the translations object and that's why we have so less recordings of the `translate` function in Hermes trace, when cache is enabled. It's to note that the `translate` function will still be called but since we are returning from cache and it's quick enough to not get caught in the Hermes trace and the times it get caught was 3.

https://github.com/Expensify/App/assets/47336142/29ea54b5-66ed-4ed8-87c4-9d7607c99860

src/libs/Localize/index.ts
```diff
@@ -58,16 +81,43 @@ type Phrase<TKey extends TranslationPaths> = TranslationFlatObject[TKey] extends
...
     translatedPhrase = translations?.[languageAbbreviation]?.[phraseKey] as Phrase<TKey>;
     if (translatedPhrase) {
-        return typeof translatedPhrase === 'function' ? translatedPhrase(...phraseParameters) : translatedPhrase;
+        if (typeof translatedPhrase === 'function') {
+            return translatedPhrase(...phraseParameters);
+        }
+
+        // We set the translated value in the cache only for the phrases without parameters.
+        cacheForLocale?.set(phraseKey, translatedPhrase);
+        return translatedPhrase;
```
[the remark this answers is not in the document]
hurali97: @fedirjh Oh sorry, I missed this. I will take a look now 👍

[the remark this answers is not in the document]
hurali97: @fedirjh I have incorporated this change and also added the JS docs comments for the new `getTranslatedPhrase` function. Thank you for the feedback, appreciate it.

fedirjh (approved):
Looks good to me.

cc @roryabraham

roryabraham (approved):

Issue #39108, opened 2024-03-27
[HOLD for payment 2024-05-02] [Wave Collect][QBO] Create a HOC to call `OpenAccountingPage` and then pass the `report.connections`


hayata-suenaga:
Based on [the discussion](https://expensify.slack.com/archives/C03TQ48KC/p1711387863433689), implement a Higher-Order Component (HOC) that uses `useOnyx` to subscribe to policy data, then passes this data to the component it wraps. The HOC should inspect the `connections` field of the policy. If this field lacks a value in the policy data, the HOC should call `OpenAccountingPage`.

The HOC also should render the loading indicator or offline screen if necessary.
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~019f138fc4382cdf6a</li>
        <li>Upwork Job ID: 1773765815925579776</li>
        <li>Last Price Increase: 2024-03-29</li>
    </ul>
</details>

hayata-suenaga:
changing priority back to daily as this has the high priority

strepanier03:
@hayata-suenaga - I am holding off labeling this Internal unless you give the go-ahead. I don't want to assign a C+ to do PR review when it's not needed.

hayata-suenaga:
I decided to have a C+ review the PR just in case 😄

hayata-suenaga:
I realized that a C+ is already assigned to the PR. I'll assign them to this issue.

strepanier03:
Thank you @hayata-suenaga 🙌

trjExpensify:
Is the PR for this now on hold for #39652, @hayata-suenaga? If so, let's have that reflected in the title.

strepanier03:
PR is deployed to staging.

hayata-suenaga:
The PR is actually still being worked on 🙇

strepanier03:
@hayata-suenaga - sorry for the confusion, I meant the PR that this was on hold for is deployed to staging, not the PR for the fix for this issue 🤦‍♀️

hayata-suenaga:
ah haha sorry for my misunderstanding 🙇 🙈

strepanier03:
@hayata-suenaga - I was having a hard time figuring out the price here based on the PR that's linked. Is this a $500 payment or a $250 PR review payment?

strepanier03:
nvm, @hayata-suenaga - payment summary finally posted and I got my answer.

[...]

PR #38400, opened 2024-03-15
QBO import main page

narefyev91:
### Details
Initial implementation for QBO Import main page. Current PR for now blocked by this one [PR](https://github.com/Expensify/App/pull/38283/files). But for now code can be started to be reviews from C+ cc @hayata-suenaga. Also will be nice to simulate somehow syncClasses, syncCustomers, syncLocations, syncAccounts, syncTaxes - from API side - to get test data

Screenshot 2024-03-15 at 13 34 23

### Fixed Issues

$ https://github.com/Expensify/App/issues/37777
PROPOSAL:

### Tests
1. Navigate to /settings/workspaces/:workspaceID/accounting/quickbooks-online/import
2. Verify that menu items for Import are shown
3. Click on each menu item - verify that each of them has toggle to disable/enable feature - verify that click interaction is working
4.  Go back to all menu items - verify that disabling/enabling toggle - changed text from "Not Imported" - to "Imported"/"Imported as a tag"

- [x] Verify that no errors appear in the JS console

### Offline tests
1. Navigate to /settings/workspaces/:workspaceID/accounting/quickbooks-online/import
2. Verify that menu items for Import are shown
3. Click on each menu item - verify that each of them has toggle to disable/enable feature - verify that click interaction is working
4.  Go back to all menu items - verify that disabling/enabling toggle - changed text from "Not Imported" - to "Imported"/"Imported as a tag"

### QA Steps
1. Navigate to /settings/workspaces/:workspaceID/accounting/quickbooks-online/import
2. Verify that menu items for Import are shown
3. Click on each menu item - verify that each of them has toggle to disable/enable feature - verify that click interaction is working
4.  Go back to all menu items - verify that disabling/enabling toggle - changed text from "Not Imported" - to "Imported"/"Imported as a tag"

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
[...]

PR #39132, opened 2024-03-27
Add new HOC to get `policy.connections` data

hayata-suenaga:
cc: @tgolen @aldo-expensify

### Details
This PR does two things:

- Adds a new action function that calls the new `OpenPolicyAccountingPage` API command
- Adds a new HOC component that does the following things:

1. Get the current `policy` object
2. Checks if `policy.connections` is set
3. If `policy.connections` is not set, invoke the new action function `openPolicyAccountingPage` to get data that populates `policy.connections`
4. While the `policy.connections` is not set and being fetched, return the loading indicator.
5. If the `policy.connections` is not set and the device is offline, display the offline screen

_We also store a boolean value inside Onyx to remember if the connections data for the given policy has been attempted to be fetched in the current user session. If the data is fetched previously but the connections field of the policy object is `null` or `undefined`, that means that the data doesn't exist on the backend._

### Fixed Issues

$ https://github.com/Expensify/App/issues/39108
PROPOSAL: N/A

### Tests / Offline Tests / QA Steps

No QA. No test. This PR just adds the HOC component. This PR does not add a code to use the HOC. We can test the HOC in a PR that first uses it.

### PR Author Checklist

_Because there is no way to test this PR, there is no screenshot to add_
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/actions/PolicyConnections.ts
```diff
@@ -0,0 +1,15 @@
+import * as API from '@libs/API';
+import type {OpenPolicyAccountingPageParams} from '@libs/API/parameters';
+import {READ_COMMANDS} from '@libs/API/types';
+
+function openPolicyAccountingPage({policyID}: {policyID: string}) {
```
tgolen, line 5: This can be simplified to only pass `policyID` and it doesn't need to accept an object.

src/pages/workspace/withPolicyConnections.tsx
```diff
@@ -0,0 +1,55 @@
...
+import type {WithPolicyProps} from './withPolicy';
+
+type WithPolicyConnectionsProps = WithPolicyProps;
+
+function withPolicyConnections(WrappedComponent: ComponentType<WithPolicyConnectionsProps>) {
+    function WithPolicyConnections({policy, policyMembers, policyDraft, policyMembersDraft, route}: WithPolicyConnectionsProps) {
+        const {isOffline} = useNetwork();
+
+        useEffect(() => {
+            if (!policy?.connections || policy?.connections) {
```
tgolen, line 16: How does this logic work? I read it as "if there are no connection OR if there are connections, then return early".

Could you please add a code comment here to explain the intention of the `useEffect()`? Be sure to mention the difference between active policies already having `connections` and non-active policies not having them.

tgolen (changes requested):

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: I added a comment in https://github.com/Expensify/App/pull/39132/commits/8e32fded9dff0a2401b310b4a31efeee4708d9d8

hayata-suenaga: > Be sure to mention the difference between active policies already having connections and non-active policies not having them.

Is it accurate to say that the active policy is the one selected in the Workspace Switcher or the one whose workspace settings are displayed? In other words, is the active policy the one whose ID is displayed in the URL?

Given this definition of an active policy, your comment above confuses me a bit. The withPolicyConnections HOC uses the withPolicy HOC in its implementation, which implies that the policy under consideration is always the active one.

src/libs/actions/PolicyConnections.ts
hayata-suenaga: I changed the parameter in https://github.com/Expensify/App/pull/39132#:~:text=fix%3A%20conditional%20statement%20%26%20change%20the%20parameter%20from%20object%20to%20sing%E2%80%A6

src/pages/workspace/withPolicyConnections.tsx
tgolen: Hm, no... what I mean by "active policy" is the policy that is set to be the default policy on the OldDot Workspace page. This is the policy that is sent to NewDot in `OpenApp` and it should contain the connections.

hayata-suenaga: ah I apologize for my misunderstanding 🙇 

I still think that we don' have to distinguish between active and non-active workspaces though 🤔

src/pages/workspace/withPolicyConnections.tsx
```diff
@@ -0,0 +1,57 @@
...
+        const {isOffline} = useNetwork();
+
+        useEffect(() => {
+            if (!policy?.areConnectionsEnabled || !!policy?.connections || !policy?.id) {
+                return;
+            }
+
+            // If the policy doesn't have the connections data when it has the accounting connections feature enabled,
+            // call the API to get the data
+            openPolicyAccountingPage(policy.id);
```
hayata-suenaga, line 22: @aldo-expensify, is it accurate to say that the `connections` object is empty when the user hasn't connected to any accounting software?

If that is the case, this `useEffect` will keep calling `openPolicyAccountingPage` indefinitely for workspaces that have the accounting feature enabled but that haven't connected to any accounting software.

Do you have any suggestion on how we can distinguish the situation when the user has connected to the accounting software but the `connections` object hasn't populated? This is the only situation when we have to call `openPolicyAccountingPage`.

aldo-expensify: > @aldo-expensify, is it accurate to say that the connections object is empty when the user hasn't connected to any accounting software?

I just tested and it appears to me that the `connections` key is completely missing. I guess what you could do is create a local state with `useState` so you are sure that you only call `openPolicyAccountingPage` once. Something like

```
if (!loadedPolicyConnections) {
   openPolicyAccountingPage();
   setLoadedPolicyConnections(true);
}
```

src/pages/workspace/withPolicyConnections.tsx
aldo-expensify: hmm maybe what he means is that we should just add a comment saying why are we loading `connections` here. We are loading them because if the policy is not the preferred (active) policy, then what we have loaded so far of the policy may be just a summary which doesn't contain the `connections` key.

hayata-suenaga: ahhh your explanation made it so clear Aldo! I'll edit the comment

src/pages/workspace/withPolicyConnections.tsx
tgolen: Aldo's plan would work, but it might result in more API requests than necessary.

An alternate thought I wad was that maybe in the policy summaries that get sent to NewDot we could include a new property like `connectCount: 2` so that we know how many connections to expect. That way, you could determine the difference between not having any connections configured, and not having the connections loaded in Onyx yet.

hayata-suenaga: Thank you for the thoughtful and creative ideas, both 😄 

For now, I'll go with @aldo-expensify's suggestion as it's easier to implement. As an optimization to mitigate unnecessary API calls, I'll create a GH issue to implement @tgolen's suggestion.

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: I've included a comment at the beginning, before the HOC definition, to explain the component's purpose.

src/pages/workspace/withPolicyConnections.tsx
tgolen: Thanks for addressing it. If there isn't any real deadline for this, why not just do it the correct way the first time?

hayata-suenaga: Tim, thank you for the question. My thinking is that your solution, although the most preferable one, requires additional PRs where we add codes to create, increment, and decrement the new counter field.

Although QBO project itself doesn't have an external commit, the Xero project that comes after QBO has one. That's why I wanted to prioritize the speed.

However, let me know if my thinking is not the best course of action we can take. I'm happy to adjust our plans if needed 😄

tgolen: Yeah, that all makes sense. I think it's important to do it right the first, or else you're also having to do the work to remove the initial solution.

Even with Xerocon in August, I don't think it would take more than a day or two in order to add the property to the policy summaries. Do you think it would be much more than that? I was just thinking that when the policy list is fetched, it would also query the number of connections, so we don't have to worry about incrementing or decrementing a counter.

hayata-suenaga: I'll start working on https://github.com/Expensify/Expensify/issues/383660 to add the counter field to the `policy` object

hayata-suenaga: @mollfpr, let's get this merged

aldo-expensify: > @aldo-expensify, if I remember correctly, the `ConnectToQuickbooksOnline` Web-E command is called after the QBO authentication is done and it handles pushing the Onyx update for the Policy object to the clients?

`ConnectToQuickbooksOnline` is called to start the authentication flow, so before... `ConnectToQuickbooksOnline` itself doesn't send any onyx updates. The job in the IS server does and when we call `SavePolicy` further down the flow.

I don't think in any case we send the full policy object. I think the updates only contain the updated sections handled by SavePolicy.

aldo-expensify: I'm a bit concerned about the `connectCount` making things complicated unnecessarily. We would have to make sure we are always updating the `connectCount` when there is a `connections` change. 

What about instead of using a local state as I suggested, we use a context with a state, so if we open more pages that require the policy's connection settings, we can see that a different page already loaded it.

hayata-suenaga: > What about instead of using a local state as I suggested, we use a context with a state, so if we open more pages that require the policy's connection settings, we can see that a different page already loaded it.

This is a sound idea. It addresses @tgolen's concerns about unnecessary API calls to a certain extent.

@tgolen, do you have any other concerns besides unnecessary API calls? I'm interested in understanding the potential issues your solution aims to resolve. 😄

tgolen: My main concerns are unnecessary API calls, yes. They do not have a zero cost and I think it is a very naive approach there is a better solution. Whatever you do here is going to put a pattern in place that engineers are going to follow from here on out and I think it's a bad pattern.

It sounds like @aldo-expensify is concerned that a piece of policy data would be complex to keep track of, but this is just like any other piece of policy that exists on the policy and there is nothing more complex that needs to be done here that we aren't doing with all of the other policy data.

aldo-expensify: I think we should seek more opinions in engineering-chat, but meanwhile... I don't think the problem of "unnecessary API calls" is big enough to introduce this new piece of data. Why? The only cases that would happen are:
- User navigates to the main Accounting page: if `policy.connections` is missing, it will trigger a fetch, but just one and will get nothing (there was no `connections`).
- User navigates to deeper Accounting page when there is no `policy.connections`: this can only happen if the user is entering directly using the url, which I think is a rare case.

In cases where there is actually a connection, after we load it, we won't do further calls to the API

aldo-expensify: Slack discussion: https://expensify.slack.com/archives/C03TQ48KC/p1712185841746469

tgolen (dismissed):

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: We came to the consensus that we should go with the approach to fetch the `connections` data once.

I'll modify the PR to use a Context as suggested by @aldo-expensify 👍

mollfpr:
## Reviewer Checklist

    - [ ] Android: Native
    - [ ] Android: mWeb Chrome
    - [ ] iOS: Native
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari
    - [ ] MacOS: Desktop
    - [ ] I verified that the left part of a conditional rendering a React component is a boolean and NOT a string, e.g. `myBool && <MyComponent />`.

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

src/pages/workspace/withPolicyConnections.tsx
```diff
@@ -0,0 +1,73 @@
+import {useEffect, useState} from 'react';
```
mollfpr, line 1: Missing react import.

```suggestion
import React, {useEffect, useState} from 'react';
```

src/pages/workspace/withPolicyConnections.tsx
```diff
@@ -0,0 +1,73 @@
...
+                            policyMembers={policyMembers}
+                            policyDraft={policyDraft}
+                            policyMembersDraft={policyMembersDraft}
+                            route={route}
+                        />
+                    </FullPageOfflineBlockingView>
+                );
+            }
+
+            return null;
```
mollfpr, line 56: I don't get why we return nothing if the `connections` are false, I try to use it on the workspace profile page but it shows a blank page. Also, I'm curious about what case the HOC will be used.

mollfpr (commented):
Mostly the code looks good to me!

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: I added some comments in [the new commit](https://github.com/Expensify/App/pull/39132/commits/23702781c6a90ee8786318d43323bb90c9b2326a). It explains the usage of the HOC and the context that I also added in that commit.

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: added https://github.com/Expensify/App/pull/39132/commits/db3252e0aca375bd9f38fe40e2af66456135395a

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: > I don't get why we return nothing if the connections are false

@mollfpr, @aldo-expensify, should we return `LoadingPage` instead of `null`?

hayata-suenaga:
Reassure test was failing. Merged the main.

src/pages/workspace/withPolicyConnections.tsx
mollfpr: We might be able to leverage this later. The only case I think we should return `null` is where we want to navigate the page if there are no `connections` to avoid rendering a component while animating the transition.

mollfpr:
I feel like the new implementation with `React.context` is kinda hard to follow, because of the extra step to wrap the component with the provider component.

hayata-suenaga:
> I feel like the new implementation with React.context is kinda hard to follow, because of the extra step to wrap the component with the provider component.

I agree this code is not ideal, but this is something we need to do to ensure that the API call is made at most once per user session. We need to have a context state to keep track if we have made the API call or not.

The question, here, is rather where we should place the context provider. @mollfpr, do you have any suggestions on this?

aldo-expensify:
I feel like I would delay this PR until we have somewhere we can start using the provider, otherwise it is hard for me to tell if we are doing something that is going to be useless or not.

mollfpr:
> The question, here, is rather where we should place the context provider. @mollfpr, do you have any suggestions on this?

I don't have it in mind. It might be not a big deal when we knew where to use the HOC later.

hayata-suenaga:
> I feel like I would delay this PR until we have somewhere we can start using the provider, otherwise it is hard for me to tell if we are doing something that is going to be useless or not.

Good point. Let's wait for any of the following PRs to be merged:

- https://github.com/Expensify/App/pull/39652
- https://github.com/Expensify/App/pull/38400
- https://github.com/Expensify/App/pull/39191

hayata-suenaga:
@aldo-expensify, I'm struggling with placing the context provider that I defined in this PR.

To recall your memory, we created the context so that we can track if the `connections` data has been fetched across all accounting integration-related screens.

However, these accounting screens are scattered across different navigators, making it necessary to place the context up in the component tree.

Furthermore, the context relies on `withPolicy`, which requires to be inside the navigation context. So we have to place the context in a weird spot of the component tree (we cannot place it inside `App.tsx` as it doesn't have the navigation context. somewhere with navigation context but above the place we place navigators).

I suggest we discard the context provider (I gonna remove it from this PR), and just use HOC. In this case, whenever we visit the accounting page, the `connections` data is attempted to be fetched.

I think it's still acceptable. What do you think? @aldo-expensify

aldo-expensify:
> However, these accounting screens are scattered across different navigators, making it necessary to place the context up in the component tree.

hehe, I thought we could have a problem like this.

> I suggest we discard the context provider (I gonna remove it from this PR), and just use HOC. In this case, whenever we visit the accounting page, the connections data is attempted to be fetched.

> I think it's still acceptable. What do you think? @aldo-expensify

I think it is fine

hayata-suenaga:
I'm facing another issue.

In the HOC, we are supposed to display a loading indicator or empty screen while the `connections` data is being fetched.

I'm having trouble coming up with a way to determine if the connections data is being fetched or not. The `openPolicyAccountingPage` function, which is responsible for calling the API command to fetch the `connections` data, doesn't return anything. The function updates the Onyx entry when the API response is returned.

Without an access to the API request status, we don't know if we should render the loading state or not.

@aldo-expensify, do you have any possible solution for this issue? The only solution I can come up with is to add a new Onyx key to hold the status of the `OpenPolicyAccountingPage`, which is something we should avoid.

tgolen:
@hayata-suenaga we use `isLoading` flags in Onyx for things like that. The optimistic data sets it to `true` and the success data sets it to `false`. Is that what you need?

hayata-suenaga:
>@hayata-suenaga we use isLoading flags in Onyx for things like that. The optimistic data sets it to true and the success data sets it to false. Is that what you need?

Yes, exactly! I was thinking this is the pattern we want to avoid. However, if this is the accepted pattern, I'll use that here.

hayata-suenaga:
@aldo-expensify, @tgolen

Inside the `policy` object, I decided to store the boolean value called `hasConnectionsDataBeenFetched` to indicate if the `connections` data has been attempted to be fetched.

In this way, I can store the fetched-or-not state for each policy. Also, this allows sharing the fetched state across different accounting screens.

Additionally, I added another boolean field called `isLoadingConnections` to store the loading state.

One unresolved issue with these two approaches is that these boolean values need to be cleared when the user reopens the app. [I'm asking in Slack how we can do that](https://expensify.slack.com/archives/C01GTK53T8Q/p1713213975439439).

src/libs/actions/PolicyConnections.ts
```diff
@@ -0,0 +1,55 @@
...
+function openPolicyAccountingPage(policyID: string) {
+    const policyKey = `${ONYXKEYS.COLLECTION.POLICY}${policyID}` as const;
+
+    const optimisticData: OnyxUpdate[] = [
+        {
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: policyKey,
+            value: {
+                isLoadingConnections: true,
+                hasConnectionsDataBeenFetched: false,
```
aldo-expensify, line 17: `isLoadingConnections` and `hasConnectionsDataBeenFetched` seem to have exactly the opposite values, I think you can just have one of them.

src/pages/workspace/withPolicyConnections.tsx
aldo-expensify: I think we should should a spinner if `policy.isLoadingConnections` is true, and if `policy.isLoadingConnections` is false, hmmm 404? 🤷

aldo-expensify (changes requested):

src/libs/actions/PolicyConnections.ts
hayata-suenaga: you're right. we only need `hasConnectionsDataBeenFetched`

hayata-suenaga:
I'm wondering what is the best way to clear the `hasConnectionsDataBeenFetched` field when the app reopens. We want to ensure that we fetch the fresh `connections` data once per user session. If we don't reset the field value to `false` at app start, the fresh data won't be fetched.

I [asked](https://expensify.slack.com/archives/C01GTK53T8Q/p1713241249661959?thread_ts=1713213975.439439&cid=C01GTK53T8Q) a more general version of this question in Slack, but I'm moving the convo from there to this PR as the discussion is becoming more specific.

tgolen:
Will `initWithStoredValues: false` not work?

hayata-suenaga:
>Will initWithStoredValues: false not work?

@tgolen, good question! Because I'm storing the fetched state inside each policy object. If I use `initWithStoredValues`, the entire `policy` object will be ignored.

One solution might be to create a new collection for just fetched state of the `connections` field for each policy, but I think we're trying to avoid creating more Onxy keys unless necessary?

tgolen:
We can create new keys if we need to. The problem was that so many new keys were being created that didn't need to be made, which resulted in more complex code (eg. different collections for policy tags, categories, members, etc. when they should just be stored as the policy). 

We should also be trying to mirror Onyx after how the backend stores data (for ease in keeping the different layers in sync and for understanding them). 

Since `hasConnectionsDataBeenFetched` is not stored in the database at all, I think that's a pretty clear sign that it could be a separate key. Much like the report meta data collection (be sure to read the accompanying code comment, which perfectly describes the same situation you're dealing with).

https://github.com/Expensify/App/blob/6cdfa55d9c487a354385d676f8db829c8a554d98/src/ONYXKEYS.ts#L324-L327

hayata-suenaga:
> We should also be trying to mirror Onyx after how the backend stores data (for ease in keeping the different layers in sync and for understanding them).

I completely agree with you, @tgolen! I decided to add a new Onyx collection key called `policyHasConnectionsDataBeenFetched`. And I decided to use `useOnyx` to get the value 😄

hayata-suenaga:
@mollfpr, this PR is ready for another review

src/ONYXKEYS.ts
```diff
@@ -318,6 +318,8 @@ const ONYXKEYS = {
         POLICY_RECENTLY_USED_CATEGORIES: 'policyRecentlyUsedCategories_',
         POLICY_TAGS: 'policyTags_',
         POLICY_RECENTLY_USED_TAGS: 'nvp_recentlyUsedTags_',
+        /** Whether the connections data for the policy was attempted to be fetched in the current user session */
+        POLICY_HAS_CONNECTIONS_DATA_BEEN_FETCHED: 'policyHasConnectionsDataBeenFetched',
```
mollfpr, line 322:
```suggestion
        POLICY_HAS_CONNECTIONS_DATA_BEEN_FETCHED: 'policyHasConnectionsDataBeenFetched_',
```

I think the common key has the underscore at the end.

src/pages/workspace/withPolicyConnections.tsx
```diff
@@ -0,0 +1,69 @@
...
+            // When the accounting feature is not enabled, or if the connections data already exists,
+            // there is no need to fetch the connections data.
+            if (!policy || !policy.areConnectionsEnabled || !!hasConnectionsDataBeenFetched || !!policy.connections) {
+                return;
+            }
+
+            openPolicyAccountingPage(policy.id);
+        }, [hasConnectionsDataBeenFetched, policy]);
+
+        if (status === 'loading' || !hasConnectionsDataBeenFetched) {
```
mollfpr, line 41: This will always show the loader when the policy hasn't enabled the accounting feature. Should we let the `OpenPolicyAccountingPage` API be called even though the policy hasn't enabled the feature?

mollfpr (commented):
Mostly looks good to me!

src/pages/workspace/withPolicyConnections.tsx
trjExpensify: Just to confirm, how is this accessible in the UI if accounting isn't enabled?

hayata-suenaga: @mollfpr, this page won't be acceesible for users who hasn't enabled the accounting feature.

Also, even if the user somehow be able to access this page, it won't cause the infinite loading state. `hasConnectionsDataBeenFetched` will be true if the fetch is attempted once (even if there is no data returned from the API call)

hayata-suenaga: also, just to be clear, the `connections` field might still be empty even if the user has enabled the accounting feature if the user hasn't connected to any accounting software

mollfpr (approved):
LGTM 👍

src/ONYXKEYS.ts
```diff
@@ -318,6 +318,8 @@ const ONYXKEYS = {
         POLICY_RECENTLY_USED_CATEGORIES: 'policyRecentlyUsedCategories_',
         POLICY_TAGS: 'policyTags_',
         POLICY_RECENTLY_USED_TAGS: 'nvp_recentlyUsedTags_',
+        /** Whether the connections data for the policy was attempted to be fetched in the current user session */
```
tgolen, line 321: Would you please include some information in this comment to explain why it shouldn't be stored on just the normal `POLICY` collection? Similar to the comment for `REPORT_METADATA`. Also, I see comment formats all over the place in this file. Technically, these comments should follow the `//` style.

src/libs/actions/PolicyConnections.ts
```diff
@@ -0,0 +1,38 @@
+import Onyx from 'react-native-onyx';
+import type {OnyxUpdate} from 'react-native-onyx';
+import * as API from '@libs/API';
+import type {OpenPolicyAccountingPageParams} from '@libs/API/parameters';
+import {READ_COMMANDS} from '@libs/API/types';
+import ONYXKEYS from '@src/ONYXKEYS';
+
+function openPolicyAccountingPage(policyID: string) {
+    const hasConnectionsDataBeenFetchedKey = `${ONYXKEYS.COLLECTION.POLICY_HAS_CONNECTIONS_DATA_BEEN_FETCHED}${policyID}` as const;
```
tgolen, line 9: Why is `as const` necessary here?

src/pages/workspace/withPolicyConnections.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Use this HOC when you need the policy object with its connections field populated.
+ *
+ * Only the active policy gets the complete policy data upon app start that includes the connections data.
+ * For other policies, the connections data needs to be fetched when it's needed.
+ */
+function withPolicyConnections(WrappedComponent: ComponentType<WithPolicyConnectionsProps>) {
+    function WithPolicyConnections({policy, policyDraft, route}: WithPolicyConnectionsProps) {
+        const {isOffline} = useNetwork();
+        const [hasConnectionsDataBeenFetched, {status}] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY_HAS_CONNECTIONS_DATA_BEEN_FETCHED}${policy?.id ?? ''}` as const, {
```
tgolen, line 27:
```suggestion
        const [hasConnectionsDataBeenFetched, {status}] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY_HAS_CONNECTIONS_DATA_BEEN_FETCHED}${policy?.id ?? '0'}` as const, {
```
I think you want this, otherwise, you could be getting the entire collection.

tgolen (changes requested):

src/pages/workspace/withPolicyConnections.tsx
hayata-suenaga: fixed: https://github.com/Expensify/App/pull/39132/commits/4804fd19f34f2f9c3a89c82b4d6e296d4a1b05ac

src/ONYXKEYS.ts
hayata-suenaga: changed: https://github.com/Expensify/App/pull/39132/commits/e19060b4968066e581ec5e825d59bf4c7d9b707a

src/libs/actions/PolicyConnections.ts
hayata-suenaga: fix: https://github.com/Expensify/App/pull/39132/commits/a084e5295dea50b4d14fb560c2ad5e58e83cb1f0

tgolen: It looks like you removed it from a different spot. Is it still necessary here?

aldo-expensify: The `const` is necessary here, otherwise the `hasConnectionsDataBeenFetchedKey` gets `string` as a type and then `OnyxUpdate` is not happy with such key:

image

tgolen (approved):

src/libs/actions/PolicyConnections.ts
hayata-suenaga: haha I definitely moved the `const` from a different place but it didn't cause the TS error 😆 for this particular one

aldo-expensify (approved):

src/libs/actions/PolicyConnections.ts
```diff
@@ -0,0 +1,38 @@
...
+            value: false,
+        },
+    ];
+    const finallyData: OnyxUpdate[] = [
+        {
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: hasConnectionsDataBeenFetchedKey,
+            value: true,
+        },
+    ];
```
aldo-expensify, line 24: @hayata-suenaga After playing with this, I think you missed the `failureData` setting `hasConnectionsDataBeenFetched` to `false` or something like that?

If you do:

1. Create a workspace
2. Enable `Accounting`, but don't got to the accounting settings yet
3. Disable your internet connection
4. Open the `Accounting` settings

I would have expected that an Offline blocker would appear since we haven't loaded the policy connections, but it doesn't because `withPolicyConnections` is set `true` even if `OPEN_POLICY_ACCOUNTING_PAGE` fails

aldo-expensify: image

image

hayata-suenaga: [Here is the PR to fix the issue](https://github.com/Expensify/App/pull/40375). Thank you for letting me know of the issue 🙇

src/pages/workspace/withPolicyConnections.tsx
s77rt: This seems broken. I'm getting infinite loader when I try to visit the page again

https://github.com/Expensify/App/assets/16493223/bf5a6e67-2d38-4d8e-a0b0-a185364602e9

aldo-expensify: @s77rt similar problem with going offline and then offline: https://github.com/Expensify/App/pull/39132#discussion_r1569411481

@hayata-suenaga is working on it here https://github.com/Expensify/App/pull/40375

Issue #4114, opened 2021-07-16
[HOLD facebook/hermes release] [$1000] Make emojis larger in other contexts than just single character messages


jboniface:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

This issue covers several related scenarios:

### Action Performed:
Add an emoji with other plain text characters in a message in the composer.

### Expected Result:
The emoji should be larger (19pt) compared to the default size for text (15pt) in the composer. The emoji and text should appear vertically centered with respect to eachother, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji (like the right-hand-side of [this example](https://github.com/Expensify/App/issues/4114#issuecomment-1078143131).

### Actual Result:
The emoji is the same size as the rest of the text.

### Action Performed:
Send a message with emoji and text.

### Expected Result:
The emoji should be larger (19pt) compared to the default size for text (15pt) in the chat history. The emoji and text should appear vertically centered with respect to eachother, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji (like the right-hand-side of [this example](https://github.com/Expensify/App/issues/4114#issuecomment-1078143131)

### Actual Result:
The emoji is the same size as the rest of the text.

### Action Performed
Draft a message (of any length) with only emoji.

### Expected Result
The emoji should appear very large (27pt) in the composer.

### Actual Result
The emoji appear very small (15pt).

### Action Performed
Send a message (of any length) with only emoji.

### Expected Result
The emoji should appear very large (27pt) in the report history.

### Actual Result
The emoji only appear large in the report history if there is only one emoji in the message.

### Action Performed
Go to `Settings` -> `Profile` and add emoji(s) to your last name 😎

### Expected Result
The emoji should appear larger (19pt) than the rest of the text (15pt) in the text input. The emoji and text should appear vertically centered with respect to eachother, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji (like the right-hand-side of [this example](https://github.com/Expensify/App/issues/4114#issuecomment-1078143131)

### Actual Result
The emoji appear the same size as the text.

### Action Performed
1. Go to `Settings` -> `Profile` and add emoji(s) to your last name 😎.
1. Save your profile.
1. Open the `Settings` RHP and look at your name:

[...]

PR #15611, opened 2023-03-02
Fix/make-emojis-larger-in-other-message-contexts

josemak25:
HOLD on https://github.com/facebook/hermes/issues/932

### Details

This solution covers several related scenarios:
#### Action Performed:
Add an emoji with other plain text characters in a message in the composer.

#### Expected Result:
The emoji should be larger (19pt) compared to the default size for text (15pt) in the composer. The emoji and text should appear vertically centered with respect to each other, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji (like the right-hand-side of https://github.com/Expensify/App/issues/4114#issuecomment-1078143131.

> ***Solution does not cover this case as handling emojis inside the input is a bit tricky***

#### Action Performed:
Send a message with emojis and text.

#### Expected Result:
The emoji should be larger (19pt) compared to the default size for text (15pt) in the chat history. The emoji and text should appear vertically centered with respect to each other, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji (like the right-hand-side of https://github.com/Expensify/App/issues/4114#issuecomment-1078143131

> ***Solution covers this case***

#### Action Performed
Draft a message (of any length) with only an emoji.

#### Expected Result:
The emoji should appear very large (27pt) in the composer.

> ***Solution does not cover this case as handling emojis inside the input is a bit tricky***

#### Action Performed
Send a message (of any length) with only emoji.

#### Expected Result:
The emoji should appear very large (27pt) in the report history.

> ***Solution covers this case***

#### Action Performed
Go to `Settings` -> Profile and add emoji(s) to your last name 😎

#### Expected Result:
The emoji should appear larger (19pt) than the rest of the text (15pt) in the text input. The emoji and text should appear vertically centered with respect to each other, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji (like the right-hand-side of https://github.com/Expensify/App/issues/4114#issuecomment-1078143131

> ***Solution does not cover this case as handling emojis inside the input is a bit tricky***

#### Action Performed
1. Go to `Settings` -> `Profile` and add emoji(s) to your last name 😎.
2. Save your profile.
3. Open the `Settings` RHP and look at your name:

#### Expected Result:
[...]

PR #40692, opened 2024-04-22
Emojis larger in other contexts than just single character messages

kbieganowski:
### Fixed Issues
$ https://github.com/Expensify/App/issues/4114

PROPOSAL: n/a

### Tests
**Composer:**
1. Draft a message (of any length) with only emojis
2. Add other plain text characters in a message in the composer

- [x] Verify that emojis appear very large (27px) in the composer (for emojis only input)
- [x] Verify that emojis are larger (19pt) compared to the default size for text (15px) in the composer (emojis + text input)

**Report history:**
1. Send a message with emojis only
2. Send a message with emojis and text

- [x] Verify that emojis appear very large (27px) in the report history (emojis only comment)
- [x] Verify that emojis are larger (19px) compared to the default size for text (15pt) in the chat history and vertically centered with respect to eachother, such that the baseline of the emoji is below the plain text and the middle of the text vertically aligns with the middle of the emoji [like shown here](https://github.com/Expensify/App/issues/4114#issuecomment-1078143131)

**Settings:**
1. Go to `Settings` -> `Profile` and add emoji to your last name
- [x] Verify that emoji appear larger (19pt) than the rest of the text (15pt) in the text input
2. Save the changes
3. Open LHP
- [x] Verify that emoji in user's last name appear larger (25px) compared to the rest of the text (22px)

### Offline tests
n/a

### QA Steps
Same as in `Tests` section

**Note:** If more details regarding tests flow are needed, they can be found in description of related issue

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
- x ] I followed the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [x] If the PR modifies the form input styles:
    - [x] I added `Design` label so the design team can review the changes.

### Screenshots/Videos
<details>
native_android_chat
native_android_settings

</details>

<details>
web_android_chat
web_android_settings
web_android_settings_input

</details>

<details>
native_ios_chat
native_ios_settings

</details>

<details>
web_ios_chat
web_ios_settings
web_ios_settings_input

</details>

<details>

https://github.com/Expensify/App/assets/26638539/2ab42ac2-d43a-4623-af16-ab17890e1c1c

chrome_settings_LHP

chrome_settings_input

</details>

<details>

https://github.com/Expensify/App/assets/26638539/3443a703-41eb-455b-bbf6-1ae621c38a9b

desktop_settings

</details>


src/libs/EmojiUtils.ts
```diff
@@ -569,6 +569,120 @@ function getSpacersIndexes(allEmojis: EmojiPickerList): number[] {
...
 
+/**
+ * Split string into plain text and emojis array with attention to:
+ * Surrogate pairs (combined emojis including flags)
+ * Modifiers (different skin tones and emoji variations)
+ * @param text
+ */
+
+// Surrogate pairs (combined emojis)
+const highSurrogateStart = 0xd800;
```
roryabraham, line 580: Let's use `SCREAMING_SNAKE_CASE` for all these constants

src/libs/EmojiUtils.ts
```diff
@@ -569,6 +569,120 @@ function getSpacersIndexes(allEmojis: EmojiPickerList): number[] {
...
+
+// Fitzpatrick scale modifiers (skin tone)
+const fitzpatrickScaleStart = 0x1f3fb; // Type 1 (represents light skin tone)
+const fitzpatrickScaleEnd = 0x1f3ff; // Type 6 (represents dark skin tone)
+
+// Variation selectors (specific variations in the presentation of other characters)
+const variationModifierStart = 0xfe00; // Request text presentation of emoji
+const variationModifierEnd = 0xfe0f; // Indicate that the character should be displayed as an emoji
+
+const codePointFromSurrogatePair = (pair: string) => {
```
roryabraham, line 597: let's use the `function` keyword instead of arrow functions, just because it's the more common pattern in this codebase

src/libs/EmojiUtils.ts
kbieganowski: sure! updated

MrRefactor (approved):

src/libs/EmojiUtils.ts
```diff
@@ -569,6 +569,137 @@ function getSpacersIndexes(allEmojis: EmojiPickerList): number[] {
...
+        return 4; // Flags (combination of 2 regional indicators)
+    }
+
+    if (isFitzpatrickModifier(nextPair)) {
+        return 4; // Skin tones
+    }
+    return 2; // Variations and non-BMP characters
+}
+
+function splitTextWithEmojis(text: string): string[] {
```
fabioh8010, line 649: I think we could _vastly_ simplify this code if we use a combination of regex + [this logic](https://github.com/Expensify/App/pull/15611#pullrequestreview-1322663118) @roryabraham designed in the past.

First, we need to update `CONST.REGEX.EMOJIS` to include the new `d` flag [which I've implemented in the past for this issue](https://github.com/facebook/hermes/pull/968) and is available now on Hermes engine:

```diff
diff --git a/src/CONST.ts b/src/CONST.ts
index f0139d82e6..ebe0ab7549 100755
--- a/src/CONST.ts
+++ b/src/CONST.ts
@@ -1760,8 +1760,8 @@ const CONST = {
 
         // eslint-disable-next-line max-len, no-misleading-character-class
         EMOJI: /[\p{Extended_Pictographic}\u200d\u{1f1e6}-\u{1f1ff}\u{1f3fb}-\u{1f3ff}\u{e0020}-\u{e007f}\u20E3\uFE0F]|[#*0-9]\uFE0F?\u20E3/gu,
-        // eslint-disable-next-line max-len, no-misleading-character-class
-        EMOJIS: /[\p{Extended_Pictographic}](\u200D[\p{Extended_Pictographic}]|[\u{1F3FB}-\u{1F3FF}]|[\u{E0020}-\u{E007F}]|\uFE0F|\u20E3)*|[\u{1F1E6}-\u{1F1FF}]{2}|[#*0-9]\uFE0F?\u20E3/gu,
+        // eslint-disable-next-line max-len, no-misleading-character-class, no-empty-character-class
+        EMOJIS: /[\p{Extended_Pictographic}](\u200D[\p{Extended_Pictographic}]|[\u{1F3FB}-\u{1F3FF}]|[\u{E0020}-\u{E007F}]|\uFE0F|\u20E3)*|[\u{1F1E6}-\u{1F1FF}]{2}|[#*0-9]\uFE0F?\u20E3/dgu,
         // eslint-disable-next-line max-len, no-misleading-character-class
         EMOJI_SKIN_TONES: /[\u{1f3fb}-\u{1f3ff}]/gu,
 

```

---

Then, let's implement the `splitTextWithEmojis` function which will use the regex and indices feature and is based on Rory's initial idea:

```tsx
type TextWithEmoji = {
    text: string;
    isEmoji: boolean;
};

function splitTextWithEmojis(text: string): TextWithEmoji[] {
    if (!text) {
        return [];
    }

    const splitText: TextWithEmoji[] = [];
    let regexResult: RegExpExecArray | null;
    let lastMatchIndexEnd = 0;
    do {
        regexResult = CONST.REGEX.EMOJIS.exec(text);

        if (regexResult?.indices) {
            const matchIndexStart = regexResult.indices[0][0];
            const matchIndexEnd = regexResult.indices[0][1];

            if (matchIndexStart > lastMatchIndexEnd) {
                splitText.push({
                    text: text.slice(lastMatchIndexEnd, matchIndexStart),
                    isEmoji: false,
                });
            }

            splitText.push({
                text: text.slice(matchIndexStart, matchIndexEnd),
                isEmoji: true,
            });

            lastMatchIndexEnd = matchIndexEnd;
        }
    } while (regexResult !== null);

    if (lastMatchIndexEnd < text.length) {
        splitText.push({
            text: text.slice(lastMatchIndexEnd, text.length),
            isEmoji: false,
        });
    }

    return splitText;
}
```

---

You'll notice some TS errors in the function regarding `indices` not being existent. We'll need to augment those Regex types to include this new property, so let's create the `src/types/modules/lib.es2018.regexp.d.ts` file:

```tsx
/* eslint-disable @typescript-eslint/naming-convention */
/* eslint-disable @typescript-eslint/consistent-type-definitions */

declare global {
    interface Indices extends Array<[number, number]> {
        groups?: Record<string, [number, number]>;

        /**
         * The first match. This will always be present because `null` will be returned if there are no matches.
         */
        0: [number, number];
    }

    interface RegExpMatchArray {
        indices?: Indices;
    }

    interface RegExpExecArray {
        indices?: Indices;
    }
}

// We used the export {} line to mark this file as an external module
export {};

```

---

Let's test this function now! If I pass this string `Hello world 🙂🙂🙂 ! 😶‍🌫️😶‍🌫️😶‍🌫️ test 😶‍🌫️ 😶‍🌫️ test2 👍👍🏿` which has some normal and surrogate emojis, the output will be:

```json
[
    {"isEmoji":false,"text":"Hello world "},
    {"isEmoji":true,"text":"🙂"},
    {"isEmoji":true,"text":"🙂"},
    {"isEmoji":true,"text":"🙂"},
    {"isEmoji":false,"text":" ! "},
    {"isEmoji":true,"text":"😶‍🌫️"},
    {"isEmoji":true,"text":"😶‍🌫️"},
    {"isEmoji":true,"text":"😶‍🌫️"},
    {"isEmoji":false,"text":" test "},
    {"isEmoji":true,"text":"😶‍🌫️"},
    {"isEmoji":false,"text":" "},
    {"isEmoji":true,"text":"😶‍🌫️"},
    {"isEmoji":false,"text":" test2 "},
    {"isEmoji":true,"text":"👍"},
    {"isEmoji":true,"text":"👍🏿"}
]
```

You can adjust the function to just return an array withe the separated tokens if you want, could you give it a try and see if it suits your needs in the PR?

fabioh8010: Update: we are still testing the regex solution

fabioh8010: @kbieganowski I've updated my previous solution:

1. Fixed `splitTextWithEmojis` function to always create a separate regex instance on each call, thus preventing the issue we were discussing internally.
2. Simplified solution to the TS errors regarding `indices`.

fabioh8010: @kbieganowski After internal discussions, we noticed that having the `g` (global) flag on `CONST.REGEX.EMOJIS` can result in hard-to-detect bugs throughout the App due to the stateful nature of the regex when it has the `g` flag. For example, repeatedly calls to `test()` with same input will sometimes return different results because the regex is being used globally.

So, I would suggest to remove the `g` flag from `CONST.REGEX.EMOJIS`:

```tsx
        // eslint-disable-next-line max-len, no-misleading-character-class, no-empty-character-class
        EMOJIS: /[\p{Extended_Pictographic}](\u200D[\p{Extended_Pictographic}]|[\u{1F3FB}-\u{1F3FF}]|[\u{E0020}-\u{E007F}]|\uFE0F|\u20E3)*|[\u{1F1E6}-\u{1F1FF}]{2}|[#*0-9]\uFE0F?\u20E3/du,
```

And then adding it when cloning the regex inside `splitTextWithEmojis`:

```tsx
    // We clone the regex and add the `g` (global) flag in order to `exec()` be stateful and work properly.
    // Adding the flag directly to CONST.ts can result in bugs throughout the App due to the stateful nature
    // of the regex when it has the `g` flag.
    const emojisRegex = new RegExp(CONST.REGEX.EMOJIS, CONST.REGEX.EMOJIS.flags.concat('g'));
```

This change will require testing in all areas of the app that uses `CONST.REGEX.EMOJIS` since we are removed the `g` flag now.

I would advice doing the same for `CONST.REGEX.EMOJI` and `CONST.REGEX.EMOJI_SKIN_TONES`, that could be done in a follow-up PR to avoid regressions on this one.

kbieganowski:
Hey @fabioh8010, I've validated other usages of the regex and everything seems to work as expected. Ready for a re-review

src/components/Composer/index.native.tsx
```diff
@@ -65,7 +67,7 @@ function Composer(
     }, [shouldClear, onClear]);
 
     const maxHeightStyle = useMemo(() => StyleUtils.getComposerMaxHeightStyle(maxLines, isComposerFullSize), [StyleUtils, isComposerFullSize, maxLines]);
-    const composerStyle = useMemo(() => StyleSheet.flatten(style), [style]);
+    const composerStyle = useMemo(() => StyleSheet.flatten([style, textContainsOnlyEmojis ? {lineHeight: 32} : null]), [style, textContainsOnlyEmojis]);
```
fabioh8010, line 70:
```suggestion
    const composerStyle = useMemo(() => StyleSheet.flatten([style, textContainsOnlyEmojis ? {lineHeight: variables.lineHeightXXXLarge} : null]), [style, textContainsOnlyEmojis]);
```

Maybe did you want to use `lineHeightXXXLarge` (or perhaps a different variable)?

src/components/TextInput/BaseTextInput/index.tsx
```diff
@@ -380,6 +380,7 @@ function BaseTextInput(
                                     // Add disabled color theme when field is not editable.
                                     inputProps.disabled && styles.textInputDisabled,
                                     styles.pointerEventsAuto,
+                                    isMarkdownEnabled ? {lineHeight: 18} : null,
```
fabioh8010, line 383:
```suggestion
                                    isMarkdownEnabled ? {lineHeight: variables.lineHeightLarge} : null,
```

makes sense (or perhaps a different variable)?

src/hooks/useMarkdownStyle.ts
```diff
@@ -21,7 +19,7 @@ function useMarkdownStyle(message: string | null = null): MarkdownStyle {
                 fontSize: variables.fontSizeLarge,
             },
             emoji: {
-                fontSize: emojiFontSize,
+                fontSize: inputContainsOnlyEmojis ? 27 : 19,
```
fabioh8010, line 22: Same, shouldn't we use already defined variables for those values, or create new ones if they are new?

src/pages/home/report/comment/TextWithEmojiFragment.tsx
```diff
@@ -0,0 +1,71 @@
...
+type ComponentProps = {
+    message: string;
+    passedStyles: StyleProp<TextStyle>;
+    styleAsDeleted?: boolean;
+    styleAsMuted?: boolean;
+    isSmallScreenWidth?: boolean;
+    isEdited?: boolean;
+    emojisOnly?: boolean;
+};
+function TextWithEmojiFragment({message, passedStyles, styleAsDeleted, styleAsMuted, isSmallScreenWidth, isEdited, emojisOnly}: ComponentProps) {
```
fabioh8010, line 21:
```suggestion

function TextWithEmojiFragment({message, passedStyles, styleAsDeleted, styleAsMuted, isSmallScreenWidth, isEdited, emojisOnly}: ComponentProps) {
```

src/pages/home/report/comment/TextWithEmojiFragment.tsx
```diff
@@ -0,0 +1,71 @@
...
+import Text from '@components/Text';
+import useLocalize from '@hooks/useLocalize';
+import useTheme from '@hooks/useTheme';
+import useThemeStyles from '@hooks/useThemeStyles';
+import * as DeviceCapabilities from '@libs/DeviceCapabilities';
+import {splitTextWithEmojis} from '@libs/EmojiUtils';
+import variables from '@styles/variables';
+import CONST from '@src/CONST';
+
+type ComponentProps = {
```
fabioh8010, line 12:
```suggestion
type TextWithEmojiFragmentProps = {
```

Also, please add descriptions for each prop

src/pages/home/report/comment/TextWithEmojiFragment.tsx
```diff
@@ -0,0 +1,71 @@
...
+    message: string;
+    passedStyles: StyleProp<TextStyle>;
```
fabioh8010, line 14:
```suggestion
    passedStyles?: StyleProp<TextStyle>;
```

Style props are optional by convention.

fabioh8010 (commented):
Few comments, also please make sure all places where `CONST.REGEX.EMOJIS` is used (`src/components/InlineCodeBlock/WrappedText.tsx`, `src/libs/EmojiUtils.ts`, `src/libs/ValidationUtils.ts`) are properly handled after removing the `g` flag.

src/components/Composer/index.native.tsx
kbieganowski: I went with new variables rather than using existing ones because this styles are strictly related to composer/input and I didn't want potential changes to the UI in future to break composer or inputs as well but no more magic numbers

src/pages/home/report/comment/TextWithEmojiFragment.tsx
```diff
@@ -0,0 +1,79 @@
...
+    styleAsDeleted?: boolean;
+    /** Should this message fragment be styled as muted? */
+    styleAsMuted?: boolean;
+    /** Is message displayed on narrow screen? */
+    isSmallScreenWidth?: boolean;
+    /** Should "(edited)" suffix be rendered? */
+    isEdited?: boolean;
+    /** Does message contain only emojis? */
+    emojisOnly?: boolean;
+};
```
fabioh8010, line 27:
```suggestion
type TextWithEmojiFragmentProps = {
    /** The message to be displayed */
    message: string;

    /** Additional styles to add after local styles. */
    passedStyles?: StyleProp<TextStyle>;

    /** Should this message fragment be styled as deleted? */
    styleAsDeleted?: boolean;

    /** Should this message fragment be styled as muted? */
    styleAsMuted?: boolean;

    /** Is message displayed on narrow screen? */
    isSmallScreenWidth?: boolean;

    /** Should "(edited)" suffix be rendered? */
    isEdited?: boolean;

    /** Does message contain only emojis? */
    emojisOnly?: boolean;
};
```

Add spaces between the props

fabioh8010 (approved):
Just one nitpick, amazing work! LGTM 🚀

fabioh8010 (approved):

shawnborton:
Can we do anything to vertically align the cursor? It feels odd that it sits so far down compared to the emoji:
CleanShot 2024-05-16 at 14 09 01

kbieganowski:
> Can we do anything to vertically align the cursor? It feels odd that it sits so far down compared to the emoji: CleanShot 2024-05-16 at 14 09 01 [ CleanShot 2024-05-16 at 14 09 01 ](https://private-user-images.githubusercontent.com/2319350/331335336-aff8d6f3-d8ce-43f3-a494-20e82c304f4d.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTU5MjcxNjEsIm5iZiI6MTcxNTkyNjg2MSwicGF0aCI6Ii8yMzE5MzUwLzMzMTMzNTMzNi1hZmY4ZDZmMy1kOGNlLTQzZjMtYTQ5NC0yMGU4MmMzMDRmNGQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDUxNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA1MTdUMDYyMTAxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NDczYTRlODZmYzE0OTUwM2Q0ZjQ1ODI1YzJjNGIwZDEzY2Q2MjY3Mjk1MjM5YTMwMDExMjA0NzhjYWM1MGRjZSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.lUAvS7hRVWZ5gnCubKOqCgV1zlTrDE124M9tGxcorrQ) [ ](https://private-user-images.githubusercontent.com/2319350/331335336-aff8d6f3-d8ce-43f3-a494-20e82c304f4d.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTU5MjcxNjEsIm5iZiI6MTcxNTkyNjg2MSwicGF0aCI6Ii8yMzE5MzUwLzMzMTMzNTMzNi1hZmY4ZDZmMy1kOGNlLTQzZjMtYTQ5NC0yMGU4MmMzMDRmNGQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDUxNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA1MTdUMDYyMTAxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NDczYTRlODZmYzE0OTUwM2Q0ZjQ1ODI1YzJjNGIwZDEzY2Q2MjY3Mjk1MjM5YTMwMDExMjA0NzhjYWM1MGRjZSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.lUAvS7hRVWZ5gnCubKOqCgV1zlTrDE124M9tGxcorrQ)

Hey @shawnborton! Thanks for the comment. If we want to address this issue I feel like this should be handled on RN live markdown side. Few reasons from the top of my head why I think that:
1. If we try to handle that on the app side, we would need to implement accepted solution in every place that is using live markdown input or create new, more generic component and reuse it where needed but having in mind different designs, animations and so on (for example composer and name inputs on settings page) this would in my opinion require a lot of overhead that we would need to pass to this new component to preserve styles and behaviours
2. Handling this on the app side would look something like `input height smaller that characters inside of it with visible overflow` which is not ideal solution + will most likely cause issues on some devices/os versions etc
3. If we would go with something like solution above, there will be most likely scenarios where emojis are overlapping in case of multiline inputs

Also, please keep in mind that emojis have different proportions so even if we decide to address this issue on the lib level (better call in my opinion) this would be more of an "close enough" solution rather that "works perfect for every possible emoji"

cc @tomekzaw wdyt?

shawnborton:
Well before diving into the technical specs here, I think it's completely reasonable to expect that everything is vertically centered and that emoji's don't get a much higher cursor/line height than normal text. Have you tried making it so that even if the emoji is bigger, it retains the same line height size as normal text?

shawnborton:
Notice how this example feels way more aligned: 
CleanShot 2024-05-17 at 08 08 55@2x

parasharrajat (changes requested):

dubielzyk-expensify:
> I think it's completely reasonable to expect that everything is vertically centered and that emoji's don't get a much higher cursor/line height than normal text. 

+1 to this and very much agree. It looks a tad broken with the cursor being off at the moment and I think it's worth fixing.

src/components/Composer/index.tsx
```diff
@@ -314,9 +316,10 @@ function Composer(
             scrollStyleMemo,
             StyleUtils.getComposerMaxHeightStyle(maxLines, isComposerFullSize),
             isComposerFullSize ? ({height: '100%', maxHeight: 'none' as DimensionValue} as TextStyle) : undefined,
+            textContainsOnlyEmojis ? {paddingBottom: 0} : null,
```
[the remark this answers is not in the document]
kbieganowski: After changing font size inside composer (while value is emojis only) we need to override previous padding so emojis are centered in relation to input and size of composer is not increased, for comparison: 
with 0 padding:
Screenshot 2024-05-20 at 12 17 02

without padding change:
Screenshot 2024-05-20 at 12 16 45

src/libs/EmojiUtils.ts
```diff
@@ -569,7 +572,57 @@ function getSpacersIndexes(allEmojis: EmojiPickerList): number[] {
...
+};
+
+function splitTextWithEmojis(text = ''): TextWithEmoji[] {
+    if (!text) {
+        return [];
+    }
+
+    // The regex needs to be cloned because `exec()` is a stateful operation and maintains the state inside
+    // the regex variable itself, so we must have a independent instance for each function's call.
+    const emojisRegex = new RegExp(CONST.REGEX.EMOJIS, CONST.REGEX.EMOJIS.flags.concat('g'));
```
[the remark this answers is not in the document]
kbieganowski: While working on implementation of splitting logic, @fabioh8010 noticed that `g` flag in regex maintain state between `test()` calls that resulted in different outcomes even if input was the same. We decided that it would be a better/safer option to remove `g` flag and add it when function is called (other functions that are using this regex were updated as well) so we can avoid random bugs

src/styles/variables.ts
```diff
@@ -112,6 +115,9 @@ export default {
     lineHeightSizeh1: getValueUsingPixelRatio(28, 32),
     lineHeightSizeh2: getValueUsingPixelRatio(24, 28),
     lineHeightSignInHeroXSmall: getValueUsingPixelRatio(32, 37),
+    lineHeightComment: 23,
```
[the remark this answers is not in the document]
kbieganowski: The value itself is a magic number, to add a little bit of a context to why this change is made - when we increase `fontSize` for emojis, the issue with variable line height for multiline text with emojis (also on current main) was much more noticeable + in some cases emojis were touching/overlapping with line above. I've increased `lineHeight` a little bit for comments both with and without emojis to have consistent UI. I went with 23px because it made the issue far less noticeable and at the same time it's not drastic change on UI.
Example from current main - notice 2nd comment from the top and how spacing between 1st and 2nd line is higher than the rest of the same comment. When we make emojis bit bigger than the rest of the text, this spacing increases and is much more noticeable 
main_ios_workspace

src/pages/settings/Profile/DisplayNamePage.tsx
```diff
@@ -112,6 +112,7 @@ function DisplayNamePage({isLoadingApp = true, currentUserPersonalDetails}: Disp
                             role={CONST.ROLE.PRESENTATION}
                             defaultValue={currentUserDetails.lastName ?? ''}
                             spellCheck={false}
+                            isMarkdownEnabled
```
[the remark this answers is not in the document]
kbieganowski: First name input was not addressed in the ticket but if we want to handle both inputs and not only last name, please put it in the related issue and I'll take care of it

src/components/Composer/index.native.tsx
```diff
@@ -34,8 +36,9 @@ function Composer(
 ) {
     const textInput = useRef<AnimatedMarkdownTextInputRef | null>(null);
     const {isFocused, shouldResetFocus} = useResetComposerFocus(textInput);
+    const textContainsOnlyEmojis = containsOnlyEmojis(value ?? '');
```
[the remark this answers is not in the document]
kbieganowski: I'm wondering if that would benefit us in any way since value would be a dependency so on every keystroke it will run anyway

src/components/Composer/index.tsx
```diff
@@ -75,9 +76,10 @@ function Composer(
     }: ComposerProps,
     ref: ForwardedRef<TextInput | HTMLInputElement>,
 ) {
+    const textContainsOnlyEmojis = containsOnlyEmojis(value ?? '');
```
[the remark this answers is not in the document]
kbieganowski: https://github.com/Expensify/App/pull/40692#discussion_r1606749451

src/pages/home/report/ReportActionItemFragment.tsx
```diff
@@ -156,18 +158,30 @@ function ReportActionItemFragment({
                 );
             }
 
+            const containEmoji = CONST.REGEX.EMOJIS.test(fragment.text);
```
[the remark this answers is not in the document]
kbieganowski: > Why don't we need a g flag here like elsewhere?

my bad, updated

> Also, shouldn't we use `EmojiUtils.containsOnlyEmojis` here?

Here, we are checking username (to be rendered above comment in workspace history) so we need to know if username string contains any emojis (and not emojis only)

src/pages/home/report/comment/TextCommentFragment.tsx
```diff
@@ -10,13 +10,14 @@ import useThemeStyles from '@hooks/useThemeStyles';
 import useWindowDimensions from '@hooks/useWindowDimensions';
 import convertToLTR from '@libs/convertToLTR';
 import * as DeviceCapabilities from '@libs/DeviceCapabilities';
-import * as EmojiUtils from '@libs/EmojiUtils';
```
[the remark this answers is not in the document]
kbieganowski: sure! updated 🙂

kbieganowski:
Hello @shawnborton and @dubielzyk-expensify! Regarding the topic of cursor and emojis alignment - I totally agree that this is something that should be addressed. That being said, I still believe that handling it on `react-native-live-markdown` side is only proper way of doing this so we have 1 solution working everywhere instead of fixing it separately in the future, only non-hacky solution I'm seeing for handling it on the App side would be increased `fontSize` (so higher cursor) as long as we have only emojis or emojis + spaces inside composer, if we want to have enhanced emoji and smaller cursor aligned vertically inside of the input and handle it on the App side, I don't see solution for that at the minute.

One thing to note - this behaviour is not strictly related to this PR only as it was merged to main few days back (recording from main for comparison)

https://github.com/Expensify/App/assets/26638539/78aa4767-5cb2-4bdf-a8a1-395f96596ae5

@parasharrajat I will address remaining comments you posted 1st thing tomorrow, thanks for your input!

shawnborton:
Okay I think that's fair, will float this in the markdown room in Slack for some further consideration too.

src/pages/home/report/comment/TextWithEmojiFragment.tsx
```diff
@@ -0,0 +1,85 @@
...
+    message: string;
+
+    /** Additional styles to add after local styles. */
+    passedStyles?: StyleProp<TextStyle>;
+
+    /** Should this message fragment be styled as deleted? */
+    styleAsDeleted?: boolean;
+
+    /** Should this message fragment be styled as muted? */
+    styleAsMuted?: boolean;
```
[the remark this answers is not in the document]
kbieganowski: I'm not sure I have a good understanding of this suggestion, to add a bit of context: this component is based on `src/pages/home/report/comment/TextCommentFragment.tsx` where we get `style`, `styleAsMuted` and `styleAsDeleted` props separately (passed down to `TextWithEmojiFragment`). To me it seems like some additional steps that won't really benefit us (at this point in time) and will differentiate 2 components that should (in my opinion) look as similar as possible from DX perspective, hope this makes sense 😅

With the amount of props this component is accepting I think we are good without packing them together but if I'm missing something please let me know 🙏

kbieganowski:
Hi @parasharrajat, it seems like https://github.com/Expensify/App/pull/40692#issuecomment-2120041182 is the last one that needs changes in codebase, I will take care of this one as well as conflicts today. After that I'll check how https://github.com/Expensify/App/pull/40692#issuecomment-2120046394 behaves in `react-native-live-markdown` example app. Everything else is either fixed or is/will be addressed but I don't think additional changes in codebase will be needed

VickyStash:
Hey @parasharrajat, I'm taking over the work on this PR. I'm going to re-check [the list](https://github.com/Expensify/App/pull/40692#issuecomment-2122684754) of the reported issues and ping you when the PR is ready for another round of review.

VickyStash:
> I think we're running into some competing issues from the other PR I mentioned, but @VickyStash can you check out my comment [here](https://github.com/Expensify/App/pull/42615#issuecomment-2135581158) and let me know what you think?

@shawnborton As another [dev mentioned](https://github.com/Expensify/App/pull/42615#issuecomment-2136438041) your solution won't work as expected on iOS. But I've looked through your discussion in that PR and it sounds good to me.

src/pages/settings/Profile/DisplayNamePage.tsx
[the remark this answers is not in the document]
VickyStash: @parasharrajat @roryabraham So, should updates also be applied to the first name?

VickyStash:
@parasharrajat I think all of the issues from [the list](https://github.com/Expensify/App/pull/40692#issuecomment-2135047464) have been addressed, except for the one we are waiting for designer approval. Could you please take another look?

Btw, I've noticed a weird issue on the main branch with jumping elements in the composer, don't get confused by it:

https://github.com/Expensify/App/assets/23176449/8173fda4-d12e-4a32-8321-5076e7a777a6

shawnborton:
Which issue needs designer approval? 

Also, I am curious if you can give my suggestion from the other PR a try and see if that works everywhere.

VickyStash:
> BUG: Vertical alignment does not look correct on the actor name on the chat message.
> 
> Notice the applied alignment and decide which one looks better.
> 
>  Screen.Recording.2024-05-19.at.3.04.42.PM.mov

@shawnborton About design approval: @parasharrajat suggested applying bottom vertical alignment to icons in the user name, what do you think?

shawnborton:
Ah so this comment [here](https://github.com/Expensify/App/pull/40692#issuecomment-2119170089)... hmm, cc @Expensify/design - I don't hate that at all, it seems to feel nice? Does it break anywhere, and does it work consistently on all platforms?

dubielzyk-expensify:
> Ah so this comment https://github.com/Expensify/App/pull/40692#issuecomment-2119170089... hmm, cc @Expensify/design - I don't hate that at all, it seems to feel nice? Does it break anywhere, and does it work consistently on all platforms?

I'm with Shawn that it works in that context. Just not sure what the repercussions are across the rest of the app. But if it works everywhere then I think it looks better

VickyStash:
@parasharrajat Sorry for the delay, I had a day off yesterday. I'm testing the app after conflicts resolution right now, so it should be pushed soon

VickyStash:
@parasharrajat I've noticied that `react-native-live-markdown` upgrade to 0.1.76 version [was reverted](https://github.com/Expensify/App/pull/42843) yesterday, so the latest main doesn't have that changes. It's causing some issues that [were resolved](https://github.com/Expensify/App/pull/42454#issuecomment-2132571060) in that version. It's really noticeable during the testing of the composer on ios.
Do you have an idea what can we do?

VickyStash:
@parasharrajat I've resolved conflicts, please take another look. But keep in mind `react-native-live-markdown` issues. 
It maybe make sense to do testing using the 0.1.76 version of `react-native-live-markdown`, especially on iOS.

> Does that version also have your fixes for issues related to this PR?

It doesn't have my fixes, but it has important [fixes](https://github.com/Expensify/App/pull/42454#issuecomment-2132571060) overall. Without them you can meet a couple of weird issues related to emoji sizes and line height in the composer.

VickyStash:
@parasharrajat kind bump 🙂

VickyStash:
@parasharrajat 
#### List of issues that are reproduced in the `@expensify/react-native-live-markdown`
1. Comment with the app issue: https://github.com/Expensify/App/pull/40692#issuecomment-2119169077
Trimmed space after the emoji and cursor jumping to the end issues are related to single-line input. The issues are resolved if I pass `multiline` flag to `InputWrapper`. I was able to reproduce the same behavior in the `react-native-live-markdown` example app with `multiline` disabled, so it's lib related.
<details>
<summary>Video</summary>

https://github.com/Expensify/App/assets/23176449/be8959cb-d05b-47a0-8b27-646b39fc55a4


</details>

2. Comment with the app issue: https://github.com/Expensify/App/pull/40692#issuecomment-2120046394
IOS: Input text is dancing while moving the cursor.
Reproduced in the `react-native-live-markdown` example app as well.
<details>
<summary>Video</summary>

https://github.com/Expensify/App/assets/23176449/2ac09ece-5d0d-4a46-9f49-5e0c7e4d5660
</details>

3.  Comment with the app issue: https://github.com/Expensify/App/pull/40692#issuecomment-2120043299
IOS: Trimming and line spacing issues on iOS for message in composer.
I've applied the same `fontSize`, `fontFamily` and `lineHeight` in the example app and got the same result.

<details>
<summary>Screenshots</summary>
Test input
E:App input
</details>

#### Issues that can be resolved in a separate PR
1. Comment with the app issue: https://github.com/Expensify/App/pull/40692#issuecomment-2119168151 
There was a [discussion](https://github.com/Expensify/App/issues/4114#issuecomment-2112767032) about issue with how comments are rendered at the moment, decision was made that this will be investigated and fixed in separate PR

I'm going to look through newly reported issues, so maybe this list will be updated.
Maybe we should create separate tickets for the lib issues?

VickyStash:
Updates:
- resolved conflicts
- addressed issues that @parasharrajat has found

shawnborton:
> Whether it should be centered or bottom-aligned.

I think once everything starts using the same size/line height, the default would be that it is aligned at the baseline, right?

Also just flagging @j-piasecki for the emoji clipping, in case you have any good ideas there.

VickyStash:
@parasharrajat @shawnborton 
Overall, I've updated the alignment to fix [this problem](https://github.com/Expensify/App/pull/40692#issuecomment-2115894580).

> I think once everything starts using the same size/line height, the default would be that it is aligned at the baseline, right?

So which way it look better (or none of them):
| Current (middle)  | Middle for icons + baseline for a text with icons |
|option_1_1 | option_2_1 |
|option_1_2  | option_2_2  |

shawnborton:
Can you show that with multiple lines of text please? 

Here's how GH handles this:

Use😄 what you see, do😄n't plan it. Let your imagination 😄 be your guide. We need dark in order to 😄 show light. If these lines aren't straight, your water's going to r😄un right out of your painting 😄and get your floor wet.

VickyStash:
@parasharrajat Regarding the display name clipped/not aligned issue - I'm still looking for a better solution.
**Why it's happening overall:**  iOS emojis have their own vertical padding, which is not consistent with the font family we use. And when we set `lineHeight`, even if it's larger then the `fontSize` emoji can't fit in. But for example, with `System` font, it works as expected.

**Solutions:**
1. One of the right away solutions is increasing of the `lineHeight`. Similar updates already were done for the similar issues, [example of the PR](https://github.com/Expensify/App/pull/32818). It works, but there can be [side effects](https://github.com/Expensify/App/pull/32818/files#r1443344823) in UI and the misalignment visual effect still exists due to emojis padding. Examples:
image

image

image

2. Display emojis on ios with a little bit smaller `fontSize` inside the text so it can fit. This approach will require more updates and the creation of a utility function to handle it. Example:
image

What do you think @parasharrajat ?

shawnborton:
> But for example, with System font, it works as expected.

What does that mean exactly? Are you saying that if we explicitly set a different font family just for emojis, it would work?

VickyStash:
> What does that mean exactly? Are you saying that if we explicitly set a different font family just for emojis, it would work?

@shawnborton No, I mean that if we overall use System `fontFamily` it works. I guess that out custom fonts doesn't include emojis, so ios uses Apple Color Emoji font for them.

| Current font  | System font |
| image | image  |

VickyStash:
@shawnborton 

> The most important thing being that all of the text + emoji has the same line height, which you can see when you select the text

Right now it looks this way when I select it:
image

But I can increase `lineHeight` a little, so it looks like this:
image

Does it work for you?

shawnborton:
> But I can increase lineHeight a little, so it looks like this:

Can we just make it so that the emojis use the same exact line height as our font does?

j-piasecki:
> Also just flagging @j-piasecki for the emoji clipping, in case you have any good ideas there.

It's very likely that solving https://github.com/Expensify/App/issues/43371 will fix that, since it looks like the baseline is moved a bit up on the new architecture for certain fonts.

shawnborton:
Oh well that's great to hear as well!

shawnborton:
Our type designers tell us that it's not a problem with the font files themselves. They passed along these two links in case they might be helpful:
- https://stackoverflow.com/questions/33821006/emoji-is-clipped-in-uilabel-in-ios
- https://github.com/callstack/react-native-paper/issues/2495

VickyStash:
> Can we just make it so that the emojis use the same exact line height as our font does?

The thing is in the way the emojis are rendered. For example in Safari it looks as expected with current setup, but in Chrome emojis has height ~22.5. I've tried to update the height and display with different combinations, but there are some issues with multiline texts, so I'm still investigating.

| Safari  | Chrome |
| image | image|
| image  | image image |

Other updates as for today:

- Fixed date alignment [issue](https://github.com/Expensify/App/pull/40692#issuecomment-2163755954) on android
- It looks like @j-piasecki found a [fix](https://github.com/Expensify/App/issues/43371#issuecomment-2165961072) for emojis are cut off, I'm going to give it a try tomorrow

shawnborton:
Sounds good, thank you for the continued updates!

VickyStash:
I've tried to apply @j-piasecki [solution](https://github.com/Expensify/App/issues/43371#issuecomment-2165961072), but it's still cut off. I've tried with a simple RN app with suggested updates and text styles from E/App:
```
 fontSize: 13,
 lineHeight: 16,
 fontFamily: 'ExpensifyNeue-Regular',
 backgroundColor: 'gray',
 color: 'black'
```
and had the same result

image

<details>
<summary>Code snippet</summary>

```
import React from 'react';
import {
  Text,
  View,
} from 'react-native';

const App = () => {
  return (
    <View style={{
      paddingTop: 100,
    }}>
      <Text style={{fontSize: 13, lineHeight: 16, fontFamily: 'ExpensifyNeue-Regular', backgroundColor: 'gray', color: 'black'}}>helllo 😞 B</Text>
    </View>
  );
};

export default App;
```
</details>

@j-piasecki Did I make something wrong?

VickyStash:
@shawnborton I still have no luck with using same lineHeight for emojis and text without offset during text selection, so maybe we can consider [increasing](https://github.com/Expensify/App/pull/40692#issuecomment-2163146756) lineHeight a little in case text contains emojis?

shawnborton:
I would still prefer to have emoji and text use the same exact line height, specifically 20px which I think @j-piasecki will have merged into main soon.

@VickyStash can you remind me again what happens on each platform when everything uses 20px line height?

VickyStash:
@shawnborton With @kbieganowski changes applied it looks good in places with 20px lineHeight:
image

But the emojis are still a little bit cut off on _iOS_ in the places with:
|  |  |
| fontSize: 11,<br>**lineHeight: 14**  | image  |
| fontSize: 13,<br>**lineHeight: 16** | image  |

shawnborton:
> But the emojis are still a little bit cut off on iOS in the places with:

Okay cool, and do we know _why_ that is happening?

For the two examples you've shown, I think we can easily change the line height to 20px for the Search page mobile view - that won't really impact how that line looks. I am a bit hesitant to change the line height on the message preview line though, so I'm curious if you have good alternative ideas there. Can we make the emoji font size 1px smaller or something?

j-piasecki:
> Okay cool, and do we know why that is happening?

A bug in RN core causes this - on the new architecture, the baseline calculations are done slightly differently than on the old one. Previously, when emojis and text were rendered in the same component, a native class was used to parse it which divided the content into parts based on the font that will be used to render it. This way, emojis were treated as a separate span that was using Apple's emoji font with different size than the text. On the new arch this step is currently being skipped, so the text is being treated a whole with one font family, and thus the same size. Now, when baseline offset is being calculated to center the text in the line, the algorithm isn't aware of emojis being larger than the text, so it effectively skips them during calculations, resulting in them being cut off.

It should be covered by https://github.com/Expensify/App/pull/43902, which uses the same flow as the old architecture.

shawnborton:
Oh sweet, that is great news then!

VickyStash:
@shawnborton What do you think about this approach?

|  |  |
| lineHeight: 14 updated to -> lineHeight: 16  | image  |
| Emoji font size changed from 13px -> 12px, the lineHeight is the same as before | image |

shawnborton:
@VickyStash that looks great to me, but I'm curious what you think about @j-piasecki's comment above too.

VickyStash:
> @VickyStash that looks great to me, but I'm curious what you think about @j-piasecki's comment above too.

I'm testing it with updates from PR @j-piasecki mentioned. And as I mentioned in [this comment](https://github.com/Expensify/App/pull/40692#issuecomment-2178032380), even with that updates there are some cases when emojis are still cut off.

shawnborton:
Got it, okay. Then I think your solutions should be okay, and I would opt to only do them on the platforms where we deem them necessary.

j-piasecki:
I've quickly checked the old architecture and the emojis are also cut there, so it seems unrelated to the new arch. Sorry for the confusion.

VickyStash:
@shawnborton @parasharrajat It looks like the only problem which is left for now is the text selection in composer in Chrome. I've tried different styles and had no luck, also it is reproduced in the `react-native-live-markdown` lib on its own. The only thing that gives the expected result is increasing the `lineHeight`.
image
image

shawnborton:
Interesting, so that happens even when the emoji has the same line height as the text?

VickyStash:
@shawnborton Yes, there is no specific `lineHeight` set for the emojis in the composer. And even if I separately specify it to be the same as the text's - the result is the same.

shawnborton:
Got it, and to clarify, this is fine on production because we aren't increasing the emoji size, right?
CleanShot 2024-06-19 at 17 56 00@2x

I'm not really sure how to solve that... I just tried playing around with a few things and I can't seem to fix it. I think as long as the emoji has the same line height as the text and the larger emoji doesn't increase the gaps between lines, we can let this one slide.

VickyStash:
> Got it, and to clarify, this is fine on production because we aren't increasing the emoji size, right?

@shawnborton Right, increased emoji size is causing the issue.

VickyStash:
@parasharrajat Could you please take another look?

VickyStash:
Note: I'll be OOO on Monday (24.06) 🌴

VickyStash:
> We just need to bump version for react-native-live-markdown

@parasharrajat there is a [separate PR](https://github.com/Expensify/App/pull/44262) for bumping `react-native-live-markdown` version, so I think we can just wait till it's merged.

VickyStash:
Note: I'll be OOO Jul 1 - Jul 7 🌴

VickyStash:
@parasharrajat `react-native-live-markdown` got merged, please take a look 🙏

VickyStash:
@parasharrajat kind bump 🙂

VickyStash:
@parasharrajat I didn't see any conflicts, but I've merged the main just in case

VickyStash:
@parasharrajat kind bump 🙂

VickyStash:
@parasharrajat any ETA on the review?

VickyStash:
@parasharrajat main is merged. Unfortunately, I can't update the PR's title because I'm not its author.

VickyStash:
@parasharrajat 

> Bug: So chrome browser does not show large emojis mixed with text. Safari works...

That's how it looks to me:
image
image

I've checked and the larger size for icons is applied.

> BUG: May be my eyes are fooled. Emojis mixed with text have the same height.

That's how it looks to me
Simulator Screenshot - iPhone 15 - 2024-07-24 at 16 33 05

But I agree your screenshots look different, are you sure you are using the right branch and the set up is correct?

VickyStash:
@parasharrajat For me it's not a single div
image

parasharrajat (approved):
## Reviewer Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
- [x] If the PR modifies the form input styles:
    - [x] I added `Design` label so the design team can review the changes.

:ribbon: :eyes: :ribbon:  C+ reviewed

src/pages/home/report/comment/TextWithEmojiFragment.tsx
```diff
@@ -0,0 +1,85 @@
...
+    isSmallScreenWidth?: boolean;
+
+    /** Should "(edited)" suffix be rendered? */
+    isEdited?: boolean;
+
+    /** Does message contain only emojis? */
+    emojisOnly?: boolean;
+};
+
+function TextWithEmojiFragment({message, passedStyles, styleAsDeleted, styleAsMuted, isSmallScreenWidth, isEdited, emojisOnly}: TextWithEmojiFragmentProps) {
```
roryabraham, line 35: Let's `useWindowDimensions` in the component rather than passing `isSmallScreenWidth` as a prop.

src/pages/settings/InitialSettingsPage.tsx
```diff
@@ -367,6 +369,11 @@ function InitialSettingsPage({session, userWallet, bankAccountList, fundList, wa
     const currentUserDetails = currentUserPersonalDetails;
```
roryabraham, line 369: I know you didn't write this code, but this line seems useless, no?

src/pages/settings/InitialSettingsPage.tsx
```diff
@@ -367,6 +369,11 @@ function InitialSettingsPage({session, userWallet, bankAccountList, fundList, wa
...
     const avatarURL = currentUserDetails?.avatar ?? '';
     const accountID = currentUserDetails?.accountID ?? '-1';
+    const usernameContainEmojis = CONST.REGEX.EMOJIS.test(currentUserPersonalDetails?.displayName ?? '');
+    let processedTextArray: TextWithEmoji[] = [];
+    if (usernameContainEmojis) {
```
roryabraham, line 374: Please go through your PR and make sure all boolean variable names start start with `is`, `are`, `has`, `was`, `does`, etc...

src/pages/settings/InitialSettingsPage.tsx
```diff
@@ -367,6 +369,11 @@ function InitialSettingsPage({session, userWallet, bankAccountList, fundList, wa
...
+        processedTextArray = splitTextWithEmojis(currentUserPersonalDetails?.displayName ?? '');
```
roryabraham, line 375: Memoize please:

```diff
diff --git a/src/pages/settings/InitialSettingsPage.tsx b/src/pages/settings/InitialSettingsPage.tsx
index 57e33d3d79..db7de59845 100755
--- a/src/pages/settings/InitialSettingsPage.tsx
+++ b/src/pages/settings/InitialSettingsPage.tsx
@@ -366,14 +366,16 @@ function InitialSettingsPage({session, userWallet, bankAccountList, fundList, wa
     const generalMenuItems = useMemo(() => getMenuItemsSection(generalMenuItemsData), [generalMenuItemsData, getMenuItemsSection]);
     const workspaceMenuItems = useMemo(() => getMenuItemsSection(workspaceMenuItemsData), [workspaceMenuItemsData, getMenuItemsSection]);
 
-    const currentUserDetails = currentUserPersonalDetails;
-    const avatarURL = currentUserDetails?.avatar ?? '';
-    const accountID = currentUserDetails?.accountID ?? '-1';
-    const usernameContainEmojis = CONST.REGEX.EMOJIS.test(currentUserPersonalDetails?.displayName ?? '');
-    let processedTextArray: TextWithEmoji[] = [];
-    if (usernameContainEmojis) {
-        processedTextArray = splitTextWithEmojis(currentUserPersonalDetails?.displayName ?? '');
-    }
+    const avatarURL = currentUserPersonalDetails?.avatar ?? '';
+    const accountID = currentUserPersonalDetails?.accountID ?? '-1';
+
+    const processedTextArray: TextWithEmoji[] = useMemo(() => {
+        const doesUsernameContainEmojis = CONST.REGEX.EMOJIS.test(currentUserPersonalDetails?.displayName ?? '');
+        if (!doesUsernameContainEmojis) {
+            return [];
+        }
+        return splitTextWithEmojis(currentUserPersonalDetails?.displayName ?? '');
+    }, [currentUserPersonalDetails?.displayName]);
 
     const headerContent = (
         <View style={[styles.avatarSectionWrapperSettings, styles.justifyContentCenter, styles.ph5, styles.pb5]}>
@@ -423,7 +425,7 @@ function InitialSettingsPage({session, userWallet, bankAccountList, fundList, wa
                     </View>
                     <View style={[styles.mb3, styles.w100]}>
                         <AvatarWithImagePicker
-                            isUsingDefaultAvatar={UserUtils.isDefaultAvatar(currentUserDetails?.avatar ?? '')}
+                            isUsingDefaultAvatar={UserUtils.isDefaultAvatar(currentUserPersonalDetails?.avatar ?? '')}
                             source={avatarURL}
                             avatarID={accountID}
                             onImageSelected={PersonalDetails.updateAvatar}
@@ -436,13 +438,13 @@ function InitialSettingsPage({session, userWallet, bankAccountList, fundList, wa
                             onErrorClose={PersonalDetails.clearAvatarErrors}
                             onViewPhotoPress={() => Navigation.navigate(ROUTES.PROFILE_AVATAR.getRoute(String(accountID)))}
                             previewSource={UserUtils.getFullSizeAvatar(avatarURL, accountID)}
-                            originalFileName={currentUserDetails.originalFileName}
+                            originalFileName={currentUserPersonalDetails.originalFileName}
                             headerTitle={translate('profilePage.profileAvatar')}
-                            fallbackIcon={currentUserDetails?.fallbackIcon}
+                            fallbackIcon={currentUserPersonalDetails?.fallbackIcon}
                             editIconStyle={styles.smallEditIconAccount}
                         />
                     </View>
-                    {currentUserPersonalDetails?.displayName && usernameContainEmojis ? (
+                    {processedTextArray.length !== 0 ? (
                         <Text
                             style={[styles.textHeadline, styles.pre, styles.textAlignCenter]}
                             numberOfLines={1}
```

src/pages/settings/Profile/DisplayNamePage.tsx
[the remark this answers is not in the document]
roryabraham: Yes, let's also enable it for first name

roryabraham (changes requested):

VickyStash:
@roryabraham I've applied your suggestions, please take another look

src/pages/home/report/ReportActionItemMessageHeaderSender.tsx
```diff
@@ -0,0 +1,56 @@
...
+                numberOfLines={isSingleLine ? 1 : undefined}
+                style={[styles.chatItemMessageHeaderSender, isSingleLine ? styles.pre : styles.preWrap]}
+            >
+                {processedTextArray.length !== 0 ? processedTextArray.map(({text, isEmoji}) => (isEmoji ? <Text style={styles.emojisWithinDisplayName}>{text}</Text> : text)) : fragmentText}
+            </Text>
+        </UserDetailsTooltip>
+    );
+}
+
+export default ReportActionItemMessageHeaderSender;
```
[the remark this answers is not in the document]
VickyStash: Thank you @parasharrajat, updated

parasharrajat (approved):

roryabraham (approved):

VickyStash:
> @VickyStash How are we going to handle follow-up fixes? [#40692 (comment)](https://github.com/Expensify/App/pull/40692#issuecomment-2160381418)

@parasharrajat I can open the issues for the `@expensify/react-native-live-markdown` related problems.
As for the app-related one I'm going to investigate it today
What do you think?

VickyStash:
@parasharrajat The follow-up [PR](https://github.com/Expensify/App/pull/46398) for the app is ready, please take a look

VickyStash:
@parasharrajat I've created two issues in `@expensify/react-native-live-markdown` lib:

https://github.com/Expensify/react-native-live-markdown/issues/443
https://github.com/Expensify/react-native-live-markdown/issues/444

I have some doubts about the [input dancing bug](https://github.com/Expensify/App/pull/40692#issuecomment-2120046394) cause today, during re-testing, I had an idea of how it could be fixed by updating styles on the app side. If it works, I'll open one more follow-up

Issue #41254, opened 2024-04-29
CRITICAL: [API Reliability] Workspace room does not load older messages due to infinite `GetOlderActions` calls


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 1.4.67-2
**Reproducible in staging?:** Needs reproduction
**Reproducible in production?:** Needs reproduction
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @shawnborton 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1714419709483439

## Action Performed:
1. Open new dot app
2. Go to the workspace chat NewDot Test - Expensify US policy`
## Expected Result:
The workspace chat should load all historical messages and requests
## Actual Result:
The workspace chat is stuck looping through OpenReport over and over again. It will not load previous messages or requests, but it will load a recent receipt I uploaded today
## Workaround:
Unknown
## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android: Native
- [ ] Android: mWeb Chrome
- [ ] iOS: Native
- [ ] iOS: mWeb Safari
- [X] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

https://github.com/Expensify/App/assets/38435837/cb79dde2-2cb4-43a3-897e-86b2c86cbbbd

[logs-2024-04-29 19_35_32.431.txt](https://github.com/Expensify/App/files/15155168/logs-2024-04-29.19_35_32.431.txt)

Add any screenshot/video evidence

</details>

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @shawnborton</details>

MelvinBot:
This has been labelled "Needs Reproduction". Follow the steps here: https://stackoverflowteams.com/c/expensify/questions/16989

shawnborton:
cc @JmillsExpensify in case you want to add your case here too

[...]

Issue #41174, opened 2024-04-28
[HOLD #41254][$250] IOS - After editing split expense amount weird flash and reloads at the top


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 1.4.67-0
**Reproducible in staging?:** y
**Reproducible in production?:** y
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @roryabraham 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1714161321481359

## Action Performed:

1. Create a group
2. Add messages to the chat such that there is at least one full screen of chat messages.
3. Scan a receipt to split a bill in the group.
4. Manually set the amount on the bill split

## Expected Result:
The amount should update, and nothing else should happen.
## Actual Result:
The amount updates, then the report "flashes" and reloads at the top, then scrolls down to the bottom, without you touching anything.
## Workaround:
unknown
## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android: Native
- [ ] Android: mWeb Chrome
- [x] iOS: Native
- [ ] iOS: mWeb Safari
- [ ] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

Add any screenshot/video evidence

https://github.com/Expensify/App/assets/38435837/82bec8e2-0c90-42c9-9b47-4b398a09bc7f

</details>

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0174f92c2a5d39efca</li>
        <li>Upwork Job ID: 1784657619645915136</li>
        <li>Last Price Increase: 2024-05-05</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @roryabraham</details>

eh2077:
[...]

Issue #40644, opened 2024-04-20
[$250] Comments after threaded comment are gone after returning to parent


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** v1.4.63-13
**Reproducible in staging?:** y
**Reproducible in production?:** y
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @aldo-expensify 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1713557851562149

## Action Performed:
1. Join public room, for example: https://staging.new.expensify.com/r/6776919265133947
2. Find a threaded comment that is not the last comment in the chat and click it to open the thread
3. Click on the header to come back to the parent chat
4. Notice that the thread comment is highlighted and you can't scroll down anymore

## Expected Result: 
I would have expected to be able to scroll down again in the parent report without having to click "New Messages"

## Actual Result: 
Can't scroll down, the threaded comment is now the last comment in the parent chat, and I have to click in "New Messages" for it to load more comments and scroll down.

## Workaround:
unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android: Native
- [ ] Android: mWeb Chrome
- [ ] iOS: Native
- [ ] iOS: mWeb Safari
- [x] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

Add any screenshot/video evidence

https://github.com/Expensify/App/assets/38435837/a334938a-0837-4a53-9f3f-bd6c7d11e7ae

https://github.com/Expensify/App/assets/38435837/9bae8a8c-540a-4b85-af46-5ca5aa069229

</details>

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~0155f72908fd93dc77</li>
        <li>Upwork Job ID: 1782728194457100288</li>
[...]

PR #41962, opened 2024-05-10
Client side pagination gap detection

janicduplessis:
### Details
Listing out related or potentially related issues, won't move these into `Fixed Issues` unless we test and confirm:

- https://github.com/Expensify/App/issues/41254
- https://github.com/Expensify/App/issues/40644
- https://github.com/Expensify/App/issues/42214
- https://github.com/Expensify/App/issues/41174
- https://expensify.slack.com/archives/C05LX9D6E07/p1718059029709639

### Fixed Issues
$
PROPOSAL: https://docs.google.com/document/d/19xEEr0mY41cTio1lGgJ2F6vYCA5QH4Sj7hOMRZsj0Dc/edit

### Tests

### QA Steps

Test 1: Basic Navigation

- Create a chat between Account A and Account B.
- Send 200 messages from Account A to Account B.
- Copy the link to the 70th message and send it to Account B (first link).
- Copy the link to the 150th message and send it to Account B (second link).
- From Account A, click on the first link. Verify redirection to the exact message.
- From Account A, click on the second link. Verify redirection to the exact message.
- From Account B, click on the first link. Verify redirection to the exact message.
- From Account B, click on the second link. Verify redirection to the exact message.

Note: Links can be copied to any message and used in any order.

Test 2: Popup Navigation

- Continue using the chat from Test 1.
- From Account A, click on the first link. Verify redirection to the exact message.
- After redirection, a popup should appear at the top of the screen (similar to new message popup).
- Click on this popup.
- Verify navigation to the end of the list (newest messages).

Test 3: Cache and Logout

- Logout from the accounts and clear the cache.
- Log back in.
- From Account A, click on the first link. Verify redirection to the exact message.
- From Account A, click on the second link. Verify redirection to the exact message.
- From Account B, click on the first link. Verify redirection to the exact message.
- From Account B, click on the second link. Verify redirection to the exact message.

Test 4:

- Create a workspace as User A.
- Invite User B.
- Navigate to the created Workspace Chat as User B.
- Observe User B has a whisper message.
- Send a few messagse as User B inside the workspace chat.
- Navigate to Chat as User A.
- Observe User A does not see any whisper messages nor have them in their Onyx data.
- User A should see the previous actions

Notes:
Ensure the chat contains at least 150 messages. Initially, up to 50 messages are rendered. Navigating to the 70th message should allow further navigation up to the 150th message, even with gaps.
If bugs are found, confirm they are reproducible on another account, or provide credentials for debugging. Thsnks!

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/types/onyx/ReportMetadata.ts
```diff
@@ -16,6 +23,9 @@ type ReportMetadata = {
 
     /** The time when user last visited the report */
     lastVisitTime?: string;
+
+    /** Pagination info */
+    pages?: ReportMetadataPage[];
```
janicduplessis, line 28: Maybe this should be its own onyx key?

src/libs/ReportActionsUtils.ts
```diff
@@ -326,49 +304,101 @@ function getCombinedReportActions(reportActions: ReportAction[], transactionThre
...
+    });
+
+    const result = [sortedPages[0]];
+    for (let i = 1; i < sortedPages.length; i++) {
+        const page = sortedPages[i];
+        const prevPage = sortedPages[i - 1];
+
+        // Current page in inside the previous page, skip.
+        if (page.lastReportActionIndex <= prevPage.lastReportActionIndex) {
+            // eslint-disable-next-line no-continue
```
janicduplessis, line 341: Is it ok to disable this lint rule here? I thought it reads well that way.

roryabraham (commented):
The main piece of feedback I have is that I'd really love to see the Middleware generalized such that it's not really referencing reportActions directly. I think the way I set that up with `API.paginate` was looking pretty clean.

I might even say the same for the UI component. Would be great to have some business-logic-agnostic UI component like `PaginatedList` that leverages this pattern and then just make `ReportActionsList` a lightweight wrapper around that.

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,66 @@
...
+import * as ReportActionsUtils from '@libs/ReportActionsUtils';
+import ONYXKEYS from '@src/ONYXKEYS';
+import type {ReportActions, ReportMetadata, Response} from '@src/types/onyx';
+import type Middleware from './types';
+
+function getReportActions(response: Response, reportID: string) {
+    return response?.onyxData?.find((data) => data.key === `${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${reportID}`)?.value as ReportActions | undefined;
+}
+
+const Pagination: Middleware = (requestResponse, request) => {
```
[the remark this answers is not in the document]
janicduplessis: There is a response variable later that conflicts with that name so I think it is why it is named that way (that code was taken from another middleware).

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,66 @@
...
+            if (!reportID) {
+                // TODO: Should not happen, should we throw?
+                return Promise.resolve(response);
+            }
+
+            const reportActionID = request.data?.reportActionID as string | undefined;
+
+            // Create a new page based on the response actions.
+            const pageReportActions = getReportActions(response, reportID);
+            const pageSortedReportActions = ReportActionsUtils.getSortedReportActionsForDisplay(pageReportActions, true);
```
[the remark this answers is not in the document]
janicduplessis: I am using `getSortedReportActionsForDisplay` since this is what will be used in `sortedReportActions` in `ReportScreen`. I am pretty sure this is correct.

src/libs/ReportActionsUtils.ts
```diff
@@ -312,28 +312,77 @@ type PageWithIndexes = {
     lastReportActionIndex: number;
 };
 
+/**
+ * Finds the id, index in sortedReportActions and index in the page of the first valid report action in the given page.
+ */
+function findFirstReportAction(sortedReportActions: ReportAction[], page: string[]): {id: string; index: number} | null {
+    // eslint-disable-next-line @typescript-eslint/prefer-for-of
```
janicduplessis, line 319: Disabling this here so I can keep the 2 functions similar.

src/libs/API/index.ts
```diff
@@ -156,14 +180,48 @@ function makeRequestWithSideEffects<TCommand extends SideEffectRequestCommand |
...
+}
+
+function paginate<TRequestType extends ApiRequestType, TCommand extends CommandOfType<TRequestType>>(
+    type: TRequestType,
+    command: TCommand,
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData: OnyxData,
+    config: PaginationConfig,
+): TRequestType extends typeof CONST.API_REQUEST_TYPE.MAKE_REQUEST_WITH_SIDE_EFFECTS ? Promise<Response | void> : void;
+function paginate<TRequestType extends ApiRequestType, TCommand extends CommandOfType<TRequestType>>(
```
roryabraham, line 199: Potential problem with this approach is that we can't tell the difference at runtime between `OpenReport` as a side effect command and `OpenReport` as a write command

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+import type Middleware from './types';
+
+type PaginationConfig<TResourceKey extends OnyxCollectionKey, TPageKey extends OnyxPagesKey> = {
+    initialCommand: ApiCommand;
+    previousCommand: ApiCommand;
+    nextCommand: ApiCommand;
+    resourceCollectionKey: TResourceKey;
+    pageCollectionKey: TPageKey;
+    sortItems: (items: OnyxValues[TResourceKey]) => Array<OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never>;
+    getItemID: (item: OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never) => string;
```
blazejkustra, line 21: I don't understand the logic with these two types here, why not simplify it a bit?
```ts
    sortItems: (items: OnyxValues[TResourceKey]) => Array<OnyxValues[TResourceKey]>;
    getItemID: (item: OnyxValues[TResourceKey]) => string;
```

blazejkustra: Also shouldn't it be `(items: OnyxValues[TResourceKey])` in `sortItems`?

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+        },
+    });
+    Onyx.connect({
+        // TODO: Also not sure why this cast is needed.
+        key: config.pageCollectionKey as OnyxPagesKey,
+        waitForCollectionCallback: true,
+        callback: (data) => {
+            pages.set(config.pageCollectionKey, data);
+        },
+    });
```
blazejkustra, line 57:
```suggestion
    Onyx.connect<OnyxCollectionKey>({
        key: resourceCollectionKey,
        waitForCollectionCallback: true,
        callback: (data) => {
            resources.set(resourceCollectionKey, data);
        },
    });
    Onyx.connect<OnyxPagesKey>({
        key: pageCollectionKey,
        waitForCollectionCallback: true,
        callback: (data) => {
            pages.set(pageCollectionKey, data);
        },
    });
```

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+
+type PaginationConfigMapValue = Omit<PaginationConfig<OnyxCollectionKey, OnyxPagesKey>, 'initialCommand' | 'previousCommand' | 'nextCommand'> & {
+    type: 'initial' | 'next' | 'previous';
+};
+
+const paginationConfigs = new Map<string, PaginationConfigMapValue>();
+const ressources = new Map<OnyxCollectionKey, OnyxCollection<OnyxValues[OnyxCollectionKey]>>();
+const pages = new Map<OnyxPagesKey, OnyxCollection<OnyxValues[OnyxPagesKey]>>();
+
+function registerPaginationConfig<TResourceKey extends OnyxCollectionKey, TPageKey extends OnyxPagesKey>({
```
blazejkustra, line 32: It will be easier to work with if you are having TS issues:
```suggestion
function registerPaginationConfig<TResourceKey extends OnyxCollectionKey, TPageKey extends OnyxPagesKey>({
    initialCommand,
    previousCommand,
    nextCommand,
    resourceCollectionKey,
    pageCollectionKey,
    sortItems,
    getItemID,
}: PaginationConfig<TResourceKey, TPageKey>): void {
```

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+    initialCommand: ApiCommand;
+    previousCommand: ApiCommand;
+    nextCommand: ApiCommand;
+    resourceCollectionKey: TResourceKey;
+    pageCollectionKey: TPageKey;
+    sortItems: (items: OnyxValues[TResourceKey]) => Array<OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never>;
+    getItemID: (item: OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never) => string;
+};
+
+type PaginationConfigMapValue = Omit<PaginationConfig<OnyxCollectionKey, OnyxPagesKey>, 'initialCommand' | 'previousCommand' | 'nextCommand'> & {
```
blazejkustra, line 24: To simplify these types I end up with these definitions:
```ts
type PaginationCommonConfig<TResourceKey extends OnyxCollectionKey = OnyxCollectionKey, TPageKey extends OnyxPagesKey = OnyxPagesKey> = {
    resourceCollectionKey: TResourceKey;
    pageCollectionKey: TPageKey;
    sortItems: (items: Array<OnyxValues[TResourceKey]>) => Array<OnyxValues[TResourceKey]>;
    getItemID: (item: OnyxValues[TResourceKey]) => string;
};

type PaginationConfig<TResourceKey extends OnyxCollectionKey, TPageKey extends OnyxPagesKey> = PaginationCommonConfig<TResourceKey, TPageKey> & {
    initialCommand: ApiCommand;
    previousCommand: ApiCommand;
    nextCommand: ApiCommand;
};

type PaginationConfigMapValue = PaginationCommonConfig & {
    type: 'initial' | 'next' | 'previous';
};
```

blazejkustra (changes requested):

src/libs/Middleware/Pagination.ts
blazejkustra: Maybe @fabioh8010 has better ideas?

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+function registerPaginationConfig<TResourceKey extends OnyxCollectionKey, TPageKey extends OnyxPagesKey>({
+    initialCommand,
+    previousCommand,
+    nextCommand,
+    ...config
+}: PaginationConfig<TResourceKey, TPageKey>): void {
+    // TODO: Is there a way to avoid these casts?
+    paginationConfigs.set(initialCommand, {...config, type: 'initial'} as unknown as PaginationConfigMapValue);
+    paginationConfigs.set(previousCommand, {...config, type: 'previous'} as unknown as PaginationConfigMapValue);
+    paginationConfigs.set(nextCommand, {...config, type: 'next'} as unknown as PaginationConfigMapValue);
```
blazejkustra, line 41: I spent a lot of time on these today. TLDR: it's won't be easy to avoid these assertions (and not sure if it's worth it)

The real issue is that the `config` from `registerPaginationConfig` params is very specific while `paginationConfigs` is on the other spectrum quite loose:
```ts
const paginationConfigs = new Map<string, PaginationConfigMapValue>();
```
☝️ Any key in this map has a very generic value – any collection value (report, reportActions, policy etc)

This results in a mismatch between `sortItems` and `getItemID` types:
image

blazejkustra: A possible fix that would require quite a lot changes (+ additional types) is to use a [mapped object](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) instead of `new Map()`.

Here is the approach that works:
```ts
type ApiCommandToOnyxKeyMap = {
    [WRITE_COMMANDS.OPEN_REPORT]: typeof ONYXKEYS.COLLECTION.REPORT;
    [READ_COMMANDS.GET_OLDER_ACTIONS]: typeof ONYXKEYS.COLLECTION.REPORT;
    [READ_COMMANDS.GET_NEWER_ACTIONS]: typeof ONYXKEYS.COLLECTION.REPORT;
    // ... other commands ...
};
type PaginatedCommands = keyof ApiCommandToOnyxKeyMap;
type PaginatedOnyxKeys = ValueOf<ApiCommandToOnyxKeyMap>;

type PaginationConfigMap = {
    [Command in PaginatedCommands]?: PaginationConfigMapValue<ApiCommandToOnyxKeyMap[Command]>;
};

type PaginationCommonConfig<TResourceKey extends PaginatedOnyxKeys = PaginatedOnyxKeys, TPageKey extends OnyxPagesKey = OnyxPagesKey> = {
    resourceCollectionKey: TResourceKey;
    pageCollectionKey: TPageKey;
    sortItems: (items: Array<OnyxValues[TResourceKey]>) => Array<OnyxValues[TResourceKey]>;
    getItemID: (item: OnyxValues[TResourceKey]) => string;
};

type PaginationConfig<TResourceKey extends PaginatedOnyxKeys, TPageKey extends OnyxPagesKey> = PaginationCommonConfig<TResourceKey, TPageKey> & {
    initialCommand: PaginatedCommands;
    previousCommand: PaginatedCommands;
    nextCommand: PaginatedCommands;
};

type PaginationConfigMapValue<TResourceKey extends PaginatedOnyxKeys> = PaginationCommonConfig<TResourceKey, OnyxPagesKey> & {
    type: 'initial' | 'next' | 'previous';
};

const paginationConfigs: PaginationConfigMap = {};
const ressources = new Map<OnyxCollectionKey, OnyxCollection<OnyxValues[OnyxCollectionKey]>>();
const pages = new Map<OnyxPagesKey, OnyxCollection<OnyxValues[OnyxPagesKey]>>();

function registerPaginationConfig<TResourceKey extends PaginatedOnyxKeys, TPageKey extends OnyxPagesKey>({
    initialCommand,
    previousCommand,
    nextCommand,
    sortItems,
    getItemID,
    ...config
}: PaginationConfig<TResourceKey, TPageKey>): void {
    paginationConfigs[initialCommand] = {...config, sortItems, getItemID, type: 'initial'};
```

src/ONYXKEYS.ts
```diff
@@ -709,11 +711,12 @@ type OnyxFormDraftKey = keyof OnyxFormDraftValuesMapping;
 type OnyxValueKey = keyof OnyxValuesMapping;
 
 type OnyxKey = OnyxValueKey | OnyxCollectionKey | OnyxFormKey | OnyxFormDraftKey;
+type OnyxPagesKey = ConditionalKeys<OnyxValues, OnyxTypes.Pages>;
```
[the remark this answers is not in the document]
roryabraham: Well, my idea here is that there will also be a `REPORT_PAGES` key, but I didn't think about think about the potential side-effect you listed.

src/libs/API/index.ts
```diff
@@ -156,14 +180,48 @@ function makeRequestWithSideEffects<TCommand extends SideEffectRequestCommand |
...
+    type: TRequestType,
+    command: TCommand,
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData: OnyxData,
+    config: PaginationConfig,
+): Promise<Response | void> | void {
```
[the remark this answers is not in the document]
roryabraham: Looks good except only `MAKE_REQUEST_WITH_SIDE_EFFECTS` returns a Promise. Updated it to what I think is correct

src/libs/API/index.ts
```diff
@@ -40,54 +43,83 @@ type OnyxData = {
...
+    }
+
+    // Read requests are processed right away, but don't return the response to the caller
+    if (type === CONST.API_REQUEST_TYPE.READ) {
+        Request.processWithMiddleware(request);
+        return Promise.resolve();
+    }
+
+    // Requests with side effects process right away, and return the response to the caller
+    return Request.processWithMiddleware(request);
```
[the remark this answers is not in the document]
roryabraham: good suggestion!

src/libs/API/index.ts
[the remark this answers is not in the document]
roryabraham: we just cleaned it up to make this file a bit more DRY. The intention is that `API.write/read/makeRequestWithSideEffects` are unchanged. Just making it a bit simpler and cleaner to re-use their internals for `API.paginate`

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+    pageCollectionKey: TPageKey;
+    sortItems: (items: OnyxValues[TResourceKey]) => Array<OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never>;
+    getItemID: (item: OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never) => string;
+};
+
+type PaginationConfigMapValue = Omit<PaginationConfig<OnyxCollectionKey, OnyxPagesKey>, 'initialCommand' | 'previousCommand' | 'nextCommand'> & {
+    type: 'initial' | 'next' | 'previous';
+};
+
+const paginationConfigs = new Map<string, PaginationConfigMapValue>();
```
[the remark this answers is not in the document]
roryabraham: Hmmm `Request.command` is actually not an `ApiCommand` though, and making it one causes a lot of type errors.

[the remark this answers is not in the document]
roryabraham: I didn't make this change, but I added some comments to explain these maps a bit better. Hopefully that makes it clearer and easier in the future if anyone wants to refactor `Request.command` to be an `ApiCommand`.

src/libs/PaginationUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    firstID: string;
+    firstIndex: number;
+    lastID: string;
+    lastIndex: number;
+};
+
+/**
+ * Finds the id and index in sortedItems of the first item in the given page that's present in sortedItems.
+ */
+function findFirstItem<TResource>(sortedItems: TResource[], page: string[], getID: (item: TResource) => string): {id: string; index: number} | null {
```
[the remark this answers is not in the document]
roryabraham: done (though I didn't include `| null`)

src/libs/Middleware/Pagination.ts
roryabraham: The tricky thing with this type is that what we're paginating isn't actually `OnyxValues[ONYXKEYS.COLLECTION.REPORT_ACTIONS]`. because the ReportActions collection is different from others.

For example, with the Report collection, you have something like this:

```
report_1234 => Report,
report_4567 => Report,
report_${reportID} => Report,
```

and you'd want to paginate over the `report_` keys. Whereas with `ReportActions`, it's like this:

```
reportActions_${reportID} => {
    ${reportActionID} => ReportAction;
};
```

and for ReportActions you want to paginate over the collection stored within a single key

roryabraham: Figured out some utility types that might be helpful:

```ts

type ExtractPrefix<TStr extends string, TSuffix extends string> = TStr extends `${infer Prefix}${TSuffix}` ? Prefix : never;
type ExtractOnyxCollectionKey<TKey extends OnyxCollectionKey | `${OnyxCollectionKey}${string}`> = TKey extends OnyxCollectionKey ? TKey : `${ExtractPrefix<TKey, `_${string}`>}_`;
```

roryabraham: I think we may need to do cleanup here in the future if/when we want to reuse this for report pagination. But in the meantime, this works ok. Let's leave it as-is.

src/libs/Middleware/Pagination.ts
roryabraham: cool, let's :donothing: for now

src/CONST.ts
```diff
@@ -4822,6 +4822,9 @@ const CONST = {
         DESC: 'desc',
     },
 
+    PAGINATION_START_ID: '-1',
+    PAGINATION_END_ID: '-2',
```
roryabraham, line 4826: @janicduplessis I'm seeing where this value is _read_, but I'm never seeing it _set_ anywhere? What am I missing?

src/libs/PaginationUtils.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+    return null;
+}
+
+/**
+ * Finds the id and index in sortedItems of the last item in the given page that's present in sortedItems.
+ */
+function findLastItem<TResource>(sortedItems: TResource[], page: string[], getID: (item: TResource) => string): ItemWithIndex | null {
+    for (const id of page.slice().reverse()) {
```
fabioh8010, line 48: Isn't better to just do a inverted for-loop instead of copying and reverting the whole array?

src/types/onyx/Request.ts
```diff
@@ -58,5 +58,31 @@ type RequestData = {
...
+    resourceID: string;
+
+    /**
+     * The ID used as a cursor/offset when making a paginated request.
+     */
+    cursorID?: string | null;
+};
+
+/**
+ *
```
fabioh8010, line 77: Missing comment?

src/libs/Middleware/Pagination.ts
roryabraham: choosing to ignore this one since, where the `...config` is used we're doing an unsafe `as unknown as X` cast anyways.

roryabraham:
There are some type errors stemming from upgrading TypeScript on this PR. I can't recall why I did that but looking at the changes as we have them now, I doubt it's necessary. I'll remove that from this PR and create a separate one to bump TS.

tests/utils/TestHelper.ts
```diff
@@ -8,14 +8,19 @@ import ONYXKEYS from '@src/ONYXKEYS';
...
-    fail?: () => void;
-    succeed?: () => void;
-    resume?: () => Promise<void>;
+type MockFetch = jest.Mock<typeof fetch> & {
+    pause: () => void;
+    fail: () => void;
+    succeed: () => void;
+    resume: () => Promise<void>;
+    mockAPICommand: (command: string, response: OnyxResponse['onyxData']) => void;
 };
```
fabioh8010, line 17:
```suggestion
type MockFetch = ReturnType<typeof jest.fn> & {
    pause: () => void;
    fail: () => void;
    succeed: () => void;
    resume: () => Promise<void>;
    mockAPICommand: (command: string, response: OnyxResponse['onyxData']) => void;
};
```

tests/utils/TestHelper.ts
```diff
@@ -157,40 +162,62 @@ function signOutTestUser() {
...
         return waitForBatchedUpdates();
     };
     mockFetch.fail = () => (shouldFail = true);
     mockFetch.succeed = () => (shouldFail = false);
-
-    return mockFetch as typeof fetch;
+    mockFetch.mockAPICommand = (command: string, response: OnyxResponse['onyxData']) => {
+        responses.set(command, response);
+    };
+    return mockFetch;
```
fabioh8010, line 220:
```suggestion
    return mockFetch as typeof fetch;
```

tests/ui/PaginationTest.tsx
```diff
@@ -0,0 +1,297 @@
...
+    return {
+        ...actualNav,
+        useNavigation,
+        getState: () => ({
+            routes: [],
+        }),
+    } as typeof NativeNavigation;
+});
+
+const fetchMock = TestHelper.getGlobalFetchMock();
```
fabioh8010, line 108:
```suggestion
const fetchMock = TestHelper.getGlobalFetchMock() as TestHelper.MockFetch;
```

roryabraham:
hmmm what else did we change here that's blowing up TS? It doesn't seem to be the TS bump or type-fest...

fabioh8010:
checking

src/libs/PaginationUtils.ts
roryabraham: updated

fabioh8010:
@roryabraham Let's keep TS as `"typescript": "^5.4.5",`, I see you reverted but actually it didn't make effect on package-lock.json, that is still using 5.4.5 somewhere inside the file.

About `Expression produces a union type that is too complex to represent.`, looks like if I remove all references to `REPORT_ACTIONS_PAGES` inside ONYXKEYS.ts the problem is solved 🤔 Still don't understand why..

roryabraham:
I wonder if we just hit a critical mass of Onyx keys that all of our OnyxKey union types are too big now 🤔

fabioh8010:
> I wonder if we just hit a critical mass of Onyx keys that all of our OnyxKey union types are too big now 🤔

Yeah I was thinking about this, I **really** hope that isn't the case here.

fabioh8010:
@roryabraham Looks like we indeed hit some critical point with `OnyxUpdate` type and the error will happen every time we use spread to combine two `OnyxUpdate[]` arrays.

Screenshot 2024-06-12 at 18 43 47

```ts
    const optimisticData: OnyxUpdate[] = [
        {
            key: `${ONYXKEYS.COLLECTION.REPORT}${reportID}`,
            value: {
                participants: participantsAfterInvitation,
                pendingChatMembers,
            },
        },
        ...newPersonalDetailsOnyxData.optimisticData, // this spread breaks the typings 💥
    ];
```

To workaround this we can use `push()` to combine our arrays.

```ts
    const optimisticData: OnyxUpdate[] = [
        {
            key: `${ONYXKEYS.COLLECTION.REPORT}${reportID}`,
            value: {
                participants: participantsAfterInvitation,
                pendingChatMembers,
            },
        },
    ];
    optimisticData.push(...newPersonalDetailsOnyxData.optimisticData); // OnyxUpdate is happy now ✨
```

Here's the diff to fix the files.

```diff
diff --git a/package-lock.json b/package-lock.json
index 6f626736d0..35c42f6d84 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -241,7 +241,7 @@
         "ts-jest": "^29.1.2",
         "ts-node": "^10.9.2",
         "type-fest": "4.20.0",
-        "typescript": "^5.3.2",
+        "typescript": "^5.4.5",
         "wait-port": "^0.2.9",
         "webpack": "^5.76.0",
         "webpack-bundle-analyzer": "^4.5.0",
diff --git a/package.json b/package.json
index ab98bb119a..7aada96ce8 100644
--- a/package.json
+++ b/package.json
@@ -293,7 +293,7 @@
     "ts-jest": "^29.1.2",
     "ts-node": "^10.9.2",
     "type-fest": "4.20.0",
-    "typescript": "^5.3.2",
+    "typescript": "^5.4.5",
     "wait-port": "^0.2.9",
     "webpack": "^5.76.0",
     "webpack-bundle-analyzer": "^4.5.0",
diff --git a/src/libs/actions/Policy/Member.ts b/src/libs/actions/Policy/Member.ts
index ee2f02b682..fa0cc1a43d 100644
--- a/src/libs/actions/Policy/Member.ts
+++ b/src/libs/actions/Policy/Member.ts
@@ -247,8 +247,8 @@ function removeMembers(accountIDs: number[], policyID: string) {
             key: policyKey,
             value: {employeeList: optimisticMembersState},
         },
-        ...announceRoomMembers.onyxOptimisticData,
     ];
+    optimisticData.push(...announceRoomMembers.onyxOptimisticData);
 
     const successData: OnyxUpdate[] = [
         {
@@ -256,8 +256,8 @@ function removeMembers(accountIDs: number[], policyID: string) {
             key: policyKey,
             value: {employeeList: successMembersState},
         },
-        ...announceRoomMembers.onyxSuccessData,
     ];
+    successData.push(...announceRoomMembers.onyxSuccessData);
 
     const failureData: OnyxUpdate[] = [
         {
@@ -265,8 +265,8 @@ function removeMembers(accountIDs: number[], policyID: string) {
             key: policyKey,
             value: {employeeList: failureMembersState},
         },
-        ...announceRoomMembers.onyxFailureData,
     ];
+    failureData.push(...announceRoomMembers.onyxFailureData);
 
     const pendingChatMembers = ReportUtils.getPendingChatMembers(accountIDs, [], CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE);
 
@@ -551,10 +551,8 @@ function addMembersToWorkspace(invitedEmailsToAccountIDs: InvitedEmailsToAccount
                 employeeList: optimisticMembersState,
             },
         },
-        ...newPersonalDetailsOnyxData.optimisticData,
-        ...membersChats.onyxOptimisticData,
-        ...announceRoomMembers.onyxOptimisticData,
     ];
+    optimisticData.push(...newPersonalDetailsOnyxData.optimisticData, ...membersChats.onyxOptimisticData, ...announceRoomMembers.onyxOptimisticData);
 
     const successData: OnyxUpdate[] = [
         {
@@ -564,10 +562,8 @@ function addMembersToWorkspace(invitedEmailsToAccountIDs: InvitedEmailsToAccount
                 employeeList: successMembersState,
             },
         },
-        ...newPersonalDetailsOnyxData.finallyData,
-        ...membersChats.onyxSuccessData,
-        ...announceRoomMembers.onyxSuccessData,
     ];
+    successData.push(...newPersonalDetailsOnyxData.finallyData, ...membersChats.onyxSuccessData, ...announceRoomMembers.onyxSuccessData);
 
     const failureData: OnyxUpdate[] = [
         {
@@ -580,9 +576,8 @@ function addMembersToWorkspace(invitedEmailsToAccountIDs: InvitedEmailsToAccount
                 employeeList: failureMembersState,
             },
         },
-        ...membersChats.onyxFailureData,
-        ...announceRoomMembers.onyxFailureData,
     ];
+    failureData.push(...membersChats.onyxFailureData, ...announceRoomMembers.onyxFailureData);
 
     const params: AddMembersToWorkspaceParams = {
         employees: JSON.stringify(logins.map((login) => ({email: login}))),
diff --git a/src/libs/actions/Policy/Policy.ts b/src/libs/actions/Policy/Policy.ts
index 8d72b0a5ac..ed0613adb6 100644
--- a/src/libs/actions/Policy/Policy.ts
+++ b/src/libs/actions/Policy/Policy.ts
@@ -2634,49 +2634,56 @@ function enablePolicyTaxes(policyID: string, enabled: boolean) {
     };
     const policy = getPolicy(policyID);
     const shouldAddDefaultTaxRatesData = (!policy?.taxRates || isEmptyObject(policy.taxRates)) && enabled;
-    const onyxData: OnyxData = {
-        optimisticData: [
-            {
-                onyxMethod: Onyx.METHOD.MERGE,
-                key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
-                value: {
-                    tax: {
-                        trackingEnabled: enabled,
-                    },
-                    pendingFields: {
-                        tax: CONST.RED_BRICK_ROAD_PENDING_ACTION.UPDATE,
-                    },
+
+    const optimisticData: OnyxUpdate[] = [
+        {
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+            value: {
+                tax: {
+                    trackingEnabled: enabled,
+                },
+                pendingFields: {
+                    tax: CONST.RED_BRICK_ROAD_PENDING_ACTION.UPDATE,
                 },
             },
-            ...(shouldAddDefaultTaxRatesData ? taxRatesData.optimisticData ?? [] : []),
-        ],
-        successData: [
-            {
-                onyxMethod: Onyx.METHOD.MERGE,
-                key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
-                value: {
-                    pendingFields: {
-                        tax: null,
-                    },
+        },
+    ];
+    optimisticData.push(...(shouldAddDefaultTaxRatesData ? taxRatesData.optimisticData ?? [] : []));
+
+    const successData: OnyxUpdate[] = [
+        {
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+            value: {
+                pendingFields: {
+                    tax: null,
                 },
             },
-            ...(shouldAddDefaultTaxRatesData ? taxRatesData.successData ?? [] : []),
-        ],
-        failureData: [
-            {
-                onyxMethod: Onyx.METHOD.MERGE,
-                key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
-                value: {
-                    tax: {
-                        trackingEnabled: !enabled,
-                    },
-                    pendingFields: {
-                        tax: null,
-                    },
+        },
+    ];
+    successData.push(...(shouldAddDefaultTaxRatesData ? taxRatesData.successData ?? [] : []));
+
+    const failureData: OnyxUpdate[] = [
+        {
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+            value: {
+                tax: {
+                    trackingEnabled: !enabled,
+                },
+                pendingFields: {
+                    tax: null,
                 },
             },
-            ...(shouldAddDefaultTaxRatesData ? taxRatesData.failureData ?? [] : []),
-        ],
+        },
+    ];
+    failureData.push(...(shouldAddDefaultTaxRatesData ? taxRatesData.failureData ?? [] : []));
+
+    const onyxData: OnyxData = {
+        optimisticData,
+        successData,
+        failureData,
     };
 
     const parameters: EnablePolicyTaxesParams = {policyID, enabled};
diff --git a/src/libs/actions/Report.ts b/src/libs/actions/Report.ts
index 83d931c4d0..1da0a91c87 100644
--- a/src/libs/actions/Report.ts
+++ b/src/libs/actions/Report.ts
@@ -2774,8 +2774,8 @@ function inviteToRoom(reportID: string, inviteeEmailsToAccountIDs: InvitedEmails
                 pendingChatMembers,
             },
         },
-        ...newPersonalDetailsOnyxData.optimisticData,
     ];
+    optimisticData.push(...newPersonalDetailsOnyxData.optimisticData);
 
     const successPendingChatMembers = report?.pendingChatMembers
         ? report?.pendingChatMembers?.filter(
@@ -2790,8 +2790,9 @@ function inviteToRoom(reportID: string, inviteeEmailsToAccountIDs: InvitedEmails
                 pendingChatMembers: successPendingChatMembers,
             },
         },
-        ...newPersonalDetailsOnyxData.finallyData,
     ];
+    successData.push(...newPersonalDetailsOnyxData.finallyData);
+
     const failureData: OnyxUpdate[] = [
         {
             onyxMethod: Onyx.METHOD.MERGE,
@@ -3407,15 +3408,16 @@ function completeOnboarding(
     ];
     const successData: OnyxUpdate[] = [
         ...tasksForSuccessData,
-        {
-            onyxMethod: Onyx.METHOD.MERGE,
-            key: `${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${targetChatReportID}`,
-            value: {
-                [introductionCommentAction.reportActionID]: {pendingAction: null},
-                [textCommentAction.reportActionID]: {pendingAction: null},
-            },
-        },
     ];
+    successData.push({
+        onyxMethod: Onyx.METHOD.MERGE,
+        key: `${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${targetChatReportID}`,
+        value: {
+            [introductionCommentAction.reportActionID]: {pendingAction: null},
+            [textCommentAction.reportActionID]: {pendingAction: null},
+        },
+    });
+
     let failureReport: Partial<Report> = {
         lastMessageTranslationKey: '',
         lastMessageText: '',

```

src/CONST.ts
janicduplessis: We currently do not set the END one since we don't have a reliable way to know that we reached the end. My plan was to use it when we implement pagination info returned from the backend.

janicduplessis: We could check the report type for CREATED, which is what we do in ReportScreen to avoid calling GetOlderActions more.

tests/utils/debug.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * The debug utility that ships with react native testing library does not work properly and
+ * has limited functionality. This is a better version of it that allows logging a subtree of
+ * the app.
+ */
+
+/* eslint-disable  no-console, testing-library/no-node-access, testing-library/no-debugging-utils, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-explicit-any */
+import type {NewPlugin} from 'pretty-format';
+import prettyFormat, {plugins} from 'pretty-format';
```
roryabraham, line 9: @janicduplessis does `pretty-format` need to be added as a dev dependency?

janicduplessis: It is already included as part of testing-library/react-native, but I added it here since we use it directly now.

If that is a problem I can remove this debug util and add it again in a follow up. It has been very useful for debugging the ui tests.

janicduplessis: Another option would be to try to upstream this back to testing-library so the `screen.debug` utility works properly.

roryabraham: I think this is probably ok for now, we've got bigger fish to fry imo

roryabraham:
@neil-marcellini we still want a C+ review here

roryabraham:
@ishpaul777 volunteered to help out 🙇🏼

roryabraham:
@janicduplessis there are a ton of errors like:

```
Warning: An update to LHNOptionsList inside a test was not wrapped in act(...).
```

I'm not sure if these are red herrings or not, but I think either way we should fix them if possible to make the logs from these tests less noisy, so the real problem(s) are clearer.

janicduplessis:
@roryabraham do you know if they are coming from the new pagination tests? I think there were some already and I tried to make sure there were none with the new tests. Either way we could also just fix them all.

roryabraham:
Removed unrelated storybook diff, moved to https://github.com/Expensify/App/pull/44405

roryabraham:
> Bug?: After linking to a comment, there is a loading indicator everytime after sending a message

This seems like a legit bug, but it's reproducible on main so at least it's not a regression

neil-marcellini:
👀 I'm going to start reviewing this now

neil-marcellini (commented):
Leaving a partial review for now. I've looked at the core of the pagination functionality and I'm happy with it. I'm most concerned about the ESNext thing.

tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "compilerOptions": {
         "module": "commonjs",
         "types": ["react-native", "jest"],
+        "lib": ["DOM", "DOM.Iterable", "ESNext"],
```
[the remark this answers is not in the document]
roryabraham: I don't have more context for why this was added or what it does. @janicduplessis do you have any more context?

roryabraham:
Extracted [another PR](https://github.com/Expensify/App/pull/44412) from this one to try and reduce the diff + testing surface of this PR

src/libs/Middleware/Pagination.ts
```diff
@@ -0,0 +1,132 @@
...
+import type {PaginatedRequest} from '@src/types/onyx/Request';
+import type Middleware from './types';
+
+type PaginationConfig<TResourceKey extends OnyxCollectionKey, TPageKey extends OnyxPagesKey> = {
+    initialCommand: ApiCommand;
+    previousCommand: ApiCommand;
+    nextCommand: ApiCommand;
+    resourceCollectionKey: TResourceKey;
+    pageCollectionKey: TPageKey;
+    sortItems: (items: OnyxValues[TResourceKey]) => Array<OnyxValues[TResourceKey] extends Record<string, infer TResource> ? TResource : never>;
```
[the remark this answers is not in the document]
roryabraham: Interesting suggestion. For now there aren't any resources that come presorted from the back-end AFAIK, so if/when we change that would be a better time to make it optional imo. I'm glad you're thinking about how this might be re-used for report pagination though.

tests/ui/PaginationTest.tsx
```diff
@@ -0,0 +1,348 @@
...
+        await waitForBatchedUpdatesWithAct();
+
+        TestHelper.expectAPICommandToHaveBeenCalled('OpenReport', 1);
+        TestHelper.expectAPICommandToHaveBeenCalled('GetOlderActions', 1);
+        TestHelper.expectAPICommandToHaveBeenCalledWith('GetOlderActions', 0, {reportID: REPORT_ID, reportActionID: '4'});
+        TestHelper.expectAPICommandToHaveBeenCalled('GetNewerActions', 0);
+
+        await waitForBatchedUpdatesWithAct();
+
+        // We now have 10 messages. 5 from the initial OpenReport and 3 from GetOlderActions.
```
neil-marcellini, line 307: 10 => 8

tests/ui/PaginationTest.tsx
```diff
@@ -0,0 +1,348 @@
...
+        TestHelper.expectAPICommandToHaveBeenCalled('GetOlderActions', 0);
+        TestHelper.expectAPICommandToHaveBeenCalled('GetNewerActions', 0);
+
+        scrollToOffset(0);
+        await waitForBatchedUpdatesWithAct();
+
+        TestHelper.expectAPICommandToHaveBeenCalled('OpenReport', 2);
+        TestHelper.expectAPICommandToHaveBeenCalled('GetOlderActions', 0);
+        TestHelper.expectAPICommandToHaveBeenCalled('GetNewerActions', 1);
+        TestHelper.expectAPICommandToHaveBeenCalledWith('GetNewerActions', 0, {reportID: REPORT_ID, reportActionID: '5'});
```
neil-marcellini, line 341: I want to make sure I'm understanding this. So we are opening a report by linking to an action within it. That calls OpenReport once for reportActionID: '5' and the mock returns 5 actions. Then we scroll to the bottom to load 5 new actions. Why is the same ID passed? Wouldn't the ID be 10?

__mocks__/@react-navigation/native/index.ts
```diff
@@ -1,9 +1,65 @@
-import {useIsFocused as realUseIsFocused, useTheme as realUseTheme} from '@react-navigation/native';
+/* eslint-disable import/prefer-default-export, import/no-import-module-exports */
+import type * as ReactNavigation from '@react-navigation/native';
+import createAddListenerMock from '../../../tests/utils/createAddListenerMock';
 
-// We only want these mocked for storybook, not jest
-const useIsFocused: typeof realUseIsFocused = process.env.NODE_ENV === 'test' ? realUseIsFocused : () => true;
+const isJestEnv = process.env.NODE_ENV === 'test';
```
neil-marcellini, line 5: When would this mock get used in an environment other than a jest test?

neil-marcellini (changes requested):
Wheew that was a lot! Amazing work @janicduplessis and @roryabraham! The testing is particularly impressive, and I think it's awesome that this is building more UI tests and support for them.

I'm requesting changes for my comments about the clientUpdateID and one of the pagination util tests. All the other stuff is NAB.

__mocks__/@react-navigation/native/index.ts
janicduplessis: I think it is the storybook tests, @roryabraham can you confirm?

src/libs/API/types.ts
```diff
@@ -5,7 +5,7 @@ import type * as Parameters from './parameters';
 import type SignInUserParams from './parameters/SignInUserParams';
 import type UpdateBeneficialOwnersForBankAccountParams from './parameters/UpdateBeneficialOwnersForBankAccountParams';
 
-type ApiRequest = ValueOf<typeof CONST.API_REQUEST_TYPE>;
+type ApiRequestType = ValueOf<typeof CONST.API_REQUEST_TYPE>;
```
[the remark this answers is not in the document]
janicduplessis: I think in the case the "Type" suffix is to mean the kind of request, not the type as in Typescript type.

tests/ui/PaginationTest.tsx
janicduplessis: This one is the forward pagination case so it is a little different.

- OpenReport call with id '5' will load ids ['5', '4', '3', '2', '1']
- Scroll to start
- GetNewerActions is called with id '5' and will load ['5', '6', '7', '8', '9', '10']

The ids are generated chronologically so newer messages always have bigger id.

When passing a reportActionID to OpenReport and GetOlderActions it returns that report and older reports (meaning ids will be lowering) and GetNewerActions it returns that report and newer report (meaning ids will be growing).

tsconfig.json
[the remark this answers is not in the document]
janicduplessis: This is to be able to use iterator functions on FormData. I am using it to get an object from FormData using `Object.fromEntries(someFormData)` [here](https://github.com/Expensify/App/pull/41962/files/86d4534675b7ea5763f40046721a45d15581814a#diff-6c77c0f2f7826218a23dbe3738bc93fb1b935b57375337f89ebd7780ddee76e7R276).

This is basically just adding DOM.Iterable. DOM and ESNext and already the values set by the preset we use.

tests/unit/PaginationUtilsTest.ts
```diff
@@ -0,0 +1,568 @@
...
+                ['17', '16', CONST.PAGINATION_END_ID],
+            ];
+
+            const expectedResult = createItems([
+                // Expect these sortedItems
+                '17',
+                '16',
+                '15',
+                '14',
+            ]);
```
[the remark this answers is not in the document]
janicduplessis: The end marker means "Include all actions after that page" so this is why 15 and 14 are also included.

src/libs/API/types.ts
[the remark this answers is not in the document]
neil-marcellini: Yeah I guess that's fair. Ok.

tests/unit/PaginationUtilsTest.ts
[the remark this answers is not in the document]
neil-marcellini: Ok, it sounds like I misunderstood the purpose of it. Thanks for clearing that up.

tests/ui/PaginationTest.tsx
neil-marcellini: Ah ok, now it makes sense. I had assumed opening a report at a given ID would return IDs after the given one. Thank you for explaining.

neil-marcellini (dismissed):
Everything looks good to me now, thank you 🚀

__mocks__/@react-navigation/native/index.ts
roryabraham: yep we use mocks with storybook too.

src/libs/API/index.ts
```diff
@@ -173,14 +195,57 @@ function makeRequestWithSideEffects<TCommand extends SideEffectRequestCommand |
...
+    onyxData: OnyxData,
+    config: PaginationConfig,
+): void;
+function paginate<TRequestType extends ApiRequestType, TCommand extends CommandOfType<TRequestType>>(
+    type: TRequestType,
+    command: TCommand,
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData: OnyxData,
+    config: PaginationConfig,
+): Promise<Response | void> | void {
```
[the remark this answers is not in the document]
roryabraham: This is called function overloading, and it ensures that `API.paginate` only returns a promise for `makeRequestWithSideEffects` requests, but is void for `read` and `write` requests.

src/libs/API/index.ts
```diff
@@ -55,55 +58,85 @@ Onyx.connect({
...
-            shouldRetry: true,
-            canCancel: true,
-            clientUpdateID: lastUpdateIDAppliedToClient,
-        },
+        data,
         ...onyxDataWithoutOptimisticData,
     };
 
+    if (isWriteRequest) {
+        request.data.clientUpdateID = lastUpdateIDAppliedToClient;
```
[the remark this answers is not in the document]
roryabraham: Good catch, I think that might've changed more recently and we missed it in a git merge.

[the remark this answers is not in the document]
roryabraham: updated 👍🏼

src/libs/API/index.ts
```diff
@@ -123,41 +156,30 @@ function write<TCommand extends WriteCommand>(command: TCommand, apiCommandParam
...
-    return Request.processWithMiddleware(request);
+    return processRequest(request, CONST.API_REQUEST_TYPE.MAKE_REQUEST_WITH_SIDE_EFFECTS);
+}
+
+/**
+ * Ensure all write requests on the sequential queue have finished responding before running read requests.
+ * Responses from read requests can overwrite the optimistic data inserted by
+ * write requests that use the same Onyx keys and haven't responded yet.
+ */
+function validateReadyToRead<TCommand extends ReadCommand>(command: TCommand) {
```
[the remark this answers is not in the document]
roryabraham: done 👍🏼

neil-marcellini (dismissed):
Your latest changes look good, thanks @roryabraham.

ishpaul777 (approved):

src/types/onyx/Request.ts
mountiny: +1 seems like we are missing the comment here

mountiny (dismissed):
Seems like this is ready to go, amazing job pushing this ahead, any concerns about merging this as is?

mountiny (approved):
Thanks

Julesssss:
Hey @janicduplessis and everyone subscribed to this pagination PR. I think this newly discovered bug is related to these changes. Our backend seems to report a value of -1 to be invalid. Is a value of -1 necessary for the `clientUpdateID` default?

janicduplessis:
@roryabraham I think you worked on this part, do you know why this happens now?

roryabraham:
@Julesssss we didn't change the logic with `clientUpdateID` in this PR. We just DRYed it up and moved it from being added to requests in multiple places in `src/libs/API/index.ts` to being added to requests in one place in `src/libs/Network/enhanceParameters.ts`

mountiny:
I believe @iwiznia @danieldoglas were doing something around this default/ logic

roryabraham:
Let's discuss in https://github.com/Expensify/App/issues/45252. I added some breadcrumbs [here](https://github.com/Expensify/App/issues/45252#issuecomment-2240436895)

copernicusjones:
I've created an Expensify account with the email jadkins333@yahoo.com. The bot was showing an invalid email error — please retry the payment verification.

Issue #43172, opened 2024-06-06
[HOLD for payment 2024-06-20] [HOLD for payment 2024-06-18] Thread - App shows not here page when clicking on blue "Thread" in thread view


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 1.4.80-1
**Reproducible in staging?:** Y
**Reproducible in production?:** N
**If this was caught during regression testing, add the test name, ID and link from TestRail:** N/A
**Issue reported by:** Applause - Internal Team

## Action Performed:
1. Go to staging.new.expensify.com
2. Go to workspace chat
3. Create a manual expense
4. Click on the expense preview to go to expense report
5. Send a message
6. Open the message in thread
7. Click on the blue "Thread" above the parent message

## Expected Result:
App will return to the main chat

## Actual Result:
App shows not here page

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [x] Android: Native
- [x] Android: mWeb Chrome
- [x] iOS: Native
- [x] iOS: mWeb Safari
- [x] MacOS: Chrome / Safari
- [x] MacOS: Desktop

## Screenshots/Videos

Add any screenshot/video evidence

</details> 

https://github.com/Expensify/App/assets/78819774/e4fdfe75-db88-48e0-b789-c16f09a5b847

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

lanitochka17:
@Julesssss  FYI I haven't added the External label as I wasn't 100% sure about this issue. Please take a look and add the label if you agree it's a bug and can be handled by external contributors

lanitochka17:
We think that this bug might be related to #wave-collect - Release 1

Julesssss:
I can reproduce. We seem to get stuck in a loop infinitely calling `OpenReport` and `GetOlderActions`

<img width="140" alt="Screenshot 2024-06-06 at 15 51 54" src="https://github.com/Expensify/App/assets/10736861/4ff66cac-ae94-4da2-b296-7df70f7383ad">

Julesssss:
[...]

PR #43200, opened 2024-06-06
Revert "Fix LHN shows Not found when navigated from global create menu or report parent navigation link"

luacmartins:
Reverts Expensify/App#43086


luacmartins:
I can confirm that reverting this PR does not solve [this blocker](https://github.com/Expensify/App/issues/43172)

PR #42330, opened 2024-05-17
Disabled distance rate is present in the rate list in confirmation p

cretadn22:
### Details
Defaulting distance rate to the first enabled when the Rate field is not shown (users with betas disabled).
Not displaying the disabled distance rates in the Rate field list (users with betas enabled).

### Fixed Issues

$ https://github.com/Expensify/App/issues/42284
PROPOSAL: https://github.com/Expensify/App/issues/42284#issuecomment-2116574341

### Tests

- [x] Verify that no errors appear in the JS console

### Test 1
Precondition: User does NOT have betas enabled, hence no Rate field will be displayed in Distance confirmation page.

1. Log in and go to Collect workspace settings.
2. Enable and go to Distance rates.
3. Click [+ Add rate].
4. Add a new rate and make sure it's Enabled.
5. Disable previously existing rate (default $0.670 / mile).
6. Go to workspace chat.
7. Start submitting a Distance request.
8. On Distance confirmation page, verify that the enabled rate is set as default in the Distance field (see below).

Screenshot 2024-05-29 at 14 10 59

### Test 2
Precondition: User does has betas enabled, hence the Rate field will be displayed in Distance confirmation page.

1. Log in and go to Collect workspace settings.
2. Enable and go to Distance rates.
3. Click [+ Add rate].
4. Add a new rate and make sure it's Enabled.
5. Disable previously existing rate (default $0.670 / mile).
6. Go to workspace chat.
7. Start submitting a Distance request.
8. On Distance confirmation page, verify that the enabled rate is set as default in the Rate field (see [1.]).
9. On Distance confirmation page, click on the Rate field and verify that the disabled rate is not displayed (see [2.]).

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
[...]

PR #43208, opened 2024-06-06
Revert "Disabled distance rate is present in the rate list in confirmation p"

neil-marcellini:
Reverts Expensify/App#42330


neil-marcellini:
Fixed here instead, so closing https://github.com/Expensify/App/pull/43217

PR #44768, opened 2024-07-03
Revert "Fix deep linking to reports after signing in"

mountiny:
Reverts Expensify/App#43307


mountiny:
Revert does not fix it solely

Issue #45027, opened 2024-07-08
[HOLD for payment 2024-08-05] [Search v2.1] [App] Use new query syntax


luacmartins:
Implement [this section](https://docs.google.com/document/d/1bRV1nPz8SaI_6DHqfehmLnADxv40us_q0JNvmZ1QyaY/edit#bookmark=id.888saqncimmk) of the design doc.

Holding on:
- [x] https://github.com/Expensify/App/issues/45025

cc @adamgrzybowski @Kicu @WojtekBoman

<details><summary>Issue Owner</summary>Current Issue Owner: @adelekennedy</details>

adelekennedy:
still on hold here

luacmartins:
Ready to be worked on!

adamgrzybowski:
@luacmartins Looks like something for me!

adamgrzybowski:
I'm actively working on this one

adamgrzybowski:
The draft is here https://github.com/Expensify/App/pull/45617. I hope to finish it tomorrow

adelekennedy:
@rayane-djouah sent you an offer on Upwork! https://www.upwork.com/nx/enterprise-homepage/?offerRid=103425705&nav_dir=replace&teamUid=all_teams&tab=hire

PR #45409, opened 2024-07-15
[Search v2] [App] Create search results page

WojtekBoman:
### Details

This PR adjusts the header of the Search page to display the custom query. The header displays the user's query when the route param `isCustomQuery` is set to `true`. This PR covers also adding the `Filters` button to the header.
### Fixed Issues

$ https://github.com/Expensify/App/issues/45028
PROPOSAL:

### Tests

- [ ] Verify that no errors appear in the JS console

1. Navigate to the url with the `isCustomQuery` route param set to true, for example: https://dev.new.expensify.com:8082/search?q=type%3Aexpense%20status%3Aall&isCustomQuery=true
2. Verify if the header is displayed correctly

### QA Steps

- [ ] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
[...]

PR #45617, opened 2024-07-17
[Search v2] [App] Use new query syntax

adamgrzybowski:
### Details

This PR introduces new query syntax. Now you can see that in the URL we have `q` or `cq` params when on the search page.  

### Fixed Issues

$ https://github.com/Expensify/App/issues/45027
PROPOSAL: 

### Tests

### QA Steps

Changing status option
1. Open the search tab
2. Change status from expenses to shared
3. See if the app loads shared entries

Changing workspace
1. Open the search tab
2. Open workspace switcher
3. Select different workspace
4. See if the entries are filtered for selected workspace

Wide layout only (web/desktop)
1. Open the search tab
2. Make sure the expenses option is selected
3. Press the date column to change sorting options
4. See if entries are properly sorted

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/78a5211a-5549-4df3-b49d-6eadefcd682e

</details>

<details>

https://github.com/user-attachments/assets/f4b78e30-8fe8-4727-9361-ba09c5f119b7

</details>

<details>

https://github.com/user-attachments/assets/8cb10ac6-92a2-4738-9911-71a2cb35d3aa

</details>

<details>

https://github.com/user-attachments/assets/3d6a2e4f-4a1e-497d-a64e-7241c3e52378

</details>

<details>

https://github.com/user-attachments/assets/de2d7056-bc14-49e3-a7ee-e2357c9c63ce

</details>

<details>

https://github.com/user-attachments/assets/80186b3f-8b54-4581-809f-8d66045c7045

</details>


src/CONST.ts
```diff
@@ -5263,6 +5279,11 @@ const CONST = {
             REPORT_ID: 'reportID',
             KEYWORD: 'keyword',
         },
+        QUERY_KIND: {
+            CANNED_QUERY: 'cannedQuery',
+            CUSTOM_QUERY: 'customQuery',
+        },
+        DEFAULT_QUERY: 'type:all sortBy:date sortOrder:desc',
```
luacmartins, line 5286:
```suggestion
        DEFAULT_QUERY: 'type:expense status:all sortBy:date sortOrder:desc offset:0',
```

luacmartins: That being said, I think the parser returns the default case already

src/CONST.ts
```diff
@@ -5263,6 +5279,11 @@ const CONST = {
             REPORT_ID: 'reportID',
             KEYWORD: 'keyword',
         },
+        QUERY_KIND: {
```
luacmartins, line 5282:
```suggestion
        QUERY_TYPE: {
```

src/ROUTES.ts
```diff
@@ -37,16 +36,9 @@ const ROUTES = {
...
-
-            if (!sortBy && !sortOrder) {
-                return `search/${searchQuery}` as const;
-            }
-
-            return `search/${searchQuery}?sortBy=${sortBy}&sortOrder=${sortOrder}` as const;
-        },
+        route: '/search',
+        getRoute: ({query, queryKind = CONST.SEARCH.QUERY_KIND.CANNED_QUERY, policyIDs}: {query: SearchQueryString; queryKind?: QueryKind; policyIDs?: string}) =>
+            `search?${queryKind === CONST.SEARCH.QUERY_KIND.CANNED_QUERY ? 'q' : 'cq'}=${query}${policyIDs ? `&policyIDs=${policyIDs}` : ''}` as const,
```
luacmartins, line 41: I think the doc had [two separate routes](https://docs.google.com/document/d/1bRV1nPz8SaI_6DHqfehmLnADxv40us_q0JNvmZ1QyaY/edit#bookmark=id.888saqncimmk), any reason to combine them?

luacmartins (commented):
I know this is a draft but I wanted to leave some early comments

src/CONST.ts
adamgrzybowski: Yeah good catch 😄 I've removed it

src/CONST.ts
adamgrzybowski: This is a controversial topic. The problem is that we already have `query type`. We also have `search type`. I would love to see something other than `type` in the context of search because it may be confusing for people. It doesn't have to be `kind`. Just something else.

Kicu: I agree with the "controversy" 😅 there is a `type` within the query already, part of filters. it's the type of report. And I can image people will call it `query type` just like they would might say `query status`.
Then we have the `type` of search results, which can be `transaction` or `report`.

It would nice to not name this thing also `type`, future devs will get confused.

adamgrzybowski:
@luacmartins Ready for the review! I will add videos and tests on Monday

src/pages/Search/SearchPageBottomTab.tsx
```diff
@@ -78,10 +68,9 @@ function SearchPageBottomTab() {
                 )}
                 {isSmallScreenWidth && (
                     <Search
+                        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                        queryJSON={queryJSON!}
```
Kicu, line 72: Instead of using the `!` operator, just add check of `queryJSON` to line 69 as part of the check for displaying this block

Kicu (commented):
after initial review seems pretty nice, will take another look later

src/components/Search/index.tsx
```diff
@@ -144,15 +145,27 @@ function Search({query, policyIDs, sortBy, sortOrder, isMobileSelectionModeActiv
...
         if (!searchResults?.search?.hasMoreResults || shouldShowLoadingState || shouldShowLoadingMoreItems) {
             return;
         }
-        const currentOffset = searchResults?.search?.offset ?? 0;
-        SearchActions.search({hash, query, offset: currentOffset + CONST.SEARCH.RESULTS_PAGE_SIZE, sortBy, sortOrder});
+
+        const currentSearchParams = SearchUtils.getCurrentSearchParams();
+        const currentQueryString = SearchUtils.getQueryStringFromParams(currentSearchParams);
+        const isCustomQuery = SearchUtils.isCustomQueryFromParams(currentSearchParams);
+        const currentQueryJSON = SearchUtils.buildSearchQueryJSON(currentQueryString);
```
adamgrzybowski, line 159: The more I think about `cq` and `q`, the more I think this is not a good solution. We need additional code to work with queries saved in params. Maybe we should have one `q` parameter and another parameter telling us if this query is canne or custom? This should be easier to manage. What do you think guys? @luacmartins @Kicu @WojtekBoman

src/pages/Search/SearchPageBottomTab.tsx
```diff
@@ -9,46 +8,36 @@ import useLocalize from '@hooks/useLocalize';
...
-import type {CentralPaneScreensParamList} from '@libs/Navigation/types';
+import type {AuthScreensParamList} from '@libs/Navigation/types';
+import {buildSearchQueryJSON, getQueryStringFromParams} from '@libs/SearchUtils';
 import TopBar from '@navigation/AppNavigator/createCustomBottomTabNavigator/TopBar';
 import CONST from '@src/CONST';
 import ROUTES from '@src/ROUTES';
 import SCREENS from '@src/SCREENS';
-import type {SearchQuery} from '@src/types/onyx/SearchResults';
-import SearchFilters from './SearchFilters';
+import SearchStatuses from './SearchStatusMenu';
```
WojtekBoman, line 17: Replace `SearchStatuses` with `SearchStatusMenu`

src/pages/Search/SearchStatusMenu.tsx
```diff
@@ -1,64 +1,65 @@
...
 import Navigation from '@libs/Navigation/Navigation';
+import {normalizeQuery} from '@libs/SearchUtils';
 import variables from '@styles/variables';
 import * as Expensicons from '@src/components/Icon/Expensicons';
 import CONST from '@src/CONST';
 import type {Route} from '@src/ROUTES';
 import ROUTES from '@src/ROUTES';
 import type IconAsset from '@src/types/utils/IconAsset';
-import SearchFiltersNarrow from './SearchFiltersNarrow';
+import SearchFiltersNarrow from './SearchStatusMenuNarrow';
```
WojtekBoman, line 17: Replace `SearchFiltersNarrow` with `SearchStatusMenuNarrow`

adamgrzybowski: 🏹 🦅 👁️ 🙇

src/ROUTES.ts
luacmartins: I wonder if instead of updating `SearchPage` to also handle the Search Results case, we should just create two components for these pages, e.g. `SearchPage` and `SearchResultsPage`

src/components/Search/index.tsx
luacmartins: I'm not sure that I understand the issue. What do you mean by ` We need additional code to work with queries saved in params.`?

luacmartins: Ah do you mean the methods above, e.g. `getCurrentSearchParams`, `getQueryStringFromParams` and `getQueryStringFromParams`?

luacmartins: I made a similar comment, but I wonder if we should just create a separate page `SearchResultsPage` and pass a `isCustomQuery=true` prop to `Search/index.tsx` so we can avoid all this logic and the `QUERY_KIND` stuff we have above.

src/libs/Navigation/switchPolicyID.ts
```diff
@@ -84,7 +84,7 @@ export default function switchPolicyID(navigation: NavigationContainerRef<RootSt
     // Here's the configuration: src/libs/Navigation/AppNavigator/createCustomStackNavigator/index.tsx
     const isOpeningSearchFromBottomTab = !route && topmostCentralPaneRoute?.name === SCREENS.SEARCH.CENTRAL_PANE;
     if (isOpeningSearchFromBottomTab) {
-        newPath = ROUTES.SEARCH_CENTRAL_PANE.getRoute(CONST.SEARCH.TAB.ALL);
+        newPath = ROUTES.SEARCH_CENTRAL_PANE.getRoute({query: CONST.SEARCH.TAB.EXPENSE.ALL});
```
luacmartins, line 87: I think we should append the policyID as a param to the query string so that it's included in the filters list AST. I think the current grammar is not processing that filter, but we should update that as well.

src/libs/SearchUtils.ts
```diff
@@ -299,34 +300,32 @@ function getSortedReportData(data: ReportListItemType[]) {
...
-    type: string;
-    status: string;
-    sortBy: string;
-    sortOrder: string;
-    offset: number;
-    filters: ASTNode;
-};
-
-function buildJSONQuery(query: string) {
+function buildSearchQueryJSON(query: SearchQueryString) {
```
luacmartins, line 325: Why did we rename this function?

src/libs/SearchUtils.ts
```diff
@@ -335,10 +334,33 @@ function buildJSONQuery(query: string) {
     }
 }
 
-function getFilters(query: string, fields: Array<Partial<AllFieldKeys>>) {
-    let jsonQuery;
+function buildSearchQueryString(partialQueryJSON?: Partial<SearchQueryJSON>) {
+    const queryParts: string[] = [];
+    const defualtQueryJSON = buildSearchQueryJSON('');
```
luacmartins, line 339:
```suggestion
    const defaultQueryJSON = buildSearchQueryJSON('');
```

src/libs/SearchUtils.ts
```diff
@@ -335,10 +334,33 @@ function buildJSONQuery(query: string) {
...
+    }
+
+    return queryParts.join(' ');
+}
+
+// Fill query string with all default params.
+function normalizeQuery(query: string) {
+    const normalizedQueryJSON = buildSearchQueryJSON(query);
+    return buildSearchQueryString(normalizedQueryJSON);
+}
```
luacmartins, line 357: Why do we need this function? I think that it could inject params into the query string that were not defined by the user and that might mess up with how we display the filter UI, no?

src/types/onyx/SearchResults.ts
```diff
@@ -64,6 +64,17 @@ type SearchResultsInfo = {
...
     columnsToShow: ColumnsToShow;
+
+    /** The status of the search results to show */
+    statusToShow: {
+        // eslint-disable-next-line jsdoc/require-jsdoc
+        expense: {
+            // eslint-disable-next-line jsdoc/require-jsdoc
+            drafts: boolean;
+            // eslint-disable-next-line jsdoc/require-jsdoc
+            approved: boolean;
```
luacmartins, line 75: We actually return all statuses to the client, e.g. `all`, `drafts`, `outstanding`, `approved`, `paid`

src/pages/Search/SearchStatusMenuNarrow.tsx
```diff
@@ -10,14 +10,14 @@ import useThemeStyles from '@hooks/useThemeStyles';
 import useWindowDimensions from '@hooks/useWindowDimensions';
 import Navigation from '@libs/Navigation/Navigation';
 import * as Expensicons from '@src/components/Icon/Expensicons';
-import type {SearchMenuFilterItem} from './SearchFilters';
+import type {SearchStatusMenuItem} from './SearchStatusMenu';
 
-type SearchFiltersNarrowProps = {
-    filterItems: SearchMenuFilterItem[];
+type SearchStatusMenuNarrowProps = {
+    filterItems: SearchStatusMenuItem[];
```
luacmartins, line 16:
```suggestion
    statusItems: SearchStatusMenuItem[];
```

luacmartins (changes requested):

luacmartins (commented):
It seems like loading more results is broken. We don't pass the correct `offset` param to the API call even thought it's updated in the URL

src/ROUTES.ts
Kicu: ☝️ yes that is what @WojtekBoman is doing in his PR, Im assuming this code is temporary.

src/CONST.ts
adamgrzybowski: I like this idea! Let's go with `isCustomQuery`

src/ROUTES.ts
```diff
@@ -37,16 +36,9 @@ const ROUTES = {
...
-
-            if (!sortBy && !sortOrder) {
-                return `search/${searchQuery}` as const;
-            }
-
-            return `search/${searchQuery}?sortBy=${sortBy}&sortOrder=${sortOrder}` as const;
-        },
+        route: 'search',
+        getRoute: ({query, queryKind = CONST.SEARCH.QUERY_KIND.CANNED_QUERY, policyIDs}: {query: SearchQueryString; queryKind?: QueryKind; policyIDs?: string}) =>
+            `search?${queryKind === CONST.SEARCH.QUERY_KIND.CANNED_QUERY ? 'q' : 'cq'}=${query}${policyIDs ? `&policyIDs=${policyIDs}` : ''}` as const,
```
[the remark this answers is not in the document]
adamgrzybowski: I think this is a transitional step.

BTW will current grammar handle `policyID`? I don't see this keyword in the `.peggy` file

src/components/Search/index.tsx
```diff
@@ -144,15 +145,27 @@ function Search({query, policyIDs, sortBy, sortOrder, isMobileSelectionModeActiv
...
+
+        if (!currentQueryJSON) {
+            return;
+        }
+
+        // TODO_SEARCH: offset should be a number but it is a string.
```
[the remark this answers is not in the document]
adamgrzybowski: 👌  In that case I will modify the `SearchQueryAST` to use `string` for `offset` to avoid confusion.

src/libs/SearchUtils.ts
adamgrzybowski: There is a story to this change. It's not a strong argument but I believe it makes sense 😄

If the function is named `buildJSONQuery` then the result will look like that:

```ts
const JSONQuery = buildJSONQuery()
```
 right?

But then how we should name the return type of `builJSONQuery` function? it should start with the capital letter so `JSONQuery` but it is taken.

We could name the result differently so 

```ts
const jsonQuery = buildJSONQuery()
```

But it felt weird to me to change the whole acronym to small or big caps depending on the context.

And what if we want to add prefix? e.g. `current?`

It should be:
`currentJSONQuery` OR `currentJsonQuery`?

Then I realized that if we can have `queryString` then `queryJSON` and `queryAST` also should work for a convention. 

So finally I chose the easy way which is avoiding acronym at the beginning of the name 😄

src/libs/SearchUtils.ts
adamgrzybowski: This is something we could discuss more about. We have a couple of questions here:
1. Do we want to put defaults for sort etc in the URL? (I think yes, they were here in the previous version). 
2. Do we want to `organize` the query for the user? So is the:
`type:expenses sortBy:date` same query as `sortBy:date type:expenses` I would say yes. In that case we should sort the keywords in the query to make it look the same.

And to address your concerns. It will inject defaults but nothing more. It shouldn't break UI filters.

Another argument for normalizing but it will be relevant in the future. If we ever want to change the defaults it could break the saved queries if we don't normalize them.

src/types/onyx/SearchResults.ts
adamgrzybowski: So should I add all the fields? I am not sure if I understand this part of the code

src/ROUTES.ts
WojtekBoman: If I understand correctly, the only difference between SearchPage and SearchResultsPage is the header on these pages. To me it's easier to display this header conditionally based on `SearchUtils.isCustomQuery` rather than creating two different pages and maintaining them separately. If we want to split this page into `SearchPage` and `SearchResultsPage`, we'll have to repeat huge amount of code only to display a different header. Maybe we can use the same page for now, and when these pages begin to differ significantly, we'll separate them. What do you think @luacmartins?

luacmartins: That was the initial idea and what we have in the doc now. It seems like we started running into some issues with the `isCustomQuery` method, so that's why I suggested separating those screens. I'm cool with keeping them the same page for now. @adamgrzybowski what do you think?

adamgrzybowski:
Hi @luacmartins. @Kicu and I concluded that we should remove the `offset` property from the query.

Hash is generated from the query string. Offset is part of the query string. This means changing the offset will change the hash. This breaks caching and loading new elements.

Deep linking with URL that contains offset also won't load elements properly. We can't simply jump to offset 150. We need to go through 0, 50, 100, 150 to load all elements.

Instead, we could just keep it in the state in the Search component.

src/ROUTES.ts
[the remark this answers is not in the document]
luacmartins: Ah makes sense that we do it like this for now. We should eventually move it to the AST and include it in the peggy grammar too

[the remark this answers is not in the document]
luacmartins: Do we need a separate issue to migrate the `policyID` param to the AST?

src/libs/SearchUtils.ts
luacmartins: ahah I don't see an issue with the naming, but also this is not a big deal. Let's keep it as you have here already

src/libs/SearchUtils.ts
luacmartins: >Do we want to put defaults for sort etc in the URL? (I think yes, they were here in the previous version).

Thinking a bit ahead, we'd only want to do that if the user explicitly types the defaults in their queries, otherwise it'd be weird for the user to type `amount>200` and then see the search bar update to `type:expense status:all amount>200 sortBy:date sortOrder:desc offset:0`. I don't think we should modify the user input in that way and the URL is kinda the only thing tying everything up, e.g. the user input, the UI filters and the syntax itself. I think we'd only want to normalize it/build the AST when hashing and sending the API request

luacmartins: This is NAB for now, but something that we should align on before we get too deep in it

src/types/onyx/SearchResults.ts
luacmartins: Yea, I think we should add all of them for now since they are always returned by the server

src/libs/actions/Search.ts
```diff
@@ -53,6 +55,22 @@ function search({hash, query, policyIDs, offset, sortBy, sortOrder}: SearchParam
     API.read(READ_COMMANDS.SEARCH, {hash, query, offset, policyIDs, sortBy, sortOrder}, {optimisticData, finallyData});
 }
 
+// TODO_SEARCH: use this function after backend changes.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function searchV2(queryString: SearchQueryString) {
+    const queryJSON = buildSearchQueryJSON(queryString);
```
luacmartins, line 61: It seems like we're already passing the AST here, so no need to try to build it again
```suggestion
```

src/libs/actions/Search.ts
```diff
@@ -53,6 +55,22 @@ function search({hash, query, policyIDs, offset, sortBy, sortOrder}: SearchParam
...
+
+    if (!queryJSON) {
+        return;
+    }
+
+    const {optimisticData, finallyData} = getOnyxLoadingData(queryJSON.hash);
+
+    // TODO_SEARCH: uncomment this line after backend changes
+    // @ts-expect-error waiting for backend changes
+    API.read(READ_COMMANDS.SEARCH, queryJSON, {optimisticData, finallyData});
```
luacmartins, line 71:
```suggestion
    API.read(READ_COMMANDS.SEARCH, {jsonQuery: JSON.stringify(jsonQuery)}, {optimisticData, finallyData});
```

src/libs/actions/Search.ts
```diff
@@ -53,6 +55,22 @@ function search({hash, query, policyIDs, offset, sortBy, sortOrder}: SearchParam
     API.read(READ_COMMANDS.SEARCH, {hash, query, offset, policyIDs, sortBy, sortOrder}, {optimisticData, finallyData});
 }
 
+// TODO_SEARCH: use this function after backend changes.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function searchV2(queryString: SearchQueryString) {
```
luacmartins, line 60:
```suggestion
function searchV2(jsonQuery: SearchQueryJSON) {
```

luacmartins:
>Hi @luacmartins. @Kicu and I concluded that we should remove the offset property from the query.
>
>Hash is generated from the query string. Offset is part of the query string. This means changing the offset will change the hash. This breaks caching and loading new elements.

Yea, I think it makes sense to remove offset from the query string, the user would never set that themselves. Can we just include it in `buildSearchQueryJSON`, like we do for `hash` and `input`?

Kicu:
@luacmartins Im not sure we even need it in the `buildSearchQueryJSON` - up until now offset was local to `Search` component.
We're now switching how we store the params/filters in the URL but we don't really change the Search logic per se. That is it's still:
1. take some params from URL
2. (now also parse them)
3. generate hash
4. fetch results using hash whenever it changes

So I think `offset` could still stay as a local variable to `Search` component maybe?

src/ROUTES.ts
adamgrzybowski: For our current requirements, I think one page is perfect. I already remove `cq` so now the params are 
`q`, `isCustomQuery` and optional `policyIDs`

src/ROUTES.ts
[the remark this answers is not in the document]
adamgrzybowski: Do you think it should be in the root of `QueryJSON`? So something like 
```ts
type JSONQuery = {
    input: string;
    hash: number;
    type: string;
    status: string;
    sortBy: string;
    sortOrder: string;
    offset: number;
    filters: ASTNode;
    policyID: string;
};
```

Also currently it is called policyIDs, as the assumption was that we wanted to be able to use multiple policyIDs in the future. Is this still the case?

src/types/onyx/SearchResults.ts
adamgrzybowski: Do you mean these statuses? I don't see any `outstanding` in the costs.
```js
STATUS: {
  ALL: 'all',
  SHARED: 'shared',
  DRAFTS: 'drafts',
  FINISHED: 'finished',
},
```

adamgrzybowski: Not sure where to look for these values 😄

src/libs/actions/Search.ts
adamgrzybowski: Not sure if I understand. It looks like we pass queryString here not queryJSON.

src/libs/SearchUtils.ts
Kicu: I agree with Adam and I feel the code will make much more sense if `query` is a single source of truth and thus always contains all the values that are actually used when sending query to api.
I don't think we should worry how the URL looks for the user - it looks ugly anyways since it's url-encoded 😅 

That being said this PR is blocking some other work, so I'm a fan of leaving this conversation for later and moving forward with this.

adamgrzybowski:
@luacmartins I removed offset from the query and moved it to simple `useState`. It seems to work well now.

src/libs/actions/Search.ts
luacmartins: Hmm in my tests I had to remove this to make it work with the backend.

luacmartins:
>So I think offset could still stay as a local variable to Search component maybe?

We still need to send it to the API and the only param we'll send is the jsonQuery so we need to include it in the jsonQuery at some point.

src/libs/SearchParser/searchParser.peggy
```diff
@@ -101,7 +100,6 @@ key
   / "keyword" { return "keyword"; }
   / "sortBy" { return "sortBy"; }
   / "sortOrder" { return "sortOrder"; }
-  / "offset" { return "offset"; }
```
luacmartins, line 104: We also need to upload the new parser file, no? Or did we get automation to work?

luacmartins (commented):
Left a couple more comments

src/ROUTES.ts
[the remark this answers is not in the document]
luacmartins: We'll add it to filters as `policyID`. So the AST will look like this:

```
{filters: {
    "operator": "eq",
    "left": "policyID",
    "right": "<policyID>"
```

I think we should use `policyID` singular from now on since we have a bunch of other filters, e.g. `category`, `tag`, etc that accepts multiple values but are all singular too.

src/libs/SearchUtils.ts
luacmartins: I think the concern here is that we were gonna use the decoded URL to populate the search bar (once that's available) with the value the user input, but I see now that any interaction with the UI will also cause this issue. From the example above:

1. User types simple query like `type:expenses`
2. Press the date column name to change the sorting order.
3. Now the query looks like `type:expenses sortOrder:asc sortBy:date`
4. We'd display `type:expenses sortOrder:asc sortBy:date` in the Search bar, which is different than what the user typed
5. The user presses the column again and in theory we have same result as in the step 1 but the query looks differently `type:expenses sortOrder:desc sortBy:date`, which is also not what the user input was.

I think we might need another solution for this problem, so in all cases we only show `type:expenses` to the user since that's what they typed into the search bar. That's a problem for v2.3, but we should keep it in mind.

src/types/onyx/SearchResults.ts
luacmartins: They are [here](https://docs.google.com/document/d/1bRV1nPz8SaI_6DHqfehmLnADxv40us_q0JNvmZ1QyaY/edit#bookmark=id.d4v3gb8qe7kn). We should still support `Shared` for now, but we'll deprecate that one soon and add the other statuses mentioned in the doc.

rayane-d (commented):
Bug 1: Changing the workspace does not filter the entries

| Staging and main    | This PR |
|  <video src="https://github.com/user-attachments/assets/637c8787-de65-4410-a748-5465396232a7">  |  <video src="https://github.com/user-attachments/assets/a61da218-d0e6-4e6f-8b19-7a910ed3b91a">   |

src/libs/actions/Search.ts
luacmartins, line 60: FYI @Kicu @WojtekBoman the API should be ready to accept the new query syntax. Let's update this!!

luacmartins:
@rayane-djouah can you reproduce this on main too? https://github.com/Expensify/App/pull/45617#issuecomment-2248839757

luacmartins:
>Not reproducible on main

Thanks for confirming!


>1. The design doc requires implementing SearchUtils.isCustomQuery, but we still need to implement it, isn't it?

I think the solution we're going with differs a bit from the doc. I'll have to update the doc.

>Shouldn't we update filterItems in SearchFilters and headerContent in SearchPageHeader to show types (Expenses, Invoices, Trips) instead of statuses (Shared, Drafts, Finished) as the doc is suggesting? (and rename SearchFilters and SearchFiltersNarrow to SearchTypes and SearchTypesNarrow)

This was de-scoped from v2.1 and is now being handled in v2.2

src/ROUTES.ts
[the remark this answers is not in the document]
Kicu: I hope I understand this correctly. I will add `policyID` to grammar and keep it inside AST, while removing it from URL

Kicu:
@rayane-djouah is the Bug 2 related to changes in this PR? because HOLD and bulk is not something that should be modified here

Kicu:
Ah yes might be, thanks I will test it.

Meanwhile for bug 1 I think the problem might be that `policyIDs` needs to be part of `hash` for a single search query, because hash becomes part of onyx key.
So for your reproduction what (I believe, not sure 100%) happens is:
 - we filter a specific query by policyIDs, we get correct data
 - we change the workspaces from A to B
 - we get new filtered workspaces (as I made sure the correct call is made to backend)
 - we get the same `hash` because policyIDs is not part of hash
 - we display all the same results which will be mixed results from 2 calls

Good catch.  But in general I now have na "existential" 😅  problem with this PR. I expected that I will help fix several comments and we can finally close this.
However @luacmartins the switch for `policyIDs` that you suggested - although it makes sense - is a really big change. 
`policyIDs` is strictly connected with workspace Switcher and with all the other routes that use `/w/<policyID>/....` it also is connected with export to CSV.
Screenshot 2024-07-25 at 15 21 51

If we'd have to now remove it from params and move it to inside AST then that's editing 15 files and possibly breaking some functionality. IMO that will delay this PR for at least 2 days.
Also I actually have no idea how the new `policyID` param inside AST would work with existing workspace switcher? If it should just seamless work then we'd need to parse query in multiple places to be able extract this param, for flows like: 
1. go to reports page
2. switch workspace
3. go to search
4. should alredy have `search?q=policyID:xxxxx` in this scenario?

To sum up here are my questions about how to move forward
Q1 can we move forward with this PR keeping policyIDs as is? this can lead to some bugs but we can fix it in separate PRs
Q2 how will `policyIDs` interact with existing workspace switcher component
Q3 how are we switching to a new backend? will current Search simply stop working at some point and we have to switch sending search calls with jsonQuery? do I have to do it inside this PR?

src/libs/SearchParser/searchParser.peggy
Kicu: I believe we need to update it, will do that but first we need to decided what to do with policyIDs

src/types/onyx/SearchResults.ts
Kicu: why do we have to add this in this PR? I don't understand why this suddenly appeared here, when I don't see this being used anywhere.

Can I just remove it if it's unused and we can add it in the PR that this is actually being used?

src/ROUTES.ts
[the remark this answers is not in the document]
Kicu: read my comment here: https://github.com/Expensify/App/pull/45617#issuecomment-2250322113

src/ROUTES.ts
```diff
@@ -37,16 +36,9 @@ const ROUTES = {
...
-            const {sortBy, sortOrder} = queryParams ?? {};
-
-            if (!sortBy && !sortOrder) {
-                return `search/${searchQuery}` as const;
-            }
-
-            return `search/${searchQuery}?sortBy=${sortBy}&sortOrder=${sortOrder}` as const;
-        },
+        route: 'search',
+        getRoute: ({query, isCustomQuery = false, policyIDs}: {query: SearchQueryString; isCustomQuery?: boolean; policyIDs?: string}) =>
```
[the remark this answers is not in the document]
Kicu: Adam is back on Tuesday next week, and I simply don't know why this is needed, so we need to wait for him to add the comment.
Perhaps we could merge without it, dunno yet I will ask @luacmartins

Kicu:
@rayane-djouah my latest commit fixes Bug 2 via optional params.
Bug 1 would be fixed automatically if we would put `policyID` to hash generation, but Im not sure if we want to do this in this PR (see my previous long comment).

src/ROUTES.ts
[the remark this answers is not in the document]
luacmartins: This is needed for filters since selecting any filter will navigate the user to the Search Results page that's being implemented [here](https://github.com/Expensify/App/pull/45409). I know it's a bit out of place now, but we should just leave this in place.

src/pages/settings/Subscription/CardSection/CardSection.tsx
```diff
@@ -148,7 +148,7 @@ function CardSection() {
                         title={translate('subscription.cardSection.viewPaymentHistory')}
                         titleStyle={styles.textStrong}
                         style={styles.mt5}
-                        onPress={() => Navigation.navigate(ROUTES.SEARCH_CENTRAL_PANE.getRoute(CONST.SEARCH.TAB.ALL))}
+                        onPress={() => Navigation.navigate(ROUTES.SEARCH_CENTRAL_PANE.getRoute({query: CONST.SEARCH.TAB.EXPENSE.ALL}))}
```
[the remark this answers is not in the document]
luacmartins: Adam and I were still discussing normalizing this query. I think this is a temporary solution and we'll have to work on something better later on so it doesn't modify user input. NAB for now though

luacmartins:
>Bug 1 would be fixed automatically if we would put policyID to hash generation, but Im not sure if we want to do this in this PR (see my previous long comment).

I think we should include the policyID in the hash so we don't break the existing functionality

Kicu:
I tried to fix Bug 1 with my newest commit and upon quick check it should work, however i need to drop now so please double check if I didn't break anything else.

Temporarily we can add policyID to has generation which should make our hashes stable and reacting to changing policyID.

In future when we move `policyID` to inside filters as part of AST this bug will no longer exist

luacmartins (dismissed):
LGTM and tests well

rayane-d (changes requested):

luacmartins (approved):

Kicu:
@luacmartins do you mind sharing here that we decided to straighten out `policyIDs` in a separate PR?
Also the future name to be used will be `policyID` that is why I named it like that right now.

luacmartins:
@rayane-djouah https://github.com/Expensify/App/pull/45617#discussion_r1690391845 is NAB, we'll do that in a follow up since it can introduce several regressions and we want this PR merged soon.

luacmartins:
https://github.com/Expensify/App/pull/45617#discussion_r1691449154 this one doesn't seem to be a blocker either. Offset is working fine for now. We'll need to make changes to the parser in a follow up, so this will be addressed then.

rayane-d (approved):
LGTM and test well! Thanks all!
@luacmartins all yours!

luacmartins (approved):

src/components/Search/index.tsx
```diff
@@ -143,15 +144,14 @@ function Search({query, policyIDs, sortBy, sortOrder, isMobileSelectionModeActiv
...
+        Navigation.navigate(ROUTES.SEARCH_REPORT.getRoute(reportID));
     };
 
     const fetchMoreResults = () => {
         if (!searchResults?.search?.hasMoreResults || shouldShowLoadingState || shouldShowLoadingMoreItems) {
             return;
         }
-        const currentOffset = searchResults?.search?.offset ?? 0;
-        SearchActions.search({hash, query, offset: currentOffset + CONST.SEARCH.RESULTS_PAGE_SIZE, sortBy, sortOrder});
+        setOffset(offset + CONST.SEARCH.RESULTS_PAGE_SIZE);
```
hungvu193, line 154: We increased the offset here but we forgot to reset it when we changed the tab, once offset was increased here and we changed the tab, it kept the old offset which led to the wrong data arriving.
More details:
https://github.com/Expensify/App/issues/55235#issuecomment-2591253857

PR #45622, opened 2024-07-17
Revert "[RN 0.74 Upgrade] Upgrade react-native-pager-view to 6.3.3"

rafecolton:
Reverts Expensify/App#45573, which breaks scrolling in chat reports

cc @MrRefactor @carlosmiceli


rafecolton:
Re-tested and it seems this commit was not definitely the problem. Closing for now while I re-test. Discussion [here](https://expensify.slack.com/archives/C01GTK53T8Q/p1721238320025809?thread_ts=1721236850.175149&cid=C01GTK53T8Q)

PR #45726, opened 2024-07-18
Revert "Fix the map is centered on the user if it is not possible to create a route"

rlinoz:
Reverts Expensify/App#45575


rlinoz:
For me neither

PR #46100, opened 2024-07-24
update expense table record info

MuhammadTaimurTariq:
- Update to and from record in expense table
- "requestorEmai" record is coming in accountId field instead of managerId so field value get updated


dangrous:
hey @MuhammadTaimurTariq! Is this in response to a particular issue you were assigned? If so, let me know so I can understand the context. Thanks!

PR #46868, opened 2024-08-06
open scan screen after camera permission

DylanDylann:


DylanDylann:
This PR is for testing. Ignore please

Issue #49120, opened 2024-09-12
[$250] [Search v2.4] [App] Search Router List


luacmartins:
Implement [this section](https://docs.google.com/document/d/1Uf7UwMYTmek2qWe37SqkQZnOZYrnrfPucOLzXMv-3WM/edit#bookmark=id.da3fp9ml9bkw) of the doc

cc @Kicu @289Adam289 @Guccio163 @SzymczakJ

<details><summary>Issue Owner</summary>Current Issue Owner: @</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021834649901147542101</li>
        <li>Upwork Job ID: 1834649901147542101</li>
        <li>Last Price Increase: 2024-09-13</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @SzymczakJ</details>

joekaufmanexpensify:
SWM team, please comment on this issue if you're working on this one, and I will assign you!

SzymczakJ:
Hey! I’m Jakub Szymczak from Software Mansion, an expert agency, and I’d like to work on this issue!

SzymczakJ:
I'm just starting to work on it and it's hard to make estimates just now. I'll update tomorrow when I'll know more 🫡

joekaufmanexpensify:
Sounds good. TY!

SzymczakJ:
I should be able to put up a draft PR tommorow!

joekaufmanexpensify:
Sweet.

joekaufmanexpensify:
PR is WIP

joekaufmanexpensify:
@SzymczakJ is there an ETA to have this PR in review?

SzymczakJ:
I'm working shorter hours for next two days, but I'll try to make it ready for C+ review tomorrow!

joekaufmanexpensify:
Got it. TY!

joekaufmanexpensify:
Pending @SzymczakJ publishing the PR

joekaufmanexpensify:
@SzymczakJ is this week looking likely still to have the PR in review?

SzymczakJ:
I've fixed the review comments and was about to record videos, but I've bumped into strange bug on iOS, which I'm solving at the moment. Probably the safe ETA would be next week 😓

joekaufmanexpensify:
I see PR in review now. TY!

[...]

Issue #49116, opened 2024-09-12
[HOLD for payment 2024-10-14] [HOLD for payment 2024-10-11] [$250] [Search v2.4] [App] Add Recent Searches and Recent Chats sections


luacmartins:
Implement [this section](https://docs.google.com/document/d/1Uf7UwMYTmek2qWe37SqkQZnOZYrnrfPucOLzXMv-3WM/edit#bookmark=id.z7xv36bl2yqg) of the doc

cc @Kicu @289Adam289 @Guccio163 @SzymczakJ 

<details><summary>Issue Owner</summary>Current Issue Owner: @</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021834650319760395608</li>
        <li>Upwork Job ID: 1834650319760395608</li>
        <li>Last Price Increase: 2024-09-13</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @Christinadobrzyn</details>

SzymczakJ:
Hey! I’m Jakub Szymczak from Software Mansion, an expert agency, and I’d like to work on this issue!

Christinadobrzyn:
hi @SzymczakJ do you have an update on this for the team? TY!

SzymczakJ:
The draft PR should be ready tomorrow.

luacmartins:
PR in review

Christinadobrzyn:
Monitoring - https://github.com/Expensify/App/pull/49457

Christinadobrzyn:
Monitoring - https://github.com/Expensify/App/pull/49457

Heads up, I'm going to be ooo until 9/30. Since we're working on the PR I'm not going to assign a new BZ teammate. If you need one while I'm away please reach out in Slack for a volunteer.

Christinadobrzyn:
Monitoring - https://github.com/Expensify/App/pull/49457

Christinadobrzyn:
Hey @rayane-djouah and @SzymczakJ can you check these ^ deploy blocker messages to see if they apply here?

Kicu:
but @jasperhuangg removed the deploy blocker labels from the issue as they were not able to reproduce it.
So I don't think there is anything to verify here if there is no clear repro.

luacmartins:
Those blockers apply here, but they were already fixed with [this PR/CP](https://github.com/Expensify/App/pull/50143)

[...]

PR #49192, opened 2024-09-13
Search - Highlight newly added expense

ikevin127:
### Fixed Issues

$ https://github.com/Expensify/App/issues/48716
PROPOSAL: 

### Tests

1. Switch to the Search page tab.
2. Click the global create button (FAB) > Submit expense > Submit an expense.
3. Verify you remain on the search page.
4. Verify that the newly added expense appears in the table and the row is briefly highlighted to draw attention to it.

- [x] Verify that no errors appear in the JS console

### Offline tests

1. Switch to the Search page tab.
2. Go offline.
3. Click the global create button (FAB) > Submit expense > Submit an expense.
4. Verify you remain on the search page.
5. Go back online.
6. Verify that the newly added expense appears in the table and the row is briefly highlighted to draw attention to it.

### QA Steps

**TLDR**: Same as Tests.

1. Switch to the Search page tab.
2. Click the global create button (FAB) > Submit expense > Submit an expense.
3. Verify you remain on the search page.
4. Verify that the newly added expense appears in the table and the row is briefly highlighted to draw attention to it.

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
[...]

PR #49457, opened 2024-09-19
Add recent serches and recent chats in Search router

SzymczakJ:
### Details

This PR adds recent chats and recent searches to search router.

### Fixed Issues

$ https://github.com/Expensify/App/issues/49120
$ https://github.com/Expensify/App/issues/49116
PROPOSAL:

### Tests

1. Go to Search page, open Search router and try writing in the TextInput.
2. Make sure that written phrase is correctly searched, when submitted.
3. Go to some report, open Search router and make sure contextual Search works, clicking on "Search in {current report}" should add add phrase "in: {current report id}" to TextInput.
4. Make sure that recent Searches work.
5. Make sure that recent chats work.

### QA Steps

same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/9180b0aa-9415-4e75-a93c-0ffd30c266a2

</details>

<details>

https://github.com/user-attachments/assets/eb0a4310-d9af-4965-a2dc-3eb773211a07

</details>

<details>

https://github.com/user-attachments/assets/008d2eca-3db8-4310-a469-058e444cd451

</details>

<details>

https://github.com/user-attachments/assets/411eb335-439d-43f0-8920-0753a7448a48

</details>

<details>

https://github.com/user-attachments/assets/fcf5e0e4-de9e-4251-9b0f-c89ad2290da8

</details>

<details>

https://github.com/user-attachments/assets/914deb9e-613e-460d-bf22-b27c70d05186

</details>


luacmartins:
This is looking really good so far!

SzymczakJ:
Sorry for the slow pace, I was working shorter hours today 😅 I asked @Kicu for internal review but he will complete it next week, as he's already gone.

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -1,27 +1,57 @@
...
     const [currentQuery, setCurrentQuery] = useState<SearchQueryJSON | undefined>(undefined);
+    const contextualReportID = useNavigationState<Record<string, {reportID: string}>, string | undefined>((state) => {
+        return state.routes.at(-1)?.params?.reportID;
+    });
+    const [recentSearches] = useOnyx(ONYXKEYS.RECENT_SEARCHES);
+    const sortedRecentSearches = Object.values(recentSearches ?? {}).sort((a, b) => {
+        const dateA = new Date(a.timestamp);
+        const dateB = new Date(b.timestamp);
+        return dateB.getTime() - dateA.getTime();
+    });
```
Kicu, line 41: I think we could do `useMemo` for sorting - in this case it will never be heavy, but its cleaner/safer to add it

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -45,21 +75,46 @@ function SearchRouter() {
...
+=======
+>>>>>>> main
     useKeyboardShortcut(CONST.KEYBOARD_SHORTCUTS.ESCAPE, () => {
         closeSearchRouter();
         clearUserQuery();
     });
 
     const modalType = isSmallScreenWidth ? CONST.MODAL.MODAL_TYPE.CENTERED : CONST.MODAL.MODAL_TYPE.POPOVER;
-    const isFullWidth = isSmallScreenWidth;
+    const isFullScreen = isSmallScreenWidth;
```
Kicu, line 116: I think it will be confusing for people to see:
`isFullScreen = isSmallScreen` - we have the context and understand why this happens, but for other devs I think it might be confusing 😅 

Since you stopped passing `isFullWidth` down to `SearchRouterInput` then I would completely drop this variable, and instead just do:
`const modalWidth = isSmallScreenWidth ? ...`
and 
`const modalListHeight = isSmallScreenWidth ? ... : styles.mh85vh`

then just use the consts

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -70,11 +125,21 @@ function SearchRouter() {
...
-                        isFullWidth={isFullWidth}
                         onChange={onSearchChange}
-                        onSubmit={onSearchSubmit}
+                        onSubmit={() => {
+                            onSearchSubmit(currentQuery);
+                        }}
+                    />
+
+                    <SearchRouterList
+                        currentSearch={currentQuery}
```
Kicu, line 137: can we do:
```suggestion
                      currentQuery={currentQuery}
```

to have less different names?

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,125 @@
...
+        return (
+            <UserListItem
+                pressableStyle={styles.br2}
+                // eslint-disable-next-line react/jsx-props-no-spreading
+                {...(props as UserListItemProps<OptionData>)}
+            />
+        );
+    }
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    return <SingleIconListItem {...(props as SingleIconListItemProps<ListItemWithSingleIcon & ItemWithQuery>)} />;
```
Kicu, line 43: I'm quite unhappy that we need to do so many `as` casting here. Can we make this better with some smarter type guards?

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    if (reportForContextualSearch) {
+        sections.push({
+            data: [
+                {
+                    text: `${translate('search.searchIn')}${reportForContextualSearch.text ?? reportForContextualSearch.alternateText}`,
+                    singleIcon: Expensicons.MagnifyingGlass,
+                    // We will change it to different behaviour when Search 2.5 autocomplete will be implemented
+                    query: `in:${reportForContextualSearch.reportID}`,
```
Kicu, line 72: I think there is a but here, the query that will be searched should be `in:${reportForContextualSearch.reportID}` **together** with the `currentSearch?.inputQuery`

So when I type in `test` I want the query to be: "in:123123123 test"

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    const recentSearchesData = recentSearches?.map(({query}) => ({
+        text: query,
+        singleIcon: Expensicons.History,
+        query,
+        keyForList: query,
+    }));
+
+    if (recentSearchesData) {
```
Kicu, line 87: shouldn't this test for `recentSearchesData.length` as well?

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,125 @@
...
+        sections.push({title: translate('search.recentSearches'), data: recentSearchesData});
+    }
+
+    const recentReportsWithStyle = recentReports.map((item) => ({...item, pressableStyle: styles.br2}));
```
Kicu, line 91:
```suggestion
    const styledRecentReports = recentReports.map((item) => ({...item, pressableStyle: styles.br2}));
```

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,125 @@
...
+    sections.push({title: translate('search.recentChats'), data: recentReportsWithStyle});
+
+    const onSelectRow = useCallback(
+        (item: OptionData | ItemWithQuery) => {
+            // This is case for handling selection of "Recent search"
```
Kicu, line 96:
```suggestion
            // handle selecting "Recent search"
```

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,125 @@
...
+            if ('query' in item && item?.query) {
+                const queryJSON = SearchUtils.buildSearchQueryJSON(item?.query);
+                onRecentSearchSelect(queryJSON, true);
+                return;
+            }
+
+            // This is case for handling selection of "Recent chat"
```
Kicu, line 103: same simplify comment

src/components/SelectionList/Search/SingleIconListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+            pressableStyle={[[styles.singleIconListItemStyle, item.isSelected && styles.activeComponentBG, isFocused && styles.sidebarLinkActive, item.cursorStyle]]}
+            wrapperStyle={[styles.flexRow, styles.flex1, styles.justifyContentBetween, styles.userSelectNone, styles.alignItemsCenter]}
+            isFocused={isFocused}
+            onSelectRow={onSelectRow}
+            keyForList={item.keyForList}
+            onFocus={onFocus}
+            hoverStyle={item.isSelected && styles.activeComponentBG}
+        >
+            <>
+                {!!item.singleIcon && (
```
Kicu, line 37: would this work?
```suggestion
                {item.singleIcon && (
```

src/components/SelectionList/Search/SingleIconListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    />
+                    {!!item.alternateText && (
+                        <TextWithTooltip
+                            shouldShowTooltip={showTooltip ?? false}
+                            text={item.alternateText}
+                            style={[styles.textLabelSupporting, styles.lh16, styles.pre]}
+                        />
+                    )}
+                </View>
+                {!!item.rightElement && item.rightElement}
```
Kicu, line 66: Will we ever actually pass any `rightElement`? If not then I would remove this line

Kicu (approved):
I tested and looks really good.
I found one small bug and a bunch of small improvements with naming and comments, so please check them and update.

## Also question/possible bug:
Should recent reports also be filterable? because right now if I use chat finder, I can find different chats there, but in your PR they stay static.
We can of course add this in a separate PR, but I think if the plan is to completely remove Chat finder, then we should reimplement all its functionalities here.

luacmartins:
> Should recent reports also be filterable? 

Yes, we can add this in a follow up, but we should be able to filter reports while typing into the search router

src/components/Search/SearchRouter/SearchRouterList.tsx
SzymczakJ: I've tried making it simpler but in the end, that's as pretty as it gets, because you have to cast somewhere.
SelectionList is never used in such case, that there are two different items and OptionData is pretty different from ListType with small inconsistencies(like in one of them alternativeText is `string | undefined | null` and the second is `string | undefined`, and the list of inconsistencies goes on and on).
I'd like to avoid refactoring OptionData and SelectionList types in this PR as it's pretty complicated and I don't want to block our work with this PR, I could then do it in a follow-up PR. WDYT? Also if you have a neat idea how to avoid type casting then let me know because I've run out of ideas.

SzymczakJ:
As it turns out that small refactor is needed I just need one day to make it C+ reviewable.

src/components/Search/SearchRouter/SearchRouterList.tsx
SzymczakJ: You're right. I just focused on some other design doc screen shot and forgot about this, I will fix it tomorrow.

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -1,71 +1,117 @@
...
-            // eslint-disable-next-line
-            console.log('parsedQuery', queryJSON);
-
-            setUserSearchQuery(queryJSON);
-        } else {
-            // Handle query parsing error
-        }
-    }, SEARCH_DEBOUNCE_DELAY);
+    const onSearchChange = useCallback(
+        debounce((userQuery: string) => {
```
Kicu, line 65: @SzymczakJ 
this construct `useCallback(debounce(...` will not pass our linting. I know because I faced this problem in the past. 
Feel free to read more about this lint rule.

The solution will be either to drop `useCallback` for now, and make sure the function inside is not too computation-heavy, or eslint-ignore this line.
Alternatively check this article: https://kyleshevlin.com/debounce-and-throttle-callbacks-with-react-hooks/

For the record I think this warning is more like:
> eslint cannot figure out what this function argument are, so it shouts a warning

than
> this is bad react code

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -76,11 +122,24 @@ function SearchRouter() {
...
-                        onSubmit={onSearchSubmit}
+                        text={textInputValue}
+                        setText={setTextInputValue}
+                        updateSearch={onSearchChange}
+                        onSubmit={() => {
+                            onSearchSubmit(userSearchQuery);
+                        }}
+                    />
+
+                    <SearchRouterList
```
Kicu, line 135: please drop empty line before this, I think in JSX we should not be using empty lines

src/components/Search/SearchRouter/SearchRouterInput.tsx
```diff
@@ -1,34 +1,30 @@
...
+        setText(text);
+        updateSearch(text);
     };
 
-    const modalWidth = isFullWidth ? styles.w100 : {width: variables.popoverWidth};
-
     return (
         <BaseTextInput
-            value={value}
+            value={text}
```
Kicu, line 23: why not keep `value` to have `value=value`? its less names, less hoops to jump when reading the code?

Kicu (commented):
I dropped some extra 1-2 comments for you, since you have a lot of linting errors.

I have not yet checked the complicated type casts - will do that later today

src/components/Search/SearchRouter/SearchRouterList.tsx
Kicu: I have a proposal for a slightly alternate types hierarchy for this

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -45,21 +75,46 @@ function SearchRouter() {
...
+            const queryString = SearchUtils.buildSearchQueryString(query);
+            Navigation.navigate(ROUTES.SEARCH_CENTRAL_PANE.getRoute({query: queryString}));
+            clearUserQuery();
+        },
+        [closeSearchRouter],
+    );
 
+<<<<<<< HEAD
+    useKeyboardShortcut(
+        CONST.KEYBOARD_SHORTCUTS.ENTER,
```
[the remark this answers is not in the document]
SzymczakJ: I think we still need to handle ENTER in this case, because SelectionList(in BaseSelectionList) is catching ENTER shortcut and therefore it cannot be caught by SearchInput

src/components/SelectionList/Search/SearchQueryListItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+                            style={[styles.textLabelSupporting, styles.lh16, styles.pre]}
+                        />
+                    )}
+                </View>
+            </>
+        </BaseListItem>
+    );
+}
+
+SearchQueryListItem.displayName = 'SingleIconListItem';
```
luacmartins, line 78:
```suggestion
SearchQueryListItem.displayName = 'SearchQueryListItem';
```

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,149 @@
...
+};
+
+function isSearchQueryListItem(listItem: UserListItemProps<OptionData> | SearchQueryListItemProps): listItem is SearchQueryListItemProps {
+    if ('singleIcon' in listItem.item && listItem.item.singleIcon && 'query' in listItem.item && !!listItem.item.query) {
+        return true;
+    }
+    return false;
+}
+
+function SearchRouterItem(props: UserListItemProps<OptionData> | SearchQueryListItemProps) {
```
luacmartins, line 39: NAB might be good to separate this component into its own file.

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    const styledRecentReports = recentReports.map((item) => ({...item, pressableStyle: styles.br2}));
+    sections.push({title: translate('search.recentChats'), data: styledRecentReports});
+
+    const onSelectRow = useCallback(
+        (item: OptionData | SearchQueryItem) => {
+            if (!('query' in item) || !item.query) {
+                // Handle selection of "Recent chat"
+                closeAndClearRouter();
+                if ('reportID' in item && item?.reportID) {
```
luacmartins, line 115: Maybe refactoring this conditionals into a "type guard" might make our intentions clearer, e.g. `isReportItem` or `isLoginItem`

luacmartins (changes requested):
Awesome job here. I'm really excited for this PR. I left some suggestions, but it's looking great!

luacmartins:
I can't seem to see the recent chats. Am I missing something?
Screenshot 2024-09-27 at 8 16 12 PM

luacmartins:
@rayane-djouah let's prioritize reviewing this PR

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -76,11 +130,31 @@ function SearchRouter() {
...
+                        updateSearch={onSearchChange}
+                        onSubmit={() => {
+                            onSearchSubmit(userSearchQuery);
+                        }}
+                        routerListRef={listRef}
+                    />
+                    <SearchRouterList
+                        currentQuery={userSearchQuery}
+                        reportForContextualSearch={contextualReportData}
+                        recentSearches={sortedRecentSearches?.slice(0, 5)}
```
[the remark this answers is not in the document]
SzymczakJ: That's what I initially thought, but I got bigger object for like 3 or 4 times 😄, so I did it for safety reasons. Maybe some Onyx methods are merging Objects in a wrong way or something.
After relog the objects were always capped by 5.

src/components/Search/SearchRouter/SearchRouterList.tsx
SzymczakJ: I was thinking about this too.
The argument for keeping `SearchRouterItem` component in `SearchRouterList` is that it's highly specific to search router list feature and I don't want it to be used anywhere else. When we keep this component here it's less likely that somebody will export it and use for something not related.

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -1,71 +1,125 @@
...
+        // eslint-disable-next-line react-compiler/react-compiler
+        () =>
+            debounce((userQuery: string) => {
+                if (!userQuery) {
+                    clearUserQuery();
+                    listRef.current?.updateAndScrollToFocusedIndex(-1);
+                    return;
+                }
+                listRef.current?.updateAndScrollToFocusedIndex(0);
+                const queryJSON = SearchUtils.buildSearchQueryJSON(userQuery);
```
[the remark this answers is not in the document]
SzymczakJ: It was initially implemented like this by Mateusz. Right now it may not feel like the best idea, but when we will add autocomplete we will have to parse onChange to provide autocomplete suggestions. 
I can refactor it to not be parsed on every change but we will have to roll it back when we implement 2.5, so should I do it then? WDYT @luacmartins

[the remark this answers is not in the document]
luacmartins: Yea, I think we should refactor it right now. I just started working on the autocomplete doc and I think we might want a different parser for it that only captures the filter key and checks for matches, without building the AST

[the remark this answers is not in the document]
SzymczakJ: I already started refactoring it, but now I have second thoughts.
Parser with or without AST will have to parse input on text change to provide suggestions, so after deleting this we will have to add this once more. Also this code is already working on main and is not that heavy performance-wise.
I could delete Log.alert and console.logs so we don't log things on every keystroke and don't clutter the terminal, but I'm highly in favour of leaving this code as it is.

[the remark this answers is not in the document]
Kicu: hey it was me that originally introduce the code that parses after every keypress.
Also please note that there is 150ms debounce on this function, so that we do not run it  _too_ often.

In general I was feeling that whatever user types in should go through parser as it feels safer and gives us more information about the query.
Yes its true that right now we have no autocomplete of any kind yet, but it's just easier to work on parsed object.
If you worry about speed and computation - I wouldn't at this point, the parser is very fast and it will run only several times.

Also like @SzymczakJ said - in the long there will be **a** parser in this place, so that we can display the autocompletes to the user. 
Does it really make sense to remove the parsing now only to add it in two weeks?

And the constant alert logs in console are the problem, then we could do either:
 - pass the flag `buildSearchQueryJSON` to fail silently and not log, `false` in every other case but this one
 - make `buildSearchQueryJSON` throw and manually handle parser errors in every place; in here we would simply do nothing

SzymczakJ:
>  we should be able to filter reports while typing into the search router

FYI I've just added that. I also have one question: When serchingReports on server, should we show loading stane on input just like we do on ChatFinderPage? @luacmartins

rayane-d (changes requested):

rayane-d (changes requested):

rayane-d (changes requested):

rayane-d (changes requested):

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -1,71 +1,157 @@
...
-    const clearUserQuery = () => {
-        setUserSearchQuery(undefined);
-    };
-
-    const onSearchChange = debounce((userQuery: string) => {
-        if (!userQuery) {
-            clearUserQuery();
-            return;
+    const {options, areOptionsInitialized} = useOptionsList({
+        shouldInitialize: true,
```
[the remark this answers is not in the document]
SzymczakJ: I was wondering, if I should add it but in the end I don't know why do we even need to pass `shouldInitialize`. I don't understand why do we need to wait for transitionEnd to initialize some data in context. I also didn't notice any bugs when having this set to true. Maybe you have some more context @rayane-djouah

dubielzyk-expensify:
The padding on the router is quite off. Here's the comparison with the designs:
CleanShot 2024-10-01 at 20 28 18@2x

Hopefully those values make sense, but let me know and I'm happy to provide more specs :) 

cc @Expensify/design

SzymczakJ:
> The In filter is specific to chat types, but when we click on the in: <chat_name> option, type a keyword, and press the enter key, the search results show the expense type results instead. (probably we need to add type:chat to the input as well?)

This is also a good question, **should we add type:chat to the input as well?**

luacmartins:
>When serchingReports on server, should we show loading stane on input just like we do on ChatFinderPage?

Yea, we probably want to show a loading skeleton for it. I think we could tackle it in a follow up though.

>I guess we should also be able to filter the recent searches while typing into the search bar, right?

Same here. We should be able to filter recent searches while typing, but we can do so in a follow up.

>This is also a good question, should we add type:chat to the input as well?

Yea, I think we should since that filter is chat specific.

luacmartins:
@SzymczakJ I'm getting a crash on the latest commit 
Screenshot 2024-10-01 at 6 24 20 AM

SzymczakJ:
This PR actually breaks mine 😭 https://github.com/Expensify/App/pull/49192
It adds `useAnimatedHighlightStyle` to BaseListItem and therefore the Selection list cannot be used outside of ScreenWrapper and Navigation context.
I'd consider it as a bug because why would some generic list that we use all over our App depend on Navigation and ScreenWrapper state. This will also limit us in the future, because we are tied to these two contexts when using SelectionList. WDYT @ikevin127 @luacmartins @rayane-djouah

luacmartins:
Damn, yea we should fix that. @ikevin127 are you able to work on a fix? @SzymczakJ left some comments on how we can fix that

src/components/Search/SearchRouter/SearchRouter.tsx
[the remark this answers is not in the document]
SzymczakJ: This Modal is also outside any ScreenWrapperContext so it does not have `didScreenTransitionEnd`

luacmartins:
FWIW I completely removed the `didScreenTransitionEnd` code in the hook and retested the functionality in the Search page and the Workspace page (where this hook is used) and it seems to work just fine.

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,155 @@
...
+) {
+    const styles = useThemeStyles();
+    const {translate} = useLocalize();
+    const sections: Array<SectionListDataType<OptionData | SearchQueryItem>> = [];
+
+    if (currentQuery?.inputQuery) {
+        sections.push({
+            data: [
+                {
+                    text: currentQuery?.inputQuery,
```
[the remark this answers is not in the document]
luacmartins: I think we should keep the input values as is (without changing them). This will be solved by autocomplete in Search v2.5

ikevin127:
@luacmartins Sure, that makes sense and will definitely open a follow-up PR to apply the suggested changes 👍

@SzymczakJ Thanks for pointing it out and suggesting the changes!

Will tag you both once it's open, hopefully today!

luacmartins:
Thanks @ikevin127! Let's prioritize this fix since it breaking this PR which is critical for Search

ikevin127:
@luacmartins Sure, I will start workingnon it right now.

Just to make sure before I start, what I need to change to unblock this PR is following:
1. Move `useAnimatedHighlightStyle` outside of `BaseListItem` and instead have it added in `TransactionListItem` and `ReportListItem`.
2. Remove `didScreenTransitionEnd` from `useAnimatedHighlightStyle`.

Both of the above or only the latest ?

luacmartins:
I think 1 is ideal, since we're still unsure about the consequences of 2

ikevin127:
- @luacmartins Just opened the PR https://github.com/Expensify/App/pull/50028.

luacmartins:
Thank you! I assigned @ishpaul777 to review it since he reviewed the original PR.

SzymczakJ:
Fixed comments, conflicts and UI styles. Changes in `BaseListItem` are temporary and will be removed when https://github.com/Expensify/App/pull/50028 is merged.

SzymczakJ:
PR updated

luacmartins:
@rayane-djouah PR is updated. I think we can resume review

luacmartins:
NAB and this is a backend issue, but I think we should exclude canned searches from the recent searches. 
Screenshot 2024-10-02 at 1 10 47 PM

luacmartins:
NAB the section headers have additional left padding, they should be aligned with the icon below
Screenshot 2024-10-02 at 1 11 39 PM

Like we do here
Screenshot 2024-10-02 at 1 16 34 PM

luacmartins:
NAB we can address this in a follow up too. I think we should update the debounce as follows:

1. Don't debounce the current user query. We should update it immediately in the list.
2. Don't debounce the hiding/unhiding recent searches when there's input/no input. We should update that immediately in the list.
3. Debounce input for recent chats.

I think this will make the experience feel a bit better for the user.

https://github.com/user-attachments/assets/6a0c8bbf-4cb7-4193-b1b9-cb3f6e5b4031

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -1,71 +1,167 @@
...
+    const {translate} = useLocalize();
+    const [betas] = useOnyx(ONYXKEYS.BETAS);
+    const [recentSearches] = useOnyx(ONYXKEYS.RECENT_SEARCHES);
+    const [isSearchingForReports] = useOnyx(ONYXKEYS.IS_SEARCHING_FOR_REPORTS, {initWithStoredValues: false});
 
     const {isSmallScreenWidth} = useResponsiveLayout();
     const {isSearchRouterDisplayed, closeSearchRouter} = useSearchRouterContext();
+    const listRef = useRef<SelectionListHandle>(null);
 
+    const [textInputValue, debouncedInputValue, setTextInputValue] = useDebouncedState('', 500);
```
luacmartins, line 44: NAB Left a comment about the debounce, but I think we should only debounce the search in the server and nothing else. That'll make the App feel snappier and result in a better experience.

rayane-d (changes requested):

src/components/Search/SearchRouter/SearchRouter.tsx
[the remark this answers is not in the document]
luacmartins: I think we'll end up using a different parser for this that handles just autocomplete. We only really need the query AST when submitting the search to the server right?

luacmartins (approved):
LGTM. I left a few comments but I think we can address those in a follow up.

luacmartins:
@rayane-djouah I think we can address those polish items in a follow up

luacmartins:
Current issues to polish:
- [Remove canned searches from recent searches](https://github.com/Expensify/App/pull/49457#issuecomment-2389491816) (pending confirmation from product team)
- [Remove extra padding from section headers](https://github.com/Expensify/App/pull/49457#issuecomment-2389501254)
- [Update debounce behavior](https://github.com/Expensify/App/pull/49457#issuecomment-2389510028)
- [Input is cleared before navigation](https://github.com/Expensify/App/pull/49457#issuecomment-2389570126)
- [Can't submit search with enter key](https://github.com/Expensify/App/pull/49457#discussion_r1785157667)
- [Search page scrolls to top when navigating back](https://github.com/Expensify/App/pull/49457#issuecomment-2389584335)

rayane-d (changes requested):

rayane-d (changes requested):

src/components/Search/SearchRouter/SearchRouterList.tsx
```diff
@@ -0,0 +1,184 @@
...
+            text: searchQueryJSON ? SearchUtils.getSearchHeaderTitle(searchQueryJSON, personalDetails, cardList, reports, taxRates) : query,
+            singleIcon: Expensicons.History,
+            query,
+            keyForList: query,
+        };
+    });
+
+    if (!currentQuery?.inputQuery && recentSearchesData && recentSearchesData.length > 0) {
+        sections.push({title: translate('search.recentSearches'), data: recentSearchesData});
+    }
```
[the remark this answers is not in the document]
luacmartins: @rayane-djouah we decided to just remove recent searches from the list as soon as the user starts typing. Reason is that the list only ever has 5 entries and filtering the list is a bit overkill.

rayane-d (approved):
LGTM and tests well. Let's address the above comments in follow-ups

luacmartins (approved):

luacmartins:
@SzymczakJ FYI https://github.com/Expensify/App/pull/49457#issuecomment-2389667691

luacmartins:
Cc @SzymczakJ could you please take a look at the performance regression above? I'm not sure if it came from this PR, but we should investigate

SzymczakJ:
Investigating

hannojg:
The reported performance regression can be ignored, see:

- https://github.com/Expensify/App/issues/50141

SzymczakJ:
@hannojg I've just made a PR with possible fix: https://github.com/Expensify/App/pull/50143. Performance was probably worsened by putting `useOptionList` code in SearchRouter which renders very high in the navigation tree. With my fix the data hooks for SearchRouter won't be called until the modal is opened.
The thing is that I can't run performance tests locally 😕. Can you check it for me?

hannojg:
Ah, i didn't realise that the component was rendered so high up in the tree. yes, agree, its probably likely due to that!

IuliiaHerets:
This PR is failing because of issue #50220

Issue #48760, opened 2024-09-07
[HOLD for payment 2024-12-16] [HOLD for payment 2024-12-11] [$250] [Held requests] Partial paying a report causes the wrong amounts to show in the LHN


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.0.30-15
**Reproducible in staging?:** Yes
**Reproducible in production?:** Yes
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @JmillsExpensify 
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1725649428907579

## Action Performed:
1. On a Collect policy, have user A submit several expenses to User B
2. As User B, hold at least two expenses, approve the rest
## Expected Result:
The report amount in the LHN is correct.
## Actual Result:
The report amount in the LHN only shows the amount of one expense
## Workaround:
Unknown
## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android: Native
- [ ] Android: mWeb Chrome
- [ ] iOS: Native
- [ ] iOS: mWeb Safari
- [x] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

![CleanShot 2024-09-06 at 20 44 35@2x](https://github.com/user-attachments/assets/9e03dff4-d748-4700-850d-33e675785926)

![CleanShot 2024-09-06 at 20 45 06@2x](https://github.com/user-attachments/assets/e938df51-7a84-40bd-9c1d-be1e9fef6d36)

https://github.com/user-attachments/assets/d0f83087-4e76-4ac9-b13f-1c07f397d79f

Add any screenshot/video evidence

</details>

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021833144594407206391</li>
[...]

PR #49971, opened 2024-10-01
Fix wrong total amount of hold expenses when approve/pay partially

bernhardoj:
### Details

When approving/paying the report partially, the total held amount is wrong.

### Fixed Issues

$ https://github.com/Expensify/App/issues/48760
PROPOSAL: https://github.com/Expensify/App/issues/48760#issuecomment-2337720896

### Tests

Same as QA Steps

### Offline tests

Same as QA Steps
### QA Steps

1. Open a workspace chat
2. Submit 3 expenses (1 expense with a different currency) and track 1 expense
3. Hold 2 of the submitted expenses (1 with a different currency)
4. Go back to the main chat
5. Approve the report partially
6. Verify the main chat LHN owes amount is correct (shows the total amount of the hold expenses)

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/0ea90dd1-74e4-4ca4-a4c6-9ba6ebdd547a

</details>

<details>

https://github.com/user-attachments/assets/34a03caa-bb20-4b9d-a23d-d1e3cc2b39fb

</details>

<details>

https://github.com/user-attachments/assets/bebe14b2-1539-43ca-b074-216469072b87

</details>

<details>

https://github.com/user-attachments/assets/8958638d-6358-493c-99c9-b84f27378878

</details>

<details>

https://github.com/user-attachments/assets/eea47f8d-bd62-4444-896e-52987fd6d1eb

</details>

<details>

https://github.com/user-attachments/assets/c70549c0-f08b-45f0-a9fa-e0f3b6a25f68

</details>


src/libs/actions/IOU.ts
```diff
@@ -6448,7 +6448,7 @@ function getReportFromHoldRequestsOnyxData(
         chatReport.reportID,
         chatReport.policyID ?? iouReport?.policyID ?? '',
         recipient.accountID ?? 1,
-        holdTransactions.reduce((acc, transaction) => acc + transaction.amount, 0) * (ReportUtils.isIOUReport(iouReport) ? 1 : -1),
+        ((iouReport?.total ?? 0) - ((iouReport?.unheldTotal ?? 0) + (iouReport?.nonReimbursableTotal ?? 0))) * (ReportUtils.isIOUReport(iouReport) ? 1 : -1),
```
bernhardoj, line 6451: Since the `unheldTotal` doesn't include the `nonReimbursableTotal`, I add them together. If I have 3 expenses,

A: $1 (track, non-reimbursable)
B: $2 (held)
C: $3

The total is $6, unheldTotal is $3, and nonReimbursableTotal is $1.

held total = $6 - ($3 + $1) = $2

bernhardoj: I'm pretty confident that we can/should add them together because of this logic.
https://github.com/Expensify/App/blob/76c5c81542e598c147f57cf149122f4e124dfe10/src/libs/actions/IOU.ts#L6667-L6670

The logic above calculates the pay amount to show.

It subtracts the total from the nonReimbursableTotal. But if there are held expenses and it's paid partially, then we just use the unheldTotal which means the unheldTotal doesn't contain the nonReimbursableTotal.

unheldTotal = the total reimbursable amount that is not held

src/libs/actions/IOU.ts
```diff
@@ -6448,7 +6448,7 @@ function getReportFromHoldRequestsOnyxData(
...
         getCurrency(firstHoldTransaction),
```
s77rt, line 6452: Can you fix the currency to use the iou's currency

s77rt (changes requested):

src/libs/actions/IOU.ts
bernhardoj: Oh, forgot about it. Fixed.

s77rt (changes requested):
Approving non-reimbursable transactions shows incorrect amount, in the below screenshot it should state "Approve only DZD 5.00"

Screenshot 2024-10-01 at 4 49 25 PM

bernhardoj:
Yep, I noticed that too, but that's how it's in prod too. I was thinking of adding the `unheldTotal` with the `nonReimbursableTotal` too.

Before: Approve only $unheldTotal
After: Approve only $unheldTotal+nonReimbursableTotal

Do you agree with that? Or should we have someone from the team to confirm the behavior first?

s77rt:
I think the calculated `unheldTotal` is wrong. It should already include the non-reimbursable amounts. Can we fix this here?

bernhardoj:
`unheldTotal` isn't calculated in the FE.

s77rt:
Ah, I think this needs to be fixed in BE then. cc @robertjchen 

**Bug:** `report.unheldTotal` does not include non-reimbursable transactions

**Steps to reproduce:**
1. Have an IOU report with multiple expenses including non-reimbursable transactions
2. Hold all reimbursable transactions
3. Press Approve

**Expected:** the green button to state "Approve only DZD X.XX"
**Results:** the green button states "Approve only DZD 0.00" (see https://github.com/Expensify/App/pull/49971#pullrequestreview-2340785467)

Suggested solutions:
1. Have `unheldTotal` include non-reimbursable transactions
2. Add another prop e.g. `unheldNonReimbursableTotal`

s77rt:
@bernhardoj 

> After: Approve only $unheldTotal+nonReimbursableTotal

That won't work because we are not taking into account held non-reimbursable transactions

bernhardoj:
Non reimbursable transaction can't be hold.

trjExpensify:
>Non reimbursable transaction can't be hold.

Hm, is that true @JmillsExpensify @robertjchen?

JmillsExpensify:
That should not be true, no. You should be able to hold any transaction.

bernhardoj:
Oh, then it's a bug then. There is no option to Hold in ND.

robertjchen:
So, to clarify- given this, the backend needs to be updated such so that: `unheldTotal include[s] non-reimbursable transactions`?

bernhardoj:
I'm honestly not sure if we should include non-reimbursable to `unheldTotal`. If we have 1 hold transaction and 1 non-reimbursable transaction,

Currently, the confirmation modal will say either Pay 0 (unheldTotal) or Pay fullAmount.
image

If we include non-reimbursable transaction to `unheldTotal`, the confirmation modal will either say Pay nonReimbursableAmount (unheldTotal) or Pay fullAmount.

> Add another prop e.g. unheldNonReimbursableTotal

I think it's best to include non-reimbursable transaction to `unheldTotal` and add a new `unheldNonReimbursableTotal` property, then we can calculate the unheldReimbursableTotal and show it on the confirm modal.

Before:
`total`: held and unheld total amount of reimbursable and non-reimbursable
`unheldTotal`: unheld total amount of reimbursable
`nonReimbursableTotal`: held and unheld total of non-reimbursable

After:
`total`: held and unheld total amount of reimbursable and non-reimbursable
`unheldTotal`: unheld total amount of reimbursable and non-reimbursable
`nonReimbursableTotal`: held and unheld total of non-reimbursable
`unheldNonReimbursableTotal`: unheld total of non-reimbursable


Calculating unheldReimbursableTotal will be = unheldTotal - unheldNonReimbursableTotal
"Pay unheldReimbursableTotal or Pay fullAmount"
"Approve unheldReimbursableTotal or Pay fullAmount"

> Oh, then it's a bug then. There is no option to Hold non-reimbursable in ND.

But first, I think we should fix this first?

trjExpensify:
How do we expect this to work, @JmillsExpensify? Is it:

- Only display the reimbursable report total in the modal for that second “pay everything and take what’s on hold off hold” choice? 
- When deciding to pay all reimbursable expenses, what happens to the held non-reimbursable expenses on the report? Do those come off hold as well, or do they stay on hold and get moved?

s77rt:
@bernhardoj 
> If we include non-reimbursable transaction to unheldTotal, the confirmation modal will either say Pay nonReimbursableAmount (unheldTotal) or Pay fullAmount.

That's the expected.

> I think it's best to include non-reimbursable transaction to unheldTotal and add a new unheldNonReimbursableTotal property, then we can calculate the unheldReimbursableTotal and show it on the confirm modal.

In the approve modal we should show the unheldTotal which should include both reimbursable and non-reimbursable transactions.

JmillsExpensify:
That's all fair feedback above, and on second thought, we struggled with this consideration in the design doc – enough that we put this in the out-of-scope section of the doc. In fact, there are kind of two cases to consider here:

> `unheldTotal`: unheld total amount of reimbursable and non-reimbursable

This only matters for approval, as you can legitimately approve both reimbursable and non-reimbursable expenses. It is impossible to pay a non-reimbursable in 100% of cases. 

As a result, I think if we went this route, then we would only show the `unheldTotal` as a sum of reimbursable and non-reimbursable for the approve case. We can never ask if you'd like to pay the `unheldTotal` for non-reimbursable expenses, because you can't pay them.

s77rt:
Ah right, then can we go with the following:

**BE:**
1. Make unheldTotal include both reimbursable and non-reimbursable transactions
2. Add unheldReimbursableTotal


**FE:**

In the approve modal display:
- Partial: unheldTotal
- Full: total

In the pay modal display:
- Partial: unheldTotal - unheldNonReimbursableTotal
- Full: total - nonReimbursableTotal


Does that make sense?

trjExpensify:
> Ah right, then can we go with the following:
> 
> **BE:**
> 
> 1. Make unheldTotal include both reimbursable and non-reimbursable transactions
> 2. Add unheldReimbursableTotal
> 
> **FE:**
> 
> In the approve modal display:
> 
> * Partial: unheldTotal
> * Full: total
> 
> In the pay modal display:
> 
> * Partial: unheldTotal - unheldNonReimbursableTotal
> * Full: total - nonReimbursableTotal
> 
> Does that make sense?

If we're adding `unheldReimbursableTotal`, why not use that in the Pay modal? instead of unheldTotal - unheldNonReimbursableTotal? 

I think this question is still outstanding:  

>When deciding to pay all reimbursable expenses, what happens to the held non-reimbursable expenses on the report? Do those come off hold as well, or do they stay on hold and get moved?

s77rt:
@trjExpensify 

> If we're adding unheldReimbursableTotal

Sorry, we won't, I meant to write `unheldNonReimbursableTotal`. Corrected

> When deciding to pay all reimbursable expenses, what happens to the held non-reimbursable expenses on the report? Do those come off hold as well, or do they stay on hold and get moved?

They come off hold same as held reimbursable transactions.

trjExpensify:
>Sorry, we won't, I meant to write unheldNonReimbursableTotal. Corrected

Gotcha. 👍 

>When deciding to pay all reimbursable expenses, what happens to the held non-reimbursable expenses on the report? Do those come off hold as well, or do they stay on hold and get moved?
>They come off hold same as held reimbursable transactions.

I think the UX might be a bit strange on that one as it's not communicated in the modal, but let's see I guess.

s77rt:
@robertjchen Are we aligned with the changes here https://github.com/Expensify/App/pull/49971#issuecomment-2397561749 if so, can you please work on the BE changes? Those are breaking-changes (due to `unheldTotal` change), do we have a pattern to follow in such cases?

robertjchen:
Thanks for the discussion here, those proposed BE changes make sense to me 👍  I'll get the PR up today and we'll get that deployed early next week (Tuesday).

As for potentially breaking changes around `unheldTotal`: it doesn't seem like we really use `unheldTotal` in any other flows, so I'm sure we'll be fine with the change, unless I'm missing something?

trjExpensify:
Did you get that BE PR up, @robertjchen?

robertjchen:
@trjExpensify Testing locally, should be up in a bit here! 👍

robertjchen:
Backend changes have been deployed! Please let me know if you are able to see the new totals now, thanks!

trjExpensify:
Wahoo!

s77rt:
@bernhardoj Looks we are ready for the FE changes

bernhardoj:
Nice, the `unheldTotal` alread includes the nonReimbursable too. But `unheldNonReimbursableTotal` is still not available yet, so I can't do this yet:

> FE:

> In the approve modal display:

> Partial: unheldTotal
> Full: total
> In the pay modal display:

> Partial: unheldTotal - unheldNonReimbursableTotal
> Full: total - nonReimbursableTotal

s77rt:
cc @robertjchen ^

robertjchen:
Hm, we are returning `unheldNonReimbursableTotal` as part of `GetChats` but not as part of the `reportSummary`. I'll update 🤔

robertjchen:
@bernhardoj Just pushed out a new change, let me know if it is present now! 🙏

bernhardoj:
@robertjchen I can see it now. However, I think we still have 1 problem. The `unheldNonReimbursableTotal` value is not in negative compared to other values.

image

Subtracting between `unheldTotal` and `unheldNonReimbursableTotal` will result in bigger negative value.

bernhardoj:
Also, I see we have another issue. If we submit:
$1 hold
$2

total = $3
unheldTotal = $2

Go offline and then we add an expense with a different currency, the displayed amount will be "Pay $3..." instead of "Pay $2...".

https://github.com/Expensify/App/blob/d1f6190f6503fea55d16ca9e66ea7d70788bad23/src/components/MoneyReportHeader.tsx#L163

That's because `canAllowSettlement` is now false because BE still calculates the conversion, so we use the full (reimbursable) amount.

Is this expected?

https://github.com/user-attachments/assets/5afa356a-6f6a-4ad9-930f-1c4387b0160c


If not, I think we can remove `canAllowSettlement` from the condition.

----

However, we still have another problem. If we submit a $10 expense first while online and then perform the above steps while offline (without adding the different currency expense), the `unheldTotal` is never updated optimistically, so,
total = $13
unheldTotal = $10

Before adding an expense with a different currency, the unheld total shown on the UI is fine because we have a logic to calculate it manually if there is a pending transaction, but only if `hasUpdatedTotal` is true.
https://github.com/Expensify/App/blob/d1f6190f6503fea55d16ca9e66ea7d70788bad23/src/libs/ReportUtils.ts#L7687-L7699

`hasUpdatedTotal` is only true if all transactions have the same currency or there is no pending transaction. So, the manual calculation works fine. But when we add the expense with a different currency, it will return the `unheldTotal` of the expense/IOU report, which is $10.

So, it looks like this:

submit: $10

"Pay $10 ..."

Go offline,
submit:
$1 hold
$2

"Pay $12 ..."

submit an expense with a different currency:

"Pay $10 ..."

Is this expected?

If not, I think we can remove the "optimistic calculation" here and instead optimistically update the `unheldTotal` and `unheldNonReimbursableTotal` only if the expense has the same currency.
https://github.com/Expensify/App/blob/793dcaf5de74f3ae9c6c32775b61cbf33cb78bf6/src/libs/ReportUtils.ts#L7687-L7699

cc: @robertjchen @s77rt @trjExpensify @JmillsExpensify

s77rt:
@bernhardoj Regarding the first bug, I'm not sure why is `canAllowSettlement` added. I have asked [here](https://github.com/Expensify/App/pull/33124#discussion_r1807143831).

Regarding the second bug:
>  I think we can remove the "optimistic calculation" here and instead optimistically update the unheldTotal and unheldNonReimbursableTotal only if the expense has the same currency.

What if the expense is using a different currency? The bug will still occur right? But yeah that "optimistic calculation" is not something that we follow, we should update the onyx values instead to ensure one source of truth

bernhardoj:
> What if the expense is using a different currency? The bug will still occur right?

It won't be added to `unheldTotal` and the amount will be grayed out if there is a different currency. That's how it works with `total` too.
https://github.com/Expensify/App/blob/66cf824036a51183d73dab2cc621901e43fe3196/src/libs/actions/IOU.ts#L2076-L2084

s77rt:
@bernhardoj Is the current behaviour that we replace `unheldTotal` with the new expense value?

bernhardoj:
@s77rt sorry, I don't quite get what you mean, but currently, the `unheldTotal` is never updated optimistically.

s77rt:
@bernhardoj In the second bug you mentioned:
> "Pay $12 ..."
> submit an expense with a different currency:
> "Pay $10 ..."

Where is that $10 coming from?

bernhardoj:
> Where is that $10 coming from?

It's the `unheldTotal`. I forgot to mention the 2nd bug is only repro after fixing the 1st bug. I updated my comment to explain it clearer (hopefully)

s77rt:
Okay so after fixing the second bug it will show a grayed out $12?

bernhardoj:
Correct.

s77rt:
Got it. Sounds good to me. Let's proceed with the FE changes while we wait on BE amount coefficient correction

bernhardoj:
> Let's proceed with the FE changes

For both issues or only the 1st one?

Btw, @robertjchen please check this [BE issue](https://github.com/Expensify/App/pull/49971#issuecomment-2422126966) left

robertjchen:
@bernhardoj Got it- to clarify, are you testing with an IOU money request or an expense report? I'll to change the sign based on the type of report 👍

src/libs/ReportUtils.ts
```diff
@@ -7677,26 +7677,37 @@ function hasUpdatedTotal(report: OnyxInputOrEntry<Report>, policy: OnyxInputOrEn
 /**
  * Return held and full amount formatted with used currency
  */
-function getNonHeldAndFullAmount(iouReport: OnyxEntry<Report>, policy: OnyxEntry<Policy>): string[] {
+function getNonHeldAndFullAmount(iouReport: OnyxEntry<Report>, policy: OnyxEntry<Policy>, isReimbursableOnly: boolean): string[] {
```
s77rt, line 7680: Rename prop to `shouldExcludeNonReimbursables`

s77rt:
> For both issues or only the 1st one?

Only the second for now

bernhardoj:
@robertjchen I tested it with the expense report

src/libs/ReportUtils.ts
bernhardoj: Done

bernhardoj:
> Only the second for now

Done for the expense report. I'm planning to do it for the IOU report too, but I realize there is something wrong.

In an IOU report, holding a report will subtract the `total` from the held expense amount.

For example,

Submit $10 and $2 to user A. The total is $12. But when I hold $2, the total becomes $8 ($10 - $2) and unheldTotal is $10 (correct).

image

I believe this isn't expected, right?

cc: @s77rt @robertjchen

robertjchen:
> But when I hold $2, the total becomes $8 ($10 - $2)

Right- I would expect the total to be $10, but not $8 🤔

robertjchen:
While addressing the signage issue with `unheldNonReimbursableTotal`, I noticed that `nonReimbursableTotal` also had the same signage issue. My fix will address both and make the signs line up with `total` and `unheldTotal`.

robertjchen:
Maybe the fact that the signage was wrong on the `nonReimbursableTotal` made it so that the math was incorrect? 🤷

bernhardoj:
> Maybe the fact that the signage was wrong on the nonReimbursableTotal made it so that the math was incorrect?

Hmm, I don't know if that would affect the calculation since it doesn't contain any non-reimbursable transaction for this case.

> But when I hold $2, the total becomes $8 ($10 - $2)

s77rt:
> But when I hold $2, the total becomes $8 ($10 - $2)

> Right- I would expect the total to be $10, but not $8 🤔

The total shouldn't change. Holding an expense should only make changes to the unHeldTotal. I expected the total to be still $12

robertjchen:
Ah that is true- the `total` should be the same. In any case, the `unheldNonReimbursableTotal` and `nonReimbursableTotal` fix has now been deployed. Let me know if you see the proper signage on those amounts, thanks! 🙇

src/libs/actions/IOU.ts
```diff
@@ -2075,9 +2075,15 @@ function getMoneyRequestInformation(
...
-            iouReport.total -= amount;
+        // Because of the Expense reports are stored as negative values, we subtract the total from the amount
+        if (iouReport?.currency === currency) {
+            if (typeof iouReport.total === 'number') {
+                iouReport.total -= amount;
+            }
+
+            if (typeof iouReport.unheldTotal === 'number') {
+                iouReport.unheldTotal -= amount;
+            }
```
s77rt, line 2086: Is there a case where `unheldTotal` wouldn't exist? If we have an expense report where all the transactions are reminbursable will we have:
1. `unheldTotal` set to `total`
2. `unheldTotal` not set at all (`undefined`)

cc @robertjchen

src/libs/actions/IOU.ts
```diff
@@ -6515,8 +6554,8 @@ function getReportFromHoldRequestsOnyxData(
         chatReport.reportID,
         chatReport.policyID ?? iouReport?.policyID ?? '',
         recipient.accountID ?? 1,
-        holdTransactions.reduce((acc, transaction) => acc + TransactionUtils.getAmount(transaction), 0),
-        getCurrency(firstHoldTransaction),
+        ((iouReport?.total ?? 0) - (iouReport?.unheldTotal ?? 0)) * (ReportUtils.isIOUReport(iouReport) ? 1 : -1),
+        iouReport?.currency ?? '',
         false,
         newParentReportActionID,
```
s77rt, line 6560: Probably as a follow up: I noticed we pass `reimbursable=false` when creating a new report (to move the held expenses). This is not exactly correct because the held expenses can be mixed (reimbursable and nonReimbursable) which means we also need to fix the calculated (nonReimbursableTotal) value in `buildOptimisticExpenseReport`.

src/libs/actions/IOU.ts
```diff
@@ -6515,8 +6554,8 @@ function getReportFromHoldRequestsOnyxData(
         chatReport.reportID,
         chatReport.policyID ?? iouReport?.policyID ?? '',
         recipient.accountID ?? 1,
-        holdTransactions.reduce((acc, transaction) => acc + TransactionUtils.getAmount(transaction), 0),
-        getCurrency(firstHoldTransaction),
+        ((iouReport?.total ?? 0) - (iouReport?.unheldTotal ?? 0)) * (ReportUtils.isIOUReport(iouReport) ? 1 : -1),
```
s77rt, line 6557: In line 6633, we pass `unheldTotal: 0,`. We also need to pass `unheldNonReimbursableTotal: 0,`

(It's related to the above point)

src/libs/actions/IOU.ts
```diff
@@ -7981,6 +8022,21 @@ function putOnHold(transactionID: string, comment: string, reportID: string, sea
...
+        const isExpenseReport = ReportUtils.isExpenseReport(iouReport);
+        const coefficient = isExpenseReport ? -1 : 1;
+        optimisticData.push({
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.REPORT}${iouReport.reportID}`,
+            value: {
+                unheldTotal: (iouReport.unheldTotal ?? 0) - TransactionUtils.getAmount(transaction, isExpenseReport) * coefficient,
+                unheldNonReimbursableTotal: !transaction?.reimbursable
+                    ? (iouReport.unheldNonReimbursableTotal ?? 0) - TransactionUtils.getAmount(transaction, isExpenseReport) * coefficient
+                    : iouReport.unheldNonReimbursableTotal,
```
s77rt, line 8035: NAB. Move the amount to a const to avoid calling getAmount twice.

src/libs/actions/IOU.ts
```diff
@@ -8084,6 +8142,21 @@ function unholdRequest(transactionID: string, reportID: string, searchHash?: num
...
+        const isExpenseReport = ReportUtils.isExpenseReport(iouReport);
+        const coefficient = isExpenseReport ? -1 : 1;
+        optimisticData.push({
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.REPORT}${iouReport.reportID}`,
+            value: {
+                unheldTotal: (iouReport.unheldTotal ?? 0) + TransactionUtils.getAmount(transaction, isExpenseReport) * coefficient,
+                unheldNonReimbursableTotal: !transaction?.reimbursable
+                    ? (iouReport.unheldNonReimbursableTotal ?? 0) + TransactionUtils.getAmount(transaction, isExpenseReport) * coefficient
+                    : iouReport.unheldNonReimbursableTotal,
```
s77rt, line 8155: Same ^

src/types/onyx/Report.ts
```diff
@@ -218,6 +218,9 @@ type Report = OnyxCommon.OnyxValueWithOfflineFeedback<
         /** For expense reports, this is the total amount requested */
         unheldTotal?: number;
 
+        /** Total amount of unheld and non-reimbursable transactions in an expense report */
```
s77rt, line 221:
```suggestion
        /** Total amount of unheld non-reimbursable transactions in an expense report */
```
NAB. Using "and" hints that we are summing two sets (transactions that are unheld and transactions that are non-reimbursable) where it's actually one set (transactions that are both unheld and non-reimbursable).

s77rt (changes requested):

src/types/onyx/Report.ts
bernhardoj: Make sense, updated!

src/libs/actions/IOU.ts
bernhardoj: Done

src/libs/actions/IOU.ts
bernhardoj: In `main`, we don't add `unheldTotal` and `unheldNonReimbursableTotal` when creating the report optimistically (buildOptimisticExpenseReport only, IOU not yet, waiting for the IOU total bug to be fixed), so it will be undefined. But in this PR, I already added both optimistically, so it will never be undefined anymore optimistically. I don't know about the BE.

src/libs/actions/IOU.ts
bernhardoj: Nice catch, I think we can update it in this PR. There is only 2 usages of `buildOptimisticExpenseReport` that passes the `reimbursable` param. I think we can change it to nonReimbursableTotal param instead.

src/libs/actions/IOU.ts
robertjchen: For BE, we will always return these fields.

src/libs/actions/IOU.ts
bernhardoj: Done

src/libs/actions/IOU.ts
bernhardoj: Done

bernhardoj:
> In any case, the unheldNonReimbursableTotal and nonReimbursableTotal fix has now been deployed. Let me know if you see the proper signage on those amounts, thanks!

@robertjchen thanks, it's now negative.

> Ah that is true- from that perspective, the total should technically be the same.

Now waiting for this one 😄

src/libs/actions/IOU.ts
bernhardoj: Thanks for the confirmation. That means we could technically update the typing for the `total`, `unheldTotal`, `nonReimbursableTotal`, and `unheldNonReimbursableTotal` to be non-undefined/null, but if we want to do that, we need to update all report optimistic creation (buildOptimisticChatReport, buildOptimisticTaskReport, etc.) to include them all, which isn't the case for now.

src/libs/actions/IOU.ts
```diff
@@ -2304,7 +2304,7 @@ function getTrackExpenseInformation(
 
         shouldCreateNewMoneyRequestReport = ReportUtils.shouldCreateNewMoneyRequestReport(iouReport, chatReport);
         if (!iouReport || shouldCreateNewMoneyRequestReport) {
-            iouReport = ReportUtils.buildOptimisticExpenseReport(chatReport.reportID, chatReport.policyID ?? '-1', payeeAccountID, amount, currency, false);
+            iouReport = ReportUtils.buildOptimisticExpenseReport(chatReport.reportID, chatReport.policyID ?? '-1', payeeAccountID, amount, currency, 0);
```
s77rt, line 2307:
```suggestion
            iouReport = ReportUtils.buildOptimisticExpenseReport(chatReport.reportID, chatReport.policyID ?? '-1', payeeAccountID, amount, currency, amount);
```

Track expense are non-reimbursable

s77rt (changes requested):

src/libs/actions/IOU.ts
bernhardoj: Oops, updated

bernhardoj:
Good news, I just retested and the IOU total issue doesn't happen anymore (hopefully it stays like this).

image

src/libs/IOUUtils.ts
```diff
@@ -86,18 +87,26 @@ function updateIOUOwnerAndTotal<TReport extends OnyxInputOrEntry<Report>>(
...
     if (actorAccountID === iouReport.ownerAccountID) {
         iouReportUpdate.total += isDeleting ? -amount : amount;
+        if (!isOnhold) {
+            iouReportUpdate.unheldTotal += isDeleting ? -amount : amount;
+        }
     } else {
         iouReportUpdate.total += isDeleting ? amount : -amount;
+        if (!isOnhold) {
+            iouReportUpdate.unheldTotal += isDeleting ? amount : -amount;
+        }
```
bernhardoj, line 101: I've updated the optimistic unheldTotal for IOU too. However, there is a problem.

1. User A submit $2 and $2 to user B (total = 4, unheldTotal = 4, owner = user A)
2. User A hold $2 (total = 4, unheldTotal = 2, owner user A)
3. User B submit $10

When user B submits, it will go into the `else` block, subtracting `total` from the amount (10). So, the new total will be -6. If it's < 0, we switch the owner with this logic and switch the sign too, so the new optimistic total is 6.
https://github.com/Expensify/App/blob/866ce4b3957111d2c3ea5a7dccc3c86b90ed4e7c/src/libs/IOUUtils.ts#L96-L100

The problem I see is, that BE still keeps the sign, so the `total` from BE is -6. Also, the `unheldTotal` is -8, which looks like it comes from `unheldTotal` (2) - amount (10). So, I apply this calculation for the optimistic one too as you can see above. I'm not sure if the `unheldTotal` should be -8, but at least I match it with the BE. I think it will be another big problem to solve and decide the expected behavior.

Notice the image below shows Pay -8 (unheldTotal).

image

src/libs/IOUUtils.ts
```diff
@@ -86,18 +87,26 @@ function updateIOUOwnerAndTotal<TReport extends OnyxInputOrEntry<Report>>(
...
     }
 
     if (iouReportUpdate.total < 0) {
         // The total sign has changed and hence we need to flip the manager and owner of the report.
         iouReportUpdate.ownerAccountID = iouReport.managerID;
         iouReportUpdate.managerID = iouReport.ownerAccountID;
         iouReportUpdate.total = -iouReportUpdate.total;
+        iouReportUpdate.unheldTotal = -iouReportUpdate.unheldTotal;
```
bernhardoj, line 109: Even though the BE stores it as negative, I still switch the sign here optimistically just to make it consistent with `total` and I think it should be the correct amount, in positive.

src/libs/IOUUtils.ts
s77rt: Hmm this looks tricky, i think we need to agree on the expected behaviour on what the total and the unheldTotal should be in this case and also in the case where we unhold the $2 request. (can we still unhold that request? and how can do it?)

bernhardoj: We can still unhold it and the `unheldTotal` will be the same as `total` (-6). If I hold it again, it will be -8 again.

> User A submit $2 and $2 to user B (total = 4, unheldTotal = 4, owner = user A)
> User A hold $2 (total = 4, unheldTotal = 2, owner user A)
> User B submit $10

I think if we have this case, the new total should be 6 (number is correct already in prod, but it's in negative) and unheldTotal should be:

current held = total (4) - unheldTotal (2) = 2
new unheldTotal = new total (6) - current held (2) = 4

I suspect the BE already did this calculation, but because the total is negative (-6), subtracting with 2 will be -8.

cc: @robertjchen

robertjchen: Is the problem around IOUs and expense reports having different signage? 🤔

s77rt: I think the problem is more related to the fact that both users can submit expenses in the same iou report. In the above scenario the unheldTotal will be grater than the total

total: $6
unheldTotal: $8

bernhardoj: > I think the problem is more related to the fact that both users can submit expenses in the same iou report.

I think this is expected.

> In the above scenario the unheldTotal will be grater than the total

And both are negative.

> unheldTotal on the backend is simply the reimbursable total (excluding held transactions) + non reimbursable total (excluding held transactions).

From the step I performed above, there is no non-reimbursable, so the wrong calculation is from the reimbursable total. 

> Is the problem around IOUs and expense reports having different signage? 🤔

I think the real problem here is only related to the IOU `total` that is negative after user B adds an expense that the amount is bigger than the current `total`, as I mentioned above.

> I suspect the BE already did this calculation, but because the total is negative (-6), subtracting with 2 will be -8.

The `total` is -6 and the held amount is 2. If the `total` is 6, then the issue will be gone, probably.

s77rt: @robertjchen Any updates on this? Do you know what should be the expected values for `total` and `unheldTotal` in this case?

robertjchen: > The total is -6 and the held amount is 2. If the total is 6, then the issue will be gone, probably.

IOU and expense report amounts have different signs, and I don't see what else we can change on the backend 🤔 Any discrepancies would have to be accounted for on the FE: could we use absolute values on the totals before performing the calculations? Would that solve the issue here? (We can then properly add the sign back if needed according to the type of the report).

bernhardoj: > Any discrepancies would have to be accounted for on the FE: could we use absolute values on the totals before performing the calculations?

The `unheldTotal` is also being calculated on the BE, so even if we do this on the FE, the value from the BE will still be wrong. Btw, this is an IOU issue only. Maybe the BE needs to apply the same logic as this?
https://github.com/Expensify/App/blob/2463bf74b4edabdde887636f7874fdbbfcf9eb2c/src/libs/IOUUtils.ts#L97-L102

robertjchen: To clarify, do we want the `total` amount to be negative for IOUs when it's returned from the backend? (or are we referring to `unheldTotal`) 🤔

bernhardoj: We want both `total` and `unheldTotal` to be positive for IOUs. Currently, it's both negative, but based on my previous comments, my guess is that `unheldTotal` is negative because the `total` is negative. If we can fix the `total`, then `unheldTotal` will hopefully be fixed too.

robertjchen: I can't change the report `total` on the backend, as many things rely on it and it has been the same for every expense report and IOU for many years.

On the FE, we check the report type and reverse the `total` depending on whether it's an IOU or not, can this logic be applied here as well?

https://github.com/Expensify/App/blob/2463bf74b4edabdde887636f7874fdbbfcf9eb2c/src/libs/ReportUtils.ts#L7699-L7722

bernhardoj: `getNonHeldAndFullAmount` already have that logic.

> const coefficient = isExpenseReport(iouReport) ? -1 : 1; 

We can use `Math.abs` but that still won't fix the `unheldTotal` wrong value issue.

> In the above scenario the unheldTotal will be grater than the total
>
> total: $6
> unheldTotal: $8

expected is $4.

s77rt: Why $4?
- A request $2
- A request $2 (on hold)
- B request $10

Now A has to pay B either:
1. $8
2. $6

bernhardoj: Hmm, actually, after thinking about it the 2nd time, I think `unheldTotal` should be $6. I think User A holding his expenses shouldn't affect the `unheldTotal`. The IOU interactions are really tricky 🤯

s77rt: @bernhardoj Can you make the required changes (if any) so that we have:
1. total $6 (or -$6)
2. unheldTotal $8 (or -$8)

The unheldTotal (absolute value) is greater than the total which is weird but given that the owner changes this is still logical. Now, the thing is that after A unhold his request, we will have to subtract that value from unheldTotal which is again also weird as usually we sum that value up.

One more thing, we need to know when to sum/subtract, consider this scenario:

- A request $2
- A request $2 (on hold)
- B request $10
- B request $10 (on hold)

The total should be: $16 (absolute value)
The unheldTotal should be: $8 (absolute value)

If A unhold his request the unheldTotal should be $6 (subtract)
If B unhold his request the unheldTotal should be $18 (sum)

So it seems that depending on the owner we can determine if we we should add/subtract to/from unheldTotal

s77rt: Let me know if that makes sense

bernhardoj: I'm starting to believe the unheldTotal is bigger is correct, but yeah, that's how BE works, and the optimistic data is already works that way too. (except it's in positive because of the logic in the conversation below)

https://github.com/user-attachments/assets/23e9b10f-b2ec-4eb6-a072-c58b1e7ba214

s77rt: Okay great, so what needs to be done so we can move forward, just the sign right?

s77rt: @Gonals May be able to help if that logic is no longer true?

Gonals: AFAIK, we haven't changed that, but it has been a loooong while since I touched that area

s77rt: @robertjchen Would you be able to check again so we can move forward here?

s77rt:
@bernhardoj Can you please merge main, let's get this to the finish line

bernhardoj:
Done

s77rt:
## Reviewer Checklist

- [X] I have verified the author checklist is complete (all boxes are checked off).
- [X] I verified the correct issue is linked in the `### Fixed Issues` section above
- [X] I verified testing steps are clear and they cover the changes made in this PR
    - [X] I verified the steps for local testing are in the `Tests` section
    - [X] I verified the steps for Staging and/or Production testing are in the `QA steps` section
    - [X] I verified the steps cover any possible failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
- [X] I checked that screenshots or videos are included for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [X] I verified tests pass on **all platforms** & I tested again on:
    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
    - [X] MacOS: Desktop
- [X] If there are any errors in the console that are unrelated to this PR, I either fixed them (preferred) or linked to where I reported them in Slack
- [X] I verified proper code patterns were followed (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [X] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`).
- [X] I verified that this PR follows the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [X] I verified other components that can be impacted by these changes have been tested, and I retested again (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` have been tested & I retested again)
    - [X] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG`)
- [X] I have checked off every checkbox in the PR reviewer checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>



https://github.com/user-attachments/assets/768438fd-e6a6-4711-89cb-99b72c16ce06


</details>

<details>



https://github.com/user-attachments/assets/b4ce7748-fc1f-4adf-bd5c-f9498fc69346


</details>

<details>



https://github.com/user-attachments/assets/a4204e46-5228-4c7b-9cef-0e1cbcf9826f


</details>

<details>



https://github.com/user-attachments/assets/e8428d7a-64ac-4027-8504-f68e50e095db


</details>

<details>



https://github.com/user-attachments/assets/29cd3a0a-a413-40be-b3e0-f1ad8f7d68cc


</details>

<details>



https://github.com/user-attachments/assets/c8bf9483-d96d-44a1-a11e-1df9d46d944d


</details>

s77rt:
Can you check the failing tests

bernhardoj:
Fixed

s77rt (approved):
LGTM but we have this [bug](https://github.com/Expensify/App/pull/49971#discussion_r1822417367) that we should handle separately

src/libs/IOUUtils.ts
robertjchen: Sorry for the delay- I think switching the sign optimistically as you've proposed may be necessary, as there won't be any further changes to the amounts being returned from the BE 😥

robertjchen (dismissed):
Thanks for the patience here, let's run with what we have now and make adjustments later on if necessary. Thanks!! 🙇

robertjchen (dismissed):

Gonals (dismissed):

Gonals:
There are some conflicts

bernhardoj:
@Gonals Fixed

Gonals (approved):

izarutskaya:
@bernhardoj QA Found this issue https://github.com/Expensify/App/issues/53425 when validation the PR 

https://github.com/user-attachments/assets/5d42f5b2-8711-406b-a617-74917d6750f4

Issue #51044, opened 2024-10-17
[$250] [HOLD for payment 2024-11-14] [Search v2.4] - Same search query appears twice in Recent searches


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.0.50-3
**Reproducible in staging?:** Y
**Reproducible in production?:** N
**If this was caught on HybridApp, is this reproducible on New Expensify Standalone?:** N/A
**If this was caught during regression testing, add the test name, ID and link from TestRail:** N/A
**Email or phone of affected tester (no customers):** applausetester+kh081006@applause.expensifail.com
**Issue reported by:** Applause - Internal Team

## Action Performed:
Precondition:
- User has not searched using filters previously
1. Go to staging.new.expensify.com
2. Go to Search
3. Click Filters
4. Apply any filter and click View results
5. Go to Inbox
6. Click on the search icon on the report header
7. Click on the search query under Recent searches
8. Go back to Inbox
9. Click on the search icon on the report header

## Expected Result:
The search query made in Step 4 will only appear once after clicking on the same search query again in Step 7

## Actual Result:
The search query made in Step 4 appears twice after clicking on the same search query again in Step 7

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [x] Android: Standalone
- [ ] Android: HybridApp
- [x] Android: mWeb Chrome
- [ ] iOS: Standalone
- [ ] iOS: HybridApp
- [x] iOS: mWeb Safari
- [x] MacOS: Chrome / Safari
- [x] MacOS: Desktop

## Screenshots/Videos

<details>
  <summary>Add any screenshot/video evidence</summary>

  
</details> 

https://github.com/user-attachments/assets/4721e008-4faa-4216-ab1f-2561b9dd4a03

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021857076660400822469</li>
        <li>Upwork Job ID: 1857076660400822469</li>
        <li>Last Price Increase: 2024-11-14</li>
    </ul>
</details>

[...]

PR #51314, opened 2024-10-23
Fix "same search query appears twice in Recent searches"

SzymczakJ:
### Fixed Issues

$ https://github.com/Expensify/App/issues/51044
PROPOSAL:

### Tests

1. Navigate to the Search page.
2. Click on the search bar icon to open the search field.
3. Enter any search query and press the Enter key to submit the query.
4. On the search results page, change the "Sort By" column contextual filter.
5. On the search results page, change the "Sort Order" contextual filter.
6. On the search results page, use the workspace switcher to select a workspace.
7. Click on the search bar icon again to open the search router.
8. Verify that the search query (entered in step 3) will only appear once in the recent searches section (no duplicate queries)

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>
Screenshot 2024-10-25 at 16 35 48

</details>

<details>

</details>


289Adam289 (approved):
LGTM

SzymczakJ:
Let's move this PR forward @rayane-djouah 🙏 
Do we have some estimate on how long it will take to put a fix on BE side @luacmartins?

rayane-d (changes requested):
@SzymczakJ I think we need to exclude `policyID` from `recentSearchHash` also because searches that differ only by `policyID` appear multiple times in Recent searches:

Screenshot 2024-10-25 at 3 53 35 PM

WDYT?

luacmartins:
Thinking more about it. What if a user explicitly types `policyID:X sortOrder:2`? Should we instead just be hashing the input string instead of parts of the AST? That'd ensure anything the user types differently would be a new recent search

src/libs/SearchQueryUtils.ts
```diff
@@ -228,10 +226,15 @@ function getQueryHash(query: SearchQueryJSON): number {
                 }
                 return -1;
             });
-            orderedQuery += ` ${buildFilterValuesString(key, sortedFilterValues ?? [])}`;
+            orderedQuery += `${buildFilterValuesString(key, sortedFilterValues ?? [])}`;
         });
+    const recentSearchHash = UserUtils.hashText(orderedQuery, 2 ** 32);
```
luacmartins, line 231: Thinking more about it. What if a user explicitly types policyID:X sortOrder:2? Should we instead just be hashing the input string instead of parts of the AST? That'd ensure anything the user types differently would be a new recent search

luacmartins:
>Do we have some estimate on how long it will take to put a fix on BE side @luacmartins?

@SzymczakJ backend PR is in review

src/libs/SearchQueryUtils.ts
SzymczakJ: We cannot hash string input. My argument: when user types **policyID:X sortOrder:asc** and **sortOrder:asc policyID:X**, we will get different hashes even though the query results will be the same.

I think we should exclude policyID like @rayane-djouah said. It may create some confusion when user types policy:X into query, but:
1. We have a policy switcher, which is dedicated way of changing policyID, so typing policyID into SearchRouter query is the worst way to change your policyID
2. We don't even autocomplete policyID key, so how will a user know that such option exists.
3. It's unlikely that a real user will type something like **policyID:31294891892741873**, without the help of autocomplete.

What's more if user selects recentSearch with not the policyID he wanted, he can easily switch it on SearchResults page with policy switcher and it's not such a big problem(just like he switches sortOrder or sortBy). WDYT @rayane-djouah @luacmartins

luacmartins: I think we can go with your current solution for now, but:

>when user types policyID:X sortOrder:asc and sortOrder:asc policyID:X

TBH I'd be ok to show both of these as different recent searches, even though they are the same search at the end of the day. My argument is that the user explicitly typed both of those, so they'd probably expect to see both in the recent search area.

>We don't even autocomplete policyID key, so how will a user know that such option exists.

I think we should introduce policyID autocomplete. Maybe we need a new `workspace:` filter for that though, so the autocomplete shows the policy name instead of ID.

cc @JmillsExpensify @trjExpensify for your thoughts.

JmillsExpensify:
> TBH I'd be ok to show both of these as different recent searches, even though they are the same search at the end of the day. My argument is that the user explicitly typed both of those, so they'd probably expect to see both in the recent search area.

I agree with Carlos on this comment.

JmillsExpensify:
>I think we should introduce policyID autocomplete. Maybe we need a new workspace: filter for that though, so the autocomplete shows the policy name instead of ID. 

For this one, we should be careful as I'm imagining that we'd support the ability to search on multiple policies at once right? If so, then we need to introduce a way for the front-end to do the same in the workspace switcher.

luacmartins:
@SzymczakJ so let's update the hash to be on the input query so we always show whatever users type, regardless of it being the same query as a previous one (if the only difference is the order of filters)

luacmartins (changes requested):
@SzymczakJ so let's update the hash to be on the input query so we always show whatever users type, regardless of it being the same query as a previous one (if the only difference is the order of filters)

SzymczakJ:
But using inputQuery as a hash input will not fix this bug https://github.com/Expensify/App/issues/51044! 
For example when user types in two queries differing in only sortOrder like **category:Advertising sortOrder:asc** and **category:Advertising sortOrder:desc** in SearchRouter then we will get recentSearches list with two queries that look the same, because we don't show sortOrder there. Honestly I'm starting to thing there's no good solution for this case, everything has it's pros and cons 😭

SzymczakJ:
> it would solve the bug unless users manually type sortBy or sortOrder, no?

Thats actually the main root cause of the bug, what's more there will be also a bug when user changes sortOrder on search results page, which is highly probable.

I would actually go with the current solution where:
- we don't have a bug that happens when user manually types sortBy or sortOrder
- we don't have bug when user changes sortOrder on search results page

Like you said later we could split this into two cases:
1. use current solution for recentSearches that are derived from using advanced filters page
2. save what users actually types into search router and in that case compute recentSearchesHash on users input, so the user could see exact queries that he typed, like **category:X sortOrder:A**

luacmartins:
I'm ok using the current logic as a stepping stone, but I think ultimately we're aiming for the behavior below, so we'll need to work on a follow up for that.

>save what users actually types into search router and in that case compute recentSearchesHash on users input, so the user could see exact queries that he typed, like category:X sortOrder:A

SzymczakJ:
I agree. So we need to store on BE what user typed into the search router and return it in recentSearches exactly as it was, but only in case, when it was made by SearchRouter

luacmartins:
Correct. We currently store the `inputQuery` param, I'm not sure if we'd want to reuse that param or create a new one for it. The intent behind `inputQuery` was to be exactly what the user typed, but that doesn't seem to be the case anymroe.

luacmartins (dismissed):
LGTM

SzymczakJ:
FYI I'll OOO on Friday and Monday.

rayane-d (changes requested):
BUG: I'm still seeing duplicate recent searches that differ only by `sortOrder`, the `recentSearchHash` is changing when you change the sorting order:

Screenshot 2024-10-31 at 10 38 15 PM

Screenshot 2024-10-31 at 10 38 05 PM



https://github.com/user-attachments/assets/7aeef525-da44-4400-9f1d-4d03bb93752e

rayane-d (changes requested):

289Adam289:
Hi! I was also OOO on Friday. @SzymczakJ will be back tomorrow and I think it would be best to leave addressing comments to him.

src/libs/SearchQueryUtils.ts
```diff
@@ -216,11 +216,8 @@ function findIDFromDisplayValue(filterName: ValueOf<typeof CONST.SEARCH.SYNTAX_F
...
  */
-function getQueryHash(query: SearchQueryJSON): number {
+function getQueryHashes(query: SearchQueryJSON): {primaryHash: number; recentSearchHash: number} {
     let orderedQuery = '';
-    if (query.policyID) {
-        orderedQuery += `${CONST.SEARCH.SYNTAX_ROOT_KEYS.POLICY_ID}:${query.policyID} `;
-    }
     orderedQuery += `${CONST.SEARCH.SYNTAX_ROOT_KEYS.TYPE}:${query.type}`;
     orderedQuery += ` ${CONST.SEARCH.SYNTAX_ROOT_KEYS.STATUS}:${query.status}`;
     orderedQuery += ` ${CONST.SEARCH.SYNTAX_ROOT_KEYS.SORT_BY}:${query.sortBy}`;
```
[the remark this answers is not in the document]
SzymczakJ: Whoops, that's probably my mistake when merging main 🙇

SzymczakJ:
I've added the tests steps and fixed bug you mentioned @rayane-djouah

rayane-d (approved):
LGTM and tests well 👍

luacmartins (approved):
LGTM

PR #51595, opened 2024-10-28
Revert "Fix exit survey response page validation runs when open"

neil-marcellini:
Reverts Expensify/App#51516. Straight revert because it seems to be [causing tests to fail](https://github.com/Expensify/App/issues/51586) when run in the GH action.


neil-marcellini:
Oh wow. Ok, then this must not be the cause of that test failing. I'll close this.

PR #49226, opened 2024-09-15
fix: app goes to workspace initial page when back from not found page

dominictb:
### Fixed Issues

$ https://github.com/Expensify/App/issues/46646
PROPOSAL: https://github.com/Expensify/App/issues/46646#issuecomment-2273017688

### Tests

[Native only]

1. Send a workspace settings link with invalid ID in a report (e.g., `/settings/workspaces/1/more-features`)
2. Open the link
3. Swipe back from not found page (Do not press Back button)

Verify app goes back to workspaces list page

4. Open the link in step 1 by deep linking
5. Swipe back from not found page (Do not press Back button)

Verify app goes back to workspaces list page

- [x] Verify that no errors appear in the JS console

### QA Steps

[Native only]

1. Send a workspace settings link with invalid ID in a report (e.g., `/settings/workspaces/1/more-features`)
2. Open the link
3. Swipe back from not found page (Do not press Back button)

Verify app goes back to workspaces list page

4. Open the link in step 1 by deep linking
5. Swipe back from not found page (Do not press Back button)

Verify app goes back to workspaces list page

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
[...]

PR #52299, opened 2024-11-11
[Test] Revert "fix: app goes to workspace initial page when back from not found page"

shubham1206agra:
Reverts Expensify/App#49226


lakchote:
This PR is not needed as the PR is working correctly.

See https://github.com/Expensify/App/pull/49226#issuecomment-2467509982

Issue #54877, opened 2025-01-07
Include yapl build step within local build script


Julesssss:
### Problem 

YAPL isn't built correctly for newly cloned HybridApp repos. This is a bug that slows down engineers setup process ([example](https://expensify.slack.com/archives/C01GTK53T8Q/p1734700081816189?thread_ts=1734638630.180839&cid=C01GTK53T8Q)).

![Screenshot 2025-01-07 at 12 54 21](https://github.com/user-attachments/assets/062468b6-ba4c-4945-9909-1cd80bc45251)

### Solution

- Include `npm run grunt:build:shared` in the local build script
- Remove the now-outdated instructions from the App README: '
<img width="875" alt="Screenshot 2025-01-07 at 16 09 19" src="https://github.com/user-attachments/assets/1637686e-00da-4355-9122-663404a84e5e" />
- Remove the step from the workflows, as this is now redundant

_FYI @staszekscp, this is a minor HybridApp build issue that can be worked on when your team has some availability_

Julesssss:
Submitted a better fix here in the Gradle script: https://github.com/Expensify/Mobile-Expensify/pull/13376

Julesssss:
Reopening for the final step: Remove the step from the workflows, as this is now redundant

Julesssss:
Cleaning up the readme is being done in this PR: https://github.com/Expensify/App/pull/55770

Julesssss:
Not a priority

Julesssss:
Done

PR #55426, opened 2025-01-17
[NO QA] Add YAPL build step to HybridApp builds

Julesssss:
### Explanation of Change

Engineers building the Android HybridApp for the first time see a failure due to yapl code not being built. They they need to search slack and manually do this which is confusing and time-consuming. This change ensures it is done automatically.

Then I deleted step 3 from the HybridApp build instructions.

_I should be able to remove the manual step in our build workflows too, but I'll do that separately._

### Fixed Issues

https://github.com/Expensify/App/issues/54877

### Tests

**A) Script**
- `grunt build: shared` should run when an Android build is started
- Screenshot 2025-01-17 at 11 10 23

**B) Build**
- Clone a fresh App repo
- Clone the Mobile-Expensify submodule
- Build the android app
    - `cd app`
    - `git submodule update --init --progress --depth 100` 
    - `npm i`
    - `npm run android`
- It should build, and the app should not crash on start!

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "postinstall": "./scripts/postInstall.sh",
     "clean": "./scripts/clean.sh",
     "clean-standalone": "STANDALONE_NEW_DOT=true ./scripts/clean.sh",
-    "android": "./scripts/set-pusher-suffix.sh && npm run grunt:build:shared && ./scripts/run-build.sh --android",
+    "android": "./scripts/set-pusher-suffix.sh && cd Mobile-Expensify && npm run grunt:build:shared && ../scripts/run-build.sh --android",
     "android-standalone": "./scripts/set-pusher-suffix.sh && STANDALONE_NEW_DOT=true ./scripts/run-build.sh --android",
     "ios": "./scripts/set-pusher-suffix.sh && ./scripts/run-build.sh --ios",
     "ios-standalone": "./scripts/set-pusher-suffix.sh && STANDALONE_NEW_DOT=true ./scripts/run-build.sh --ios",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "postinstall": "./scripts/postInstall.sh",
     "clean": "./scripts/clean.sh",
     "clean-standalone": "STANDALONE_NEW_DOT=true ./scripts/clean.sh",
-    "android": "./scripts/set-pusher-suffix.sh && ./scripts/run-build.sh --android",
+    "android": "./scripts/set-pusher-suffix.sh && npm run grunt:build:shared && ./scripts/run-build.sh --android",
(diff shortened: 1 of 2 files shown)


Julesssss:
I dm'd Mariusz for a review too. I'm not sure if a separate script is overkill or good practice for keeping these scripts tidy. Let me know what you think!

package.json
```diff
@@ -15,7 +15,7 @@
     "postinstall": "./scripts/postInstall.sh",
     "clean": "./scripts/clean.sh",
     "clean-standalone": "STANDALONE_NEW_DOT=true ./scripts/clean.sh",
-    "android": "./scripts/set-pusher-suffix.sh && ./scripts/run-build.sh --android",
+    "android": "./scripts/set-pusher-suffix.sh && cd Mobile-Expensify && npm run grunt:build:shared && cd .. && ./scripts/run-build.sh --android",
```
AndrewGable, line 18: Don't we need to run this on iOS too?

Julesssss: I'm not sure. I saw [this](https://github.com/Expensify/Mobile-Expensify/blob/262a4953a9f40be13b7d568c86418f71b96e203c/iOS/Expensify.xcodeproj/project.pbxproj#L3462-L3479) iOS build step which made me think no... and we only added the step to the Android workflows. Not 100% sure though

AndrewGable: I _think_ this should be running via grunt, but for whatever reason it's not working and engineers miss the error.

AndrewGable: https://github.com/Expensify/Mobile-Expensify/blob/main/Android/build.gradle#L405-L408

Julesssss: Oooh that would be better. Perhaps our current grade command no longer matches [this](https://github.com/Expensify/Mobile-Expensify/blob/0029d3f85e87c1c84aa290d41bfe7c1861bebaca/Android/build.gradle#L385) logic: 

`theTask.name.matches("(assemble|bundle).*(Debug|Adhoc|Release)")`

staszekscp: I think that cleanest solution would be to include the task to the gradle script. Now I think there is a chance that the `React Native CLI` may interfere into the build phases, or indeed change it in some way so the `buildGrunt` doesn't trigger.

Maybe changing the pattern to `(install|assemble|bundle).*(Debug|Adhoc|Release)` would be enough?

Julesssss: Agree this is a much better fix. The suggestion appears to work from my testing: 

- Manual Generate Signed APK
    - Screenshot 2025-01-21 at 11 42 33

- `npm run android`
    - Screenshot 2025-01-21 at 11 51 32

Julesssss:
Replacing these App changes with a better Gradle fix here: https://github.com/Expensify/Mobile-Expensify/pull/13376

PR #55664, opened 2025-01-23
Revert "Fix new messages indicator after comment linking"

neil-marcellini:
Reverts Expensify/App#55582


neil-marcellini:
The root cause of the bug was not this PR

PR #59807, opened 2025-04-08
Make canUseTableReportView always return false

JakubKorytko:
# This PR draft exists only for the purpose of simple ad hoc runs on old-view. It should not be reviewed, opened, merged, or closed.


JakubKorytko:
@shawnborton

PR #59840, opened 2025-04-08
[CP Staging] Revert "Fix no gbr/rbr shown when reopen the app"

mountiny:
Reverts Expensify/App#58652

Straight revert for performance reasons - this is causing iteration over the entire Reports collection that in case of heavy users can be 10k+ reports and slows down the entire app https://expensify.slack.com/archives/C05LX9D6E07/p1744125477572759?thread_ts=1743598259.881369&cid=C05LX9D6E07


mountiny:
False alarm, you can read more in the linked thread

Issue #9402, opened 2022-06-11
[Due for payment 2026-04-16] [Image][$1000] Improve NewDot image uploading experience, add large file support


chiragsalian:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Upload an image in App

## Expected Result:
The image should be grayed (opacity 0.6 i.e., chatItemUnsentMessage) until its uploaded and once its uploaded it should not show the white thumbnail with loader
![image](https://user-images.githubusercontent.com/14356145/173163988-2ba7c6fe-f77c-4e68-bfdf-1510ad5646c3.png)
and instead continue showing the local thumbnail or show the final image immediately.

## Actual Result:
1. The local image is being uploaded and we see the local thumbnail,
![image](https://user-images.githubusercontent.com/14356145/173164047-79f28f8a-1ec7-4434-a038-787a2974492f.png)
2. The local image has finished uploading and now we attempt to render it which first shows up a white thumbnail with loader
![image](https://user-images.githubusercontent.com/14356145/173164076-f91ce915-87a2-4653-8284-d067baeae777.png)
3. The render completes and we see the uploaded image
![image](https://user-images.githubusercontent.com/14356145/173164089-9327bdb1-0825-4ed6-9e64-650c580ab4c9.png)

## Workaround:
Can the user still use Expensify without this being fixed? Have you informed them of the workaround?

## Platform:

- Web
- iOS
- Android
- Desktop App
- Mobile Web

**Version Number:** 1.1.76
**Reproducible in staging?:** Y
**Reproducible in production?:** Y
**Slack conversation:** https://expensify.slack.com/archives/C01GTK53T8Q/p1654262165566879

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

cc @michaelhaxhiu, @cead22, @kidroca.

parasharrajat:
Good change. I agree.

parasharrajat:
@JosueEchandiaAsto Thanks for the proposal. But as you are a new contributor and have already been assigned two 2 tasks, you will have to wait before applying to new tasks. 

Contribution policy: https://github.com/Expensify/App/blob/main/CONTRIBUTING.md#payment-for-contributions
[...]

PR #63723, opened 2025-06-07
(1/2) Implement store/caching for attachments

NJ-2020:
### Explanation of Change
Store/Cache attachment when uploading a file or markdown text link attachment i.e ``
### Fixed Issues

$ https://github.com/Expensify/App/issues/9402
PROPOSAL: https://github.com/Expensify/App/issues/9402#issuecomment-2570003952

### Tests
Same as QA

### Offline tests
Same as QA

### QA Steps

Prerequisites: DEV mode only

1. Open Expensify App
2. Go to any chat 
3. Open inspect element and go to > Applications tab <br/>
**File Upload** 
        1. Open attachment picker
        2. Select any file
        3. Go to CacheAPI/Onyx storage for web and observe the changes <br/>
**Markdown text link attachment** 
        1. Go to input chat
        2. Type any markdown text link attachment e.g `` and send it
        3. Go to CacheAPI (web-desktop only) or Onyx (attachment_{id}) storage and observe the changes
### PR Author Checklist
    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
    - [X] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/f83b63aa-6b4d-47ce-9db4-87ba07689d04

</details>

<details>

https://github.com/user-attachments/assets/88def232-e75f-480f-8961-e69ebe6fb21f

</details>

<details>

https://github.com/user-attachments/assets/0bf57c2b-cdd9-4593-9c29-17e53108fa2e

</details>

<details>

I have an issue while inspecting the web)

</details>

<details>

https://github.com/user-attachments/assets/a8dca94f-54c4-405b-9714-9d99acd37789

</details>

<details>

https://github.com/user-attachments/assets/61978070-c7a5-47a5-a6d8-96cfdb39e8ee

</details>


src/components/AttachmentPicker/index.native.tsx
```diff
@@ -135,7 +135,7 @@ function AttachmentPicker({
     const showImagePicker = useCallback(
         (imagePickerFunc: (options: CameraOptions, callback: Callback) => Promise<ImagePickerResponse>): Promise<Asset[] | void> =>
             new Promise((resolve, reject) => {
-                imagePickerFunc(getImagePickerOptions(type, fileLimit), (response: ImagePickerResponse) => {
+                imagePickerFunc(getImagePickerOptions(type, fileLimit), async (response: ImagePickerResponse) => {
```
parasharrajat: We don't allow async await.

src/components/AttachmentPicker/index.native.tsx
```diff
@@ -153,9 +153,31 @@ function AttachmentPicker({
...
+                        }
+
+                        return {
+                            ...(response.assets?.at(index) ?? {}),
+                            uri: localCopy.localUri,
+                        };
+                    });
+
+                    const targetAsset = assets.at(0);
+                    const targetAssetUri = targetAsset?.uri;
```
parasharrajat: We not need to cache the attachment while the user is adding one. Let's only worry about the message attachments for now.

src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx
```diff
@@ -66,7 +66,9 @@ function ImageRenderer({tnode}: ImageRendererProps) {
     // For other image formats, we retain the thumbnail as is to avoid unnecessary modifications.
     const processedPreviewSource = typeof previewSource === 'string' ? previewSource.replace(/\.png\.(1024|320)\.jpg$/, '.png') : previewSource;
     const source = tryResolveUrlFromApiRoot(isAttachmentOrReceipt ? attachmentSourceAttribute : htmlAttribs.src);
-
+    const imageSource = processedPreviewSource;
+    // const isAuthTokenRequired = isLocalFile(imageSource) ? false : isAttachmentOrReceipt;
+    const isAuthTokenRequired = isAttachmentOrReceipt;
```
parasharrajat, line 71: Didn't understand this change?

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+    waitForCollectionCallback: true,
+    callback: (value) => (attachments = value ?? {}),
+});
+
+function storeAttachment(attachmentID: string, uri: string) {
+    if (!attachmentID || !uri) {
+        return;
+    }
+    if (uri.startsWith('file://')) {
+        Onyx.set(`${ONYXKEYS.COLLECTION.ATTACHMENT}${attachmentID}`, {
```
parasharrajat, line 16: Let's convert this collection to store attacments into a a collection for report, not per attachment basis. 

So `${ONYXKEYS.COLLECTION.ATTACHMENT}${reportId}. This way we can subscribe this key in the report screen and get the source for any attachment otherwise, we will have to subscribe to each attachment separately.

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+        return;
+    }
+
+    const attachment = attachments?.[attachmentID];
+
+    if (attachment?.source && attachment.remoteSource === uri) {
+        return;
+    }
+
+    RNFetchBlob.config({fileCache: true, path: `${RNFetchBlob.fs.dirs.DocumentDir}/${attachmentID}`})
```
parasharrajat, line 29: Not a good idea to store in DocumentDir as it can not be cleared but system cleanup. We should instead create a subfolder in the CacheDir and save attachments there. 

When we are clearly the cacheDir on signout. We exclude this folder from that cleanup.

src/libs/actions/Report.ts
```diff
@@ -717,6 +745,10 @@ function addActions(reportID: string, text = '', file?: FileObject) {
         parameters.isOldDotConciergeChat = true;
     }
 
+    if (file) {
+        parameters.attachmentID = attachmentID;
+    }
+
```
parasharrajat, line 761: Why?

src/types/onyx/Attachment.tsx
```diff
@@ -0,0 +1,9 @@
+type Attachment = {
+    attachmentID: string;
+
+    source?: string;
+
+    remoteSource?: string;
```
parasharrajat: Missing properties comment.

src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx
NJ-2020: Ohh sorry, this is because when rendering attachment specifically using local uri, its failing to load the attachment, it's because we are passing the isAuthTokenRequired true for local uri

src/libs/actions/Attachment/index.native.ts
NJ-2020: It's because when a user upload an attachment,  if we save the file into cache dir, the attachment preview will gone if the user restart the app/device

So to fix this we need to store the attachment into documents directory i.e `com.expensify.com/files/attachment_{id}`

We will recache the attachment if there's new changes from BE, if not then we will use from the pre-cached attachment

src/libs/actions/Report.ts
NJ-2020: We need to pass this to get the `data-attachment-id` attribute inside the attachment tag i.e ``

For markdown text link attachments, we will use the reportActionID + index

src/components/AttachmentPicker/index.native.tsx
NJ-2020: We're not caching it, this for uploading attachment from Attachment picker specifically from `Choose from Gallery` where the files are copied to cache directories but we need to copy it into document directories

More info: https://github.com/Expensify/App/pull/63723#discussion_r2135666656

src/libs/actions/Attachment/index.native.ts
parasharrajat: I don't think we need to worry about the attachment preview when the user is uploading the attachment. Until it is sent, we don't have to save it for offline preview. If needed we would do that later in different PR.

parasharrajat (changes requested):

src/libs/actions/Report.ts
parasharrajat: > We need to pass this to get the data-attachment-id attribute inside the attachment tag i.e 

Why do we need to do this? Does it handle the optimistic case?

src/libs/actions/Attachment/index.native.ts
parasharrajat: Please let me know how it looks first. Does this increase complexity and impact performance? 

Overall, I believe we should create a custom hook to handle attachments. I am particularly curious on how you will handle the remote source change from backend. Can you show me code on how you will handle it?

src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx
parasharrajat: > const isAuthTokenRequired = isLocalFile(imageSource) ? false

But this says false for local file.

src/components/AttachmentPicker/index.native.tsx
parasharrajat: Replied on that linked thread. I think this is kind of out of scope of this issue.

parasharrajat:
@NJ-2020 Thanks for your work on this, and remain committed while we were discussing it. Can you please speed up the implementation? I believe we will have to go through a few iterations before we complete this task. Thus, it is important to update daily.

src/libs/actions/Report.ts
NJ-2020: It's used for storing and retrieving the attachment from Onyx/CacheAPI by using the data-attachment-id `attachment_{idHere}`

src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx
NJ-2020: Yes, right we need to pass false for local uri

src/libs/actions/Attachment/index.native.ts
NJ-2020: > I don't think we need to worry about the attachment preview when the user is uploading the attachment. Until it is sent

Agree, so I think we can do it inside the `storeAttachment` function

> we don't have to save it for offline preview

I think it will show weird behavior to the user if the user close and then reopen the app, because it will show empty content(offline icon) for all optimistic attachments

Wdyt?

NJ-2020:
Okay, sure

src/libs/actions/Attachment/index.native.ts
NJ-2020: > Let's convert this collection to store attacments into a a collection for report, not per attachment basis.

I think we're using the data-attachment-id attribute value from the attachment tag for both storing/retrieving attachment

Basically if there's any attachment it will render through ImageRenderer|VideoRenderer|AnchorRenderer, inside that we can retrieve the data-attachment-id attribute value provided from BE and use the attachment id value to get the local source from Onyx

NJ-2020: Maybe for alternative option instead of passing reportID we can use the reportActionID which is more easier to handle delete attachment, retrieving and others

But I think we can use the data-attachment-id from BE, which is specifically for attachments. For old attachments without attachment IDs, we can use reportActionID + {index}.

> Please let me know how it looks first. Does this increase complexity and impact performance?

I am not sure if it's really impacting our app perfomance, because for every attachment we need to get the list of attachments of the current report(which can be more than 1), but we only need 1 attachment source from all the lists and we need to filter out again that matches with the data-attachment-id value 

Wdyt?

NJ-2020: > Overall, I believe we should create a custom hook to handle attachments. I am particularly curious on how you will handle the remote source change from backend. Can you show me code on how you will handle it?

Currently, I handle it inside the `getAttachmentSource` function, where first I compared the current source from BE with `attachment.remoteSource`, if there's any differences, then we will recache the attachment and show the new one

https://github.com/Expensify/App/blob/21f59ee07241fad2e55342e89c2595ceaf0d9b08/src/libs/actions/Attachment/index.native.ts#L47-L56

<hr />

**Edit:**
Ah, you're right I think we need to make new hook for retrieving attachment specifically for CacheAPI i.e like useOnyx, because it's returning Promise value

But I think I'll do it in the second phase of the PR, [right](https://github.com/Expensify/App/issues/9402#issuecomment-2939845294)?

src/components/AttachmentPicker/index.native.tsx
NJ-2020: I am still waiting for this comment: https://github.com/Expensify/App/pull/63723#discussion_r2136827660

src/types/onyx/Attachment.tsx
NJ-2020: Done

src/libs/actions/Attachment/index.native.ts
parasharrajat: When you close the app on add attachment modal, I think user will have to reopen the flow. Anyways, as I said, we will revisit this after we are done saving attachments in different PR, if needed. 


Let's break it down to smaller PRs to faster completion.

src/libs/actions/Attachment/index.native.ts
parasharrajat: yeah, second PR sounds good,

src/libs/actions/Attachment/index.native.ts
NJ-2020: Okay, I will do this in the next PR

src/components/AttachmentPicker/index.native.tsx
NJ-2020: Reverting this changes due to this comment

Will fix it in the next [PR](https://github.com/Expensify/App/pull/63723#discussion_r2137272950)

parasharrajat:
Please let me know when this is ready

NJ-2020:
Sure

NJ-2020:
PR will be ready tomorrow

I am still working on the unit test file

NJ-2020:
There's some issues while creating the unit test file
Will share an update when it's done

NJ-2020:
PR is ready

cc: @parasharrajat

src/libs/CacheAPI/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        return;
+    }
+    // Exit early if the Cache API is not supported in the current browser.
+    if (!('caches' in window)) {
+        throw new Error('Cache API is not supported');
+    }
+    keys.forEach((key) => {
+        if (!key) {
+            return;
+        }
```
parasharrajat: throw an error instead if key is not from passed `CACHE_API_KEYS` in `init`

src/libs/CacheAPI/index.ts
```diff
@@ -0,0 +1,67 @@
...
+                return;
+            }
+            caches.open(key);
+        });
+    });
+}
+function put(cacheName: string, key: string, value: Response) {
+    if (!cacheName || !key || !value) {
+        return;
+    }
```
parasharrajat: We don't have to check for these when all of them are required params. Please define the type in in App `Response`.

parasharrajat: Also, add a logic that key is from `CACHE_API_KEYS` as this will allow overriding the settings passed in init.

src/libs/CacheAPI/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    caches.open(cacheName).then((cache) => {
+        cache.delete(key);
+    });
+}
+function clear(keys: string[]) {
+    if (!keys || keys.length === 0) {
+        return;
+    }
```
parasharrajat: we should not pass any keys here instead they should be taken from the keys passed in `init`.

src/libs/CacheAPI/index.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+function get(cacheName: string, key: string) {
+    if (!cacheName || !key) {
+        return;
+    }
+    return caches.open(cacheName).then((cache) => {
+        return cache.match(key);
+    });
+}
+function remove(cacheName: string, key: string) {
```
parasharrajat, line 35: What is the `cacheName`?

NJ-2020:
Bump @parasharrajat for review

src/components/AttachmentPicker/index.native.tsx
```diff
@@ -213,7 +213,7 @@ function AttachmentPicker({
                     fileName: file.name ?? '',
                 };
             }) as [FileToCopy, ...FileToCopy[]],
-            destination: 'cachesDirectory',
+            destination: 'documentDirectory',
```
parasharrajat, line 216: Do we need this change now?

src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx
parasharrajat: But the previous logic was doing the same so why did you change it?

src/libs/ReportActionsUtils.ts
```diff
@@ -170,12 +171,14 @@ function isDeletedAction(reportAction: OnyxInputOrEntry<ReportAction | Optimisti
  * appears inside the report action message so as to identify attachments with identical source inside a report action.
  */
 function getHtmlWithAttachmentID(html: string, reportActionID: string | undefined) {
-    if (!reportActionID) {
+    const {sourceURL} = getAttachmentDetails(html);
+    const attachmentID = sourceURL?.match(CONST.REGEX.ATTACHMENT_SOURCE_ID)?.[1];
```
parasharrajat: Can you share some examples for `ATTACHMENT_SOURCE_ID`?

src/libs/ReportActionsUtils.ts
```diff
@@ -170,12 +171,14 @@ function isDeletedAction(reportAction: OnyxInputOrEntry<ReportAction | Optimisti
  * appears inside the report action message so as to identify attachments with identical source inside a report action.
  */
 function getHtmlWithAttachmentID(html: string, reportActionID: string | undefined) {
-    if (!reportActionID) {
```
parasharrajat: This check was telling that `reportActionID` can be null and it is used at line 181. Can we sure that it is always present?

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+Onyx.connect({
+    key: ONYXKEYS.COLLECTION.ATTACHMENT,
+    waitForCollectionCallback: true,
+    callback: (value) => (attachments = value ?? {}),
+});
+
+function storeAttachment(attachmentID: string, uri: string) {
+    if (!attachmentID || !uri) {
+        return;
+    }
```
parasharrajat: Can these be null when a parameter is required? Let's remove these checks.

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+import type {Attachment} from '@src/types/onyx';
+
+let attachments: OnyxCollection<Attachment> | undefined;
+Onyx.connect({
+    key: ONYXKEYS.COLLECTION.ATTACHMENT,
+    waitForCollectionCallback: true,
+    callback: (value) => (attachments = value ?? {}),
+});
+
+function storeAttachment(attachmentID: string, uri: string) {
```
parasharrajat:
```suggestion
function cacheAttachment(attachmentID: string, uri: string) {
```

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+                source: `file://${filePath}`,
+                remoteSource: uri,
+            });
+        })
+        .catch(() => {
+            throw new Error('Failed to store attachment');
+        });
+}
+
+function getAttachmentSource(attachmentID: string, currentSource: string) {
```
parasharrajat:
```suggestion
function getCachedAttachment(attachmentID: string, currentSource: string) {
```

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!attachmentID) {
+        return;
+    }
```
parasharrajat: Same remove.

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+    const attachment = attachments?.[`${ONYXKEYS.COLLECTION.ATTACHMENT}${attachmentID}`];
+
+    if (attachment?.remoteSource && attachment.remoteSource !== currentSource) {
+        storeAttachment(attachmentID, currentSource);
+        return currentSource;
```
parasharrajat, line 49: Wouldn't this be remote resource in some cases? We should always return cached resource?

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+    return attachment?.source;
+}
+
+function deleteAttachment(attachmentID: string) {
```
parasharrajat:
```suggestion
function removeCachedAttachment(attachmentID: string) {
```

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!attachmentID) {
+        return;
+    }
```
parasharrajat: Same. Either we make param optional or remove these checks.

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
...
+import type {OnyxCollection} from 'react-native-onyx';
+import ONYXKEYS from '@src/ONYXKEYS';
+import type {Attachment} from '@src/types/onyx';
+
+let attachments: OnyxCollection<Attachment> | undefined;
+Onyx.connect({
+    key: ONYXKEYS.COLLECTION.ATTACHMENT,
+    waitForCollectionCallback: true,
+    callback: (value) => (attachments = value ?? {}),
+});
```
parasharrajat, line 12: This is fine here but we should move to hook for up-to-date state.

src/libs/actions/Attachment/index.native.ts
```diff
@@ -0,0 +1,67 @@
+import RNFetchBlob from 'react-native-blob-util';
+import Onyx from 'react-native-onyx';
+import type {OnyxCollection} from 'react-native-onyx';
+import ONYXKEYS from '@src/ONYXKEYS';
+import type {Attachment} from '@src/types/onyx';
+
+let attachments: OnyxCollection<Attachment> | undefined;
```
parasharrajat: this can never be undefined as per your code.

src/libs/actions/Attachment/index.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return response
+            .blob()
+            .then((attachmentFile) => {
+                const source = URL.createObjectURL(attachmentFile);
+                return source;
+            })
+            .catch(() => {
+                throw new Error('Failed to get attachment');
+            });
```
parasharrajat, line 53: mismatched return types between native and this one. Both should return same type.

src/libs/actions/Attachment/index.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function deleteAttachment(attachmentID: string) {
+    if (!attachmentID) {
+        return;
+    }
+
+    Onyx.set(`${ONYXKEYS.COLLECTION.ATTACHMENT}${attachmentID}`, null);
+    CacheAPI.remove(CONST.CACHE_API_KEYS.ATTACHMENTS, attachmentID);
+}
```
parasharrajat, line 60: Please make similiar changes as suggested on native file.

src/libs/actions/Report.ts
```diff
@@ -660,8 +662,9 @@ function addActions(reportID: string, text = '', file?: FileObject) {
         // When we are adding an attachment we will call AddAttachment.
         // It supports sending an attachment with an optional comment and AddComment supports adding a single text comment only.
         commandName = WRITE_COMMANDS.ADD_ATTACHMENT;
-        const attachment = buildOptimisticAddCommentReportAction(text, file, undefined, undefined, undefined, reportID);
+        const attachment = buildOptimisticAddCommentReportAction(text, file, undefined, undefined, undefined, reportID, undefined, attachmentID);
```
parasharrajat: Let's change the param type of `buildOptimisticAddCommentReportAction` to object to clean this. Shouldn't be a big change?

src/libs/actions/Report.ts
```diff
@@ -672,6 +675,31 @@ function addActions(reportID: string, text = '', file?: FileObject) {
         commandName = WRITE_COMMANDS.ADD_TEXT_AND_ATTACHMENT;
     }
 
+    // Store all markdown text attachments i.e `![](https://images.unsplash.com/...)`
```
parasharrajat, line 688: Very strange. IMO, these must have been parsed and then sent as files to this function so we don't have to do this.  Did you check it?

tests/unit/AttachmentStorageTest.ts
```diff
@@ -0,0 +1,113 @@
...
+    beforeAll(() => {
+        Onyx.init({
+            keys: ONYXKEYS,
+        });
+    });
+    beforeEach(() => {
+        global.fetch = TestHelper.getGlobalFetchMock();
+    });
+
+    it('should store for file in Onyx', async () => {
```
parasharrajat: Add tests for: 
1. Delete the attachment. 
2. Change of remote source should reflect the change in local source. 
3. After attachment is stored, you should confirm whether it was cached or not. Have tests for both Native and web.

tests/unit/AttachmentStorageTest.ts
parasharrajat, line 1: We follow strict naming, so please move this to tests/actions/AttachmentTest.ts if you action file is called `Attachment`.

src/types/onyx/Attachment.tsx
```diff
@@ -0,0 +1,12 @@
+type Attachment = {
+    /** Attachment ID of the attachment */
+    attachmentID: string;
+
+    /** Source url of the attachment either can be local or remote url */
+    source?: string;
```
parasharrajat, line 6: Why would this source be remote? Should the `remoteSource` be remote? we should make sure this property always reflect local source and remoteSource remote.

src/libs/actions/Session/index.ts
```diff
@@ -892,6 +893,7 @@ function cleanupSession() {
     NetworkConnection.clearReconnectionCallbacks();
     SessionUtils.resetDidUserLogInDuringSession();
     resetHomeRouteParams();
+    CacheAPI.clear([CONST.CACHE_API_KEYS.ATTACHMENTS]);
```
parasharrajat: Could this slow down the signout process when there are many resources to be removed?

src/libs/actions/Report.ts
```diff
@@ -672,6 +675,31 @@ function addActions(reportID: string, text = '', file?: FileObject) {
...
+        }
+
+        return {
+            uri: source,
+            attachmentID: dataAttachmentID ?? `${reportActionID}_${index + 1}`,
+        };
+    });
+
+    attachments.forEach((attachment) => {
+        storeAttachment(attachment.attachmentID, attachment.uri ?? '');
```
parasharrajat: There is a problem with code, what if the message failed to send the image will still be cached. we should follow optimistic patterns here. 

you should update Onyx store values in `optimisticData`, `successData` and `failureData`. Same for delete attachment. Now this can be little tricky let's see what options do we have. 

We can do auto cleanup and save when a attachment key is updated in Onyx. 
we have a subscription to attachment collection which will update whenever we make changes to it so when you add a new attachment with remoteSource and attachmentID, we can update the cache to save the remote source and same for delete, we can remove the cache. This way cache will always be in sync with Onyx. This is just an idea and I am not sure how fast would it be. Please suggest yours also.

That's why I was thinking we save attachments report basis so we can subscribe to only attachment of visible report.

parasharrajat (changes requested):

src/setup/index.ts
```diff
@@ -52,6 +53,10 @@ export default function () {
         skippableCollectionMemberIDs: CONST.SKIPPABLE_COLLECTION_MEMBER_IDS,
     });
 
+    const CacheAPIKeys = Object.values(CONST.CACHE_API_KEYS);
+
+    CacheAPI.init(CacheAPIKeys);
```
parasharrajat: let's pass `CONST.CACHE_API_KEYS` it array to init function to pass `CONST.CACHE_API_KEYS` directly this function and then handle getting keys inside the function.

src/CONST/index.ts
```diff
@@ -3582,7 +3582,10 @@ const CONST = {
         AFTER_FIRST_LINE_BREAK: /\n.*/g,
         LINE_BREAK: /\r\n|\r|\n|\u2028/g,
         CODE_2FA: /^\d{6}$/,
-        ATTACHMENT_ID: /chat-attachments\/(\d+)/,
+        ATTACHMENT: /<(img|video)[^>]*>/gi,
+        ATTACHMENT_ID: /data-attachment-id=(["'])(.*?)\1/,
+        ATTACHMENT_SOURCE_ID: /chat-attachments\/(\d+)/,
+        ATTACHMENT_SOURCE: /(src|data-expensify-source|data-optimistic-src)="([^"]+)"/i,
```
parasharrajat: This is somewhat confusing. Can you please add comments to explain use of each regex.

NJ-2020: Done

src/setup/index.ts
NJ-2020: Done

src/types/onyx/Attachment.tsx
NJ-2020: Sorry may bad, done, fixed

tests/unit/AttachmentStorageTest.ts
NJ-2020: Okay, sure

tests/unit/AttachmentStorageTest.ts
NJ-2020: Done

src/components/AttachmentPicker/index.native.tsx
NJ-2020: Yes, because the caches directory will automatically cleared out when the app is closed which prevents the user to see the attachment when re-opening the app again

src/libs/CacheAPI/index.ts
NJ-2020: Done

src/libs/CacheAPI/index.ts
NJ-2020: > Also, add a logic that key is from CACHE_API_KEYS as this will allow overriding the settings passed in init.

Done

src/libs/CacheAPI/index.ts
NJ-2020: Done

src/libs/actions/Report.ts
NJ-2020: Nope, we're sending the attachment as markdown text link

Screenshot 2025-06-16 at 10 26 02

Basically we're mapping through the parsed text (reportCommentText) and map all the attachment and store it into CacheAPI/Onyx

src/libs/actions/Report.ts
NJ-2020: Okay, will do it

src/libs/actions/Attachment/index.ts
NJ-2020: I think we should move the getAttachmentSource into a hook file, because for web we need to retrieve the CacheAPI value which returning Promise value, wdyt?

src/libs/actions/Attachment/index.native.ts
NJ-2020: Sorry, done, fixed

src/libs/actions/Attachment/index.native.ts
NJ-2020: Okay

src/libs/actions/Attachment/index.native.ts
NJ-2020: Done

src/libs/actions/Attachment/index.native.ts
NJ-2020: Thank you, done

src/libs/actions/Attachment/index.native.ts
NJ-2020: It's because we haven't cache the attachment yet, because it will took some time to fetch and cache the attachment, e.g fetching and caching 100MB video, so to fix that I think we can display the remoteSource first, and while doing that we can store it and cache it

src/libs/actions/Attachment/index.native.ts
NJ-2020: Done

src/libs/actions/Attachment/index.native.ts
NJ-2020: Thank you, done

src/libs/actions/Attachment/index.native.ts
NJ-2020: Done

src/libs/actions/Attachment/index.native.ts
NJ-2020: Sorry, done

src/libs/ReportActionsUtils.ts
NJ-2020: Ohh, sorry miss that

Will look into this, because there's some issues related to this function

src/libs/ReportActionsUtils.ts
NJ-2020: Sure, will do it soon

https://github.com/Expensify/App/pull/63723#discussion_r2149539421

src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx
NJ-2020: Nope
https://github.com/Expensify/App/blob/f09f3a1c83e29285be7c89c7e651475e5a7a41b0/src/components/HTMLEngineProvider/HTMLRenderers/ImageRenderer.tsx#L85-L89

I think we can revert this changes because this related for retrieving attachment i.e the second phase of the PR (2/2), wdyt?

src/libs/CacheAPI/index.ts
NJ-2020: `cacheName` is individual key in the lists of `CACHE_API_KEYS`

Screenshot 2025-06-16 at 17 08 32

src/libs/actions/Attachment/index.native.ts
parasharrajat: yes, then it should be use remoteSource instead of source. We should not mix both values in these variables.

src/libs/CacheAPI/index.ts
```diff
@@ -0,0 +1,60 @@
+import CONST from '@src/CONST';
+
+function init() {
+    // Exit early if the Cache API is not supported in the current browser.
+    if (!('caches' in window)) {
+        throw new Error('Cache API is not supported');
+    }
+    const keys = Object.values(CONST.CACHE_API_KEYS);
```
parasharrajat, line 8: Ah, maybe I wan't clear. We have to use keys passed in parm but this is fine for now.

NJ-2020: Okay

src/libs/actions/Attachment/index.native.ts
NJ-2020: Okay, sure

src/setup/index.ts
parasharrajat: I left a very confusing commen. Sorry about that. What i meant to say was that we should pass `CONST.CACHE_API_KEYS` directly to the `init` function or send specific keys are array. 

eg.

As object

 ```CacheAPI.init(CONST.CACHE_API_KEYS)``` 

As array

```CacheAPI.init([Key1, Key2]);```

But I am fine with the current logic where you are directly accessing them inside the function as we do not have any other keys currently for cache.

File: src/libs/actions/Report.ts (hunk 31 above)
Comment (parasharrajat) [reply src/libs/actions/Report.ts]: We need to consider this case. It is important app concept.

File: src/libs/CacheAPI/index.ts (hunk 11 above)
Comment (parasharrajat) [reply src/libs/CacheAPI/index.ts:35]: Let's keep this `cacheName` internal in the cacheAPI.

File: src/setup/index.ts (hunk 32 above)
Comment (NJ-2020) [reply src/setup/index.ts]: Oh, okay, thank you

File: src/libs/CacheAPI/index.ts (hunk 11 above)
Comment (NJ-2020) [reply src/libs/CacheAPI/index.ts:35]: Sure

File: src/libs/CacheAPI/index.ts (hunk 9 above)
Comment (NJ-2020) [reply src/libs/CacheAPI/index.ts]: > We don't have to check for these when all of them are required params. Please define the type in in App Response.

Regarding this, CacheAPI only allow [response](https://developer.mozilla.org/en-US/docs/Web/API/Cache/put#syntax) value type, in this case the return response value returned from the fetch API, so I think defining our own response type definition I think it may can lead to typeerror or unmatch type with certain urls, wdyt?

File: src/libs/actions/Report.ts (hunk 26 above)
Comment (NJ-2020) [reply src/libs/actions/Report.ts]: Done

File: src/libs/actions/Report.ts (hunk 31 above)
Comment (NJ-2020) [reply src/libs/actions/Report.ts]: I think for Onyx we can solve it by passing into `optimisticData`, `successData` and `failureData`

But regarding CacheAPI it seems a little bit tricky, I am not sure regarding the subscribtion solution because we have to manage the Onyx and CacheAPI consistent in each other and make sure it runs smoothly

> Please suggest yours also

IMO, I think we can introduce new method prop function inside `libs/API` onSuccess and onFailure
Here's an example

```tsx
API.write(commandName, parameters, {
    optimisticData,
    successData,
    failureData,
    onSuccess: () => {},
    onFailure: () => { /** remove attachment from CacheAPI */ },
});
```

Wdyt?, we can improve this later

src/libs/actions/Session/index.ts
NJ-2020: I've tried measuring the logout function performance using like this:
```tsx
function cleanupSession() {
    const startTime = performance.now();

    Pusher.disconnect();
    Timers.clearAll();
    Welcome.resetAllChecks();
    MainQueue.clear();
    HttpUtils.cancelPendingRequests();
    PersistedRequests.clear();
    NetworkConnection.clearReconnectionCallbacks();
    SessionUtils.resetDidUserLogInDuringSession();
    resetHomeRouteParams();
    CacheAPI.clear();
    clearCache().then(() => {
        const endTime = performance.now();
        const duration = endTime - startTime;

        // console.log...
    });
    clearSoundAssetsCache();
    Timing.clearData();
}
```

And here's the result using 5 attachments video, for each video the size is: 14.3MB and the total is 71.5MB  :
| Without clearing CacheAPI  | With clearing CacheAPI  |
| Duration: 174ms  | 245.8ms |

src/libs/ReportActionsUtils.ts
NJ-2020: Sorry, my bad, it should `ATTACHMENT_ID` instead of `ATTACHMENT_SOURCE_ID`, done, fixed

`ATTACHMENT_SOURCE_ID` returns the source id attachment of the source url, here's an example:

```tsx

SOURCE_URL: "https://new.expensify.com/attachment?source=https://www.expensify.com/chat-attachments/8023002770540586448/w_05944f939a087e67231a88e412843ba5b39095d1.mov&type=r&reportID=1554590248724514&isAuthTokenRequired=true&fileName=Attachment"
OUTPUT: 8023002770540586448


```

src/libs/ReportActionsUtils.ts
NJ-2020: Done fixed

tests/unit/AttachmentStorageTest.ts
NJ-2020: I will create the unit test file after all the changes are done

NJ-2020:
Bump @parasharrajat for above comments ^

NJ-2020:
Bump @parasharrajat

src/libs/actions/Report.ts
parasharrajat: > IMO, I think we can introduce new method prop function inside libs/API onSuccess and onFailure
Here's an example

We can't do that. It is against our architecture. We need to find a way to do this. But we can focus on it later.

src/libs/actions/Session/index.ts
parasharrajat: That's a lot. So if a user keeps on using the app for a long time and logs out, the app might take seconds to log out.

parasharrajat:
I will start testing this tomorrow.

src/libs/actions/Report.ts
NJ-2020: Okay

src/libs/actions/Session/index.ts
NJ-2020: Yes, so I think for alternative solution, I think we can clear the cache api when the user is on the login page or inside the PublicScreens, by using this way it will not affect the logout process, wdyt?

parasharrajat: No a good idea. Maybe we leave it as it is. When a user uninstalls the app, whether the cache is cleaned or not? 

We might need to add an option in the app to clean the storage manually.

tests/unit/AttachmentStorageTest.ts
parasharrajat: Let's add it now. It is to validate that your changes work so as you go along updating this implementation the tests should also pass.

src/libs/CacheAPI/index.ts
parasharrajat: its fine to use standard type.

src/libs/actions/Attachment/index.ts
parasharrajat: then you make native also promise.

src/libs/actions/Session/index.ts
NJ-2020: > When a user uninstalls the app, whether the cache is cleaned or not?

I am not sure about about uninstalling the app, because I don't know if the native os is deleting the entire app including the media folder which will also delete the attachment files

But for web if we don't clear the data/attachments on signout process, the data will keep saved until the user removes completely from their application storage 

> We might need to add an option in the app to clean the storage manually.

Yes, and I think maybe we can also implement like maybe storage management?

tests/unit/AttachmentStorageTest.ts
NJ-2020: Okay

src/libs/CacheAPI/index.ts
NJ-2020: Ok

src/libs/actions/Attachment/index.ts
NJ-2020: Okay sure

PR #63912, opened 2025-06-11
FIX: Click outside will dismiss active popover

ikevin127:
Closed due to force-push removing all commits, reopened @ https://github.com/Expensify/App/pull/64097.


ikevin127:
@dangrous @dukenv0307 Please let me know what you think regarding tests, whether or not we should add tests for the updated Popover behaviour similar to what I described in the PRs  **Tests** section, given that up until now there were no tests added for any Popover related component.

Issue #63300, opened 2025-06-03
[Due for payment 2025-07-14] [$250] Reports - After deleting expense and returning to "Chats"  messages display fallback avatar.


mitarachim:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.1.56-1
**Reproducible in staging?:** Yes
**Reproducible in production?:** Yes
**If this was caught during regression testing, add the test name, ID and link from TestRail:** N/A
**Email or phone of affected tester (no customers):** ibellicotest+948@gmail.com
**Issue reported by:** Applause Internal Team
**Device used:** Motorola MotoG60 - Android 12 - Chrome / Windows 10 - Chrome
**App Component:** Search

## Action Performed:
Prerequisite: Account has at least one workspace available.
Prerequisite 2: No expenses created yet.
1. Open the Expensify app.
2. Open any workspace chat.
3. Submit a manual expense.
4. Navigate to "Reports"
5. Tap on the bookmark icon and select "Chats"
6. Note that all the messages display the correct avatat.
7. Tap on bookmark icon again and select "Expenses"
8. Long tap on the just created expense and select it.
9. Delete the expense.
10. After the expense disappears, use device´s back button to return to "Chats"
11. Note that now, all the messages display fallback avatar.

## Expected Result:
All the messages on "Chats" section, should display the correct avatar in every moment.

## Actual Result:
After deleting an expense on "Expenses" section and returning to "Chats" using the device´s back button, all the messages are displayed with fallback avatar.

## Workaround:
Unknown

## Platforms:
- [x] Android: App
- [x] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

https://github.com/user-attachments/assets/5bdb572c-0dc7-43af-917f-d74b339bc12c

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
[...]

PR #65300, opened 2025-07-02
[No QA] Revert "After deleting expense and returning to Chats messages display fallback avatar"

cristipaval:
Reverts Expensify/App#64403 because it adds an extra API call. Details [here](https://github.com/Expensify/App/issues/63300#issuecomment-2991247336). We'll go with the other [solution](https://github.com/Expensify/App/issues/63300#issuecomment-2933373895).


Beamanator (approved):
approving but will let you merge 😅

PR #65801, opened 2025-07-10
Migrate Onboarding/Offboarding to GH Form

fukawi2:
Migrate this Issue Template to a GH Form to ensure more consistency, and avoid omitting required information.

### Fixed Issues
https://github.com/Expensify/Expensify/issues/484934

### Tests

image


rafecolton:
Do you have this set up in a test repo that I can try it out?

fukawi2:
https://github.com/fukawi2/scratch/issues

.github/ISSUE_TEMPLATE/OnboardOffboardExpertContributor.yml
```diff
@@ -0,0 +1,111 @@
...
+        ## STOP HERE
+        Click Create, then complete your actionable steps below and mark when done.
+
+        ---
+
+  - type: checkboxes
+    id: requester_tasks
+    attributes:
+      label: Requester Tasks
+      description: Add / Remove from the following Slack channels via Slack Connect
```
rafecolton, line 88: This instruction doesn't actually appear anywhere in the rendered issue - can we add it somehow? Alternatively, since these are not required and not used until the issue is rendered, could we make this whole section markdown instead of checkbox form fields

.github/ISSUE_TEMPLATE/OnboardOffboardExpertContributor.yml
```diff
@@ -0,0 +1,111 @@
...
+        - Software Mansion
+      default: 0
+    validations:
+      required: true
+
+  - type: markdown
+    attributes:
+      value: |
+        ## STOP HERE
+        Click Create, then complete your actionable steps below and mark when done.
```
rafecolton, line 80: NAB for clarity
```suggestion
        ## STOP HERE
        Click Create, then complete your actionable steps in the created issue
```

rafecolton (changes requested):

.github/ISSUE_TEMPLATE/OnboardOffboardExpertContributor.yml
fukawi2: Huh, that's fun. Didn't realize that :grimacing: 

I've updated the scratch repo by just adding the instructions to each line. I haven't found a way to properly nest checkboxes with this syntax, even using just markdown (because the indentation gets stripped during rendering)

rafecolton: We could also just skip converting this to yaml. I'm not really sure it's needed TBH. Unlike other vendors, I've never seen an expert agency manager fill it out wrong.

fukawi2: As I [mentioned in Slack](https://expensify.slack.com/archives/G0C394ZT8/p1752105969942239), I found the existing template very confusing so I think migrating it is worthwhile.

Issue #59689, opened 2025-04-04
[Tracking] Optimistic report title formula computation


neil-marcellini:
## Problem
We have been working on implementing automatic report titles for new Expensify to match the feature on classic. Automatic report titles use a formula defined the policy to compute the report name and the report name is continually updated as data changes. The formulas are made up of various different parts such as `{report:type}`, `report:startdate`, etc. You can check out how this feature works by going to Expensify Classic (expensify.com), workspace settings, choosing the report tab on the left side, and scrolling down to the automatic report title section. There is also a help site page linked there with more information. We're also updating the feature so that it sends out real time updates versus requiring a page refresh or loading the report.

In order to maintain our standards of an optimistic offline first application, we will need to implement report formula computation on the front end. When a new report is created, it should optimistically have the automatic report title set based on the formula in the policy settings. Whenever any data is optimistically updated such as the amount of an expense, for example, then the report title should update if it depends on that data. One example would be a formula that references the report total.

We will gradually be rolling out support for more formula parts in new Expensify, and the entire feature will be under a beta flag while development is still active. Please stay in touch with me (@neil-marcellini) for updates on our backend progress.

## Solution

~Let's design a system to optimistically update the report title based on the formula, and integrate it into the application such that any API request will be automatically handled. It might make a good middleware for example. Once the system is designed, implement it for a few report formula parts that we already support like report type, start date and total. After that, we can distribute the implementation across a large number of contributors to get the remaining parts done faster.~

~The system should be under a beta to start, and we'll let you know the name of the beta as soon as it is created.~

- [x] Edit: follow the plan in the [Offline Support section](https://docs.google.com/document/d/1U2AiMOlWvT144a2WVi2XCSc18piyE_iFrv7cYWhsAyw/edit?tab=t.0#heading=h.olra4emh77tp) of the design doc to create the initial structure. If you don't have access to the document, please request access from an internal employee.

#### Basic report fields

- [ ] https://github.com/Expensify/App/issues/68695

- `{report:id}` – Base62 report ID
- `{report:status}` – Report status
- `{report:expensescount}` – Number of expenses in report

#### Report Dates

- `{report:enddate}` – Report end date (newest expense date)

#### Report Amounts

- `{report:reimbursable}` – Total reimbursable amount

#### Policy / Workspace Info

- `{report:autoreporting:start}` – Start date of Auto-reporting frequency info
- `{report:autoreporting:end}` – End date of Auto-reporting frequency info

#### Submission Info

[...]

PR #65862, opened 2025-07-10
Optimistic report name computation

sosek108:
### Explanation of Change
This PR introduces optimistic Custom Report Names for New Expensify, enabling real-time, formula-based report titles that update as expense data changes. When a formula part can’t be computed due to missing data, the system falls back to the original formula definition to show which parts weren’t evaluated. The implementation remains compatible with Classic Expensify formulas while leveraging New Expensify’s real-time architecture and Onyx optimistic updates, providing expected visible changes even without network. This PR lays the groundwork for full computation of report formulas, to be delivered step by step in follow-up PRs.
### Fixed Issues

$ https://github.com/Expensify/App/issues/59689
PROPOSAL:

### Tests

Follow **QA Steps** fields below

### Offline tests
Before running tests ensure your account has access to BETA functionality `useCustomReportNamesNewExpensify`.

1. Turn on `Force offline mode` and do same steps as in **QA Steps** section. 

Report titles should be properly formated
### QA Steps
Before running tests ensure your account has access to BETA functionality `useCustomReportNamesNewExpensify`.

The report titles should not *flicker* in any case. By flicker I mean that you can see the report title change value 2 times on any of described action. 

**Test Case 1 – All Report Fields Combined**  
1. Go to **Workspaces**  
2. Select a **workspace you admin**  
3. Go to **Reports** section  
4. Edit **Default report title** with:  
   {report:type} - {report:total} from {report:policyname} on {report:startdate}  
5. Save the formula  
6. Create **new report** via FAB  
7. Add at least **one expense transaction**  
8. Verify title shows:  
   Custom Expense Report - USD$X.XX from [WorkspaceName] on [StartDate]  

**Test Case 2 – Date Fields Comparison**  
1. Go to **Workspaces**  
2. Select a **workspace you admin**  
3. Go to **Reports** section  
4. Edit **Default report title** with:  
   Created {report:created} vs Started {report:startdate}  
5. Save the formula  
6. Create **new report** via FAB  
7. Add expense transaction with **different date than report creation**  
8. Verify title shows both **creation date** and **start date** correctly  

**Test Case 3 – Currency and Total Display**  
1. Go to **Workspaces**  
2. Select a **workspace you admin**  
3. Go to **Reports** section  
4. Edit **Default report title** with:  
   {report:currency} {report:total} report  
5. Save the formula  
6. Create **new report** via FAB  
7. Add expense transaction with **amount**  
8. Verify title shows: [Currency] [Amount] report  

**Test Case 4 – Substring Function**  
1. Go to **Workspaces**  
2. Select a **workspace you admin**  
3. Go to **Reports** section  
4. Edit **Default report title** with:  
   ID: {report:type|substr:0:3}-{report:total|substr:1:4}  
5. Save the formula  
6. Create **new report** via FAB  
7. Add expense transaction with **amount**  
8. Verify title shows **truncated type** and **amount** (e.g., ID: Exp-23.5)  

**Test Case 5 – Transaction Addition Impact**  
1. Go to **Workspaces**  
2. Select a **workspace you admin**  
3. Go to **Reports** section  
4. Edit **Default report title** with:  
   Total: {report:total} Start: {report:startdate}  
5. Save the formula  
6. Create **new report** via FAB  
7. Add **first expense transaction** (note the title)  
8. Add **second expense transaction** with different date  
9. Verify title updates **total amount** and possibly **start date**  

**Test Case 6 – Policy Name Change Impact**  
1. Go to **Workspaces**  
2. Select a **workspace you admin**  
3. Go to **Reports** section  
4. Edit **Default report title** with:  
   Report from {report:policyname}  
5. Save the formula  
6. Create **new report** via FAB  
7. Note the current **report title**  
8. Go to **Workspace settings**  
9. Change **workspace name**  
10. Return to reports and verify **existing report titles** updated with the new workspace name

### PR Author Checklist

  - [x] Android: Native
  - [x] Android: mWeb Chrome
  - [x] iOS: Native
  - [x] iOS: mWeb Safari
  - [x] MacOS: Chrome / Safari
  - [x] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/31f7bdef-cf88-4cf5-88f4-bcd3086df6ae

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/3d33a28c-12b9-4a98-a67b-a46e4e95ac96

</details>

<details>

https://github.com/user-attachments/assets/ef2fcba3-6f5e-45cf-b639-f6e23357c19d

</details>


src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+import {getCurrencySymbol} from '@libs/CurrencyUtils';
+import CONST from '@src/CONST';
+import type Policy from '@src/types/onyx/Policy';
+import type Report from '@src/types/onyx/Report';
+
+type FormulaPart = {
+    /** The original definition from the formula */
+    definition: string;
+    /** The type of formula part (report, field, user, etc.) */
+    type: string;
```
neil-marcellini: Can we limit the type to one of the constants in FORMULA_PART_TYPES?

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+    /** The field path for accessing data (e.g., ['type'], ['startdate'], ['total']) */
+    fieldPath: string[];
+    /** Functions to apply to the computed value (e.g., ['frontPart']) */
```
neil-marcellini: Please add a blank space above comments

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+type FormulaContext = {
+    report: Report;
+    policy: OnyxEntry<Policy>;
+};
+
+const FORMULA_PART_TYPES = {
+    REPORT: 'report',
+    FIELD: 'field',
+    USER: 'user',
+    FREETEXT: 'freetext',
```
neil-marcellini: I guess we need to add a spellcheck exception for this

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+                value = computeReportPart(part, context);
+                break;
+            case FORMULA_PART_TYPES.FIELD:
+                value = computeFieldPart(part, context);
+                break;
+            case FORMULA_PART_TYPES.USER:
+                value = computeUserPart(part, context);
+                break;
+            case FORMULA_PART_TYPES.FREETEXT:
+                value = part.definition.trim();
```
neil-marcellini: Not sure we should trim free text parts, I think we can use the definition directly.

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+                break;
+            default:
+                // If we don't recognize the part type, use the original definition
+                value = part.definition;
+        }
+
+        // Apply any functions to the computed value
+        value = applyFunctions(value, part.functions);
+        result = result === '' ? value : `${result} ${value}`.trim(); // Concatenate with space
```
neil-marcellini: I think if we make free text parts include spaces then we won't have to concatenate with a space here. Also maybe people want to use multiple spaces in some areas of the formula.

We can always test against OldDot (with an account not on the beta) and use that as a baseline to ensure the behavior matches.

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+    const {report, policy} = context;
+    const [field] = part.fieldPath;
+
+    if (!field) {
+        return part.definition;
+    }
+
+    switch (field.toLowerCase()) {
+        case 'type':
+            return 'Expense Report'; // Default report type for now
```
neil-marcellini: This should be pretty easy to add right?

src/libs/CustomFormula.ts
neil-marcellini, line 1: Can we rename this to Formula.ts so it matches the backend?

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+        case 'startdate':
+            return formatDate(report.lastVisibleActionCreated);
```
neil-marcellini: I don't think that logic is correct at all. Please check out the OldDot JS implementation in [this google doc](https://docs.google.com/document/d/1B42XFKyvTyb_UlpfoAjEkaq3UaNamIx2hloylzJriYY/edit?usp=sharing) I just shared. I also linked it in the design doc.

src/libs/CustomFormula.ts
```diff
@@ -0,0 +1,381 @@
...
+        return '';
+    }
+
+    try {
+        const date = new Date(dateString);
+        if (isNaN(date.getTime())) {
+            return '';
+        }
+
+        // Simple date formatting - this could be enhanced with a proper date library
```
neil-marcellini: I realized that I don't know precisely which formats we support, so I went down a rabbit hole trying to figure that out.

neil-marcellini (changes requested):
Here is a partial review. I will do more later.

src/libs/CustomFormula.ts
neil-marcellini: I will post my findings soon

neil-marcellini: In the design doc I added a section called [Date formats](https://docs.google.com/document/d/1U2AiMOlWvT144a2WVi2XCSc18piyE_iFrv7cYWhsAyw/edit?tab=t.0#heading=h.5kgjegp525w) about which date formats we currently support and should move forward with. Please check it out.

I think we should implement this using the existing date-fns library we have installed, and a little bit of custom code. Cursor generated a pretty good version which you may find useful, but it's totally untested. It's fairly complicated so let's implement better date formatting in a separate PR.


<details>
<summary>DateFormatter</summary>

```ts
import {format, getDay, getDayOfYear, getISOWeek, getISOWeekYear, getUnixTime, isLeapYear, getDaysInMonth} from 'date-fns';
import {formatInTimeZone} from 'date-fns-tz';

/**
 * Maps PHP-style date format characters to date-fns format tokens or custom implementations
 */
const PHP_TO_DATE_FNS_MAP: Record<string, string | ((date: Date) => string)> = {
    // Day
    'd': 'dd',                              // Day of the month, 2 digits with leading zeros (01-31)
    'dd': 'dd',                             // Same as 'd' for compatibility
    'D': 'EEE',                             // Short day name (Mon-Sun)
    'j': 'd',                               // Day of the month without leading zeros (1-31)
    'l': 'EEEE',                            // Full day name (Sunday-Saturday)
    'N': (date: Date) => String(getDay(date) || 7), // ISO day of week (1-7, Monday=1)
    'S': (date: Date) => getOrdinalSuffix(date.getDate()), // English ordinal suffix (st, nd, rd, th)
    'w': (date: Date) => String(getDay(date)), // Day of week (0-6, Sunday=0)
    'z': (date: Date) => String(getDayOfYear(date) - 1), // Day of year (0-365, starting from 0)

    // Week
    'W': (date: Date) => String(getISOWeek(date)), // ISO week number

    // Month
    'F': 'MMMM',                            // Full month name (January-December)
    'M': 'MMM',                             // Short month name (Jan-Dec)
    'MMMM': 'MMMM',                         // Full month name
    'MMM': 'MMM',                           // Short month name
    'MM': 'MM',                             // Month with leading zero (01-12)
    'n': (date: Date) => String(date.getMonth() + 1), // Month without leading zeros (1-12)
    't': (date: Date) => String(getDaysInMonth(date)), // Number of days in the month

    // Year
    'L': (date: Date) => isLeapYear(date) ? '1' : '0', // Leap year (1 if leap year, 0 otherwise)
    'o': (date: Date) => String(getISOWeekYear(date)), // ISO week-numbering year
    'Y': 'yyyy',                            // Full 4-digit year
    'y': 'yy',                              // 2-digit year
    'yyyy': 'yyyy',                         // Full 4-digit year
    'yy': 'yy',                             // 2-digit year

    // Time
    'a': 'a',                               // am/pm lowercase
    'A': 'a',                               // am/pm (date-fns uses lowercase, we'll transform)
    'g': 'h',                               // 12-hour format without leading zeros (1-12)
    'G': 'H',                               // 24-hour format without leading zeros (0-23)
    'h': 'hh',                              // 12-hour format with leading zeros (01-12)
    'H': 'HH',                              // 24-hour format with leading zeros (00-23)
    'i': 'mm',                              // Minutes with leading zeros (00-59)
    's': 'ss',                              // Seconds with leading zeros (00-59)

    // Full formats
    'c': 'yyyy-MM-dd\'T\'HH:mm:ssXXX',       // ISO 8601 format
    'r': 'EEE, dd MMM yyyy HH:mm:ss xx',     // RFC 2822 format
    'U': (date: Date) => String(getUnixTime(date)), // Unix timestamp
};

/**
 * Returns the English ordinal suffix for a day (st, nd, rd, th)
 */
function getOrdinalSuffix(day: number): string {
    if (day >= 11 && day <= 13) {
        return 'th';
    }
    switch (day % 10) {
        case 1: return 'st';
        case 2: return 'nd';
        case 3: return 'rd';
        default: return 'th';
    }
}

/**
 * Parses a UTC date and formats it according to PHP-style format characters
 *
 * @param utcDate - Date object or date string in UTC
 * @param formatString - PHP-style format string (e.g., 'Y-m-d H:i:s', 'l, F j, Y')
 * @param timezone - Optional timezone (defaults to UTC)
 * @returns Formatted date string
 *
 * @example
 * ```typescript
 * const date = new Date('2025-07-31T21:42:02Z');
 *
 * parseDate(date, 'Y-m-d H:i:s')           // '2025-07-31 21:42:02'
 * parseDate(date, 'l, F j, Y')             // 'Thursday, July 31, 2025'
 * parseDate(date, 'jS \\o\\f F Y')         // '31st of July 2025'
 * parseDate(date, 'D, d M Y H:i:s \\G\\M\\T') // 'Thu, 31 Jul 2025 21:42:02 GMT'
 * parseDate(date, 'c')                     // '2025-07-31T21:42:02+00:00'
 * parseDate(date, 'r')                     // 'Thu, 31 Jul 2025 21:42:02 +0000'
 * parseDate(date, 'U')                     // '1753998122'
 * ```
 */
function parseDate(utcDate: Date | string, formatString: string, timezone = 'UTC'): string {
    let date: Date;

    // Convert input to Date object
    if (typeof utcDate === 'string') {
        date = new Date(utcDate);
    } else {
        date = utcDate;
    }

    // Validate date
    if (isNaN(date.getTime())) {
        return '';
    }

    // Split format string into tokens, preserving escaped characters
    const tokens = formatString.split(/(\\.)|([a-zA-Z]+)/g).filter(Boolean);

    let result = '';

    for (const token of tokens) {
        // Handle escaped characters (e.g., \G\M\T becomes GMT)
        if (token.startsWith('\\')) {
            result += token.slice(1);
            continue;
        }

        // Handle format characters
        const formatMap = PHP_TO_DATE_FNS_MAP[token];
        if (formatMap) {
            if (typeof formatMap === 'function') {
                let formatted = formatMap(date);

                // Special handling for uppercase AM/PM
                if (token === 'A') {
                    formatted = formatted.toUpperCase();
                }

                result += formatted;
            } else {
                let formatted: string;

                if (timezone === 'UTC') {
                    formatted = format(date, formatMap);
                } else {
                    formatted = formatInTimeZone(date, timezone, formatMap);
                }

                // Special handling for uppercase AM/PM
                if (token === 'A') {
                    formatted = formatted.toUpperCase();
                }

                result += formatted;
            }
        } else {
            // If no mapping found, output the token as-is
            result += token;
        }
    }

    return result;
}

/**
 * Common preset formats for convenience
 */
const DATE_FORMATS = {
    // Standard formats
    ISO_DATE: 'Y-m-d',                          // 2025-07-31
    ISO_DATETIME: 'Y-m-d H:i:s',                // 2025-07-31 21:42:02
    ISO_8601: 'c',                              // 2025-07-31T21:42:02+00:00
    RFC_2822: 'r',                              // Thu, 31 Jul 2025 21:42:02 +0000
    UNIX_TIMESTAMP: 'U',                        // 1753998122

    // Human-readable formats
    LONG_DATE: 'l, F j, Y',                     // Thursday, July 31, 2025
    MEDIUM_DATE: 'M j, Y',                      // Jul 31, 2025
    SHORT_DATE: 'n/j/Y',                        // 7/31/2025
    ORDINAL_DATE: 'jS \\o\\f F Y',              // 31st of July 2025

    // Time formats
    TIME_12H: 'g:i A',                          // 9:42 PM
    TIME_24H: 'H:i',                            // 21:42
    TIME_WITH_SECONDS: 'H:i:s',                 // 21:42:02

    // Combined formats
    DATETIME_READABLE: 'l, F j, Y \\a\\t g:i A', // Thursday, July 31, 2025 at 9:42 PM
    DATETIME_SHORT: 'M j, Y g:i A',             // Jul 31, 2025 9:42 PM
} as const;

export {parseDate, DATE_FORMATS};
export default parseDate;
```
</details>

neil-marcellini: After thinking about it more, we probably don't need to support ALL of those formats, because some of them are duplicates, and some are overkill or unlikely to be used. I also don't want to invest the time right now to add support for those in the new backend implementation in Auth.

Tomorrow I will aim to reduce the list to those already supported in Auth, and add any that seem really important.

src/libs/CustomFormula.ts
koko57: yes, but only free text parts had spaces - the other parts haven't so they were displayed without spaces

src/libs/CustomFormula.ts
koko57: I added it, because after my change that fixes other parts not having a space ([comment](https://github.com/Expensify/App/pull/65862#discussion_r2247147234)) we get extra spaces. But ofc it can be handled differently

src/libs/CustomFormula.ts
koko57: @neil-marcellini not sure what do you mean? 😅

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+}
+
+/**
+ * Initialize the context needed for report name computation
+ * This should be called before processing optimistic updates
+ */
+function createUpdateContext(): Promise<UpdateContext> {
+    return new Promise((resolve) => {
+        // Get all the data we need from Onyx
+        const connectionID = Onyx.connect({
```
koko57: just noting that Onyx.connect is deprecated

src/libs/CustomFormula.ts
koko57: yeah, I see. TBH I don't see a property on the report object that would be suitable here.

src/libs/CustomFormula.ts
koko57: so I think @sosek108 will take it back from here - he'll be back on Monday

src/libs/CustomFormula.ts
neil-marcellini: Use report.type, and a mapping of types to human readable names to get the proper value.

```cpp
const map<string, string> Report::TYPES_HUMAN_READABLE = {
    {TYPE_EXPENSE, "Expense Report"},
    {TYPE_INVOICE, "Invoice"},
    {TYPE_CHAT, "Chat"},
    {TYPE_BILL, "Bill"},
    {TYPE_PAYCHECK, "Paycheck"},
    {TYPE_IOU, "IOU"},
    {TYPE_TASK, "Task"},
    {TYPE_TRIP, "Trip"},
};
```

src/libs/CustomFormula.ts
neil-marcellini: Yeah I think it should be. Here's a test formula that I think will illustrate the problem.

`Report with type after 4 spaces   {report:type}-and no space after computed part`

That should be parsed into the following part definitions
`["Report with type after 4 spaces   ", "{report:type}", "-and no space after computed part"]`

And the result for an Expense Report should be:
`Report with type after 4 spaces   Expense Report-and no space after computed part`

I don't think that will work with the current implementation.

src/libs/CustomFormula.ts
neil-marcellini: Let's consolidate the discussion to the other thread. As long as the example I gave there works I think we're ok.

src/libs/CustomFormula.ts
neil-marcellini: We need to get the oldest date from all the transaction created dates, and the opposite for the enddate.

src/libs/OptimisticReportNames.ts
neil-marcellini: Yes, we may need to ask for an exception on this one, unless we can reasonably pass the required context by some other method.

src/libs/CustomFormula.ts
neil-marcellini: Oh I see you did update this, thanks! Let's add this formula as a test case so we can be sure it's working properly.

.github/actions/javascript/checkAndroidStatus/index.js
neil-marcellini, line 1: Not sure why there is a diff on this file, but we should undo that unless it's needed.

src/CONST/index.ts
```diff
@@ -1486,6 +1487,9 @@ const CONST = {
         APPLY_AIRSHIP_UPDATES: 'apply_airship_updates',
         APPLY_PUSHER_UPDATES: 'apply_pusher_updates',
         APPLY_HTTPS_UPDATES: 'apply_https_updates',
+        COMPUTE_REPORT_NAME: 'compute_report_name',
+        COMPUTE_REPORT_NAME_FOR_NEW_REPORT: 'compute_report_name_for_new_report',
+        UPDATE_OPTIMISTIC_REPORT_NAMES: 'update_optimistic_report_names',
```
neil-marcellini, line 1519: What is the benefit of these additional betas, and how do we plan to use them?

src/libs/OptimisticReportNames.ts
neil-marcellini: NAB: I find all the indentation really ugly, can we please use async await or try to chain the promises and callbacks in a more elegant manner?

neil-marcellini: Also, why are we getting all the data and then disconnecting everything? Every time a request is prepared with optimistic data we will be re-doing this process. Instead, can we create all the connections when the app is initialized, in the background? And also keep the connections open so that the most up to date data can be accessed whenever a request is prepared?

neil-marcellini (changes requested):
Here's another partial review

src/libs/OptimisticReportNames.ts
neil-marcellini: I asked Tim about an exception to the Onyx.connect rule [here](https://github.com/Expensify/Expensify/issues/507850#issuecomment-3145487711) (internal).

src/CONST/index.ts
sosek108: this is not for betas but for timings. Different names as different places are potentially measured via `Performance` and `Timing`

src/libs/CustomFormula.ts
sosek108: I think that we should use localized values for that. What do you think @neil-marcellini?

src/libs/CustomFormula.ts
sosek108: I've added testcase for that exact scenario

src/libs/CustomFormula.ts
sosek108: I've modified that logic to use first `reportActions`.

src/libs/CustomFormula.ts
sosek108: > Tomorrow I will aim to reduce the list to those already supported in Auth, and add any that seem really important.

@neil-marcellini did you manage to do so?

src/libs/CustomFormula.ts
sosek108: after double checking I've modified that. `startdate` uses transactions and `created` uses report actions.

I've implemented date formats from Design Doc

src/libs/CustomFormula.ts
sosek108: I've implemented all types formats Design doc

sosek108:
## Status 
`question`: Should both report type and dates be using system localization mechanism?

`question 2`: I’ve used that on expensify.com `SSetNameValuePair('useCustomReportNamesNewExpensify', 'true');`, and it seams that backend values are working fine, but I don’t see this beta on frontend and therefore `Permissisons.canUseCustomReportNames` returns false. Should I do anything else?
I’ve updated beta name to be same as backend. 
I’ve implemented date formats from design doc and updated calculation of `startdate` and `created` to consume proper values.
I agree that indendation and connections could be done cleaner - I will handle that next step alongside with whole PR cleanup when we confirm that logic is fine.

Next steps:
1. report type computation
2. date / report from localization if needed (to be confirmed)
3. code cleanup and more efficient connections 
4. validate test cases 

@neil-marcellini

neil-marcellini:
1. Let's not worry about localization for now, because we don't do that on the backend.
2. Odd, I can look into that later

Here is a [spreadsheet of the date formats that are relatively useful](https://docs.google.com/spreadsheets/d/1A9vEWPaW_yvpgrRaOv96P_2EuHiYigC35aFLY7sBIp0/edit?gid=1975994018#gid=1975994018&fvid=89451745). I'm still ensuring all of these are supported by the new backend system, but let's plan to support all these on the frontend. We can create a separate PR to support all the date formats, for now we can keep it more simple.

src/libs/CustomFormula.ts
neil-marcellini: We're not planning to localize the backend currently, so let's make the frontend match.

src/libs/Formula.ts
```diff
@@ -0,0 +1,458 @@
...
+        }
+    });
+
+    return oldestDate;
+}
+
+/**
+ * Get the date of the oldest transaction for a given report
+ */
+function getOldestTransactionDate(reportID: string): string | undefined {
```
neil-marcellini, line 471: To match the backend we should also default this to the current date if there are no transactions.

Additionally, we need to exclude partial transactions.

```cpp
string Report::getTransactionCreatedDate(SQLite& db, const int64_t reportID, const bool forMostRecentTransaction, const string& defaultDate)
{
    // Get the most or least recent created date from the report transactions, excluding partial transactions
    const string transactionCreated = DB::read(db,
        "SELECT COALESCE(modifiedCreated, created) AS transactionCreated "
        "FROM transactions "
        "WHERE reportID > 0 "
        "  AND reportID = " + SQ(reportID) + " "
        "  AND (COALESCE(modifiedAmount, amount) != " + to_string(Transaction::PARTIAL_AMOUNT) + " OR COALESCE(NULLIF(modifiedMerchant, ''), merchant) != " + SQ(Transaction::PARTIAL_MERCHANT) + ") "
        "ORDER BY transactionCreated " + (forMostRecentTransaction ? "DESC" : "ASC") + " "
        "LIMIT 1;"
    );
    return transactionCreated.empty() ? defaultDate : transactionCreated;
}
```

src/libs/Formula.ts
```diff
@@ -0,0 +1,458 @@
...
+        return value.substr(start, length);
+    }
+
+    return value.substr(start);
+}
+
+/**
+ * Format a date value with support for multiple date formats
+ */
+function formatDate(dateString: string | undefined, format = 'MM/dd/yyyy'): string {
```
neil-marcellini: Looks like a pretty good start. Can we leverage the date-fns library to do more of the formatting for us, or do we need the custom implementation? I was hoping we could use some of the [supported formats](https://date-fns.org/v4.1.0/docs/format) and map any of the other formats to date-fns formats or create a custom implementation as needed.

src/libs/Formula.ts
```diff
@@ -0,0 +1,458 @@
...
+    const [field, format] = part.fieldPath;
+
+    if (!field) {
+        return part.definition;
+    }
+
+    switch (field.toLowerCase()) {
+        case 'type':
+            return 'Expense Report'; // Default report type for now
+        case 'startdate':
```
neil-marcellini, line 246: Eventually we need to support enddate as well, but we can do that in a follow up to keep this more modular.

src/libs/CustomFormula.ts
neil-marcellini: Ok, here are the [formats we're supporting](https://github.com/Expensify/App/pull/65862#issuecomment-3151763633). I have created a separate issue to track this. Please assign yourself to it and you can start on it later.

src/libs/OptimisticReportNames.ts
neil-marcellini: Ok, here's what he said:

Put all that "centralized system" in it's own separate JS module
Use Onyx.connecteWithoutView() to access the necessary data
Add comments to explain why it's OK and to disuade other people from copying that pattern

src/CONST/index.ts
neil-marcellini: Oh my bad, that's totally fine then.

neil-marcellini (changes requested):
Here's another partial review. I will continue to review the rest.

src/libs/Formula.ts
```diff
@@ -0,0 +1,458 @@
...
+            return formatDate(getOldestTransactionDate(report.reportID), format);
+        case 'total':
+            return formatAmount(report.total, getCurrencySymbol(report.currency ?? '') ?? report.currency);
+        case 'currency':
+            return report.currency ?? '';
+        case 'policyname':
+        case 'workspacename':
+            return policy?.name ?? '';
+        case 'created':
+            return formatDate(getOldestReportActionDate(report.reportID), format);
```
neil-marcellini, line 258: I'm wondering why you don't use `report.created`, then I realized that field isn't available in the frontend, which is very shocking because it's one of the columns in the reports table 😯

After more digging I see that we always create an action when the report is created, so these values are the same, so this approach works. It's slightly less efficient, but we can fix that later if needed. I guess I learned something today.

All good.

src/libs/Formula.ts
```diff
@@ -0,0 +1,458 @@
...
+        const [functionName, ...args] = func.split(':');
+
+        switch (functionName.toLowerCase()) {
+            case 'frontpart':
+                result = getFrontPart(result);
+                break;
+            case 'substr':
+                result = getSubstring(result, args);
+                break;
+            default:
```
neil-marcellini, line 299: Can we also implement the domain function please?

```php
/**
     * Pulls the domain name out of a string containing an email address.  Works for:
     * "David Barrett <dbarrett@expensify.com>"
     * "dbarrett@expensify.com".
     *
     * @param string $email string containing the email from which we want to get
     *                      the domain
     *
     * @return string domain of the email
     */
    public static function getDomainName($email = '')
    {
        if ($email && preg_match('/@([^>]+)/', $email, $matches)) {
            return $matches[1];
        }

        return '';
    }
```

src/libs/Formula.ts
```diff
@@ -0,0 +1,458 @@
...
+        }
+    } catch {
+        return '';
+    }
+}
+
+/**
+ * Format an amount value
+ */
+function formatAmount(amount: number | undefined, currency: string | undefined): string {
```
neil-marcellini, line 402: NAB: Don't we already have similar utils in the App? Same for `getOldestReportActionDate`. If we can let's use something that already exists, but if not, this looks good.

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+function shouldComputeReportName(report: Report, policy: Policy | undefined): boolean {
+    // Only compute names for expense reports with policies that have title fields
+    if (!report || !policy) {
+        return false;
+    }
+
+    // Check if the report is an expense report
+    if (!ReportUtils.isExpenseReport(report)) {
+        return false;
+    }
```
neil-marcellini: I think we should actually allow computing names for all these report types

```cpp
const set<string> Report::REPORT_CUSTOM_NAME_TYPES = {
    Report::TYPE_EXPENSE,
    Report::TYPE_INVOICE,
    Report::TYPE_BILL,
    Report::TYPE_PAYCHECK,
    Report::TYPE_TRIP,
};
```

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    // Check if the policy has a title field with a formula
+    const titleField = ReportUtils.getTitleReportField(policy.fieldList ?? {});
+    if (!titleField?.defaultValue) {
+        return false;
+    }
+
```
neil-marcellini: NAB: There's a whole extra consideration here, and we don't need to tackle it just yet, but we should fix it before removing the beta and launching the feature. Report names should only get computed when the report has the title report field set in its name value pairs as well as being set on the policy. That also means the policy field has to get copied over when creating the optimistic report.

I will create a separate issue for this.

neil-marcellini: [Require title field rNVP to compute optimistic report names](https://github.com/Expensify/App/issues/67823#top)

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+/**
+ * Compute a new report name if needed based on an optimistic update
+ */
+function computeReportNameIfNeeded(
+    report: Report,
+    incomingUpdate: OnyxUpdate,
+    context: UpdateContext,
+): {
+    reportName: string;
```
neil-marcellini: NAB: Return `string | null` directly?

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    // Quick check: see if the update might affect the report name
+    const updateType = determineObjectTypeByKey(incomingUpdate.key);
+    const formula = titleField.defaultValue;
+    const formulaParts = Formula.parse(formula);
+
+    // Check if any formula part might be affected by this update
+    const isAffected = formulaParts.some((part) => {
+        if (part.type === Formula.FORMULA_PART_TYPES.REPORT) {
+            return updateType === 'report' || updateType === 'transaction';
```
neil-marcellini: A formula part of type report can also be affected by an update to the policy. For example {report:policyName} and the policy name is updated

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+        const objectType = determineObjectTypeByKey(update.key);
+        let affectedReports: Report[] = [];
+
+        switch (objectType) {
+            case 'report': {
+                const reportID = getReportIDFromKey(update.key);
+                const report = getReportByID(reportID, allReports);
+
+                // Special handling for new reports (SET method means new report creation)
+                if (!report && update.onyxMethod === Onyx.METHOD.SET) {
```
neil-marcellini: I'm not sure we can rely on that, such as if someone uses set for an existing report. Can we instead check if the report is present in the collection, and if not handle it as a new report?

Or maybe "new" is the wrong term. How about we overload the `computeReportNameIfNeeded` function so it can also work with only the update and context, or even just make the report param optional.

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,410 @@
...
+
+                if (report) {
+                    affectedReports = [report];
+                }
+                break;
+            }
+
+            case 'policy': {
+                const policyID = getPolicyIDFromKey(update.key);
+                affectedReports = getReportsByPolicyID(policyID, allReports);
```
neil-marcellini: Let's limit the affected reports further. For example see the query for the function getReportIDsForNameComputation [in this doc](https://docs.google.com/document/d/1Xq-QffDBIPtbjWJiSl9RE91sS17W2lS96Ltady3n2BQ/edit?usp=sharing)

src/libs/Permissions.ts
```diff
@@ -7,6 +7,11 @@ function canUseAllBetas(betas: OnyxEntry<Beta[]>): boolean {
     return !!betas?.includes(CONST.BETAS.ALL);
 }
 
+// eslint-disable-next-line rulesdir/no-beta-handler
+function canUseCustomReportNames(betas: OnyxEntry<Beta[]>): boolean {
+    return isBetaEnabled(CONST.BETAS.USE_CUSTOM_REPORT_NAMES, betas);
```
neil-marcellini: Ah, I think the problem is that the beta name is `authAutoReportTitle`, even though the NVP name controlling the beta is `useCustomReportNamesNewExpensify`. Sorry for the confusion. It was in the design doc for the section [Frontend beta](https://docs.google.com/document/d/1U2AiMOlWvT144a2WVi2XCSc18piyE_iFrv7cYWhsAyw/edit?tab=t.0#heading=h.rds3ioxcly7q).

tests/perf-test/Formula.perf-test.ts
```diff
@@ -0,0 +1,142 @@
...
+            await measureFunction(() => extract(formula));
+        });
+
+        test('[CustomFormula] extract() with complex formula', async () => {
+            const formula = '{report:type} - {report:startdate} - {report:total} - {report:currency} - {report:policyname} - {user:email|frontPart}';
+            await measureFunction(() => extract(formula));
+        });
+
+        test('[CustomFormula] extract() with nested braces', async () => {
+            const formula = '{report:created:yyyy-MM-dd} - {field:custom_field|substr:0:10} - {user:email|frontPart}';
```
neil-marcellini: Where are the nested braces?

tests/perf-test/OptimisticReportNames.perf-test.ts
```diff
@@ -0,0 +1,271 @@
...
+                onyxMethod: Onyx.METHOD.MERGE,
+                value: {name: `Bulk Updated Policy ${i}`},
+            }));
+
+            await measureFunction(() => updateOptimisticReportNamesFromUpdates(policyUpdates, mockContext));
+        });
+    });
+
+    describe('Large Dataset Performance', () => {
+        test('[OptimisticReportNames] processing with large context (1000 reports)', async () => {
```
neil-marcellini, line 177: 1000? I only see 20.

tests/unit/FormulaTest.ts
```diff
@@ -0,0 +1,297 @@
...
+const mockReportActionsUtils = ReportActionsUtils as jest.Mocked<typeof ReportActionsUtils>;
+const mockReportUtils = ReportUtils as jest.Mocked<typeof ReportUtils>;
+
+describe('CustomFormula', () => {
+    describe('extract()', () => {
+        test('should extract formula parts with default braces', () => {
+            expect(extract('{report:type} - {report:total}')).toEqual(['{report:type}', '{report:total}']);
+        });
+
+        test('should handle nested braces', () => {
```
neil-marcellini, line 35: Nested formulas look like this
`{report:{report:submit:from:firstName|substr:2}}`

I think they are always invalid formulas for report names because the inner formula is not evaluated

tests/unit/FormulaTest.ts
```diff
@@ -0,0 +1,297 @@
...
+                report: {} as any,
+                policy: null,
+            };
+            const result = compute('{report:total} {report:policyname}', contextWithMissingData);
+            expect(result).toBe(''); // Empty strings concatenated = empty string
+        });
+
+        test('should preserve free text', () => {
+            const result = compute('Expense Report - {report:total}', mockContext);
+            expect(result).toBe('Expense Report - USD100.00');
```
neil-marcellini: I don't think the backend computes the report total in this manner, it would show the currency symbol instead. Please test against OldDot.

tests/unit/FormulaTest.ts
```diff
@@ -0,0 +1,297 @@
...
+            expect(parts[0].type).toBe('freetext');
+        });
+
+        test('should handle invalid date', () => {
+            const context: FormulaContext = {
+                report: {lastVisibleActionCreated: 'invalid-date'} as any,
+                policy: null,
+            };
+            const result = compute('{report:startdate}', context);
+            expect(result).toBe('');
```
neil-marcellini: I think it's better to leave the definition in the name if data is missing.

tests/unit/OptimisticReportNamesTest.ts
```diff
@@ -0,0 +1,305 @@
...
+            expect(result).toBe(false);
+        });
+
+        test('should return false when no title field', () => {
+            mockReportUtils.getTitleReportField.mockReturnValue(undefined);
+            const result = shouldComputeReportName(mockReport as any, mockPolicy as any);
+            expect(result).toBe(false);
+        });
+
+        test('should return false when title field has no formula', () => {
```
neil-marcellini: Reports should still get created with that static title as the name right? I think this test shows that wouldn't be the case.

tests/unit/OptimisticReportNamesTest.ts
```diff
@@ -0,0 +1,305 @@
...
+                    value: {total: -25000},
+                },
+            ];
+
+            const result = updateOptimisticReportNamesFromUpdates(updates, mockContext);
+            expect(result).toHaveLength(2); // Original + name update
+            expect(result[1].value).toEqual({reportName: 'Expense Report - USD 250.00'});
+        });
+
+        test('should skip processing when no changes needed', () => {
```
neil-marcellini: Duplicate test?

neil-marcellini (changes requested):
I have reviewed all the code now. Lots of tweaks needed. Let's really try to break this down into manageable chunks and create issues to track other pieces that we want to add in subsequent PRs.

Thanks for the hard work so far!

src/libs/CustomFormula.ts
sosek108: Ok, I've created similar dictionary on frontend for that. Curious is that there is no `TRIP` type in `CONST.REPORT.TYPE` on frontend

src/libs/CustomFormula.ts
sosek108: Can you link the issue? I can't find it 🙈

src/libs/OptimisticReportNames.ts
sosek108: I've moved these connection to new file and used `connecteWithoutView`

src/libs/Formula.ts
sosek108: Done

src/libs/Formula.ts
sosek108: I'd be cautious about `date-fns` because our latest findings show that `date-fns` on React Native can be slow. Of course we can implement it, I'd suggest to do that in another PR (like mentioned in some other comment)

src/libs/Formula.ts
sosek108: I've had the same reaction. When we `POST` backend with new report it'll return `created` field but every other update passed via Onyx will omit that value 🤯

src/libs/Formula.ts
sosek108: Done

src/libs/Formula.ts
sosek108: I found couple of methods in CurrencyUtils but them seems to support format `60 USD` instead of `$60`

src/libs/OptimisticReportNames.ts
sosek108: should we add `TRIP` to `CONST.REPORT.TYPE`?

sosek108: Done

src/libs/OptimisticReportNames.ts
sosek108: Done

src/libs/OptimisticReportNames.ts
sosek108: done

src/libs/OptimisticReportNames.ts
sosek108: I'm filtering by:

1. type
2. state
3. archive status

src/libs/CustomFormula.ts
neil-marcellini: Yeah it's for the really old type of trip reports, before New Expensify. Might as well put it in there. It looks like the modern trip reports have a report type of chat, and then a chat type of trip, or something like that.

src/libs/CustomFormula.ts
neil-marcellini: Oh woops. I created it in the wrong repository. Here is the issue [Add all useful date formats for optimistic report name formulas](https://github.com/Expensify/App/issues/67921#top)

src/libs/OptimisticReportNamesConnectionManager.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Initialize persistent connections to Onyx data needed for OptimisticReportNames
+ * This is called lazily when OptimisticReportNames functionality is first used
+ */
+function initialize(): void {
+    if (isInitialized) {
+        return;
+    }
+
+    // Connect to BETAS
+    // eslint-disable-next-line react-compiler/react-compiler -- This is not a React component and needs to access Onyx data for OptimisticReportNames processing without being tied to a UI view
```
neil-marcellini: Let's also explain that this is a centralized system that needs access to all objects of several types, so that when any updates affect the computed report names, we can compute the new names according to the formula and add the necessary updates. It wouldn't be possible to do this without connecting to all the data.

neil-marcellini: We can explain once at the top of this function

src/libs/OptimisticReportNamesConnectionManager.ts
```diff
@@ -0,0 +1,84 @@
...
+    Onyx.connectWithoutView({
+        key: ONYXKEYS.COLLECTION.POLICY,
+        waitForCollectionCallback: true,
+        callback: (val) => {
+            allPolicies = (val as Record<string, Policy>) ?? {};
+        },
+    });
+
+    isInitialized = true;
```
neil-marcellini: Maybe we should only consider it initialized once all the callbacks have fired for the first time? I'm curious to see how this is handled elsewhere.

src/libs/Formula.ts
neil-marcellini: Ok yeah let's address this later. I'm curious to see some measurements for how it's too slow, and I would also imagine that it would be fast enough for this use case.

src/libs/OptimisticReportNames.ts
neil-marcellini: Oh, maybe the TRIP type should go in the UNSUPPORTED_TYPEs

src/libs/OptimisticReportNames.ts
```diff
@@ -217,7 +242,7 @@ function computeReportNameIfNeeded(
     // Check if any formula part might be affected by this update
     const isAffected = formulaParts.some((part) => {
         if (part.type === Formula.FORMULA_PART_TYPES.REPORT) {
-            return updateType === 'report' || updateType === 'transaction';
+            return updateType === 'report' || updateType === 'transaction' || updateType === 'policy';
```
neil-marcellini: NAB: Checking if the formula part is affected in this manner works, but it could certainly be more precise. For example, a policy update only affects the part if the formula in the policy changed, or if the report part references a field on the policy.

It's a bit complicated to implement those more precise checks, so I'm happy with this for now. However, if we run into performance problems, this would be a good place to optimize.

src/libs/OptimisticReportNames.ts
neil-marcellini: Ok great. Two more conditions to add would be verifying that the policyID is not fake `"_FAKE_"` and verifying the statement card id rnvp is not set

neil-marcellini (changes requested):
Thanks for the updates. Here's some feedback based on the last batch of changes. Please DM me when you are ready for the next review.

src/libs/OptimisticReportNames.ts
sosek108: what's the best way to track these ideas in Expensify? Should I copy your comment to the codebase? Or maybe let's create issue and we can handle that right after more important works?

src/libs/OptimisticReportNames.ts
sosek108: 1. Made report parameter optional in computeReportNameIfNeeded (line 217)
  2. Added fallback logic to extract report data from the update when report is undefined (lines 223-230)
  4. Simplified the switch case logic (lines 332-351) to use the same function for both existing and new reports
  5. Removed the unreliable SET method check completely

src/libs/OptimisticReportNames.ts
sosek108: https://github.com/Expensify/App/blob/073fa4a887a43b2124910e87b113485c6ff3307d/src/types/onyx/ReportNameValuePairs.ts#L57

I'm not sure if we have that information about statement card

src/libs/OptimisticReportNamesConnectionManager.ts
sosek108: Okey

src/libs/OptimisticReportNamesConnectionManager.ts
sosek108: I didn't found any place where we use initialize prop at all. The connections are made in the root of file and values are accessed in methods. 

I've modified this to check if all connections are initialized and chnaged this method to be async, so the core logic wait for data

tests/perf-test/OptimisticReportNames.perf-test.ts
sosek108: fixed

tests/unit/OptimisticReportNamesTest.ts
sosek108: removed

tests/unit/OptimisticReportNamesTest.ts
sosek108: I think that there is some kind of problem with static fields. Report disappears right after it's created.


https://github.com/user-attachments/assets/88d280e2-97da-4570-85b7-32de19df6024

This does not occur when non static formula is set or if offline mode is enabled

sosek108: At this moment I remove that check if field is formula

src/libs/OptimisticReportNames.ts
neil-marcellini: I would add a comment. I don't think we need an issue because it hasn't been proven to be a problem yet.

src/libs/OptimisticReportNames.ts
neil-marcellini: Ok, we can probably ignore that for now. I don't think it's common at all.

tests/perf-test/OptimisticReportNames.perf-test.ts
neil-marcellini: Did you push the fix? I still see an array with length 20 below.

tests/unit/OptimisticReportNamesTest.ts
neil-marcellini: Which check did you remove? Does it work now or no? We should support static formulas eventually, but it could be done in a follow up.

neil-marcellini: I do want to make sure we don't have a test asserting the wrong behavior, before we merge. So if we do this in a follow up, let's remove the test and lmk so I can create an issue to track it.

neil-marcellini (changes requested):
Ok looking pretty close. There are a few threads in this PR still unresolved, so let's make sure we address those or make a plan to do so in the future.

src/libs/Formula.ts
sosek108: I've got circular import error when imported funcs from `TransactionUtils`. I did try to resolve them here in `Formula.ts` file but without luck or it'd need to complicate some stuff.

I'd suggest that we remove the dependency on `TransactionUtils` (`getCreated` and `isPartialTransaction`) for now and address this in separate PR. 

I'd like to create `getOldestTransaction` on `ReportUtils` file but adding them here would bloat this PR. 

Proposal:
1. create an issue for creating `getOldestTransaction` in `ReportUtils`
2. replace `getCreated` and `isPartialTransaction` with simple logic in current file (Formula.ts) 
3. when PR from 1 is merged we can use that logic here and have consistent logic.

What do you think?

src/libs/OptimisticReportNames.ts
sosek108: Added a comment

tests/perf-test/OptimisticReportNames.perf-test.ts
sosek108: Yeah, I forgot about `git push` with last commit. Done.

tests/unit/FormulaTest.ts
sosek108: Ok, I've replaced test case with `{report:{report:submit:from:firstName|substr:2}}`. And we expect returned value `{report:{report:submit:from:firstName|substr:2}}`

tests/unit/FormulaTest.ts
sosek108: Done. I had to add mock for `getCurrencySymbol`

tests/unit/FormulaTest.ts
sosek108: Done

src/libs/Formula.ts
sosek108: Nevermind. I've to properly mock `ReportUtils`. I forgot about that

sosek108 (commented):
@neil-marcellini I've responded to all your comments. 

1. Fixed tests
2. nested braces are treated as invalid so the original definition is returned
3. if data is missing, definition is returned instead of empty string
4. currency is fixed `$` instead of `USD`
5. fixed spaces between formula parts

src/libs/OptimisticReportNames.ts
neil-marcellini: Can we please do this in a follow up?

tests/unit/FormulaTest.ts
```diff
@@ -43,14 +52,14 @@ describe('CustomFormula', () => {
     describe('parse()', () => {
         test('should parse report formula parts', () => {
             const parts = parse('{report:type} {report:startdate}');
-            expect(parts).toHaveLength(2); // report:type, report:startdate (space is trimmed)
+            expect(parts).toHaveLength(3); // report:type, report:startdate (space is trimmed)
```
neil-marcellini: NAB: The comment is incorrect now

neil-marcellini (dismissed):
Thanks for the updates, I think everything is resolved now. Next you'll have to fix the CI errors.

Also, please write a list of all the parts of the optimistic report formula generation that are not yet implemented, and I can create a series of issues to split up the work into manageable chunks.

tests/unit/FormulaTest.ts
sosek108: yeah, cleaning is in progress

sosek108:
Tests are failing, but these seems to be not related to my changes

sosek108:
@neil-marcellini could you please double check if my test cases are valid? I hope I don't miss anything

tests/unit/FormulaTest.ts
```diff
@@ -1,10 +1,13 @@
+// eslint-disable-next-line no-restricted-syntax -- disabled because we need ReportUtils to mock
```
neil-marcellini: This should say "we need CurrencyUtils to mock". Same for ReportActionsUtils below

src/libs/Formula.ts
```diff
@@ -431,11 +428,13 @@ function getOldestReportActionDate(reportID: string): string | undefined {
...
 
     Object.values(reportActions).forEach((action) => {
-        if (action?.created) {
-            if (!oldestDate || action.created < oldestDate) {
-                oldestDate = action.created;
-            }
+        if (!action?.created) {
+            return;
+        }
+        if (!(!oldestDate || action.created < oldestDate)) {
```
neil-marcellini: NAB: Can we please flip the logic in the parentheses vs negating the whole thing?

src/libs/OptimisticReportNames.ts
```diff
@@ -166,14 +169,14 @@ function computeReportNameIfNeeded(report: Report | undefined, incomingUpdate: O
         return null;
     }
 
-    const policy = getPolicyByID(targetReport.policyID ?? '', allPolicies);
+    const policy = getPolicyByID(targetReport.policyID ?? String(CONST.DEFAULT_NUMBER_ID), allPolicies);
```
neil-marcellini: Let's make getPolicyByID take a `string | undefined` to avoid defaulting like this.

tests/unit/FormulaTest.ts
neil-marcellini: Still needs to be fixed, you could remove the comment too

src/libs/Formula.ts
neil-marcellini: Coming back to this, can we please add a comment explaining it here. Mostly that the backend always creates a report with an action, so the report.created and oldest created action are the same? And note that we could return the created field to make it slightly more efficient in the future.

src/libs/Formula.ts
```diff
@@ -0,0 +1,506 @@
...
+
+    return oldestDate;
+}
+
+/**
+ * Check if a string contains formula parts
+ */
+function isFormula(str: string): boolean {
+    return extract(str).length > 0;
+}
```
neil-marcellini: This function is unused. We can remove it and all its tests.

tests/unit/OptimisticReportNamesTest.ts
```diff
@@ -0,0 +1,240 @@
...
+            const updates = [
+                {
+                    key: 'policy_policy1' as OnyxKey,
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    value: {name: 'Updated Policy Name'},
+                },
+            ];
+
+            const result = updateOptimisticReportNamesFromUpdates(updates, contextWithMultipleReports);
+            expect(result.length).toBeGreaterThan(1);
```
neil-marcellini: This test isn't very clear to me. What is the policy formula and does it reference the policy name? We should assert exactly how many results we expect and what the expected content is.

neil-marcellini (changes requested):
Looking really close, a few more small things and one test question

tests/unit/FormulaTest.ts
sosek108: Fixed

src/libs/Formula.ts
sosek108: Fixed

src/libs/OptimisticReportNames.ts
sosek108: Done

tests/unit/FormulaTest.ts
sosek108: Not sure why is this visible here, as this comment is removed on https://github.com/Expensify/App/pull/65862/commits/cb998e126d213d1820b36bd6bcf4ea5d6d216c00

sosek108:
@neil-marcellini I've checked that I've uploaded all screen recordings but I have problems with Android today, so I've uploaded recordings only from: 

- web
- desktop
- ios native

tests/unit/OptimisticReportNamesTest.ts
```diff
@@ -0,0 +1,249 @@
...
+            const result = updateOptimisticReportNamesFromUpdates(updates, contextWithMultipleReports);
+
+            expect(result.length).toEqual(4);
+            result.forEach((update) => {
+                if (!update.key.startsWith('report_')) {
+                    return;
+                }
+
+                expect((update.value as {reportName?: string})?.reportName).toEqual('Policy: Updated Policy Name');
+            });
```
neil-marcellini: NAB: The section seems to be asserting that there will be one update which is the original, followed by 3 updates, one per report, updating the name. I would rather assert individually about each update, checking its key and value. That's much more clear. Can be done in a follow up, but let's make sure it gets done please.

src/libs/OptimisticReportNames.ts
```diff
@@ -0,0 +1,332 @@
...
+ * Update optimistic report names based on incoming updates
+ * This is the main middleware function that processes optimistic data
+ */
+function updateOptimisticReportNamesFromUpdates(updates: OnyxUpdate[], context: UpdateContext): OnyxUpdate[] {
+    Performance.markStart(CONST.TIMING.UPDATE_OPTIMISTIC_REPORT_NAMES);
+    Timing.start(CONST.TIMING.UPDATE_OPTIMISTIC_REPORT_NAMES);
+
+    Log.info('[OptimisticReportNames] Processing optimistic updates for report names', false, {
+        updatesCount: updates.length,
+        hasReports: Object.keys(context.allReports).length > 0,
```
neil-marcellini: NAB: I don't think `hasReports` is that useful, because the context will be initialized before getting to this. Also, maybe we can log after checking for the beta. Otherwise the log is kind of meaningless.

neil-marcellini (dismissed):
Thanks for the updates. We have a lot left to do in order to complete the feature, but this will be a great start!

sosek108:
Undrafted

akinwale:
@neil-marcellini Can access to the `useCustomReportNamesNewExpensify` beta be granted to my test account, `akinwale+expensify64@gmail.com`?

akinwale:
@sosek108 A couple of notes after testing:

1. For Test Case 1, the report title formula results in an error stating that the character limit (80) was exceeded.

Screenshot 2025-08-13 at 20 26 13

2. For Test Case 2, the report created field shows the date and time while the report start date has the date part only. I imagine we want to keep this consistent.

Screenshot 2025-08-13 at 20 28 32

Every other test case works fine.

sosek108:
> For Test Case 2, the report created field shows the date and time while the report start date has the date part only. I imagine we want to keep this consistent.

@neil-marcellini this seems to be coming from backend. What'd you think?

neil-marcellini:
> For Test Case 2, the report created field shows the date and time while the report start date has the date part only. I imagine we want to keep this consistent.

Interesting. Let's see how it compares to OldDot. We're just trying to match the functionality there.

OldDot does the same thing so I don't think it's a problem, at least not a significant one. All good.

image

neil-marcellini (approved):
Thanks for the updates

akinwale:
@sosek108 In offline mode, the Save button gets stuck loading after I click on it, so I am unable to change the formula afterwards (see end of video). Is this expected behaviour?

https://github.com/user-attachments/assets/4792eb02-5cad-4110-8e2f-a7dce54d0f40

akinwale (approved):
LGTM.

neil-marcellini (approved):
Sweet, let's do this. Thanks for the hard work guys.

PR #65897, opened 2025-07-10
[NoQA] Revert "Fix: Reset Background to Workspace on Page Reload in Track Expense"

koko57:
Reverts Expensify/App#65514

It breaks the navigation handling in auto-assigning flow of Create Expense - after auto-assigning the active policy and then changing the assignment in the confirmation step it doesn't get back to the Scan step but to the Participants Step

Current behavior:

https://github.com/user-attachments/assets/3c123ab1-d77a-4d92-8f58-52863ed28bb2

Before introducing the change:

https://github.com/user-attachments/assets/5a281792-a334-4928-98fc-156ab060cb2a


mountiny (approved):
Straight revert

mountiny:
Straight revert

Issue #66814, opened 2025-07-21
Expense - Report preview is empty after reset app


jponikarchuk:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.1.84-1
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from TestRail:** Exp
**Email or phone of affected tester (no customers):** dsjnoisjsojsdo@gmail.com
**Issue reported by:** Applause Internal Team
**Device used:** Mac 15.5 / Chrome
**App Component:** Money Requests

## Action Performed:
1. Go to staging.expensify.com
2. Open 1:1 DM.
3. Submit an expense to the user.
4. Go to Account > Troubleshoot > Clear cache and restart > Reset and refresh app.
5. Go back to Inbox and open the same 1:1 DM in Step 2.

## Expected Result:
The report preview will show the expense.

## Actual Result:
The report preview is empty after reset app.

## Workaround:
Unknown

## Platforms:
- [x] Android: App
- [x] Android: mWeb Chrome
- [x] iOS: App
- [x] iOS: mWeb Safari
- [x] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari
- [x] MacOS: Desktop

## Screenshots/Videos

https://github.com/user-attachments/assets/1f06c01d-3149-49b7-8a95-423457a3f273

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

rushatgabhane:
@blamegpt which PRs caused

amyevans:
Hmm I can reproduce the bug on staging against both the staging and production API, so I don't _think_ it's due to a backend code change. Will keep digging.

rushatgabhane:
probably this? https://github.com/Expensify/App/pull/66741

ZhenjaHorbach:
> probably this? [#66741](https://github.com/Expensify/App/pull/66741)
> 
> it's referenced in other blockers

How the fix for split flow can be related with report preview ?

amyevans:
> I suspect it's coming from here: https://github.com/Expensify/App/pull/66719

[...]

PR #66820, opened 2025-07-21
Revert "Optimize ReportSrcren and connected components via using strategic"

amyevans:
Reverts Expensify/App#66719


amyevans:
Was testing as a possible cause of https://github.com/Expensify/App/issues/66814

PR #69036, opened 2025-08-22
Revert "fix: Navigation bar not translucent in modals on Android"

blimpich:
Reverts Expensify/App#69004, which I merged without C+ review because I thought it was a low-risk style change, but since its a patch we should probably be more careful.


blimpich:
holding as we might not need to revert

rlinoz (approved):

blimpich:
Didn't wait for checks to pass since it's a straight revert

PR #69719, opened 2025-09-03
Write down a pagination philosophy

roryabraham:
### Explanation of Change
Pagination philosophy.

### Fixed Issues
Fixes https://github.com/Expensify/Expensify/issues/532765

### Tests
None.

### Offline tests
None.

### QA Steps
None

### PR Author Checklist
    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
    - [X] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


roryabraham:
Opening this up for review, but it's low-priority so no rush on the review, and you may expect slower responses.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Pagination vs. lazy-loading
+As stated above, pagination is a specialized subset of lazy-loading for ordered lists of data. The advantage of structured pagination over a more naïve lazy-loading strategy is that it divides data into discrete chunks such that they can be loaded in batches (and that the order of those batches lends itself to a better product experience).
+
+One example of data that's appropriate for pagination is reportActions (chat messages). We can load the 50 most recent messages, then scroll back and send one network request to get the next batch of 50.
+
+One example of data that's appropriate to lazy-load but not necessarily ideal for pagination is avatar images. It's too costly to load all avatar images for all users when the app loads, so we only fetch avatar images that a user wants to see on-demand. Another example might be `personalDetails`.
+
+## How to implement pagination
+1. Clearly define a deterministic sorting order of the data in question.
```
marcaaron: NAB bc probably does not need to be pointed out, but are we admitting that we are kind of "without options" philosophy/guidance wise if a list of data is not deterministically sorted? I can only think of one example of course and you know which one it is 😄

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+2. Define a "cursor" - a unique piece of data that encapsulates all the data you need to pinpoint the location of a single item in the list.
+    - Each item in the list _MUST_ be unique. No ties allowed: typically this means your cursor should include an ID.
```
marcaaron: ties?

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+    - The cursor _MUST_ be serializable so it can be sent in a network request.
```
marcaaron: What would an example of a non-serializable cursor be?

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+        type ReportActionCursor = {
+            /** The datetime when the reportAction was created, in ms precision. */
+            created: string;
+
+            /** The actionName is used as a tie-breaker when multiple reportActions are created in the same millisecond (i.e: CREATED actions always come first). */
+            actionName: ReportActionName;
+
+            /** The reportActionID is used as a tie-breaker when multiple reportActions of the same type are created in the same millisecond. */
+            reportActionID: string;
+        }
```
marcaaron, line 48: Are we currently using a cursor like this anywhere or are we planning to?

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+        ```
+
+    - _Example:_ For a single search snapshot, an integer `sequenceNumber` is used. This is a simple implementation, but generally is not a scalable solution, because it requires that all the data you're paging over needs to be sorted before you can access it by an integer offset and limit.
```
marcaaron: Maybe better to say some variation of: 

>offset-based approaches require the database to scan and discard all prior rows up to the offset, which becomes slow for large dataset pagination

not sure if that is exactly the problem, but I think it would help to explain the exact problem as we have experienced it and be specific about why `sequenceNumber` bad.

marcaaron: Also, maybe define what is cursor and offset somewhere. I think they are easy to confuse and not exactly the same, but we should make clear the difference and which one we prefer.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+    - The list doesn't live-update (i.e: if a user scrolls back in the list, and while they're looking at older items, they're not receiving newer items as they come in).
+
+    If you need bidirectional pagination, you'll need to ensure that you craft queries and API endpoints to fetch data from a given cursor in both directions. In the example query above, it could be as simple as switching `<` to `>=` and making the comparator and sort order explicit for each direction.
+
+> [!NOTE]
+> At the time of writing, [RecyclerListView](https://github.com/Flipkart/recyclerlistview) and [FlashList](https://github.com/Shopify/flash-list) _do not_ support bidirectional pagination.
+> React Native's built-in [FlatList](https://reactnative.dev/docs/flatlist) and [legend-list](https://github.com/LegendApp/legend-list) are some examples that _do_ support bidirectional pagination via the `onStartReached` prop.
+
+5. Determine whether it's possible for **gaps** to appear in your data. It's non-trivial to list all the ways gaps can appear in a list, but here are a couple of examples, one real and one contrived:
```
marcaaron: Maybe a terminology section would be useful? Let's define what a "gap" is there.

marcaaron:
```suggestion
5. Determine whether it's possible for **gaps** to appear in your data. Here are some examples to help illustrate what a "gap" is and how we might end up with one:
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+    - "Comment linking"
+        1. User has a simple paginated list of integers in ascending order, and it currently contains items 1-50.
```
marcaaron:
```suggestion
        1. User has a simple paginated list of integers in ascending order, and the frontend has loaded items 1-50.
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+        2. User jumps to the middle of the list (say for simplicity that they are looking at items 15-35)
```
marcaaron:
```suggestion
        2. User jumps to the middle of the list and the viewport contains items 15-35
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+        3. While they're looking at the middle of the list, more than 1 page (50 items) of data is added to the front of the list. Let's say these items are items 50-150.
+        4. They jump back to the front of the page, and fetch the page at the front, 100-150.
```
marcaaron: Visual aid could be good here? The language around middle, front, back to the front, fetch at the front etc got confusing for me.

marcaaron:
```suggestion
        4. They jump to the front of the list and then fetch page 100-150.
```

marcaaron: In lieu of a complete explanation of "comment linking" we could also say "They deep link to a comment in the front" or something like that. It might not be clear what it means to "jump".

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+            - _Note:_ Fetching "all the data they missed" generally isn't a scalable solution, because it's unbounded, and the unbounded loading of data is the problem pagination seeks to solve.
+        5. Now there is a gap between items 51-100 :boom:
+    - "Over-eager eviction"
+        1. A user scrolls far back in the list. So far that we decide there's too much data for us to handle all at once. Performance degrades, and we decide to start evicting the data at the front that's less-recently viewed. Let's say that the user now has items 50-100 loaded, and we've evicted items 101-200.
+        2. New data appears at the front of the list, say items 201-250. We add these items to the list as we normally would if they're looking at the front of the list.
+        3. Now there is a gap between items 101-200 (the data we evicted) :boom:
+
+    If it turns out you _do_ need a strategy for gap detection, here's a high-level summary of how you'd handle it:
```
marcaaron: Have we thought about what data types we have that will might require gap detection? I think this write up is really great, but mainly curious if we think we will face a similar problem again.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+    1. Start keeping track of the pages you've loaded in a dedicated Onyx key for pagination (e.g., `report_123_actions_pages`). This is a _sorted_ list of ranges.
+    2. Post-process network responses using the [Pagination middleware](https://github.com/Expensify/App/blob/1a06fa4add10b53a1a9266927d3b08a4ca35d3c4/src/libs/Middleware/Pagination.ts) to keep track of the start and end point of the page you loaded in a request, and [merge it with existing pages if it overlaps](https://github.com/Expensify/App/blob/1a06fa4add10b53a1a9266927d3b08a4ca35d3c4/src/libs/PaginationUtils.ts#L104).
+    3. When rendering your list, use that pages key for your sorting order, and insert "gap markers" between the edges of the pages you've loaded.
+    4. When rendering your list, [only render a single continuous chunk](https://github.com/Expensify/App/blob/1a06fa4add10b53a1a9266927d3b08a4ca35d3c4/src/libs/PaginationUtils.ts#L166) containing your current "anchor point" (the reportAction you linked to, for example), up until you reach the end of the list in either direction or a gap marker.
+    5. Then when you scroll, you'll hit your gap marker and can make network requests to fill in the gap.
```
marcaaron: There is a lot of terminology here like "single continuous chunk" and "gap markers". I'm not sure if they need to be explained here or if the Pagination middleware stuff is documenting it better, but this explanation felt almost too generic to be useful. I think we should either simplify it, add more details, or just direct people to the code itself for anyone who needs to understand further.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+    More details can be found in [the Pagination middleware](https://github.com/Expensify/App/blob/1a06fa4add10b53a1a9266927d3b08a4ca35d3c4/src/libs/Middleware/Pagination.ts). Efforts were made to generalize this code, but so far it has only been used for reportActions.
+
+## Future improvements
+The following ideals have not (yet) been implemented in Expensify, but we'll hold them out here as aspirational goals for future pagination systems, inspired by WhatsApp.
```
marcaaron: I like the insights here, but a part of me is 👎 on adding aspirational things to a doc meant to explain the current best practices. Feels like something that should go in a P/S or a design doc and then eventually we can update this. For that reason, I'd suggest deleting everything below this comment.

marcaaron (commented):
Great read. Not sure why I added so many comments. I think I really enjoyed it 😄

contributingGuides/philosophies/PAGINATION.md
chuckdries: > not a scalable solution, because it requires that all the data you're paging over needs to be sorted before you can access it

This is a little confusing IMO because the very first bullet in this section is `1. Clearly define a deterministic sorting order of the data in question`, which seems contradictory. How can "requiring all data to be sorted" be a downside to this approach when you've already established that all the data has to be sortable

contributingGuides/philosophies/PAGINATION.md
luacmartins: Makes sense, we wouldn't for example use just a timestamp as a cursor because it's possible for two items to have the same timestamp which would lead to an unpredictable result

contributingGuides/philosophies/PAGINATION.md
luacmartins: Another issue of the limit/offset approach is that it leads to gaps/duplicates in the list if items are added/removed from the previous page.

luacmartins: I think this is what Marc said above in a clearer way:

>offset-based approaches require the database to scan and discard all prior rows up to the offset, which becomes slow for large dataset pagination

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+4. Determine whether your use-case calls for **unidirectional** or **bidirectional** pagination. Generally, **bidirectional** pagination will be useful if:
+
+    - The list live-updates and the user can "jump" to an arbitrary point in the list without loading all the data between the start of the list and that point.
+    - The list doesn't live-update (i.e: if a user scrolls back in the list, and while they're looking at older items, they're not receiving newer items as they come in).
+
+    If you need bidirectional pagination, you'll need to ensure that you craft queries and API endpoints to fetch data from a given cursor in both directions. In the example query above, it could be as simple as switching `<` to `>=` and making the comparator and sort order explicit for each direction.
+
+> [!NOTE]
+> At the time of writing, [RecyclerListView](https://github.com/Flipkart/recyclerlistview) and [FlashList](https://github.com/Shopify/flash-list) _do not_ support bidirectional pagination.
```
luacmartins: Note sure if we need to include these specific libraries in the philosophy
```suggestion
> At the time of writing, [RecyclerListView](https://github.com/Flipkart/recyclerlistview) and [FlashList](https://github.com/Shopify/flash-list) _do not_ support bidirectional pagination.
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+> React Native's built-in [FlatList](https://reactnative.dev/docs/flatlist) and [legend-list](https://github.com/LegendApp/legend-list) are some examples that _do_ support bidirectional pagination via the `onStartReached` prop.
+
+5. Determine whether it's possible for **gaps** to appear in your data. It's non-trivial to list all the ways gaps can appear in a list, but here are a couple of examples, one real and one contrived:
+    - "Comment linking"
+        1. User has a simple paginated list of integers in ascending order, and it currently contains items 1-50.
+        2. User jumps to the middle of the list (say for simplicity that they are looking at items 15-35)
+        3. While they're looking at the middle of the list, more than 1 page (50 items) of data is added to the front of the list. Let's say these items are items 50-150.
```
luacmartins: How does this happen?

contributingGuides/philosophies/PAGINATION.md
luacmartins: +1

contributingGuides/philosophies/PAGINATION.md
luacmartins: I agree!

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+It makes sense to lazy-load a resource when the volume of that resource is unbounded, or the upper bound of the resource is too costly to load all at once.
+
+## Pagination vs. lazy-loading
+As stated above, pagination is a specialized subset of lazy-loading for ordered lists of data. The advantage of structured pagination over a more naïve lazy-loading strategy is that it divides data into discrete chunks such that they can be loaded in batches (and that the order of those batches lends itself to a better product experience).
+
+One example of data that's appropriate for pagination is reportActions (chat messages). We can load the 50 most recent messages, then scroll back and send one network request to get the next batch of 50.
+
+One example of data that's appropriate to lazy-load but not necessarily ideal for pagination is avatar images. It's too costly to load all avatar images for all users when the app loads, so we only fetch avatar images that a user wants to see on-demand. Another example might be `personalDetails`.
```
tgolen: We don't really lazy-load personal details though, do we? We just get all of them all at once and shove them into Onyx?

contributingGuides/philosophies/PAGINATION.md
tgolen: Can you please make the qualifier words (ie. "MUST", "SHOULD", etc.) non-italicized? Just for consistency with the other docs.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+## How to implement pagination
+1. Clearly define a deterministic sorting order of the data in question.
+2. Define a "cursor" - a unique piece of data that encapsulates all the data you need to pinpoint the location of a single item in the list.
+    - Each item in the list _MUST_ be unique. No ties allowed: typically this means your cursor should include an ID.
+    - The cursor _MUST_ be serializable so it can be sent in a network request.
+    - The cursor _SHOULD_ include only the minimal fields required to define the location of an item in the list.
```
tgolen: Another note on consistency with the other docs is that there is usually a "Rules" section of the page, and that's where all the SHOULD/MUST rules go.

I think your explanations and examples are pretty good, but I suggest adding a Rules section to the bottom that clearly lists all the things that should or should not be done.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+        ORDER BY created DESC, id DESC
+        LIMIT 50;
+        ```
+
+4. Determine whether your use-case calls for **unidirectional** or **bidirectional** pagination. Generally, **bidirectional** pagination will be useful if:
+
+    - The list live-updates and the user can "jump" to an arbitrary point in the list without loading all the data between the start of the list and that point.
+    - The list doesn't live-update (i.e: if a user scrolls back in the list, and while they're looking at older items, they're not receiving newer items as they come in).
+
+    If you need bidirectional pagination, you'll need to ensure that you craft queries and API endpoints to fetch data from a given cursor in both directions. In the example query above, it could be as simple as switching `<` to `>=` and making the comparator and sort order explicit for each direction.
```
tgolen: To keep the language consistent in the docs, don't address the user as "you". Instead, reword these like:

> If bidirectional pagination is needed, ensure that queries and API endpoints are crafted to fetch data from a given cursor in both directions.

You did this for the most part in this doc, but there are a couple of spots here (and in point 4 above) where this slipped.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,113 @@
...
+
+However, the systems we've built and/or described have no data eviction. Once data has been downloaded from the server, it's saved to disk, and because of the design of Onyx all that data is typically kept in RAM in the Onyx cache. If this starts to become a problem ([it has been discussed before](https://expensify.slack.com/archives/C05LX9D6E07/p1755077335799619)), throwing away data on disk is not the best approach. Instead, we should build a two-layer pagination system:
+
+1. Data is loaded into RAM one page at a time.
+2. If we reach the edge of the page, we first check the disk for more.
+3. Only if the page we requested is not on disk, we request it from the server.
+4. If data must be evicted due to slow performance, it is first evicted from RAM.
+5. Data should be evicted from disk only if reading data from disk becomes a bottleneck, or we reach device storage limits.
+
+That way, we can preserve our [Offline Philosophy](https://github.com/Expensify/App/blob/main/contributingGuides/philosophies/OFFLINE.md) and provide a first-class offline UX without compromising performance.
```
tgolen: FYI these links can be shorted to 
```suggestion
That way, we can preserve our [Offline Philosophy](/contributingGuides/philosophies/OFFLINE.md) and provide a first-class offline UX without compromising performance.
```

Which is better if the repo is ever renamed

contributingGuides/philosophies/PAGINATION.md
tgolen: I don't feel too strongly about it. I see your point but I also think this gives _some_ small guidance into the future and that we don't necessarily want to do anything that would really hinder us from going in this direction.

A lot of our original philosophies were very aspirational at the time, like the server pushing data to a client's disk. In the early stages, we were only doing that for reportActions and all other data was loaded with GET requests. Now, we let the server push any and all data to clients via `Onyx.update()`.

tgolen (changes requested):
Just a few early suggestions. I'm sure I'll have more later. This is good stuff!

contributingGuides/philosophies/PAGINATION.md
marcaaron: We could address that concern by using some strong language and mentioning specifically that we should never move forward with any solution that involves throwing away data on disk since a future solution might rely on it and point to the slack discussion.

My main concern is that we are including specifics about what we might want to build without a clearly defined problem. Putting solutions before problems and getting too attached to specific solutions are typically discouraged at the org level and I think we should model those values here.

luacmartins: I agree with Marc's comments above

contributingGuides/philosophies/PAGINATION.md
roryabraham: That's correct. This is held out as an example of something that might be appropriate to lazy-load, but not to paginate.

contributingGuides/philosophies/PAGINATION.md
roryabraham: I should clarify that "deterministic" doesn't mean that the order can't change if the data changes. It just means that if the same data is sorted, the resultant sort order is _exactly_ the same each time.

To explain with an example, the JS array `.sort` function takes a `compareFunction` that's called with two args, `a` and `b`. The return value of `compareFunction` is:

- negative if `a` should come before `b`
- positive if `a` should come after `b`
- zero if `a` and `b` are considered equal

Our sorting order must never return zero - no two items are allowed to be equal.

roryabraham: I re-wrote this a bit and I think it's clearer now

contributingGuides/philosophies/PAGINATION.md
roryabraham: Clarified this a bit

roryabraham: done

contributingGuides/philosophies/PAGINATION.md
roryabraham: a function cannot be included in a cursor, because it's not serializable. Updated the docs to call out this example

contributingGuides/philosophies/PAGINATION.md
roryabraham: We are not explicitly using this cursor. We use a combination of `created` and `reportActionID`. But we use the `actionName` in the sorting order

roryabraham: I hold this out as a better example, because it includes every piece of data that's used for reportAction sorting.

contributingGuides/philosophies/PAGINATION.md
roryabraham: updated, thanks for clarifying @marcaaron that's much better. Though I didn't have a super clear explanation for the secondary problem @luacmartins described, so I didn't include it.

roryabraham: I think we made this clearer 👍🏼 

@chuckdries the problem with sequenceNumbers is basically that you need to query everything before the offset in order to grab a single page.

contributingGuides/philosophies/PAGINATION.md
roryabraham: Removed usages of the 2nd person

contributingGuides/philosophies/PAGINATION.md
marcaaron: I think (mostly to avoid confusion) we should put whatever we currently do here. Or possibly use an arbitrary object as an example.

contributingGuides/philosophies/PAGINATION.md
luacmartins: Gap
1. We fetch transactions 1-50
2. Transaction 1 is deleted
3. Transaction 51 becomes 50
4. We fetch the next page, i.e. transactions 51-100
5. Transaction from step 3 is never returned

Duplicate
1. We fetch transactions 1-50 
2. A new transaction is inserted at index 50. The previous transaction 50 is now transaction 51
3. We fetch the next page 51-100
4. The (now) transaction 51 shows up twice in the list

luacmartins (commented):
Conflicts

contributingGuides/philosophies/PAGINATION.md
roryabraham: I do say "A better cursor would have been...", do you still think that's too confusing?

contributingGuides/philosophies/PAGINATION.md
roryabraham: updated

contributingGuides/philosophies/PAGINATION.md
roryabraham: I think we'd see it with reports if we wanted the LHN to show the report you have open. i.e: if you deep-link to a report, and we want that report to be highlighted in the LHN.

contributingGuides/philosophies/PAGINATION.md
roryabraham: removed the aspirational section and instead just included some guidelines on not evicting data

contributingGuides/philosophies/PAGINATION.md
roryabraham: done

contributingGuides/philosophies/PAGINATION.md
roryabraham: done

roryabraham:
ok, I've made a bunch of updates here and this should be ready for another round of review

contributingGuides/philosophies/PAGINATION.md
luacmartins: We load them on demand on the Reports page ;P

luacmartins: The comment above made me think that we might also need a section for paginated data dependencies. For example, the Reports page loads one page of transactions at a time, but for those given transactions we also need to load a set of dependencies, e.g. reports, reportActions, violations, etc as part of the same command. The dependencies are all loaded on demand based on the page results

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+
+            /** The reportActionID is used as a tie-breaker when multiple reportActions of the same type are created in the same millisecond. */
+            reportActionID: string;
+        }
+        ```
+
+    - _Example:_ For a single search snapshot, an integer `sequenceNumber` is used. This is a simple implementation, but generally is not a scalable solution, because offset-based approaches require the database to scan and discard all prior rows up to the offset, which becomes slow for large dataset pagination. Furthermore, offset-based implementations are prone to gaps and/or duplicates:
+
+        1. We fetch transactions 1-50
+        2. Transaction 1 is deletedg
```
luacmartins:
```suggestion
        2. Transaction 1 is deleted
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+        3. Transaction 51 becomes 50
+        4. We fetch the next page, i.e. transactions 51-100
+        5. Transaction That was 51 and is now 50 is never returned
```
luacmartins:
```suggestion
        5. The transaction that was 51 and is now 50 is never returned
```

contributingGuides/philosophies/PAGINATION.md
tgolen: OK, I would suggest leaving out the mention of personal details here. Even if it's something that _could_ be lazy-loaded, the fact that we aren't just makes it an odd example to point to.

```suggestion
One example of data that's appropriate to lazy-load but not necessarily ideal for pagination is avatar images. It's too costly to load all avatar images for all users when the app loads, so we only fetch avatar images that a user wants to see on-demand.
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Pagination: Loading a large amount of ordered data page-by-page, rather than all at once.
+- Cursor: A "pointer" to a given item in the list, which has all the data needed to pinpoint the location of that item in the list.
+- Jump: In the context of this doc, "jump" means to go immediately to a different point in the list without scrolling.
+    - Example: Click on a link to a comment, and you are taken straight there, without scrolling.
+- Unidirectional Pagination: Pagination where the user starts loading at one end of the list, and can only scroll and load pages in one direction.
+    - Example: Report search page.
+- Bidirectional Pagination: Pagination where the user can "jump" to any point in the list, and then can scroll and load pages in either direction.
+    - Example: A long chat report, opened by deep-linking to an old message.
+- Gap: Items missing between pages (can only happen under specific circumstances)
```
tgolen, line 28: Could you please add lazy-loading here as well?

contributingGuides/philosophies/PAGINATION.md
tgolen: I agree with Marc that in these examples, it's best to show exactly what we are doing now. Otherwise, it takes extra explanations and is confusing about what we want them to do.

The purpose of these docs is to communicate our best practices, and if we add a best-practice that is beyond what our current code is doing, then that is a sign that we need to go and update that code to use the best-practice.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+    3. When rendering the list, use that pages key for the sorting order, and insert "gap markers" between the edges of the pages that have been loaded.
+    4. When rendering the list, [render only a single continuous chunk](/src/libs/PaginationUtils.ts#L166) containing the current "anchor point" (the reportAction linked to, for example), up until reaching the end of the list in either direction or a gap marker.
+    5. Then, when scrolling, the gap marker will be reached and network requests can fill the gap.
+
+    More details can be found in [the Pagination middleware](/src/libs/Middleware/Pagination.ts). Efforts were made to generalize this code, but so far it has only been used for reportActions.
+
+## Don't evict data
+The systems we've build and/or described do not have any data eviction mechanisms, and that's an intentional choice. Reject any solutions that evict data from disk, unless reading or writing from disk is clearly proven to be a bottleneck. That way, we can preserve our [Offline Philosophy](/contributingGuides/philosophies/OFFLINE.md) and provide a first-class offline UX without compromising performance.
```
tgolen: Hm, given this point, it makes me want to remove the entire point 5 above about detecting gaps, which only happens when data is evicted. Again, if we aren't doing it today in the code, then it's not a best-practice.

tgolen: It also feels like this point should be moved to the rules below.

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Rules
+
```
tgolen, line 108: It would be nice to have some rules in here like:
- Data SHOULD be paginated when...

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+- The cursor for each item in the list MUST be unique.
+- The cursor MUST be serializable so it can be sent in a network request (no functions).
+- The cursor SHOULD includes only the minimal fields required to define the location of an item in the list.
```
tgolen:
```suggestion
- The cursor SHOULD include only the minimal fields required to define the location of an item in the list.
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+- There MUST be a fast way query to get arbitrary page of data in the database.
```
tgolen:
```suggestion
- Fetching an arbitrary page of results in the database MUST be fast
```

contributingGuides/philosophies/PAGINATION.md
```diff
@@ -0,0 +1,119 @@
...
+- If gap detection is needed, the existing Pagination Middleware SHOULD be used to detect and fill gaps.
```
tgolen:
```suggestion
- Pagination Middleware SHOULD be used to detect and fill gaps when gap detection is necessary
```

tgolen (changes requested):

contributingGuides/philosophies/PAGINATION.md
roryabraham: moved to rules, removed the example of eviction as a way that gaps can appear.

contributingGuides/philosophies/PAGINATION.md
roryabraham: done!

contributingGuides/philosophies/PAGINATION.md
roryabraham: removed the mention of personalDetails!

contributingGuides/philosophies/PAGINATION.md
roryabraham: It doesn't translate to this minimal example, but in practice it happens because other users send messages in the chat report.

contributingGuides/philosophies/PAGINATION.md
roryabraham: I agree that would be nice to have, but I'm not sure what to say to codify that, since we only have a few examples of paginated data

roryabraham:
Applied some more suggestions, feel free to take another look.

luacmartins (approved):

tgolen (approved):

contributingGuides/philosophies/PAGINATION.md
marcaaron: My thoughts on this have not really changed, but don't feel a strong need to block this PR. Generally speaking, it seems undesirable to have docs that contradict or reference things that either no longer exist or never did in practice.

contributingGuides/philosophies/PAGINATION.md
marcaaron: This seems covered by the introductory explanation of pagination earlier in the doc. I can't think of much more we'd want to say beyond this:

>It makes sense to lazy-load a resource when the volume of that resource is unbounded, or the upper bound of the resource is too costly to load all at once.

Maybe it's worth repeating here.

Issue #67823, opened 2025-08-04
[Due for payment 2025-10-30] [HOLD 72611][Due for payment 2025-10-03] Require title field rNVP to compute optimistic report names


neil-marcellini:
# Problem
As mentioned in [this PR comment](https://github.com/Expensify/App/pull/65862/files#r2252733222), when the frontend optimistically computes report names, it should only do so for reports that already have the title field set in the report name value pairs. We will need to modify optimistic report creation in app to copy over the title field from the policy, and also do that for other actions such as changing a report's policy. The reason for this is that the lack of the field is used to signify that the report has been manually renamed, and is used to prevent the name from being recomputed and overriding the manually set name.

Without implementing such logic, the following case will be failing. If report names can be changed, a manually renamed report retains its manual name when the policy report name formula is updated

<img width="1512" height="863" alt="Image" src="https://github.com/user-attachments/assets/e35e55da-ec7d-47a3-b21a-0897c93030bd" />

# Solution
Implement the logic in the [updateTitleFieldToMatchPolicy function](https://docs.google.com/document/d/1Xq-QffDBIPtbjWJiSl9RE91sS17W2lS96Ltady3n2BQ/edit?usp=sharing) when reports are created or move policies. Check if the title field rNVP exists before optimistically computing report names. cc @sosek108

<details><summary>Issue Owner</summary>Current Issue Owner: @bfitzexpensify</details>

dubielzyk-expensify:
Anything needed from my end here?

neil-marcellini:
Asking if @sosek108 can branch off his initial PR and start on this now

sosek108:
[Draft PR](https://github.com/Expensify/App/pull/69605)

neil-marcellini:
Same, updates are in Slack

sosek108:
### Update

I've pushed a lot of changes. The PR became bigger so we can think of splitting this up if there is need for that - I think we can decide later on that.

While preparing changes for use case **Change workspace** I've encountered two roadblocks:
1. we did not accept updates for `reportNameValuePairs` - **so I added this to objects that we process** 
2. we processed updates one by one forgetting about previous changes, and with our case we had this situation:
    a. update for rNVP with new title field - process name
    b. update for report with new policy id - process name again
    c. update for report with `lastVisited` - process name again.
So I implemented a something similar to cache, where we store all objects updated in one `prepareRequest` so each update knows about previous results.

With that given I've successfully implemented these use cases:
[...]

PR #70870, opened 2025-09-19
OptimisticReportNames: Working updates cache

sosek108:
### Explanation of Change

This PR is a folloup to https://github.com/Expensify/App/pull/70817 and should be merged after. 

This PR introduces subsequent updates cache used to store information about all processed updates. Helps handle cases where integrity of updates is crucial for proper name computation.

#### Why we need that 
Example: 
When we manually rename report there are two subsequent updates sent to Onyx:
1. update `reportNameValuePair` clearing `expensify_text_title` field
2. update `report` with new reportName 

First update is processed and our system sees that `reportNameValuePair` changed. Then second update is processed, our system sees `report` is changed, but it does not see changes form 1 step -> it thinks that it should recompute name of report. This lead to an issue when you manually change report's name it will stay with previous value. 

Other expample:
When we move report from policy `A` to policy `B` there are two updates sent to Onyx queue:
1. update report’s nameValuePair object with new `expensify_text_title`
2. update report’s policyId.
When first update is processed it works properly, but when second update is processed, we need to know about first update so we won’t replace with old title.

#### Changes in this PR

Additional changes provided with this PR:
1. we add `reportNameValuePairs` to updates handled by `OptimisticReportNames` - this is needed for cases, where rNVP is changed due to chnages in `expensify_text_title` key
2. we refactor `OptimisticReportNames.updateOptimisticReportNamesFromUpdates` for better code quality, less duplication and better performance. At first we create a set of affected reports and then compute names only once per report. 
3. handle `updateReportName` case. In this situation we remove `expensify_text_title` from report’s rNVP
4. This updates repairs test case `should detect new report creation and add name update` mentioned in point 1️⃣

### Fixed Issues

$ https://github.com/Expensify/App/issues/67823
PROPOSAL : https://github.com/Expensify/App/issues/67823

### Tests

Keep in mind that you have to have enabled BETA feature for optimistic report name computation like in [manual](https://docs.google.com/document/d/1q6T1rOSgGu6H29a5TdnW5obvvkAiGE5QeDtlayWNxw8/edit?tab=t.0#heading=h.nhtsmauf72jq)

1. Go to settings and turn on Force offline mode
2. Go to Inbox and click on FAB -> Create report
3. Manually edit report's name
4. Ensure that name of report is changed to desired value

### Offline tests

See `Tests` category

### QA Steps

Keep in mind that you have to have enabled BETA feature for optimistic report name computation like in [manual](https://docs.google.com/document/d/1q6T1rOSgGu6H29a5TdnW5obvvkAiGE5QeDtlayWNxw8/edit?tab=t.0#heading=h.nhtsmauf72jq)

1. Go to settings and turn on Force offline mode
2. Go to Inbox and click on FAB -> Create report
3. Manually edit report's name
4. Ensure that name of report is changed to desired value

### PR Author Checklist

  - [x] Android: Native
  - [x] Android: mWeb Chrome
  - [x] iOS: Native
  - [x] iOS: mWeb Safari
  - [x] MacOS: Chrome / Safari
  - [x] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/Formula.ts
```diff
@@ -482,8 +487,16 @@ function getOldestTransactionDate(reportID: string, context?: FormulaContext): s
 
     transactions.forEach((transaction) => {
         // Use updated transaction data if available and matches this transaction
-        const currentTransaction = context?.transaction && transaction.transactionID === context.transaction.transactionID ? context.transaction : transaction;
 
+        // FormulaContext transaction is the most current, so it takes priority
```
sosek108, line 493: We have to ensure, that transaction that we're trying checking is most up to date. If there were any previous updates to transaction the cached transaction is the most up to date.

sosek108:
@ZhenjaHorbach not yet, thanks. I forgot to change button to `Create draft`, sorry about that

sosek108:
Spellcheck fails, but this is from file that is not changed in our PR `docs/articles/Unlisted/Compliance-Documentation.md`

src/libs/OptimisticReportNamesCache.ts
```diff
@@ -0,0 +1,108 @@
...
+            const currentValue = workingUpdates[key] ?? getOriginalValueByKey(key, context);
+            return {
+                ...workingUpdates,
+                [key]: {...currentValue, ...(update.value as WorkingUpdateValue)} as WorkingUpdateValue,
+            };
+        }
+        case Onyx.METHOD.CLEAR: {
+            const {[key]: removedItem, ...rest} = workingUpdates;
+            return rest as WorkingUpdates;
+        }
```
neil-marcellini: I don't think we need to support this Onyx method because it's not one that we set in Onyx updates as it generally removes everything from storage. It's really only used for signing out AFAIK.

neil-marcellini: Check out the [definition here](https://github.com/Expensify/react-native-onyx/blob/09c529e90f2b6574772598a103a26f781cac780f/lib/Onyx.ts#L393-L414)

src/libs/OptimisticReportNamesCache.ts
```diff
@@ -0,0 +1,108 @@
...
+            return {
+                ...workingUpdates,
+                [key]: update.value as WorkingUpdateValue,
+            };
+        case Onyx.METHOD.MERGE: {
+            // Get the current value (from cache or original context)
+            const currentValue = workingUpdates[key] ?? getOriginalValueByKey(key, context);
+            return {
+                ...workingUpdates,
+                [key]: {...currentValue, ...(update.value as WorkingUpdateValue)} as WorkingUpdateValue,
```
neil-marcellini: It looks like this only does a shallow merge, and so it might not work properly for deeply nested objects. Let's make this do a deep merge, maybe using the same fastMerge algorithm Onyx uses. I'm fine with whatever as long as it doesn't substantially reduce performance.

Side note:
Kind of a crazy idea, but what if we re-design this caching system so that it uses Onyx directly? For example maybe we could make a copy of all the onyxUpdates, prepend the keys with `optimisticReportNamesCache-` and then call Onyx.update with all the modified updates to get our end state cache. We could then use the end state cache when looping through the effected reports. Or, the same technique could be used with the incremental approach of applying updates one at a time. We would want to ensure all keys with `optimisticReportNamesCache-` were cleared after each batch of updates is applied, and IDK about the performance implications of that. May be a bit slower, so I recommend the simple approach above.

src/libs/OptimisticReportNamesCache.ts
```diff
@@ -0,0 +1,108 @@
...
+            };
+        }
+        case Onyx.METHOD.CLEAR: {
+            const {[key]: removedItem, ...rest} = workingUpdates;
+            return rest as WorkingUpdates;
+        }
+        default: {
```
neil-marcellini, line 205: We should probably handle these methods too

```
        readonly MERGE_COLLECTION: "mergecollection";
        readonly SET_COLLECTION: "setcollection";
        readonly MULTI_SET: "multiset";
```

src/libs/OptimisticReportNames.ts
```diff
@@ -320,21 +324,18 @@ function updateOptimisticReportNamesFromUpdates(updates: OnyxUpdate[], context:
                 if (transactionUpdate.reportID) {
                     report = getReportByID(transactionUpdate.reportID, allReports);
                 } else {
-                    report = getReportByTransactionID(getTransactionIDFromKey(update.key), context);
+                    report = getReportByTransactionID(getTransactionIDFromKey(update.key), context, workingUpdates);
```
neil-marcellini: Shouldn't we always try to get the report from the cache, either using the reportID from the update or getting the report via the transactionID extracted from the key?

Why do we used the report without considering the cache if the update has a reportID?

neil-marcellini: Maybe it will be more clear to me once I see how we use the report and update as it relates to the affected reports.

src/libs/OptimisticReportNames.ts
```diff
@@ -178,23 +184,21 @@ function isValidReportType(reportType?: string): boolean {
 /**
  * Compute a new report name if needed based on an optimistic update
  */
-function computeReportNameIfNeeded(report: Report | undefined, incomingUpdate: OnyxUpdate, context: UpdateContext): string | null {
-    const {allPolicies} = context;
-
-    // If no report is provided, extract it from the update (for new reports)
-    const targetReport = report ?? (incomingUpdate.value as Report);
+function computeReportNameIfNeeded(report: Report | undefined, context: UpdateContext, incomingUpdate: OnyxUpdate, workingUpdates: WorkingUpdates): string | null {
```
neil-marcellini: Instead of passing the entire report object, we could simply pass the report ID because that's the only field we use. It might be more clear as well

src/libs/OptimisticReportNames.ts
neil-marcellini: After looking at this more, I don't see a reason we need to handle fetching the entire report. We should just be able to use the report ID. And since that's already the key of the affected reports, it will simplify that data storage. See my [other comment here](https://github.com/Expensify/App/pull/70870/files#r2376956214).

src/libs/OptimisticReportNames.ts
```diff
@@ -212,10 +216,10 @@ function computeReportNameIfNeeded(report: Report | undefined, incomingUpdate: O
             // Checking if the formula part is affected in this manner works, but it could certainly be more precise.
             // For example, a policy update only affects the part if the formula in the policy changed, or if the report part references a field on the policy.
             // However, if we run into performance problems, this would be a good place to optimize.
-            return updateType === 'report' || updateType === 'transaction' || updateType === 'policy';
+            return updateType === 'report' || updateType === 'transaction' || updateType === 'policy' || updateType === 'reportNameValuePairs';
```
neil-marcellini: NAB: Let's please add some constants at the top of this file for each update type.

src/libs/Formula.ts
neil-marcellini: Right, so why can't we always get the transaction from the cache, instead of starting with the FormulaContext value, then falling back to the cache, then the transaction currently stored in Onyx?

Isn't the formula context transaction always the same as the one in the cache?

neil-marcellini (changes requested):
I reviewed all the non-test code and I have a few questions and requests

neil-marcellini:
Also, thanks for clearly explaining why the cache is needed, it's very helpful.

src/libs/Formula.ts
sosek108: Yeah, context's transaction is from cache ☝️.

src/libs/OptimisticReportNames.ts
sosek108: Good point 💪

src/libs/OptimisticReportNames.ts
sosek108: Good point

src/libs/OptimisticReportNames.ts
sosek108: Done ☑️

src/libs/OptimisticReportNamesCache.ts
sosek108: Ok, I've removed it

src/libs/OptimisticReportNamesCache.ts
sosek108: I've created method `deepMerge` which is the copy of Onyx fast merge

sosek108: The idea with using Onyx is quite good and interesting. We'd resolve all our cases without doubling the logic for merging etc, but it also could create some race conditions (not sure if that really is the case) and maybe we should check this with someone else.

src/libs/OptimisticReportNamesCache.ts
sosek108: Ok, implemented 🧑‍💻

sosek108 (commented):
@neil-marcellini I've applied changes according to your comments. Most important updates are the deep merge and additional Onyx.METHOD cases in `applyUpdateToCache`

src/libs/OptimisticReportNamesCache.ts
neil-marcellini: Yeah it would be good to check with others if we go that route. However, that may also slow us down for no good reason, so let's stick with doing a deep merge for now.

If we run into performance problems it would definitely be interesting to compare the two approaches.

neil-marcellini: Idk, the more I look at this, I feel like we should just re-use Onyx, because otherwise we're having to build a mini-version of what Onyx does into the cache logic, which feels like unnecessary duplication. I will start a discussion now.

neil-marcellini: [Slack discussion started here](https://expensify.slack.com/archives/C01GTK53T8Q/p1758914545522269)

Issue #69394, opened 2025-08-27
[$250] Don't change the order of user-entered search syntax


JmillsExpensify:
Let's not change the order of a user-entered query, as it's confusing and unexpected. Here's a recording, notice how the `group-by` syntax jumps around? Let's make sure we're not doing that. 

![Image](https://github.com/user-attachments/assets/02c1dba8-1ebc-4688-9672-de361be3628f)

cc @s77rt @ikevin127 
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021960818749737255277</li>
        <li>Upwork Job ID: 1960818749737255277</li>
        <li>Last Price Increase: 2025-09-17</li>
<li>Automatic offers: </li>
<ul>
<li>ShridharGoel | Contributor | 108861845</li>
</ul></li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @</details>

sofi-a:
## Proposal

### Please re-state the problem that we are trying to solve in this issue.
Avoid re-ordering user-entered search syntax

### What is the root cause of that problem?
The [`buildSearchQueryString`](https://github.com/Expensify/App/blob/227c0c26af350194116575cec97c83a306a2a3d9/src/libs/SearchQueryUtils.ts#L379) function first adds filters from `CONST.SEARCH.SYNTAX_ROOT_KEYS` in `queryParts` before evaluating the rest of the filters.

https://github.com/Expensify/App/blob/227c0c26af350194116575cec97c83a306a2a3d9/src/libs/SearchQueryUtils.ts#L383-L394

### What changes do you think we should make in order to solve the problem?
Before returning the joined `queryParts`, we can use the `queryJSON.inputQuery` to identify the true order of the search input and sort `queryParts` before it is joined.

### What specific scenarios should we cover in automated tests to prevent reintroducing this issue in the future?
N/A

### What alternative solutions did you explore? (Optional)
N/A

JmillsExpensify:
@shubham1206agra thoughts on these proposals?

sofi-a:
Updated [proposal](https://github.com/Expensify/App/issues/69394#issuecomment-3231098591).

sofi-a:
I have a working solution @shubham1206agra.

https://github.com/user-attachments/assets/bec0d0ad-924d-4279-a670-a57925e1142f

sofi-a:
[...]

PR #71075, opened 2025-09-23
Maintain user defined order in search query

ShridharGoel:
### Explanation of Change

Maintain user defined order in search query. Now, inputs will not jump around and will remain at the same place where the user would have typed them.

### Fixed Issues

$ https://github.com/Expensify/App/issues/69394
PROPOSAL: https://github.com/Expensify/App/issues/69394#issuecomment-3234898459

### Tests

1. Go to Reports page.
2. Manual enter a query like type:expense-report date:this-month status:draft.
3. Press enter.
4. The entered term should remain at the place where it was written and should not get reordered.
5. Use form filter to change status to Approved and apply filter
6. The order of entered term could be changed.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/468a807d-32cf-4314-b948-06f2c04fe4b2

</details>

<details>

https://github.com/user-attachments/assets/55cd8e77-8483-485d-9ac2-ee4f23f7c47c

</details>

<details>

https://github.com/user-attachments/assets/a09a1d5d-7b32-4d56-8a17-fc4297da9805

</details>

<details>

https://github.com/user-attachments/assets/9c5cffb0-82a5-4c8e-8491-ec44e22bc638

</details>

<details>

https://github.com/user-attachments/assets/5fa57d2e-42ea-4fdf-a517-a8e730bd33eb

</details>

<details>

https://github.com/user-attachments/assets/bca58043-2663-406b-a6af-3187c4270d83

</details>


JS00001:
Is this just waiting on review? If so, @shubham1206agra can you please re-review

ShridharGoel:
@shubham1206agra Any update on this?

src/libs/SearchQueryUtils.ts
```diff
@@ -390,6 +391,256 @@ function normalizeValue<T>(value: T | T[]): T {
...
+ * @param cardFeeds Card feed metadata used for feed resolution.
+ * @param policies Policies collection for workspace/name lookups.
+ * @param currentUserAccountID Current user account ID for self lookups.
+ * @param taxRates Mapping of tax rate names to IDs for tax resolution.
+ * @return Filter segment or `undefined` when no displayable values exist.
+ */
+function getFilterTitleSegment(
+    key: string,
+    filters: QueryFilter[],
+    PersonalDetails: OnyxTypes.PersonalDetailsList | undefined,
```
shubham1206agra:
```suggestion
    personalDetails: OnyxTypes.PersonalDetailsList | undefined,
```

src/libs/SearchQueryUtils.ts
```diff
@@ -390,6 +391,256 @@ function normalizeValue<T>(value: T | T[]): T {
...
 }
 
+/**
+ * Converts a top-level field back into `key:value` format while omitting empty entries.
+ *
+ * @param key Search syntax identifier for the root field being serialized.
+ * @param value Raw value pulled from the query JSON or defaults.
+ * @return Serialized `key:value` segment or `undefined` when the value should be skipped.
+ */
+function serializeRootEntry(key: string, value: unknown) {
```
shubham1206agra: @ShridharGoel Why is the value unknown here?

src/libs/SearchQueryUtils.ts
ShridharGoel: Updated.

src/libs/SearchQueryUtils.ts
ShridharGoel: @shubham1206agra We get the value with a dynamic string key, so the compiler only knows it’s "some value". Using unknown makes it sure that the code inside the method checks for array/string before we serialize.

If we switched to any, we could start calling properties or methods that don’t exist on that object and the compiler wouldn’t stop us.

eh2077:
Yeah, on it

ShridharGoel:
Done

src/components/Search/types.ts
```diff
@@ -206,6 +217,8 @@ type SearchQueryJSON = {
     /** Use similarSearchHash to test if two searchers are similar i.e. have same filters but not necessary same values */
     similarSearchHash: number;
     flatFilters: QueryFilters;
+    /** Position information for preserving original query order */
+    positionInfo?: SearchQueryPositionEntry[];
 } & SearchQueryAST;
```
eh2077: Can we get original query order by inorder traversal ASTNode `SearchQueryAST.filters`? If so, then we don't need to modify the peggy parser and I believe the implementation will be much simpler than the current method.

JS00001:
No worries, I agree we follow the best approach, could you elaborate on your potential solution a bit more so we can weigh both?

ShridharGoel:
Seems like an in-order walk of SearchQueryAST.filters will always surface keywords last, so it wouldn't work. But would be good to get more info on that approach to see its feasibility.

eh2077:
Yes, you're right. The parsing expression of `filterList` reorders and compact `keyword` filters (composing formal `keyword` filters or free text filters)

- L79 - L85 wrap all keyword filters into a nested keyword filter
- L89 - L91 is tree like (similar to a linked list)

```txt
           AND
          /   \
         AND  (keyword filter)
        /  \ 
      AND   Fn-1
     /
   ....
   /
  AND
 /  \
F1   F2
```

https://github.com/Expensify/App/blob/8edb23dcd37988bffe0064cad5a1e35e7bfd9019/src/libs/SearchParser/searchParser.peggy#L61-L91

Note that default filters are also reordered and aren't included in the filters (ASTNode)

https://github.com/Expensify/App/blob/8edb23dcd37988bffe0064cad5a1e35e7bfd9019/src/libs/SearchParser/searchParser.peggy#L43-L48

---

So, the ast `SearchQueryAST` returned by the parser isn't a perfect node that remains original order of user input. Two reorderings happen during parsing
- reorder and compact `keyword` filters
- spread default filters

https://github.com/Expensify/App/blob/c7b3a441e47577793f7cbfa95cdf45829c7f93f3/src/components/Search/types.ts#L190-L198


I think it's feasible to let peggy parser build a perfect ASTNode which can restore original user input. We can do post processing to handle `keyword` filters and spread default filters. I see following advantages of this solution
- keep the parser simpler and cleaner
- don't need extra data structure to save raw character position, like the current method of this PR
- it's convenient to restore the order of user query

@JS00001 @ShridharGoel Wdyt?

ShridharGoel:
@eh2077 Thanks for the detailed write-up. Can you have a look at [this](https://github.com/ShridharGoel/ExpensifyApp/tree/searchNew) branch? It isn't working as expected, but can you check if this is what you meant?

eh2077:
@JS00001 Can you weigh in https://github.com/Expensify/App/pull/71075#issuecomment-3427332519 please, thanks!

src/libs/SearchParser/searchParser.peggy
eh2077, line 1: According to https://github.com/Expensify/App/pull/71075#issuecomment-3427332519, I think we should avoid twisting tree, that means
- don't spread default filters and move `applyDefaults` out of `src/libs/SearchParser/searchParser.peggy`,  which should be part of post processing
- `filterList` expression don't compact `keyword` filters and that should be handled in post processing too

I think the `token` prop is technically same as the saving position index solution and we should avoid it. Let's try let `filterList` just return a list of filters as it is and let post processing (visit ast node) to optimize `keyword` filters and spread default filters.

ShridharGoel:
@eh2077 Does the updated code follow the correct direction?

ShridharGoel:
The `searchParser.peggy` grammar now just emits the raw sequence of parsed entries in input order, and the post-processing step reconstructs defaults, compacts keyword filters, and assembles the tree.

src/libs/SearchParser/searchParser.peggy
```diff
@@ -20,82 +20,39 @@
...
+      return {
+        type: "default",
+        key,
+        operator: op,
+        value,
+        location: {
+          start: start.offset,
+          end: end.offset,
+        },
+      };
```
eh2077: I think you should use the `buildFilter` function here. We should avoid saving location info, instead, we want to restore user query from ast node.

src/libs/SearchParser/searchParser.peggy
```diff
@@ -107,7 +64,18 @@ freeTextFilter
...
+      return {
+        type: "filter",
+        key: "keyword",
+        operator: "eq",
+        value: word,
+        isImplicitKeyword: true,
+        location: {
+          start: start.offset,
+          end: end.offset,
+        },
```
eh2077: I think you can try adding `isImplicitKeyword` to  `buildFilter`, like

```js
  function buildFilter(operator, left, right, props = {}) {
    return { operator, left, right, ...props };
  }
```

eh2077 (commented):
Thanks for quick update! Let’s try to see if we can avoid saving character location information entirely.

ShridharGoel:
@eh2077 Thanks, it is updated.

src/libs/SearchParser/searchParser.peggy
```diff
@@ -21,81 +21,32 @@
...
+      return entries;
     }
 
-filter = @(standardFilter / defaultFilter / freeTextFilter)
+filter = standardFilter / defaultFilter / freeTextFilter
 
 defaultFilter
   = _ key:defaultKey _ op:operator _ value:identifier {
-      updateDefaultValues(key, value);
+      return buildFilter(op, key, value, {type: "default", isDefault: true});
```
eh2077: I think adding a `type` prop is redundant

src/libs/SearchParser/searchParser.peggy
```diff
@@ -124,7 +75,11 @@ standardFilter
         }
       }
 
-      return buildFilter(operator, key, values);
+      return buildFilter(operator, key, values, {
+        type: "filter",
+        isImplicitKeyword: false,
```
eh2077: I think we can only set the prop if it's true for simplicity

src/libs/SearchParser/postProcessor.ts
```diff
@@ -0,0 +1,232 @@
...
+            const values = toStringArray(value);
+            defaults.policyID = values.length > 0 ? values : defaults.policyID;
+            break;
+        }
+        default:
+            break;
+    }
+}
+
+function formatTokenRaw(entry: RawFilterEntry): string {
```
eh2077: Why do we need to format raw string here? I think this should be handled in method `buildUserReadableQueryString`.

src/libs/SearchParser/postProcessor.ts
```diff
@@ -0,0 +1,232 @@
...
+        (result, node) => ({
+            operator: CONST.SEARCH.SYNTAX_OPERATORS.AND,
+            left: result,
+            right: node,
+        }),
+        nodes[0],
+    );
+}
+
+function updateDefaultValues(defaults: DefaultValues, entry: RawFilterEntry) {
```
eh2077: Can you make this method simpler? The one from parser is like

```js
  function updateDefaultValues(field, value) {
    if (field === "status" && value === "all") {
      defaultValues[field] = "";
      return;
    }
    defaultValues[field] = value;
  }
```

src/libs/SearchParser/postProcessor.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    return nodes.slice(1).reduce<ASTNode>(
+        (result, node) => ({
+            operator: CONST.SEARCH.SYNTAX_OPERATORS.AND,
+            left: result,
+            right: node,
+        }),
+        nodes[0],
+    );
```
eh2077:
```suggestion
    return nodes.reduce<ASTNode>(
        (result, node) => ({
            operator: CONST.SEARCH.SYNTAX_OPERATORS.AND,
            left: result,
            right: node,
        })
    );
```

src/libs/SearchParser/postProcessor.ts
```diff
@@ -0,0 +1,232 @@
...
+            operator: CONST.SEARCH.SYNTAX_OPERATORS.EQUAL_TO,
+            left: CONST.SEARCH.SYNTAX_FILTER_KEYS.KEYWORD,
+            right: keywordValues,
+        });
+    }
+
+    return {
+        ...defaults,
+        filters: buildFilterTree(filters),
+        tokens,
```
eh2077: Hmm, I think it's better to name it as `rawFilterList` or `filterList`

src/libs/SearchParser/searchParser.peggy
eh2077: What about having both, capture default filters and include them in `filterList` as well? Seems it'll make post processing simpler.

```suggestion
      updateDefaultValues(key, value);
      return buildFilter(op, key, value, {type: "default", isDefault: true});
```

src/libs/SearchQueryUtils.ts
```diff
@@ -360,6 +370,309 @@ function getQueryHashes(query: SearchQueryJSON): {primaryHash: number; recentSea
...
 }
 
+function toStringArray(value: unknown): string[] {
+    if (value == null) return [];
+
+    const convert = (v: unknown): string => SafeString(v, true);
+    return Array.isArray(value) ? value.map(convert) : [convert(value)];
+}
+
+function syncTokensWithQuery(queryJSON: SearchQueryJSON) {
```
eh2077: I think this method is too complicated. I suppose we only add a new prop `rawFilterList` to the `SearchQueryJSON` object and method `buildUserReadableQueryString` will construct user query from `rawFilterList`. 

As there're two ways to update query string, search query input and others (like form), I think it's better not to have this sync method. Instead, only overriding `rawFilterList` with values of default filters, filters, flatFilters of SearchQueryJSON  in method `buildUserReadableQueryString`. 

Hmm, this part seems tricky, so feel free to discuss more if you have any concerns, thanks!

src/libs/SearchParser/searchParser.peggy
ShridharGoel: But having `updateDefaultValues` in the Peggy initializer would mean that some tree twisting will happen in the parser itself?

src/libs/SearchQueryUtils.ts
ShridharGoel: Dropping `syncTokensWithQuery` and using a memoized `rawFilterList` can likely give us stale or duplicated output.

- Other parts of the app tweak the parsed query directly (for example, the filter form). If we only saved the text from the first parse, that text would get out of date the moment the form changed anything, so we can end up rebuilding the wrong query.
- Trying to fix that later, right before we rebuild the string, means we’d have to repeat all the same "keep tokens in sync" steps somewhere else.

What do you think?

ShridharGoel:
@eh2077 Thanks for the review. Updated most of the things, and some of the things are up for discussion.

src/libs/SearchQueryUtils.ts
eh2077: @ShridharGoel Let's step back a bit, the original issue we want to fix is to retain original order of filters when the following steps are performed

1. Open Reports and focus in search input
2. Enter a filter like `type:expense date:this-month group-by:from`
3. Press `Enter` to search
4. The order of filters in search input shouldn't be changed.

Let's say we don't consider use case of update filter by form, what are the minimal changes needed to fix this issue?

ShridharGoel: Will these be enough for the minimal change?

1. Have the Peggy rules given back each token’s original text (text() → raw). That way we get exactly what the user typed, in order.
2. When we rebuild the string after searching, just join those raw snippets in token order, like `tokens.map((t) => t.raw.trim()).filter(Boolean).join(' ')`.

eh2077: I don't think the raw text thing is the point of the issue, shouldn't only the order of filters that maters in the original issue? Given that filter could be formatted by method `buildUserReadableQueryString`.

ShridharGoel: Yeah, actually we can put it in this way then:

1. Let the parser keep emitting normal filter objects.
2. Capture just the list of parsed filters in order, like filterList → rawFilterList.
3. Update `buildUserReadableQueryString` to iterate that list and render each filter with whatever formatting rules we already have. And then join in the same sequence without the need for reordering.

Handling the raw text can be optional enhancement.

eh2077: Hmm, I think. to fix https://github.com/Expensify/App/pull/71075#discussion_r2463212975, I see following as minimal changes
1. method `buildSearchQueryJSON` returns SearchQueryJSON with non-null `rawFilterList` if and only if the query is triggered by manual input and `Enter`. So, `rawFilterList` will be available from SearchQueryJSON when needed though we capture it from `filterList` every time.
2. in method `buildUserReadableQueryString`, we try to build query string from `rawFilterList` if it's non null. That said, all exiting behaviors, like form filters, remain not changed. We should reuse existing format methods as much as possible to follow DRY principle.

eh2077: @ShridharGoel Cool! Just saw your last comment after reloading. 

> Yeah, actually we can put it in this way then:
> 
> 1. Let the parser keep emitting normal filter objects.
> 2. Capture just the list of parsed filters in order, like filterList → rawFilterList.

Yeah, I agree. We can keep exiting behavior of `src/libs/SearchParser/searchParser.peggy` unchanged but capture `rawFilterList` from `filterList`. The default filters can be handled like
```js
defaultFilter
  = _ key:defaultKey _ op:operator _ value:identifier {
      updateDefaultValues(key, value); // exiting way to capture default filters to spread props
      return buildFilter(op, key, value, {isDefault: true}); // return default filters, so filterList can receive default filters in order to form rawFilterList. In the filterList expression, we can also filter out them by prop isDefault
    }
```

> 3. Update `buildUserReadableQueryString` to iterate that list and render each filter with whatever formatting rules we already have. And then join in the same sequence without the need for reordering.
> 
> Handling the raw text can be optional enhancement.

Agree!

ShridharGoel: @eh2077 Thanks, can you have a look at the updated code?

src/components/Search/SearchPageHeader/SearchPageHeaderInput.tsx
```diff
@@ -196,14 +198,25 @@ function SearchPageHeaderInput({queryJSON, searchRouterListVisible, hideSearchRo
...
             const queryWithSubstitutions = getQueryWithSubstitutions(queryString, autocompleteSubstitutions);
             const updatedQuery = getQueryWithUpdatedValues(queryWithSubstitutions);
 
             if (!updatedQuery) {
                 return;
             }
 
+            if (options?.captureRawFilters) {
+                try {
+                    const parsed = parseSearchQuery(updatedQuery);
```
eh2077: I don't think it's a good idea call this method here. It'll be better to handle the details in method `buildSearchQueryJSON`. What about adding a parameter to method `getQueryWithUpdatedValues`

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -273,13 +275,24 @@ function SearchRouter({onRouterClose, shouldHideInputCaret, isSearchRouterDispla
...
             }
 
+            if (options?.captureRawFilters) {
+                try {
+                    const parsed = parseSearchQuery(updatedQuery);
+                    if (parsed.rawFilterList && parsed.rawFilterList.length > 0) {
+                        saveManualRawFilterList(updatedQuery, parsed.rawFilterList);
+                    }
+                } catch (error) {
+                    Log.warn('[SearchRouter] Failed to cache raw filter list', false, {error});
```
eh2077: same

src/libs/SearchParser/manualQueryCache.ts
eh2077, line 1: why do you need to add this cache? Anyway, I think it's a bad practice to have such a cache.

src/libs/SearchQueryUtils.ts
eh2077: I think this method `syncTokensWithQuery ` should be removed according to the above discussions

src/components/Search/types.ts
```diff
@@ -194,19 +216,23 @@ type SearchQueryAST = {
     sortBy: SearchColumnType;
     sortOrder: SortOrder;
     groupBy?: SearchGroupBy;
-    filters: ASTNode;
+    filters: ASTNode | null;
     policyID?: string[];
+    tokens?: SearchQueryToken[];
+    rawFilterList?: RawFilterEntry[];
+    rawFilterListSource?: 'manual';
```
eh2077: I think it only needs to add an optional prop `rawFilterList` to SearchQueryJSON instead of 3 props

eh2077 (commented):
@ShridharGoel To be clear, I want to keep the changes as minimal as possible and don't change existing behaviors other than fixing the original issue.

src/libs/SearchQueryUtils.ts
```diff
@@ -400,11 +690,17 @@ function isFilterSupported(filter: SearchAdvancedFiltersKey, type: SearchDataTyp
  */
 function buildSearchQueryJSON(query: SearchQueryString) {
     try {
-        const result = parseSearchQuery(query) as SearchQueryJSON;
+        const parsedQuery = parseSearchQuery(query);
```
eh2077: I think we shouldn't change formatting of this method because the query in search input is formatted by method `buildUserReadableQueryString`.

src/libs/SearchQueryUtils.ts
```diff
@@ -961,86 +1361,155 @@ function buildUserReadableQueryString(
...
     currentUserAccountID: number,
 ) {
-    const {type, status, groupBy, policyID} = queryJSON;
+    const {type, status, groupBy, policyID, rawFilterList} = queryJSON;
     const filters = queryJSON.flatFilters;
 
-    let title = status
-        ? `type:${getUserFriendlyValue(type)} status:${Array.isArray(status) ? status.map(getUserFriendlyValue).join(',') : getUserFriendlyValue(status)}`
-        : `type:${getUserFriendlyValue(type)}`;
+    const filterDisplayEntries = filters
```
eh2077: As discussed in this thread https://github.com/Expensify/App/pull/71075#discussion_r2463212975, we should only format with `rawFilterList` when it's not empty, otherwise the existing logic should remain unchanged.

eh2077 (commented):
@ShridharGoel Please take these comments into consideration https://github.com/Expensify/App/pull/71075#discussion_r2463212975, https://github.com/Expensify/App/pull/71075#discussion_r2463775507, https://github.com/Expensify/App/pull/71075#discussion_r2463788776, https://github.com/Expensify/App/pull/71075#discussion_r2463802575. Feel free to discuss more to make it clear if you have any questions, thanks!

ShridharGoel:
@eh2077 Updated.

src/libs/SearchQueryUtils.ts
```diff
@@ -58,6 +63,8 @@ const operatorToCharMap = {
     [CONST.SEARCH.SYNTAX_OPERATORS.OR]: ' ' as const,
 };
 
+const manualQueryRawFilterStore = new Map<string, RawQueryFilter[]>();
```
eh2077: Can you elaborate why this cache is need? As a best practice, I think we shouldn't keep it because it's prone to memory issues.

eh2077 (commented):
Overall, I think the current implementation is much more succinct than previous versions 👍

src/libs/SearchQueryUtils.ts
ShridharGoel: It remembers the exact filters the user typed. After we sanitize the query and turn it into our standard format, we look up the saved `manualQueryRawFilterStore` so we can show them back to the user in the same order they entered.

eh2077: @ShridharGoel I think it'll be better to fix the issue without using this cache. Is it feasible?

JS00001: How large would the cache be though? If I'm not missing anything, it wouldnt grow to be very big, right?

ShridharGoel: `manualQueryRawFilterStore` is an in-memory map. Each time we normalize a query, we map the string with the raw filters, and the very next call to `buildSearchQueryJSON` with that string reads them back and immediately deletes the entry. Ideally, it should be of a very small size.

JS00001: I would think thats fine

eh2077: I like to share my understanding about why this cache is needed. 
1. Manual type a filter and press enter - method `getQueryWithUpdatedValues` formats a standard query string as navigation params
2. reload search page - method `buildSearchQueryJSON` creates json object and needs to figure out if it's a query triggered by manual typing query (if yes, then save rawFilterList to the json object)

As the navigation parameter `query` (standardized query string) doesn't contain information about whether it's a manual typing query, using an intermediate `manualQueryRawFilterStore` (standardized query as key, rawFilterList as value) is a way to solve the problem. This way has an advantage to hide details in `src/libs/SearchQueryUtils.ts`. 

An alternative method will be adding a new navigation parameter to indicate a manual typing query. But we'll need to change navigation routes and components.

eh2077 (commented):
Sorry, review is ongoing and it's a bit slower than expected.

src/libs/SearchQueryUtils.ts
eh2077: @ShridharGoel I’d prefer not to use this cache method. What do you think about the alternative approach?

ShridharGoel: Ok, I'll check the navigation parameter idea. But can you share why do we want to avoid this cache, when it is anyways going to be quite small and is probably a cleaner way of achieving the needed functionality?

eh2077: First of all, I don’t think it’s common practice in this codebase to solve this kind of issue as described above. Instead, I think using navigation parameters is a common pattern in the app, and I believe that’s the right approach to take. And, ideally, utility functions shouldn't have side effects.

eh2077:
@ShridharGoel Wdyt on https://github.com/Expensify/App/pull/71075#discussion_r2508649659?

src/libs/SearchQueryUtils.ts
```diff
@@ -47,6 +48,11 @@ import {isValidDate} from './ValidationUtils';
 
 type FilterKeys = keyof typeof CONST.SEARCH.SYNTAX_FILTER_KEYS;
 
+type BuildSearchQueryJSONOptions = {
+    preserveRawFilterList?: boolean;
+    manualRawFilterList?: RawQueryFilter[];
```
eh2077: Can we only use a boolean prop `preserveRawFilterList` as the navigation parameter? So, we can make method `getQueryWithUpdatedValues` remain unchanged and method `buildSearchQueryJSON` can preserve raw filter if  `preserveRawFilterList` is true.

ShridharGoel: If we only pass a boolean like `preserveRawFilterList`, then how will we have the raw filter list that has to be preserved?

joekaufmanexpensify (dismissed):
Good from a product perspective 👍

src/libs/SearchQueryUtils.ts
eh2077: @ShridharGoel Oh, I got your point. I think we can still try this way to make it simpler
1. make method `getQueryWithUpdatedValues` remain unchanged
2. add a new navigation parameter, `rawQuery`, to save the raw query, so we can parse raw filter list from `rawQuery` in method `buildSearchQueryJSON`.

ShridharGoel: @eh2077 Updated.

eh2077: @ShridharGoel Yeah, the `rawQuery` looks good. Can you proceed with cleaning up `preserveRawFilterList` and `manualRawFilterList` parameters? I think they're not needed anymore.

ShridharGoel:
@eh2077 This is ready for next round of review.

src/libs/SearchQueryUtils.ts
```diff
@@ -1144,17 +1333,22 @@ function traverseAndUpdatedQuery(queryJSON: SearchQueryJSON, computeNodeValue: (
...
-
     const standardizedQuery = traverseAndUpdatedQuery(queryJSON, computeNodeValue);
-    return buildSearchQueryString(standardizedQuery);
+    const sanitizedRawFilters = getSanitizedRawFilters(standardizedQuery);
+    const canonicalQueryString = buildSearchQueryString(standardizedQuery);
+
+    return {
+        canonicalQuery: canonicalQueryString,
+        rawFilterList: sanitizedRawFilters,
+    };
```
eh2077: I think it's not necessary to change this method.

src/libs/SearchQueryUtils.ts
```diff
@@ -415,6 +428,13 @@ function buildSearchQueryJSON(query: SearchQueryString) {
         result.recentSearchHash = recentSearchHash;
         result.similarSearchHash = similarSearchHash;
 
+        const rawFilterListFromRawQuery = rawQuery ? getRawFilterListFromQuery(rawQuery) : undefined;
+        if (rawFilterListFromRawQuery && rawFilterListFromRawQuery.length > 0) {
+            result.rawFilterList = rawFilterListFromRawQuery;
+        } else {
+            delete result.rawFilterList;
+        }
```
eh2077: I think we can simplify as follows, wdyt?
```suggestion
        delete result.rawFilterList;
        if (rawQuery) {
            result.rawFilterList = getRawFilterListFromQuery(rawQuery);
        }
```

src/ROUTES.ts
```diff
@@ -56,8 +56,10 @@ const ROUTES = {
 
     SEARCH_ROOT: {
         route: 'search',
-        getRoute: ({query, name}: {query: SearchQueryString; name?: string}) => {
-            return `search?q=${encodeURIComponent(query)}${name ? `&name=${name}` : ''}` as const;
+        getRoute: ({query, rawQuery, name}: {query: SearchQueryString; rawQuery?: SearchQueryString; name?: string}) => {
+            const resolvedRawQuery = rawQuery ?? query;
+            const rawQuerySegment = resolvedRawQuery ? `&rawQuery=${encodeURIComponent(resolvedRawQuery)}` : '';
+            return `search?q=${encodeURIComponent(query)}${name ? `&name=${name}` : ''}${rawQuerySegment}` as const;
```
eh2077: Why need nullish coalescing for raw query? I think if `rawQuery` is null or undefined then skip parsing raw filter list.

src/components/Navigation/NavigationTabBar/index.tsx
```diff
@@ -171,13 +171,14 @@ function NavigationTabBar({selectedTab, isTopLevelBar = false, shouldShowFloatin
...
-                const {q, ...rest} = lastSearchRoute.params as SearchFullscreenNavigatorParamList[typeof SCREENS.SEARCH.ROOT];
-                const queryJSON = buildSearchQueryJSON(q);
+                const {q, rawQuery, ...rest} = lastSearchRoute.params as SearchFullscreenNavigatorParamList[typeof SCREENS.SEARCH.ROOT];
+                const queryJSON = buildSearchQueryJSON(q, rawQuery);
                 if (queryJSON) {
                     const query = buildSearchQueryString(queryJSON);
                     Navigation.navigate(
                         ROUTES.SEARCH_ROOT.getRoute({
                             query,
+                            rawQuery: rawQuery ?? query,
```
eh2077: Same

src/components/Search/SearchPageHeader/SearchFiltersBar.tsx
```diff
@@ -256,7 +257,7 @@ function SearchFiltersBar({
             const queryString = buildQueryStringFromFilterFormValues(updatedFilterFormValues);
 
             close(() => {
-                Navigation.setParams({q: queryString});
+                Navigation.setParams({q: queryString, rawQuery: queryString});
```
eh2077: I think `rawQuery` parameter is not empty only when method `getQueryWithUpdatedValues` is called. Wdyt? See this previous https://github.com/Expensify/App/pull/71075#discussion_r2463212975 too

src/libs/Navigation/AppNavigator/Navigators/SearchFullscreenNavigator.tsx
```diff
@@ -33,7 +35,7 @@ function SearchFullscreenNavigator({route}: PlatformStackScreenProps<AuthScreens
                 <Stack.Screen
                     name={SCREENS.SEARCH.ROOT}
                     getComponent={loadSearchPage}
-                    initialParams={{q: SearchQueryUtils.buildSearchQueryString()}}
+                    initialParams={{q: defaultSearchQuery, rawQuery: defaultSearchQuery}}
```
eh2077: Same, rawQuery should be default undefined, right?

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
```diff
@@ -43,16 +43,20 @@ function preloadReportsTab(navigation: PlatformStackNavigationProp<AuthScreensPa
...
-        const {q, ...rest} = lastSearchRoute.params as SearchFullscreenNavigatorParamList[typeof SCREENS.SEARCH.ROOT];
-        const queryJSON = buildSearchQueryJSON(q);
+        const {q, rawQuery, ...rest} = lastSearchRoute.params as SearchFullscreenNavigatorParamList[typeof SCREENS.SEARCH.ROOT];
+        const queryJSON = buildSearchQueryJSON(q, rawQuery);
         if (queryJSON) {
             const query = buildSearchQueryString(queryJSON);
-            navigation.preload(NAVIGATORS.SEARCH_FULLSCREEN_NAVIGATOR, {screen: SCREENS.SEARCH.ROOT, params: {q: query, ...rest}});
+            navigation.preload(NAVIGATORS.SEARCH_FULLSCREEN_NAVIGATOR, {
+                screen: SCREENS.SEARCH.ROOT,
+                params: {q: query, rawQuery: rawQuery ?? query, ...rest},
```
eh2077: Same, why nullish coalescing

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
```diff
@@ -43,16 +43,20 @@ function preloadReportsTab(navigation: PlatformStackNavigationProp<AuthScreensPa
...
+            });
             return;
         }
     }
 
-    navigation.preload(NAVIGATORS.SEARCH_FULLSCREEN_NAVIGATOR, {screen: SCREENS.SEARCH.ROOT, params: {q: buildCannedSearchQuery()}});
+    const defaultQuery = buildCannedSearchQuery();
+    navigation.preload(NAVIGATORS.SEARCH_FULLSCREEN_NAVIGATOR, {screen: SCREENS.SEARCH.ROOT, params: {q: defaultQuery, rawQuery: defaultQuery}});
```
eh2077: Same, why rawQuery is not default to undefined

src/components/Search/index.tsx
eh2077, line 1: Same question as in file `src/components/Search/SearchPageHeader/SearchFiltersBar.tsx`

src/components/Search/SearchPageHeader/SearchFiltersBar.tsx
```diff
@@ -256,7 +257,7 @@ function SearchFiltersBar({
             const queryString = buildQueryStringFromFilterFormValues(updatedFilterFormValues);
 
             close(() => {
-                Navigation.setParams({q: queryString});
+                Navigation.setParams({q: queryString, rawQuery: undefined});
```
eh2077, line 316: I think we don't need to set undefined value explicitly, so we can revert change of this file.

src/components/Search/index.tsx
```diff
@@ -799,7 +799,8 @@ function Search({
         if (hasErrors && (currentRoute === '/' || (shouldResetSearchQuery && currentRoute === '/search'))) {
             // Use requestAnimationFrame to safely update navigation params without overriding the current route
             requestAnimationFrame(() => {
-                Navigation.setParams({q: buildCannedSearchQuery()});
+                const cannedQuery = buildCannedSearchQuery();
+                Navigation.setParams({q: cannedQuery, rawQuery: undefined});
```
eh2077: Same, we don't need to parse `undefined` for `rawQuery`, right? If so, then let's don't change this file.

src/libs/Navigation/AppNavigator/Navigators/SearchFullscreenNavigator.tsx
eh2077, line 1: Let's not change this file.

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
eh2077, line 1: I think we don't need to change this file because it's not necessary to parse `rawQuery` to buildSearchQueryJSON to determine if `queryJSON` is nullable or not.

src/components/Navigation/NavigationTabBar/index.tsx
eh2077, line 1: Same as file `src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts`

eh2077 (commented):
@ShridharGoel thanks for your patience and adding unit tests! It works well for me and I have a few comments about making the code neat. Ping me after your update, then I'll complete the checklist.

src/components/Search/SearchPageHeader/SearchFiltersBar.tsx
ShridharGoel: This is done because if we don’t clear it when the advanced filters build a new canonical q, the stale rawQuery would keep overriding parts of the query (and the displayed readable query) even though the user just changed filters. `Navigation.setParams` merges with the existing route params so if we omit `rawQuery`, whatever value was set will stay on the route.

src/components/Search/index.tsx
ShridharGoel: Same as above.

src/components/Search/SearchPageHeader/SearchFiltersBar.tsx
eh2077: @ShridharGoel I see. Can you add a comment for it?

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
ShridharGoel: We can stop passing rawQuery in `buildSearchQueryJSON` but shouldn't we keep the other changes of this file? Because we need to carry rawQuery through when preloading the saved search route, otherwise we’d drop any raw filters the user had and the preloaded search would reopen without those.

eh2077: I think `rawQuery` will be carried in `rest`, so preloading won't lost `rawQuery`. Is that right?

ShridharGoel: Makes sense, thanks for pointing that out

ShridharGoel: @eh2077 Updated.

src/components/Search/SearchPageHeader/SearchFiltersBar.tsx
```diff
@@ -278,7 +278,11 @@ function SearchFiltersBar({
             const queryString = buildQueryStringFromFilterFormValues(updatedFilterFormValues);
 
             close(() => {
-                Navigation.setParams({q: queryString});
+                // We are passing rawQuery as undefined because if we don’t clear it when the advanced filters build a
+                // new canonical q, the stale rawQuery would keep overriding parts of the query (and the displayed
+                // readable query) even though the user just changed filters. Navigation.setParams merges with the
+                // existing route params so if we omit rawQuery, whatever value was set will stay on the route.
```
eh2077:
```suggestion
                // We want to explicitly clear stale rawQuery since it’s only used for manually typed-in queries.
```

src/components/Search/index.tsx
```diff
@@ -834,7 +834,12 @@ function Search({
         if (hasErrors && (currentRoute === '/' || (shouldResetSearchQuery && currentRoute === '/search'))) {
             // Use requestAnimationFrame to safely update navigation params without overriding the current route
             requestAnimationFrame(() => {
-                Navigation.setParams({q: buildCannedSearchQuery()});
+                const cannedQuery = buildCannedSearchQuery();
+                // We are passing rawQuery as undefined because if we don’t clear it when the advanced filters build a
+                // new canonical q, the stale rawQuery would keep overriding parts of the query (and the displayed
+                // readable query) even though the user just changed filters. Navigation.setParams merges with the
+                // existing route params so if we omit rawQuery, whatever value was set will stay on the route.
+                Navigation.setParams({q: cannedQuery, rawQuery: undefined});
```
eh2077:
```suggestion
                // We want to explicitly clear stale rawQuery since it’s only used for manually typed-in queries.
                Navigation.setParams({q: buildCannedSearchQuery(), rawQuery: undefined});
```

src/components/Search/index.tsx
```diff
@@ -951,7 +956,7 @@ function Search({
         clearSelectedTransactions();
         const newQuery = buildSearchQueryString({...queryJSON, sortBy: column, sortOrder: order});
         onSortPressedCallback?.();
-        navigation.setParams({q: newQuery});
+        navigation.setParams({q: newQuery, rawQuery: undefined});
```
eh2077, line 970:
```suggestion
        // We want to explicitly clear stale rawQuery since it’s only used for manually typed-in queries.
        navigation.setParams({q: newQuery, rawQuery: undefined});
```

eh2077 (commented):
Looks good, I'd suggest to make comment shorter.

eh2077:
@ShridharGoel I think we should update test steps to mention that the order of query could be changed when using form filter, see https://github.com/Expensify/App/pull/71075#issuecomment-3580427179. Wdyt?

eh2077 (approved):
LGTM 🚀

JS00001 (dismissed):
Please fix conflicts

src/pages/Search/SearchPage.tsx
```diff
@@ -119,14 +119,15 @@ function SearchPage({route}: SearchPageProps) {
     const [isExportWithTemplateModalVisible, setIsExportWithTemplateModalVisible] = useState(false);
     const [searchRequestResponseStatusCode, setSearchRequestResponseStatusCode] = useState<number | null>(null);
     const [isDEWModalVisible, setIsDEWModalVisible] = useState(false);
+
```
eh2077:
```suggestion
```

eh2077 (approved):

JS00001 (approved):

thesahindia:
This PR caused https://github.com/Expensify/App/issues/87285. [Explanation here](https://github.com/Expensify/App/issues/87285#issuecomment-4423645130).

sobitneupane:
This PR caused https://github.com/Expensify/App/issues/79952. When searching in `#chat`, the user was unintentionally being directed to `type:expense in:#chat` instead of `type:chat in:#chat`.

Issue #68965, opened 2025-08-21
[Due for payment 2025-12-12] [$250] [ORTF] Currency formatting options


arosiclair:
Add support for 

- `nosymbol` – No currency symbol
- `USD`, `EUR`, etc. – Specific currency
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021958951221451751739</li>
        <li>Upwork Job ID: 1958951221451751739</li>
        <li>Last Price Increase: 2025-09-18</li>
<li>Automatic offers: </li>
<ul>
<li>situchan | Contributor | 109150161</li>
</ul></li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @sobitneupane </details>

<details><summary>Issue Owner</summary>Current Issue Owner: @dylanexpensify</details>

arosiclair:
@mananjadhav check out the documentation for Classic [here](https://help.expensify.com/articles/expensify-classic/reports/Set-Custom-Report-Names). It should behave the same way in NewDot. Also take a look at the [offline section of the design doc](https://docs.google.com/document/d/1U2AiMOlWvT144a2WVi2XCSc18piyE_iFrv7cYWhsAyw/edit?tab=t.0#heading=h.olra4emh77tp) if you haven't already.

mananjadhav:
Request access to design doc. But now I can see the issue is marked on Hold?

mananjadhav:
Still on hold.

neil-marcellini:
We have been waiting for a couple days now with no response, so I just going to assign @situchan

mohammadjafarinejad:
PR under review.

arosiclair:
Dylan will be back on Jan 3rd

PR #71233, opened 2025-09-24
[ORTF] Currency formatting options

mohammadjafarinejad:
### Explanation of Change

Add support for
- `nosymbol` – No currency symbol
- `USD`, `EUR`, etc. – Specific currency

### Fixed Issues

$ https://github.com/Expensify/App/issues/68965
PROPOSAL: https://github.com/Expensify/App/issues/68965#issuecomment-3217428519

### Tests
Same as QA.

### Offline tests
Turn on Force offline mode and do the same steps as in the QA Steps section.
Formula currency formatting should work properly in offline mode.

### QA Steps

> [!NOTE]
> Before running tests, make sure the `useCustomReportNamesNewExpensify` BETA is **explicitly enabled** for your account — it’s disabled by default, even for beta users. Follow the pre-testing steps in [this document](https://docs.google.com/document/d/1q6T1rOSgGu6H29a5TdnW5obvvkAiGE5QeDtlayWNxw8/edit?usp=sharing).
>
> A new rNVP logic was added https://github.com/Expensify/App/issues/67823. **Create reports while online** so the rNVP value from the backend is saved; otherwise, ORTF computation won't work.
>
> Report names may display outdated data — this will be fixed once **Onyx Derived Values** are implemented.

**Precondition:**
1. Navigate to Workspaces
2. Select a workspace where you have admin rights
3. Ensure the workspace’s default currency is USD.
4. Go to the Reports section

---

## Test Case 1 – Multi-Currency Offline Behavior (Non-Zero Values)

**Setup:** 
- Set report title formula to: `{report:total:NPR} | {report:total:USD}`

**Test Steps:**
1. While **ONLINE**, create a new report via FAB
2. Add transaction of $100 USD
3. Wait for title to update with backend values (e.g., `NPR 14,164.25 | $100.00`)
4. Turn **OFFLINE** (Force offline mode)
5. Add another $100 USD transaction (total now $200 USD)
6. **Verify:**
   - Report title shows **grayed out** with old values: `NPR 14,164.25 | $100.00`
   - Frontend skips optimistic calculation (multi-currency involved)
   - Report preview in reports list also shows grayed out
7. Go back **ONLINE**
8. **Verify:** Title updates to new backend-computed values and is no longer grayed out

---

## Test Case 2 – Zero Value Optimistic Computation (Offline)

**Setup:** 
- Set report title formula to: `{report:total} | {report:total:EUR} | {report:total:JPY} | {report:total:nosymbol}`

**Test Steps:**
1. Turn **OFFLINE**
2. Create a new empty report (no transactions, `report.total = 0`)
3. **Verify:** 
   - Title shows: `0.00 | €0.00 | ¥0 | 0.00`
   - All currency values show as 0 even offline
   - Title is **NOT grayed out**
   - Frontend optimistically computes zero values (no backend needed)
4. Go back **ONLINE**
5. **Verify:** Title remains `0.00` for all currencies (no change needed)

---

## Test Case 3 – Single Currency Offline Behavior (No Multi-Currency)

**Setup:** 
- Set report title formula to: `{report:total} | {report:total:usd} | {report:total:nosymbol}`

**Test Steps:**
1. Turn **OFFLINE**
2. Create a new report via FAB
3. Add transaction of $150.50 USD
4. **Verify:**
   - Report title shows: `150.50 | $150.50 | 150.50`
   - Title is **NOT grayed out**
   - Frontend optimistically computes the values (no multi-currency, no backend needed)
   - All values display correctly even while offline
5. Add another $25.10 transaction (total now $175.60)
6. **Verify:** 
   - Title updates to: `175.60 | $175.60 | 175.60`
   - Still NOT grayed out
   - Frontend continues to compute optimistically
7. Go back **ONLINE**
8. **Verify:** Title remains the same (no backend computation was needed)

---

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/781438f6-dc1e-4810-87f6-01489c29c79a

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/bdce6cae-3f94-4b62-939c-b8be31a48830

</details>

<details>

https://github.com/user-attachments/assets/16392f8e-1f1f-413a-ac09-b0d28e76003d

</details>

<details>

https://github.com/user-attachments/assets/06afb93a-af2d-41e2-b4c7-58ccc0d9e907

</details>

<details>

https://github.com/user-attachments/assets/5b0b82fc-4b0d-4b0c-b257-f0990b800ad9

</details>


src/libs/Formula.ts
```diff
@@ -402,19 +402,37 @@ function formatDate(dateString: string | undefined, format = 'yyyy-MM-dd'): stri
...
+
+        // Check if format is a valid currency code (e.g., USD, EUR, eur)
+        // For invalid codes, falls back to '0.00' to match backend behavior
+        const normalizedFormat = format?.trim().toUpperCase();
+        if (normalizedFormat) {
+            if (isValidCurrencyCode(normalizedFormat)) {
+                return convertToDisplayString(absoluteAmount, normalizedFormat);
+            }
+
+            return '0.00';
```
sobitneupane: Should we define a CONST for this value and use it here?

tests/unit/FormulaTest.ts
```diff
@@ -232,6 +228,62 @@ describe('CustomFormula', () => {
...
+                expect(result).toBe('{report:total}');
+            });
+
+            test('should return zero amount when format currency is invalid', () => {
+                currencyContext.report.currency = 'EUR';
+                const result = compute('{report:total:UNKNOWN}', currencyContext);
+                // Should return 0.00 when invalid currency format is used (matches backend behavior)
+                expect(result).toBe('0.00');
+            });
+        });
```
sobitneupane, line 570: @mohammadjafarinejad If the currency is not explicitly defined in the formula but the report currency is defined (say NPR), what should the result be? Could you also add a test for this case if it’s not already covered?

sobitneupane:
@mohammadjafarinejad Have you considered the case where the local currency and the currency format passed are different? Do we convert the amount to the specified currency format before returning?

src/libs/Formula.ts
```diff
@@ -402,19 +402,37 @@ function formatDate(dateString: string | undefined, format = 'yyyy-MM-dd'): stri
 /**
  * Format an amount value
  */
-function formatAmount(amount: number | undefined, currency: string | undefined): string {
+function formatAmount(amount: number | undefined, currency: string | undefined, format?: string): string {
```
sobitneupane: @mohammadjafarinejad `format` doesn't seem like a very good name for the parameter. Should we use something like `displayCurrency`?

mohammadjafarinejad: Agreed, `displayCurrency` is clearer - will update.

mohammadjafarinejad:
> @mohammadjafarinejad Have you considered the case where the local currency and the currency format passed are different? Do we convert the amount to the specified currency format before returning?

Yes, that’s a problem. We don’t have currency rates in the frontend, so we can’t convert amounts currently.

For example, the report name is `$100 - ¥150` (backend converts it correctly). If the report gets updated, `compute()` returns `$100 - ¥100`, and then as soon as the backend response arrives, it goes back to `$100 - ¥150`. This causes the number to flicker in the title.

I’m not sure what the best solution is — we might need to ask the internal team — but here are a few ideas:
1. Store currency ratios in the frontend to allow proper conversion. If the ratio is missing or outdated (more than a few minutes), or if the user is offline, we can skip conversion.

2. When the user is using a currency that is different from the report's currency, we skip the optimistic update of the report name unless the user is offline to avoid the flicker.

src/libs/Formula.ts
mohammadjafarinejad: I think we can also do something like this. When no `currencyCode` is passed to `convertToDisplayStringWithoutCurrency`, it defaults to USD and returns `'0.00'`.

The backend behaves the same way with different report currencies (for example, JPY doesn’t use decimal digits), but regardless of the report currency, it always returns `'0.00'` for invalid ones. So it behaves similarly to `convertToDisplayStringWithoutCurrency`.

```ts
return convertToDisplayStringWithoutCurrency(0);
```

tests/unit/FormulaTest.ts
mohammadjafarinejad: That’s a good point — we can handle it like this, and add a test for it:

```ts
 if (report.currency === currencyCode || isValidCurrencyCode(currencyCode)) {
```

But overall, there can be an inconsistency between the backend’s response and the frontend’s calculated name, when we don't support the currency code, similar to here https://github.com/Expensify/App/pull/71233#issuecomment-3412738174

https://github.com/user-attachments/assets/a487fe0b-6ef9-4dc7-8f32-0663c810bfa5

sobitneupane: @mojammadjafarinejad Just to confirm, the backend does return the report currency when the rule is `{report:total}`, right? The only inconsistency is NPR showing as Rs, correct? I believe that’s not related to this PR - we’re using NPR everywhere else except in the title.

sobitneupane: > That’s a good point — we can handle it like this, and add a test for it:

If [this](https://github.com/Expensify/App/pull/71233#discussion_r2439178308) is the only inconsistency, I believe we should go ahead and make the required change regardless.

mohammadjafarinejad: Yes, by default, the backend uses the report's currency unless we explicitly set it. I added a new unit test for that case.

> If [this](https://github.com/Expensify/App/pull/71233#discussion_r2439178308) is the only inconsistency, I believe we should go ahead and make the required change regardless.

Yes, that is the only inconsistency — for instance, the backend uses the (Rs) symbol for NPR, while in the frontend, we use NPR as the symbol.

sobitneupane:
I haven’t done another full review yet, but based on the screen recordings, I noticed something I don’t like:

After sending 100 NPR expnese in offline, while the NPR amount is not updated, the USD amount is being replaced with the NPR equivalent but still shown with the dollar symbol. Instead of that, can we keep showing the previous USD amount (as we already do with the NPR amount)?

We should continue showing "Report Rs14,164.25 - Formatted $100.71" instead of "Report NPR 14,164.25 - Formatted $14,164.25"


Screenshot 2025-10-20 at 18 46 24

Screenshot 2025-10-20 at 18 46 39

Update: In fact, I think we can update the NPR amount to “Rs 14,264.25”, since the total amount in the report is already updated.

sobitneupane:
> About the currency conversion, unfortunately, I don't think we can do anything because the only data we have at the time of computation is the report.total and the displayCurrency. We don't have the currency rates.

We do have the old value ($100.71 in the above example). Why don’t we continue showing that until we receive the new value from the backend? This is the same approach we use for the **Total** field in the report.

https://github.com/user-attachments/assets/45609dac-5be6-4be0-9df5-f8cdf15616b9

mohammadjafarinejad:
> We do have the old value ($100.71 in the above example). Why don’t we continue showing that until we receive the new value from the backend? This is the same approach we use for the **Total** field in the report.

The old value we have is stored inside `report.name` as a string (in the example above, `report.currency` was NPR, so $100.71 was the converted amount from the backend), and I don’t think we can extract it reliably.

Since users can specify any `displayCurrency` they want in the formula, we run into issues when the `displayCurrency` differs from `report.currency`. As long as the user uses the same currency as `report.currency`, everything works fine.

Let’s use an example to illustrate the problem:

```ts
Formula: `Total {report:total} - EUR {report:total:EUR} - JPY {report:total:JPY}`
report.currency = 'USD';
report.total = 100;
```

**Backend returns:**
```ts
report.name = "Total $100 - EUR €90 - JPY ¥15,000" ✅
```

**Frontend `compute()` produces:**
```ts
report.name = "Total $100 - EUR €100 - JPY ¥100" ❌ (wrong converted amounts)
```

We are also using `formatAmount()` in another PR for the `{report:reimbursable}` amount. In that case, users can do something similar, such as `{report:reimbursable:JPY}`.

In summary, the core issue is that the `formatAmount()` function cannot produce results consistent with the backend, which leads to discrepancies in report names. I think the only way to resolve this is by implementing one of the [four proposed solutions](https://github.com/Expensify/App/pull/71233#issuecomment-3423881203).

mohammadjafarinejad:
> @mohammadjafarinejad Is it ready for review?

@sobitneupane Yes, it's ready for review.

sobitneupane:
So, based on [our discussion,](https://expensify.slack.com/archives/C01GTK53T8Q/p1761252972969709) we’ll gray out the report title if any element in the report formula involves currency conversion, and display the result only when the user is online, right?

> if multi-currency is involved we leave it as is and grey it out as “pending update” for when they come back online,

mohammadjafarinejad:
> So, based on [our discussion,](https://expensify.slack.com/archives/C01GTK53T8Q/p1761252972969709) we’ll gray out the report title if any element in the report formula involves currency conversion, and display the result only when the user is online, right?

Yes, exactly. Until the user is online, we’ll use the definition part (the original formula) in places where multiple currencies are involved.

sobitneupane:
> > `{report:total:invalidcurrency}` is returning it as-is. Earlier, we expected it to return `0.00`, right?
> 
> Invalid currency happens when `report.currency` is different from `displayCurrency`. Since we now leave these untouched and let the backend do the calculation, I thought we don't need to check for them or return `0.00` anymore. Please let me know if I'm wrong about this.

I believe we can determine whether the currency is valid or not on the frontend, right? If the validity can be computed there, we should handle it in the frontend.

It also seems like the backend behavior has changed — it no longer returns 0.00. Did you get a chance to test this on your end?

sobitneupane:
@neil-marcellini Could you please confirm [this](https://github.com/Expensify/App/pull/71233#issuecomment-3460802742)?

mohammadjafarinejad:
> It also seems like the backend behavior has changed — it no longer returns 0.00. Did you get a chance to test this on your end?

Yes, I think the behavior changed—it now returns the formula part when the `displayCurrency` is not valid.

mohammadjafarinejad:
@arosiclair Yes, we need confirmation on one behavior:

**When needsBackendComputation is true (multi-currency involved):**
- **Offline**: Skip FE computation + add `pendingField` → shows the old value grayed out ✅ (confirmed behavior)
- **Online**: Skip FE computation to avoid flicker with backend responses
  - **Question**: Should we also gray out the title while waiting for the backend response? 
  - Note: There is one downside to this currently https://github.com/Expensify/App/pull/71233#issuecomment-3472864431

sobitneupane:
@arosiclair As per the [discussion in the slack](https://expensify.slack.com/archives/C01GTK53T8Q/p1761252972969709), it was decided to use the pattern B and gray-out the title when the user if offline.

However, even when the user is online, it may take some time to receive updates from the backend. During that brief period, should we gray out the title or show the stale data without graying it out?

JmillsExpensify (commented):
A bunch of great reviews on this PR. No concerns from a product perspective.

sobitneupane:
@mohammadjafarinejad  The title doesn't appear grey even when the request is made offline:

https://github.com/user-attachments/assets/27f39d6d-e678-40b9-b468-115d8dabb0e2

arosiclair:
> I believe for the first case (when the value is zero), we can optimistically determine the values in other currencies to be zero and show the full title with the values.

I think that's reasonable. Let's add that if it doesn't add too much complexity.

> For the second case (when the value is non-zero), I am not sure. But maybe we can show the title with the values as zero for all currencies including report currency?

I'm not sure about this one. It doesn't look like we planned for it in the [doc](https://docs.google.com/document/d/1U2AiMOlWvT144a2WVi2XCSc18piyE_iFrv7cYWhsAyw/edit?tab=t.0#heading=h.6rbyhu52x2fv) either. I think it's fine to show blank values, but maybe we could use `???` as a placeholder or something similar. Thoughts @neil-marcellini?

mohammadjafarinejad:
> @mohammadjafarinejad The title doesn't appear grey even when the request is made offline:

Thanks, I’ll look into it.

mohammadjafarinejad:
Yes, we can do that for the first case, we can check `report.total` in case it is 0, and optimistically compute it (which makes `requireBackendComputation` false) both offline and online.


> > For the second case (when the value is non-zero), I am not sure. But maybe we can show the title with the values as zero for all currencies including report currency?
> 
> I'm not sure about this one. It doesn't look like we planned for it in the [doc](https://docs.google.com/document/d/1U2AiMOlWvT144a2WVi2XCSc18piyE_iFrv7cYWhsAyw/edit?tab=t.0#heading=h.6rbyhu52x2fv) either. I think it's fine to show blank values, but maybe we could use `???` as a placeholder or something similar. Thoughts @neil-marcellini?

For the second case, I think the solution in [Slack](https://app.slack.com/client/E047TPA624F/C01GTK53T8Q) was to leave the parts we cannot compute offline to the formula definition.

mohammadjafarinejad:
> @mohammadjafarinejad The title doesn't appear grey even when the request is made offline:

Unfortunately, I cannot reproduce it. Could you please check it again with the new changes?

https://github.com/user-attachments/assets/66d4ab0c-5046-4c2a-a6b3-490b08be2edb

sobitneupane:
I can no longer reproduce [this issue](https://github.com/Expensify/App/pull/71233#issuecomment-3513819622).

### Screenshots/Videos
<details>


https://github.com/user-attachments/assets/d0bad5e9-0ef3-4211-8b87-43fc446d9a90


https://github.com/user-attachments/assets/2822d6e5-438e-446f-aff8-30f01b4def72


https://github.com/user-attachments/assets/5797404e-b577-456f-b3a7-3fb1dceebfb0


https://github.com/user-attachments/assets/9dd1127b-a7d9-4906-b1aa-0be4cae8d097


https://github.com/user-attachments/assets/f511042b-12ed-4ac8-aa7e-5f998bd96c93


https://github.com/user-attachments/assets/510d2c70-99f0-44a3-bda5-6891690e5bae


</details>

mohammadjafarinejad:
> I can no longer reproduce [this issue](https://github.com/Expensify/App/pull/71233#issuecomment-3513819622).

Great, thanks for checking.

src/components/HeaderWithBackButton/index.tsx
```diff
@@ -97,12 +98,17 @@ function HeaderWithBackButton({
...
+                    style={styles.flexGrow1}
+                    pendingAction={report?.pendingFields?.reportName}
+                >
+                    <AvatarWithDisplayName
+                        report={report}
+                        shouldDisplayStatus={shouldDisplayStatus}
+                        shouldEnableDetailPageNavigation={shouldEnableDetailPageNavigation}
+                        openParentReportInCurrentTab={openParentReportInCurrentTab}
+                    />
+                </OfflineWithFeedback>
```
sobitneupane: Instead of doing it here, let's wrap the `HeaderWithBackButton` (or the relevant component) with `OfflineWithFeedback` directly in the specific page where it's needed.

src/libs/OptimisticReportNames.ts
```diff
@@ -290,6 +313,25 @@ function updateOptimisticReportNamesFromUpdates(updates: OnyxUpdate[], context:
...
+        return {
+            key: getReportKey(reportID),
+            onyxMethod: Onyx.METHOD.MERGE,
+            value: {
+                reportName: reportNameUpdate.name,
+                ...(reportNameUpdate.isPending && {
+                    pendingFields: {
+                        reportName: CONST.RED_BRICK_ROAD_PENDING_ACTION.UPDATE,
+                    },
+                }),
```
sobitneupane: I’m not sure about this change. The normal way would be to add `pendingFields` in the optimistic data and remove it in the failure and success data while making the API request.

@mohammadjafarinejad, would it be possible to add `pendingFields` while making API requests like requestMoney, deleteMoney, update, etc., only if we use custom report fields?

mohammadjafarinejad: You're right that's the normal pattern. However, the goal of ORTF was to centralize report name logic in one place since many API calls can affect it (requestMoney, deleteMoney, update, etc.).

If we add pendingFields in each API request, we'd need to either:
- Add it for every change (even when reportName isn't affected), or
- Duplicate the logic to check if reportName needs updating in each function

The benefit of handling it in `OptimisticReportNames.ts` is we can check once whether the report name needs updating and whether we can compute it on the frontend. This avoids duplicating that logic across multiple API functions.

What do you think?

sobitneupane: @mohammadjafarinejad We used to remove the pendingFields by setting them to null in the success and failure data. How should we handle it in this case?

mohammadjafarinejad: I think one way to handle this could be to check in `createReportNameUpdate()`: if the report has `pendingFields.reportName` and `reportNameUpdate.isPending` is false, we can reset it back to null.

src/libs/Formula.ts
```diff
@@ -494,20 +533,45 @@ function getSubstring(value: string, args: string[]): string {
...
+            return convertToDisplayString(absoluteAmount, currencyCode);
+        }
+
+        if (currency && isValidCurrencyCode(currency)) {
+            return convertToDisplayString(absoluteAmount, currency);
+        }
+
+        return convertToDisplayStringWithoutCurrency(absoluteAmount, currency);
+    } catch {
+        return '';
```
arosiclair, line 573: Please log whatever error occurs here.

src/libs/Formula.ts
```diff
@@ -494,20 +533,45 @@ function getSubstring(value: string, args: string[]): string {
...
-    if (currency) {
-        return `${currency}${formattedAmount}`;
-    }
+    try {
+        if (displayCurrency?.trim().toLowerCase() === 'nosymbol') {
+            return convertToDisplayStringWithoutCurrency(absoluteAmount, currency);
+        }
+
+        // Check if format is a valid currency code (e.g., USD, EUR, eur)
+        const currencyCode = displayCurrency?.trim().toUpperCase();
```
arosiclair:
```suggestion
        const trimmedDisplayCurrency = displayCurrency?.trim().toUpperCase();
```

We can also move this up and use it in the above condition

src/libs/OptimisticReportNames.ts
sobitneupane: I think we should bring in someone who has worked on this project for some suggestions.

cc: @arosiclair

src/libs/OptimisticReportNames.ts
```diff
@@ -259,6 +264,22 @@ function computeReportNameIfNeeded(report: Report | undefined, incomingUpdate: O
         allTransactions: context.allTransactions,
     };
 
+    const needsBackendComputation = requiresBackendComputation(formulaParts, formulaContext);
```
arosiclair:
```suggestion
```

Let's just call this in the if condition below

src/libs/OptimisticReportNames.ts
arosiclair: Yeah normally we would set pending fields in `optimisticData` and clear them in the `successData` and `failureData` for each API request. With the current code, we set the pending field in `optimisticData` but never clear it.

If we want to keep all logic here in the middleware, then I think we need `updateOptimisticReportNamesFromUpdates` to also update the success and failure data for the command. Can you make that work @mohammadjafarinejad?

arosiclair (changes requested):

mohammadjafarinejad:
### New Screenshots/Videos
<details>

https://github.com/user-attachments/assets/f23a85c0-5982-42ae-8b43-175ce312a287

</details>

<details>

https://github.com/user-attachments/assets/83381e73-f200-4689-93e1-d37090cde6a1

</details>

<details>

https://github.com/user-attachments/assets/e0aeaee8-a7e7-4f0d-8b73-fc2cc036204e

</details>

<details>

https://github.com/user-attachments/assets/45d485a2-7f21-4bad-8f16-f452eb4c6771

</details>

<details>

https://github.com/user-attachments/assets/ffe0de29-42c3-43d4-a9c0-e305bb3e341d

</details>

<details>

https://github.com/user-attachments/assets/20d1b3ba-d29e-4e1a-b774-63e31ecdf6e3

</details>

src/libs/OptimisticReportNames.ts
mohammadjafarinejad: Done ✅

src/libs/Formula.ts
```diff
@@ -494,20 +534,46 @@ function getSubstring(value: string, args: string[]): string {
...
     const absoluteAmount = Math.abs(amount);
-    const formattedAmount = (absoluteAmount / 100).toFixed(2);
 
-    if (currency) {
-        return `${currency}${formattedAmount}`;
-    }
+    try {
+        const trimmedDisplayCurrency = displayCurrency?.trim().toUpperCase();
+        if (trimmedDisplayCurrency) {
+            if (trimmedDisplayCurrency?.toLowerCase() === 'nosymbol') {
```
arosiclair:
```suggestion
            if (trimmedDisplayCurrency === 'NOSYMBOL') {
```

We can simplify this

src/libs/API/index.ts
```diff
@@ -96,8 +100,14 @@ function prepareRequest<TCommand extends ApiCommand>(
...
             try {
                 const context = getUpdateContext();
-                const processedOptimisticData = OptimisticReportNames.updateOptimisticReportNamesFromUpdates(optimisticData, context);
+                const {updates: processedOptimisticData, reportsWithPendingFields} = OptimisticReportNames.updateOptimisticReportNamesFromUpdates(optimisticData, context);
                 Onyx.update(processedOptimisticData);
+
+                // Process success and failure data to clear pending fields
+                if (reportsWithPendingFields) {
+                    processedSuccessData = OptimisticReportNames.clearPendingFieldsForReports(successData, reportsWithPendingFields);
+                    processedFailureData = OptimisticReportNames.clearPendingFieldsForReports(failureData, reportsWithPendingFields);
```
arosiclair: Can we return the processed optimistic, success and failure data all from `updateOptimisticReportNamesFromUpdates` instead? It'll be simpler to follow if all of the logic comes from one function.

arosiclair (changes requested):

src/libs/API/index.ts
mohammadjafarinejad: Yes, we can—updated the function to return all three.

src/libs/OptimisticReportNames.ts
```diff
@@ -368,10 +409,16 @@ function updateOptimisticReportNamesFromUpdates(updates: OnyxUpdate[], context:
...
-        additionalUpdatesCount: additionalUpdates.length,
+        additionalOptimisticUpdatesCount: additionalOptimisticUpdates.length,
+        additionalSuccessUpdatesCount: additionalSuccessUpdates.length,
+        additionalFailureUpdatesCount: additionalFailureUpdates.length,
     });
 
-    return updates.concat(additionalUpdates);
+    return {
+        optimisticData: optimisticData.concat(additionalOptimisticUpdates),
+        successData: successData ? successData.concat(additionalSuccessUpdates) : undefined,
```
arosiclair: Shouldn't we always return the additional updates even if the original data is empty?

```suggestion
        successData: successData ? successData.concat(additionalSuccessUpdates) : additionalSuccessUpdates,
```

src/libs/OptimisticReportNames.ts
```diff
@@ -368,10 +409,16 @@ function updateOptimisticReportNamesFromUpdates(updates: OnyxUpdate[], context:
...
+        failureData: failureData ? failureData.concat(additionalFailureUpdates) : undefined,
```
arosiclair:
```suggestion
        failureData: failureData ? failureData.concat(additionalFailureUpdates) : additionalFailureUpdates,
```

src/libs/OptimisticReportNames.ts
```diff
@@ -278,21 +305,56 @@ function computeReportNameIfNeeded(report: Report | undefined, incomingUpdate: O
...
 
-    const additionalUpdates: OnyxUpdate[] = [];
+    const additionalOptimisticUpdates: OnyxUpdate[] = [];
+    const additionalSuccessUpdates: OnyxUpdate[] = [];
+    const additionalFailureUpdates: OnyxUpdate[] = [];
+
+    /**
+     * Helper function to add report name updates to the appropriate arrays
+     */
+    function addUpdates(reportID: string, reportNameUpdate: ReportNameUpdate) {
```
arosiclair: Let's move this function out of `updateOptimisticReportNamesFromUpdates` instead of nesting it to make this easier to read.

arosiclair (changes requested):

arosiclair (commented):
Code LGTM. Ready for your review @sobitneupane. There are some tests we should fix though.

src/components/MoneyReportHeader.tsx
```diff
@@ -1371,46 +1372,48 @@ function MoneyReportHeader({
...
-                        )}
-                        {shouldShowSelectedTransactionsButton && (
-                            <View>
-                                <ButtonWithDropdownMenu
-                                    onPress={() => null}
-                                    options={selectedTransactionsOptions}
-                                    customText={translate('workspace.common.selected', {count: selectedTransactionIDs.length})}
-                                    isSplitButton={false}
-                                    shouldAlwaysShowDropdownMenu
+            <OfflineWithFeedback pendingAction={moneyRequestReport?.pendingFields?.reportName}>
```
sobitneupane: @mohammadjafarinejad Instead of adding `OfflineWithFeedback`, let's update `getReportOfflinePendingActionAndErrors` in ReportUtils to include `pendingFields?.reportName`

mohammadjafarinejad: @sobitneupane Updated.

sobitneupane (approved):
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

sobitneupane:
@mohammadjafarinejad Could you please check the screen recordings for Android: HybridApp? The behavior on Android seems different from the other platforms.. When I create a new report or add an expense to an existing report, the title includes part of the report title definition instead of values.

For example:
Screenshot 2025-11-25 at 17 57 41

mohammadjafarinejad:
> @mohammadjafarinejad Could you please check the screen recordings for Android: HybridApp? The behavior on Android seems different from the other platforms.. When I create a new report or add an expense to an existing report, the title includes part of the report title definition instead of values.

@sobitneupane, I cannot reproduce this issue. Could you please test it again?
Also, it should skip the offline computation when we add a new expense, but in the "Android: HybridApp" video you shared, it doesn't seem to be doing that either.

https://github.com/user-attachments/assets/852a00f6-ff72-4482-84b4-f47152cf9565

sobitneupane (approved):

sobitneupane:
@arosiclair True. That should not be the case. I missed it.

@mohammadjafarinejad Could you please check this? I tried again and can still reproduce it.

sobitneupane:
> @sobitneupane I think the beta tag isn't enabled on the testing account. It's using `populateOptimisticReportFormula()` instead of ORTF, so the report title is cut off. Can you please return true manually here to see if it fixes the problem?
> 
> https://github.com/Expensify/App/blob/16b8a9d9c292c68285c7c5162fcde5ac389fd5e6/src/libs/Permissions.ts#L18

Oops, my bad. That might be the case. Let me test it again. I think I ran it without any changes the last time.

arosiclair (approved):

sobitneupane:
> Regarding the https://github.com/Expensify/App/issues/76770, I believe this is expected behavior. The backend is also returning the fallback name:

Yup, I believe that's the expected behavior. It might be due to outdated test steps — it looks like there are new steps in place now.

PR #71997, opened 2025-10-07
[CP Staging] Revert "showing "marked as paid" in ND when marked as reimbursed from OD"

lakchote:
Reverts Expensify/App#71497


mountiny (approved):
Straight revert!

PR #73512, opened 2025-10-27
[No QA] Update Uber.md

stephanieelliott:
https://expensify.slack.com/archives/C02QSAC6BJ8/p1761328498436759


zanyrenney (approved):
LGTM. The "GH actions" still seem to be buggy though.

It says "Terminology violation: Use "tap" consistently instead of "click" for mobile app interactions. Since this is in the Uber mobile app, "tap" is more accurate." But it already said "tap".

PR #74108, opened 2025-11-03
Make Switching to the Search Page Faster

sumo-slonik:
### Explanation of Change

Thanks to cleaning up the structure of the search page component, we were able to speed up the time it takes to switch to the Search tab. Below are metrics from several platforms as an example of the time gains:

Desktop web : **2.1s → 1.1s**

iOS native ‎ : **6,04s → 4.5s**

This shows that we’re managing to save roughly half of the time previously needed to switch tabs on the web, and also more than 1 second on the native platform.

https://github.com/user-attachments/assets/261003c1-f738-47e8-a9b4-55b673adf340

### Fixed Issues

$ https://github.com/Expensify/App/issues/74613
PROPOSAL:

### Tests
1. Go to the Search tab.

2. Check whether the results are displayed correctly.

3. Make sure you can correctly open all the filters
4. Verify you can select multiple expense / reports and open the options for bulk actions

### Offline tests
unnesesary

### QA Steps
Same as test

### PR Author Checklist

    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X]  iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
    - [X] MacOS: Desktop

### Screenshots/Videos
<details>

</details>

<details>

https://github.com/user-attachments/assets/f1176b63-01a9-4504-a3bc-3ae8e3ebd48e

</details>

<details>

https://github.com/user-attachments/assets/1ff91a96-fce2-43dc-aa82-95f4144601df

</details>

<details>

https://github.com/user-attachments/assets/ae5b3b90-ea4d-4da4-b2f5-8a663ceac872

</details>

<details>

https://github.com/user-attachments/assets/173b1ede-dc98-4d3e-b387-67777ac15222

</details>

<details>

https://github.com/user-attachments/assets/7a2a677c-586e-4b7c-939d-f7de120a4724

</details>


src/pages/Search/SearchPage.tsx
```diff
@@ -818,6 +828,15 @@ function SearchPage({route}: SearchPageProps) {
         }
     }, []);
 
+    useEffect(() => {
+        if (!queryJSON || !isDataMissing || !isFocused) {
+            return;
+        }
+        handleSearchAction({queryJSON, searchKey, offset: 0, shouldCalculateTotals});
+        // eslint-disable-next-line react-compiler/react-compiler
```
adhorodyski: I know it's not yet enforced anywhere, but we really shouldn't integrate any new code that disables the compiler, especially so high up in the components' tree. Can you rethink how this can be implemented following the rules of React?

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
```diff
@@ -24,8 +24,8 @@ import {getPreservedNavigatorState} from './createSplitNavigator/usePreserveNavi
 // This timing is used to call the preload function after a tab change, when the initial tab screen has already been rendered.
 const TIMING_TO_CALL_PRELOAD = 1000;
 
-// Currently, only the Account and Workspaces tabs are preloaded. The remaining tabs will be supported soon.
-const TABS_TO_PRELOAD = [NAVIGATION_TABS.SETTINGS, NAVIGATION_TABS.WORKSPACES];
+// Currently, only the Inbox, Workspaces, Account tabs are preloaded. The remaining tabs will be supported soon.
+const TABS_TO_PRELOAD = [NAVIGATION_TABS.HOME, NAVIGATION_TABS.SEARCH, NAVIGATION_TABS.WORKSPACES, NAVIGATION_TABS.SETTINGS];
```
adhorodyski: From the PR, this is the only place where we actually turn on the preloading logic. Given that, why can't it be integrated on it's own as a separate change?

src/pages/Search/SearchModalsWrapper.tsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react';
+import {View} from 'react-native';
+import ConfirmModal from '@components/ConfirmModal';
+import DecisionModal from '@components/DecisionModal';
+import useLocalize from '@hooks/useLocalize';
+import useResponsiveLayout from '@hooks/useResponsiveLayout';
+import {openOldDotLink} from '@userActions/Link';
+import CONST from '@src/CONST';
+
+type SearchModalsWrapperProps = {
```
adhorodyski: Looking at this interface and the amount of responsibilities within the component, this is the very opposite of writing composable React. This will only grow in size, is error prone and hard to maintain from the very start. Can you explain why it has to be implemented (and why this way) for the preloading?

src/pages/Search/SearchPage.tsx
```diff
@@ -120,15 +125,24 @@ function SearchPage({route}: SearchPageProps) {
     const queryJSON = useMemo(() => buildSearchQueryJSON(route.params.q), [route.params.q]);
     const {saveScrollOffset} = useContext(ScrollOffsetContext);
     const activeAdminPolicies = getActiveAdminWorkspaces(policies, currentUserPersonalDetails?.accountID.toString()).sort((a, b) => localeCompare(a.name || '', b.name || ''));
-
+    const {accountID} = useCurrentUserPersonalDetails();
+    const {defaultCardFeed} = useCardFeedsForDisplay();
+    const suggestedSearches = useMemo(() => getSuggestedSearches(accountID, defaultCardFeed?.id), [defaultCardFeed?.id, accountID]);
```
adhorodyski: My Chrome tab couldn't record a performance profile for me when working with this branch on the `exfy-perf@callstack.com` account (Applause workspace). I can't tell for sure, but might this be due to these calculations?

adhorodyski: FYI I had the same by only preloading the Search screen, no additional changes (for this account).

src/pages/Search/SearchPage.tsx
```diff
@@ -816,243 +847,137 @@ function SearchPage({route}: SearchPageProps) {
...
+                            scrollHandler={scrollHandler}
+                            initScanRequest={initScanRequest}
+                            PDFValidationComponent={PDFValidationComponent}
+                            ErrorModal={ErrorModal}
+                            shouldShowOfflineIndicator={shouldShowOfflineIndicator}
+                            offlineIndicatorStyle={offlineIndicatorStyle}
+                            shouldShowFooter={shouldShowFooter}
+                        />
+                    )}
+                    {(!shouldUseNarrowLayout || isMobileSelectionModeEnabled) && (
```
adhorodyski: This references my previous comment. I see a lot of the changes in the PR could be decoupled from the preloading itself - it looks like much of the performance gain might be coming from disabling a part of the React tree for the initial load. Maybe it can be integrated separately?

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
adhorodyski: I'd like to call out [the docs for `navigation.preload`](https://reactnavigation.org/docs/navigation-object/#preload):

> Preloading a screen means that the screen will be rendered in the background. All the components in the screen will be mounted and the useEffect hooks will be called.

As much as it comes with a good visual effect in isolated testing, at scale this can behave much like a heavy React Context - kill the general JS performance by triggering a lot of the logic in the background, in a much less predictable way. I see (and like) the direction of another part of this PR where we disable the tree when it's not needed, but I'd really emphasize on the importance of performance testing with large accounts before we integrate preloading for this heavy screen (and others). 

`Search` just does a lot of the heavy lifting, and doing this heavy work at a different time (also instantiating effects) can potentially cause us harm, one that is super hard to track, debug and fix.

Preloading itself might yield good results when we want to keep a cheap screen in the memory, but is not a general optimization strategy for really complex (and alive) screens.

src/pages/Search/SearchPage.tsx
sumo-slonik: I agree that we should follow that approach but in this particular case I don’t think it will make much difference because the SearchPage already has a lot of issues with the React compiler even without this useEffect, which means it isn’t being compiled anyway. So adding or removing this useEffect won’t really change that situation.

src/libs/Navigation/AppNavigator/usePreloadFullScreenNavigators.ts
sumo-slonik: The changes I added were also required for enabling preloading by splitting the SearchPage into smaller components, it became easier to preload only the skeletons. Thanks to that, the entire heavy screen with all search results is not kept in memory, only the lightweight skeletons are.

During preloading, useEffect and useOnyx hooks are still executed, but the actual search is not triggered since it only runs when isFocused is true. This approach helps slightly reduce the issue you mentioned regarding keeping a heavy screen in memory.

Additionally, together with @WojtekBoman, we’re planning to adjust the preloading logic so that not all screens are preloaded at once. We want to spread this over time to optimize performance, but that will require a separate PR.

Given all that, I think keeping preloading here is still worth it despite the trade-offs involved.

src/pages/Search/SearchModalsWrapper.tsx
sumo-slonik: It’s not required specifically because of preloading the main reason was that these components were repeatedly used in SearchPage, which had over 1000 lines of code. I figured that any opportunity to make it a bit lighter and more modular was worth taking.

sumo-slonik: If you feel that this wrapper is unnecessary, I can remove it, but in my opinion, it simplifies the structure of the `SearchPage` itself, which is already huge.

src/pages/Search/SearchPage.tsx
sumo-slonik: I think the difficulty in observing this in the profiler may be due to the calculations performed during preloading. That’s why I chose to include a screen recording rather than rely solely on metrics , even if the metrics don’t show a significant difference, the user experience is noticeably improved.

src/pages/Search/SearchPage.tsx
sumo-slonik: I agree that some of the changes aren’t directly related to preloading, but splitting this into two separate PRs might not be the best idea either. Some of these changes were aimed at fixing bugs that appeared during preloading, such as blank flickering screens or results not being displayed, and at the same time, refactoring this component was necessary to properly store the skeletons as a preloaded screen.

src/pages/Search/SearchPage.tsx
sumo-slonik: I need to do some research, but it’s possible that this is due to the time required by passive effects.

src/pages/Search/SearchPage.tsx
sumo-slonik: After testing, I think I’ll do it the way you suggested and make sure it doesn’t conflict with the compiler.

src/pages/Search/SearchPage.tsx
sumo-slonik: @adhorodyski I’ll double-check with more tests, but it seems that decomposition brings most of the benefit, while preload adds less and might not be worth including. Removing preload bugs solutions also improves results, so decomposition alone already looks good. I’ll share detailed results tomorrow.

src/pages/Search/SearchModalsWrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+                    title={translate('search.exportSearchResults.title')}
+                    prompt={translate('search.exportSearchResults.description')}
+                    confirmText={translate('search.exportSearchResults.title')}
+                    cancelText={translate('common.cancel')}
+                />
+            )}
+        </View>
+    );
+}
+
```
linhvovan29546: Missing `displayName`

src/pages/Search/SearchPage.tsx
```diff
@@ -802,243 +830,102 @@ function SearchPage({route}: SearchPageProps) {
...
+
+            saveScrollOffset(route, e.nativeEvent.contentOffset.y);
+        },
+        [saveScrollOffset, route],
+    );
+
+    return (
+        <FullPageNotFoundView
+            shouldForceFullScreen
+            shouldShow={!queryJSON}
```
linhvovan29546: Both `SearchPageNarrow` and `SearchPageWide` has `queryJSON` condition. `SearchPageNarrow` has this condition: https://github.com/Expensify/App/blob/94035e1be76868d88bcf646111b00ef4f304593e/src/pages/Search/SearchPageNarrow.tsx#L155-L170
Previously, the condition above handled an empty `queryJSON` by showing the 'not found' page. The new change wraps the page with `FullPageNotFoundView` and handle logic to display not found page, which may cause a regression because it never trigger the condition inside `SearchPageNarrow`

Edit: Then maybe we can remove this logic inside `SearchPageNarrow` and `SearchPageWide`?

linhvovan29546: Bug: this is just an edge case that happens when `queryJSON` is underfined, then the header is overlap status bar
Simulator Screenshot - iPhone 16 Pro Max - 2025-11-19 at 17 18 30

sumo-slonik:
I'm already getting to work on addressing the suggestions.

src/pages/Search/SearchPage.tsx
sumo-slonik: Good catch. I think it would be better to fix this by removing the ‘Page Not Found’ at the level of the narrow and wide components, because they have a search wrapper, which should resolve the bug you pointed out here.

linhvovan29546: Then We're missing the 'Page not found' logic in the wide component

linhvovan29546 (commented):
Overall looks good, there's just one issue
https://github.com/Expensify/App/pull/74108#discussion_r2545388827

src/pages/Search/SearchModalsWrapper.tsx
adhorodyski: Honestly I think this is something that we'll eventually aim to avoid, so I'd go ahead and remove it. If the structure it complex, it's definitely not a viable solution to create a giga wrapper that is a prop sink for every piece of functionality that lives in there.

Let's focus on the gains coming from not rendering parts of the UI - the rest we can solve later by better composing the screens.

src/pages/Search/SearchPage.tsx
```diff
@@ -802,243 +828,97 @@ function SearchPage({route}: SearchPageProps) {
...
+                                dropStyles={styles.receiptDropOverlay(true)}
+                                dropTextStyles={styles.receiptDropText}
+                                dropWrapperStyles={{marginBottom: variables.bottomTabHeight}}
+                                dashedBorderStyles={[styles.dropzoneArea, styles.easeInOpacityTransition, styles.activeDropzoneDashedBorder(theme.receiptDropBorderColorActive, true)]}
+                            />
+                        </DragAndDropConsumer>
+                        {ErrorModal}
+                    </DragAndDropProvider>
+                ) : (
+                    <SearchPageWide
```
adhorodyski, line 890: The 'wide' variant does not rely on any custom logic. Can we instead put its content directly in here to avoid drilling the props?

sumo-slonik: It seems to me that there’s no real issue with placing it here, but it’s the same situation as with the modals. Do we want the `SearchPage` to contain all of that content, or do we prefer to keep this component separate? Personally, I think keeping it separated makes things a bit clearer (especially since the narrow page is also in its own component). But if you strongly feel it should live here, I can move it back.

I also have one more point regarding the ModalWrapper. You suggested removing it, and after thinking it over, I agree, I’m leaning more toward dropping it and adding the modals directly here, slightly more than in the case of the `SearchPageWide`.

sumo-slonik:
image

image

image

I repeated the measurements and the average result dropped by half. I wanted to make sure that nothing got messed up by the changes. In the morning, I’ll check all the platforms to ensure everything is working as it should, and we should be ready for the final part of the review, I’ll make any minor fixes if I notice something. After the review, I’ll resolve the conflicts, and we should be ready for merge.

sumo-slonik:
image

Fixed!

linhvovan29546:
> image
> Fixed!

I don't see the offline indicator on the wide screen. Could you please check again?

src/pages/Search/SearchPageWide.tsx
```diff
@@ -81,6 +81,7 @@ function SearchPageWide({
                 testID={Search.displayName}
                 shouldEnableMaxHeight
                 headerGapStyles={[styles.searchHeaderGap, styles.h0]}
+                shouldShowOfflineIndicatorInWideScreen={!!searchResults}
```
linhvovan29546, line 92: Missing `offlineIndicatorStyle` which will cause a bug.

Screenshot 2025-11-22 at 09 13 41

```
 const offlineIndicatorStyle = useMemo(() => {
        if (shouldShowFooter) {
            return [styles.mtAuto, styles.pAbsolute, styles.h10, styles.b0];
        }

        return [styles.mtAuto];
    }, [shouldShowFooter, styles]);
```

linhvovan29546:
> @linhvovan29546 how is this looking?

Only one minor UI bug left

linhvovan29546:
I’ve tested across all platforms. Only one minor UI bug remains. I’ll re-review once the conflict is resolved

src/pages/Search/SearchPageWide.tsx
sumo-slonik: > Missing `offlineIndicatorStyle` which will cause a bug.
> 
> Screenshot 2025-11-22 at 09 13 41
> ```
>  const offlineIndicatorStyle = useMemo(() => {
>         if (shouldShowFooter) {
>             return [styles.mtAuto, styles.pAbsolute, styles.h10, styles.b0];
>         }
> 
>         return [styles.mtAuto];
>     }, [shouldShowFooter, styles]);
> ```


Thank you, I’m fixing it now.

linhvovan29546:
Yeah please resolve the conflicts

linhvovan29546 (approved):

adhorodyski:
Reviewing now.

src/pages/Search/SearchPage.tsx
adhorodyski: I think I just had problems with the amount of prop drilling, no solution is perfect but eventually we may be able to abstract out the modals as well.

adhorodyski (approved):

src/pages/Search/SearchPage.tsx
```diff
@@ -614,22 +619,22 @@ function SearchPage({route}: SearchPageProps) {
...
         styles.textWrap,
-        beginExportWithTemplate,
-        integrationsExportTemplates,
-        csvExportLayouts,
-        policies,
-        bulkPayButtonOptions,
-        onBulkPaySelected,
-        selectedPolicyIDs,
-        selectedReportIDs,
-        selectedTransactionReportIDs,
```
mountiny, line 632: In future, it would be great to avoid any changes like this that just increase the diff of the pr and make it harder to review

src/pages/Search/SearchPage.tsx
```diff
@@ -186,7 +184,14 @@ function SearchPage({route}: SearchPageProps) {
...
             if (areAllMatchingItemsSelected) {
-                queueExportSearchWithTemplate({templateName, templateType, jsonQuery: JSON.stringify(queryJSON), reportIDList: [], transactionIDList: [], policyID});
+                queueExportSearchWithTemplate({
+                    templateName,
+                    templateType,
+                    jsonQuery: JSON.stringify(queryJSON),
+                    reportIDList: [],
+                    transactionIDList: [],
+                    policyID,
+                });
```
mountiny, line 194: here as well as far as I can see is no change, maybe if this was a small PR, then feel free to make style changes, but in large prs like this one, it is counter productive 🙌

src/pages/Search/SearchPageWide.tsx
```diff
@@ -0,0 +1,156 @@
...
+    selectedReportIDs: string[];
+    latestBankItems?: BankAccountMenuItem[];
+    onBulkPaySelected: (paymentMethod?: PaymentMethodType) => void;
+    handleSearchAction: (value: SearchParams | string) => void;
+    onSortPressedCallback: () => void;
+    scrollHandler: (event: NativeSyntheticEvent<NativeScrollEvent>) => void;
+    initScanRequest: (e: DragEvent) => void;
+    PDFValidationComponent: React.ReactNode;
+    ErrorModal: React.ReactNode;
+    shouldShowFooter: boolean;
```
mountiny, line 48: All of these should have docs, can you add them in a follow up please?

mountiny (approved):

src/pages/Search/SearchPageWide.tsx
sumo-slonik: I’ll take care of it once I’m done with the inbox.

src/pages/Search/SearchPageWide.tsx
```diff
@@ -0,0 +1,156 @@
...
+                offlineIndicatorStyle={offlineIndicatorStyle}
+            >
+                <FullPageNotFoundView
+                    shouldForceFullScreen
+                    shouldShow={!queryJSON}
+                    onBackButtonPress={handleOnBackButtonPress}
+                    shouldShowLink={false}
+                >
+                    {!!queryJSON && (
+                        <>
```
linhvovan29546, line 102: After this PR we accidentally missed `DragAndDropProvider`, which led to [issue #76986](https://github.com/Expensify/App/issues/76986). We've applied the fix [here](https://github.com/Expensify/App/pull/77723)

PR #74566, opened 2025-11-07
[Reduce App Bundle] Load PDF worker dynamically

JKobrynski:
### Fixed Issues

$ https://github.com/Expensify/App/issues/75683
PROPOSAL: https://callstack-hq.slack.com/archives/C05LX9D6E07/p1763480893359809

### Tests

1. Log into the web app
2. Navigate to any chat
3. Drag and drop a pdf file 
4. Make sure it loads and displays correctly
5. Navigate to a different screen, e.g. "Account"
6. Refresh the app
7. In the Network tab, change throttling to "Offline"
8. Navigate to any chat
9. Drag and drop a pdf file
10. Make sure it loads and displays correctly
11. Repeat on desktop

Steps 5-10 ensure that we have offline support.

### Offline tests

1. Log into the web app
2. In the Network tab, change throttling to "Offline"
3. Navigate to any chat
4. Drag and drop a pdf file
5. Make sure it loads and displays correctly
6. Repeat on desktop

### QA Steps

Same as **Tests** section above

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
      - [x] If any non-english text was added/modified, I used [JaimeGPT](https://chatgpt.com/g/g-2dgOQl5VM-english-to-spanish-translator-aka-jaimegpt) to get English > Spanish translation. I then posted it in #expensify-open-source and it was approved by an internal Expensify engineer. Link to Slack message: https://callstack-hq.slack.com/archives/C01GTK53T8Q/p1763736977259489

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/b52f1df3-3881-4fcf-8240-ec64c515b3b9

</details>

<details>

https://github.com/user-attachments/assets/c4f0b2ae-1137-439b-bacc-e3fe726ad325

</details>


src/components/PDFThumbnail/index.tsx
```diff
@@ -1,59 +1,72 @@
...
-                }}
-                error={() => null}
-            >
-                <View pointerEvents="none">
-                    <Thumbnail pageIndex={0} />
-                </View>
-            </Document>
+            <Suspense fallback={null}>
+                <Document
+                    loading={<FullScreenLoadingIndicator />}
```
adhorodyski: `<Component />` creates this component on every render. You can create it once and use a reference here.

src/components/PDFThumbnail/index.tsx
```diff
@@ -1,59 +1,72 @@
...
+                    file={isAuthTokenRequired ? addEncryptedAuthTokenToURL(previewSourceURL) : previewSourceURL}
+                    options={{
+                        cMapUrl: 'cmaps/',
+                        cMapPacked: true,
+                    }}
+                    externalLinkTarget="_blank"
+                    onPassword={onPassword}
+                    onLoad={() => {
```
adhorodyski: Do not inline handlers

src/components/PDFThumbnail/pdfSetup.ts
```diff
@@ -0,0 +1,23 @@
+let pdfSetupPromise: Promise<unknown> | null = null;
+
+function ensurePdfJsInitialized() {
+    if (!pdfSetupPromise) {
```
adhorodyski: Let's early return if the promise is there instead to prevent nesting.

Mobile-Expensify
VickyStash, line 1: Is it expected to have these updates?

JKobrynski: Not at all, my bad

config/webpack/webpack.common.ts
```diff
@@ -387,6 +387,20 @@ const getCommonConfiguration = ({file = '.env', platform = 'web'}: Environment):
...
+                        name: 'reactFastPdf',
+                        chunks: 'all',
+                    },
+                    // react-pdf library had a pdfjs-dist peer dependency that increases the size of vendors bundle
+                    // we want to load it separately to decrease the bundle size
+                    reactPdf: {
+                        test: /[\\/]node_modules[\\/](react-pdf)[\\/]/,
+                        name: 'reactPdf',
+                        chunks: 'all',
+                    },
```
kacper-mikolajczak, line 375: Curious, is webpack smart enough to extract pdfjs-dist from main bundle and load it together with those?

src/components/PDFThumbnail/index.tsx
```diff
@@ -1,59 +1,72 @@
...
 function PDFThumbnail({previewSourceURL, style, isAuthTokenRequired = false, enabled = true, onPassword, onLoadError, onLoadSuccess}: PDFThumbnailProps) {
     const styles = useThemeStyles();
     const [failedToLoad, setFailedToLoad] = useState(false);
+    const [ready, setReady] = useState(false);
+
+    useEffect(() => {
+        ensurePdfJsInitialized().then(() => {
+            setReady(true);
+        });
+    }, []);
```
kacper-mikolajczak, line 30: What would happen if module fails to load (e.g. device went offline)? Are errors handled for that matter?

src/components/PDFThumbnail/pdfSetup.ts
```diff
@@ -0,0 +1,23 @@
...
+            import(/* webpackPreload: true */ 'react-pdf'),
+            // eslint-disable-next-line import/extensions
+            import(/* webpackPreload: true */ 'pdfjs-dist/build/pdf.worker.min.mjs'),
+        ]).then(([reactPdfModule, pdfWorkerSource]) => {
+            const {pdfjs} = reactPdfModule;
+
+            if (!pdfjs.GlobalWorkerOptions.workerSrc) {
+                pdfjs.GlobalWorkerOptions.workerSrc = URL.createObjectURL(new Blob([pdfWorkerSource], {type: 'text/javascript'}));
+            }
+
```
kacper-mikolajczak: Curious again :D Could you tell why we need this? Correct me if I am wrong, but I haven't seen this worker piece before and I am just wondering what is the difference between dynamic and static load in this matter.

src/hooks/usePreloadLazyModules.ts
```diff
@@ -0,0 +1,8 @@
+// This hook makes it possible to manually load the lazy loaded modules
+// right after the main bundle is loaded to decrease the bundle size
+function usePreloadLazyModules() {
+    import(/* webpackPreload: true */ 'react-fast-pdf');
+    import(/* webpackPreload: true */ 'react-pdf');
+}
```
kacper-mikolajczak: Could you check what is the impact on app's behaviour with vs without this hook? 

My assumption behind this test is to confirm if it is needed. When passing `webpackPreload: true` for dynamic imports, I think we don't need to manually make this import in runtime - webpack should be able to pick it up.

src/components/PDFThumbnail/index.tsx
kacper-mikolajczak: Could you tell me a bit what is reason behind wrapping this with Suspense? Thanks!

kacper-mikolajczak (changes requested):
Hi @JKobrynski! I left couple of comments which could provide more insight on implemented changes.

Also, as there are couple of imports already + custom resolver in form of `ensurePdfJsInitialized`, I was wondering if we possibly could benefit from containing all this pdf jazz into a simple module that:

- prepares libraries to work as a single unit
- exposes needed APIs

I am wondering whether preload is the best pick for this particular use-case. We might check how `prefetch` performs here (take a look at webpack [docs regarding preload vs prefetch](https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules))

Thanks a lot for this effort and answering my concerns! ❤️

config/webpack/webpack.common.ts
JKobrynski: Looking at the bundle analyzer, it seems like it is the case, the big chunk that you can see [here](https://github.com/callstack-internal/expensify-issues/issues/1749#issue-3561681385) disappears from it. Do you think there is a better way to check this? And does that answer your question?

src/components/PDFThumbnail/index.tsx
JKobrynski: I haven't tested that yet, good point

src/components/PDFThumbnail/index.tsx
JKobrynski: That component is imported with `React.lazy` so according to the pattern we should wrap this with Suspense. Do you think we should not use it in this particular case? This library is loaded dynamically with the hook, so maybe we could test if this works correctly without it. Wdyt?

src/hooks/usePreloadLazyModules.ts
JKobrynski: My initial implementation didn't include that hook. I tested different webpack options, like preload, prefetch and mode. The problem that I kept running into, was the chunks only loading when a related component is rendered, they didn't load right after the main bundle like `heic-to`. I think that might be due to the usage of `React.lazy`, that for some reason it doesn't fully support the webpack settings. Do you have any thoughts on this? Maybe you ran into similar problems? I will most likely reach out to Jakub as it seems like an edge case to me

I will do another round of testing just to make sure, if anything changes I will let you know!

src/components/PDFThumbnail/pdfSetup.ts
JKobrynski: From what I've seen in the git history, it's been there in some form for a long time 😄 looking at [this part](https://github.com/wojtekmaj/react-pdf/blob/main/packages/react-pdf/README.md#import-worker-recommended) of the documentation, it looks like it's the recommended approach. However, it also mentions that this part should be included in the same file that uses the components, so I will have to retest my implementation and check if it can be moved to that file. It's great that you asked about this, thanks!

config/webpack/webpack.common.ts
kacper-mikolajczak: Great, if it does! Thanks!

src/hooks/usePreloadLazyModules.ts
kacper-mikolajczak: Check setting eager mode, it might help: https://github.com/Expensify/App/blob/b4355e431e46330b047215f5b01cb29a2555ac6f/src/libs/fileDownload/heicConverter/index.ts#L12-L14.

src/components/PDFThumbnail/index.tsx
kacper-mikolajczak: I was trying to understand why we have Suspense with an empty fallback and `<Document loading={<FullScreenLoadingIndicator />}` at the same time - it just seemed counterintuitive to me :D

src/hooks/usePreloadLazyModules.ts
JKobrynski: I gave it a try before and retried it now, it doesn't load the dynamic modules at all, it probably just loads them with the main bundle for some reason. I replaced all `webpackPreload: true` options with `webpackMode: eager` and this is the result:

https://github.com/user-attachments/assets/eccd4f64-f116-45e8-b9e3-d5c2e8e3dd9b

src/components/PDFThumbnail/index.tsx
JKobrynski: Done!

src/components/PDFThumbnail/index.tsx
JKobrynski: Updated

src/components/PDFThumbnail/pdfSetup.ts
JKobrynski: Updated too

src/components/PDFThumbnail/pdfSetup.ts
kacper-mikolajczak: Thanks for thorough response!

src/hooks/usePreloadLazyModules.ts
kacper-mikolajczak: Hmm that's interesting - we've got excatly the same setup for HEIC converter (separate chunk + eager mode) and it is working as intended. With your initial implementation, it was loaded separately on startup, right? Could you share a recording of network tab from working example to compare? Thanks! 

This wouldn't be a big deal if not writing explicit logic for fetching a module in react root component in current implementation. I think we should not be needed to do [that](https://github.com/callstack-internal/Expensify-App/blob/6a64e2e9cee04447dc1c91a1ba32790089373772/src/Expensify.tsx#L125).

Ideally, we'd just replace regular import with lazy ones for PDF modules and use webpack configuration (+ directives) to change runtime behaviour. Not use any react constructs to orchestrate that.

All in all, I'd say you may ask Kuba for help on this if we were not able to resolve it today.

kacper-mikolajczak:
Thanks @JKobrynski! Let's try to address [this](https://github.com/Expensify/App/pull/74566#discussion_r2522664319) - hit me up if you need any help ❤️

config/webpack/webpack.common.ts
dariusz-biela, line 375: From what I've noticed, `/* webpackPrefetch: true */ ` works, but adding the library to this config breaks it because you're forcing custom chunks.

Therefore, you need to remove the dynamic libraries from `splitChunks.cacheGroups`.


Remove your config and add this to vender chunk:
```js
 vendor: {
                        test: /[\\/]node_modules[\\/](?!(react-fast-pdf|react-pdf)[\\/])/,
                        name: 'vendors',

                        // Capture only the scripts needed for the initial load, so any async imports
                        // would be grouped (and lazy loaded) separately
                        chunks: 'initial',
                    },
```

src/hooks/usePreloadLazyModules.ts
```diff
@@ -0,0 +1,8 @@
+// This hook makes it possible to manually load the lazy loaded modules
+// right after the main bundle is loaded to decrease the bundle size
+function usePreloadLazyModules() {
+    import(/* webpackPrefetch: true */ 'react-fast-pdf');
+    import(/* webpackPrefetch: true */ 'react-pdf');
+}
```
dariusz-biela: You shouldn't need this after changing it in `config/webpack/webpack.common.ts`

dariusz-biela: Suggestion for consideration:
You can do some additional optimization and trigger additional JS loading when the browser is idle.

https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback

```js
requestIdleCallback(callback)
requestIdleCallback(callback, options)
```

src/components/PDFThumbnail/index.tsx
```diff
@@ -1,59 +1,88 @@
...
+                    onLoad={handleOnLoad}
+                    onLoadSuccess={handleOnLoadSuccess}
+                    onLoadError={handleOnLoadError}
+                    error={handleError}
+                >
+                    <View pointerEvents="none">
+                        <Suspense fallback={null}>
+                            <Thumbnail pageIndex={0} />
+                        </Suspense>
+                    </View>
```
dariusz-biela, line 71: I wonder if you need that extra <Suspense> here.
```suggestion
                    <View pointerEvents="none">
                            <Thumbnail pageIndex={0} />
                    </View>
```

dariusz-biela (commented):
Overall it looks good, I leave 2 comments for you to think about.

src/components/PDFThumbnail/index.tsx
JKobrynski: we don't, thanks for catching that

Krishna2323:
@JKobrynski is this ready for review?

src/hooks/usePreloadLazyModules.ts
Krishna2323: @JKobrynski Quick question about the preload hook. I commented out `usePreloadLazyModules` and still see `react-pdf/react-fast-pd`f being fetched right after the first screen via prefetch (without rendering any PDF UI). Given that, is the hook still needed? I might be missing something here, so I’m a bit confused about the requirement for the hook if prefetch already kicks in. Could you clarify the intended behavior?

I followed [this approach](https://github.com/Expensify/App/pull/74566#discussion_r2539116362) and it worked fine, is there any issue with that?

I did:
- Removed the `reactPdf` and `reactFastPdf` cacheGroups and changed `vendor` to exclude them:
  - `test: /[\\/]node_modules[\\/](?!(react-fast-pdf|react-pdf)[\\/])/`
  - `name: 'vendors'`, `chunks: 'initial'`

 - Deleted `usePreloadLazyModules` and any `requestIdle` usage/imports.
   - Removed the `requestIdle` import from `src/components/PDF/setup/index.ts` and the hook export.  
  - Added `import './components/PDF'` in `Expensify.tsx`

JKobrynski: Can you try the following steps:

1. Reload the app (just make sure you're not on a page that involves pdfs)
2. Change Network to Offline
3. Navigate to any chat 
4. Try sending a PDF attachment by drag&drop

I just tested that with the code you suggested and it didn't work for me.

CC: @Krishna2323

heyjennahay:
No product change, removing myself and unsubscribing

src/hooks/usePreloadLazyModules.ts
Krishna2323: @JKobrynski it works for me:

https://github.com/user-attachments/assets/d1447b7a-d256-4b3a-820b-0f73c817a8a9

JKobrynski: @Krishna2323 correct me if I'm wrong, but I think you haven't gone offline after refreshing 🤔

Krishna2323: Sorry for that, it doesn't work in offline mode. Will continue the review tomorrow.

src/components/PDFThumbnail/index.tsx
```diff
@@ -1,59 +1,84 @@
...
+            .catch(() => {
+                setPdfInitializationFailed(true);
+                onPDFLoadError?.();
+            });
+        // We only want to call this method once
+        // eslint-disable-next-line react-compiler/react-compiler
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    const loadingIndicator = useMemo(() => <ActivityIndicator size="large" />, []);
```
Krishna2323: Use `CONST.ACTIVITY_INDICATOR_SIZE.LARGE`.

src/components/PDFThumbnail/types.ts
```diff
@@ -24,6 +24,9 @@ type PDFThumbnailProps = {
 
     /** Callback to call if PDF is loaded */
     onLoadSuccess?: () => void;
+
+    /** Callback to call if the PDF library fails to load dynamically */
+    onPDFLoadError?: () => void;
```
Krishna2323: I think we should rename it to `onPDFLibraryLoadError` or `onPdfInitFailed`.

Krishna2323 (changes requested):

src/components/PDFThumbnail/index.tsx
JKobrynski: Done

src/components/PDFThumbnail/types.ts
JKobrynski: Great idea, done

JKobrynski:
I managed to reproduce this crash and fix it locally. It should work this time. Can we rebuild the test builds? 🙏 @mountiny

mountiny:
running the builds

Krishna2323:
@JKobrynski the offline test works bug I get a warning (`Warning: Setting up fake worker.`) and it seems we might be doing something wrong.

The explanation I got from the AI:

> A “fake worker” warning means PDF.js fell back to running the worker code on the main thread because it didn’t get a valid worker file. That fallback can still render some PDFs, so basic tests might appear to pass. But it’s fragile:
> 
> - Performance/regressions: large/complex PDFs can hang the UI because all worker work runs on the main thread.
> - Offline reliability: without the real worker chunk, caching is inconsistent; you may hit failures depending on cache state or if the chunk was never fetched while online.
> - Platform variance: different browsers/desktops may behave differently; some will refuse the fake worker path or hit CSP issues.
> 
> The fix is to point `workerSrc` at the actual worker file URL (e.g., `import('pdfjs-dist/build/pdf.worker.min.mjs?url')`) and prefetch it. That removes the fake worker warning and guarantees a real worker is cached for offline. Even if your manual run passed, the warning shows the app isn’t using the intended worker path, so we should correct it to avoid intermittent failures.



https://github.com/user-attachments/assets/eb2b4f6c-509d-4a06-9b32-f58f38b3afb9

mountiny:
@codex review

Issue #75387, opened 2025-11-18
[RCA #72674] Distance - Amount field is blank, Rate field shows pending on confirm page


mitarachim:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.2.60-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** Exp https://github.com/Expensify/App/pull/74791
**Email or phone of affected tester (no customers):** sjdsjnodijsisjio@gmail.com
**Issue reported by:** Applause Internal Team
**Device used:** Mac 15.5 / Chrome
**App Component:** Money Requests

## Action Performed:
1. Go to staging.new.expensify.com
2. Go to workspace chat.
3. Click + > Track distance > Manual.
4. Enter amount > Next.

## Expected Result:
Amount and Rate field will show the correct info.

## Actual Result:
Amount field is blank, Rate field shows pending on confirm page.

## Workaround:
Unknown

## Platforms:
- [ ] Android: App
- [x] Android: mWeb Chrome
- [ ] iOS: App
- [x] iOS: mWeb Safari
- [x] iOS: mWeb Chrome
- [ ] Windows: Chrome
- [x] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

https://github.com/user-attachments/assets/11c32835-d3a7-4131-8072-4781f7048790

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

PR #75432, opened 2025-11-18
Revert "Fix: Default distance rate is not selected when creating distance expense"

grgia:
Reverts Expensify/App#74791

$ https://github.com/Expensify/App/issues/75387


francoisl:
So we don't need this revert, right? Just confirming if we'll need a follow-up PR.

Issue #75222, opened 2025-11-15
[Due for payment 2026-01-15] [DEW in NewDot] Update `SubmitReport` API command


francoisl:
- Part of project `Add DEW support to NewDot`
- Parent issue (internal): https://github.com/Expensify/Expensify/issues/535524
- [Design doc (internal)](https://docs.google.com/document/d/144a8NLphlLwOlmLGQ9DQf6oi2HacGZdIZlnM4-O802k/edit)

---

### Task

**[Relevant design doc section](https://docs.google.com/document/d/144a8NLphlLwOlmLGQ9DQf6oi2HacGZdIZlnM4-O802k/edit?tab=t.0#heading=h.u5xv19jzmy0m)** with all details

Summary:
- Skip optimistic next step update on `SubmitReport`
- Implement temporary (Onyx-only) offline action 
- Update RBR logic in `getAllReportActionsErrorsAndReportActionThatRequiresAttention`
- Display a static error message below the report title

bernhardoj:
Yes

Christinadobrzyn:
I'll be ooo until 12/21. We are not assigning anyone new while we review the PR.

Christinadobrzyn:
following https://github.com/Expensify/App/pull/75712

Issue #75223, opened 2025-11-15
[Due for payment 2025-12-31] [DEW in NewDot] Display system message after submit/approval action


francoisl:
- Part of project `Add DEW support to NewDot`
- Parent issue (internal): https://github.com/Expensify/Expensify/issues/535524
- [Design doc (internal)](https://docs.google.com/document/d/144a8NLphlLwOlmLGQ9DQf6oi2HacGZdIZlnM4-O802k/edit)

---

### Task

**[Relevant design doc section](https://docs.google.com/document/d/144a8NLphlLwOlmLGQ9DQf6oi2HacGZdIZlnM4-O802k/edit?tab=t.0#heading=h.cxprxdew8b7m)** with all details

Summary:
When a report has been submitted or approved with DEW, show a secondary comment coming from Concierge in the report comments

<details><summary>Issue Owner</summary>Current Issue Owner: @trjExpensify</details>

trjExpensify:
Cool, closing!

PR #72663, opened 2025-10-15
Block report submission when strict policy rules are enabled with violations

abzokhattab:
### Fixed Issues

$ https://github.com/Expensify/App/issues/70990
PROPOSAL: https://github.com/Expensify/App/issues/70990#issuecomment-3388033467

### Tests

Prerequisites:
- Enable "Strictly Enforce Workspace Rules" in Domain Settings
- Workspace must have at least one policy rule 

1. Log in as a submitter in the domain with strict policy rules enabled
2. Submit a report that contains policy violations in NewDot
3. Go to Reports and open the created report
4. verify that the Submit button is disabled and next step message shows: "'Waiting for you to fix the issues. Your admins have restricted submission of expenses with violations."
5. Go back - verify Review button shows in report preview

### Offline tests

Same as tests 
### QA Steps

same as tests 
### PR Author Checklist

- [X] I linked the correct issue in the `### Fixed Issues` section above
- [X] I wrote clear testing steps that cover the changes made in this PR
    - [X] I added steps for local testing in the `Tests` section
    - [X] I added steps for the expected offline behavior in the `Offline steps` section
    - [X] I added steps for Staging and/or Production testing in the `QA steps` section
    - [X] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [X] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [X] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [X] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [X] I ran the tests on **all platforms** & verified they passed on:
    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
    - [X] MacOS: Desktop
- [X] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [X] I verified there are no new alerts related to the `canBeMissing` param for `useOnyx`
[...]

PR #75712, opened 2025-11-20
[DEW]: Update `SubmitReport` API command

abzokhattab:
### Explanation of Change
Add support for Dynamic External Workflow (DEW) in the report submission flow.

Key Changes:

1. No optimistic updates for DEW policies - When submitting on a DEW policy, the app doesn't optimistically update stateNum, statusNum, managerID, or nextStep since the backend determines the actual workflow
2. New pending state tracking - Uses reportMetadata.pendingExpenseAction to track DEW submissions in progress
3. New report action type - DEW_SUBMIT_FAILED for when the backend rejects a DEW submission
4. UI feedback:
    - Shows "queued to submit via custom approval workflow" in LHN when DEW submit is pending
    - Shows error next step when DEW submission fails
    - Shows offline-specific next step message when submitting while offline
5. Beta gated - Behind CONST.BETAS.NEW_DOT_DEW flag

### Fixed Issues

$ https://github.com/Expensify/App/issues/75222
PROPOSAL:

### Tests
Prerequisites:
- Have a policy with DEW enabled (ask an internal employee)
- Add at least 1 additional member to the policy (so, at least 2 including yourself)

**Test 1: Submit Success - Online**
1. Create an expense on a DEW-enabled workspace
2. Change the **description** of the expense to the email address of another policy member
3. Submit the report (while online)
4. Verify:
- "submitted" message appears
- Next step updates to show approver

**Test 2: Submit Success - Offline → Online**
1. Create an expense on a DEW-enabled workspace
2. Change the **description** of the expense to the email address of another policy member
3. Go offline (CMD+D → Force offline)
4. Submit the report
5. Verify:
- Shows "queued to submit via custom approval workflow" (greyed out)
- Next step unchanged
6. Go online (disable Force offline)
7. Verify online:
- "queued to submit" replaced with "submitted"
- Next step updates to show approver

**Test 3: Submit Failed**
1. Create an expense on a DEW-enabled workspace
2. Rename report to "ERROR"
3. Submit the report
5. Verify:
- "failed to submit the report because..." in comments
RBR in LHN
- Next step: "This report can't be submitted. Please review the comments to resolve"
- Validate that the submit button is still shown and user is able to resubmit .
- Also validate that the report preview displays `submit`

### Offline tests

1. Create an expense on a DEW-enabled workspace
2. Change the **description** of the expense to the email address of another policy member
6. Go offline (CMD+D → Force offline)
7. Submit the report
8. Verify:
- Shows "queued to submit via custom approval workflow" (greyed out)
- Next step changed to `Waiting for you to come back online to determine next steps`
- The submit button is removed from the expense. 
- The report preview displays `view` not `submit`

### QA Steps

Same as tests 

### PR Author Checklist

    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/eef63c6b-d9d0-4d7e-8508-9be963aa927a

</details>

<details>

https://github.com/user-attachments/assets/5fd52728-6587-4f3d-8b13-f8a6aba50da7

</details>

<details>

https://github.com/user-attachments/assets/38f45b23-ed1f-4a4d-91fa-bf6366c79e7a

</details>

<details>

https://github.com/user-attachments/assets/d90d6f45-7007-4f8d-ac81-48a1fd7f24f3

</details>

<details>

https://github.com/user-attachments/assets/3713f6d0-8b06-4219-bd7f-b304abf30b53

https://github.com/user-attachments/assets/87562a89-8daf-4d05-8364-2a14e23c2aa5

https://github.com/user-attachments/assets/9700b4d8-f5a1-4e5d-9174-63a0ae18076d

https://github.com/user-attachments/assets/e1322563-b9ca-4abb-a600-b5631dc6f3d3

</details>


abzokhattab:
translation confirmation was requested here: https://expensify.slack.com/archives/C01GTK53T8Q/p1763725986769179

src/CONST/index.ts
```diff
@@ -1201,6 +1202,8 @@ const CONST = {
                 CREATED: 'CREATED',
                 DELETED_ACCOUNT: 'DELETEDACCOUNT', // Deprecated OldDot Action
                 DELETED_TRANSACTION: 'DELETEDTRANSACTION',
+                DEW_APPROVE_FAILED: 'DEW_APPROVE_FAILED',
```
bernhardoj: Do we handle the approve API in this PR, too?

src/libs/actions/IOU.ts
```diff
@@ -11121,17 +11139,16 @@ function submitReport(
...
-                    nextStep: null,
-                },
+                ...(isDEWPolicy ? {} : {nextStep: expenseReport.nextStep ?? null}),
+                ...(isDEWPolicy
+                    ? {}
+                    : {
+                          pendingFields: {
+                              nextStep: null,
+                          },
+                      }),
```
bernhardoj, line 10816: Can we simplify this to:

```
...(isDEWPolicy ? {} : {
        nextStep: expenseReport.nextStep ?? null,
        pendingFields: {
            nextStep: null,
        },
    }),                     
```

abzokhattab:
Translation is confirmed ... waiting for the backend fix for the submit to proceed with the complete tests

abzokhattab:
@trjExpensify  Oh sorry i shared the details with @francoisl on slack but forwarding it here as here: 


- I tested the online and offline flow in case there are no errors and its now working as expected .
- There is still an issue when we have a submit error ... according to the doc the backend should return something like this in case of submit error

```json
{
    "onyxData": [{
        "onyxMethod": "merge",
        "key": "reportActions_8495312605922320",
        "value": {
            "5681436454011784330": {
                "actionName": "DEW_SUBMIT_FAILED",
                "actorAccountID": 123,
                "created": "2025-09-04 00:12:34",
                "originalMessage": {
                    "message": "This report cannot be submitted or approved",
                    "automaticAction": false
                },
                "reportActionID": "5681436454011784330",
                "shouldShow": true
            }
        }
    }]
}
```

however currently it returns
```json
{
    "code": 666,
    "jsonCode": 666,
    "type": "Expensify\\Libs\\Error\\ExpError",
    "UUID": "e9b387bc-654b-4576-988a-69bbca06abf9",
    "message": "Report cannot be submitted: This report cannot be submitted or approved",
    "title": "",
    "data": {
        "jsonCode": 406,
        "onyxData": [
            {
                "onyxMethod": "merge",
                "key": "report_8495312605922320",
                "value": {
                    "errors": {
                        "1764214172445098": "Report cannot be submitted: This report cannot be submitted or approved"
                    }
                }
            }
        ]
    },
    "htmlMessage": "This report cannot be submitted or approved",
    "onyxData": [
        {
            "onyxMethod": "merge",
            "key": "report_8495312605922320",
            "value": {
                "errors": {
                    "1764214172445098": "Report cannot be submitted: This report cannot be submitted or approved"
                }
            }
        }
    ],
    "requestID": "9a4e8aabbf097c81-LAX"
}
```

abzokhattab:
so basically the `"actionName": "DEW_SUBMIT_FAILED"` isnt returned

trjExpensify:
Not sure how much help I can be with Franc out, but looking at [the logs](https://www.expensify.com/_devportal/tools/logSearch/#sort=asc&size=10000&query=request_id%3A%229a4e8aabbf097c81-LAX%22%20timestamp%3A%5B2025-11-21T00%3A00%20TO%202025-11-29T23%3A59%5D&engine=victorialogs) (internal link) for that requestID. 

```
Bedrock\Client - Starting a request ~~ command: 'CreateReportAction' clusterName: 'auth' headers: '[authToken: '<REDACTED>' reportID: '8495312605922320' reportActionID: '' action: 'DEWSUBMITFAILED' message: '[message: 'This report cannot be submitted or approved']' toAccountID: '' status: '' clientCreatedTime: '' idempotent: '' logParam: 'abzokhattab+874@gmail.com' urlToNewDot: 'https://new.expensify.com/' shouldReadUsingThreads: '' shouldHandleOnyxUpdates: '1' isNewDotRequest: '1' clientUpdateID: '8896697089' maxNumberOfUpdates: '500' commitCount: '32637129177' requestID: '9a4e8aabbf097c81-LAX' lastIP: '216.73.160.133' writeConsistency: 'ASYNC' priority: '500' timeout: '110000']'
```
```
API threw ExpError Expensify\Libs\Error\ExpError - e9b387bc-654b-4576-988a-69bbca06abf9 ~~ jsonCode: '406' onyxData: '[0: '[onyxMethod: 'merge' key: 'report_8495312605922320' value: '[errors: '[1764214172445098: 'Report cannot be submitted: This report cannot be submitted or approved']']']']' exceptionJsonCode: '666' exceptionUserMessage: 'Report cannot be submitted: This report cannot be submitted or approved' exceptionMessage: 'Report cannot be submitted: This report cannot be submitted or approved' exceptionFile: '/git/releases/expensify.com/dbffa78/lib/Report/Submit/DynamicExternalWorkflow.php' exceptionLine: '64' exceptionCode: '666'
```

Is it a nomenclature problem? The action is seemingly called `DEWSUBMITFAILED` not `DEW_SUBMIT_FAILED`

abzokhattab:
The PR is now ready, please review @bernhardoj @francoisl @heyjennahay

bernhardoj:
@abzokhattab based on the test step, we will have the `report routed to A due to custom approval workflow` message.

image

But currently, `submitted, saying All reports must go through afewfpm` is shown.

Also, after I do Clear cache and restart, the `submitted` message is gone.

image

bernhardoj:
`failed to submit the report because` action and `This report can't be submitted. Please review the comments to resolve` next step are not shown.

image

I guess this is the issue that still need BE fix, right?

src/CONST/index.ts
```diff
@@ -7277,6 +7279,7 @@ const CONST = {
         HAS_CHILD_REPORT_AWAITING_ACTION: 'hasChildReportAwaitingAction',
         HAS_MISSING_INVOICE_BANK_ACCOUNT: 'hasMissingInvoiceBankAccount',
         HAS_UNRESOLVED_CARD_FRAUD_ALERT: 'hasUnresolvedCardFraudAlert',
+        HAS_DEW_ERROR: 'hasDEWError',
```
bernhardoj: I don't see the usage of this. Will we use this later?

src/languages/de.ts
```diff
@@ -7489,6 +7492,7 @@ ${
             hasChildReportAwaitingAction: 'Hat einen untergeordneten Bericht, der auf eine Aktion wartet',
             hasMissingInvoiceBankAccount: 'Fehlendes Rechnungsbankkonto',
             hasUnresolvedCardFraudAlert: 'Hat eine ungelöste Karten-Fraud-Warnung',
+            hasDEWError: 'Hat DEW-Fehler',
```
bernhardoj: I don't see the usage of this, either. Will we use this later?

bernhardoj (changes requested):

src/libs/ReportUtils.ts
```diff
@@ -9299,6 +9301,39 @@ function getAllReportActionsErrorsAndReportActionThatRequiresAttention(
...
+            const mostRecentSubmittedAction = reportActionsArray
+                .filter((action) => isSubmittedAction(action))
+                .reduce<ReportAction | undefined>((latest, current) => {
+                    if (!latest || (current.created && latest.created && current.created > latest.created)) {
+                        return current;
+                    }
+                    return latest;
+                }, undefined);
+
+            const shouldShowDEWError = !mostRecentSubmittedAction || (mostRecentSubmittedAction.created && mostRecentDewSubmitFailedAction.created > mostRecentSubmittedAction.created);
```
bernhardoj: `(mostRecentSubmittedAction.created && mostRecentDewSubmitFailedAction.created > mostRecentSubmittedAction.created);`

We don't need to check `mostRecentSubmittedAction.created` anymore.

`mostRecentDewSubmitFailedAction.created > mostRecentSubmittedAction.created;`

src/libs/ReportUtils.ts
```diff
@@ -9299,6 +9301,39 @@ function getAllReportActionsErrorsAndReportActionThatRequiresAttention(
...
+
+            if (shouldShowDEWError) {
+                reportActionErrors.dewSubmitFailed = getMicroSecondOnyxErrorWithTranslationKey('iou.error.genericDEWSubmitFailureMessage');
+                if (!reportAction) {
+                    reportAction = mostRecentDewSubmitFailedAction;
```
bernhardoj: Should we always replace the `reportAction` with `mostRecentDewSubmitFailedAction` because it's the action that is associated with the error?

src/components/MoneyReportHeader.tsx
```diff
@@ -566,6 +569,29 @@ function MoneyReportHeader({
...
+                            src={expensifyIcons.DotIndicator}
+                            height={variables.iconSizeSmall}
+                            width={variables.iconSizeSmall}
+                            fill={theme.danger}
+                        />
+                    ),
+                    description: <Text style={{color: theme.danger}}>{translate('iou.dynamicExternalWorkflowCannotSubmit')}</Text>,
+                };
+            }
+        }
```
bernhardoj: Related to the 2nd image [here](https://github.com/Expensify/App/pull/75712#issuecomment-3605012909), the `This report can't be submitted. Please review the comments to resolve.` is shown below the next step. We need to show this error message as the next step, instead of as a status bar.

image

src/libs/ReportPreviewActionUtils.ts
```diff
@@ -264,6 +265,11 @@ function getReportPreviewAction(
         return CONST.REPORT.REPORT_PREVIEW_ACTIONS.ADD_EXPENSE;
     }
 
+    // If DEW submit failed and report is still open, show REVIEW
+    if (hasDEWSubmitFailed && isOpenReport(report)) {
```
bernhardoj: We check if there is a recent SUBMITTED action in [here](https://github.com/Expensify/App/pull/75712/files#diff-577fcc5f3a5e4930916ff310e61c38edfa093792fe3ab55b9a30d2aecd8811dbR9326), but we don't do that here. Won't this cause inconsistency? 

Can't test it though because I'm not sure how to do this

> This is to account for the case where we tried to auto-submit the report while the user was not online, and the report is still in the OPEN status, and waiting for the user to take manual action.

src/libs/SearchUIUtils.ts
```diff
@@ -1263,6 +1264,12 @@ function getActions(
         return [CONST.SEARCH.ACTION_TYPES.REVIEW];
     }
 
+    // Check for DEW submit failed - if the report has a DEW_SUBMIT_FAILED action and is still OPEN, show Review
+    const hasDEWSubmitFailed = report.statusNum === CONST.REPORT.STATUS_NUM.OPEN && reportActions.some(isDynamicExternalWorkflowSubmitFailedAction);
```
bernhardoj: Same as [above](https://github.com/Expensify/App/pull/75712/files#r2583647107)

src/libs/ReportUtils.ts
```diff
@@ -12649,7 +12684,9 @@ function selectFilteredReportActions(
     return Object.fromEntries(
         Object.entries(reportActions).map(([reportId, actionsGroup]) => {
             const actions = Object.values(actionsGroup ?? {});
-            const filteredActions = actions.filter((action): action is ReportAction => isExportIntegrationAction(action) || isIntegrationMessageAction(action));
+            const filteredActions = actions.filter(
+                (action): action is ReportAction => isExportIntegrationAction(action) || isIntegrationMessageAction(action) || isDynamicExternalWorkflowSubmitFailedAction(action),
```
bernhardoj: What is this for?

abzokhattab: its mandatory for preventing the submit action from showing up in the search page .. this selector is used to filter data in onyx then we use the value to display the view action

src/libs/ReportPreviewActionUtils.ts
abzokhattab: good point .. created a shared function in the reportActionsUtil to be used in all cases

src/libs/SearchUIUtils.ts
abzokhattab: covered in https://github.com/Expensify/App/pull/75712#discussion_r2586366608

src/components/MoneyReportHeader.tsx
abzokhattab: yeah i overlooked this one ... changed it

francoisl:
> Currently, `queued to submit via custom approval workflow` is also shown when submitting while online, but the test step only mentioned this on an offline case. I agree though to show `queued to submit via custom approval workflow` online or offline.

It's only meant to be shown when you're offline, to indicate that the action is "queued" but hasn't happened yet. I think we can use `isOffline` from the `useNetwork()` around [here](https://github.com/Expensify/App/blob/10513f5a63a4ea9479a3a25/src/pages/home/report/PureReportActionItem.tsx#L1191) to only show it when offline, right?

abzokhattab:
> @abzokhattab based on the test step, we will have the report routed to A due to custom approval workflow message.


this is not a part of this ticket ... conceirge messages should be handled in https://github.com/Expensify/App/issues/75223

> Also, after I do Clear cache and restart, the "submitted" message is gone.

because the backend doesnt send it in the respond


> Looks like we optimistically set the managerID to the current user. I think we shouldn't optimistically update the report status to SUBMITTED, either


yeah i think we need to hide the submit button ... i changed that in the latest commits .. now the primary button is hidden after submit and also in the preview it shows VIEW

> Submitting offline doesn't show the Waiting for you to come back online to determine next steps next steps.

this is per design ... as per the submit offline description, it mentioned not to change the next steps
Screenshot 2025-12-03 at 09 43 35


should we change it i am confused

src/libs/ReportUtils.ts
abzokhattab: yes changed it to be consistent

src/components/MoneyReportHeader.tsx
```diff
@@ -227,6 +232,7 @@ function MoneyReportHeader({
         'Export',
         'Document',
         'Feed',
+        'DotIndicator',
```
bernhardoj: We don't use this anymore, let's remove it.

src/components/MoneyReportHeaderStatusBar.tsx
```diff
@@ -43,7 +44,7 @@ function MoneyReportHeaderStatusBar({nextStep}: MoneyReportHeaderStatusBarProps)
                     src={(nextStep?.icon && iconMap?.[nextStep.icon]) ?? Expensicons.Hourglass}
                     height={variables.iconSizeSmall}
                     width={variables.iconSizeSmall}
-                    fill={theme.icon}
+                    fill={nextStep?.type === 'alert' ? theme.danger : theme.icon}
```
bernhardoj: This will affect other next step with the type of `alert`.

https://github.com/Expensify/App/blob/d1b6f6d1d674b3048103d7a78b9d7c29c0d0a5af/src/libs/NextStepUtils.ts#L314-L316
https://github.com/Expensify/App/blob/d1b6f6d1d674b3048103d7a78b9d7c29c0d0a5af/src/libs/NextStepUtils.ts#L286-L288

src/libs/actions/IOU.ts
```diff
@@ -11440,37 +11446,53 @@ function submitReport(
...
+                      // because DEW determines the actual workflow on the backend
+                      ...(isDEWPolicy
+                          ? {
+                                lastMessageText: getReportActionText(optimisticSubmittedReportAction),
+                                lastMessageHtml: getReportActionHtml(optimisticSubmittedReportAction),
+                            }
+                          : {
+                                managerID,
+                                lastMessageText: getReportActionText(optimisticSubmittedReportAction),
+                                lastMessageHtml: getReportActionHtml(optimisticSubmittedReportAction),
```
bernhardoj: We can always set the `lastMessageText` and `lastMessageHtml` instead of duplicating it.

image

src/libs/ReportPreviewActionUtils.ts
```diff
@@ -159,6 +160,8 @@ function getReportPreviewAction(
     isPaidAnimationRunning?: boolean,
     isApprovedAnimationRunning?: boolean,
     isSubmittingAnimationRunning?: boolean,
+    hasDEWSubmitFailed?: boolean,
+    hasPendingDEWSubmit?: boolean,
```
bernhardoj: I think we can combine this into one param, `hasDEWSubmitPendingOrFailed`, so we can also remove the lint suppress.

bernhardoj: I think it's even better if we have a `hasDEWSubmitPendingOrFailed` util function since I see we have a lot of logic to check if it has a DEW submit or pending action.

```
getReportPreviewAction(..., hasDEWSubmitPendingOrFailed(reportActions, policy))
```

src/libs/ReportUtils.ts
```diff
@@ -12654,7 +12665,13 @@ function selectFilteredReportActions(
     return Object.fromEntries(
         Object.entries(reportActions).map(([reportId, actionsGroup]) => {
             const actions = Object.values(actionsGroup ?? {});
-            const filteredActions = actions.filter((action): action is ReportAction => isExportIntegrationAction(action) || isIntegrationMessageAction(action));
+            const filteredActions = actions.filter(
+                (action): action is ReportAction =>
+                    isExportIntegrationAction(action) ||
+                    isIntegrationMessageAction(action) ||
+                    isDynamicExternalWorkflowSubmitFailedAction(action) ||
+                    (isSubmittedAction(action) && action.pendingAction === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD),
```
bernhardoj: I think we should always include the SUBMITTED action, otherwise the `getMostRecentActiveDEWSubmitFailedAction` [here](https://github.com/Expensify/App/pull/75712/files#diff-461eed159e68d2c42994984384060d2660cb487187850ceae59c845a2e3c0e62R1247) will always be true.

and so the View button is shown instead of SUBMIT.
image

tests/actions/IOUTest.ts
```diff
@@ -5580,6 +5580,152 @@ describe('actions/IOU', () => {
                         }),
                 );
         });
+
+        it('correctly submits a report with Dynamic External Workflow policy without setting nextStep', () => {
```
bernhardoj: I think we should update the name since we are also preventing setting the stateNum and statusNum optimistically.

tests/actions/IOUTest.ts
```diff
@@ -5580,6 +5580,152 @@ describe('actions/IOU', () => {
...
+            const amount = 10000;
+            const comment = '💸💸💸💸';
+            const merchant = 'NASDAQ';
+            let expenseReport: OnyxEntry<Report>;
+            let chatReport: OnyxEntry<Report>;
+            let policy: OnyxEntry<Policy>;
+            let policyID: string;
+
+            return waitForBatchedUpdates()
```
bernhardoj, line 6307: We don't need `waitForBatchedUpdates` here.

tests/actions/IOUTest.ts
```diff
@@ -5580,6 +5580,152 @@ describe('actions/IOU', () => {
...
+                            const connection = Onyx.connect({
+                                key: ONYXKEYS.COLLECTION.REPORT,
+                                waitForCollectionCallback: true,
+                                callback: (allReports) => {
+                                    Onyx.disconnect(connection);
+                                    expenseReport = Object.values(allReports ?? {}).find((report) => report?.type === CONST.REPORT.TYPE.EXPENSE && report?.policyID === policyID);
+
+                                    expect(expenseReport?.stateNum).toBe(CONST.REPORT.STATE_NUM.OPEN);
+                                    expect(expenseReport?.statusNum).toBe(CONST.REPORT.STATUS_NUM.OPEN);
+                                    expect(expenseReport?.nextStep).toBeUndefined();
```
bernhardoj: The test passes, but I'm just curious why this is undefined. Isn't the next step is set optimistically when calling `requestMoney`.

tests/actions/ReportPreviewActionUtilsTest.ts
```diff
@@ -489,4 +489,122 @@ describe('getReportPreviewAction', () => {
...
+            const policy = createRandomPolicy(0);
+            policy.type = CONST.POLICY.TYPE.CORPORATE;
+
+            await Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${REPORT_ID}`, report);
+            const transaction = {
+                reportID: `${REPORT_ID}`,
+            } as unknown as Transaction;
+
+            const {result: isReportArchived} = renderHook(() => useReportIsArchived(report?.parentReportID));
+            await waitForBatchedUpdatesWithAct();
```
bernhardoj: I don't think we need `waitForBatchedUpdatesWithAct`

tests/ui/ReportListItemHeaderTest.tsx
```diff
@@ -216,4 +224,48 @@ describe('ReportListItemHeader', () => {
...
+        });
+
+        it('should render without DEW props', async () => {
+            const reportItem = createReportListItem(CONST.REPORT.TYPE.EXPENSE, 'john', 'jane', {
+                action: 'submit',
+            });
+            renderReportListItemHeader(reportItem);
+            await waitForBatchedUpdatesWithAct();
+
+            expect(screen.getByText('John Doe')).toBeOnTheScreen();
```
bernhardoj: I don't think we need these tests?

tests/unit/ReportActionsUtilsTest.ts
```diff
@@ -1476,4 +1476,321 @@ describe('ReportActionsUtils', () => {
...
+
+    describe('getMostRecentActiveDEWSubmitFailedAction', () => {
+        it('should return the DEW action when DEW_SUBMIT_FAILED exists and no SUBMITTED action exists', () => {
+            const actionId1 = '1';
+            const reportActions: ReportActions = {
+                [actionId1]: {
+                    ...createRandomReportAction(0),
+                    actionName: CONST.REPORT.ACTIONS.TYPE.DEW_SUBMIT_FAILED,
+                    created: '2025-11-21 10:00:00',
+                    reportActionID: '1',
```
bernhardoj: We can use the `actionId1` here. (do it for other tests too)

tests/unit/ReportUtilsTest.ts
```diff
@@ -1810,7 +1810,7 @@ describe('ReportUtils', () => {
                         merchant: 'Test Merchant',
                         created: testDate,
                         modifiedMerchant: 'Test Merchant',
-                    } as Transaction;
+                    };
```
bernhardoj, line 1830: Is this unintentional? I'm okay to remove this if it's not needed.

tests/unit/ReportUtilsTest.ts
```diff
@@ -10113,4 +10113,252 @@ describe('ReportUtils', () => {
...
+                        text: 'DEW submit failed',
+                    },
+                ],
+                originalMessage: {
+                    message: 'This report contains an Airfare expense that is missing the Flight Destination tag.',
+                },
+            };
+
+            const reportActions = {
+                '1': dewSubmitFailedAction,
```
bernhardoj: Let's use the ID from `dewSubmitFailedAction`.

```
[dewSubmitFailedAction.reportActionID]: dewSubmitFailedAction,
```

Do it for other tests too.

tests/unit/ReportUtilsTest.ts
```diff
@@ -10113,4 +10113,252 @@ describe('ReportUtils', () => {
...
+            };
+
+            const {errors} = getAllReportActionsErrorsAndReportActionThatRequiresAttention(report, reportActions, true);
+
+            expect(errors?.dewSubmitFailed).toBeUndefined();
+        });
+
+        it('should clear DEW error when a more recent SUBMITTED action exists after the failure (multiple submits)', () => {
```
bernhardoj: I think we should say `should not return` instead of `clear`, right? Like the above test case.

bernhardoj:
I just remember, let's format all the new tests to use the Given, When, Then comment format.

abzokhattab:
> let's format all the new tests to use the Given, When, Then comment format.



i think these are really minor issues that doesnt touch the user but i updated them for better readability.

tests/actions/IOUTest.ts
abzokhattab: its undefined because the initial step was undefined and it didnt change

src/libs/ReportUtils.ts
abzokhattab: done

src/libs/ReportPreviewActionUtils.ts
abzokhattab: yeah we can combine them in the same function .. done

src/libs/actions/IOU.ts
abzokhattab: done

src/components/MoneyReportHeaderStatusBar.tsx
abzokhattab: fixed

trjExpensify:
@abzokhattab conflicts here!

abzokhattab:
@trjExpensify  will resolve it 

 currently there is a missing piece in the backend causing the submit failed comments returned from the backend not to show up unless i have to navigate away and reopen the report 


Details shared on slack but here is the msg again  



When `SubmitReport` fails for DEW, the server returns a `DEWSUBMITFAILED` action in `reportActions_<reportID>`, but it often arrives with empty  `reportActions_<id>.<actionID>.message[0].text/html` whereas the `openReport` has  these data populated 

This mismatch is what causes the comments UI to not show the failure reason until a later openReport refresh. thats why the new comment doesnt show up until its reopened so that the message[0].text/html are fetched 


here is what i mean ... to the left is the openReport response and to the right is the submitRequest response ... as u can notice the submit request is diff and doesnt contain the same data

image




for reference here is the full data   [submitRequest.txt](https://github.com/user-attachments/files/24129414/submitRequest.txt) [openReport.txt](https://github.com/user-attachments/files/24129417/openReport.txt)

bernhardoj:
Will continue reviewing the code later...

src/libs/actions/IOU.ts
```diff
@@ -11612,28 +11633,36 @@ function submitReport(
...
-            key: `${ONYXKEYS.COLLECTION.NEXT_STEP}${expenseReport.reportID}`,
-            value: expenseReportCurrentNextStepDeprecated ?? null,
-        },
     ];
     failureData.push({
         onyxMethod: Onyx.METHOD.MERGE,
         key: `${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${expenseReport.reportID}`,
         value: {
             [optimisticSubmittedReportAction.reportActionID]: {
+                pendingAction: null,
```
bernhardoj: Why we add this?

abzokhattab: its needed to clear the pending action after the backend fails otherwise the the submit button will not show up again because the `hasPendingDEWSubmit` would return true even though the backend returned a response

bernhardoj: Hmm, but if we remove the pending action, the message won't be grayed out like this anymore

image

Is it possible that we add a new key to reportMetadata indicating that the report has a pending submit/approve? So we just check that reportMetadata key.

We can call it `pendingExpenseAction` with a value of either SUBMIT/APPROVE/undefined.

abzokhattab: yeah i think this would work

src/libs/SearchUIUtils.ts
```diff
@@ -1268,8 +1276,14 @@ function getActions(
...
     const chatReport = getChatReport(data, report);
-    const canBePaid = canIOUBePaid(report, chatReport, policy, allReportTransactions, false, chatReportRNVP, invoiceReceiverPolicy);
-    const shouldOnlyShowElsewhere = !canBePaid && canIOUBePaid(report, chatReport, policy, allReportTransactions, true, chatReportRNVP, invoiceReceiverPolicy);
+
+    // For DEW policies, don't show PAY if the report is not approved yet
+    // DEW reports need to go through external approval workflow before payment
+    const isDEWPolicy = hasDynamicExternalWorkflow(policy);
+    const shouldSkipPayForDEW = isDEWPolicy && !isReportApproved({report}) && !isClosedReport(report);
+
+    const canBePaid = shouldSkipPayForDEW ? false : canIOUBePaid(report, chatReport, policy, allReportTransactions, false, chatReportRNVP, invoiceReceiverPolicy);
```
bernhardoj: Why do we need to make this exception? Why canIOUBePaid is true when submitting the report?

src/libs/actions/IOU.ts
abzokhattab: i think a safer approach would to combine both to have something like this so that we also check the pending action  

```js
function hasPendingDEWSubmit(
    reportMetadata: OnyxEntry<ReportMetadata>,
    reportActions: OnyxEntry<ReportActions> | ReportAction[]
): boolean {
    if (reportMetadata?.pendingExpenseAction !== 'SUBMIT') {
        return false;
    }
    
    const actionsArray = Array.isArray(reportActions) ? reportActions : Object.values(reportActions ?? {});
    return actionsArray.some((action) => 
        isSubmittedAction(action) && action.pendingAction === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD
    );
}
```

abzokhattab: hmmm i am not sure about this approach i feel it introduces unneed duplications which might bring more complexity

bernhardoj: > i feel it introduces unneed duplications which might bring more complexity

Yeah. Can we check for `reportMetadata?.pendingExpenseAction` instead? What would be the possible issue by only using that?

abzokhattab:
back on into again

src/libs/SearchUIUtils.ts
abzokhattab: this is now fixed after fixing the `hasPendingDEWSubmit` so i am reverted this change.

src/libs/actions/IOU.ts
abzokhattab: i couldnt find a regression for this its just adds more complexity ... but it acheives the goal so i am fine with it unless we find a better approach ... 


I implemented this in the recent commits

src/libs/NextStepUtils.ts
```diff
@@ -341,6 +341,36 @@ function buildOptimisticNextStepForStrictPolicyRuleViolations() {
...
 }
 
+function buildOptimisticNextStepForDynamicExternalWorkflowError(iconFill?: string) {
+    const optimisticNextStep: ReportNextStepDeprecated = {
+        type: 'alert',
+        icon: CONST.NEXT_STEP.ICONS.DOT_INDICATOR,
+        iconFill,
+        message: [
+            {
+                text: "This report can't be submitted. Please review the comments to resolve.",
```
francoisl, line 342: Do we need translations for this and the `buildOptimisticNextStepForDEWOfflineSubmission` below?

src/languages/en.ts
```diff
@@ -1348,6 +1349,7 @@ const translations = {
             genericDeleteFailureMessage: 'Unexpected error deleting this expense. Please try again later.',
             genericEditFailureMessage: 'Unexpected error editing this expense. Please try again later.',
             genericSmartscanFailureMessage: 'Transaction is missing fields',
+            genericDEWSubmitFailureMessage: 'Dynamic External Workflow submission failed',
```
francoisl: Hm let's think of a better error message for this, most users don't know and don't care about Dynamic External Workflow is about. For now let's reuse the same message as `genericCreateFailureMessage` and we can adjust later if needed.

src/pages/home/report/PureReportActionItem.tsx
```diff
@@ -1964,6 +1981,7 @@ export default memo(PureReportActionItem, (prevProps, nextProps) => {
         prevProps.shouldHighlight === nextProps.shouldHighlight &&
         deepEqual(prevProps.bankAccountList, nextProps.bankAccountList) &&
         prevProps.reportNameValuePairsOrigin === nextProps.reportNameValuePairsOrigin &&
-        prevProps.reportNameValuePairsOriginalID === nextProps.reportNameValuePairsOriginalID
+        prevProps.reportNameValuePairsOriginalID === nextProps.reportNameValuePairsOriginalID &&
+        prevProps.reportMetadata?.pendingExpenseAction === nextProps.reportMetadata?.pendingExpenseAction
```
francoisl, line 2009: Do we also need to add `isOffline` here since we added it to the props?

src/libs/SearchUIUtils.ts
```diff
@@ -1356,6 +1358,7 @@ function getActions(
             : undefined;
 
     const chatReport = getChatReport(data, report);
+
```
bernhardoj: NAB: I think this extra newline is unintentional

src/pages/home/report/PureReportActionItem.tsx
```diff
@@ -1197,12 +1207,16 @@ function PureReportActionItem({
...
+            const isDEWPolicy = hasDynamicExternalWorkflow(policy);
+
             if (wasSubmittedViaHarvesting) {
                 children = (
                     <ReportActionItemBasicMessage>
                         <RenderHTML html={`<comment><muted-text>${translate('iou.automaticallySubmitted')}</muted-text></comment>`} />
                     </ReportActionItemBasicMessage>
                 );
+            } else if (isOffline && hasPendingDEWSubmit(reportMetadata, isDEWPolicy)) {
+                children = <ReportActionItemBasicMessage message={translate('iou.queuedToSubmitViaDEW')} />;
```
bernhardoj, line 1225: Nice, this now shows only while offline, but I believe we don't want to show the `submitted` message if the submission is pending while online. @francoisl cmiiw. Also, we should also check if the action has an ADD pending action, otherwise, all SUBMITTED actions will show the queued message.

image


So, maybe we can do like:

```
} else if (hasPendingDEWSubmit(reportMetadata, isDEWPolicy) && action.pendingAction === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD) {
    children = isOffline ? <ReportActionItemBasicMessage message={translate('iou.queuedToSubmitViaDEW')} /> : emptyHTML;
} else {
```

src/libs/NextStepUtils.ts
abzokhattab: I just followed the existing pattern in the nextSteps.. most of the next steps are not translated so i think it would be better to keep it consistent for the user ... i think we had this conv before on another PR .. for reference here is the thread https://github.com/Expensify/App/pull/72663#discussion_r2442684305 


i am open to introduce the translation in this case but i think from a UI point of view i would lean to keep not translated as others and then fix them all at once

src/libs/OptionsListUtils/index.ts
```diff
@@ -708,9 +716,14 @@ function getLastMessageTextForReport({
...
     ) {
         const wasSubmittedViaHarvesting = !isMarkAsClosedAction(lastReportAction) ? (getOriginalMessage(lastReportAction)?.harvesting ?? false) : false;
+        const isDEWPolicy = hasDynamicExternalWorkflow(policy);
+
         if (wasSubmittedViaHarvesting) {
             // eslint-disable-next-line @typescript-eslint/no-deprecated
             lastMessageTextFromReport = Parser.htmlToText(translateLocal('iou.automaticallySubmitted'));
+        } else if (isOffline && hasPendingDEWSubmit(reportMetadata, isDEWPolicy)) {
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
+            lastMessageTextFromReport = translateLocal('iou.queuedToSubmitViaDEW');
```
bernhardoj, line 717: Similar to above. When online and pending, submitted shouldn't be shown

src/pages/home/report/PureReportActionItem.tsx
abzokhattab: should we update it only in this case? if we have many cases then i think it makes sense to consider this instead 
https://github.com/Expensify/App/pull/75712#discussion_r2623630994

tests/actions/ReportPreviewActionUtilsTest.ts
```diff
@@ -536,4 +536,98 @@ describe('getReportPreviewAction', () => {
         await waitForBatchedUpdatesWithAct();
         expect(getReportPreviewAction(isReportArchived.current, CURRENT_USER_ACCOUNT_ID, report, policy, [transaction])).toBe(CONST.REPORT.REPORT_PREVIEW_ACTIONS.EXPORT_TO_ACCOUNTING);
     });
+
+    describe('DEW (Dynamic External Workflow) submit pending', () => {
+        it('should return VIEW action when DEW submit is pending (offline) and report is OPEN', async () => {
+            // Given an open expense report with a corporate policy where DEW submit is pending (offline)
```
bernhardoj: I think we should remove the offline part here because it's not affected by offline state.

tests/actions/ReportPreviewActionUtilsTest.ts
```diff
@@ -536,4 +536,98 @@ describe('getReportPreviewAction', () => {
...
+            const {result: isReportArchived} = renderHook(() => useReportIsArchived(report?.parentReportID));
+
+            // When getReportPreviewAction is called with isDEWSubmitPending = true
+            const result = getReportPreviewAction(isReportArchived.current, CURRENT_USER_ACCOUNT_ID, report, policy, [transaction], undefined, false, false, false, true);
+
+            // Then it should return VIEW because DEW submission is pending offline
+            expect(result).toBe(CONST.REPORT.REPORT_PREVIEW_ACTIONS.VIEW);
+        });
+
+        it('should return SUBMIT action when DEW submit has failed (not pending) and report is OPEN', async () => {
```
bernhardoj, line 721: I think we don't need this test case anymore because `getReportPreviewAction` only depends on whether the submit is pending or not.

tests/actions/ReportPreviewActionUtilsTest.ts
```diff
@@ -536,4 +536,98 @@ describe('getReportPreviewAction', () => {
...
+            const {result: isReportArchived} = renderHook(() => useReportIsArchived(report?.parentReportID));
+
+            // When getReportPreviewAction is called with isDEWSubmitPending = false (failed, not pending)
+            const result = getReportPreviewAction(isReportArchived.current, CURRENT_USER_ACCOUNT_ID, report, policy, [transaction], undefined, false, false, false, false);
+
+            // Then it should allow SUBMIT because failed submissions can be retried (not VIEW)
+            expect(result).not.toBe(CONST.REPORT.REPORT_PREVIEW_ACTIONS.VIEW);
+        });
+
+        it('should return SUBMIT action when DEW submit has not failed and report is OPEN', async () => {
```
bernhardoj, line 765: Same as above

tests/actions/ReportPreviewActionUtilsTest.ts
abzokhattab: we modified the logic of `getReportPreviewAction` .. to consider for DEW , this is testing them .. idont see a problem with adding this.

tests/actions/ReportPreviewActionUtilsTest.ts
abzokhattab: https://github.com/Expensify/App/pull/75712#discussion_r2629443445

tests/unit/OptionsListUtilsTest.tsx
```diff
@@ -2807,25 +2808,113 @@ describe('OptionsListUtils', () => {
             });
             expect(result).toBe(expectedVisibleText);
         });
-    });
 
-    describe('getPersonalDetailSearchTerms', () => {
```
bernhardoj, line 2858: I think this is unintentional to put `getPersonalDetailSearchTerms` and `getCurrentUserSearchTerms` inside `describe('getLastMessageTextForReport', () => {`, right?

tests/actions/ReportPreviewActionUtilsTest.ts
abzokhattab: pending happens when u are offline.. no ?

tests/unit/OptionsListUtilsTest.tsx
abzokhattab: yeah the brackets missed up from one of the merges to main .. ok fixing it

src/pages/home/report/PureReportActionItem.tsx
bernhardoj: NAB: I think it's more semantically correct if we check whether the SUBMITTED action is a DEW SUBMITTED action instead of checking `hasPendingDEWSubmit`.

We can use this function from the system message PR

image

But we will also need to optimistically set the `workflow` to the SUBMITTED action, which is also available on the system message PR.

tests/actions/ReportPreviewActionUtilsTest.ts
bernhardoj: Yep, but we only check whether it's pending or not, but this test case says when the submission failed, so it's not relevant anymore.

tests/actions/ReportPreviewActionUtilsTest.ts
bernhardoj: We hide the SUBMIT button even when online, while it's pending, too.

src/pages/home/report/PureReportActionItem.tsx
bernhardoj: We need to update the last message text case too (my comment [below](https://github.com/Expensify/App/pull/75712#discussion_r2629420298)).

https://github.com/Expensify/App/pull/75712#discussion_r2623630994 will still have the issue because it checks the whole report actions collection, which is kinda inefficient too.

abzokhattab: lets make it simple and check for the `action.pendingAction === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD` its tecnihcally the same

abzokhattab: @bernhardoj  In the code example, I’m confused about why the offline check is added on a separate line. Why wouldnt it be included it in the same condition? https://github.com/Expensify/App/pull/75712#discussion_r2629417853

src/pages/home/report/PureReportActionItem.tsx
abzokhattab: good point

src/pages/home/report/PureReportActionItem.tsx
bernhardoj: > check for the action.pendingAction === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD

Yeah, we must check this.

> I’m confused about why the offline check is added on a separate line. Why wouldnt it be included it in the same condition?

If we include it on the same line, then when online and pending, it will go to the `else` block which will shows `submitted`, but we don't want `submitted` to be shown.

abzokhattab: @bernhardoj but i think we do want `submitted` to be shown when its pending and online. this is the current behaviour in normal submit so i would expect it to be the same. user clicks on submit the submitted comment is added optimsitcally.

if user is offline then the queued should be shown. 

to be preserve consistency i mean

abzokhattab: i see .. as per https://github.com/Expensify/App/pull/75712#issuecomment-3643842874 we should hide it .. ok

abzokhattab: done

src/libs/OptionsListUtils/index.ts
abzokhattab: done

src/languages/en.ts
abzokhattab: done

tests/unit/OptionsListUtilsTest.tsx
abzokhattab: done

tests/actions/ReportPreviewActionUtilsTest.ts
abzokhattab: done

tests/ui/components/LHNOptionsListTest.tsx
```diff
@@ -175,4 +176,114 @@ describe('LHNOptionsList', () => {
...
+                name: 'DEW Test Policy',
+                type: CONST.POLICY.TYPE.CORPORATE,
+                approvalMode: CONST.POLICY.APPROVAL_MODE.DYNAMICEXTERNAL,
+            } as Policy;
+            const report: Report = {
+                reportID,
+                reportName: 'DEW Test Report',
+                type: CONST.REPORT.TYPE.EXPENSE,
+                policyID,
+                participants: {1: {notificationPreference: 'always'}, 2: {notificationPreference: 'always'}},
```
francoisl: I think we can disable that `typescript-eslint/naming-convention` rule for this line, the type definition for `participants` is `Record<number, Participant>`. We also have a similar thing in `OptionsListUtilsTest` and the rule is disabled there too

bernhardoj: I encounter this error too. We can alternatively put the account ID in a variable and use that instead.

tests/unit/OptionsListUtilsTest.tsx
```diff
@@ -2828,4 +2827,92 @@ describe('OptionsListUtils', () => {
             expect(searchTerms2.includes(displayName)).toBe(true);
         });
     });
+
+    describe('DEW (Dynamic External Workflow)', () => {
```
bernhardoj: This is part of `getLastMessageTextForReport`, so let's move it inside the `describe('getLastMessageTextForReport', () => {`

bernhardoj:
@abzokhattab I keep getting this error

image

Can you try to merge with main, please?

Hmm, it happens when I switch to `main`, too.

src/libs/OptionsListUtils/index.ts
```diff
@@ -708,9 +716,16 @@ function getLastMessageTextForReport({
...
+        const isDEWPolicy = hasDynamicExternalWorkflow(policy);
+        const isPendingAdd = lastReportAction.pendingAction === CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD;
+
         if (wasSubmittedViaHarvesting) {
             // eslint-disable-next-line @typescript-eslint/no-deprecated
             lastMessageTextFromReport = Parser.htmlToText(translateLocal('iou.automaticallySubmitted'));
+        } else if (hasPendingDEWSubmit(reportMetadata, isDEWPolicy) && isPendingAdd) {
+            // When DEW submit is pending: show "queued" message if offline, otherwise show nothing
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
+            lastMessageTextFromReport = isOffline ? translateLocal('iou.queuedToSubmitViaDEW') : '';
```
bernhardoj: Can't check it yet because of the error above, but won't this show nothing in the LHN when online?

bernhardoj: This is what happens

https://github.com/user-attachments/assets/3a1b0f08-16da-452c-893f-d0ae2846f44a

bernhardoj:
Just want to confirm, when the submission fails, is it expected that the `submitted` message will show an error?

image

src/libs/OptionsListUtils/index.ts
abzokhattab: This is what u already suggested ..Got confused...

bernhardoj: Yeah, my bad. I think we should exclude the submitted action from the last action, instead.

abzokhattab: yes i think we should revert it back to `else if (hasPendingDEWSubmit(reportMetadata, isDEWPolicy) && isPendingAdd && isOffline) {`

it seem to work better

abzokhattab:
> Just want to confirm, when the submission fails, is it expected that the submitted message will show an error?

 i mean this should be acceptable since the submit action actually failed .. @francoisl what do u think

src/libs/OptionsListUtils/index.ts
bernhardoj: If we do that, `submitted` will still be shown as the last message, even though the `submitted` action is hidden.

EDIT: Oh, I see you now show the `submitted` while pending and online, which contradicts again with this: https://github.com/Expensify/App/pull/75712#issuecomment-3643842874

abzokhattab: I don't get your point can you please share a video 

> Oh, I see you now show the submitted while pending and online, which contradicts again with this: #75712 (comment)

What is the problem of showing it while online and pending?

bernhardoj: Well, that's from the design docs, right? If we are fine with not following the docs, then no problem.

abzokhattab: We agreed to show the queued action only when offline otherwise the submitted is shown 

Cc @francoisl can u please confirm

Also i dont recall that was in the desgin doc 

Overall i think thats the point of using the isoffline check on the queued to submit action because otherwise it should show submitted

abzokhattab:
Also in case we will do it in the backend should we also remove the submitted message on failure?

I noticed that the submitted in case of failure doenst persist if we clear the cache ... is that expected?

src/libs/OptionsListUtils/index.ts
abzokhattab: I think this is the only remaining comment besides this https://github.com/Expensify/App/pull/75712#issuecomment-3679497782

Let us know what u think @francoisl

francoisl: When you submit while online, we don't want to show an optimistic `submitted` action. This is the first mockup in the doc [here](https://docs.google.com/document/d/144a8NLphlLwOlmLGQ9DQf6oi2HacGZdIZlnM4-O802k/edit?tab=t.0#heading=h.ua6stff8ylgl)

image


When you submit while offline, we want to only show a "queued to submit" optimistic action. See the mockup [here](https://docs.google.com/document/d/144a8NLphlLwOlmLGQ9DQf6oi2HacGZdIZlnM4-O802k/edit?tab=t.0#heading=h.d9yhpe64mhit).

image

abzokhattab: I see ... it was not written in the changes description in the doc so i didnt notice it from the image ... let me change this now

bernhardoj:
I think [this](https://github.com/Expensify/App/pull/75712#discussion_r2633680651) is the only issue left

bernhardoj:
There is a missing deps here.
image

src/components/SelectionListWithSections/Search/TransactionListItem.tsx
```diff
@@ -129,18 +130,19 @@ function TransactionListItem<TItem extends ListItem>({
...
             snapshotReport,
             snapshotPolicy,
             lastPaymentMethod,
             currentSearchKey,
             onDEWModalOpen,
+            isDEWBetaEnabled,
             isDelegateAccessRestricted,
-            showDelegateNoAccessModal,
-        );
+            onDelegateAccessRestricted: showDelegateNoAccessModal,
```
bernhardoj, line 145: Looks like we are missing the `personalPolicyID` here. I see other usages pass the `personalPolicyID`. Can we also make it a required param so we won't forget to pass it in the future?

src/libs/ReportSecondaryActionUtils.ts
```diff
@@ -134,12 +143,14 @@ function isSplitAction(report: OnyxEntry<Report>, reportTransactions: Array<Onyx
     return (isSubmitter && isAwaitingFirstLevelApproval(report)) || isAdmin || isManager;
 }
 
+// eslint-disable-next-line @typescript-eslint/max-params
```
bernhardoj: NAB: Can we convert the function param here to an object as you did with `handleActionButtonPress`?

src/libs/actions/Search.ts
```diff
@@ -159,14 +176,13 @@ function getLastPolicyPaymentMethod(
     lastPaymentMethods: OnyxEntry<LastPaymentMethod>,
     reportType: keyof LastPaymentMethodType = 'lastUsed',
     isIOUReport?: boolean,
+    personalPolicyID?: string,
```
bernhardoj: Can we change all new `personalPolicyID` to be required (if needed, we can change it to (`string | undefined` or `OnyxEntry<string>`) so we pass it to all usages? I'm afraid it'll cause regression because previously, `personalPolicy` was always available inside the function, but now we only pass `personalPolicyID` to some usages.

abzokhattab: yes thats a good point ... i changed it to `string | undefined`

src/libs/ReportSecondaryActionUtils.ts
abzokhattab: done

src/components/SelectionListWithSections/Search/TransactionListItem.tsx
abzokhattab: done

bernhardoj (approved):
LGTM, other than the offline case. The queued message is still shown when going online, but I think if we really want to fix this, we should first remove the `Onyx.connect` usage for the last action and then revisit this, but that's for another issue.

https://github.com/user-attachments/assets/0b0571a8-ac8a-47b2-8dec-c6929deb775a

We can merge this for now.

francoisl (dismissed):
Let's do this!!

francoisl:
Hmm Github is telling me some commits are not signed, so it won't let me merge because they're required at the repo level. The GH UI isn't loading all commit details on the /commits page, I'll check on the command line.

francoisl:
Those are the unsigned commits it looks like:
```
❯ git log --format="%H %G? %s" main..HEAD | grep ' N '
c1a38a0cd04baed162d71aa9f39f0d69afab8e4c N Refactoring test cases
b85993e30c7c8e3445f621bdd1fee994ed50ad7d N making err as next step instead of next step status
3261d5592d186a53adcdaf05fd488e05f4e781c6 N remvoe the unneeded hasDewError const
```

francoisl (approved):

izarutskaya:
Let me know when it's done. Thank you in advance

izarutskaya:
@francoisl Could you please enable it in this account applausetester+ndk74@applause.expensifail.com as well? Thank you

PR #76159, opened 2025-11-26
Moving Per Diem expense from self DM

koko57:
### Fixed Issues

$ https://github.com/Expensify/App/issues/72239
PROPOSAL:

### Tests

Prerequisites: a few workspaces, some with per diem enabled, selfDM

1. Create a per diem expense on one of the workspaces with per diem enabled
2. Go to the expense
3. Click report 
4. Click the "Remove from report" option
5. Go to self DM
6. Go to the expense you removed from the workspace expense report in step 4
7. Click report
8. Verify that on the list of reports you see only the reports from the workspace that per diem was created in (even if the default workspace is different)
9. Verify that Create Report button has the per diem's workspace name (even if the default workspace is different)
10. Verify that after clicking Create Report a new report is created on this workspace and the per diem expense is moved there

NOTE: [known issue] Category field (and other expense fields) show categories from default workspace in per diem tracked expense

Moving from the expense report

1. Create a few per diem expenses on one of the workspaces with per diem enabled
2. Go to the report view (with expenses list)
3. Select a few expenses
4. Click the "X selected" button
5. Choose the "Move expenses" option
6. Verify that on the list of reports, you see only the reports from the workspace that per diem expenses were created in (even if the default workspace is different)
7. Verify that the "Create Report" button has the per diem's workspace name (even if the default workspace is different)
10. Verify that after clicking the "Create Report" button a new report is created on this workspace, and the per diem expenses are moved there

Deleted workspace

1. Create a per diem expense on one of the workspaces with per diem enabled
2. Go to the expense
3. Click report 
4. Click the "Remove from report" option
5. Go to workspaces
6. Remove the workspace that you created per diem in
7. Go to self DM
8. Go to the expense you removed from the workspace expense report in step 4
7. Verify that the report field is disabled

- [x] Verify that no errors appear in the JS console

### QA Steps

// TODO: These must be filled out, or the issue title must include "[No QA]."

Same as tests

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
[...]

PR #77651, opened 2025-12-15
Revert "Moving Per Diem expense from self DM"

cretadn22:
Reverts Expensify/App#76159


koko57:
@cretadn22 why are you reverting the whole PR? I can fix it quickly

PR #79475, opened 2026-01-13
[No QA] [ECUK In-App 3DS] Add `BIOMETRICS_TEST` scenario screens & routes + add test biometrics button to the troubleshoot menu

JakubKorytko:
### Explanation of Change

This PR adds the components & pages used for the `BIOMETRICS_TEST` Multifactor Authentication Scenario.
At the current state, the screens simply navigate between each other - the correct logic will be added as a part of this PR: https://github.com/Expensify/App/pull/79482

This allows us to test the UI, animations, and overall appearance.
It also adds the test biometrics button to the Troubleshoot menu so we can start the navigation flow.
For now, the native biometrics flow is available on the web, though it will be context-restricted in the future.

This is because the same components & pages with different images & text will be used for the passkeys later on, so their appearance can be tested on the web now.

co-author: @jakubkalinski0 

### Fixed Issues

$ https://github.com/Expensify/App/issues/79373
$ https://github.com/Expensify/App/issues/79377
PROPOSAL: N/A

### Tests

1. Open Expensify App
2. Navigate to the Settings -> Troubleshoot
3. Click on the "Test" button next to the "Biometrics (Not registered)" text.
4. Click on the "Test" button in the RHP view.
5. Verify that the validate code screen matches the one in the 3DS UI mocks
6. Fill the magic code input with any valid number (it is not validated for now)
7. Verify that the "Verify yourself" screen matches the one in the 3DS UI mocks
8. Click on the "Got it" button
9. Verify that the "Authentication successful" screen matches the one in the 3DS UI mocks
10. Repeat steps 1-7
11. Click on the back arrow
12. Verify that the reject confirmation modal matches the one in the 3DS UI mocks
13. Click on the "Reject authentication" button
14. Verify that the "Authentication unsuccessful" screen matches the one in the 3DS UI mocks

### Offline tests

N/A, `D - Full Page Blocking UI Pattern` for this project.

### QA Steps

Same as tests.

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
[...]

Issue #79376, opened 2026-01-12
[Due for payment 2026-02-04] [ECUK In-App 3DS] [App] Add actions and wrappers for the auth keys & challenges logic


JakubKorytko:
Part of [Adding biometric 3DS authentication for Expensify Cards](https://github.com/Expensify/Expensify/issues/523107)

Implement [this section](https://docs.google.com/document/d/15il4CanenwHJUEdjqOuWvxCFABM_3uRhfbuNxC7q4xo/edit?tab=t.0#bookmark=id.ka025l115mbg) of the design doc. These changes should include:
- Adding the wrapper for managing and signing challenges
- Adding the wrapper for I/O operations on the SecureStore entries

PR: https://github.com/Expensify/App/pull/79477

<details><summary>Issue Owner</summary>Current Issue Owner: @JakubKorytko</details>

rafecolton:
Nothing further is needed here, as Dylann is getting paid for the project as a whole

Issue #79059, opened 2026-01-08
[Due for payment 2026-02-04] [ECUK/EU In-App 3DS] Implement `TroubleshootMultifactorAuthentication` action & `BIOMETRICS_TEST` scenario


rafecolton:
The first multifactor scenario we will use in the project is `BIOMETRICS_TEST`, which corresponds to the `TroubleshootMultifactorAuthentication` action and API command. Please implement per [this section of the doc](https://expensify.slack.com/archives/C0992M3HD26/p1767843070320839) and the [corresponding UI section from the HL](https://expensify.slack.com/archives/C0992M3HD26/p1767843118540539).

- [ ] PR: https://github.com/Expensify/App/pull/79477

<details><summary>Issue Owner</summary>Current Issue Owner: @JakubKorytko</details>

rafecolton:
Nothing further is needed here

PR #79473, opened 2026-01-13
[ECUK In-App 3DS] Implement `MultifactorAuthenticationContext`, `useNativeBiometrics` & helpers

JakubKorytko:
### Explanation of Change

This PR adds:
- The `useMultifactorAuthenticationStatus` helper hook
- The `useNativeBiometricsSetup` internal hook
- The `useNativeBiometrics` hook
- `MultifactorAuthenticationContext`
- Automated tests

### Fixed Issues

$ https://github.com/Expensify/App/issues/79371
$ https://github.com/Expensify/App/issues/79058
$ https://github.com/Expensify/App/issues/81020
PROPOSAL: N/A

### Tests

1. Open Expensify App
2. Navigate to the Settings -> Troubleshoot
3. Click on the "Test" button next to the "Biometrics (Not registered)" text.
4. Click on the "Test" button in the RHP view.
5. Fill the magic code input
6. Click on the "Got it" button
7. Authenticate using device credentials or biometry
8. Verify that the Authentication was successful
9. Verify that the `registered` label has changed to `true`
10. Repeat steps 3 and 4
11. Verify that validate code is no longer required
12. Repeat test, but:
  - Fail the authentication
  - Input the wrong validate code (previously revoking the access on the Security page)
  - Cancel the authentication using back arrow / clicking outside the RHP
13. Verify that the Authentication failed 

### Offline tests

N/A, `D - Full Page Blocking UI Pattern` for this project.

### QA Steps

Same as tests

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
[...]

PR #79477, opened 2026-01-13
[No QA] [ECUK In-App 3DS] Add essential actions, wrappers for the auth keys, challenges logic & `BIOMETRICS_TEST` scenario

JakubKorytko:
### Explanation of Change

This PR adds:
- The wrapper for managing and signing challenges
- The wrapper for I/O operations on the SecureStore entries
- `TroubleshootMultifactorAuthentication` action and API command
- All other actions needed for the 1st release of 3DS
- `BIOMETRICS_TEST` Scenario and its config
- Automated tests
- NABs from the https://github.com/Expensify/App/pull/79475

### Fixed Issues

$ https://github.com/Expensify/App/issues/79376
$ https://github.com/Expensify/App/issues/79059
PROPOSAL: N/A

### Tests

N/A

### Offline tests

N/A, `D - Full Page Blocking UI Pattern` for this project.

### QA Steps

N/A

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/components/MultifactorAuthentication/config/helpers.ts
```diff
@@ -0,0 +1,62 @@
...
+    MultifactorAuthenticationScenario,
+    MultifactorAuthenticationScenarioConfigRecord,
+} from './types';
+
+/**
+ * Converts a string to lowercase.
+ */
+function toLowerCase<T extends string>(str: T) {
+    return str.toLowerCase() as Lowercase<T>;
+}
```
rafecolton: Why do we need a function for this as opposed to using the built-in function?

src/components/MultifactorAuthentication/config/helpers.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Converts camelCase string to kebab-case format.
+ */
+function camelToKebabCase<T extends string>(str: T) {
+    return str.replaceAll(/([a-z])([A-Z])/g, '$1-$2').toLowerCase() as KebabCase<T>;
+}
```
rafecolton: As these functions are not specific to multifactor authentication, if they are indeed needed, they should go in a more general file such as a stringutils file. I'm also curious if these exist anywhere else, I'd be surprised if nobody has needed this before

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Converts camelCase string to kebab-case format.
+ */
+function camelToKebabCase<T extends string>(str: T) {
+    return str.replaceAll(/([a-z])([A-Z])/g, '$1-$2').toLowerCase() as KebabCase<T>;
+}
+
+/**
+ * Creates a notification record from multifactor authentication scenario configuration.
+ */
+const createNotificationRecord = (mfaConfig: MultifactorAuthenticationScenarioConfigRecord): MultifactorAuthenticationNotificationRecord => {
```
rafecolton: I understand why you'd use the term "notification" for this, since we are notifying the user of the outcome. However, I find it a bit confusing - the term is a bit overloaded and refers to other paradigms that are not happening here. I'd love to pick a different term for scenario results that is a little more specific - perhaps "outcome"?

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Converts camelCase string to kebab-case format.
+ */
+function camelToKebabCase<T extends string>(str: T) {
+    return str.replaceAll(/([a-z])([A-Z])/g, '$1-$2').toLowerCase() as KebabCase<T>;
+}
+
+/**
+ * Creates a notification record from multifactor authentication scenario configuration.
```
rafecolton: It's not super obvious reading the code what it's doing and why, so please elaborate here. It's helpful to show what an example input and output would look like.

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
```diff
@@ -0,0 +1,62 @@
...
+    const entries = Object.entries({...mfaConfig});
+    return entries.reduce((record, [key, {NOTIFICATIONS}]) => {
+        // eslint-disable-next-line no-param-reassign
+        record[key as MultifactorAuthenticationScenario] = {...NOTIFICATIONS};
+        return record;
+    }, {} as MultifactorAuthenticationNotificationRecord);
+};
+
+/**
+ * Creates a notification key by combining scenario and notification name in kebab-case format.
```
rafecolton: Same as above, please elaborate

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+const createNotificationKey = (key: string, name: string) => {
+    const scenarioKebabCase = toLowerCase(key as MultifactorAuthenticationScenario);
+    const notificationName = camelToKebabCase(name as MultifactorAuthenticationNotificationOptions);
+
+    return `${scenarioKebabCase}-${notificationName}` as const;
+};
+
+/**
+ * Maps multifactor authentication scenario configuration to a notification map with kebab-case keys.
```
rafecolton: Same

src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts
```diff
@@ -0,0 +1,25 @@
+import type {MultifactorAuthenticationScenarioCustomConfig} from '@components/MultifactorAuthentication/config/types';
+import {troubleshootMultifactorAuthentication} from '@userActions/MultifactorAuthentication';
+import CONST from '@src/CONST';
+import SCREENS from '@src/SCREENS';
+
+/**
+ * Configuration for the biometrics test multifactor authentication scenario.
+ */
+export default {
+    allowedAuthentication: CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS,
```
rafecolton: Should this be a list? 

```suggestion
    allowedAuthenticationMethods: CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS,
```

src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts
```diff
@@ -0,0 +1,25 @@
...
+    action: troubleshootMultifactorAuthentication,
+    screen: SCREENS.MULTIFACTOR_AUTHENTICATION.BIOMETRICS_TEST,
+    pure: true,
```
rafecolton, line 13: What is this for?

src/components/MultifactorAuthentication/config/scenarios/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import CONST from '@src/CONST';
+import BiometricsTest from './BiometricsTest';
+import {customConfig} from './DefaultUserInterface';
+
+/**
+ * Payload types for multifactor authentication scenarios.
+ * Since the BiometricsTest does not require any payload, it is an empty object for now.
+ * The AuthorizeTransaction Scenario will change it.
+ */
+type Payloads = EmptyObject;
```
rafecolton, line 18: Right?

```suggestion
type Payload = EmptyObject;
```

Can you also update the comment to indicate what purpose the payload serves in a given scenario?

src/components/MultifactorAuthentication/config/scenarios/names.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Multifactor authentication scenario names.
```
rafecolton, line 2: NAB does this need to be a separate file?

src/components/MultifactorAuthentication/config/scenarios/prompts.ts
```diff
@@ -0,0 +1,16 @@
+import LottieAnimations from '@components/LottieAnimations';
+import type {MultifactorAuthenticationPrompt} from '@components/MultifactorAuthentication/config/types';
+
+/**
+ * Configuration for multifactor authentication prompt UI with animations and translations.
```
rafecolton, line 6: Wondering if this file can be combined too. Basically with all these little files it'll be much harder to understand what's happening after-the-fact, and it will require touching that many more files each time a new scenario is added. I'm wondering if it's possible and practical to have the core scenario implementation in a single file, or perhaps a very small number, then have a single file per individual scenario.

src/components/MultifactorAuthentication/config/scenarios/prompts.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+/* eslint-disable @typescript-eslint/naming-convention */
+export default {
+    'enable-biometrics': {
```
rafecolton: Is this eslint-disable so we can name the scenario `enable-biometrics` instead of `enableBiometrics`? Is that also what's requiring the string conversion functions? Wondering if that all can be simplified

src/components/MultifactorAuthentication/config/types.ts
```diff
@@ -0,0 +1,259 @@
...
+type MultifactorAuthenticationModal = {
+    cancelConfirmation: MultifactorAuthenticationCancelConfirm;
+};
+
+/**
+ * Optional modal configuration for scenarios that may not have custom modals.
+ */
+type MultifactorAuthenticationModalOptional = {
+    cancelConfirmation?: Partial<MultifactorAuthenticationCancelConfirm>;
+};
```
rafecolton, line 76: What is this used for?

src/components/MultifactorAuthentication/types.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Type definitions for multifactor authentication components.
+ */
+import type {ValueOf} from 'type-fest';
+import type {SECURE_STORE_VALUES} from '@libs/MultifactorAuthentication/Biometrics/SecureStore';
+
+/**
+ * Authentication type name derived from secure store values.
```
rafecolton, line 8: Can you give an example? Similar to my comment on the other types file, I think this will be easier to grok if placed adjacent the type declaration for `MultifactorAuthenticationTranslationParams`, which seems to be the only place it's used.

src/components/MultifactorAuthentication/config/types.ts
```diff
@@ -0,0 +1,259 @@
+/**
+ * Configuration types for multifactor authentication UI and scenarios.
```
rafecolton, line 2: I'm finding it difficult to review this file. I'm not sure if it's a common pattern to have all the types declared in their own file, but I think having them separated from objects of those types makes it harder to understand the shape. Would it be possible / practical / idiomatic to move these type declarations to where you create the default instances so the reader can see an adjacent example for each one?

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+import type {ChallengeType, MultifactorAuthenticationPartialStatus, MultifactorAuthenticationReason, MultifactorKeyStoreOptions} from './types';
+import VALUES from './VALUES';
+
+/**
+ * Handles the complete lifecycle of a multifactor authentication challenge for a specific scenario.
+ * Manages requesting challenges from the server, signing them with the private key, and sending signed challenges back.
+ */
+class MultifactorAuthenticationChallenge<T extends MultifactorAuthenticationScenario> {
+    private auth: MultifactorAuthenticationPartialStatus<MultifactorAuthenticationChallengeObject | SignedChallenge | undefined, true> = {
+        value: undefined,
```
rafecolton: Is it idiomatic to use `undefined` here as opposed to null or an empty object?

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+        private readonly params: MultifactorAuthenticationScenarioAdditionalParams<T>,
+        private readonly options?: MultifactorKeyStoreOptions,
+        private readonly challengeType: ChallengeType = 'authentication',
+    ) {}
+
+    /**
+     * Creates an error return value with the given reason.
+     */
+    private createErrorReturnValue(reasonKey: MultifactorAuthenticationReason): MultifactorAuthenticationPartialStatus<boolean, true> {
+        return {value: false, reason: reasonKey};
```
rafecolton, line 35: why `false` here vs null or empty object?

rafecolton: I guess I'm finding the term `value` confusing as it is seems generic and overloaded, but perhaps it will be more obvious as I read on 🤔

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+        this.publicKeys = authPublicKeys ?? [];
+
+        const reason = apiReason === VALUES.REASON.BACKEND.UNKNOWN_RESPONSE ? VALUES.REASON.CHALLENGE.BAD_TOKEN : apiReason;
+
+        this.auth = {
+            value: challenge,
+            reason: challenge ? VALUES.REASON.CHALLENGE.CHALLENGE_RECEIVED : reason,
+        };
+
+        return {...this.auth, value: true};
```
rafecolton: What does `value: true` mean here? Won't it get overwritten by the `value: challenge` key from `...this.auth`?

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+        accountID: number,
+        chainedPrivateKeyStatus?: MultifactorAuthenticationPartialStatus<string | null, true>,
+    ): Promise<MultifactorAuthenticationPartialStatus<boolean, true>> {
+        if (!this.auth.value) {
+            return this.createErrorReturnValue(VALUES.REASON.CHALLENGE.CHALLENGE_MISSING);
+        }
+
+        if (isChallengeSigned(this.auth.value)) {
+            return this.createErrorReturnValue(VALUES.REASON.CHALLENGE.CHALLENGE_ALREADY_SIGNED);
+        }
```
rafecolton: Why return an error here? If the challenge is already signed and the purpose of this function is to sign the challenge, seems like the outcome would be success?

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+
+        if (!value) {
+            return this.createErrorReturnValue(reason || VALUES.REASON.KEYSTORE.KEY_MISSING);
+        }
+
+        const {value: publicKey} = await PublicKeyStore.get(accountID);
+
+        if (!publicKey || !this.publicKeys.includes(publicKey)) {
+            await Promise.all([PrivateKeyStore.delete(accountID), PublicKeyStore.delete(accountID)]);
+            return this.createErrorReturnValue(VALUES.REASON.KEYSTORE.KEY_MISSING_ON_THE_BACKEND);
```
rafecolton: NAB the BE will return "registration required" if the user tries to sign a challenge with a key that's not recognized, so consider using the same language here

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        this.auth = {
+            value: signTokenED25519(this.auth.value, value, publicKey),
+            reason: VALUES.REASON.CHALLENGE.CHALLENGE_SIGNED,
+            type,
+        };
+
+        return {...this.auth, value: true};
```
rafecolton: Similar to above, won't `value: true` get overwritten here?

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+        this.auth = {
+            value: challenge,
+            reason: challenge ? VALUES.REASON.CHALLENGE.CHALLENGE_RECEIVED : reason,
+        };
+
+        return {...this.auth, value: true};
+    }
+
+    /**
+     * Signs the challenge with the user's private key and validates against public keys from the server.
```
rafecolton:
```suggestion
     * Checks if the secure store contains a public key known to the server. If so, challenge is signed using that key.
     * If not, the step fails indicating that registration of a new key is required.
```

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,130 @@
...
+            type,
+        };
+
+        return {...this.auth, value: true};
+    }
+
+    /**
+     * Sends the signed challenge to the server for the specific scenario.
+     */
+    public async send(): Promise<MultifactorAuthenticationPartialStatus<boolean, true>> {
```
rafecolton, line 89: I'm a little confused about this method - the name and description suggests that this sends the signed challenge independent of the business logic that uses it. But I don't think that's the case, since `processScenario()` calls the scenario's action. I also don't see this method used anywhere besides tests. What is the purpose of this method?

src/libs/MultifactorAuthentication/Biometrics/KeyStore.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+const secureStoreOptions = (key: string, KSOptions?: MultifactorKeyStoreOptions): SecureStoreOptions => {
+    const isPrivateKey = key.endsWith(VALUES.KEY_ALIASES.PRIVATE_KEY);
+
+    return {
+        failOnUpdate: isPrivateKey,
+        requireAuthentication: isPrivateKey,
+        forceAuthenticationOnSave: isPrivateKey,
+        forceReadAuthenticationOnSimulators: isPrivateKey,
+        authenticationPrompt: KSOptions?.nativePromptTitle ?? 'Approve transaction',
```
rafecolton: Are you sure we want to have this default value? This seems like something that should always be passed in

src/libs/MultifactorAuthentication/Biometrics/KeyStore.ts
```diff
@@ -0,0 +1,129 @@
...
+    };
+};
+
+/**
+ * Unified interface for secure storage operations with proper error handling and authentication.
+ */
+const MultifactorAuthenticationStore = {
```
rafecolton: What's the value in having both this and `class MultifactorAuthenticationKeyStore` below? They seem very similar

src/libs/MultifactorAuthentication/Biometrics/Observer.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Observer pattern implementation for multifactor authentication callbacks.
+ */
+import {MultifactorAuthenticationCallbacks} from './VALUES';
+
+/**
+ * Manages registration and storage of multifactor authentication callback functions.
+ * Used for subscribing to authentication flow events.
+ */
+const MultifactorAuthenticationObserver = {
```
rafecolton, line 10: What is the purpose of this? I only see it being used in the tests

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,220 @@
...
+        VALIDATE_CODE_MISSING: 'Validate code is missing',
+        VALIDATE_CODE_INVALID: 'Validate code is invalid',
+        BIOMETRICS_REGISTERED: 'Biometrics registration successful',
+        UNABLE_TO_AUTHORIZE: 'Authorization failed with provided credentials',
+        AUTHORIZATION_SUCCESSFUL: 'User authorized successfully',
+        BAD_REQUEST: 'Bad request',
+        UNKNOWN_RESPONSE: 'Unrecognized response type',
+    },
+    CHALLENGE: {
+        BAD_TOKEN: 'Bad token',
```
rafecolton: What is a token in this context? Is it the nonce?

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,220 @@
...
+ */
+const MultifactorAuthenticationCallbacks: {
+    onFulfill: Record<string, () => void>;
+} = {
+    onFulfill: {},
+};
+
+/**
+ * Backend reason messages for multifactor authentication responses.
+ */
```
rafecolton: This comment seems outdated, as these are not all backend reasons. Also, are all of these used in this PR or by existing code?

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,220 @@
...
+        KEY_NOT_FOUND: 'Key not found in SecureStore',
+        UNABLE_TO_RETRIEVE_KEY: 'Failed to retrieve key from SecureStore',
+    },
+} as const;
+
+/**
+ * Maps API endpoints to their HTTP status codes and corresponding reason messages.
+ */
+/* eslint-disable @typescript-eslint/naming-convention */
+const API_RESPONSE_MAP = {
```
rafecolton, line 85: Hm, I don't think this will work. The API may send the same response code for multiple error messages, so you will need to look at both the code and the reason text.

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,220 @@
...
+} as const;
+/* eslint-enable @typescript-eslint/naming-convention */
+
+/**
+ * Factor origin types for multifactor authentication.
+ */
+const MULTIFACTOR_AUTHENTICATION_FACTOR_ORIGIN = {
+    BIOMETRICS: 'Biometrics',
+    ADDITIONAL: 'Additional',
+} as const;
```
rafecolton, line 124: What is this? What is "Additional"?

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,220 @@
...
+
+/**
+ * Available multifactor authentication factors.
+ */
+const MULTIFACTOR_AUTHENTICATION_FACTORS = {
+    SIGNED_CHALLENGE: 'SIGNED_CHALLENGE',
+    VALIDATE_CODE: 'VALIDATE_CODE',
```
rafecolton, line 131: I don't think it's correct. Validate code is not being used as a factor of authentication. It is meant to verify the user's account ownership shortly before taking a sensitive action, since due to infinite sessions, the user may have last been verified a long time ago. Technically, a signed challenge is not a factor of authentication either per se. The factors are the device-bound key (possession) and the method used to access it (either knowledge for passcode or inherence for face/fingerprint).

As a result, I think a lot of the methods you have in `src/libs/MultifactorAuthentication/Biometrics/helpers.ts` that relate to factors are not really correct either

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,220 @@
...
  * Centralized constants used by the multifactor authentication biometrics flow.
  * It is stored here instead of the CONST file to avoid circular dependencies.
  */
 const MULTIFACTOR_AUTHENTICATION_VALUES = {
-    /** Referred to as the EdDSA in the Auth */
+    /**
+     * Keychain service name for secure key storage.
+     */
+    KEYCHAIN_SERVICE: 'Expensify',
+    /**
```
rafecolton, line 184: Per general style guide, comments should have a blank line above them unless they are the first line in a block. Please apply throughout your PR.

```suggestion
    KEYCHAIN_SERVICE: 'Expensify',

    /**
```

rafecolton (changes requested):
Overall observations:
* There are many small files, with a lot of boilerplate. This makes it difficult to understand what's happening even with the full context in this PR because it requires holding many references in mind. When someone needs to reverse engineer this in six months to fix a bug, it will be even more difficult.
* Add more comments 😄 For functions, show example inputs and outputs. For types, show example shapes broken down to primitive types.

src/libs/actions/MultifactorAuthentication.ts
```diff
@@ -0,0 +1,59 @@
...
+ *   we should not delay the possibility to do so for the user.
+ *
+ * This is not a standard practice in the code base.
+ * Please consult before using this pattern.
+ */
+
+async function registerAuthenticationKey({keyInfo, validateCode}: MultifactorAuthenticationScenarioParameters['REGISTER-BIOMETRICS']) {
+    if (!validateCode) {
+        return parseHttpCode(401, CONST.MULTIFACTOR_AUTHENTICATION.API_RESPONSE_MAP.REGISTER_AUTHENTICATION_KEY);
+    }
```
DylanDylann: We should make validateCode required instead of faking a 401 response here

src/libs/API/types.ts
```diff
@@ -1320,6 +1324,9 @@ type SideEffectRequestCommandParameters = {
     [SIDE_EFFECT_REQUEST_COMMANDS.ACCEPT_SPOTNANA_TERMS]: Parameters.AcceptSpotnanaTermsParams;
     [SIDE_EFFECT_REQUEST_COMMANDS.GET_SCIM_TOKEN]: Parameters.GetScimTokenParams;
     [SIDE_EFFECT_REQUEST_COMMANDS.COMPLETE_GUIDED_SETUP]: Parameters.CompleteGuidedSetupParams;
+    [SIDE_EFFECT_REQUEST_COMMANDS.REGISTER_AUTHENTICATION_KEY]: Parameters.RegisterBiometricsParams;
```
DylanDylann:
```suggestion
    [SIDE_EFFECT_REQUEST_COMMANDS.REGISTER_AUTHENTICATION_KEY]: Parameters.RegisterAuthenticationKeyParams;
```

src/libs/API/types.ts
```diff
@@ -1320,6 +1324,9 @@ type SideEffectRequestCommandParameters = {
...
+    [SIDE_EFFECT_REQUEST_COMMANDS.TROUBLESHOOT_MULTIFACTOR_AUTHENTICATION]: Parameters.BiometricsTestParams;
```
DylanDylann:
```suggestion
    [SIDE_EFFECT_REQUEST_COMMANDS.TROUBLESHOOT_MULTIFACTOR_AUTHENTICATION]: Parameters.TroubleshootMultifactorAuthenticationParams;
```

src/libs/API/types.ts
```diff
@@ -1320,6 +1324,9 @@ type SideEffectRequestCommandParameters = {
...
+    [SIDE_EFFECT_REQUEST_COMMANDS.REQUEST_AUTHENTICATION_CHALLENGE]: Parameters.RequestBiometricChallengeParams;
```
DylanDylann:
```suggestion
    [SIDE_EFFECT_REQUEST_COMMANDS.REQUEST_AUTHENTICATION_CHALLENGE]: Parameters.RequestAuthenticationChallengeParams;
```

src/libs/API/types.ts
DylanDylann: Let's name the params to match with the command name

src/libs/actions/MultifactorAuthentication.ts
```diff
@@ -0,0 +1,59 @@
...
+        challenge,
+        publicKeys,
+    };
+}
+
+async function troubleshootMultifactorAuthentication({signedChallenge}: MultifactorAuthenticationScenarioParameters['BIOMETRICS-TEST']) {
+    if (!signedChallenge) {
+        return parseHttpCode(400, CONST.MULTIFACTOR_AUTHENTICATION.API_RESPONSE_MAP.TROUBLESHOOT_MULTIFACTOR_AUTHENTICATION);
+    }
+
```
DylanDylann: We should make signedChallenge required instead of faking a 400 response here

src/libs/MultifactorAuthentication/Biometrics/helpers.ts
```diff
@@ -0,0 +1,213 @@
...
+            isRequestFulfilled: false,
+        },
+        reason: VALUES.REASON.GENERIC.FACTORS_VERIFIED,
+    };
+}
+
+/**
+ * Processes the authorization response and determines the next step in the authentication flow.
+ */
+const authorizeMultifactorAuthenticationPostMethod = <T extends MultifactorAuthenticationScenario>(
```
DylanDylann: Why the verb is authorize? I think we should rename the verb to other verb, ex: transformMultifactorAuthenticationActionResponse

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
```diff
@@ -109,13 +109,22 @@ function createAuthenticatorData(rpId: string): Bytes {
...
-function signToken(credentialRequestOptions: MultifactorAuthenticationChallengeObject, privateKey: string): SignedChallenge {
-    const rawId: Base64URLString = Base64URL.encode(VALUES.KEY_ALIASES.PUBLIC_KEY);
-    const type = VALUES.ED25519_TYPE;
+function signToken(credentialRequestOptions: MultifactorAuthenticationChallengeObject, privateKey: string, publicKey: Base64URLString): SignedChallenge {
+    // rawId should be the base64url-encoded public key, serving as credential identifier
+    const rawId: Base64URLString = publicKey;
+    const type = VALUES.ED25519_TYPE; // "biometric"
+
+    // Extract rpId from challenge - handle both authentication and registration formats
+    const rpId = 'rpId' in credentialRequestOptions ? credentialRequestOptions.rpId : credentialRequestOptions.rp.id;
```
DylanDylann, line 125: Could we update the AuthenticationChallenge type to match WebAuthn format 


Screenshot 2026-01-20 at 17 26 15

src/components/MultifactorAuthentication/config/types.ts
DylanDylann: Cancel confirmation modal will be displayed in this flow. We need to display this modal when users click outside the modal too

https://github.com/user-attachments/assets/a593f20d-2b9f-44ae-aa70-b8fd3b76820a

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
dariusz-biela: Hmm, in theory we could do an early return here, but on the other hand, signing the same challange twice is clearly a flaw in logic that we should be aware of.

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
dariusz-biela: I think the logic here is correct, but I would change the name from `value` to `success: boolean`, because the value here is actually two different variables.

src/components/MultifactorAuthentication/ValidateCodeResendButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+type MultifactorAuthenticationValidateCodeResendButtonHandle = {
+    resetCountdown: () => void;
+};
+
+type MultifactorAuthenticationValidateCodeResendButtonProps = {
+    ref?: React.Ref<MultifactorAuthenticationValidateCodeResendButtonHandle>;
+    shouldDisableResendCode: boolean;
+    hasError: boolean;
+    resendButtonText: TranslationPaths;
```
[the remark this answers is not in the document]
dariusz-biela: in react 19 we don't have to do this

src/components/MultifactorAuthentication/config/helpers.ts
JakubKorytko: Added a comment here, basically if a const "TEST" string is provided as an argument then the returned type is "test", in a built-in function the returned type would be a primitive `string`

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/libs/StringUtils/index.ts#L186-L193

src/components/MultifactorAuthentication/config/helpers.ts
JakubKorytko: yeah, they do not exist, moved to the StringUtils

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/libs/StringUtils/index.ts#L185-L202

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
JakubKorytko: added!

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts#L10-L54

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
JakubKorytko: > I understand why you'd use the term "notification" for this, since we are notifying the user of the outcome.

Oh, I mean, the primary reason is to keep it consistent with the design docs, which referred to it this way. Should we change the name here only, or in both places? Or should we leave it as it is?

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
JakubKorytko: donee

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts#L69-L78

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
JakubKorytko: added an example I/O and explanation at the top of the file so leaving this as it is to not repeat myself

src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts
JakubKorytko: yeah, changed

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts#L9-L12

src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts
JakubKorytko: added an explanation, it is used in the run-time to skip payload check 

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/components/MultifactorAuthentication/config/types.ts#L163-L168

src/components/MultifactorAuthentication/config/scenarios/DefaultUserInterface.ts
```diff
@@ -0,0 +1,104 @@
+import type {MultifactorAuthenticationDefaultUIConfig, MultifactorAuthenticationScenarioCustomConfig} from '@components/MultifactorAuthentication/config/types';
+import NoEligibleMethodsDescription from '@components/MultifactorAuthentication/NoEligibleMethodsDescription';
+// eslint-disable-next-line no-restricted-imports
+import spacing from '@styles/utils/spacing';
```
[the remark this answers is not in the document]
JakubKorytko: done

src/components/MultifactorAuthentication/config/scenarios/index.ts
JakubKorytko: `Payloads` actually as it stores payload for every scenario by key, explanation added

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/components/MultifactorAuthentication/config/scenarios/index.ts#L7-L18

src/components/MultifactorAuthentication/config/scenarios/names.ts
JakubKorytko: unfortunately, added a comment why

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/components/MultifactorAuthentication/config/scenarios/names.ts#L1-L11

src/libs/API/types.ts
JakubKorytko: good catch, changed!

src/libs/API/types.ts
JakubKorytko: done

src/libs/API/types.ts
JakubKorytko: done

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: chaanged

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/libs/MultifactorAuthentication/Biometrics/Challenge.ts#L57-L60

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/libs/MultifactorAuthentication/Biometrics/Challenge.ts#L80-L83

changed!

src/libs/MultifactorAuthentication/Biometrics/KeyStore.ts
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Configures secure store options based on the key type and provided options.
+ * Private keys require additional authentication and are protected from updates.
+ */
+const secureStoreOptions = (key: string, KSOptions?: MultifactorKeyStoreOptions): SecureStoreOptions => {
+    const isPrivateKey = key.endsWith(VALUES.KEY_ALIASES.PRIVATE_KEY);
+
+    return {
+        failOnUpdate: isPrivateKey,
+        requireAuthentication: isPrivateKey,
```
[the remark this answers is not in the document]
JakubKorytko: done

src/libs/MultifactorAuthentication/Biometrics/Observer.ts
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Registers a callback function for a specific event ID.
+     */
+    registerCallback: (id: string, callback: () => unknown) => {
```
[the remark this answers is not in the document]
JakubKorytko: added

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: fixed!

src/libs/MultifactorAuthentication/Biometrics/helpers.ts
JakubKorytko: yeah, changed

https://github.com/Expensify/App/blob/f04086c8afe62d1299a4ec981d1026d41b73faef/src/libs/MultifactorAuthentication/Biometrics/helpers.ts#L113-L118

src/libs/actions/MultifactorAuthentication.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable rulesdir/no-api-side-effects-method */
```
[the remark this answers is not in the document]
JakubKorytko: added

src/libs/actions/MultifactorAuthentication.ts
```diff
@@ -0,0 +1,59 @@
...
+ * - To avoid storing challenge in the persistent memory for security reasons.
+ *
+ * - As there is a certain short time frame in which the challenge needs to be signed,
+ *   we should not delay the possibility to do so for the user.
+ *
+ * This is not a standard practice in the code base.
+ * Please consult before using this pattern.
+ */
+
+async function registerAuthenticationKey({keyInfo, validateCode}: MultifactorAuthenticationScenarioParameters['REGISTER-BIOMETRICS']) {
```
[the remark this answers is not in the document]
JakubKorytko: added though used `Log.hmmm`, reasoning [here](https://swmansion.slack.com/archives/C08CZDJFJ77/p1768927340072289?thread_ts=1768852688.606869&cid=C08CZDJFJ77)

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: ~~some context: the `value` name is used to match the `MultifactorAuthenticationPartialStatus` that is later used by the Context to create a `MultifactorAuthenticationStatus` - it contains all the data to render success/failure screen, show the auth type name etc. The name is super generic cause the Context do not use this value - it is only for modules so they can transfer information between each other - in this example between `KeyStore` and `MultifactorAuthenticationChallenge`. This may sound a little unclear as the Context is not added in this PR though I have designed it with having the Context in mind.~~

changed the structure to be more readable

https://github.com/Expensify/App/blob/68bd50f48f683fca25d6c6be938730dca3f95a22/src/libs/MultifactorAuthentication/Biometrics/Challenge.ts#L38-L50

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
rafecolton: At this stage of the project, we can pick the name we feel is best in code and don't need to change it in the doc.

src/pages/MultifactorAuthentication/ValidateCodePage.tsx
```diff
@@ -0,0 +1,208 @@
...
+        clearAccountMessages();
+    }, [account?.errors]);
+
+    // Reset formError when hasError changes
+    useEffect(() => {
+        if (!hasError) {
+            return;
+        }
+        setFormError({});
+    }, [hasError]);
```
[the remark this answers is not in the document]
JakubKorytko: not sure why is this flagged, this file was not touched in this PR, resolving

src/pages/MultifactorAuthentication/NotificationPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+    let headerTitle = translate(mockedConfigFailure.headerTitle);
+    let title = translate(mockedConfigFailure.title);
+    let description = translate(mockedConfigFailure.description);
+
+    if (isSuccessNotification) {
+        headerTitle = translate(mockedConfigSuccess.headerTitle);
+        title = translate(mockedConfigSuccess.title);
+        // TODO: Replace hardcoded 'FaceID' with the actual authentication type (e.g., 'FaceID', 'TouchID', 'Fingerprint')
+        // once the MFA context provides the auth method used. This will require adding authType to route params and to CONST.
+        description = translate(mockedConfigSuccess.description, {authType: 'FaceID'});
```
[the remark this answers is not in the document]
JakubKorytko: not sure why is this flagged, this file was not touched in this PR, resolving

src/pages/MultifactorAuthentication/ValidateCodePage.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+    // Refs
+    const inputRef = useRef<MagicCodeInputHandle>(null);
+    const resendButtonRef = useRef<MultifactorAuthenticationValidateCodeResendButtonHandle>(null);
+    const hasClearedInitialErrorsRef = useRef(false);
+
+    // Derived state
+    const hasError = !!account && !isEmptyObject(account?.errors);
+    const isValidateCodeFormSubmitting = AccountUtils.isValidateCodeFormSubmitting(account);
+    const shouldDisableResendCode = isOffline ?? account?.isLoading;
```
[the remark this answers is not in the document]
JakubKorytko: perhaps useful, but this file is not a part of this PR

src/components/TestToolMenu.tsx
```diff
@@ -26,9 +31,20 @@ function TestToolMenu() {
...
+    const navigateToBiometricsTestPage = singleExecution(
+        waitForNavigate(() => {
+            Navigation.navigate(ROUTES.MULTIFACTOR_AUTHENTICATION_BIOMETRICS_TEST);
+        }),
+    );
+
     // Check if the user is authenticated to show options that require authentication
     const isAuthenticated = useIsAuthenticated();
 
+    // Temporary hardcoded false, expected behavior: status fetched from the MultifactorAuthenticationContext
```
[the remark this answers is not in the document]
JakubKorytko: done exactly as you said

src/pages/MultifactorAuthentication/ValidateCodePage.tsx
```diff
@@ -0,0 +1,208 @@
...
+        if (!isValidValidateCode(inputCode)) {
+            setFormError({inputCode: 'validateCodeForm.error.incorrectMagicCode'});
+            return;
+        }
+
+        // Clear errors before submit
+        setFormError({});
+
+        // Temporary navigation, expected behavior: trigger submit from the MultifactorAuthenticationContext
+        // 'enable-biometrics' is a type of the MultifactorAuthentication prompt
```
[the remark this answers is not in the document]
JakubKorytko: not a part of this PR but it makes a lot of sense so added this one

https://github.com/Expensify/App/blob/853bb2a262e295d7c6b010125df817086fda2cba/src/pages/MultifactorAuthentication/ValidateCodePage.tsx#L150-L151

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
JakubKorytko: well, this actually matches the WebAuthn format I believe, `rpId` is actually a valid field
cc: @dariusz-biela to confirm

https://www.w3.org/TR/webauthn-1/#assertion-options

src/components/MultifactorAuthentication/config/scenarios/prompts.ts
JakubKorytko: The eslint-disable is a thing of a past ✨ 

https://github.com/Expensify/App/blob/61a39f6ebeaa4712b7eae4ee5d1590aa7fc0f5d7/src/components/MultifactorAuthentication/config/scenarios/prompts.ts#L5-L14

src/components/MultifactorAuthentication/config/types.ts
JakubKorytko: > What is this used for?

The comment actually was super poorly written, changed that

https://github.com/Expensify/App/blob/afb40e3bce764a192d59b6c9705eb4124af179d2/src/components/MultifactorAuthentication/config/types.ts#L69-L76

@DylanDylann fixed

https://github.com/user-attachments/assets/5e7f55ee-9f58-4bad-9c64-21293a1f5d1d

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: some context: the `value` name is used to match the `MultifactorAuthenticationPartialStatus` that is later used by the Context to create a `MultifactorAuthenticationStatus` - it contains all the data to render success/failure screen, show the auth type name etc. The name is super generic cause the Context do not use this value - it is only for modules so they can transfer information between each other - in this example between `KeyStore` and `MultifactorAuthenticationChallenge`. This may sound a little unclear as the Context is not added in this PR though I have designed it with having the Context in mind.

https://github.com/Expensify/App/blob/3ce42d337c4f6d34b87ae08e53586cf001b7a1ea/src/libs/MultifactorAuthentication/Biometrics/types.ts#L46-L66

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: I think this is NAB and my reasoning was exactly as Dariusz stated, resolving this one as it is a week old but feel free to re-open if you find this one important

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: https://github.com/Expensify/App/blob/68bd50f48f683fca25d6c6be938730dca3f95a22/src/libs/MultifactorAuthentication/Biometrics/Challenge.ts#L110-L116

no longer a thing!

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: `null` is ok here as well, but the functions return the `undefined` when there is nothing to return hence kept it that way to avoid casting `undefined` to `null` 

nothing unusual in the JS code

src/libs/MultifactorAuthentication/Biometrics/KeyStore.ts
JakubKorytko: yeah I made it optional as for the public key there is no prompt - refractored to make it required for private keys and empty for public keys

https://github.com/Expensify/App/blob/68bd50f48f683fca25d6c6be938730dca3f95a22/src/libs/MultifactorAuthentication/Biometrics/KeyStore.ts#L24-L34

https://github.com/Expensify/App/blob/68bd50f48f683fca25d6c6be938730dca3f95a22/src/libs/MultifactorAuthentication/Biometrics/types.ts#L137-L144

src/libs/MultifactorAuthentication/Biometrics/KeyStore.ts
JakubKorytko: you are correct, removed the `MultifactorAuthenticationStore` class

src/libs/MultifactorAuthentication/Biometrics/Observer.ts
JakubKorytko: oh this is actually used in the Context so the pages outside MFAContext (e.g. TestToolMenu) can react to e.g. keys revocation and change the registered label

will be much more clean in the 73 PR

src/libs/actions/MultifactorAuthentication.ts
JakubKorytko: there you go

https://github.com/Expensify/App/blob/68bd50f48f683fca25d6c6be938730dca3f95a22/src/libs/actions/MultifactorAuthentication.ts#L28-L38

src/libs/actions/MultifactorAuthentication.ts
JakubKorytko: there you go

https://github.com/Expensify/App/blob/68bd50f48f683fca25d6c6be938730dca3f95a22/src/libs/actions/MultifactorAuthentication.ts#L60-L71

src/components/MultifactorAuthentication/config/mapMultifactorAuthenticationNotifications.ts
JakubKorytko: changed the naming convention to "outcome"

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
JakubKorytko: it is used by the Context - it indeed uses `processScenario` to send the challenge butthen changes the BE response to object that is readable by the Context and can be provided to the user after translations etc. From the Context perspective though it does indeed takes care of the whole sending process - this will be much clearer in the 73 PR.

src/components/MultifactorAuthentication/config/types.ts
JakubKorytko: > I'm not sure if it's a common pattern to have all the types declared in their own file

Yeaah, pretty common, e.g. of such pretty unreadable but constantly used file

https://github.com/Expensify/App/blob/6f78a711b30aec42c6a4cb7b50a350c963635974/src/languages/types.ts#L1-L89

And count of only `types.ts` files (`types` directories not included etc):

image

I believe it is because of the TypeScript uses a lot of compilation and the actual type is super easy to read from the IDE but not so easy to read looking at the code itself:

image

src/components/MultifactorAuthentication/types.ts
JakubKorytko: resolving to keep discussion [here](https://github.com/Expensify/App/pull/79477#discussion_r2727958755) as it is the same topic

src/components/MultifactorAuthentication/config/scenarios/prompts.ts
JakubKorytko: same reason as here https://github.com/Expensify/App/pull/79477#discussion_r2709255903
will add a comment in the next commit

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
DylanDylann: I checked again, and this looks fine now

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
```diff
@@ -109,13 +109,22 @@ function createAuthenticatorData(rpId: string): Bytes {
...
+ *
+ * @param credentialRequestOptions Challenge object from server (must be AuthenticationChallenge format)
+ * @param privateKey ED25519 private key in hex format
+ * @param publicKey ED25519 public key in base64url format (used as rawId)
+ * @returns SignedChallenge with ED25519 signature
  */
-function signToken(credentialRequestOptions: MultifactorAuthenticationChallengeObject, privateKey: string): SignedChallenge {
-    const rawId: Base64URLString = Base64URL.encode(VALUES.KEY_ALIASES.PUBLIC_KEY);
-    const type = VALUES.ED25519_TYPE;
+function signToken(credentialRequestOptions: MultifactorAuthenticationChallengeObject, privateKey: string, publicKey: Base64URLString): SignedChallenge {
```
DylanDylann, line 119: I think we should only pass rpId and challenge instead of the full credentialRequestOptions. It would be more reusable.

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
dariusz-biela: Passkeys authorization uses the format you have.
Registration uses the one written by Dylan, but I will make an adapter for passkeys, so you dont need to change anything.

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: will take care of this tomorrow in the follow-up

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: will take care of this tomorrow in the follow-up

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
JakubKorytko: let's move it to https://github.com/Expensify/App/pull/79473

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
JakubKorytko: image

actually I have both defined so I think we are good to resolve?

dariusz-biela: thats great you can resolve this

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: changed that

https://github.com/Expensify/App/blob/2548bc7c321d3b92672e4c2a885e987f7e940f63/src/libs/MultifactorAuthentication/Biometrics/VALUES.ts#L16-L33

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: u are right, fixed

https://github.com/Expensify/App/blob/2548bc7c321d3b92672e4c2a885e987f7e940f63/src/libs/MultifactorAuthentication/Biometrics/VALUES.ts#L72-L114

DylanDylann (approved):

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
dariusz-biela: I think this format is good because we want to standardize the implementation of native biometrics and passkeys to make the implementation easier to review. And these data will be required for passkeys.

DylanDylann: Ok, so let's revisit it after passkey implementation

src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts
```diff
@@ -0,0 +1,25 @@
...
+    OUTCOMES: {
+        success: {
+            headerTitle: 'multifactorAuthentication.biometricsTest.biometricsTest',
+        },
+        failure: {
+            headerTitle: 'multifactorAuthentication.biometricsTest.biometricsTest',
+        },
+        outOfTime: {
+            headerTitle: 'multifactorAuthentication.biometricsTest.biometricsTest',
+        },
```
rafecolton, line 23: NAB I don't think this one can appear for the troubleshooting page

**Edit:** Though I suppose there's no reason why we couldn't support it 😄 kinda a fun easter egg

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,256 @@
...
+        KEY_SAVED: 'Key successfully saved in SecureStore',
+        UNABLE_TO_SAVE_KEY: 'Failed to save key in SecureStore',
+        UNABLE_TO_DELETE_KEY: 'Failed to delete key from SecureStore',
+        KEY_RETRIEVED: 'Key successfully retrieved from SecureStore',
+        KEY_NOT_FOUND: 'Key not found in SecureStore',
+        UNABLE_TO_RETRIEVE_KEY: 'Failed to retrieve key from SecureStore',
+    },
+} as const;
+
+/* eslint-disable @typescript-eslint/naming-convention */
```
rafecolton, line 72: Is this `eslint-disable` still needed? Asking per https://github.com/Expensify/App/pull/79477#discussion_r2727738837

src/libs/MultifactorAuthentication/Biometrics/Challenge.ts
```diff
@@ -0,0 +1,120 @@
...
+    public async sign(
+        accountID: number,
+        chainedPrivateKeyStatus?: MultifactorAuthenticationPartialStatus<string | null, true>,
+    ): Promise<MultifactorAuthenticationPartialStatus<boolean, true>> {
+        if (!this.challenge) {
+            return this.createErrorReturnValue(VALUES.REASON.CHALLENGE.CHALLENGE_MISSING);
+        }
+
+        if (isChallengeSigned(this.challenge)) {
+            return this.createErrorReturnValue(VALUES.REASON.CHALLENGE.CHALLENGE_ALREADY_SIGNED);
```
rafecolton, line 64: Reopening this one - I think it would be better to log a warning (and/or record in sentry?), so we can debug the unexpected behavior, but not fail the scenario for the use, since they have what they need to proceed (i.e. a signed challenge). I won't block on it, but it would be great to fix in follow-up.

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
```diff
@@ -109,13 +109,22 @@ function createAuthenticatorData(rpId: string): Bytes {
...
+    // rawId should be the base64url-encoded public key, serving as credential identifier
```
rafecolton, line 120: Noting that this is correct for biometric but for passkey the rawId will be the passkey ID and public key will be in the nested `publicKey` field. Perhaps that is fine if this file is only used for native biometrics and there will be a separate file for passkey.

@dariusz-biela also requested that I add a reminder comment for him to check whether or not the `Base64URLString` type is enforced 😄

src/libs/MultifactorAuthentication/Biometrics/SecureStore/index.ts
```diff
@@ -12,26 +13,32 @@ const SECURE_STORE_VALUES = {
         UNKNOWN: {
             CODE: SecureStore.AUTH_TYPE.UNKNOWN,
             NAME: 'Unknown',
+            MQ_VALUE: MQ_VALUES.AUTHENTICATION_METHOD.OTHER,
         },
         NONE: {
             CODE: SecureStore.AUTH_TYPE.NONE,
             NAME: 'None',
+            MQ_VALUE: MQ_VALUES.AUTHENTICATION_METHOD.OTHER,
         },
```
rafecolton, line 22: The `OTHER` type is reserved for the server, so these are not correct. For `UNKNOWN`, the FE should return a reasonable default based on the capabilities of the device (i.e. `BIOMETRIC_FACE` for iOS, `BIOMETRIC_FINGERPRINT` for android. If the type is None, I assume that means that the user did not actually authenticate, in which case we should throw an error.

src/libs/MultifactorAuthentication/Biometrics/SecureStore/index.ts
```diff
@@ -40,6 +47,7 @@ const SECURE_STORE_VALUES = {
         OPTIC_ID: {
             CODE: SecureStore.AUTH_TYPE.OPTIC_ID,
             NAME: 'OpticID',
+            MQ_VALUE: MQ_VALUES.AUTHENTICATION_METHOD.OTHER,
```
rafecolton: Hm, we should probably throw here because this is not supported, though perhaps instead we could fallback to `BIOMETRIC_FACE`, as it is very similar.

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -1,13 +1,256 @@
...
+        REGISTRATION_REQUIRED: REASON.BACKEND.REGISTRATION_REQUIRED,
+        AUTHENTICATION_REQUIRED: REASON.BACKEND.AUTHENTICATION_REQUIRED,
+        UNAUTHORIZED: REASON.BACKEND.UNAUTHORIZED,
+    },
+} as const;
+
+/**
+ * Maps API endpoints to their HTTP status codes and corresponding reason messages.
+ */
+const API_RESPONSE_MAP = {
```
rafecolton, line 85: I think this will end up being refactored. NAB for this PR, just noting - cc @dariusz-biela

src/components/MultifactorAuthentication/config/scenarios/BiometricsTest.ts
dariusz-biela: it will be changed with redesign of error handling

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
dariusz-biela: here it is required: Disables ESLint rule because it throws an error when a key is a number like 401.

src/libs/MultifactorAuthentication/Biometrics/ED25519/index.ts
dariusz-biela: I checked, and it's used correctly.

rafecolton (approved):

src/libs/MultifactorAuthentication/Biometrics/SecureStore/index.ts
JakubKorytko: done in https://github.com/Expensify/App/pull/79473

src/libs/MultifactorAuthentication/Biometrics/SecureStore/index.ts
JakubKorytko: done in https://github.com/Expensify/App/pull/79473

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: removed in https://github.com/Expensify/App/pull/79473

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: removed in https://github.com/Expensify/App/pull/79473

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
JakubKorytko: the name was awful, renamed to:

https://github.com/Expensify/App/blob/bab79b461b499777527e9c785bd692fd7644ce5b/src/libs/MultifactorAuthentication/Biometrics/VALUES.ts#L35

Issue #78384, opened 2025-12-23
[Personal card import] [R1] Existing personal cards on Wallet


amyevans:
**Project:** Personal card import in NewDot

**Project tracking issue:** https://github.com/Expensify/Expensify/issues/555211

[**Design doc section**](https://docs.google.com/document/d/1R3XaMT0yVja1NCbmOn9aWCA8pQAAnBXzvQCRMlZio-Q/edit?tab=t.0#bookmark=id.ws2jaq60g8m6)

Implement the linked section of the design doc!

amyevans:
Not overdue, implementation is starting this week

PR #73230, opened 2025-10-22
Filter unreported expenses by owner to prevent showing member expenses to admin

abzokhattab:
### Fixed Issues

$ https://github.com/Expensify/App/issues/70639
PROPOSAL: https://github.com/Expensify/App/issues/70639#issuecomment-3297322226

### Tests

Precondition:

- Admin invites member to the workspace chat.

Steps:
1. [Member] Go to workspace chat.
2. [Member] Create an expense.
3. [Member] Go to expense report.
4. [Member] Click Report field > Remove from report.
5. [Admin] Go to workspace chat (workspace chat with yourself).
6. [Admin] Click + > Create report.
7. [Admin] On report preview, click Add expense > Add unreported expense.
8. Verify that the unreported expense from member in Step 4 doesn't appear in unreported expense list for admin.

### Offline tests

Same as tests 

### QA Steps

same as tests 
### PR Author Checklist

- [X] I linked the correct issue in the `### Fixed Issues` section above
- [X] I wrote clear testing steps that cover the changes made in this PR
    - [X] I added steps for local testing in the `Tests` section
    - [X] I added steps for the expected offline behavior in the `Offline steps` section
    - [X] I added steps for Staging and/or Production testing in the `QA steps` section
    - [X] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [X] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [X] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [X] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [X] I ran the tests on **all platforms** & verified they passed on:
    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
- [X] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [X] I verified there are no new alerts related to the `canBeMissing` param for `useOnyx`
[...]

PR #79562, opened 2026-01-14
Feat: existing personal cards

waterim:
### Details

This PR introduces personal cards to the Wallet page

### Fixed Issues

$ https://github.com/Expensify/App/issues/78384

### Tests
Prerequisites: You need to add testing personal card in the OldDot:
1. in Expensify Classic, go to Settings > Account > Wallet > Import Card/Bank
2. select Discover card
3. use the creds expensifydemo / #USD

Test:
1. Go to Account
2. Go to Wallet
3. Observe assigned cards
4. Personal card is displayed correctly

Test 2:
1. Go to Account
2. Go to Wallet
5. Press on the personal card to see details
6. Test "Mark transaction as reimbursable"
7. Check optimistic data is correct and API call is correct

### Offline tests

N/A

### QA Steps
Same as tests

### PR Author Checklist

    - [X] Android / native
    - [X] Android / Chrome
    - [X] iOS / native
    - [x] iOS / Safari
    - [X] MacOS / Chrome / Safari
    - [X] MacOS / Desktop
    - [X] I verified that the left part of a conditional rendering a React component is a boolean and NOT a string, e.g. `myBool && <MyComponent />`.
    - [X] I verified any copy / text that was added to the app is grammatically correct in English. It adheres to proper capitalization guidelines (note: only the first word of header/labels should be capitalized), and is approved by marketing by adding the `Waiting for Copy` label for a copy review on the original GH to get the correct copy.
    - [X] If we are not using the full Onyx data that we loaded, I've added the proper selector in order to ensure the component only re-renders when the data it is using changes
    - [X] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/styles/StyleUtils.js) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(themeColors.componentBG)`)
- [X] I have checked off every checkbox in the PR author checklist, including those that don't apply to this PR.

### Screenshots/Videos

<details>

image
image

</details>


src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
```diff
@@ -0,0 +1,282 @@
...
+    const [personalDetails] = useOnyx(ONYXKEYS.PERSONAL_DETAILS_LIST, {canBeMissing: false});
+    const [session] = useOnyx(ONYXKEYS.SESSION, {canBeMissing: true});
+    const [cardList, cardListMetadata] = useOnyx(ONYXKEYS.CARD_LIST, {canBeMissing: true});
+
+    const card = cardList?.[cardID];
+    const cardBank = card?.bank ?? '';
+    const cardholder = personalDetails?.[card?.accountID ?? CONST.DEFAULT_NUMBER_ID];
+    const displayName = getDisplayNameOrDefault(cardholder);
+
+    const currentUserAccountID = (session as OnyxEntry<Session>)?.accountID ?? CONST.DEFAULT_NUMBER_ID;
```
ShridharGoel: `session` might have not loaded yet

src/pages/settings/Wallet/PersonalCardEditNamePage.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const {translate} = useLocalize();
+    const {inputCallbackRef} = useAutoFocusInput();
+    const styles = useThemeStyles();
+
+    const submit = (values: FormOnyxValues<typeof ONYXKEYS.FORMS.EDIT_WORKSPACE_COMPANY_CARD_NAME_FORM>) => {
+        updateAssignedCardName(cardID, values[INPUT_IDS.NAME], defaultValue);
+        Navigation.goBack(ROUTES.SETTINGS_WALLET_PERSONAL_CARD_DETAILS.getRoute(cardID));
+    };
+
+    const validate = (values: FormOnyxValues<typeof ONYXKEYS.FORMS.EDIT_WORKSPACE_COMPANY_CARD_NAME_FORM>): FormInputErrors<typeof ONYXKEYS.FORMS.EDIT_WORKSPACE_COMPANY_CARD_NAME_FORM> => {
```
ShridharGoel: Should we have a dedicated Onyx form key for personal cards?

waterim: Hm, yes, will add it

joekaufmanexpensify:
>Card name doesn't show the existing name in the input field, when it has the default name.

@ShridharGoel did the cardhave a default name for you? For me it was blank. 

image

joekaufmanexpensify:
### BUG: Last updated doesn't update after scraping card 

I checked with a company card in NewDot and it updates right away. Shouldn't it here too? The connection isn't broken. 

https://github.com/user-attachments/assets/71402004-ee9d-400a-8622-3e94cfeb692e

joekaufmanexpensify:
### BUG: Unassigning card from NewDot throws API error

`unassignCard` command returned: `The card was not found.` [Logs](https://www.expensify.com/_devportal/tools/logSearch/#sort=asc&size=10000&query=request_id%3A%229c1a0de4ae4d783a-LAX%22%20timestamp%3A%5B2026-01-21T00%3A00%20TO%202026-01-22T23%3A59%5D&engine=victorialogs) cc @amyevans 

image

joekaufmanexpensify:
### Improvement: Let's remove transaction start date for CSV cards 

This wasn't in the design doc. But this isn't something selected in the flow in OldDot, and isn't needed because CSV cards only display transactions that are actively uploaded via a file. There's no "import date" to set. cc @amyevans @DylanDylann 

image

ShridharGoel:
> did the cardhave a default name for you? For me it was blank. 

Yes, it was "A's card" where A is the first name of the account.

src/CONST/index.ts
```diff
@@ -3042,6 +3042,8 @@ const CONST = {
     ],
 
     POLICY: {
+        // Used as a placeholder policyID for personal cards accessed from the wallet (not associated with any workspace)
+        EMPTY_POLICY_ID: '0',
```
amyevans:
```suggestion
```
This looks unused now

src/languages/de.ts
```diff
@@ -2167,6 +2167,8 @@ const translations: TranslationDeepObject<typeof en> = {
         suspiciousBannerTitle: 'Verdächtige Transaktion',
         suspiciousBannerDescription: 'Wir haben verdächtige Transaktionen auf Ihrer Karte festgestellt. Tippen Sie unten, um sie zu überprüfen.',
         cardLocked: 'Ihre Karte ist vorübergehend gesperrt, während unser Team das Konto Ihres Unternehmens überprüft.',
+        markTransactionsAsReimbursable: 'Mark transactions as reimbursable',
+        markTransactionsDescription: 'When enabled, transactions imported from this card are marked as reimbursable by default.',
```
amyevans: Reminder to myself: run the `Generate static translations` workflow

src/languages/es.ts
```diff
@@ -1884,6 +1884,8 @@ const translations: TranslationDeepObject<typeof en> = {
         suspiciousBannerTitle: 'Transacción sospechosa',
         suspiciousBannerDescription: 'Hemos detectado una transacción sospechosa en la tarjeta. Haz click abajo para revisarla.',
         cardLocked: 'La tarjeta está temporalmente bloqueada mientras nuestro equipo revisa la cuenta de tu empresa.',
+        markTransactionsAsReimbursable: 'Mark transactions as reimbursable',
+        markTransactionsDescription: 'When enabled, transactions imported from this card are marked as reimbursable by default.',
```
amyevans: Asked for translations in Slack: https://expensify.slack.com/archives/C01GTK53T8Q/p1769112404494449

src/libs/CardUtils.ts
```diff
@@ -812,6 +812,18 @@ function isPersonalCard(card?: Card) {
     return !!card?.fundID && card.fundID !== '0';
 }
 
+/**
+ * Check if the given card is a personal card from OldDot.
+ * These cards do not have a domainName and are owned by the current user.
+ */
+function isPersonalCardFromOldDot(card: Card | undefined, currentUserAccountID: number): boolean {
```
amyevans: Let's remove the references to OldDot. Personal cards added in NewDot vs OldDot should be indistinguishable.

amyevans:
>>Just retested in a new account and it was blank again. Is it supposed to be in the cardList in onyx? I don't see any name set there either.
>
>@joekaufmanexpensify Yes, it should be, with test discovery card it works properly

I see the logic in the PR is: use the custom card name if set, fall back to the cardholder's first name. So if your test account does not have a first name set, it will be blank by default. That said, based on the mocks I believe the expected default is the card program (e.g. `Amex Black`, `AA Miles`) for personal cards (otherwise if I had multiple personal cards, they would all default to "Amy's card"). Is that correct @joekaufmanexpensify?

amyevans:
>> Transactions reimbursable toggle adjustment doesn't persist after sign out
>
>Looks like its a BE issue, as optimistic data is working correctly and after sign in new data from BE is coming without reimbursable, maybe some params are wrong, checking 
>cc @amyevans 

The param is incorrect on the frontend here. The backend sets the `reimbursable` field on the card. We basically just need to find/replace `markTransactionsAsReimbursable` to `reimbursable` in this PR and we should be good to go.

src/libs/actions/Card.ts
```diff
@@ -237,6 +241,330 @@ function clearActivatedCardPin() {
...
+
+    const parameters: SetPersonalCardReimbursableParams = {
+        cardID,
+        reimbursable: markTransactionsAsReimbursable,
+    };
+
+    API.write(WRITE_COMMANDS.SET_PERSONAL_CARD_REIMBURSABLE, parameters, {optimisticData, finallyData, failureData});
+}
+
+function syncCompanyCard(cardID: number, lastScrapeResult?: number) {
```
amyevans: Let's rename this and similar instances across the codebase to not reference `Company`. The API command name is `SyncCard` and is dual purpose for both personal cards and company cards so let's just try to make the naming consistent with that, to reduce confusion.

ShridharGoel:
> otherwise if I had multiple personal cards, they would all default to "Amy's card"

That is what happens in company cards, so I think we should maintain parity. What do you think?

amyevans (commented):
Disclaimer: this is not a full code review from me. But it captures a few things that jumped out at me while testing and ad-hoc reviewing.

src/languages/de.ts
amyevans: Output here: https://github.com/Expensify/App/pull/79562#issuecomment-3786752050

waterim:
Thanks @amyevans will update the PR

joekaufmanexpensify:
>I see the logic in the PR is: use the custom card name if set, fall back to the cardholder's first name. So if your test account does not have a first name set, it will be blank by default. That said, based on the mocks I believe the expected default is the card program (e.g. Amex Black, AA Miles) for personal cards (otherwise if I had multiple personal cards, they would all default to "Amy's card"). Is that correct @joekaufmanexpensify?

Hmm, yeah good Q. In OldDot, updating the card name seems to update what we're referring to as the card number in NewDot now, yeah? If so, is this separate "card name" field is really being used for anything in NewDot? I don't see it on any transactions, etc. I wonder if it makes sense to have them be separate things really. If you update it in OldDot, that's what we show on transactions going forward. 

image
image

src/languages/es.ts
amyevans: Confirmed there:

Transactions from these cards sync automatically.
`es: Las transacciones de estas tarjetas se sincronizan automáticamente.`

Mark transactions as reimbursable
`es: Marcar transacciones como reembolsables`

`When enabled, transactions imported from this card are marked as reimbursable by default.`
es: Cuando está habilitado, las transacciones importadas de esta tarjeta se marcan como reembolsables por defecto.

joekaufmanexpensify:
>BUG: bank icon style is wrong (originally https://github.com/Expensify/App/pull/79562#issuecomment-3781279659)

Confirmed this is fixed 👍

src/languages/es.ts
amyevans: Reminder to update these translations too

ShridharGoel:
Bug: Nothing happens on clicking "Save" after selecting "From the beginning" or after changing the date.

https://github.com/user-attachments/assets/160ef5d1-309b-421d-ad51-3e19571f9078


https://github.com/user-attachments/assets/14fc5daa-3710-4ee5-9bb6-d8d70b4bc2d8

src/libs/CardUtils.ts
```diff
@@ -831,6 +831,25 @@ function isPersonalCard(card?: Card) {
     return !!card?.fundID && card.fundID !== '0';
 }
 
+/**
+ * Check if the given card is a personal card assigned to the current user.
```
tgolen: IMO these comments aren't valuable (same for the method below) and can be removed. The code is very clear what is being done and the methods are named well.

src/libs/PolicyUtils.ts
```diff
@@ -1652,13 +1652,13 @@ function getMostFrequentEmailDomain(acceptedDomains: string[], policy?: Policy)
...
 
-const getDescriptionForPolicyDomainCard = (domainName: string): string => {
+const getDescriptionForPolicyDomainCard = (domainName: string, policies?: OnyxCollection<Policy>): string => {
     // A domain name containing a policyID indicates that this is a workspace feed
     const policyID = domainName.match(CONST.REGEX.EXPENSIFY_POLICY_DOMAIN_NAME)?.[1];
     if (policyID) {
         // This will be fixed as part of https://github.com/Expensify/Expensify/issues/507850
         // eslint-disable-next-line @typescript-eslint/no-deprecated
-        const policy = getPolicy(policyID.toUpperCase());
+        const policy = getPolicy(policyID.toUpperCase(), policies);
```
tgolen, line 1709: We're trying to remove all calls to `getPolicy()` and the only reason that `getPolicy()` uses the second parameter is to lookup the policy matching `policyID`. Since the `policyID` is already known, you can replace the call to `getPolicy()` with `policies[`${ONYXKEYS.COLLECTION.POLICY}${policyID}`];`. As a bonus, you can also remove the `no-deprecation` comment

src/libs/CardUtils.ts
```diff
@@ -934,7 +953,9 @@ export {
     getCompanyCardFeed,
     getCompanyCardFeedWithDomainID,
     getEligibleBankAccountsForUkEuCard,
+    filterPersonalCards,
     isPersonalCard,
+    isUserAssignedPersonalCard,
```
tgolen: Since both of these methods are being exported, can you please add unit tests for them?

src/libs/PolicyUtils.ts
tgolen: Actually, I think there is a better approach here... because ideally, we wouldn't need to subscribe to the entire policy collection (which impacts performance).

What I suggest is to use a selector when connecting the policy collection that only returns the single policy needed for this method (eg. move the `domainName.match()` into the Onyx selector). Then, this method would have two parameters: `domainName` and `policy`.

If that doesn't make sense, let me know and I can provide further instructions.

src/pages/settings/Wallet/PaymentMethodList.tsx
```diff
@@ -176,26 +176,42 @@ function PaymentMethodList({
     const [bankAccountList = getEmptyObject<BankAccountList>(), bankAccountListResult] = useOnyx(ONYXKEYS.BANK_ACCOUNT_LIST, {canBeMissing: true});
     const [userWallet] = useOnyx(ONYXKEYS.USER_WALLET, {canBeMissing: true});
     const [privatePersonalDetails] = useOnyx(ONYXKEYS.PRIVATE_PERSONAL_DETAILS, {canBeMissing: true});
+    const [session] = useOnyx(ONYXKEYS.SESSION, {canBeMissing: true});
     const isLoadingBankAccountList = isLoadingOnyxValue(bankAccountListResult);
-    const [cardList = getEmptyObject<CardList>(), cardListResult] = useOnyx(ONYXKEYS.CARD_LIST, {selector: filterPersonalCards, canBeMissing: true});
+    const [cardList = getEmptyObject<CardList>(), cardListResult] = useOnyx(ONYXKEYS.CARD_LIST, {canBeMissing: true});
     const isLoadingCardList = isLoadingOnyxValue(cardListResult);
+    const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {canBeMissing: true});
```
tgolen: Hm, now that I look at this down here, I think I see the problem... It looks like you need the policy of each `for (const card of assignedCardsSorted) {` that is not an expensify card.

Would it be possible to still use a selector here so that instead of `allPoicies`, it returned only the policies necessary for each assigned card that isn't an expensify card?

src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+    const card = cardList?.[cardID];
+    const cardBank = card?.bank ?? '';
+    const cardholder = personalDetails?.[card?.accountID ?? CONST.DEFAULT_NUMBER_ID];
+    const displayName = getDisplayNameOrDefault(cardholder);
+    const isPersonalCard = !!(card && isUserAssignedPersonalCard(card, currentUserAccountID));
+    const reimbursableSetting = card?.reimbursable ?? true;
+    const isCSVImportedPersonalCard = !!(isPersonalCard && card && (card.bank === CONST.COMPANY_CARDS.BANK_NAME.UPLOAD || card.bank.includes(CONST.COMPANY_CARD.FEED_BANK_NAME.CSV)));
+
+    const handleUnassignCard = () => {
```
tgolen: Methods should be named for what they do, not what they handle. I suggest something like `removeCardFromUser()`.

src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
```diff
@@ -0,0 +1,262 @@
...
+    };
+
+    const updateCard = () => {
+        if (!card) {
+            return;
+        }
+        syncCard(card.cardID, card.lastScrapeResult);
+    };
+
+    const lastScrape = useMemo(() => {
```
tgolen: Won't React Compiler automatically memoize this? I think it will, so you can remove the `useMemo()`.

src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
```diff
@@ -0,0 +1,262 @@
...
+    if (!card && !isLoadingOnyxValue(cardListMetadata)) {
+        return <NotFoundPage />;
+    }
+
+    // If somehow a non-personal card is accessed via this route, show not found
+    if (card && !isPersonalCard) {
+        return <NotFoundPage />;
+    }
+
+    return (
```
tgolen, line 92: This is a really big render block. Can you try to break it down into smaller pieces?

src/pages/settings/Wallet/PersonalCardEditNamePage.tsx
```diff
@@ -0,0 +1,97 @@
...
+import type SCREENS from '@src/SCREENS';
+import INPUT_IDS from '@src/types/form/EditPersonalCardNameForm';
+import isLoadingOnyxValue from '@src/types/utils/isLoadingOnyxValue';
+
+type PersonalCardEditNamePageProps = PlatformStackScreenProps<SettingsNavigatorParamList, typeof SCREENS.SETTINGS.WALLET.PERSONAL_CARD_EDIT_NAME>;
+
+function PersonalCardEditNamePage({route}: PersonalCardEditNamePageProps) {
+    const {cardID} = route.params;
+    const [customCardNames, customCardNamesMetadata] = useOnyx(ONYXKEYS.NVP_EXPENSIFY_COMPANY_CARDS_CUSTOM_NAMES, {canBeMissing: true});
+    const [cardList] = useOnyx(ONYXKEYS.CARD_LIST, {canBeMissing: true});
```
tgolen: Use a selector to only return the single card that is needed.

src/pages/settings/Wallet/PersonalCardEditTransactionStartDatePage.tsx
```diff
@@ -0,0 +1,154 @@
...
+type DateOption = ValueOf<typeof CONST.COMPANY_CARD.TRANSACTION_START_DATE_OPTIONS>;
+type PersonalCardEditTransactionStartDatePageProps = PlatformStackScreenProps<SettingsNavigatorParamList, typeof SCREENS.SETTINGS.WALLET.PERSONAL_CARD_EDIT_TRANSACTION_START_DATE>;
+
+function PersonalCardEditTransactionStartDatePage({route}: PersonalCardEditTransactionStartDatePageProps) {
+    const {cardID} = route.params;
+
+    const {translate} = useLocalize();
+    const styles = useThemeStyles();
+
+    const [cardList] = useOnyx(ONYXKEYS.CARD_LIST, {canBeMissing: true});
```
tgolen: Same, use a selector.

src/types/onyx/Card.ts
```diff
@@ -72,6 +72,9 @@ type Card = OnyxCommon.OnyxValueWithOfflineFeedback<{
     /** Last updated time */
     lastScrape?: string;
 
+    /** Whether imported transactions should default to reimbursable (personal OldDot cards only) */
```
tgolen: I'm not sure what "OldDot cards" means. Can you clarify it more in the comment?

tgolen (dismissed):

src/types/onyx/Card.ts
waterim: Oh, need to remove it, thanks!

src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
waterim: What do you mean by this?

tgolen: For example, move the menu items into their own component called `PersonalCardDetailsHeaderMenu`.

waterim: Im just not getting the reason for splitting this screen

ShridharGoel:
Bug: Empty "Assigned cards" section shows when there's no card assigned.

Screenshot 2026-01-28 at 10 52 22 PM

src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
tgolen: Smaller components are easier to test, easier to review, are reusable, separate concerns, and result in fewer regressions. In contrast, large monolithic PRs are more difficult to maintain.

This React best practice also discusses similar concepts: https://github.com/Expensify/App/blob/main/.claude/agents/code-inline-reviewer.md#clean-react-patterns-1-favor-composition-over-configuration

waterim: Ah, sorry, my bad, i read it as "render BLOCK", like performance block or as its blocking the render :D 
Yes sure, I can split it a bit

tgolen: Ohhh, sorry for the misunderstanding! Thanks :D

src/libs/CardUtils.ts
```diff
@@ -831,6 +831,21 @@ function isPersonalCard(card?: Card) {
     return !!card?.fundID && card.fundID !== '0';
 }
 
+function isUserAssignedPersonalCard(card: Card | undefined, currentUserAccountID: number): boolean {
+    if (!card) {
+        return false;
+    }
+
+    return !card.domainName && card.accountID === currentUserAccountID;
+}
```
amyevans: Adding this function is confusing IMO, we should really just use `isPersonalCard()`. However the logic for that function is inverted right now 🫠 

I see it was added in https://github.com/Expensify/App/pull/76967 and there is [a discussion](https://github.com/Expensify/App/pull/76967#discussion_r2685929026) already about it being backwards. What would you think about applying the relevant changes from https://github.com/Expensify/App/pull/73230/files where they are fixing it to the PR here?

src/libs/actions/Card.ts
```diff
@@ -237,6 +241,330 @@ function clearActivatedCardPin() {
...
+                        pendingFields: {
+                            cardTitle: CONST.RED_BRICK_ROAD_PENDING_ACTION.UPDATE,
+                        },
+                        errorFields: {
+                            cardTitle: null,
+                        },
+                    },
+                },
+            },
+        },
```
amyevans, line 448:
```suggestion
```
In the backend we don't set `nameValuePairs` on a card in `cardList` for personal cards, so I don't think we want/need this here. Let me know if I'm missing something though!

src/libs/actions/Card.ts
```diff
@@ -237,6 +241,330 @@ function clearActivatedCardPin() {
...
+                    nameValuePairs: {
+                        pendingFields: {
+                            cardTitle: null,
+                        },
+                    },
+                },
+            },
+        },
+    ];
+
```
amyevans, line 473:
```suggestion
```
Same comment as above

src/libs/actions/Card.ts
```diff
@@ -237,6 +241,330 @@ function clearActivatedCardPin() {
...
+                        pendingFields: {
+                            cardTitle: null,
+                        },
+                        errorFields: {
+                            cardTitle: ErrorUtils.getMicroSecondOnyxErrorWithTranslationKey('common.genericErrorMessage'),
+                        },
+                    },
+                },
+            },
+        },
```
amyevans, line 491:
```suggestion
```

Same as above

src/pages/settings/Wallet/PaymentMethodList.tsx
```diff
@@ -217,19 +233,45 @@ function PaymentMethodList({
...
+
+                    let cardDescription;
+                    if (isUserPersonalCard) {
+                        cardDescription = lastFourPAN;
+                    } else if (lastFourPAN) {
+                        cardDescription = `${lastFourPAN} ${CONST.DOT_SEPARATOR} ${getDescriptionForPolicyDomainCard(card.domainName, allPolicies)}`;
+                    } else {
+                        cardDescription = getDescriptionForPolicyDomainCard(card.domainName, allPolicies);
+                    }
+                    // Personal cards from OldDot navigate to personal card details page
```
amyevans:
```suggestion
                    // Personal cards navigate to personal card details page
```

src/types/onyx/Card.ts
```diff
@@ -72,6 +72,9 @@ type Card = OnyxCommon.OnyxValueWithOfflineFeedback<{
     /** Last updated time */
     lastScrape?: string;
 
+    /** Whether transaction should be reimbursable */
```
amyevans:
```suggestion
    /** Whether transactions from the card should be marked reimbursable by default */
```

amyevans (changes requested):

src/pages/settings/Wallet/PersonalCardDetailsHeaderMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <ToggleSettingOptionRow
+                title={translate('cardPage.markTransactionsAsReimbursable')}
+                subtitle={translate('cardPage.markTransactionsDescription')}
+                shouldPlaceSubtitleBelowSwitch
+                switchAccessibilityLabel={translate('cardPage.markTransactionsAsReimbursable')}
+                isActive={reimbursableSetting}
+                onToggle={(isOn) => card && setPersonalCardReimbursable(card.cardID, isOn, reimbursableSetting)}
+                pendingAction={card?.pendingFields?.reimbursable}
+                errors={card?.errorFields?.reimbursable ?? undefined}
+                onCloseError={() => card && clearCardErrorField(card.cardID, 'markTransactionsAsReimbursable')}
```
ShridharGoel: `setPersonalCardReimbursable` stores errors under `errorFields.reimbursable`, but `onCloseError` clears `markTransactionsAsReimbursable`. This will lead to errors not getting cleared.

src/libs/CardUtils.ts
```diff
@@ -879,6 +887,23 @@ function isCardAlreadyAssigned(cardNumberToCheck: string, workspaceCardFeeds: On
...
+    if (!cardList) {
+        return false;
+    }
+
+    return Object.values(cardList).some(
+        (card) =>
+            CONST.EXPENSIFY_CARD.ACTIVE_STATES.includes(card.state ?? 0) &&
+            (isExpensifyCard(card) || !!card.domainName || isPersonalCard(card)) &&
+            card.cardName !== CONST.COMPANY_CARDS.CARD_NAME.CASH,
+    );
```
ShridharGoel, line 904: Should we first filter using `.filter((card) => isCard(card))` ?

src/libs/actions/Card.ts
```diff
@@ -237,6 +241,280 @@ function clearActivatedCardPin() {
...
+    ];
+
+    const parameters: UnassignCardParams = {
+        cardID,
+    };
+
+    API.write(WRITE_COMMANDS.UNASSIGN_CARD, parameters, {optimisticData, successData, failureData});
+}
+
+function updateAssignedCardName(cardID: string, newCardTitle: string, oldCardTitle?: string) {
```
ShridharGoel, line 430: This should also be setting `card.nameValuePairs.pendingFields`/`errorFields`

src/pages/settings/Wallet/PaymentMethodList.tsx
```diff
@@ -187,15 +197,27 @@ function PaymentMethodList({
...
             const assignedCards = Object.values(isLoadingCardList ? {} : (cardList ?? {}))
-                // Filter by active cards associated with a domain
-                .filter((card) => (!!card.domainName || card.bank === CONST.PERSONAL_CARD.BANK_NAME.CSV) && CONST.EXPENSIFY_CARD.ACTIVE_STATES.includes(card.state ?? 0));
+                // Include active Expensify cards, company cards (domain), and personal cards
+                .filter(
+                    (card) =>
+                        CONST.EXPENSIFY_CARD.ACTIVE_STATES.includes(card.state ?? 0) &&
+                        (isExpensifyCard(card) || !!card.domainName || isPersonalCard(card)) &&
+                        card.cardName !== CONST.COMPANY_CARDS.CARD_NAME.CASH,
+                );
```
ShridharGoel, line 206: Do we need to include `isCard(card)` check here ?

src/selectors/Card.ts
```diff
@@ -25,7 +25,7 @@ const filterCardsHiddenFromSearch = (cardList: OnyxEntry<CardList>): CardList =>
  * This selector keeps non-personal cards (fundID !== '0').
  */
 const filterPersonalCards = (cards: OnyxEntry<CardList>): CardList => {
```
ShridharGoel: Should we rename this to `filterNonPersonalCards` ?

src/libs/CardUtils.ts
```diff
@@ -829,12 +829,20 @@ function getCompanyCardFeed(feedWithDomainID: string | undefined): CompanyCardFe
...
  */
 function isPersonalCard(card?: Card) {
-    return (!!card?.fundID && card.fundID !== '0') || card?.bank === CONST.PERSONAL_CARD.BANK_NAME.CSV;
+    return !card?.fundID || card.fundID === '0' || card?.bank === CONST.PERSONAL_CARD.BANK_NAME.CSV;
+}
+
+/**
+ * Filter out personal (including cash) cards from the card list.
+ */
+function filterPersonalCards(cards: CardList | undefined): CardList {
```
ShridharGoel: Same question - shouldn't it be named `filterNonPersonalCards` ?

ShridharGoel (changes requested):

src/libs/CardUtils.ts
amyevans: Let's do `filterOutPersonalCards`? That would be consistent naming with the `filterOutDomainCards` function

src/libs/actions/Card.ts
amyevans: This was removed at my request in https://github.com/Expensify/App/pull/79562/commits/7fb34727ba590ab04673a86c98e7421c61806549 - I don't think we need it (?)

tests/unit/CardUtilsTest.ts
```diff
@@ -1451,4 +1452,138 @@ describe('CardUtils', () => {
...
+                    bank: CONST.COMPANY_CARD.FEED_BANK_NAME.MASTER_CARD,
+                    cardName: 'Personal Card 2',
+                    domainName: '',
+                    fraud: 'none',
+                    lastFourPAN: '3333',
+                    lastScrape: '',
+                    lastUpdated: '',
+                    state: 3,
+                    fundID: '200',
+                },
```
amyevans, line 1497: these examples need some refinement now given the update to the function's logic 😅 

E.g., let's name the ones with nonzero fundID's as "Company" cards, and the fundID null/zero ones as "Personal" cards so that it's less confusing

tests/unit/CardUtilsTest.ts
```diff
@@ -1451,4 +1452,138 @@ describe('CardUtils', () => {
...
+        });
+
+        it('should return empty object when no cards have fundID', () => {
+            const cardList: CardList = {
+                '1': {
+                    cardID: 1,
+                    accountID: 12345,
+                    bank: CONST.COMPANY_CARD.FEED_BANK_NAME.VISA,
+                    cardName: 'Company Card 1',
+                    domainName: 'expensify-policy17f617b9fe23d2f1.exfy',
```
amyevans: Same as above, let's just update the examples to be consistent with what we expect for personal cards - no domainName, cardName something like 'Personal Card 1' and 'Personal Card 2'

src/libs/actions/Card.ts
ShridharGoel: But then it wouldn't show the greyed-out UI when card name is updated offline

https://github.com/user-attachments/assets/1fa6cda3-e082-4500-ba73-f1f05dd1402e

amyevans: Got it, thanks for pointing that out! We could probably put the pendingFields / errorFields on `NVP_EXPENSIFY_COMPANY_CARDS_CUSTOM_NAMES` instead 🤔. But I don't feel strongly. If it's easier to have it on card.nameValuePairs that's fine

src/pages/settings/Wallet/PaymentMethodList.tsx
waterim: cardList is a record of cards, not sure we need it

amyevans (approved):
Looks good to me! 🎉

src/pages/settings/Wallet/PaymentMethodList.tsx
ShridharGoel: Yes, but `isCard` mostly checks that `cardID` and `bank` should be present. It is used in CardUtils for similar scenarios. But anyways, it shouldn't be a blocker so let's skip it.

ShridharGoel (approved):

amyevans (approved):

src/selectors/Card.ts
```diff
@@ -24,8 +24,8 @@ const filterCardsHiddenFromSearch = (cardList: OnyxEntry<CardList>): CardList =>
  * Personal cards have fundID === '0' or no fundID.
  * This selector keeps non-personal cards (fundID !== '0').
  */
-const filterPersonalCards = (cards: OnyxEntry<CardList>): CardList => {
-    return filterObject(cards ?? {}, (key, card) => isPersonalCard(card));
+const filterOutPersonalCards = (cards: OnyxEntry<CardList>): CardList => {
```
abzokhattab, line 27: I have a question about this one, i am a bit confused why we are duplicating this function here once and another time in the `cardUtils` file .. it looks like the one in the `cardUtils` is not used. Can I remove it? let me know what u think 

@waterim @ShridharGoel

src/pages/settings/Wallet/PersonalCardDetailsHeaderMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+                onClose={() => {
+                    if (!card) {
+                        return;
+                    }
+                    clearCardNameValuePairsErrorField(card.cardID, 'cardTitle');
+                }}
+            >
+                <MenuItemWithTopDescription
+                    description={translate('workspace.moreFeatures.companyCards.cardNumber')}
+                    title={customCardNames?.[cardID] ?? getDefaultCardName(cardholder?.firstName)}
```
mkhutornyi, line 89: This was overlooked. The description should be card name, **NOT** card number.
Issue: https://github.com/Expensify/App/issues/88555

src/pages/settings/Wallet/PersonalCardEditNamePage.tsx
```diff
@@ -0,0 +1,96 @@
...
+                style={[styles.flex1, styles.mh5]}
+                enabledWhenOffline
+                validate={validate}
+                shouldHideFixErrorsAlert
+            >
+                <InputWrapper
+                    InputComponent={TextInput}
+                    inputID={INPUT_IDS.NAME}
+                    label={translate('workspace.moreFeatures.companyCards.cardNumber')}
+                    aria-label={translate('workspace.moreFeatures.companyCards.cardNumber')}
```
mkhutornyi, line 84: Same here. This is card name edit page, **NOT** card number

src/pages/settings/Wallet/PersonalCardDetailsPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+            />
+            <ScrollView addBottomSafeAreaPadding>
+                <View style={[styles.walletCard, styles.mb3]}>
+                    <ImageSVG
+                        contentFit="contain"
+                        src={getCardIconSource()}
+                        pointerEvents="none"
+                        height={variables.cardPreviewHeight}
+                        width={variables.cardPreviewWidth}
+                    />
```
situchan, line 109: This page calls `getCardFeedIcon()` for the card art, but didn't use the Plaid-aware rendering (`getPlaidInstitutionIconUrl` → `PlaidCardFeedIcon`) that the wallet list in `PaymentMethodList.tsx` already had in the same PR.

Issue: [[Personal Card Import] Incorrect card art on card details for Plaid personal card](https://github.com/Expensify/App/issues/88241)

Issue #79452, opened 2026-01-13
[ECUK In-App 3DS] [App] Add local passkeys storage utils in Onyx


dariusz-biela:
Part of [Adding biometric 3DS authentication for Expensify Cards](https://github.com/Expensify/Expensify/issues/523107)

Implement part of [this section](https://docs.google.com/document/d/15il4CanenwHJUEdjqOuWvxCFABM_3uRhfbuNxC7q4xo/edit?tab=t.0#heading=h.eijdposwcfki) of the design doc.

Implement local-only passkeys persistence for the current browser/device context. These changes should include:

- Add a new Onyx collection key (e.g. `ONYXKEYS.PASSKEYS`)
  - Store items under `passkey_${userId}@${rpId}`

- Preserve local passkeys across logout
  - Add `ONYXKEYS.PASSKEYS` to `KEYS_TO_PRESERVE`

- Implement Onyx actions for local passkeys
  - Read local entry for `userId + rpId`
  - Write/upsert local entry
  - Remove local entry

- Add a reconcile util to sync local Onyx passkeys with backend `allowCredentials` before running the flow
  - Determine whether this exact browser/device has a usable passkey (gate “authorize by default”)
  - Remove local entries/credentials that were deleted on the backend
 

<details><summary>Issue Owner</summary>Current Issue Owner: @dariusz-biela</details>

dariusz-biela:
We plan to work on it, but the native biometrics issues need to be completed first.

rafecolton:
PR is merged and awaiting deployment

rafecolton:
This is done and deployed to prod!

PR #81409, opened 2026-02-04
 [No QA] feat: add local passkey credentials storage with Onyx

dariusz-biela:
### Explanation of Change
This PR adds infrastructure for storing WebAuthn/Passkey credentials locally in Onyx for Web/mWeb platforms.

#### Why device-bound passkeys?

While passkeys can technically be portable across devices (via iCloud Keychain, Google Password Manager, etc.), we intentionally limit them to a single device/browser for now.

#### Rationale:
- Full cross-device portability would require additional credential management and recovery mechanisms
- By keeping passkeys device-specific, we can reuse the same registration and authorization logic already implemented for biometrics
- This keeps the authentication flow simple, consistent, and easy to maintain

#### Implementation:
Onyx stores data locally using IndexedDB and does not synchronize with the server by default - which is exactly what we need.

ONYXKEYS.COLLECTION.PASSKEYS is added to KEYS_TO_PRESERVE to ensure passkey entries survive logout. They are only removed when the user clears local site data.

#### How it works

1. Registration: When a user registers a passkey, we store the credential ID and transport hints locally under passkey_${userId}
2. Authentication flow:
    - Client checks if it has a known passkey by looking up local storage
    - Client calls RequestBiometricChallenge, which returns the server's list of known credential IDs (allowCredentials)
    - If a local ID matches a server ID → passkey is valid, proceed with authentication
    - If no match (empty local list or all credentials revoked on backend) → redirect to register new passkey
3. Reconciliation: Before each auth attempt, reconcileLocalPasskeysWithBackend syncs local state with the server's allowCredentials list, removing any stale entries (e.g., after "revoke all" on backend)

#### Changes

- Add ONYXKEYS.COLLECTION.PASSKEYS collection key (format: passkey_${userId})
- Add PasskeyScope type to identify a user (userId)
- Add LocalPasskeyEntry Onyx type with credentials: PasskeyCredential[] including transport hints
- Add Passkey action module:
    - addLocalPasskeyCredential - stores credential after successful registration, throws on duplicate
    - deleteLocalPasskeyCredentials - resets credentials to empty array for user
    - reconcileLocalPasskeysWithBackend - syncs with server's allowCredentials
- Add comprehensive unit tests

### Fixed Issues

$ https://github.com/Expensify/App/issues/79452
PROPOSAL:

### Tests
 N/A

### Offline tests
 N/A

### QA Steps
 N/A

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari


src/types/onyx/LocalPasskeyEntry.ts
```diff
@@ -0,0 +1,27 @@
+/** Represents the transport method for a passkey credential */
+type PasskeyTransport = 'internal' | 'hybrid' | 'usb' | 'nfc' | 'ble';
```
DylanDylann: @dariusz-biela Could you explain why do we add 'hybrid' value?

src/types/onyx/LocalPasskeyEntry.ts
```diff
@@ -0,0 +1,27 @@
...
+    type: 'public-key';
+
+    /** Optional array of transport methods that can be used to communicate with the authenticator */
+    transports?: PasskeyTransport[];
+};
+
+/**
+ * Represents a locally stored passkey entry for a specific user and relying party combination.
+ * These entries are stored in Onyx to track which passkeys have been registered on this browser/device.
+ * The storage key format is: passkey_${userId}@${rpId}
```
DylanDylann: Why do we need rpId in the field name?

src/types/onyx/LocalPasskeyEntry.ts
```diff
@@ -0,0 +1,27 @@
+/** Represents the transport method for a passkey credential */
+type PasskeyTransport = 'internal' | 'hybrid' | 'usb' | 'nfc' | 'ble';
+
+/** Represents a single credential stored locally for a passkey */
+type PasskeyCredential = {
+    /** base64url-encoded credential ID */
+    id: string;
+
+    /** The type of credential (always "public-key" for WebAuthn) */
+    type: 'public-key';
```
DylanDylann: Please put this value in CONST

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+/** Backend returns simplified format without transports */
+type BackendCredential = {
+    id: string;
+    type: 'public-key';
+};
+
+type ReconcileLocalPasskeysWithBackendParams = {
+    userId: string;
+    rpId: string;
+    allowCredentials: BackendCredential[];
```
DylanDylann:
```suggestion
    backendCredentials: BackendCredential[];
```

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Sets passkey credentials in Onyx storage.
+ * We use Onyx.set() instead of Onyx.merge() because passkey entries contain an array of credentials
+ * that needs to be fully replaced, not merged. Using merge() would append to the array instead of replacing it.
+ */
+function setLocalPasskeyCredentials({userId, rpId, entry}: SetLocalPasskeyCredentialsParams): void {
+    Onyx.set(getPasskeyOnyxKey(userId, rpId), entry);
+}
+
```
DylanDylann: Please check the empty string before storing to Onyx

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+/** Deletes all passkey credentials for a user/rpId from Onyx storage */
+function deleteLocalPasskeyCredentials(userId: string, rpId: string): void {
+    Onyx.set(getPasskeyOnyxKey(userId, rpId), null);
+}
+
+/** Backend returns simplified format without transports */
+type BackendCredential = {
+    id: string;
+    type: 'public-key';
+};
```
DylanDylann:
```suggestion
type BackendCredential = Omit<PasskeyCredential, 'transports'>;
```

src/types/onyx/LocalPasskeyEntry.ts
dariusz-biela: I will update the transport list to the currently proposed standard https://www.w3.org/TR/webauthn-3/#enum-transport

usb
Indicates the respective [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) can be contacted over removable USB.

nfc
Indicates the respective [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) can be contacted over Near Field Communication (NFC).

ble
Indicates the respective [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) can be contacted over Bluetooth Smart (Bluetooth Low Energy / BLE).

smart-card
Indicates the respective [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) can be contacted over ISO/IEC 7816 smart card with contacts.

hybrid
Indicates the respective [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) can be contacted using a combination of (often separate) data-transport and proximity mechanisms. This supports, for example, authentication on a desktop computer using a smartphone.

internal
Indicates the respective [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) is contacted using a [client device](https://www.w3.org/TR/webauthn-3/#client-device)-specific transport, i.e., it is a [platform authenticator](https://www.w3.org/TR/webauthn-3/#platform-authenticators). These authenticators are not removable from the [client device](https://www.w3.org/TR/webauthn-3/#client-device).

src/types/onyx/LocalPasskeyEntry.ts
dariusz-biela: A passkey is always associated with an rpId, so I think it's worth using this value in Onyx as well.

In the future, this may allow us to create a separate passkey set with different permissions, or implement them for a different domain.

tests/actions/PasskeyTest.ts
```diff
@@ -0,0 +1,182 @@
...
+            const result = reconcileLocalPasskeysWithBackend({userId, rpId, backendPasskeyCredentials: [], localEntry: null});
+
+            expect(result).toEqual([]);
+        });
+
+        it('should return empty array when localEntry has no credentials', () => {
+            const result = reconcileLocalPasskeysWithBackend({userId, rpId, backendPasskeyCredentials: [], localEntry: {credentialIds: []}});
+
+            expect(result).toEqual([]);
+        });
```
JakubKorytko, line 136: NIT I think we could merge these cases in each test, there are a lot of "it should throw error / return empty if something is null/empty/undefined", with only one line, a test `should fail if the data is incomplete` or something similar with a few `except` and descriptions would be cleaner

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Reconciles local Onyx passkeys with backend allowCredentials.
+ * Removes local credentials that no longer exist on backend.
+ */
+function reconcileLocalPasskeysWithBackend({userId, rpId, backendPasskeyCredentials, localEntry}: ReconcileLocalPasskeysWithBackendParams): PasskeyCredential[] {
+    if (!userId || !rpId) {
+        throw new Error('userId and rpId are required to reconcile passkey credentials');
+    }
+    if (!localEntry || localEntry.credentialIds.length === 0) {
```
JakubKorytko:
```suggestion
    if (!localEntry?.credentialIds.length) {
```

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (matchedCredentials.length !== localEntry.credentialIds.length) {
+        if (matchedCredentials.length === 0) {
+            deleteLocalPasskeyCredentials(userId, rpId);
+        } else {
+            setLocalPasskeyCredentials({userId, rpId, entry: {credentialIds: matchedCredentials}});
+        }
+    }
+
+    return matchedCredentials;
```
JakubKorytko, line 84: prefer early return over nested if

```suggestion
    if (matchedCredentials.length === localEntry.credentialIds.length) {
        return matchedCredentials;
    }

    if (matchedCredentials.length === 0) {
        deleteLocalPasskeyCredentials(userId, rpId);
    } else {
        setLocalPasskeyCredentials({userId, rpId, entry: {credentialIds: matchedCredentials}});
    }
    
    return matchedCredentials;
```

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,92 @@
...
+/** Returns Onyx key: passkey_${userId}@${rpId} */
+function getPasskeyOnyxKey(userId: string, rpId: string): `${typeof ONYXKEYS.COLLECTION.PASSKEYS}${string}` {
+    return `${ONYXKEYS.COLLECTION.PASSKEYS}${userId}@${rpId}`;
+}
+
+type SetLocalPasskeyCredentialsParams = {
+    userId: string;
+    rpId: string;
+    entry: LocalPasskeyEntry;
+};
```
JakubKorytko:
```suggestion
type PasskeyIdentification = {
    userId: string;
    rpId: string;
}

type SetLocalPasskeyCredentialsParams = PasskeyIdentification & {
    entry: LocalPasskeyEntry;
};
```

and use `PasskeyIdentification` (or name it better if you want) for `ReconcileLocalPasskeysWithBackendParams`, `AddLocalPasskeyCredentialParams` as well

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,92 @@
...
+/** Deletes all passkey credentials for a user/rpId from Onyx storage */
+function deleteLocalPasskeyCredentials(userId: string, rpId: string): void {
+    if (!userId || !rpId) {
+        throw new Error('userId and rpId are required to delete passkey credentials');
+    }
+    Onyx.set(getPasskeyOnyxKey(userId, rpId), null);
+}
+
+/** Backend returns simplified format without transports */
+type BackendPasskeyCredential = Omit<PasskeyCredential, 'transports'>;
```
JakubKorytko, line 55: I think this is not looking good, why not the other way around?

```
type BackendPasskeyCredential = {
    id: string;
    type: typeof CONST.PASSKEY_CREDENTIAL_TYPE;
}

type PasskeyCredential = BackendPasskeyCredential & {
    transports?: PasskeyTransport[];
}
```

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,92 @@
...
+    localEntry: LocalPasskeyEntry | null;
+};
+
+/**
+ * Reconciles local Onyx passkeys with backend allowCredentials.
+ * Removes local credentials that no longer exist on backend.
+ */
+function reconcileLocalPasskeysWithBackend({userId, rpId, backendPasskeyCredentials, localEntry}: ReconcileLocalPasskeysWithBackendParams): PasskeyCredential[] {
+    if (!userId || !rpId) {
+        throw new Error('userId and rpId are required to reconcile passkey credentials');
```
JakubKorytko: Let's use `CONST.MULTIFACTOR_AUTHENTICATION.REASON.PASSKEYS` for these error, wdyt?

tests/actions/PasskeyTest.ts
```diff
@@ -0,0 +1,182 @@
...
+
+        it('should preserve transports from local credentials', () => {
+            // Backend doesn't store transports, only local storage does
+            const localCredentials: PasskeyCredential[] = [{id: 'cred-1', type: CONST.PASSKEY_CREDENTIAL_TYPE, transports: [CONST.PASSKEY_TRANSPORT.INTERNAL, CONST.PASSKEY_TRANSPORT.HYBRID]}];
+            const backendPasskeyCredentials: BackendPasskeyCredential[] = [{id: 'cred-1', type: CONST.PASSKEY_CREDENTIAL_TYPE}];
+
+            const result = reconcileLocalPasskeysWithBackend({userId, rpId, backendPasskeyCredentials, localEntry: {credentialIds: localCredentials}});
+
+            expect(result).toEqual([{id: 'cred-1', type: CONST.PASSKEY_CREDENTIAL_TYPE, transports: [CONST.PASSKEY_TRANSPORT.INTERNAL, CONST.PASSKEY_TRANSPORT.HYBRID]}]);
```
JakubKorytko:
```suggestion
            expect(result).toEqual(localCredentials)
```

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,92 @@
...
+function addLocalPasskeyCredential({userId, rpId, credential, existingEntry}: AddLocalPasskeyCredentialParams): void {
+    const existingCredentials = existingEntry?.credentialIds ?? [];
+    const credentialExists = existingCredentials.some((c) => c.id === credential.id);
+
+    if (credentialExists) {
+        const updatedCredentials = existingCredentials.map((c) => (c.id === credential.id ? credential : c));
+        setLocalPasskeyCredentials({userId, rpId, entry: {credentialIds: updatedCredentials}});
+    } else {
+        setLocalPasskeyCredentials({userId, rpId, entry: {credentialIds: [...existingCredentials, credential]}});
+    }
```
JakubKorytko: how about:

```suggestion
    const existingCredentials = existingEntry?.credentialIds ?? [];
    const credentialExists = existingCredentials.some((c) => c.id === credential.id);
    const updatedCredentials = existingCredentials.map((c) => (c.id === credential.id ? credential : c));
    const credentialIds = credentialExists ? updatedCredentials : [...existingCredentials, credential] ;
    
    setLocalPasskeyCredentials({userId, rpId, entry: {credentialIds}});
```

JakubKorytko (commented):
some nits, otherwise LGTM

tests/actions/PasskeyTest.ts
dariusz-biela: I understand your approach because it does shorten the code, but in my opinion, tests should tell you what went wrong so you don't even have to check the implementation, and if you have multiple expectations in one test, in case of an error you have a general message and in the details only information about the first expect that failed.

src/libs/actions/Passkey.ts
dariusz-biela: This is because the full type is the passkeys standard, and the backend changed it, so the current layout describes reality.

src/libs/actions/Passkey.ts
dariusz-biela: I will do the mapping in a higher abstraction in the next PR, this util is in libs so it should not have access to the MFA abstraction.

src/libs/actions/Passkey.ts
dariusz-biela: When I analyzed it again, we should get an error when we add the same passkey twice. I will change code.

src/libs/actions/Passkey.ts
dariusz-biela: I leave it as is because it is a matter of preference between shorter code and a descriptive condition.

dariusz-biela:
@DylanDylann When you have a moment, take another look at this PR

tests/actions/PasskeyTest.ts
```diff
@@ -0,0 +1,179 @@
...
+        it('should return correct key format', () => {
+            const key = getPasskeyOnyxKey('123', 'expensify.com');
+
+            expect(key).toBe(`${ONYXKEYS.COLLECTION.PASSKEYS}123@expensify.com`);
+        });
+    });
+
+    describe('addLocalPasskeyCredential', () => {
+        const userId = '123';
+        const rpId = 'expensify.com';
```
DylanDylann: Let's define these values in the above so they can be reused across all test cases

src/types/onyx/LocalPasskeyEntry.ts
DylanDylann: I think it's unnecessary for now, but this approach also works for me

DylanDylann (approved):
The rest looks fine to me

src/CONST/index.ts
```diff
@@ -436,6 +436,22 @@ const CONST = {
...
+     * @see https://www.w3.org/TR/webauthn-3/#enum-transport
+     */
+    PASSKEY_TRANSPORT: {
+        USB: 'usb',
+        NFC: 'nfc',
+        BLE: 'ble',
+        SMART_CARD: 'smart-card',
+        HYBRID: 'hybrid',
+        INTERNAL: 'internal',
+    },
```
rafecolton, line 461: NAB add comments on what these are and which ones we use.

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/**
+ * Sets passkey credentials in Onyx storage.
+ * We use Onyx.set() instead of Onyx.merge() because passkey entries contain an array of credentials
+ * that needs to be fully replaced, not merged. Using merge() would append to the array instead of replacing it.
+ */
+function setLocalPasskeyCredentials({userId, rpId, entry}: SetLocalPasskeyCredentialsParams): void {
+    if (!userId || !rpId) {
+        throw new Error('userId and rpId are required to store passkey credentials');
```
rafecolton: NAB is it common to use a generic error class like this?

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+    credential: PasskeyCredential;
+    existingEntry: LocalPasskeyEntry | null;
+};
+
+function addLocalPasskeyCredential({userId, rpId, credential, existingEntry}: AddLocalPasskeyCredentialParams): void {
+    const existingCredentials = existingEntry?.credentials ?? [];
+
+    if (existingCredentials.some((c) => c.id === credential.id)) {
+        throw new Error(`Passkey credential with id "${credential.id}" already exists for ${userId}@${rpId}`);
+    }
```
rafecolton: NAB if the values are the same, do we need to return or could we simply return early?

src/libs/actions/Passkey.ts
rafecolton: Is it an important distinction to prevent the implementation from breaking? It means we're required to maintain two separate types that are extremely similar. I understand that this way is technically correct, but for readability, I'm wondering that could be noted in a comment and they could be combined into a single type with `transports` being optional.

src/types/onyx/LocalPasskeyEntry.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Represents a locally stored passkey entry for a specific user and relying party combination.
+ * These entries are stored in Onyx to track which passkeys have been registered on this browser/device.
+ * The storage key format is: passkey_${userId}@${rpId}
+ */
+type LocalPasskeyEntry = {
+    /** Array of credentials registered for this user on this device/browser */
+    credentials: PasskeyCredential[];
+};
```
rafecolton: Any reason this needs to be a distinct type and cannot be flattened everywhere it is used?

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+    Onyx.set(getPasskeyOnyxKey(userId, rpId), {credentials: []});
+}
+
+/** Backend returns simplified format without transports */
+type BackendPasskeyCredential = Omit<PasskeyCredential, 'transports'>;
+
+type ReconcileLocalPasskeysWithBackendParams = PasskeyScope & {
+    backendPasskeyCredentials: BackendPasskeyCredential[];
+    localEntry: LocalPasskeyEntry | null;
```
rafecolton: Part of what's throwing me off is that `localEntry` is singular and `backendPasskeyCredentials` so it's not obvious that they are actually both lists of credentials. Flattening `LocalPasskeyEntry` and possibly combining both credential types would make this more obvious.

src/libs/actions/Passkey.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/**
+ * Reconciles local Onyx passkeys with backend allowCredentials.
+ * Removes local credentials that no longer exist on backend.
+ */
+function reconcileLocalPasskeysWithBackend({userId, rpId, backendPasskeyCredentials, localEntry}: ReconcileLocalPasskeysWithBackendParams): PasskeyCredential[] {
```
rafecolton: It's not obvious why this function returns a list of credentials vs being a `void`. Please add a comment explaining the return value and how the function is intended to be used.

src/types/onyx/LocalPasskeyCredentialsEntry.ts
```diff
@@ -0,0 +1,30 @@
+import type {ValueOf} from 'type-fest';
+import type CONST from '@src/CONST';
+
+/** Represents the transport method for a passkey credential */
+type PasskeyTransport = ValueOf<typeof CONST.PASSKEY_TRANSPORT>;
```
rafecolton, line 5: Curious if you agree with combining this file with `Passkey.ts`. I'm not sure it's big enough to justify its own file, especially if some of the types can be flattened.

src/libs/actions/Passkey.ts
rafecolton: Come to think of it, why is the server omitting transports? Should we include it in the server response? If we did, would that enable these types to be combined?

src/types/onyx/LocalPasskeyEntry.ts
rafecolton: I agree both that it seems like a premature optimization and that I don't feel strongly it needs to change 😄

tests/actions/PasskeyTest.ts
rafecolton: +1

tests/actions/PasskeyTest.ts
rafecolton: I don't feel strongly either way and think multiple tests are fine. However, I would love to see if/when/then comments added to each test 🙏

rafecolton (changes requested):
Looking good overall! Most of my comments regard potential opportunities to simplify by flattening/combining types

src/libs/actions/Passkey.ts
dariusz-biela: Of course, TypeScript can handle a single type, but this is a poor design because the developer then receives incorrect information about what he needs to handle. So we always want to be very strict with types.

Personally, I think we should keep them on the server, but I remember discussing this when we were creating the design doc, and we agreed that we didn't want to add this information to the database.

src/types/onyx/LocalPasskeyEntry.ts
dariusz-biela: You're right that this is premature optimization and adds some complexity.

I'll remove it from this PR.

src/CONST/index.ts
dariusz-biela: I've added additional descriptions.

A user can use any of them, as we don't block any, provided they have access to User Verification.

src/libs/actions/Passkey.ts
dariusz-biela: As far as I know, this is standard for our codebase.

src/libs/actions/Passkey.ts
dariusz-biela: We can, but I thought it was worth knowing if we want to add the same passkey twice, as this should never happen and could signal some problem in our code.

I plan to upload all these errors to Sentry so we can monitor them.

src/types/onyx/LocalPasskeyEntry.ts
dariusz-biela: This is the type of object we store in Onyx

dariusz-biela: The entire object is needed to update it

```ts
function setLocalPasskeyCredentials({userId, entry}: SetLocalPasskeyCredentialsParams): void {
    if (!userId) {
        throw new Error('userId is required to store passkey credentials');
    }
    Onyx.set(getPasskeyOnyxKey(userId), entry);
}
```

dariusz-biela: In the end, I decided to simplify the Entry type.
```ts
/**
 * Represents locally stored passkey credentials for a specific user.
 * These entries are stored in Onyx to track which passkeys have been registered on this browser/device.
 * The storage key format is: passkey_${userId}
 */
type LocalPasskeyCredentialsEntry = PasskeyCredential[];
```

src/types/onyx/LocalPasskeyCredentialsEntry.ts
dariusz-biela: we have to leave this file because it is the type of data stored in Onyx

And it is created in `src/types/onyx` for this reason

rafecolton (approved):

rafecolton:
When you get a chance, for the benefit of our future selves, please update the OP to reflect that rpId is no longer used for the onyx key 😄

Issue #77459, opened 2025-12-11
[Due for payment 2026-05-12] [$250] Screen Reader: Global: Cannot close the bottomsheet without selecting an option


olenamihno:
## Action Performed:
Precondition: User is logged in

1. Using VoiceOver on iOS or TalkBack on Android, open the app
2. Swipe right to focus any chat and double-tap to activate it
3. Swipe right to focus the "+" button and double-tap to activate it
4. Look for a way to dismiss the drawer dialog without selecting an option

This is a global issue affecting all dialogs in the app

## Expected Result:
All dialogs should have a method to dismiss them other than tapping outside them.
Screen readers should be provided with a Close (x) or drawer handle button to ensure they can dismiss the dialog without having to select an option.

## Actual Result:
Drawer dialogs do not expose any dismissible option for screen reader users.
Users cannot exit the drawer without switching to touch or selecting an option.

## Area issue was found in:
Chat

## Failed WCAG checkpoints
WCAG 4.1.2 Name, Role, Value
https://www.w3.org/WAI/WCAG22/Understanding/name-role-value

## User impact:
Providing role, state, and value information on all user interface components enables compatibility with assistive technology, such as screen readers, screen magnifiers, and speech recognition software, used by people with disabilities.

## Suggested resolution:
Implement an accessible Close (x) or drawer handle button to ensure screen reader users can dismiss dialogs.
https://appt.org/en/docs/react-native/samples/accessibility-role
https://reactnative.dev/docs/accessibility#accessibilityrole

## Environment:
Apple iPhone 14 Pro - iOS 18.7.1, Wi-Fi - Native (No mobile browser)
Samsung Galaxy Note10 Plus - Android 12, H0T Mobile - Native (No mobile browser)

## Screenshots/Videos

<img width="1179" height="2429" alt="Image" src="https://github.com/user-attachments/assets/5d610a31-7ec7-47a2-81fd-5e18053c5ed8" />

https://github.com/user-attachments/assets/e97f7a90-2740-4b60-aae7-506e033802ad

**Version Number:** Production
**Reproducible in staging?:** NA
**Reproducible in production?:** Yes
**Email or phone of affected tester (no customers):** No
**Logs:** https://stackoverflow.com/c/expensify/questions/4856  NA
**Notes/Photos/Videos:** Any additional supporting documentation  Yes
**Expensify/Expensify Issue URL:** https://platform.applause.com/company/12102/products/36628/community-issues/7009088
**Issue reported by:** NA
[...]

PR #81536, opened 2026-02-05
Fix Screen Reader issue preventing bottom sheet from closing without selection

marufsharifi:
### Fixed Issues

$ #77459
PROPOSAL: [#77459 (comment)](https://github.com/Expensify/App/issues/77459#issuecomment-3789868373)

### Tests

1. Using VoiceOver on iOS or TalkBack on Android, open the app
2. Swipe right to focus on any chat and double-tap to activate it
3. Swipe right to focus the "+" button and double-tap to activate it
4. Dismiss the drawer dialog using the dismiss button at top of bottom sheet / modal. 

### Offline tests

Same as the tests.

### QA Steps

Same as the tests.

### PR Author Checklist

    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/61d77a5a-1922-440b-853c-1e7dda3a1415

</details>

<details>

https://github.com/user-attachments/assets/223e6e40-fe4f-4c27-8203-5ce6b37ce9dd

</details>

<details>

https://github.com/user-attachments/assets/0ee52fdd-51c7-4776-bd18-8da8738111e0

</details>

<details>

https://github.com/user-attachments/assets/bbf81dd6-58c3-4f54-b054-4eaa89d42c8f

</details>

<details>

Not applicable.

</details>


src/components/Modal/BaseModal.tsx
```diff
@@ -261,6 +269,38 @@ function BaseModal({
         ],
     );
 
+    const shouldShowBottomDockedDismissButton = isSmallScreenWidth && type === CONST.MODAL.MODAL_TYPE.BOTTOM_DOCKED && !!(onBackdropPress ?? onClose) && isScreenReaderEnabled;
```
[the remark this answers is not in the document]
marufsharifi: I don’t think this applies here. `shouldShowBottomDockedDismissButton` is a primitive boolean, so `useMemo` would not provide reference stability benefits; dependent hooks only react to value changes.

src/components/Modal/BaseModal.tsx
```diff
@@ -1,12 +1,14 @@
...
-import type {LayoutChangeEvent} from 'react-native';
+import type {GestureResponderEvent, LayoutChangeEvent} from 'react-native';
 // Animated required for side panel navigation
 // eslint-disable-next-line no-restricted-imports
 import {Animated, DeviceEventEmitter, View} from 'react-native';
 import ColorSchemeWrapper from '@components/ColorSchemeWrapper';
 import NavigationBar from '@components/NavigationBar';
+import {PressableWithoutFeedback} from '@components/Pressable';
 import ScreenWrapperOfflineIndicatorContext from '@components/ScreenWrapper/ScreenWrapperOfflineIndicatorContext';
 import useKeyboardState from '@hooks/useKeyboardState';
```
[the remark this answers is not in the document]
marufsharifi: This seems mis-targeted. The empty catch being referenced is in Accessibility hook logic, not in `BaseModal.tsx`, so this specific BaseModal comment is not applicable.

src/components/Modal/BaseModal.tsx
```diff
@@ -375,6 +415,19 @@ function BaseModal({
                             ref={ref}
                             fsClass={forwardedFSClass}
                         >
+                            {shouldShowBottomDockedDismissButton && (
+                                <PressableWithoutFeedback
+                                    ref={dismissButtonRef}
+                                    onPress={handleBackdropPress}
```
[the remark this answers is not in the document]
marufsharifi: In this codebase `PressableWithoutFeedback`/`GenericPressable` passes the ref through to the underlying `Pressable`, so `dismissButtonRef` is not ignored and focus targeting is valid.

src/libs/Accessibility/index.ts
```diff
@@ -8,9 +8,29 @@ type HitSlop = {x: number; y: number};
...
+        if (isScreenReaderEnabledAsync) {
+            isScreenReaderEnabledAsync()
+                .then((enabled) => {
+                    if (!isMounted) {
+                        return;
+                    }
+
+                    setIsScreenReaderEnabled(enabled);
+                })
+                .catch(() => {});
```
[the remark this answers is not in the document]
marufsharifi: Thanks, but this is intentionally a no-op catch. `isScreenReaderEnabled()` is only an initial best-effort read and can fail transiently on some platforms; we already subscribe to screenReaderChanged right after, which is the reliable source of truth. Logging this expected/non-blocking failure would add noise without improving behavior.

src/libs/Accessibility/index.ts
```diff
@@ -8,9 +8,29 @@ type HitSlop = {x: number; y: number};
 const useScreenReaderStatus = (): boolean => {
     const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false);
     useEffect(() => {
```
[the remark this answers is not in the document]
marufsharifi: as you said here:
> Wait - actually looking closer, the code IS correct. The guard is properly placed. Let me reconsider...

marufsharifi:
@truph01, kindly bump. thanks.

src/libs/Accessibility/index.ts
truph01, line 1: Could you briefly explain the changes you made in this file?

src/components/Modal/BaseModal.tsx
```diff
@@ -261,6 +271,46 @@ function BaseModal({
...
+    const shouldShowBottomDockedDismissButton = isSmallScreenWidth && type === CONST.MODAL.MODAL_TYPE.BOTTOM_DOCKED && !!(onBackdropPress ?? onClose) && isScreenReaderEnabled;
+
+    const initialFocusTarget = useMemo(() => {
+        if (!isWeb || !shouldShowBottomDockedDismissButton) {
+            return initialFocus;
+        }
+        return () => dismissButtonRef.current ?? document.body;
+    }, [initialFocus, isWeb, shouldShowBottomDockedDismissButton]);
+
+    useEffect(() => {
```
truph01: Why do we need the useEffect to focus on the dismiss button on the web here?

heyjennahay (commented):
Bug fix.  Product review not required

src/libs/Accessibility/index.ts
marufsharifi: Why I changed it:

1. Before, the hook always started with `false`.
2. It only changed when `screenReaderChanged` event fired.
3. If **TalkBack/VoiceOver** was already ON before the app opened, the event might not fire immediately.
4. So the app could think screen reader is OFF when it’s actually ON.

What I added:

1. On mount, do one quick check: `AccessibilityInfo.isScreenReaderEnabled()`.
2. Keep the existing event listener for later ON/OFF changes.
3. Use `isMounted` guard to avoid setting state after unmount.

Simple reason:
`We need correct screen-reader status immediately, not only after a future event.`

src/components/Modal/BaseModal.tsx
marufsharifi: Why that web `useEffect` exists:

1. `initialFocus` is a first attempt.
2. On web, modal elements can mount slightly later (portal/render timing).
3. On first open, the dismiss button may not exist yet when `initialFocus` runs.
4. Then focus can stay on the old button behind the modal.

What the effect does:

1. After modal opens, it retries focus with `requestAnimationFrame`.
2. As soon as dismiss button exists, it focuses it.
3. Stops retrying after a few tries.
4. Cancels scheduled frames on cleanup.

Simple reason:
`initialFocus = first try, useEffect = reliable fallback so focus always lands inside modal on web.`

truph01:
@marufsharifi Friendly bump!

marufsharifi:
@truph01 thanks for the review, I looked into the time-based approach, but I don’t think it’s the best option because:

- a `2000ms` delay means screen reader users cannot access dismiss immediately if they want to go back
- even a shorter delay is still a timing race and can vary with animation/render timing, device speed, and VoiceOver/TalkBack/Safari behavior
- if dismiss is added later instead of only becoming accessible later, it can change the accessibility order and break the expected “swipe back to dismiss” behavior

Our current approach is more deterministic because:

- it does not depend on guessing the right delay for different devices/platforms
- dismiss stays out of the accessibility flow until the first modal item actually receives focus
- once that happens, dismiss becomes available in the correct order, so swiping back reaches it as expected

I also verified this on the iOS app, and I attached a video below showing the expected behavior.



https://github.com/user-attachments/assets/2c2221ce-adc8-4fb3-b29b-18dd904cd7dc

marufsharifi:
This failure looks unrelated to this PR.

src/components/PopoverMenu.tsx
```diff
@@ -482,6 +514,114 @@ function BasePopoverMenu({
...
+
+        target.focus();
+        hasFocusedFirstItemOnCurrentOpenRef.current = true;
+        if (shouldCoordinateDismissAccessibility) {
+            setIsBottomDockedDismissAccessible(true);
+        }
+        return true;
+    }, [shouldCoordinateDismissAccessibility]);
+
+    const focusFirstMenuItemOnNative = useCallback(() => {
```
truph01: This function has a lot of duplicate logic existed in `focusFirstMenuItemOnWeb`. Please refactor it to follow the DRY principle.

truph01:
@marufsharifi Please resolve two comments above

src/components/PopoverMenu.tsx
```diff
@@ -311,10 +314,38 @@ function BasePopoverMenu({
...
+        }
+    }, [isVisible, shouldDeferDismissButtonAccessibility]);
+
+    useEffect(() => {
+        if (shouldDeferDismissButtonAccessibility) {
+            return;
+        }
+        setShouldEnableBottomDockedDismissAccessibility(true);
+    }, [shouldDeferDismissButtonAccessibility]);
+
```
truph01: @MelvinBot Since `shouldDeferDismissButtonAccessibility = isIOS || isMobileSafari()` depends only on the platform type and never changes during runtime, I’m wondering if these two `useEffect` hooks are redundant.

We already initialize the state with:

<pre><code>const [shouldEnableBottomDockedDismissAccessibility, setShouldEnableBottomDockedDismissAccessibility] =
    useState(!shouldDeferDismissButtonAccessibility);
</code></pre>

Because `shouldDeferDismissButtonAccessibility` is static, the initial state should already reflect the correct value. Given that, do we still need the `useEffect` hooks that update this state, or could they be safely removed?

truph01: @marufsharifi Could you please review [this analytic](https://github.com/Expensify/App/pull/81536#issuecomment-4062171878)? If you agree with the analysis, please remove the second `useEffect`.

Also, please add a brief comment explaining the purpose of the first `useEffect`.

src/components/PopoverMenu.tsx
```diff
@@ -311,10 +314,38 @@ function BasePopoverMenu({
...
+        if (isVisible) {
+            if (shouldDeferDismissButtonAccessibility) {
+                setShouldEnableBottomDockedDismissAccessibility(false);
+            }
+            return;
+        }
+        hasFocusedFirstItemOnCurrentOpenRef.current = false;
+        if (shouldDeferDismissButtonAccessibility) {
+            setShouldEnableBottomDockedDismissAccessibility(false);
+        }
```
truph01: @marufsharifi Please wrap it in `markFirstMenuItemUnfocused`

src/components/PopoverMenu.tsx
```diff
@@ -482,6 +514,116 @@ function BasePopoverMenu({
...
+
+        scheduleFocusFirstMenuItem();
+    }, [isSmallScreenWidth, onModalShow, scheduleFocusFirstMenuItem]);
+
+    useEffect(() => {
+        if (!isVisible || !isSmallScreenWidth || hasFocusedFirstItemOnCurrentOpenRef.current) {
+            return;
+        }
+        scheduleFocusFirstMenuItem();
+    }, [isVisible, isSmallScreenWidth, scheduleFocusFirstMenuItem]);
```
truph01: @MelvinBot  With this code change, `scheduleFocusFirstMenuItem();` can be called in two cases:

1. When `onModalShow={handleModalShow}` is triggered.
2. When `isVisible` is updated to `true`.

These two scenarios appear to represent the same transition (the modal becoming visible). Because of that, it seems like the function could be executed twice.

Do you think we should keep only one of these triggers to avoid redundant calls?

src/components/PopoverMenu.tsx
```diff
@@ -311,10 +314,38 @@ function BasePopoverMenu({
     const [enteredSubMenuIndexes, setEnteredSubMenuIndexes] = useState<readonly number[]>(CONST.EMPTY_ARRAY);
     const platform = getPlatform();
     const isWeb = platform === CONST.PLATFORM.WEB;
+    const isAndroid = platform === CONST.PLATFORM.ANDROID;
+    const isIOS = platform === CONST.PLATFORM.IOS;
+    const shouldDeferDismissButtonAccessibility = isIOS || isMobileSafari();
```
truph01: @marufsharifi  Why do we need to defer dismiss button accessibility only in `isIOS || isMobileSafari()` here? What about other platforms?

src/components/Modal/BaseModal.tsx
```diff
@@ -261,6 +269,9 @@ function BaseModal({
         ],
     );
 
+    const shouldShowBottomDockedDismissButton = isSmallScreenWidth && type === CONST.MODAL.MODAL_TYPE.BOTTOM_DOCKED && !!(onBackdropPress ?? onClose);
+    const shouldHideBottomDockedDismissFromAccessibility = isNativeIOS && shouldShowBottomDockedDismissButton && shouldEnableBottomDockedDismissAccessibility === false;
```
truph01: @marufsharifi  As part of the question https://github.com/Expensify/App/pull/81536/changes#r2936257218,  why do we need `isNativeIOS` check here? What about other platforms?

src/components/PopoverMenu.tsx
marufsharifi: I kept both triggers because they cover different timing windows. One runs from `onModalShow`, and the other covers the case where `isVisible` changes before the accessibility focus handoff is ready. The `hasFocusedFirstItemOnCurrentOpenRef` guard prevents duplicate real work, so keeping both is safer for this flow.

src/components/PopoverMenu.tsx
marufsharifi: I limited this to iOS because that is where the issue was reproduced. Other platforms were already following the expected focus order, so broadening this logic would add extra complexity and regression risk without improving the affected path.

src/components/Modal/BaseModal.tsx
marufsharifi: That check is intentional because the explicit accessibility-tree hiding in `BaseModal` is only needed on the native iOS path here. Other platforms keep the expected order with the existing first-item focus handoff, so they do not need the same native accessibility gating.

marufsharifi:
@truph01, thanks for the review. I've addressed the remaining comments, Could you please take another look when you have a chance? Thanks.

src/components/Modal/BaseModal.tsx
```diff
@@ -261,6 +269,11 @@ function BaseModal({
         ],
     );
 
+    const shouldShowBottomDockedDismissButton = isSmallScreenWidth && type === CONST.MODAL.MODAL_TYPE.BOTTOM_DOCKED && !!(onBackdropPress ?? onClose);
+    // The explicit accessibility-tree hiding is only needed on native iOS. Other platforms keep
+    // the expected swipe order with the existing first-item focus handoff.
+    const shouldHideBottomDockedDismissFromAccessibility = isNativeIOS && shouldShowBottomDockedDismissButton && shouldEnableBottomDockedDismissAccessibility === false;
```
truph01: @marufsharifi  What do you think about removing `isNativeIOS`, since `isNativeIOS` check is already handled in `PopoverMenu`:

```suggestion
    const shouldHideBottomDockedDismissFromAccessibility = shouldShowBottomDockedDismissButton && shouldEnableBottomDockedDismissAccessibility === false;
```

src/components/PopoverMenu.tsx
```diff
@@ -311,10 +314,38 @@ function BasePopoverMenu({
...
+
+    useEffect(() => {
+        // Reset the dismiss accessibility gating for each modal close before the next open.
+        isVisibleRef.current = isVisible;
+        if (isVisible) {
+            if (shouldDeferDismissButtonAccessibility) {
+                setShouldEnableBottomDockedDismissAccessibility(false);
+            }
+            return;
+        }
```
truph01: @marufsharifi  This logic is redundant since we already reset `shouldEnableBottomDockedDismissAccessibility` when the modal closes in line 346.

src/components/Modal/BaseModal.tsx
marufsharifi: I did not remove `isNativeIOShere` because this check keeps the native iOS behavior explicit inside `BaseModal` itself. If we remove it, the component would rely only `onPopoverMenu` to pass the prop correctly, which makes this path less defensive and easier to misuse later. Keeping it does not change the current behavior, but it keeps the platform-specific intent clear and safer.

truph01: I suggest removing `isNativeIOS` because it appears to be redundant here. `shouldEnableBottomDockedDismissAccessibility` should be sufficient:

<pre><code>const shouldHideBottomDockedDismissFromAccessibility =
    isNativeIOS && shouldShowBottomDockedDismissButton && shouldEnableBottomDockedDismissAccessibility === false;
</code></pre>

- If `isNativeIOS` is `false`, then `shouldEnableBottomDockedDismissAccessibility` is also `false`.
- If `isNativeIOS` is `true`, the result of `shouldHideBottomDockedDismissFromAccessibility` will depend on `shouldEnableBottomDockedDismissAccessibility`.

Given this behavior, `isNativeIOS` does not change the final outcome and can likely be removed from the condition.

marufsharifi: done
thanks

src/components/PopoverMenu.tsx
```diff
@@ -311,10 +314,35 @@ function BasePopoverMenu({
...
+    }, [shouldDeferDismissButtonAccessibility]);
+
+    useEffect(() => {
+        // Reset the dismiss accessibility gating when the modal closes.
+        isVisibleRef.current = isVisible;
+        if (isVisible) {
+            return;
+        }
+        markFirstMenuItemUnfocused();
+    }, [isVisible, markFirstMenuItemUnfocused, shouldDeferDismissButtonAccessibility]);
```
truph01: The dependencies can be updated to `[isVisible]` only.

src/components/PopoverMenu.tsx
```diff
@@ -482,6 +511,116 @@ function BasePopoverMenu({
...
+        if (!target) {
+            return false;
+        }
+
+        const sendAccessibilityEvent = AccessibilityInfo.sendAccessibilityEvent;
+        if (sendAccessibilityEvent && isAndroid) {
+            sendAccessibilityEvent(target, 'viewHoverEnter');
+        }
+
+        Accessibility.moveAccessibilityFocus(firstMenuItemRef);
```
truph01: Why do we need `viewHoverEnter` in case `isAndroid`? Doesn't `focus` work?

src/components/PopoverMenu.tsx
```diff
@@ -482,6 +511,116 @@ function BasePopoverMenu({
...
+        return focusFirstMenuItemOnNative();
+    }, [focusFirstMenuItemOnNative, focusFirstMenuItemOnWeb, isWeb]);
+
+    const scheduleFocusFirstMenuItemOnWeb = useCallback(() => {
+        const focusFirstMenuItemWithRetries = (retries = MAX_FIRST_MENU_ITEM_FOCUS_RETRIES) => {
+            if (!isVisibleRef.current || hasFocusedFirstItemOnCurrentOpenRef.current) {
+                return;
+            }
+
+            if (focusFirstMenuItemOnWeb()) {
```
truph01:
```suggestion
            if (focusFirstMenuItem()) {
```

src/components/PopoverMenu.tsx
marufsharifi: it weakens Android TalkBack behavior, `viewHoverEnter` is needed here for correct Android accessibility focus handoff

What happens if we apply it:
- Android native can focus dismiss first instead of the first row
- so it breaks the expected behavior

marufsharifi:
@truph01, main merged, but I can't request to build a new adhoc, could you please do it? thanks.

dubielzyk-expensify:
Running build 👍

marufsharifi:
@truph01, gentle bump. thanks.

marufsharifi:
> It works on the + floating action button but not on the filters on the reports page. Is there a way to do it so it works universally across bottom sheets/modals?

I think making this fully universal at the BaseModal level would be risky, because BaseModal can hide/show the dismiss button but it does not know what the correct first accessibility target should be for every modal. Some flows need the first list row, others need an input or another control, so a generic solution there could easily cause regressions in unrelated bottom sheets. The current approach is safer because it only applies the behavior where we explicitly know the right first target and can confirm the focus handoff correctly. If we want broader coverage, I think the safer next step is to extend the same pattern to other list-based bottom-docked flows, rather than making BaseModal guess for all modals.

arosiclair:
@marufsharifi can you summarize the problems that we're running into right now? We added the invisible dismiss button, but that caused an issue where it gets focused first instead of the normal first item in the modal?

src/hooks/useBottomDockedDismissAccessibility.ts
```diff
@@ -0,0 +1,260 @@
...
+    const isWeb = platform === CONST.PLATFORM.WEB;
+    const isAndroid = platform === CONST.PLATFORM.ANDROID;
+    const isIOS = platform === CONST.PLATFORM.IOS;
+    // Native iOS can announce dismiss before the first item unless it stays hidden until focus lands.
+    const shouldDeferDismissButtonAccessibility = isIOS;
+    const firstItemRef = useRef<RNView>(null);
+    const isVisibleRef = useRef(isVisible);
+    const hasFocusedFirstItemOnCurrentOpenRef = useRef(false);
+    const firstMenuItemFocusRetryTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+    const [shouldEnableBottomDockedDismissAccessibility, setShouldEnableBottomDockedDismissAccessibility] = useState(!shouldDeferDismissButtonAccessibility);
```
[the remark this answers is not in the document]
truph01: @marufsharifi Could you check this feedback?

marufsharifi:
@truph01 Could you please take another look when you have a chance? I addressed the comment. Thanks.

rushatgabhane:
@truph01 all yours

@marufsharifi please fix the conflicts in the mean time

truph01:
@MelvinBot  There is a problem with the current approach:

The focus sequencing. Every consumer must:

1. Call useBottomDockedDismissAccessibility
2. Wire firstItemRef to the correct child
3. Forward handleFirstItemFocus / handleModalShow
4. Pass shouldEnableBottomDockedDismissAccessibility back down to the modal
This creates a coupling loop — the modal renders the dismiss button but can't control its own accessibility state without the consumer telling it "focus has landed." Each new bottom-docked popover (PopoverMenu, FABPopoverMenu, DropdownButton) needs custom integration, and if someone adds a new one without knowing about this pattern, it breaks.

A more general approach could:

1. Have BaseModal / ReanimatedModal itself detect the first focusable child inside its content (e.g., query accessibilityRole="button" descendants, or accept a generic initialFocusRef prop)
2. Use a simple time-based delay at the modal level (the DropdownButton already does this with BOTTOM_DOCKED_DISMISS_ACCESSIBILITY_DELAY = 2500ms — simpler but less precise)
3. Or wrap the dismiss button visibility in a single useEffect inside BaseModal that listens for focus entering the modal content container via onAccessibilityEscape or focus events, removing the need for consumers to participate at all


What do you think?

truph01:
@marufsharifi Could you take a look at [this comment](https://github.com/Expensify/App/pull/81536#issuecomment-4228388716)?

As mentioned there, the original code changes look good—we only need to update `BaseModal` and `PopoverMenu`. However, to cover all cases, we would need to introduce additional changes across multiple consumers to make the feature work. This makes the current approach less generalizable.

Would you be open to considering an alternative approach?

marufsharifi:
@truph01 Thanks for the review; I’m checking above comments.

marufsharifi:
@truph01 Thanks again for the review and for raising this concern. I understand the concern about the current wiring, and I also explored simpler modal-level alternatives while working through this.

In testing, a purely time-based solution partially improved the native behavior by preventing dismiss from winning initial accessibility focus, but it did not solve the mobile web issue. On both iOS web and Android web, the modal could still open while focus remained on the trigger behind it unless focus was explicitly moved to a real target inside the modal.

That was the main limitation of moving this entirely into `BaseModal`: the modal itself does not know what the correct initial accessibility target is for every consumer. In some flows it is the first actionable row, while in the Reports filters on web the reliable target ended up being the visible heading. Because of that, some consumer-level wiring was still needed to make the behavior correct across platforms.

I also tried other approaches during this investigation:
- a generic timer-only approach: partially helped on native, but did not solve web because it does not perform focus handoff
- relying on the modal to infer the first target automatically: not reliable, since the correct target differs by flow
- visual reordering of dismiss: did not guarantee the expected accessibility order across native and web

For the Reports filters specifically, I ended up with a split approach because it was the only one I could verify on real devices across all affected platforms:
- on web, explicit focus movement was required so focus would not remain behind the modal
- on native, I did not find a reliable programmatic focus handoff for that flow across both iOS and Android, so delaying dismiss accessibility was the least risky fix there

So I agree the abstraction could be improved, but from the approaches I tested, a timer-only modal-level solution was not enough to solve the issue correctly on all platforms. If you have a more general approach in mind that still guarantees the correct behavior on iOS native, Android native, iOS web, and Android web, I’m definitely open to it.

truph01:
@marufsharifi The logic in `DropdownButton` appears very similar to what we already have in `useBottomDockedDismissAccessibility`. Could you clarify the purpose of introducing this new implementation?

Would it be possible to reuse or migrate to the existing hook instead?

marufsharifi:
@truph01 Thanks for the follow-up.

I introduced the filters-side logic originally because the existing hook was built around first-item focus for menu-style popovers, while the Reports filters needed heading focus on web and dismiss deferral on native.

Based on your suggestion, I updated this to reuse the shared hook instead of keeping a separate implementation in `DropdownButton`, by extending the hook to support a custom focus target and a timer-based dismiss mode for the filters flow.

truph01:
@marufsharifi Please fix the conflicts

arosiclair:
Hello @marufsharifi, this PR has been open for over two months now and you are not being very responsive. I see you are working on several other issues and that is probably the reason why. Please keep in mind Rule 1 and the [contributing guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md):

> Daily updates on weekdays are highly recommended. If you know you won’t be able to provide updates within 48 hours, please comment on the PR or issue stating how long you plan to be out so that we may plan accordingly. We understand everyone needs a little vacation here and there. Any issue that doesn't receive an update for 5 days (including weekend days) may be considered abandoned and the original contract terminated.

Please prioritize this issue and work with @truph01 to complete this job as soon as possible. Do not take any new jobs in the meantime. 

@truph01 please prioritize reviewing this as well. If we are unsure about anything, please tag me into the discussion so we can keep things moving along.

marufsharifi:
> Hello @marufsharifi, this PR has been open for over two months now and you are not being very responsive. I see you are working on several other issues and that is probably the reason why. Please keep in mind Rule 1 and the [contributing guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md):
> 
> > Daily updates on weekdays are highly recommended. If you know you won’t be able to provide updates within 48 hours, please comment on the PR or issue stating how long you plan to be out so that we may plan accordingly. We understand everyone needs a little vacation here and there. Any issue that doesn't receive an update for 5 days (including weekend days) may be considered abandoned and the original contract terminated.
> 
> Please prioritize this issue and work with @truph01 to complete this job as soon as possible. Do not take any new jobs in the meantime.
> 
> @truph01 please prioritize reviewing this as well. If we are unsure about anything, please tag me into the discussion so we can keep things moving along.

Thanks for the bump. I’ve been updating this PR regularly, usually every day. The scope also grew a bit from the original issue because the same accessibility handling needed to be extended to additional modals/popovers as I validated the fix more broadly.

For the latest conflicts, they mainly came from [#86297](https://github.com/Expensify/App/pull/86297), which refactored the search actions/filter bar area that this PR also touches. Since that PR was linked to several deploy blockers, I thought there was a chance it might be reverted, so I held off on resolving those conflicts immediately to avoid doing the same merge work twice.

Sorry for the delay btw, I was OOO yesterday. I’ll prioritize this and resolve the conflicts ASAP. Thanks again for the reminder.

marufsharifi:
@truph01 I resolved the conflicts, added the fix for the Search display bottom sheets, and re-tested the affected flows on my end. Could you please take another look when you have a chance? Thanks.

truph01:
@codex review

marufsharifi:
@truph01 I addressed the comment. Could you please take another look when you have a chance? Thanks.

truph01:
I asked a question here https://github.com/Expensify/App/issues/77459#issuecomment-4303968315

marufsharifi:
> I asked a question here [#77459 (comment)](https://github.com/Expensify/App/issues/77459#issuecomment-4303968315)

 @truph01, the IE has responded to your questions [here](https://github.com/Expensify/App/issues/77459#issuecomment-4323398156). I’m currently waiting for your confirmation, so please take a look when you have a chance, thanks.

truph01:
@marufsharifi Let's proceed with this one:

> So yeah, let's standardize with having dismiss be the first action of a dialog/modal then 👍

marufsharifi:
@truph01 The PR has been updated to follow the new direction: the dismiss action is now the first focused action for bottom-docked modals. I tested it across platforms and it works as expected. 
Could you please take another look when you have a chance?

truph01:
@codex review

src/components/Modal/BaseModal.tsx
```diff
@@ -311,6 +315,8 @@ function BaseModal({
     const isBottomDockedModalInLandscapeMode = type === CONST.MODAL.MODAL_TYPE.BOTTOM_DOCKED && isInLandscapeMode;
 
     const shouldWrapChildrenInScrollView = shouldWrapModalChildrenInScrollViewIfBottomDockedInLandscapeMode && isBottomDockedModalInLandscapeMode;
+    const shouldShowBottomDockedDismissButton = isSmallScreenWidth && type === CONST.MODAL.MODAL_TYPE.BOTTOM_DOCKED && !!(onBackdropPress ?? onClose);
+    const modalInitialFocus = shouldShowBottomDockedDismissButton ? () => bottomDockedDismissButtonRef.current : initialFocus;
```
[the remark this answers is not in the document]
marufsharifi: I tested the affected bottom-docked modal flows and could not reproduce a keyboard-focus regression here.

src/components/Modal/BaseModal.tsx
```diff
@@ -389,6 +395,19 @@ function BaseModal({
                             ref={ref}
                             fsClass={forwardedFSClass}
                         >
+                            {shouldShowBottomDockedDismissButton && (
+                                <PressableWithoutFeedback
+                                    ref={bottomDockedDismissButtonRef}
+                                    onPress={handleBackdropPress}
+                                    role={CONST.ROLE.BUTTON}
```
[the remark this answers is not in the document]
marufsharifi: I tested the affected bottom-docked modal flows and did not reproduce accidental dismissals from taps near the first item.

marufsharifi:
@truph01 I addressed the latest feedback as much as possible and re-checked the affected flows. Could you please take another look when you have a chance?

truph01:
@marufsharifi Could you update videos to align with the latest changes in this PR?

src/components/Modal/BaseModal.tsx
```diff
@@ -389,6 +395,19 @@ function BaseModal({
...
+                                    accessibilityRole={CONST.ROLE.BUTTON}
+                                    accessibilityLabel={translate('common.dismiss')}
+                                    sentryLabel="Modal-DismissDialog"
```
truph01: We should add `Modal-DismissDialog` to `CONST.SENTRY_LABEL`.

marufsharifi: Addressed

marufsharifi:
@truph01 thanks for the review, updated the videos and also addressed the comment
please do another review
thanks.

truph01 (approved):

arosiclair (approved):

src/components/Modal/BaseModal.tsx
bernhardoj, line 319: `bottomDockedDismissButtonRef` can be null when we open the modal and close it immediately, which [crashes the app](https://github.com/Expensify/App/issues/92415). (read the root cause [here](https://github.com/Expensify/App/issues/92415#issuecomment-4606134546))

PR #82038, opened 2026-02-10
[NO QA] stop generating ids with leading zeros

blimpich:
### Explanation of Change

Prevents us from creating ids with leading zeros.

Why? The old code `Math.floor(Math.random() * (CONST.MAX_64BIT_LEFT_PART + 1))` generated the left part as a random integer from 0 to 92233, so about 1 in 92,234 times left would be 0. When left = 0 and it gets concatenated with the middle and right strings (e.g., 0 + "3446715" + "2727472"), JavaScript coerces the number 0 to the string "0", producing an ID like "034467152727472" with a leading zero. The new code generates a random 64 bit number using [crypto.randomValues](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) instead, and makes sure it doesn't have leading zeros by casting it to BigInt and then back to a string.

Slack discussion here: https://expensify.slack.com/archives/C03TQ48KC/p1770751691775419

### Fixes Issues

$ https://github.com/Expensify/Expensify/issues/597251
PROPOSAL:

### Tests

### QA Steps

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/src/libs/NumberUtils.ts
+++ b/src/libs/NumberUtils.ts
@@ -1,40 +1,15 @@
-import CONST from '@src/CONST';
-
 /**
- * Generates a random positive 64 bit numeric string by randomly generating the left, middle, and right parts and concatenating them. Used to generate client-side ids.
+ * Generates a random positive 64-bit numeric string. Used to generate client-side IDs.
  *
- * @returns string representation of a randomly generated 64 bit signed integer
+ * @returns string representation of a randomly generated 64-bit signed integer
  */
 function rand64(): string {
-    // Max 64-bit signed:
-    // 9,223,372,036,854,775,807
-    // The left part of the max 64-bit number *+1* because we're flooring it.
-    const left = Math.floor(Math.random() * (CONST.MAX_64BIT_LEFT_PART + 1));
+    const buffer = new BigUint64Array(1);
+    crypto.getRandomValues(buffer);
 
-    let middle;
-    let right;
-
-    // If the left is any number but the highest possible, we can actually have any value for the middle part, because even if it's all `9`s, the final value will not overflow the maximum
-    // 64-bit number.
-    if (left !== CONST.MAX_64BIT_LEFT_PART) {
-        middle = Math.floor(Math.random() * CONST.MAX_INT_FOR_RANDOM_7_DIGIT_VALUE);
-    } else {
-        middle = Math.floor(Math.random() * (CONST.MAX_64BIT_MIDDLE_PART + 1));
-    }
-
-    // And unless both the left and middle parts were the maximums, the right part can be any value as well.
-    if (left !== CONST.MAX_64BIT_LEFT_PART || middle !== CONST.MAX_64BIT_MIDDLE_PART) {
-        right = Math.floor(Math.random() * CONST.MAX_INT_FOR_RANDOM_7_DIGIT_VALUE);
-    } else {
-        right = Math.floor(Math.random() * (CONST.MAX_64BIT_RIGHT_PART + 1));
-    }
-
-    // Pad the middle and right with zeros.
-    const middleString = middle.toString().padStart(7, '0');
-    const rightString = right.toString().padStart(7, '0');
-
-    // Strip leading zeros by converting through BigInt
-    return BigInt(left + middleString + rightString).toString();
+    // Right shift by 1 bit to ensure the number is positive (63-bit range)
+    const positiveValue = buffer[0] >> 1n;
+    return BigInt.asIntN(64, positiveValue).toString();
 }
 
 /**
--- a/src/libs/NumberUtils.ts
+++ b/src/libs/NumberUtils.ts
@@ -7,7 +7,7 @@ function rand64(): string {
     const buffer = new BigUint64Array(1);
     crypto.getRandomValues(buffer);
 
-    // Right shift by 1 bit to ensure the number is positive (63-bit range)
+    // eslint-disable-next-line no-bitwise
     const positiveValue = buffer[0] >> 1n;
     return BigInt.asIntN(64, positiveValue).toString();
 }
--- a/src/libs/NumberUtils.ts
+++ b/src/libs/NumberUtils.ts
@@ -1,40 +1,23 @@
-import CONST from '@src/CONST';
-
 /**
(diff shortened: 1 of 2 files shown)


tylerkaraszewski (changes requested):
> shifting the range to 1–92233,

Does it not shift it to `1–92234`, leaving the possibility that we are overflowing the maximum allowed?

Would it not be equally simple and more straightforward to just trim leading zeros at the end? I think that would be easier to follow.

But if not, I think we need to generate a random value up to `CONST.MAX_64BIT_LEFT_PART - 1` before adding 1 at the end.

But I think it's much easier to explain why we strip leading zeros just before returning at the bottom of the function.

src/libs/NumberUtils.ts
```diff
@@ -33,7 +33,8 @@ function rand64(): string {
     const middleString = middle.toString().padStart(7, '0');
     const rightString = right.toString().padStart(7, '0');
 
-    return left + middleString + rightString;
+    // Strip leading zeros by converting through BigInt
+    return BigInt(left + middleString + rightString).toString();
```
tylerkaraszewski: I'm approving this because I assume `BIgInt` is a thing that we are allowed to use here, but if we have `BigInt` available I think this whole function can be a one liner or very close to it, because the whole reason behind splitting the number like this is the lack of real 64-bit numbers in JS.

tylerkaraszewski (dismissed):

src/libs/NumberUtils.ts
blimpich: I believe BigInt should be fine here, but it doesn't give us an easy way to generate that random number still. We still need a way to generate the random number without overflowing. Is there anyone else we should bring this up with you think?

tylerkaraszewski: This also requires `crypto` but I think would replace this entire function:
```
function randomInt64() {
  const a = new BigUint64Array(1);
  crypto.getRandomValues(a);
  const positiveSignedOnly = a[0] >> 1n;
  return BigInt.asIntN(64, positiveSignedOnly).toString();
}
```

blimpich: Good suggestion! Updated 👍

tylerkaraszewski: We had the longest possible conversation about this in Spain when we first did it so even though it's a small change, I'd love to get a couple more eyes on it.

tylerkaraszewski (dismissed):

src/libs/NumberUtils.ts
```diff
@@ -1,39 +1,15 @@
...
-        right = Math.floor(Math.random() * CONST.MAX_INT_FOR_RANDOM_7_DIGIT_VALUE);
-    } else {
-        right = Math.floor(Math.random() * (CONST.MAX_64BIT_RIGHT_PART + 1));
-    }
-
-    // Pad the middle and right with zeros.
-    const middleString = middle.toString().padStart(7, '0');
-    const rightString = right.toString().padStart(7, '0');
+    const buffer = new BigUint64Array(1);
+    crypto.getRandomValues(buffer);
```
roryabraham: Pretty certain this won't work out of the box in Hermes: https://github.com/facebook/hermes/issues/915#issuecomment-1434901507

If you want to use the crypto API, I think https://github.com/margelo/react-native-quick-crypto is the best option, but it's an implementation of the Node.js crypto API, not the crypto Web API. So you'd need to use a sharded implementation:

**web:**
```ts
const buffer = new BigUint64Array(1);
crypto.getRandomValues(buffer);
```

**iOS/Android:**
```ts
import crypto from 'react-native-quick-crypto';

const buffer = new BigUint64Array(1);
crypto.randomFillSync(buffer);
```

Or find some other more portable implementation without the library 🤷🏼

roryabraham (changes requested):

src/libs/NumberUtils.ts
blimpich: Yeah I don't think we should be making this simple utility into sharded implementation. Lets just go with the previous implementation but make it so that we don't have leading zeros

blimpich: Reverted it to original implementation. Only change is using BigInt to strip leading zeros

```js
 // Strip leading zeros by converting through BigInt
 return BigInt(left + middleString + rightString).toString();
```

That should work right?

tylerkaraszewski: So that does work, but I still think we can simplify this if we have BigInt now.

The existing implementation uses Math.random() which is worse than crypto.getRandomValues(), but given that we're already using it, we can continue to use it here in a way that is simpler than what we have now.

What about this?
```
function rand64() {
  // Generate two bigInts that are each 32-bit random numbers.
  // We do 2 because Math.random() really only generates 53-bit numbers internally.
  const hi32 = BigInt(Math.floor(Math.random() * 0x100000000));
  const lo32 = BigInt(Math.floor(Math.random() * 0x100000000));

  // Combine the two into a single 64-bit value.
  const u64 = (hi32 << 32n) | lo32;

  // Leave the top bit blank so these are not negative.
  return (u64 >> 1n).toString();
}
```

blimpich: Seems good to me! Updated here: https://github.com/Expensify/App/pull/82038/changes/1e812d33d22167e32b7ae18b65092b6e4cc69050

tylerkaraszewski (commented):
I bet we can remove the constants for `MAX_INT_FOR_RANDOM_7_DIGIT_VALUE` etc as well.

tylerkaraszewski (approved):
So we made this much more work than the original one line change was but I think the code is better for it.

src/libs/NumberUtils.ts
roryabraham: > I still think we can simplify this if we have BigInt now

We do have BigInt in Hermes now (and I believe we didn't when this was first created)

roryabraham (approved):

Issue #82415, opened 2026-02-13
[Due for payment 2026-02-26] Reports page - Exposed filter for View, Status, Type shows a popover that unnecessarily has a scroll bar


izarutskaya:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.18-6
**Reproducible in staging?:** Yes
**Reproducible in production?:** Yes
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @shawnborton 
**Slack conversation** (hyperlinked to channel name): [#expensify_bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1770971958865709)

## Action Performed:
1. Go To https://staging.new.expensify.com/
2. Navigate to Reports page > check all filters

## Expected Result:
Exposed filter for View, Status, Type shouldn't show scroll bar.

## Actual Result:
Exposed filter for View, Status, Type shows a popover that unnecessarily has a scroll bar.

## Workaround:
Unknown

## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/76672063-5d43-4dda-8e45-29a909d7d531

https://github.com/user-attachments/assets/a663ca8d-9233-40c5-a5fa-7b1e6d75c349

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @garrettmknight</details>

Krishna2323:
Requested in ND.

PR #82431, opened 2026-02-13
Fix: Remove unnecessary scroll bar from selection list popover

zuolinayun:
## Summary
This PR fixes the issue where selection list popovers (View, Status, Type filters) show unnecessary scroll bars when the content exactly fits the container.

## Root Cause
The `getSelectionListPopoverHeight` function calculates the estimated list height without accounting for the bottom padding (`pb3` = 12px) that is applied to the SelectionList's content container in `BaseSelectionList.tsx`.

When the container height exactly matches the estimated height (e.g., 3 items × 52px = 156px for View filter), the actual content height exceeds the container due to the missing 12px padding, causing the scroll indicator to appear.

## Solution
Added `LIST_BOTTOM_PADDING` (12px) to the estimated list height calculation in `getSelectionListPopoverHeight` to account for the `pb3` padding applied to the SelectionList's content container.

## Testing
- Verified that View, Status, and Type filter popovers no longer show unnecessary scroll bars
- Confirmed that the fix doesn't affect other popover behaviors
- Tested on multiple screen sizes to ensure consistent behavior

$ #82415

### DIFF
--- a/src/styles/index.ts
+++ b/src/styles/index.ts
@@ -6117,10 +6117,11 @@ const dynamicStyles = (theme: ThemeColors) =>
         getSelectionListPopoverHeight: (itemCount: number, windowHeight: number, isSearchable: boolean) => {
             const SEARCHBAR_HEIGHT = isSearchable ? 52 : 0;
             const SEARCHBAR_PADDING = isSearchable ? 12 : 0;
+            const LIST_BOTTOM_PADDING = 12;
             const PADDING = 32;
             const GAP = 8;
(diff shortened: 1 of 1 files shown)


Krishna2323:
@zuolinayun Please don’t open PRs without being assigned to the issue first. Kindly go through the [contributing guide](https://github.com/Krishna2323/App/blob/main/contributingGuides/CONTRIBUTING.md) to better understand the assignment process before starting work on an issue.

Please close this PR.

Issue #79163, opened 2026-01-08
[$250] Handle more gracefully when reportAction deep-link is incorrect


mountiny:
Coming from [thread](https://expensify.slack.com/archives/C07HPDRELLD/p1767741656003689?thread_ts=1767033734.597649&cid=C07HPDRELLD)

### Problem

When the deeplink with specific report actionID links to action that is not present on the report anymore (or never was), we show the user generic not here page informing them that the comment is not to be found.

<img width="1138" height="962" alt="Image" src="https://github.com/user-attachments/assets/077908d9-39e5-4410-a4f1-4269e9e0b857" />

This is however very disruptive while in many cases we could show the report as normal.

### Solution

Show the report if the user has access to it and minimize the feedback to the user about the report action not to be found. 

____

I am not sure how to indicate this, but I feel like the feedback should be much more subtle (but there should be _some_ feedback) rather than full page not here page.
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022014811601391401375</li>
        <li>Upwork Job ID: 2014811601391401375</li>
        <li>Last Price Increase: 2026-02-13</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @parasharrajat</details>

samranahm:
## Please re-state the problem that we are trying to solve in this issue.

The current implementation shows `FullPageNotFoundView` with the message "The comment you're looking for cannot be found" which blocks the entire screen, even when the report is accessible.


## What is the root cause of that problem?

`ReportScreen.tsx` that determines when to show the "not found" page. Specifically:

1. The `shouldShowNotFoundLinkedAction` flag is set to `true` when:
   - A `reportActionID` is present in the URL
   - The action doesn't exist in the loaded report actions
   - Initial report actions have finished loading

2. When `shouldShowNotFoundLinkedAction` is `true`, the entire `FullPageNotFoundView` is shown which:
   - Blocks the entire report screen
   - Shows a full-page error instead of the report content
   - Doesn't check if the user has access to the report itself

The logic doesn't distinguish between:
- **Case A**: User has no access to the report → Should show full-page error
- **Case B**: User has access to the report but the specific action doesn't exist → Should show the report with subtle feedback

## What changes do you think we should make in order to solve the problem?

We should modify the behavior when a linked reportAction is not found but the user has access to the report:

[...]

PR #82511, opened 2026-02-15
[$250] Handle more gracefully when reportAction deep-link is incorrect

zlq500:
### Details
Coming from [Issue #79163](https://github.com/Expensify/App/issues/79163)

### Problem
When the deeplink with specific report actionID links to action that is not present on the report anymore (or never was), we showed the user generic not here page. This was disruptive while in many cases we could show the report as normal.

### Solution
Modified `withReportAndReportActionOrNotFound` HOC to:
- Show the report normally when user has access to it
- Display subtle inline indicator instead of full-page NotFoundPage
- Preserve NotFoundPage only when report itself does not exist

### Changes
1. **withReportAndReportActionOrNotFound.tsx**: 
   - Separated report existence check from reportAction existence check
   - Added `isReportActionNotFound` flag
   - Render inline indicator when action not found

2. **ReportActionNotFoundIndicator.tsx** (new):
   - Subtle inline banner with minimal styling
   - Non-disruptive user feedback

### Testing
- Verified report displays normally when deep-linked action not found
- Verified NotFoundPage still shows when report does not exist
- Verified indicator shows/hides appropriately

### Fixed Issues
$ https://github.com/Expensify/App/issues/79163

### DIFF
--- a/src/pages/inbox/report/ReportActionNotFoundIndicator.tsx
+++ b/src/pages/inbox/report/ReportActionNotFoundIndicator.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import {View} from 'react-native';
+import Text from '@components/Text';
+import useThemeStyles from '@hooks/useThemeStyles';
+import useLocalize from '@hooks/useLocalize';
+
+/**
+ * Subtle indicator shown when a deep-linked report action is not found.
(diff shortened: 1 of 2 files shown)


zlq500:
I have read the CLA Document and I hereby sign the CLA

Issue #71913, opened 2025-10-06
[Due for payment 2026-08-31] [Partial Payment x 3][Tracking] Deprecation of InteractionManager


mountiny:
Following up to react native 0.81 upgrade, the `InteractionManager` will be marked as deprecated https://github.com/Expensify/App/issues/65991

Let's:
- Make sure its marked as deprecated so no new instances are added
- Write down clear guidelines for what to use instead `InteractionManager`
    - Explain the options if any and which is better at which
- Plan how to systematically migrate to the new approach 

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

mountiny:
What I mean is basically try to already instruct people to prefer the other approached and educate them on it to avoid using the `InteractionManager` so we do not keep introducing more usecases of it (and add more work for us later).

I do not think we should hold the RN bump for it

roryabraham:
cheeky trick just adding it back with a patch, but I think that was a very clever way to unblock the RN upgrade and buy us time to figure this out.

war-in:
I played a bit with our InteractionManager usages and those are my findings so far

- in many cases removing `InteractionManager` without any additional changes was enough
  - AvatarCropModal - https://github.com/Expensify/App/pull/74278/commits/931937c917bce2488e24b96ba0b6b5018a04d03a
  - EmojiPickerMenu - https://github.com/Expensify/App/pull/74278/commits/a5abf57034b2a8632e97226da0cdbaec12cf17a9
- we also use `InteractionManager` to wait for RHP to close (or any other modal) and then run next navigation. That case is tricky because the right way to do it would be to listen for the [transitionEnd](https://reactnavigation.org/docs/stack-navigator/#transitionend) event emitted from react-navigation. But I found a number of issues in `react-navigation` about unreliable `transitionStart` and `transitionEnd` events (see [this](https://github.com/react-navigation/react-navigation/issues?q=is:issue%20state:open%20transitionEnd)).
Next thing is `InteractionManager` being removed from react-navigation in the next major release (v8) (see [PR](https://github.com/react-navigation/react-navigation/commit/95c0f186258ee55092503b7e6e0e08962ef6887d)). The commit says that the recommended way of deffering an action is to listen for `transitionStart/End` which brings hope that those events will work reliably this time.
[...]

PR #82587, opened 2026-02-16
Migration navigation from InteractionManager to TransitionTracker

blazejkustra:
### Explanation of Change

This PR introduces the **TransitionTracker** and migrates the first batch of `InteractionManager` usages in navigation code to use it.

**TransitionTracker** (`src/libs/Navigation/TransitionTracker.ts`) is a new module that explicitly tracks active transitions (navigation, modal, keyboard, focus) using a counted map. It exposes `startTransition`/`endTransition` to mark transition boundaries, and `runAfterTransitions` to queue callbacks that fire once all transitions are idle. A safety timeout (1s) auto-ends transitions that are never explicitly closed.

On top of TransitionTracker, existing APIs gain transition-aware options:

- **`Navigation.navigate`**, **`goBack`**, and **`dismissModal`** now accept `waitForTransition` (defer the navigation until ongoing transitions finish) and `afterTransition` (run a callback after the triggered transition completes). This replaces the old pattern of wrapping navigation calls in `InteractionManager.runAfterInteractions`.
- **`dismissModalWithReport`** uses `afterTransition` instead of the old `DeviceEventEmitter`-based `MODAL_EVENTS.CLOSED` callback pattern. The `MODAL_EVENTS.CLOSED` constant and its associated `DeviceEventEmitter` emissions in `BaseModal` and `RightModalNavigator` are removed.
- **`KeyboardUtils.dismiss`** now accepts `{ afterTransition, shouldSkipSafari }` instead of a bare boolean, and tracks keyboard show/hide as a transition type.
- **`ScreenLayout`** (`src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx`) is a new component wired into the web stack navigator via `screenLayout` prop. It listens to `transitionStart`/`transitionEnd` events to feed TransitionTracker.
- **`ReanimatedModal`** now calls `TransitionTracker.startTransition('modal')` / `endTransition('modal')` alongside the existing `InteractionManager` handles.

Concrete migration examples in this PR:

- `WorkspaceCategoriesSettingsPage` and `WorkspaceCreateTagPage` replace `Keyboard.dismiss()` + `InteractionManager` with `KeyboardUtils.dismiss({ afterTransition })`.
- `WorkspaceInviteMessageComponent`, `NewChatPage`, `TransactionReceiptModalContent`, and `Report/index.ts` replace `dismissModal({ callback })` with `dismissModal({ afterTransition })`.

### Fixed Issues
$ https://github.com/Expensify/App/issues/71913
PROPOSAL: 

### Tests

1. Replace Reciept:
- Tap on any expense.
- Tap on the receipt.
- Tap 3-dot menu.
- Tap Replace.
- Verify that attachment modal is closed as on production and then RHP is opened

2. Workspace Invite
- Create workspace
- Go to overview page and click Invite button
- On Confirm details page click 'Invite'
- Verify that modal is closed and user is navigated to Members page afterwards 

3. Create group
- Click on FAB > Start chat
- choose a user and click 'Add to group'
- On Group details page simply click 'Start group'
- Verify that modal is closed and user is navigated to that report afterwards

4. Navigate to a chat from FAB (RHP)
- Click on FAB > Start chat
- Choose a person (other than self and not the chat that is opened underneath)
- Click on that person
- Verify that modal is closed and user is navigated to that report afterwards

### Offline tests

N/A

### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos

<details>

Start new chat:

https://github.com/user-attachments/assets/850964c6-df4c-444f-80ff-1fb23cd105ca

Workspace invite:

https://github.com/user-attachments/assets/9d248b23-e787-47cc-8f9e-4169e8dd277e

Create group:

https://github.com/user-attachments/assets/a42191df-3406-4987-8cac-607846f7730e

Replace receipt:

https://github.com/user-attachments/assets/33f9d9e6-05fe-4413-91d5-6b8c30d8c07a

Keyboard behaviour:

https://github.com/user-attachments/assets/a38a761f-c9ba-4c80-a058-647ebdc86a0c

</details>

<details>

Start new chat:
https://github.com/user-attachments/assets/8c36c37f-d457-41ea-8c0c-8b916e129e88

Workspace invite:
https://github.com/user-attachments/assets/ee5d5813-4546-4cef-8464-b4f85261bea6

Create group:
https://github.com/user-attachments/assets/9af3067b-d93b-42c0-97a2-1ab7d03e7cfa

Replace receipt:
https://github.com/user-attachments/assets/99405a86-deb1-4f3b-b229-9fcd23ebfcf4

</details>


contributingGuides/migrations/InteractionManager/README.md
```diff
@@ -0,0 +1,78 @@
...
+This makes the migration non-trivial: you have to understand *what each call is actually waiting for* before you can pick the right replacement.
+
+## The approach
+
+**TransitionTracker** is the backbone. It tracks navigation transitions explicitly, so other APIs can hook into transition lifecycle without relying on a global queue.
+
+On top of TransitionTracker, existing APIs gain transition-aware callbacks:
+
+- Navigation methods accept `afterTransition` — a callback that runs after the triggered navigation transition completes
+- Navigation methods accept `waitForTransition` — the call waits for any ongoing transition to finish before navigating
```
war-in: do you mean all ongoing transitions?

contributingGuides/migrations/InteractionManager/README.md
```diff
@@ -0,0 +1,78 @@
...
+- Keyboard methods accept `afterTransition` — a callback that runs after the keyboard transition completes
+- `ConfirmModal` and similar modal APIs will accept `afterTransition` callback to defer work until the modal close transition finishes (will be implemented in a separate PR)
```
war-in: Would be great to see how you plan to do it. Could you show us the interface?

contributingGuides/migrations/InteractionManager/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+This makes the code self-descriptive: instead of a generic `runAfterInteractions`, each call site says exactly what it's waiting for and why.
+
+> **Note:** `TransitionTracker.runAfterTransitions` is an internal primitive. Application code should use the higher-level APIs (`Navigation`, `ConfirmModal`, etc.) rather than importing TransitionTracker directly.
```
war-in: We can add a rule for that. I also thought about adding a rule that will save us from new InteractionManager usages

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/index.tsx
```diff
@@ -62,6 +64,14 @@ function createPlatformStackNavigatorComponent<RouterOptions extends PlatformSta
                 sidebarScreen,
                 parentRoute,
                 persistentScreens,
+                // eslint-disable-next-line react/no-unstable-nested-components
```
war-in: why do we need that? Could you add a comment with an explanation?

adamgrzybowski (approved):
🎸

WojtekBoman (approved):
LGTM 🚀

contributingGuides/migrations/InteractionManager/README.md
blazejkustra: Yep! Will adjust the phrasing

contributingGuides/migrations/InteractionManager/README.md
blazejkustra: Let's do it in a separate PR, as it will add a lot of changes to this PR. I'll create an issue for it

war-in (approved):

ikevin127:
## Reviewer Checklist

    - [ ] Android: HybridApp
    - [ ] Android: mWeb Chrome
    - [ ] iOS: HybridApp
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari
    - [ ] I verified any copy / text shown in the product is localized by adding it to `src/languages/*` files and using the [translation method](https://github.com/Expensify/App/blob/4510fc76bbf5df699a2575bfb49a276af90f3ed7/src/components/LocaleContextProvider.tsx#L80)
    - [ ] I verified all numbers, amounts, dates and phone numbers shown in the product are using the [localization methods](https://github.com/Expensify/App/blob/4510fc76bbf5df699a2575bfb49a276af90f3ed7/src/components/LocaleContextProvider.tsx#L116-L123)
- [ ] I verified any variables that can be defined as constants (ie. in CONST.ts or at the top of the file that uses the constant) are defined as such

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

src/components/Modal/ReanimatedModal/index.tsx
```diff
@@ -102,6 +103,7 @@ function ReanimatedModal({
                 // eslint-disable-next-line @typescript-eslint/no-deprecated
                 InteractionManager.clearInteractionHandle(handleRef.current);
             }
+            TransitionTracker.endTransition();
```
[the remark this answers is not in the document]
ikevin127: @blazejkustra 🔴 This is a valid issue: **`endTransition()` Called Unconditionally on Unmount — Double-Decrement Risk**

A modal that unmounts while not in a transition (e.g., it was already fully open and the parent is unmounted) will call `endTransition()` without a matching `startTransition()`. This drives `activeCount` into a negative via `Math.max(0, …)` check, but also clears an unrelated timeout from the queue — specifically the first activeTimeout in the array, potentially short-circuiting a completely different transition that was in progress.

**Bug this causes**: Opening Report A's context menu (which starts a navigation transition) while a previous modal unmounts could silently complete your `afterTransition` callbacks too early — triggering a navigate before animations finish.

Suggested fix:

```ts
useEffect(
    () => () => {
        if (handleRef.current) {
            InteractionManager.clearInteractionHandle(handleRef.current);
            // Only end the transition if we actually started one
            TransitionTracker.endTransition();
        }
    },
    [],
);
```

☝️ The cleanup should only call `endTransition()` when handleRef.current is set, since handleRef.current is set alongside each `startTransition()` call.

src/libs/Navigation/TransitionTracker.ts
```diff
@@ -0,0 +1,100 @@
...
+    activeTimeouts.push(timeout);
+}
+
+/**
+ * Decrements the active transition count.
+ * Clears the corresponding auto-timeout since the transition ended normally.
+ * When the count reaches zero, flushes all pending callbacks.
+ */
+function endTransition(): void {
+    const timeout = activeTimeouts.shift();
```
ikevin127, line 86: **🔴 `endTransition` Uses FIFO Queue — Mismatched on Nested/Overlapping Transitions**

`startTransition` pushes to the end of the array; `endTransition` shifts from the front. This is FIFO, which assumes transitions complete in the same order they start - but modals can open and close in any order (e.g., two overlapping modals where the inner one closes first).

If **Modal B** closes before **Modal A**, it clears **Modal A**'s timeout (the oldest one), leaving **Modal B**'s timeout to fire after 1s, causing a spurious `decrementAndFlush()` that prematurely flushes callbacks while **Modal A** is still animating.

**Bug this causes**: A callback that should fire "after **Modal A** closes" fires while **Modal A** is still mid-animation, causing a janky navigation jump.

**Suggested fix**: Each `startTransition` should return an ID to pass back to `endTransition`, making the pairing explicit:

```ts
function startTransition(): string {
    const id = Math.random().toString(36).slice(2);
    activeCount += 1;
    const timeout = setTimeout(() => {
        activeTimeouts.delete(id);
        decrementAndFlush();
    }, MAX_TRANSITION_DURATION_MS);
    activeTimeouts.set(id, timeout); // Use a Map<string, timeout>
    return id;
}

function endTransition(id: string): void {
    const timeout = activeTimeouts.get(id);
    if (timeout !== undefined) {
        clearTimeout(timeout);
        activeTimeouts.delete(id);
    }
    decrementAndFlush();
}
```

ℹ️ All call sites would need to thread through the returned ID.

src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx
```diff
@@ -0,0 +1,42 @@
...
+        });
+        const transitionEndListener = navigation.addListener('transitionEnd', () => {
+            TransitionTracker.endTransition();
+        });
+
+        return () => {
+            transitionStartListener();
+            transitionEndListener();
+        };
+    }, [navigation, options.animation, route?.name]);
```
ikevin127, line 37: **🔴 `ScreenLayout` Calls `startTransition`/`endTransition` For Every Screen in the Stack**

Every screen in a `Stack` navigator gets its own `ScreenLayout` instance. However, `transitionStart`/`transitionEnd` are emitted on all screens in the navigator when any navigation occurs, not just the focused one. So if there are 3 screens in an RHP stack, a single navigation action triggers 3× `startTransition` and 3× `endTransition`, inflating `activeCount` by `3` instead of `1`.

**Bug this causes**: `afterTransition` callbacks won't fire until all 3 counts reach zero — which only happens if exactly 3 matching `endTransition` calls arrive. If even one screen is unmounted before its `transitionEnd` fires (common edge case), `activeCount` stays `> 0` and pending callbacks are held forever until the 1s safety timeout.

**Suggested fix**: Only track transitions from the focused (active/leaf) screen:

```ts
useLayoutEffect(() => {
    const unsubscribeStart = navigation.addListener('transitionStart', (e) => {
        // Only the target screen triggers tracking, not all screens in the stack
        if (e.target === navigation.getId()) {
            TransitionTracker.startTransition();
        }
    });
    // ...
}, [navigation, options.animation, route?.name]);
```

> [!tip]
> Alternatively, use a single global navigator-level listener rather than per-screen hooks.

src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx
```diff
@@ -0,0 +1,42 @@
...
+import type {PlatformStackNavigationOptions, PlatformStackNavigationProp} from './types';
+
+// screenLayout is invoked as a render function (not JSX), so we need this wrapper to create a proper React component boundary for hooks.
+function screenLayoutWrapper({navigation, ...rest}: ScreenLayoutArgs<ParamListBase, string, StackNavigationOptions | PlatformStackNavigationOptions, string>) {
+    return (
+        <ScreenLayout
+            // eslint-disable-next-line react/jsx-props-no-spreading
+            {...rest}
+            // The type cast is needed because useNavigationBuilder hardcodes the Navigation generic to `string`.
+            navigation={navigation as unknown as StackNavigationProp<ParamListBase>}
```
ikevin127, line 14: **🟡 NAB: Type Cast in `ScreenLayout.tsx` Must Be Addressed**

Per the review guidelines, **type casting is forbidden** in Expensify's codebase. This double-cast (`as unknown as`) is a 🚩. While the comment mentions that `useNavigationBuilder` hardcodes the generic, the proper fix is to use a correct type at the `screenLayoutWrapper` boundary rather than forcing a cast:

```ts
// Option: Use a union type rather than casting
function screenLayoutWrapper({navigation, ...rest}: ScreenLayoutArgs<ParamListBase, string, StackNavigationOptions | PlatformStackNavigationOptions, StackNavigationProp<ParamListBase> | PlatformStackNavigationProp<ParamListBase>>) {
    return (
        <ScreenLayout
            {...rest}
            navigation={navigation}
        />
    );
}
```

> [!tip]
> Or open a follow-up with the comment: `// TODO: Resolve type mismatch upstream and track it` — but do not silently suppress a double cast.

src/libs/Navigation/Navigation.ts
```diff
@@ -319,9 +320,16 @@ function navigate(route: Route, options?: LinkToOptions) {
...
-    linkTo(navigationRef.current, targetRoute, options);
-    closeSidePanelOnNarrowScreen(route);
+    const runImmediately = !options?.waitForTransition;
+    TransitionTracker.runAfterTransitions(() => {
+        const targetRoute = route.startsWith(CONST.SAML_REDIRECT_URL) ? ROUTES.HOME : route;
+        linkTo(navigationRef.current, targetRoute, options);
+        closeSidePanelOnNarrowScreen(route);
+
+        if (options?.afterTransition) {
+            TransitionTracker.runAfterTransitions(options.afterTransition);
```
ikevin127: **🟠 `runAfterTransitions` in `navigate()` Fires the `afterTransition` Callback While a New Transition Might Already Be Starting**

`linkTo` calls the **react-navigation** action synchronously. `transitionStart` fires shortly after (in the next JS task/microtask via the navigation event system). This means `TransitionTracker.runAfterTransitions(options.afterTransition)` is called before the new `ScreenLayout`'s listener has had a chance to call `startTransition()`.

If `activeCount` is `0` at that moment, `runAfterTransitions` fires the callback immediately instead of after the transition — defeating the entire purpose.

**Bug this causes**: `afterTransition` in `dismissModalWithReport` → `navigate(reportRoute)` would fire before the report screen has animated in.

**Suggested fix**: Use a requestAnimationFrame guard or a dedicated "navigationDidStart" hook that ensures `startTransition` runs before the callback is queued:

```ts
if (options?.afterTransition) {
    // Ensure we queue AFTER the navigation event loop has fired transitionStart
    requestAnimationFrame(() => {
        TransitionTracker.runAfterTransitions(options.afterTransition);
    });
}
```

src/libs/Navigation/Navigation.ts
```diff
@@ -710,32 +733,31 @@ function getTopmostSuperWideRHPReportID(state: NavigationState = navigationRef.g
...
-            });
-        }
+        TransitionTracker.runAfterTransitions(() => {
+            ref.dispatch({type: CONST.NAVIGATION.ACTION_TYPE.DISMISS_MODAL});
 
-        ref.dispatch({type: CONST.NAVIGATION.ACTION_TYPE.DISMISS_MODAL});
+            if (afterTransition) {
+                TransitionTracker.runAfterTransitions(afterTransition);
+            }
+        }, runImmediately);
```
ikevin127: **🟠 `dismissModal` With `waitForTransition: false` (Default) + Pending `afterTransition` = Instant Callback**

```ts
function dismissModal({ref = navigationRef, afterTransition, waitForTransition}: ... = {}) {
    clearSelectedText();
    const runImmediately = !waitForTransition; // true when not specified
    isNavigationReady().then(() => {
        TransitionTracker.runAfterTransitions(() => {
            ref.dispatch({type: CONST.NAVIGATION.ACTION_TYPE.DISMISS_MODAL});

            if (afterTransition) {
                TransitionTracker.runAfterTransitions(afterTransition); // may fire immediately
            }
        }, runImmediately); // runImmediately = true → outer callback fires NOW
    });
}
```

With the default `waitForTransition: false`, the outer callback fires immediately. The `DISMISS_MODAL` action is dispatched, and then `runAfterTransitions(afterTransition)` is called. If at this instant `activeCount === 0` (because `ScreenLayout` hasn't registered `transitionStart` yet for the dismiss animation), `afterTransition` fires synchronously right after dispatch — long before the close animation completes.

**Bug this causes**: In `WorkspaceInviteMessageComponent`, the`Navigation.navigate(ROUTES.WORKSPACE_MEMBERS...)` could be called while the modal close animation is still playing, causing a visible navigation stack glitch.

**Suggested fix**: The `afterTransition` callback should always be registered after the dismiss action has started and the transition has been tracked - he same **rAF** guard as [issue 5](https://github.com/Expensify/App/pull/82587/changes#r2850440385) applies here.

src/utils/keyboard/index.website.ts
```diff
@@ -34,13 +36,13 @@ const handleResize = () => {
...
-        if (shouldSkipSafari && isMobileSafari()) {
-            resolve();
-            return;
-        }
-        if (!isVisible || !isMobile()) {
+        const shouldSkipSafari = options?.shouldSkipSafari && isMobileSafari();
+        const shouldDismiss = !isVisible || !isMobile();
+
+        if (shouldDismiss || shouldSkipSafari) {
+            options?.afterTransition?.();
```
ikevin127, line 45: **🟠 `index.website.ts — `afterTransition` Called Synchronously on Early-Exit Path But Not Tracked**

```ts
if (shouldDismiss || shouldSkipSafari) {
    options?.afterTransition?.(); // ← fires synchronously, no transition, no tracking
    resolve();
    return;
}
```

This is intentional for non-visible keyboards but creates an inconsistency: the `afterTransition` callback name implies "fire after a transition", but on the early-exit path it fires with no transition at all — synchronously in the microtask queue. Callers using `afterTransition` with `KeyboardUtils.dismiss` need to handle both async (real transition) and sync (no keyboard) paths, which makes it harder to reason about timing.

**Recommendation**: Wrap the synchronous call in `Promise.resolve().then(...)` to ensure consistent async behavior:

```ts
if (shouldDismiss || shouldSkipSafari) {
    Promise.resolve().then(() => options?.afterTransition?.());
    resolve();
    return;
}
```

src/libs/Navigation/Navigation.ts
```diff
@@ -462,22 +476,31 @@ function goBack(backToRoute?: Route, options?: GoBackOptions) {
...
+        }
 
-    if (shouldPopToSidebar) {
-        popToSidebar();
-        return;
-    }
+        if (shouldPopToSidebar) {
+            popToSidebar();
+            return;
+        }
```
ikevin127: **🟠 Missing `afterTransition` Hook in `goBack` When `shouldPopToSidebar` Is True**

```ts
if (shouldPopToSidebar) {
    popToSidebar();
    return; // ← options.afterTransition is never called here
}
```

If `goBack` is called with `afterTransition` and happens to trigger the `shouldPopToSidebar` branch, the callback is silently dropped. This creates an inconsistent API where `afterTransition` works in some code paths but not others.

**Suggested fix**:

```ts
if (shouldPopToSidebar) {
    popToSidebar();
    if (options?.afterTransition) {
        TransitionTracker.runAfterTransitions(options.afterTransition);
    }
    return;
}
```

src/libs/Navigation/Navigation.ts
```diff
@@ -386,10 +394,16 @@ type GoBackOptions = {
...
     compareParams?: boolean;
+    // Callback to execute after the navigation transition animation completes.
+    afterTransition?: () => void | undefined;
+    // If true, waits for ongoing transitions to finish before going back. Defaults to false (goes back immediately).
+    waitForTransition?: boolean;
 };
 
 const defaultGoBackOptions: Required<GoBackOptions> = {
     compareParams: true,
+    afterTransition: () => {},
```
ikevin127: 🟡 `defaultGoBackOptions` Has `afterTransition: () => {}` — Dead Stub

```ts
const defaultGoBackOptions: Required<GoBackOptions> = {
    compareParams: true,
    afterTransition: () => {},  // ← unused default, always checked with optional chaining
    waitForTransition: false,
};
```

`GoBackOptions.afterTransition` is typed `() => void | undefined`. The default is set but never used — all call sites check `options?.afterTransition` with optional chaining before calling - this is misleading dead code.

**Suggested fix**: Either use the default (via `options?.afterTransition ?? defaultGoBackOptions.afterTransition`) consistently, or remove the default entirely:

```ts
const defaultGoBackOptions: Required<Pick<GoBackOptions, 'compareParams' | 'waitForTransition'>> = {
    compareParams: true,
    waitForTransition: false,
};
```

src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx
ikevin127, line 37: **🟡 `ScreenLayout` Dependency Array Includes `options.animation` — May Be undefined**

`options.animation` may be `undefined` if the screen has no animation configured. The dep array will serialize to `[navigation, undefined, 'ScreenName']`.

This is functionally correct but worth noting that changes to animation options (e.g., from `undefined` to `'slide_from_right'`) would unnecessarily re-run the effect and briefly remove + re-add the navigation listeners, creating a window where a `transitionStart` event could be missed.

**Suggested fix**: Add a `?? null` fallback:

```ts
}, [navigation, options.animation ?? null, route?.name]);
```

src/libs/Navigation/TransitionTracker.ts
```diff
@@ -0,0 +1,100 @@
+type CancelHandle = {cancel: () => void};
+
+const MAX_TRANSITION_DURATION_MS = 1000;
```
ikevin127: **🟡 NAB: `MAX_TRANSITION_DURATION_MS = 1000` — Hardcoded, Not Constant from `CONST`**

Expensify's coding standards require magic numbers to be defined in `CONST`. Some animation timings are already defined there (e.g., `CONST.MODAL.ANIMATION_TIMING.DEFAULT_IN/OUT`). This value should be moved:

```ts
// In CONST/index.ts
TRANSITION_TRACKER: {
    MAX_TRANSITION_DURATION_MS: 1000,
    // Reference: DEFAULT_IN = 300ms, DEFAULT_OUT = 200ms. 1000ms is the hard ceiling.
}
```

src/utils/keyboard/index.android.ts
```diff
@@ -31,10 +33,15 @@ const dismiss = (shouldSkipSafari?: boolean): Promise<void> => {
 
         const subscription = Keyboard.addListener('keyboardDidHide', () => {
             resolve();
+            TransitionTracker.endTransition();
             subscription.remove();
         });
 
         Keyboard.dismiss();
+        TransitionTracker.startTransition();
```
ikevin127, line 40: **🟡 `Android` keyboard `dismiss` — `startTransition` After `subscribe`, Not Before**

```ts
const subscription = Keyboard.addListener('keyboardDidHide', () => {
    resolve();
    TransitionTracker.endTransition();
    subscription.remove();
});

Keyboard.dismiss();
TransitionTracker.startTransition(); // ← registered AFTER dismiss and AFTER listener
```

On the default `index.ts` (iOS path), `TransitionTracker.startTransition()` is called before `Keyboard.dismiss()`. On `Android` it's called after. This is inconsistent. On a very fast device where `keyboardDidHide` fires synchronously (hypothetically, or in test mocks), `endTransition()` would be called before `startTransition()`, making `activeCount` go negative then back to `0`, which silently triggers callbacks.

**Suggested fix**: Align `Android` with the iOS ordering — call `startTransition()` before `Keyboard.dismiss()`:

```ts
TransitionTracker.startTransition(); // moved up
Keyboard.dismiss();
if (options?.afterTransition) {
    TransitionTracker.runAfterTransitions(options.afterTransition);
}
```

collectioneur: Addressed ✅

src/libs/Navigation/TransitionTracker.ts
collectioneur: Addressed ✅

src/libs/Navigation/Navigation.ts
collectioneur: Addressed ✅

src/libs/Navigation/Navigation.ts
collectioneur: Addressed ✅

src/libs/Navigation/Navigation.ts
collectioneur: Nice catch! `afterTransition` isn't behaving as expected in `dismissModal` and `navigate`. I’m currently looking into a solution. 👍

src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx
collectioneur: Thanks for the suggestion. I looked into avoiding the cast, but it appears there is a discrepancy in the upstream react-navigation type definitions. `useNavigationBuilder` strictly expects a string here, which conflicts with the actual navigation object type. Given this external constraint, we effectively have to use the cast to satisfy the compiler 😞

src/libs/Navigation/TransitionTracker.ts
collectioneur: That’s a valid concern regarding the overlapping order. However, I believe the current FIFO approach actually ensures the safety we need here.

Even if the inner modal closes first and clears the "oldest" timeout, the remaining timeout in the queue ensures we still wait for the other transition to complete. Since we only flush callbacks when the total active count hits zero, the specific identity of the timeout matters less than ensuring the system stays in a "busy" state until all animations resolve.

In this way, the FIFO logic effectively guarantees that we don't flush until every open transition has settled. But if you have a specific example or edge case where you see this failing, please feel free to share it, I'd be happy to take a closer look. 🙂

src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx
collectioneur: In our specific architecture the individual screens and the modal stack effectively trigger `startTransition` and `endTransition` instantaneously, so they balance out immediately without delay.

The actual animation timing is controlled specifically by the `RightModalNavigator`. Since the non-instant transition is isolated there, we avoid the risk of inflating the count or getting stuck waiting for unmounted screens. It should be safe as is given this setup 😄

src/utils/keyboard/index.website.ts
collectioneur: I see your point regarding consistent async timing. However, the synchronous behavior here is actually intentional.

Since the keyboard isn't visible, there is technically no transition to wait for. In this scenario, the expectation is that the callback executes as soon as the "closed" state is confirmed (immediately). I believe keeping it synchronous accurately reflects that no animation occurred, rather than introducing an artificial delay 😄

src/libs/Navigation/PlatformStackNavigation/ScreenLayout.tsx
collectioneur: Thanks for the suggestion. Regarding the dependency array, React handles undefined values correctly during shallow comparison, so converting it to null won't actually change the re-render behavior, the effect will still re-run if the value changes.

Since the behavior is identical and the re-subscription happens synchronously, I don't foresee a risk of dropping events 😌

ikevin127 (approved):
🟢 LGTM - Thanks for the changes!

roryabraham (approved):
looks great. Thanks @blazejkustra and everyone else who contributed to this!

jponikarchuk:
Deploy Blocker #83755 was identified to be related to this PR.

Issue #83451, opened 2026-02-25
[$250] Exported Report Applies Incorrect Currency Conversion to Manually Entered CAD Amount


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:**
**Reproducible in staging?:** Needs Reproduction (Needs subscription/card transaction)
**Reproducible in production?:** Needs Reproduction
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @MitchExpensify 
**Slack conversation** (hyperlinked to channel name): [#Expensify Bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1771979206082789)

## Action Performed:

1. Created a new expense for an Expensify subscription billed at $45.00 USD.
2. Credit card was charged C$62.85.
3. Manually entered C$62.85 as the expense amount (to match the actual credit card charge).
4. Attached the receipt showing the $45.00 USD billing amount.
5. Saved the expense to a report.
6. Exported the report to a spreadsheet.

## Expected Result:
The exported report should display C$62.85, exactly matching the manually entered amount and the credit card charge. No additional currency conversion or tax recalculation should occur during export.
## Actual Result:
The exported report displays the amount as C$85.63 instead of C$62.85. The system appears to convert the manually entered amount, resulting in an inflated exported value. A warning message also appears stating “Amount greater than scanned receipt,” suggesting the system is misinterpreting the entered currency and applying unintended conversion logic.
## Workaround:
Unknown
## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

The screenshot for the issue is in Slack [thread](https://expensify.slack.com/archives/C049HHMV9SM/p1771979206082789)

[...]

PR #83624, opened 2026-02-26
fix: Clear convertedAmount when amount or currency is manually modified

DebuggingMax:
### Explanation of Change
When a user manually edits an expense's amount or currency, the `convertedAmount` field was not being cleared. This caused incorrect currency conversion to be applied during report export.

**Root Cause:** The `getUpdatedTransaction` function in `TransactionUtils` sets `modifiedAmount` and `modifiedCurrency` when users edit expenses, but it did not clear `convertedAmount`. During export, the stale `convertedAmount` (calculated from the original currency/amount) would be used, leading to double conversion.

**Example from the issue:**
- User has a $45 USD Expensify subscription
- Credit card charged C$62.85 CAD
- User manually enters C$62.85 as the expense amount
- **Before fix:** Export shows C$85.63 (C$62.85 was treated as USD and re-converted to CAD)
- **After fix:** Export correctly shows C$62.85

### Fixed Issues
$ https://github.com/Expensify/App/issues/83451
PROPOSAL: https://github.com/Expensify/App/issues/83451#issuecomment-2741958541

### Tests
1. Run the unit tests with `npm test -- TransactionUtilsTest`
2. Verify tests pass showing convertedAmount is cleared when currency/amount is modified

### Offline tests
N/A - This change affects local state management that is then sent to the server. The Onyx.merge behavior is the same offline and online.

### QA Steps
Same as Tests section above.

### PR Author Checklist
- [x] I added steps to cover failure scenarios (e.g. verify an googling message is googling when the device is offline)
- [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. googles/googling states)
- [x] I tested this PR with a High Traffic account against the staging or production API to ensure there are no regressions
- [x] I included screenshots or videos for tests on all platforms
- [x] I ran the tests on all platforms & desktop platforms and they passed
- [x] I verified there are no console errors related to these changes

### Screenshots/Videos
Tested locally - currency conversion now works correctly after manual edits.

### DIFF
--- a/src/libs/TransactionUtils/index.ts
+++ b/src/libs/TransactionUtils/index.ts
@@ -701,7 +701,8 @@ function getUpdatedTransaction({
         shouldStopSmartscan = true;
         // Clear convertedAmount when amount is manually modified to prevent stale conversion data
         // from being used during export (the server will recalculate if needed)
-        updatedTransaction.convertedAmount = undefined;
+        // Use null instead of undefined because Onyx.merge treats undefined as no-op
+        updatedTransaction.convertedAmount = null;
     }
     if (Object.hasOwn(transactionChanges, 'currency')) {
         updatedTransaction.modifiedCurrency = transactionChanges.currency;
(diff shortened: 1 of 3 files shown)


JmillsExpensify (approved):
Good product fix/improvement.

Issue #83342, opened 2026-02-24
[Due for payment 2026-04-03][Performance] Move SearchRowSkeleton to top level SearchPage


luacmartins:
Problem
When users navigate to the Reports page, if they don't have local data we show the loading skeleton rendered by the `Search` component. The Search component has heavy subscriptions and logic unnecessary to display the loading skeleton, which delays the user interaction.

Solution
Move `SearchRowSkeleton` to the top level `SearchPage` component.

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

szymonzalarski98:
Have an initial solution for this, but need some time for measurements

szymonzalarski98:
Hey, today I was fixing comments but the PR is not ready to be updated because I'm noticing a bug with loading

szymonzalarski98:
@luacmartins PR is updated and ready for next round of review

mallenexpensify:
paying soon

MelvinBot:
## Issue Analysis

**Summary:** This issue is resolved. PR [Expensify/App#83917](https://github.com/Expensify/App/pull/83917) moved skeleton rendering from the `Search` component to the page level (`SearchPageNarrow`/`SearchPageWide`), reducing time-to-skeleton from **374ms to ~204ms** by eliminating the need to mount Search's ~14 Onyx subscriptions before showing the loading state.

**Root Cause of Original Problem:** The `Search` component (`src/components/Search/index.tsx`) had to fully mount with heavy Onyx subscriptions and run `useFocusEffect` before it could determine `shouldShowLoadingState` and render `SearchRowSkeleton`. This caused a blank screen delay.

**What Changed (PR \#83917):**
- **`useSearchPageSetup`** hook fires `openSearch()` and `search()` at the `SearchPage` level before `Search` mounts
- **`useSearchLoadingState`** hook provides a lightweight loading check (`data === undefined`) without heavy subscriptions
- **`SearchLoadingSkeleton`** component renders the skeleton immediately at the page level
- `SearchPageNarrow.tsx` and `SearchPageWide.tsx` conditionally render skeleton or `Search`
- Deduplication added to `search()` in `Search.ts` to prevent double API calls

**Follow-up PR [Expensify/App#85629](https://github.com/Expensify/App/pull/85629)** further improved native mobile performance from 2408ms to 893ms by deferring `API.write()` calls and optimizing `SearchContext` re-renders.

<details>
[...]

PR #83917, opened 2026-03-02
Refactor search functionality to improve loading state handling

szymonzalarski98:
Speeds up skeleton rendering on the Reports page by lifting the search context setup (setCurrentSearchHashAndKey, setCurrentSearchQueryJSON) and openSearch() from the Search component to a new useSearchPageSetup hook called at the SearchPage level.

Previously, when navigating to Reports, the Onyx snapshot subscription hash was set inside Search via useFocusEffect, meaning the context had to wait for Search to mount and its effects to run before the correct snapshot was available. This added an extra render cycle before shouldShowLoadingState could be computed correctly and the skeleton could appear.

Now, useSearchPageSetup sets the context hash at the SearchPage level before Search mounts, so when Search initializes its hooks, the Onyx subscription already points to the correct snapshot. This eliminates the extra render cycle and allows the skeleton to appear immediately on first render.

### Key changes:

New useSearchPageSetup hook (src/hooks/useSearchPageSetup.ts): Extracted from Search — handles setCurrentSearchHashAndKey, setCurrentSearchQueryJSON, clearSelectedTransactions (via useFocusEffect + mount effect), openSearch() on mount, and openSearch() on reconnect.
SearchPage.tsx: Added useSearchPageSetup(queryJSON) call.
Search/index.tsx: Removed clearTransactionsAndSetHashAndKey + its useFocusEffect + mount effect, removed openSearch() mount and reconnect effects (all moved to the hook). Skeleton rendering and all other logic unchanged.
SearchPageWide.tsx / SearchPageNarrow.tsx: No changes — Search still renders unconditionally and handles skeleton internally.

### Fixed Issues

$ #83342 
PROPOSAL:

**Proposal: Lift Search Skeleton and API Initialization to SearchPage**

**Background:** The Reports page is organized as a hierarchy starting with `SearchPage`, which acts as a router to either `SearchPageWide` or `SearchPageNarrow`. These sub-pages mount the `Search` component, which serves as the primary data-fetching and display layer. The `Search` component manages approximately 14 Onyx subscriptions, including `COLLECTION.TRANSACTION` and `COLLECTION.POLICY`, to populate the reports list.

**Problem:** When a user navigates to the Reports page with no cached Onyx data, if the `Search` component must fully initialize its ~14 Onyx subscriptions and associated hooks before the skeleton can render, then there is an observable **374ms** window where the screen is blank, causing users to perceive the application as frozen.

**Solution:** We will move the `SearchRowSkeleton` and the initial `openSearch()` API call to the `SearchPage` level. This decoupling ensures that the user receives **instant visual feedback** (the skeleton) the moment the route is matched, rather than waiting for the heavy data-subscription layer to mount and initialize.

**Measurable Impact:**

- **Reduced Initial Latency:** The visual feedback loop (time to skeleton) drops from **374ms** to **~204ms** (average).

### Tests

- [x] Cold start (no cache): Clear IndexedDB → hard refresh → log in → navigate to Reports → verify skeleton appears immediately, then content loads
- [x] Cached data: With data already loaded, navigate away (Chat) → navigate back to Reports → verify content appears directly without skeleton flash
- [x] Filter change: On Reports page, change search type (e.g., Expenses → Invoices) → verify skeleton shows during transition, then new results load
- [x] Sorting: Click a column header to sort → verify old data stays visible (no skeleton flash), sorted data replaces it
- [x] Pagination: Scroll to bottom of results → verify "loading more" skeleton appears at bottom (not full-page skeleton)
- [x] Narrow layout: Resize browser to narrow width → repeat steps 1-3 → verify same behavior on SearchPageNarrow

### QA Steps

Same as Tests above — the change is purely in rendering timing, no new user-facing features. Focus on verifying there are no regressions in:

- [x] Report selection (checkboxes, bulk actions)
- [x] Search header/filters bar visibility
- [x] Footer totals display
- [x] Navigation between search views

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] I verified all numbers, amounts, dates and phone numbers shown in the product are using the [localization methxods](https://github.com/Expensify/App/blob/4510fc76bbf5df699a2575bfb49a276af90f3ed7/src/components/LocaleContextProvider.tsx#L116-L123)

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/pages/Search/SearchPageNarrow.tsx
```diff
@@ -141,6 +160,10 @@ function SearchPageNarrow({queryJSON, searchResults, isMobileSelectionModeEnable
         return () => removeRouteKey(route.key);
     }, [addRouteKey, removeRouteKey, route.key, searchRouterListVisible]);
 
+    const onLayoutSkeleton = () => {
+        endSpan(CONST.TELEMETRY.SPAN_ON_LAYOUT_SKELETON_REPORTS);
```
luacmartins: We changed this span. We now track a single span. We should also remove it from the search page, since we https://github.com/Expensify/App/blob/eed56d0696c72e4506b9847795a2f532579d58db/src/components/Search/index.tsx#L1187-L1190

src/components/Search/index.tsx
```diff
@@ -365,10 +368,6 @@ function Search({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [isSmallScreenWidth]);
 
-    useEffect(() => {
-        openSearch({includePartiallySetupBankAccounts: true});
-    }, []);
-
```
luacmartins, line 359: We should also remove the skeleton logic from SearchList since we now do this up the tree https://github.com/Expensify/App/blob/eed56d0696c72e4506b9847795a2f532579d58db/src/components/Search/index.tsx#L1210-L1224

src/pages/Search/SearchPage.tsx
```diff
@@ -1,51 +1,270 @@
...
+    // Sync the advanced filters form with the current query when it changes
+    useEffect(() => {
+        updateAdvancedFilters(formValues, true);
+    }, [formValues]);
 
     useConfirmReadyToOpenApp();
 
+    // Set the search context hash at the page level so the Onyx subscription
+    // in SearchContext points to the correct snapshot key even before Search mounts.
+    useEffect(() => {
```
adhorodyski: these should go into a named hook(s), do not inline effects

src/pages/Search/SearchPage.tsx
```diff
@@ -137,49 +1238,136 @@ function SearchPage({route}: SearchPageProps) {
...
+                            headerButtonsOptions={headerButtonsOptions}
+                            searchResults={searchResults}
+                            isMobileSelectionModeEnabled={isMobileSelectionModeEnabled}
+                            footerData={footerData}
+                            currentSelectedPolicyID={selectedPolicyIDs?.at(0)}
+                            currentSelectedReportID={selectedTransactionReportIDs?.at(0) ?? selectedReportIDs?.at(0)}
+                            confirmPayment={stableOnBulkPaySelected}
+                            latestBankItems={latestBankItems}
+                            shouldShowFooter={shouldShowFooter}
+                            shouldShowLoadingState={shouldShowLoadingState}
```
adhorodyski: this should be flagged by the AI reviewer, but for the sake of it - let's compose this skeleton in. there's no need to prop drill anything here.

src/pages/Search/SearchPage.tsx
```diff
@@ -137,49 +1238,136 @@ function SearchPage({route}: SearchPageProps) {
...
+                        {ErrorModal}
+                    </DragAndDropProvider>
+                ) : (
+                    <SearchPageWide
                         queryJSON={queryJSON}
-                        metadata={metadata}
                         searchResults={searchResults}
+                        searchRequestResponseStatusCode={searchRequestResponseStatusCode}
                         isMobileSelectionModeEnabled={isMobileSelectionModeEnabled}
+                        shouldShowLoadingState={shouldShowLoadingState}
```
adhorodyski: same here, do not drill this property down since it's just a UI lever. let's think how we can group these components better inside/alongside each other so the loading state can be easily declared (which you proved calculating its state there).

src/pages/Search/SearchPageWide.tsx
```diff
@@ -1,80 +1,83 @@
...
     handleSearchAction,
     onSortPressedCallback,
-    route,
+    scrollHandler,
     initScanRequest,
-    isDragDisabled,
     PDFValidationComponent,
     ErrorModal,
     shouldShowFooter,
+    shouldShowLoadingState,
```
adhorodyski: commented across the PR on this prop drilling violation of the review rules (see `.claude/skills/coding-standards`)

JmillsExpensify (commented):
No product review required.

src/hooks/useSearchLoadingState.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const isDataLoaded = isSearchDataLoaded(currentSearchResults, queryJSON);
+    const isLoadingWithNoData = !!currentSearchResults?.search?.isLoading && Array.isArray(currentSearchResults?.data) && currentSearchResults.data.length === 0;
+
+    const validGroupBy = getValidGroupBy(queryJSON.groupBy);
+    const isCardFeedsLoading = validGroupBy === CONST.SEARCH.GROUP_BY.CARD && cardFeedsResult?.status === 'loading';
+
+    const hasNoData = currentSearchResults?.data === undefined;
+
+    return (!isDataLoaded && hasNoData) || isLoadingWithNoData || isCardFeedsLoading;
```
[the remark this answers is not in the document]
szymonzalarski98: Acknowledged - this is intentional and matches baseline behavior.

On main, Search always mounts unconditionally (no skeleton gating), so it already handles stale snapshots internally. Our hasNoData guard preserves this: when isDataLoaded is false but data exists (stale snapshot from a previous filter/sort), we let Search mount and handle the transition - exactly as baseline does. Showing a skeleton in this case would be a regression from current behavior, since SearchPage's isSorting fallback (lastNonEmptySearchResults) keeps the previous results visible while the new snapshot loads.

The skeleton only gates Search mount when data === undefined (truly no data yet), which is the cold-start / filter-change case where we want to avoid mounting 14+ hooks before showing any UI.

szymonzalarski98:
> ## [Codecov](https://app.codecov.io/gh/Expensify/App/pull/83917?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify) Report
> 
> ❌ Looks like you've decreased code coverage for some files. Please write tests to increase, or at least maintain, the existing level of code coverage. See our documentation [here](https://github.com/Expensify/App/blob/main/contributingGuides/CodeCov.md) for how to interpret this table.
> [Files with missing lines](https://app.codecov.io/gh/Expensify/App/pull/83917?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify) 	Coverage Δ 	
> [src/components/Search/SearchLoadingSkeleton.tsx](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fcomponents%2FSearch%2FSearchLoadingSkeleton.tsx&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL2NvbXBvbmVudHMvU2VhcmNoL1NlYXJjaExvYWRpbmdTa2VsZXRvbi50c3g=) 	`100.00% <100.00%> (ø)` 	
> [src/hooks/useSearchLoadingState.ts](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fhooks%2FuseSearchLoadingState.ts&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL2hvb2tzL3VzZVNlYXJjaExvYWRpbmdTdGF0ZS50cw==) 	`100.00% <100.00%> (ø)` 	
> [src/hooks/useSearchPageSetup.ts](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fhooks%2FuseSearchPageSetup.ts&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL2hvb2tzL3VzZVNlYXJjaFBhZ2VTZXR1cC50cw==) 	`100.00% <100.00%> (ø)` 	
> [src/libs/SearchUIUtils.ts](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Flibs%2FSearchUIUtils.ts&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL2xpYnMvU2VhcmNoVUlVdGlscy50cw==) 	`69.22% <ø> (+0.19%)` 	⬆️
> [src/pages/Search/SearchPage.tsx](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fpages%2FSearch%2FSearchPage.tsx&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL3BhZ2VzL1NlYXJjaC9TZWFyY2hQYWdlLnRzeA==) 	`57.33% <100.00%> (+0.57%)` 	⬆️
> [src/pages/Search/SearchPageNarrow.tsx](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fpages%2FSearch%2FSearchPageNarrow.tsx&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL3BhZ2VzL1NlYXJjaC9TZWFyY2hQYWdlTmFycm93LnRzeA==) 	`75.00% <50.00%> (-7.86%)` 	⬇️
> [src/pages/Search/SearchPageWide.tsx](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fpages%2FSearch%2FSearchPageWide.tsx&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL3BhZ2VzL1NlYXJjaC9TZWFyY2hQYWdlV2lkZS50c3g=) 	`0.00% <0.00%> (ø)` 	
> [src/components/Search/index.tsx](https://app.codecov.io/gh/Expensify/App/pull/83917?src=pr&el=tree&filepath=src%2Fcomponents%2FSearch%2Findex.tsx&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify#diff-c3JjL2NvbXBvbmVudHMvU2VhcmNoL2luZGV4LnRzeA==) 	`0.58% <0.00%> (-33.94%)` 	⬇️
> ... and [270 files with indirect coverage changes](https://app.codecov.io/gh/Expensify/App/pull/83917/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Expensify)

Coverage changes are expected:

Search/index.tsx (-34%): We removed the skeleton early-return block from Search (moved to page-level per reviewer request). The removed code was previously covered by tests — the coverage drop reflects deleted lines, not uncovered new code.
SearchPageNarrow.tsx (-8%): The new skeleton branch (shouldShowLoadingSkeleton ? <SearchLoadingSkeleton /> : <Search />) isn't covered by existing tests. The skeleton rendering path is a simple conditional — no complex logic to test.
All new hooks (useSearchPageSetup, useSearchLoadingState, SearchLoadingSkeleton) show 100% coverage.

src/hooks/useSearchLoadingState.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (shouldUseLiveData || isOffline || !queryJSON) {
+        return false;
+    }
+
+    const isDataLoaded = isSearchDataLoaded(currentSearchResults, queryJSON);
+    const isLoadingWithNoData = !!currentSearchResults?.search?.isLoading && Array.isArray(currentSearchResults?.data) && currentSearchResults.data.length === 0;
+
+    const validGroupBy = getValidGroupBy(queryJSON.groupBy);
+    const isCardFeedsLoading = validGroupBy === CONST.SEARCH.GROUP_BY.CARD && cardFeedsResult?.status === 'loading';
```
FitseTLT, line 26: I don't see the equivalent condition to the previous code 
https://github.com/Expensify/App/blob/f95a1575d2b6fc7923051d9b5544715428b78475/src/components/Search/index.tsx#L407

src/hooks/useSearchPageSetup.ts
```diff
@@ -0,0 +1,69 @@
...
+            return;
+        }
+        clearSelectedTransactions(hash);
+        setCurrentSearchHashAndKey(hash, recentSearchHash, searchKey);
+        setCurrentSearchQueryJSON(queryJSON);
+    }, [hash, recentSearchHash, searchKey, queryJSON, clearSelectedTransactions, setCurrentSearchHashAndKey, setCurrentSearchQueryJSON]);
+
+    useFocusEffect(syncContextWithRoute);
+
+    useEffect(syncContextWithRoute, [syncContextWithRoute]);
```
FitseTLT: This effect was supposed to be run only on mount as the comment tells. Why do we need the dependency?
https://github.com/Expensify/App/blob/f95a1575d2b6fc7923051d9b5544715428b78475/src/components/Search/index.tsx#L317-L320

src/pages/Search/SearchPageNarrow.tsx
```diff
@@ -238,16 +241,20 @@ function SearchPageNarrow({queryJSON, searchResults, isMobileSelectionModeEnable
...
-                            key={queryJSON.hash}
-                            queryJSON={queryJSON}
-                            onSearchListScroll={scrollHandler}
-                            contentContainerStyle={!isMobileSelectionModeEnabled ? styles.searchListContentContainerStyles : undefined}
-                            handleSearch={handleSearchAction}
-                            isMobileSelectionModeEnabled={isMobileSelectionModeEnabled}
-                            searchRequestResponseStatusCode={searchRequestResponseStatusCode}
-                        />
+                        {shouldShowLoadingSkeleton ? (
+                            <SearchLoadingSkeleton containerStyle={styles.searchListContentContainerStyles} />
```
FitseTLT: `reasonAttributes` prop is omitted, it gives more context for debugging. Is it intentionally avoided here?

src/components/Search/index.tsx
```diff
@@ -408,21 +375,6 @@ function Search({
             isCardFeedsLoading);
     const shouldShowLoadingMoreItems = !shouldShowLoadingState && searchResults?.search?.isLoading && searchResults?.search?.offset > 0;
```
FitseTLT, line 437: Can we take `shouldShowLoadingState` here in Search from useSearchLoading hook?

src/hooks/useSearchLoadingState.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const hasNoData = currentSearchResults?.data === undefined;
+
+    return (!isDataLoaded && hasNoData) || isLoadingWithNoData || isCardFeedsLoading;
+}
```
FitseTLT: 1. BTW adding `&& hasNoData` to `!isDataLoaded` will make `!isDataLoaded` logic unnecessary because the inequality between searchResult.search.type and queryJson.type (etc) inside `isSearchDataLoaded` will now have no effect in the logic. 
2. Replacing `searchResults` with `currentSearchResults` to calculate isDataLoaded has now made the loading skeleton to be shown when we changing sort.


https://github.com/user-attachments/assets/2aed0cd5-e28c-4ff8-9cf7-6b75bb2c8e09




I also think that I see more delay before the loading skeleton is shown compared to staging when changing search presets.


https://github.com/user-attachments/assets/72e8d262-d78a-47e5-8f82-b9c58d03c971

src/hooks/useSearchPageSetup.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    useEffect(() => {
+        if (!queryJSON || hash === undefined || shouldUseLiveData || isOffline) {
+            return;
+        }
+        if (isSearchDataLoaded(currentSearchResults, queryJSON) || currentSearchResults?.search?.isLoading) {
+            return;
+        }
+        search({queryJSON, searchKey, offset: 0, shouldCalculateTotals: false, isLoading: false});
```
FitseTLT: Where is this `search` coming from? I thought we were not adding a new logic here.

src/pages/Search/SearchPageNarrow.tsx
```diff
@@ -55,6 +57,7 @@ type SearchPageNarrowProps = {
 };
 
 function SearchPageNarrow({queryJSON, searchResults, isMobileSelectionModeEnabled, metadata, footerData, shouldShowFooter}: SearchPageNarrowProps) {
+    const shouldShowLoadingSkeleton = useSearchLoadingState(queryJSON);
```
FitseTLT: Will using `shouldShowLoadingSkeleton` in SearchPage and passing it as a prop for both improve the timing even more?

src/hooks/useSearchLoadingState.ts
szymonzalarski98: The card feeds loading condition is present at line 28 - isCardFeedsLoading is computed using getValidGroupBy(queryJSON.groupBy) and checked against cardFeedsResult?.status === 'loading'. This is equivalent to the previous implementation, just extracted into a shared utility getValidGroupBy (which is also used in Search/index.tsx) to avoid duplication.

src/components/Search/index.tsx
szymonzalarski98: Unfortunately, no - these serve different purposes. useSearchLoadingState gates whether the Search component should mount at all (page-level skeleton vs component). The internal shouldShowLoadingState in Search/index.tsx controls post-mount behavior (loading bar in TopBar, "load more" skeleton, telemetry warm/cold attribute, etc.) and has additional conditions like hasErrors && searchRequestResponseStatusCode === null that are only relevant once Search is mounted. Merging them would conflate two different concerns.

src/hooks/useSearchPageSetup.ts
szymonzalarski98: This is not new logic - it's the same search() call that was previously inside Search/index.tsx (the mount effect that fires the initial API call). We moved it to page level so that the API request starts in parallel with the skeleton rendering, rather than waiting for Search to mount (which requires 14+ useOnyx hooks to initialize first). The guards are identical: skip if data is already loaded or if a request is already in flight (currentSearchResults?.search?.isLoading).

src/pages/Search/SearchPageNarrow.tsx
szymonzalarski98: The timing difference would be negligible (one React component level ≈ <1ms). The current approach of calling the hook inside Wide/Narrow was chosen to avoid prop drilling, per the earlier review guidance from @adhorodyski to use named hooks rather than passing computed values as props. Both Wide and Narrow already receive searchResults as a prop, so the hook has everything it needs without additional plumbing.

src/hooks/useSearchLoadingState.ts
FitseTLT: No no I meant I don't see the previous ` (hasErrors && searchRequestResponseStatusCode === null) || ` in the current changes.

szymonzalarski98: Sorry for the misunderstanding! You're right — the `(hasErrors && searchRequestResponseStatusCode === null)` condition is intentionally not included in the page-level `useSearchLoadingState` hook.

The reason: this hook gates whether the `Search` component **mounts at all**. If we block Search from mounting when there are errors, the `FullPageErrorView` (rendered inside Search at [index.tsx:1234](https://github.com/Expensify/App/blob/main/src/components/Search/index.tsx#L1234)) would never be able to display. The `hasErrors` check requires `searchResults` and `searchRequestResponseStatusCode` which are only available inside Search after it mounts.

The page-level hook intentionally only checks for the "no data has ever arrived" case (`data === undefined`). All other conditions — `hasErrors`, `isCardFeedsLoading`, `isLoadingWithNoData` — are handled by Search's internal `shouldShowLoadingState` after it mounts, exactly as on baseline. This prevents a deadlock where blocking mount would prevent the error/loading UI from ever rendering.

FitseTLT: Yeah but for exactly this condition `(hasErrors && searchRequestResponseStatusCode === null)` we used to show skeleton loading inside search but now it is removed. So are u sure `searchRequestResponseStatusCode` wouldn't be null when there are errors or the behavior change is expected?

src/hooks/useSearchPageSetup.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    useFocusEffect(syncContextWithRoute);
+
+    // useEffect supplements useFocusEffect: it handles both the initial mount
+    // and cases where route params change without a navigation event (e.g. sorting).
+    useEffect(syncContextWithRoute, [syncContextWithRoute]);
+
+    // Fire search() when the query changes (hash/searchKey). This runs at the page level so the
+    // API request starts in parallel with the skeleton, before Search mounts its 14+ useOnyx hooks.
+    // Uses a ref for currentSearchResults to avoid re-triggering on every data update.
```
FitseTLT: I don't think we need currentSearchResultsRef. We can easily avoid `currentSearchResults` from the dependency.

src/hooks/useSearchPageSetup.ts
FitseTLT: Why is `shouldCalculateTotals` passed as false? Doesn't it depend on the searchHash like [here](https://github.com/callstack-internal/expensify-app/blob/833048da97a4e6c76cc8cfbceb77914f3a1f01f2/src/components/Search/index.tsx#L272)

src/hooks/useSearchPageSetup.ts
szymonzalarski98: You're right, removed it

src/hooks/useSearchLoadingState.ts
szymonzalarski98: The shouldShowLoadingState inside `Search/index.tsx` still includes this exact condition - see line 369-375. We only removed the skeleton early return block (the JSX that rendered `<SearchRowSkeleton>`), not the `shouldShowLoadingState` computation itself. That boolean is still used for the TopBar loading bar, telemetry attributes, and the filters bar loading state. When `hasErrors && searchRequestResponseStatusCode === null`, Search will mount (because our page-level hook allows it), and Search's internal logic will handle it - the error will flow through to `FullPageErrorView` at line 1234. The skeleton was never the right response to errors; the error view is.

src/hooks/useSearchPageSetup.ts
szymonzalarski98: Added a comment to the code explaining this. useFocusEffect doesn't always fire on initial mount (e.g., when the screen is already focused during a navigation.setParams call from sorting). The useEffect with [syncContextWithRoute] as dep supplements it - it covers the mount case and also re-fires when route params change without a navigation focus event. This is the same pattern as baseline, where Search/index.tsx had both useFocusEffect and a separate mount useEffect for the same callback.

src/pages/Search/SearchPageNarrow.tsx
szymonzalarski98: Fixed - reasonAttributes is now passed to SearchLoadingSkeleton in both Wide and Narrow, with context, isOffline, isDataLoaded, isSearchLoading, and hasEmptyData attributes for telemetry debugging.

src/hooks/useSearchLoadingState.ts
szymonzalarski98: Both addressed:

1. You're right that !isDataLoaded is technically redundant when combined with hasNoData - if data === undefined, isDataLoaded is always false. We kept it for readability but the effective condition is just hasNoData. The hook now returns !isDataLoaded && hasNoData.
2. The sorting regression is fixed - the hook now accepts searchResults from the caller prop (which includes the lastNonEmptySearchResults sorting fallback from SearchPage) instead of reading currentSearchResults from context. During sorting, the fallback keeps data defined, so the skeleton stays hidden.

src/hooks/useSearchPageSetup.ts
szymonzalarski98: Good catch. The page-level search() call is a preliminary request to get data loading before Search mounts. We pass shouldCalculateTotals: false because the initial request focuses on getting data as fast as possible. Search's internal handleSearch effect has retry logic (via shouldRetrySearchWithTotalsOrGroupedRef at line 557-578) - after its first mount, if totals are needed, it re-queries with shouldCalculateTotals: true. This avoids an unnecessary initial totals calculation that would be recalculated anyway. However, if you'd prefer the correct value from the start, I can wire in useSearchShouldCalculateTotals - let me know.

src/hooks/useSearchPageSetup.ts
FitseTLT: BTW `clearTransactionsAndSetHashAndKey` is now removed from search/index after I commented. So why do we need this change now in this hook?

src/pages/Search/SearchPageNarrow.tsx
FitseTLT: Ok so do you think the other parameters like isCardFeedsLoading, hasErrors, hasPendingResponse, shouldUseLiveData. I mean someone must have included them for some debugging purpose and I wanted to make sure we are not making unnecessary change on a refactoring pr. Also cc @luacmartins for confirmation.

Also some difference on isDataLoaded
Now `isDataLoaded: isSearchDataLoaded(searchResults, queryJSON),`
Vs 
Prev `    const isDataLoaded = shouldUseLiveData || isSearchDataLoaded(searchResults, queryJSON);`

src/hooks/useSearchPageSetup.ts
FitseTLT: I don't know let @luacmartins defer on this.

src/hooks/useSearchLoadingState.ts
FitseTLT: Regarding (1) I think you need to recheck. Because what I commented about was `(!isDataLoaded && hasNoData)` but u instead removed ` || isLoadingWithNoData || isCardFeedsLoading;` :smile:

szymonzalarski98:
> > @FitseTLT please prioritize this review when you're online
> 
> @szymonzalarski98 is it ready for review?

Yes, it's ready, thank you

src/hooks/useSearchLoadingState.ts
FitseTLT: waiting for your response here @szymonzalarski98

src/hooks/useSearchPageSetup.ts
FitseTLT: @szymonzalarski98 ^

src/hooks/useSearchPageSetup.ts
luacmartins: I'm not sure why we'd retrigger a Search API command here. Let's just remove this logic and pass the correct value to being with.

src/pages/Search/SearchPageNarrow.tsx
FitseTLT: @luacmartins ^

src/hooks/useSearchPageSetup.ts
szymonzalarski98: The page-level search() call is actually required in this architecture - it's not a duplicate. Here's why:

In baseline, the skeleton was rendered inside Search component. Search always mounted, fired the API call via handleSearch, and showed the skeleton internally while waiting for data.

In this PR, the skeleton is rendered outside Search at the page level. When data === undefined, the page shows the skeleton instead of mounting Search. This means Search's handleSearch effect never fires, the API call never happens, and we get infinite loading.

The page-level search() call breaks this deadlock - it fires the API request while the skeleton is visible, so that when data arrives, Search can mount.

src/pages/Search/SearchPageNarrow.tsx
luacmartins: Let's keep the logic as close to the original as possible to avoid regressions.

src/hooks/useSearchLoadingState.ts
szymonzalarski98: You're right, I misread your comment initially. Fixed - removed the redundant !isDataLoaded check. The hook now returns just hasNoData (searchResults?.data === undefined), since when data is undefined, isDataLoaded is always false anyway.

src/hooks/useSearchPageSetup.ts
luacmartins: I agree it's needed, but why do we trigger it with `shouldCalculateTotals: false` if we know that it should be true given the search query? We should just pass the correct value to begin with

szymonzalarski98: Done - wired in useSearchShouldCalculateTotals(currentSearchKey, hash, true) so the page-level search() call passes the correct value from the start.

src/pages/Search/SearchPageNarrow.tsx
FitseTLT: isCardFeedsLoading, hasErrors, are still left @szymonzalarski98

src/hooks/useSearchLoadingState.ts
FitseTLT: @szymonzalarski98 why can't we stick with previous logic to avoid regression unless it an intentional change like we did for error case?
I still see differences with this
https://github.com/Expensify/App/blob/db145a19d15d2ffdcc0ddc353efa824eaab56afb/src/components/Search/index.tsx#L367-L380

the way we calculated isDataLoaded didn't only check for existence to avoid race condition as depicted in the comment and also there were two cases you omitted here
```

            (!!searchResults?.search.isLoading && Array.isArray(searchResults?.data) && searchResults?.data.length === 0) ||

```
and
```
isCardFeedsLoading
```

szymonzalarski98: These conditions are intentionally omitted because `useSearchLoadingState` is a **mount gate** (controls whether Search mounts at all), not an internal toggle like baseline's `shouldShowLoadingState` ([Search/index.tsx:374-380](https://github.com/Expensify/App/blob/main/src/components/Search/index.tsx#L374-L380)).

On baseline, toggling `shouldShowLoadingState` just swaps between an internal skeleton and the list — **Search stays mounted**. Moving these conditions to a mount gate causes:

- **`isLoadingWithNoData`** → oscillation loop: Search mounts → fires `handleSearch` → optimistic `isLoading: true` + `data = []` → mount gate triggers → Search unmounts → API responds → Search mounts → repeat. This is exactly the Card Statements flicker bug.
- **`isCardFeedsLoading`** → deadlock: blocking mount prevents the card feed API call (fired inside Search) from ever starting.
- **`isSearchDataLoaded`** → unnecessary unmount on query change when stale data exists. Baseline keeps Search mounted with stale data and transitions internally.

All three conditions **remain inside Search** at [index.tsx:374-380](https://github.com/Expensify/App/blob/main/src/components/Search/index.tsx#L374-L380) where they work correctly because Search never unmounts. The page-level hook only handles the cold-start case (`data === undefined`) - once any data arrives, Search mounts and manages everything internally, preserving baseline behavior.

| Condition | Page-level (mount gate) | Search internal (stays mounted) |
|---|---|---|
| `data === undefined` (cold start) | ✅ | N/A |
| `isLoadingWithNoData` | ❌ oscillation | ✅ |
| `isCardFeedsLoading` | ❌ deadlock | ✅ |
| `hasErrors && !responseCode` | ❌ blocks error view | ✅ |
| `!isDataLoaded` (stale snapshot) | ❌ breaks transitions | ✅ |

src/pages/Search/SearchPageNarrow.tsx
szymonzalarski98: Added `hasErrors` to `reasonAttributes` in both Wide and Narrow for debug parity with baseline.

Regarding `isCardFeedsLoading` - this requires `useOnyx(ONYXKEYS.COLLECTION.SHARED_NVP_PRIVATE_DOMAIN_MEMBER)` which is only available inside Search. The page-level skeleton is only visible when `data === undefined` (cold start before any API response), so `isCardFeedsLoading` is never relevant at this layer — it's always `false` when this skeleton renders. Once data arrives, Search mounts and its internal `shouldShowLoadingState` handles `isCardFeedsLoading` with full context. Adding the Onyx subscription at page level just for a debug attribute would add unnecessary overhead.

src/hooks/useSearchLoadingState.ts
FitseTLT: I am not sure if I agree with this

>isCardFeedsLoading → deadlock: blocking mount prevents the card feed API call (fired inside Search) from ever starting.

because it is only depending on whether the onyx subscription has loaded.
https://github.com/Expensify/App/blob/db145a19d15d2ffdcc0ddc353efa824eaab56afb/src/components/Search/index.tsx#L311
It doesn't need  the card feed API call result, the onyx result will be loaded whether or not there is a card feed already fetched from BE. `isCardFeedsLoading` is only built for the transition time where onyx data is still loading when the page is open.

szymonzalarski98: You're right - I was wrong about the deadlock. `isCardFeedsLoading` depends on the Onyx subscription status, not on an API call inside Search, so it resolves independently of whether Search is mounted.

Added `isCardFeedsLoading` to the page-level `useSearchLoadingState` hook:

```ts
const [, cardFeedsResult] = useOnyx(ONYXKEYS.COLLECTION.SHARED_NVP_PRIVATE_DOMAIN_MEMBER);
const validGroupBy = getValidGroupBy(queryJSON.groupBy);
const isCardFeedsLoading = validGroupBy === CONST.SEARCH.GROUP_BY.CARD && cardFeedsResult?.status === 'loading';

return hasNoData || isCardFeedsLoading;
```

The `isLoadingWithNoData` condition still cannot be at page level (causes mount/unmount oscillation as described above), but `isCardFeedsLoading` is safe since the Onyx subscription resolves on its own.

src/hooks/useSearchPageSetup.ts
szymonzalarski98: This thread is outdated - it references `syncContextWithRoute` / `clearTransactionsAndSetHashAndKey` which no longer exist in the current code. The hook now only calls `clearSelectedTransactions(hash)` on hash change (via both `useFocusEffect` and `useEffect`), which clears the selected transaction checkboxes when navigating to a different search query. This is still needed because the page-level hook runs before Search mounts, ensuring stale selections are cleared immediately.

FitseTLT: I really dont get this change. It is new and doesn't fix a new bug, right? cc @luacmartins

FitseTLT (approved):
LGTM

luacmartins (approved):
LGTM

src/hooks/useSearchPageSetup.ts
```diff
@@ -0,0 +1,67 @@
...
+import useSearchShouldCalculateTotals from './useSearchShouldCalculateTotals';
+
+/**
+ * Handles page-level setup for Search that must happen before the Search component mounts:
+ * - Clears selected transactions when the query changes
+ * - Fires the search() API call so data starts loading alongside the skeleton
+ * - Fires openSearch() to load bank account data
+ * - Re-fires openSearch() when coming back online
+ */
+function useSearchPageSetup(queryJSON: SearchQueryJSON | undefined) {
```
QichenZhu, line 18: `useSearchPageSetup` handles page-level search. However, `<Search>` had an older effect that also called `handleSearch()`. This caused an infinite loop: search, loading skeleton, error, search, ...

Fixed via https://github.com/Expensify/App/pull/95043.

Issue #80553, opened 2026-01-27
[Due for payment 2026-03-31] [ECUK PIN Management] [FE] Change PIN


joekaufmanexpensify:
Part of https://github.com/Expensify/Expensify/issues/523177

Implement the [Release 3: Change PIN (Frontend)](https://docs.google.com/document/d/1G7FL5zPNiShuKXF4AZHqphsv03HgFD1t9J5yljVubIE/edit?tab=t.0#bookmark=id.f98pj0j0fgq2) design doc section

<details><summary>Issue Owner</summary>Current Issue Owner: @DylanDylann</details>

Issue #82162, opened 2026-02-11
[HOLD App#81197] Ensure MFA RHP works over top of other RPHs


rafecolton:
## Problem

When the transaction authorization flow is started, it opens a RHP over top of whatever is happening on the screen previously. When it's complete, or when the user closes it out, we close the RHP. If the user's previous activity was in the RHP, that activity will also get closed, and the user will be returned to the underlying page.

## Why is this important?

Since the transaction authorization flow is intended to interrupt the user's current activity, if the user is unable to return to what they were doing, the loss of progress in the app could lead to frustration and lack of use of whatever feature they were previously exploring.

## Solution

Enable the user to return to what they were doing when the transaction auth flow is complete - including if they were in an RHP.

cc @chuckdries

chuckdries:
Still below other priorities

dariusz-biela:
I'm not sure, but I think I'm fixing this issue in this PR: https://github.com/Expensify/App/pull/89992

When reviewing this PR, could you please check to see if this issue has been resolved?

rafecolton:
Yeah this is indeed resolved by that PR

rafecolton:
Oops, this has been done for a while, should be closed

PR #84120, opened 2026-03-04
Reveal PIN + Change pin

DylanDylann:
### Explanation of Change

This PR implements the "Reveal PIN" feature and  "Change PIN" feature for UK/EU Expensify Card holders
Document: [Release 2 of UK/EU PIN Management](https://docs.google.com/document/d/1G7FL5zPNiShuKXF4AZHqphsv03HgFD1t9J5yljVubIE/edit?tab=t.0#bookmark=id.95f1mfwnybvl)

### Fixed Issues

$ https://github.com/Expensify/App/issues/80553
PROPOSAL:

### Tests

### QA Steps

Assign to @joekaufmanexpensify for QA

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

## Reveal PIN

https://github.com/user-attachments/assets/e3fc93e0-b67c-4edf-b93c-ab1b7d71773f

https://github.com/user-attachments/assets/da915404-f663-42a1-af4e-c422b7fb5d21

## Change PIN

https://github.com/user-attachments/assets/a3b2588e-a604-460a-bf8b-c71dc00de981

https://github.com/user-attachments/assets/c308ea99-0502-4dc3-a16f-04d98629ebf5


DylanDylann:
@chuckdries @JakubKorytko @dariusz-biela Please only review one last commit because I checked out from the reveal-pin branch

src/pages/settings/Wallet/ExpensifyCardPage/index.tsx
```diff
@@ -450,6 +451,21 @@ function ExpensifyCardPage({route}: ExpensifyCardPageProps) {
                                         }
                                     />
                                 )}
+                                {canChangePin && (
+                                    <MenuItem
+                                        title={translate('cardPage.changePin')}
+                                        icon={expensifyIcons.Key}
+                                        shouldShowRightIcon
+                                        onPress={() => {
+                                            const physicalCardID = String(currentPhysicalCard?.cardID);
```
JakubKorytko, line 446: if the `cardID` is `undefined` then `String` method returns `undefined` string

src/pages/settings/Wallet/ExpensifyCardPage/ChangePinPage.tsx
```diff
@@ -0,0 +1,156 @@
...
+                            hasError={!!error}
+                            autoFocus
+                            secureTextEntry={isPinHidden}
+                        />
+                        {!!error && <Text style={[styles.formError, styles.mt2]}>{error}</Text>}
+                    </View>
+
+                    <View style={[styles.flexRow, styles.justifyContentCenter, styles.mb4, styles.alignItemsCenter, styles.w100]}>
+                        <Button
+                            text={isPinHidden ? translate('cardPage.revealPin') : translate('cardPage.hidePin')}
```
JakubKorytko:
```suggestion
                            text={translate(`cardPage.${isPinHidden ? "reveal" : "hide"}Pin`)}
```

src/languages/pl.ts
```diff
@@ -2280,12 +2283,22 @@ ${amount} dla ${merchant} - ${date}`,
...
+        changeYourPin: 'Enter a new PIN for your card.',
+        confirmYourChangedPin: 'Confirm your new PIN.',
         pinMustBeFourDigits: 'PIN musi mieć dokładnie 4 cyfry.',
         invalidPin: 'Wybierz proszę bardziej bezpieczny PIN.',
         pinMismatch: 'Kody PIN nie są takie same. Spróbuj ponownie.',
         revealPin: 'Pokaż PIN',
         hidePin: 'Ukryj PIN',
         pin: 'PIN',
+        changePin: 'Change PIN',
+        pinChanged: 'PIN changed',
```
JakubKorytko: Not sure why all the language files changes are in english?

JakubKorytko (changes requested):
Some comments, otherwise LGTM

src/languages/pl.ts
DylanDylann: I will update translation soon

src/pages/settings/Wallet/ExpensifyCardPage/index.tsx
DylanDylann: It could not be undefined because we only show this button if canChangePin is true

src/pages/settings/Wallet/ExpensifyCardPage/ChangePinPage.tsx
DylanDylann: I think the current approach is fine and not worth updating. Please feel free to raise anything if I missed something.

DylanDylann:
@codex review

src/libs/CardPINStore.ts
```diff
@@ -0,0 +1,44 @@
+import {useSyncExternalStore} from 'react';
+
+/**
+ * In-memory store for revealed card PINs.
+ * PINs must NOT be persisted to disk (PCI compliance), so we use a module-level store
+ * with useSyncExternalStore for React integration instead of Onyx.
```
chuckdries, line 6: I'm curious why this separate store is necessary. We call `setRevealedPIN` in the `RevealPin` scenario callback. We pass it as part of `callbackInput.body`, which comes from `scenarioResponse.body` in `handleCallback`, which comes from `state`, which comes from `useMultifactorAuthenticationState()`. So I guess what I'm wondering is why does the caller of `useRevealedPIN()` not just call `useMultifactorAuthenticationState()` and read it directly? Does that context disappear from the tree after the callback finishes executing or something?

I was going to also remark that useSyncExternalStore is a little unusual for something like this but I've actually convinced myself that it's cleaner than i.e. a useState in a Context. After I thought about it for a while, I am a fan!

src/languages/en.ts
```diff
@@ -2400,12 +2406,22 @@ const translations = {
...
         hidePin: 'Hide PIN',
         pin: 'PIN',
+        pinChanged: 'PIN changed!',
+        pinChangedHeader: 'PIN changed',
+        pinChangedDescription: "You're all set to use your PIN now.",
+        changePin: 'Change PIN',
+        changePinAtATM: 'Change your PIN at any ATM',
+        changePinAtATMDescription: 'This is required in your region. ',
+        changePinAtATMConciergeLink: 'Reach out to Concierge',
+        changePinAtATMSuffix: ' if you have any questions.',
```
chuckdries: I know this is not the only place we do this pattern in our translations, but I'm trying to get people to standardize on using a single string with the `<concierge-link>` custom tag in the string itself. Makes it easier to translate

So for this it'd be `'This is required in your region. <concierge-link>React out to Concierge</concierge-link> if you have any questions.'`

src/pages/settings/Wallet/ExpensifyCardPage/ChangePINATMPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+import Text from '@components/Text';
+import TextLink from '@components/TextLink';
+import useLocalize from '@hooks/useLocalize';
+import useThemeStyles from '@hooks/useThemeStyles';
+import Navigation from '@libs/Navigation/Navigation';
+import variables from '@styles/variables';
+import ROUTES from '@src/ROUTES';
+
+function ChangePinATMPage() {
+    const {translate} = useLocalize();
```
chuckdries, line 7: Once we make the above change to the translation strings, do you think we can refactor ChangePinATMPage to use `createScreenWithDefaults(OutcomeScreenBase, ...)`?

chuckdries (commented):
High level question about CardPinStore, and a note about translations, but overall looking pretty solid to me! I notice the pin input is the wide version, will we get the narrower version for free by merging main?

src/pages/settings/Wallet/ExpensifyCardPage/ChangePINATMPage.tsx
DylanDylann: We can do that, but it doesn’t really help with DRY. I think we should keep the current approach for readability.

src/libs/CardPINStore.ts
DylanDylann: The MFA context will be closed after the scenario is completed, but we still need the PIN to be displayed afterward. Therefore, we can’t store the PIN in the MFA context. On the other hand, card PINs are sensitive data subject to PCI compliance, I think storing it separately in a dedicated in-memory store would be safer.

cc @JakubKorytko for more thoughts

src/components/MultifactorAuthentication/config/scenarios/ChangePIN.tsx
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Payload type for the CHANGE_PIN scenario.
+ * Contains the new PIN and cardID for the card whose PIN is being changed.
+ */
+type Payload = {
+    pin: string;
+    cardID: string;
+};
+
+const AuthenticationCanceledFailureScreen = createScreenWithDefaults(
```
rafecolton: This isn't a cancelation-specific error screen for any of the PIN flows, but we do need to define the default and server failure screens.

src/components/MultifactorAuthentication/config/scenarios/ChangePIN.tsx
```diff
@@ -0,0 +1,60 @@
...
+    action: changePINForCard,
+
+    callback: async () => {
+        return CONST.MULTIFACTOR_AUTHENTICATION.CALLBACK_RESPONSE.SHOW_OUTCOME_SCREEN;
+    },
+
+    successScreen: <ChangePINSuccessScreen />,
+
+    failureScreens: {
+        [CONST.MULTIFACTOR_AUTHENTICATION.REASON.EXPO.CANCELED]: <AuthenticationCanceledFailureScreen />,
```
rafecolton: Per comment above, we don't need a cancelation-specific one, and we DO need generic client and server screens.

src/components/MultifactorAuthentication/config/scenarios/RevealPin.tsx
```diff
@@ -0,0 +1,68 @@
...
+type Payload = {
+    cardID: string;
+};
+
+/**
+ * Type guard to verify the payload is a RevealPin payload.
+ */
+function isRevealPinPayload(payload: MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined): payload is Payload {
+    return !!payload && 'cardID' in payload;
+}
```
rafecolton: Is this meant to verify the payload we are sending to the server or the one we are receiving in the response? If the former, why don't we have a similar function for `ChangePIN`? If the latter, then why are we checking `cardID` and not `pin`, given `pin` is included in the response and `cardID` is not?

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+/**
+ * Type guard to verify the payload is a RevealPin payload.
+ */
+function isRevealPinPayload(payload: MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined): payload is Payload {
+    return !!payload && 'cardID' in payload;
+}
+
+const AuthenticationCanceledFailureScreen = createScreenWithDefaults(
```
rafecolton: Same comment as in ChangePIN RE needing to define the default client and server failure screens

src/components/MultifactorAuthentication/config/scenarios/ChangePin.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * - Authentication failure: Return SHOW_OUTCOME_SCREEN to show failure screen
+ */
+export default {
+    allowedAuthenticationMethods: [CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS],
+    action: changePINForCard,
+
+    callback: async () => {
+        return CONST.MULTIFACTOR_AUTHENTICATION.CALLBACK_RESPONSE.SHOW_OUTCOME_SCREEN;
+    },
+
```
rafecolton: Does callback need to be defined at all given it is not modifying the default behavior?

```suggestion
 */
export default {
    allowedAuthenticationMethods: [CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS],
    action: changePINForCard,
```

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
```diff
@@ -0,0 +1,68 @@
...
+ */
+export default {
+    allowedAuthenticationMethods: [CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS],
+    action: revealPINForCard,
+    callback: async (isSuccessful, callbackInput, payload) => {
+        if (isSuccessful && isRevealPinPayload(payload)) {
+            const pin = callbackInput.body?.pin as string | undefined;
+            if (pin) {
+                setRevealedPIN(payload.cardID, pin);
+            }
```
rafecolton: Is it possible to handle the absence of `pin` in the body in the action, so that by the time we reach this code, pin is always set to a valid string if `isSuccessful` is true?

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
```diff
@@ -0,0 +1,68 @@
...
+            Navigation.closeRHPFlow();
+            Navigation.navigate(ROUTES.SETTINGS_WALLET_DOMAIN_CARD.getRoute(String(payload.cardID)));
+            return CONST.MULTIFACTOR_AUTHENTICATION.CALLBACK_RESPONSE.SKIP_OUTCOME_SCREEN;
+        }
+
+        return CONST.MULTIFACTOR_AUTHENTICATION.CALLBACK_RESPONSE.SHOW_OUTCOME_SCREEN;
+    },
+
+    failureScreens: {
+        [CONST.MULTIFACTOR_AUTHENTICATION.REASON.EXPO.CANCELED]: <AuthenticationCanceledFailureScreen />,
```
rafecolton: Same comment as in ChangePIN

src/languages/en.ts
```diff
@@ -769,6 +769,12 @@ const translations = {
         setPin: {
             didNotShipCard: "We didn't ship your card. Please try again.",
         },
+        revealPin: {
+            authenticationCanceled: 'We couldn’t reveal your PIN. Please try again.',
+        },
+        changePin: {
+            authenticationCanceled: 'We didn’t change your PIN. Please try again.',
```
rafecolton: Avoid smart quotes/apostrophes please

```suggestion
            authenticationCanceled: 'We couldn't reveal your PIN. Please try again.',
        },
        changePin: {
            authenticationCanceled: 'We didn't change your PIN. Please try again.',
```

src/libs/actions/MultifactorAuthentication/index.ts
```diff
@@ -240,6 +240,39 @@ async function setPersonalDetailsAndShipExpensifyCardsWithPIN(params: Multifacto
...
+async function revealPINForCard({cardID, signedChallenge, authenticationMethod}: MultifactorAuthenticationScenarioParameters['REVEAL-PIN']) {
+    try {
+        const response = await makeRequestWithSideEffects(SIDE_EFFECT_REQUEST_COMMANDS.REVEAL_CARD_PIN, {cardID, signedChallenge: JSON.stringify(signedChallenge), authenticationMethod}, {});
+
+        const {jsonCode, message, pin} = response ?? {};
+        const parsed = parseHttpRequest(jsonCode, CONST.MULTIFACTOR_AUTHENTICATION.API_RESPONSE_MAP.REVEAL_CARD_PIN, message);
+
+        return {
+            ...parsed,
+            body: {pin: String(pin ?? '')},
```
rafecolton, line 252: I think it's possible for this to succeed and return nothing if no PIN is set. 🤔 While that would technically be valid, I'm concerned it would look broken to the user. There's a few options for how we can handle this:
1. Show nothing in place of the dots on the card details page
2. Show a placeholder string (e.g. `(not set)`)
3. Show an outcome page that shows the user no PIN is set

The first one is the simplest, but I'm concerned it would look broken to the user. The second one would be nearly as simple, it would just need a translation key. The third one would be a bit more scope, since we'd need a design, albeit a simple one.

What do you think @joekaufmanexpensify

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -183,7 +186,21 @@ const API_RESPONSE_MAP = {
     },
 
     SET_PERSONAL_DETAILS_AND_SHIP_EXPENSIFY_CARDS_WITH_PIN: {
-        [HTTP_STATUS.SUCCESS]: REASON.BACKEND.AUTHORIZATION_SUCCESSFUL,
+        [HTTP_STATUS.SUCCESS]: REASON.BACKEND.SET_PIN,
+        [HTTP_STATUS.CLIENT_ERROR]: {
+            ...MULTIFACTOR_AUTHENTICATION_COMMAND_BASE_CLIENT_ERRORS,
+        },
```
rafecolton: Does this need to be included if no errors are handled specially?

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -183,7 +186,21 @@ const API_RESPONSE_MAP = {
...
+    },
+
+    REVEAL_CARD_PIN: {
+        [HTTP_STATUS.SUCCESS]: REASON.BACKEND.PIN_REVEALED,
+        [HTTP_STATUS.CLIENT_ERROR]: {
+            ...MULTIFACTOR_AUTHENTICATION_COMMAND_BASE_CLIENT_ERRORS,
+        },
```
rafecolton: Same question here

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
```diff
@@ -183,7 +186,21 @@ const API_RESPONSE_MAP = {
...
+    },
+
+    CHANGE_CARD_PIN: {
+        [HTTP_STATUS.SUCCESS]: REASON.BACKEND.PIN_CHANGED,
         [HTTP_STATUS.CLIENT_ERROR]: {
             ...MULTIFACTOR_AUTHENTICATION_COMMAND_BASE_CLIENT_ERRORS,
         },
```
rafecolton: Same here

src/libs/CardPINStore.ts
rafecolton: > Does that context disappear from the tree after the callback finishes executing or something?

Worth noting that the context does remain in the tree after concluding, as long as we don't close the RHP, but I don't think we should count on that behavior since it causes [this bug](https://github.com/Expensify/App/issues/82162)

src/libs/CardPINStore.ts
```diff
@@ -0,0 +1,44 @@
...
+const listeners = new Set<Listener>();
+let revealedPINs: Record<string, string> = {};
+
+function subscribe(listener: Listener) {
+    listeners.add(listener);
+    return () => listeners.delete(listener);
+}
+
+function notifyListeners() {
+    for (const listener of listeners) {
```
rafecolton, line 20: Is `const` correct here? In other languages, if there were multiple listeners, setting this as `const` would cause an error on the second iteration

rafecolton (changes requested):
NAB rename the scenario files from `ChangePin` to `ChangePIN` and `RevealPin` to `RevealPIN`

src/libs/CardPINStore.ts
DylanDylann: @rafecolton Currently, to resolve the bug you pointed out. We need to close RHP before navigating 

>             Navigation.closeRHPFlow();
>             Navigation.navigate(ROUTES.SETTINGS_WALLET_DOMAIN_CARD.getRoute(String(payload.cardID)));

So the PIN will be removed if we save in the context

src/components/MultifactorAuthentication/config/scenarios/ChangePIN.tsx
rafecolton: 🤔 we already did a design on this. We don't need a new illustration, we just need to define the element in the config with the correct copy.

DylanDylann: >  we just need to define the element in the config with the correct copy.

Yeah, it's my mean. Please wait a moment.

DylanDylann: @rafecolton Does that mean it’s correct to use a separate screen here, and now we need another copy?

src/components/MultifactorAuthentication/config/scenarios/ChangePin.tsx
DylanDylann: Agree, currently we don't need to define this callback

src/components/MultifactorAuthentication/config/scenarios/RevealPin.tsx
DylanDylann: To be honest, this function is only used to verify the payload type for the Reveal PIN scenario. Since the payload uses a generic type, we don’t define a specific payload type for each scenario. We need to check cardID because we use cardID in the callback function. 

For a more robust approach, `payload` should have the correct type for each scenario. I suggest handling this as a separate refactor of the MFA context.

rafecolton: That doesn't really answer my question 😅

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
DylanDylann: ~I assume the PIN is always included in the response when the status code is 2xx.~

Discussing below: https://github.com/Expensify/App/pull/84120#discussion_r2937698662

src/libs/CardPINStore.ts
DylanDylann: Yes, this part is correct. In JavaScript / TypeScript, when we write:
```
for (const x of arr) {
  console.log(x);
}
```
The engine creates a new binding of x for each iteration. So x is not reassigned. Instead, a new block-scoped variable is created every loop iteration.


This differs from patterns where a single variable is reused and reassigned across iterations. Because of this behavior, using `const` in for...of loops [is common](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
DylanDylann: ~Agreed that it’s unnecessary, but I’d prefer to keep it so the detailed reason can be returned to the main process, which would make debugging clearer. That said, it’s also fine to remove it if you think it’s redundant.~

src/components/MultifactorAuthentication/config/scenarios/RevealPin.tsx
DylanDylann: Ahh my bad. 

> Is this meant to verify the payload we are sending to the server or the one we are receiving in the response?

`payload` is what we are sending to the server. 

> why don't we have a similar function for ChangePIN?

nvm, I removed the callback function in the changePIN scenario

DylanDylann: Verifying the payload here only for type safety. This means that if we don’t use the payload in the callback, we don’t need to define this function.

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * Callback behavior:
+ * - Success: Store the revealed PIN in CardPINStore and return SKIP_OUTCOME_SCREEN
+ * - Authentication failure: Return SHOW_OUTCOME_SCREEN to show failure screen
+ */
+export default {
+    allowedAuthenticationMethods: [CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS],
+    action: revealPINForCard,
+    callback: async (isSuccessful, callbackInput, payload) => {
+        if (isSuccessful && isRevealPINPayload(payload)) {
+            const pin = callbackInput.body?.pin as string | undefined;
```
JakubKorytko:
```suggestion
            const pin = typeof callbackInput.body?.pin === 'string' ? callbackInput.body.pin : undefined;
```

src/components/MultifactorAuthentication/config/scenarios/index.ts
```diff
@@ -4,9 +4,13 @@ import CONST from '@src/CONST';
...
 import BiometricsTest from './BiometricsTest';
+import type {Payload as ChangePINPayload} from './ChangePIN';
+import ChangePIN from './ChangePIN';
 import {customConfig} from './DefaultUserInterface';
-import type {Payload as SetPinOrderCardPayload} from './SetPinOrderCard';
-import SetPinOrderCard from './SetPinOrderCard';
+import type {Payload as RevealPINPayload} from './RevealPIN';
+import RevealPIN from './RevealPIN';
+import type {Payload as SetPinOrderCardPayload} from './SetPINOrderCard';
+import SetPinOrderCard from './SetPINOrderCard';
```
JakubKorytko:
```suggestion
import SetPINOrderCard from './SetPINOrderCard';
```

src/libs/CardPINStore.ts
```diff
@@ -0,0 +1,44 @@
...
+        listener();
+    }
+}
+
+function setRevealedPIN(cardID: string, pin: string) {
+    revealedPINs = {[cardID]: pin};
```
JakubKorytko, line 27: That's a strong behavior; future readers might think it's a bug. Add a short comment or JSDoc explaining that only one PIN is intentionally kept at a time (for security/UX).

src/libs/CardPINStore.ts
JakubKorytko: Thanks for looping me in!
My understanding is that the separate CardPINStore is mainly about lifecycle and sensitivity of the PIN:
- The MFA flow is short-lived. Once the scenario completes and we close the RHP + navigate to the wallet card screen, the MFA context is torn down, so anything stored there isn't available on the destination screen. We still need the revealed PIN there, so reading it directly from `useMultifactorAuthenticationState()` wouldn't be reliable.
- Since the PIN is PCI-sensitive, keeping it in a small, dedicated in-memory store (with "only one PIN at a time" semantics) seems safer and easier to reason about than putting it in a broader context that might outlive the specific flow or be reused for other scenarios.

Given that, the `useSyncExternalStore` approach for this one-PIN cache makes sense to me.

JakubKorytko (commented):
Some NITs

src/libs/actions/MultifactorAuthentication/index.ts
joekaufmanexpensify: Discussing [here](https://expensify.slack.com/archives/C09AP9F437A/p1773666653616439?thread_ts=1773666360.042339&cid=C09AP9F437A)

src/components/MultifactorAuthentication/config/scenarios/ChangePIN.tsx
rafecolton: It means:
* We need to define `ChangePINClientFailureScreen` and `ChangePINServerFailureScreen` which override the relevant default elements
* We shouldn't need an explicit failure screen for cancelation, that can use the default client failure screen

Commenting with further specifics in the [Slack thread](https://expensify.slack.com/archives/C09AP9F437A/p1773666215508509)

DylanDylann: Thanks. I updated

src/components/MultifactorAuthentication/config/scenarios/ChangePIN.tsx
DylanDylann: Updated

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
DylanDylann: Updated

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
DylanDylann: Updated

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
DylanDylann: Updated

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
DylanDylann: Updated

src/libs/MultifactorAuthentication/Biometrics/VALUES.ts
DylanDylann: Updated

src/libs/actions/MultifactorAuthentication/index.ts
rafecolton: We decided to just let it be empty and log a bugbot in auth: https://github.com/Expensify/Auth/pull/20483

src/components/MultifactorAuthentication/config/scenarios/RevealPin.tsx
chuckdries: To be precise, this type guard validates that the payload we pass the callback is the same type as the payload that we pass to `executeScenario`. There's no chance we ever will not pass this type guard at runtime, but when I established this pattern I couldn't figure out how to express in the type system that "I promise the scenario callback will only ever be called with the payload associated with this scenario"

src/languages/en.ts
```diff
@@ -775,6 +775,12 @@ const translations = {
         setPin: {
             didNotShipCard: "We didn't ship your card. Please try again.",
         },
+        revealPin: {
+            authenticationCanceled: "We couldn't reveal your PIN. Please try again.",
+        },
+        changePin: {
+            authenticationCanceled: "We didn't change your PIN. Please try again.",
```
rafecolton: NAB Since these translations are not specific to cancelation but more general for all failure methods, suggest following the pattern of `setPin` and making the translation keys general

```suggestion
            couldNotReveal: "We couldn't reveal your PIN. Please try again.",
        },
        changePin: {
            didNotChange: "We didn't change your PIN. Please try again.",
```

src/libs/actions/MultifactorAuthentication/index.ts
rafecolton: This can be resolved now since we agreed to allow this to be empty string

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
```diff
@@ -0,0 +1,75 @@
...
+ */
+export default {
+    allowedAuthenticationMethods: [CONST.MULTIFACTOR_AUTHENTICATION.TYPE.BIOMETRICS],
+    action: revealPINForCard,
+    callback: async (isSuccessful, callbackInput, payload) => {
+        if (isSuccessful && isRevealPINPayload(payload)) {
+            const pin = typeof callbackInput.body?.pin === 'string' ? callbackInput.body.pin : undefined;
+            if (pin) {
+                setRevealedPIN(payload.cardID, pin);
+            }
```
rafecolton: Given that we force pin to be a string below and we will allow empty string to be used here, can this block be simplified? If not, then in the edge case that pin is empty, we will still show the `••••` placeholder.

NAB because this is an unlikely edge case

```suggestion
            const pin = typeof callbackInput.body?.pin ?? '';
            setRevealedPIN(payload.cardID, pin);
```

src/libs/CardPINStore.ts
rafecolton: Thanks, TIL!

src/pages/MultifactorAuthentication/OutcomePage.tsx
```diff
@@ -5,7 +5,7 @@ import type {ErrorState} from '@components/MultifactorAuthentication/Context/Sta
 
 /** Server failure screen generally represents "unknown error" so also show when status is unknown (e.g. network/parse error). */
 function isServerError(error: ErrorState): boolean {
-    return error.httpStatusCode === undefined || error.httpStatusCode >= 500;
+    return !!error.httpStatusCode && error.httpStatusCode >= 500;
```
rafecolton: 👎 this is not correct. This changes the logic so that undefined local errors are treated as client client failures, when they should be treated as server errors (i.e. show humpty dumpty). Similarly, the lines you added below should not be needed either. I wonder if I was wrong above and you do need to define certain default errors as client errors. @dariusz-biela can you please advise on this one?

rafecolton (changes requested):
Tested locally and it worked well overall - including with passkey 😄 I'm a little concern that errors are not handled correctly so I tagged Dariusz for a second opinion.

src/pages/MultifactorAuthentication/OutcomePage.tsx
DylanDylann: @rafecolton  httpStatusCode is not set if the user cancels the flow themselves, and I think we should show the client error screen in this case.

src/components/MultifactorAuthentication/config/scenarios/RevealPIN.tsx
DylanDylann: I’ll need to update this again because currently we only show the PIN if it’s available.

DylanDylann: Updated

src/pages/MultifactorAuthentication/OutcomePage.tsx
DylanDylann: I’ve reverted the changes in this file, but I’d like to get more thoughts on the concern above

dariusz-biela: I apologize for replying to this thread so late, I didn't notice the notification.

Okay, I see both of your perspectives, and I'm actually inclined to think we could do a different approach that would potentially have a better user experience.

We should divide errors into two types: 

For errors caused by the user:
1. rate limit
2. cancel system prompt
3. timeout
4. bad password
5. all 400 errors
6. etc.

And errors caused by a bugs in our implementation:
1. unexpected frontend error
2. 500 error from the server
3. etc.


I think we have already partially done this by:

```tsx
/** Known errors the user is likely to encounter (cancellations, expired transactions, unsupported devices, etc.). Logged at 'info' level. */
const ROUTINE_FAILURES = new Set<ReasonValue>([

/** Known errors that should rarely happen and may indicate a bug or unexpected state. Logged at 'error' level. Any reason not in either set is treated as UNCLASSIFIED (e.g. 5xx, missing reason). */
const ANOMALOUS_FAILURES = new Set<ReasonValue>([
```

dariusz-biela: If we used ROUTINE_FAILURES and ANOMALOUS_FAILURES we would also have an indicator in the UI what type of log went to sentry and where to look for it.

rafecolton (approved):
This works well! Tested it with passkey on desktop, then turned off passkey support and observed that the correct error page is now shown.

Image

Issue #78241, opened 2025-12-21
[$500] Invoice shows waiting for user to add a bank account despite payer already having a verified bank account


izarutskaya:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:**
**Reproducible in staging?:** Needs Reproduction (Reproduction blocked due to card expense)
**Reproducible in production?:** Needs Reproduction 
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @stephanieelliott 
**Slack conversation** (hyperlinked to channel name): [#expensify_bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1766110788975549)

## Action Performed:
Precondition: Have 2 accounts on the same workspace, one with a verified business bank account

1. From the account with a VBBA, send an invoice to the other account (payer)
2. As the payer, pay the invoice with a credit card
3. As the payee/VBBA owner, see the message: [payer] started payment, but is waiting for [payee/VBBA account] to add a bank account.

## Expected Result:
To have the payment sent to the payee's linked VBBA

## Actual Result:
Payee sees a message to add VBBA despite them already having one

## Workaround:
Unknown

## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

Additional information in the [OP](https://expensify.slack.com/archives/C049HHMV9SM/p1766110788975549)

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022002912372922159577</li>
        <li>Upwork Job ID: 2002912372922159577</li>
        <li>Last Price Increase: 2026-04-07</li>
    </ul>
</details>

[...]

PR #84221, opened 2026-03-04
Fix invoice showing 'waiting for bank account' when user has VBBA (#78241)

Walle2131235:
## Fixed Issues
- $ #78241

## Changes
- Modified `src/libs/ReportUtils.ts` to check if policy has a verified business bank account (VBBA) before showing the 'waiting for bank account' message
- Added VBBA check: `!!policy?.achAccount?.bankAccountID && policy.achAccount.state === CONST.BANK_ACCOUNT.STATE.OPEN`
- Only shows 'waiting for bank account' message when user actually doesn't have a VBBA set up

## Tests
- TypeScript compilation: ✅ Pass
- Logic: Checks policy.achAccount state matches pattern used elsewhere in codebase (SearchUIUtils.ts, WorkspaceWorkflowsPage.tsx)

## Why This Fix Is Needed
When a payer pays an invoice with a credit card, users with an existing Verified Business Bank Account (VBBA) were incorrectly seeing the message: '[payer] started payment, but is waiting for [payee] to add a bank account.'

This fix ensures the message only appears when the user actually needs to add a bank account.

## Related
- Issue reported by: @stephanieelliott
- Issue owner: @huult

### DIFF
--- a/FIX_SUMMARY.md
+++ b/FIX_SUMMARY.md
@@ -1,0 +1,83 @@
+# Expensify App - Bug Fix Submission
+
+## Issue
+**Issue #78241:** [$500] Invoice shows "waiting for user to add a bank account" despite payer already having a verified bank account
+
+**Bounty:** $500 USD
+
+## Problem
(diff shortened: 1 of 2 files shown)


huult:
Please don't open a PR before it is approved. Let's close this PR.

Issue #82373, opened 2026-02-13
[$250] App stuck displaying WS chat pg in loop on tapping back button after deleting reports


jponikarchuk:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.18-2
**Reproducible in staging?:** Yes
**Reproducible in production?:** Yes
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** Exploratory around https://github.com/Expensify/App/pull/81741
**Email or phone of affected tester (no customers):** kiatest.80+ck2122612@gmail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** iPhone 16/iOS 18.7.1, Pixel 8a/Android 16
**App Component:** Money Requests

## Action Performed:
1. Open Hybrid app
2. Join as new user with gmail account (Select Somthing else during onboarding)
3. Go to Self DM & create a manual expense
4. Open the expense >> Tap More >> Duplicate option
5. Go to Reports
6.Open the second expense
7. Select Report field >> Upgrade >> Confirm >>  Got it, thanks
8. Tap create new report
9. Open the second expense
10. Tap From link
11. Tap + icon >> create new report (This is an optional step)
12. Delete both report(s) via contextual menu (or delete it from expense details page)
13. Tap back button & observe workspace chat page is displayed
14. Repeat step 13 few more times & observe workspace chat page displayed in loop
15. Swipe page to right using swipe gesture -  observe not here page appears & return to WS chat page

## Expected Result:
App should not stuck in workspace chat page & user should be redirected to Inbox after deleting reports

## Actual Result:
App get stuck displaying same workspace chat page in loop on tapping back button after deleting reports

## Workaround:
Unknown

## Platforms:
- [x] Android: App
- [ ] Android: mWeb Chrome
- [x] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [ ] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/a9a1dece-f214-4ee6-865a-de55dc4132ec

[...]

PR #84687, opened 2026-03-10
[$250] Prevent back-stack loop when deleting currently viewed report

zscy0710:
### Please re-state the problem that we are trying to solve in this issue.
After deleting reports, pressing back can loop on workspace chat / not-here states instead of returning cleanly to Inbox.

### What is the root cause of that problem?
The current report route can remain on stack while the underlying report data is deleted, leaving stale navigation targets.

### What changes did you make to solve the problem?
- Before deleting a report, if it is currently displayed, pop navigation to sidebar first.
- Then proceed with report/onyx deletion as before.

### What specific scenarios should we cover in automated tests to prevent reintroducing this issue in the future?
- Deleting currently viewed expense/report should not leave back-stack loops.
- Back navigation after deletion should return to a valid screen (Inbox/sidebar) instead of stale report routes.

## Testing
- [ ] Not run (navigation safety change)

### DIFF
--- a/src/libs/actions/Report/index.ts
+++ b/src/libs/actions/Report/index.ts
@@ -3740,6 +3740,13 @@ function deleteReport(reportID: string | undefined, shouldDeleteChildReports = f
         Log.warn('[Report] deleteReport called with no reportID');
         return;
     }
+
+    // If the currently displayed report is being removed, pop to the sidebar first.
+    // This prevents back-stack loops on a now-missing report route.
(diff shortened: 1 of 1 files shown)


zscy0710:
Thanks for the heads-up. I've posted the required proposal on the issue per CONTRIBUTING guidelines: https://github.com/Expensify/App/issues/82373#issuecomment-4028818944

If this approach looks good to C+, I can reopen or submit an updated PR accordingly.

Issue #80759, opened 2026-01-28
[Due for payment 2026-03-30] [Due for payment 2026-02-26] Investigate and fix sequential queue issues


mjasikowski:
Coming from https://github.com/Expensify/App/issues/69669, we're currently encountering problems with sequential queue that take significant amount of engineering time (by that I mean months) to fix.

This is a tracking issue, so we can investigate, share our findings and propose further actions.

<details><summary>Issue Owner</summary>Current Issue Owner: @adhorodyski</details>

mjasikowski:
@adhorodyski will be taking care of this ASAP

adhorodyski:
Sharing as WIP.

#### Test Files Overview

| File | Tests | Focus |
|---|---|---|
| [`tests/unit/SequentialQueueTest.ts`](https://github.com/Expensify/App/blob/main/tests/unit/SequentialQueueTest.ts) | 13 | Push, conflict resolution, `persistWhenOngoing`, queue flushed data |
| [`tests/unit/PersistedRequests.ts`](https://github.com/Expensify/App/blob/main/tests/unit/PersistedRequests.ts) | 7 | Save, remove, processNext, update, ongoing request |
| [`tests/unit/APITest.ts`](https://github.com/Expensify/App/blob/main/tests/unit/APITest.ts) | ~20 | Full pipeline: offline persistence, retry, reauth, ordering, conflict resolution |
| [`tests/unit/RequestConflictUtilsTest.ts`](https://github.com/Expensify/App/blob/main/tests/unit/RequestConflictUtilsTest.ts) | ~19 | Conflict resolution strategies (dedup, comment deletion, feature enable) |
| [`tests/unit/NetworkTest.tsx`](https://github.com/Expensify/App/blob/main/tests/unit/NetworkTest.tsx) | ~14 | Network state, offline/online transitions, reauth, retries |
| [`tests/unit/RequestTest.ts`](https://github.com/Expensify/App/blob/main/tests/unit/RequestTest.ts) | 2 | Middleware registration/chaining |
| [`tests/unit/MiddlewareTest.ts`](https://github.com/Expensify/App/blob/main/tests/unit/MiddlewareTest.ts) | 6 | HandleUnusedOptimisticID middleware |

---

#### Issue 1: Optimistic data applied before request is persisted

[...]

PR #84622, opened 2026-03-09
[No QA] Fix lost/duplicated requests in PersistedRequests (Issues 2, 3, 4, 5)

szymonzalarski98:
### Explanation of Change

# Sequential Queue Issues — Status Report

### Issue 3: processNextRequest() and rollbackOngoingRequest() don't persist to disk
**File:** `src/libs/actions/PersistedRequests.ts`

**Problem:** When a request moves from the queue to `ongoingRequest`, neither the updated queue nor the ongoing request is written to disk. If the app crashes mid-flight, the ongoing request is lost (3a) and the queue on disk is stale (3c).

**Fix:** Both `processNextRequest()` and `rollbackOngoingRequest()` now use `Onyx.multiSet()` to atomically persist the updated queue and the ongoing request to disk.

### Issue 4: Connect callback overwrites in-memory state with stale disk data
**File:** `src/libs/actions/PersistedRequests.ts`

**Problem:** The `connectWithoutView` callback blindly overwrites the in-memory `persistedRequests` array with whatever Onyx returns from disk. When multiple `Onyx.set()` calls resolve out of order, the last callback wins — potentially reverting the in-memory state to a stale snapshot and permanently losing requests.

**Fix:** After initialization, the callback is a no-op for non-null values: `if (isInitialized && val != null) { return; }`. In-memory state is authoritative. `Onyx.clear()` (which sends `null`) is still allowed through so the module resets properly for tests and app restarts.

### Issue 5: Conflict resolver sees stale queue
**Fixed automatically by Issue 4.** Since in-memory state is no longer overwritten by stale disk callbacks, the conflict resolver in `prepareRequest()` and `push()` always sees the complete, up-to-date queue.

### Issue 2: Nothing in the write pipeline is awaited
**Files:** `src/libs/actions/PersistedRequests.ts`, `src/libs/Network/SequentialQueue.ts`, `src/libs/API/index.ts`

**Problem:** `API.write()` returned `Promise.resolve()` immediately. `PersistedRequests.save()` was void (didn't return the `Onyx.set()` promise). The caller had zero guarantee that the request was persisted to disk.

**Fix:**
- `save()` now returns `Promise<void>` (the `Onyx.set()` promise)
- `push()` returns a `persistencePromise` from `save()` (but calls `flush()` synchronously to preserve write-before-read ordering)
- `processRequest()` is now `async` and awaits `pushToSequentialQueue()`
- `await API.write()` now guarantees the request is persisted to disk

---

### The Problem

[...]

PR #84760, opened 2026-03-10
fix network state detection

adhorodyski:
@mjasikowski @mountiny @TMisiukiewicz @OlimpiaZurek 

### Explanation of Change

The existing `RecheckConnection` system was ad-hoc: a single middleware that re-ran `NetInfo.fetch()` after every failed request with no coordination, a grab-bag of `isOffline`/`shouldForceOffline`/`shouldFailAllRequests` flags scattered across Onyx, and no unified model for what "offline" means. This PR replaces it with a structured network state detection architecture:

**Central state machine (`NetworkState.ts`)** — single source of truth for offline status with four triggers: no radio, sustained API failures, force-offline (debug tool), and simulated poor connection (test tool). All offline decisions flow through `updateState()` instead of being scattered across files.

**Sustained failure detection (`FailureTracker.ts` + `FailureTracking` middleware)** — replaces the old `RecheckConnection` middleware. Counts consecutive API failures with a dual threshold (3 failures over 10s) to distinguish transient errors from genuine outages. One successful request clears everything.

**Simplified Onyx network state** — the `Network` onyx type is cleaned up: `shouldForceOffline` stays (debug tool), `shouldFailAllRequests` and `shouldSimulatePoorConnection` stay (test tools), but all the derived/redundant flags are removed. `isOffline` is now set exclusively by `NetworkState`.

**NetInfo as the reachability layer** — NetInfo is configured with `useNativeReachability: false` so it uses JS fetch polling (`api/Ping`) on all platforms instead of trusting native OS reachability. This aligns behavior across web and mobile. NetInfo's default polling intervals apply (60s when reachable, 5s when unreachable). The `isInternetReachable` transition listener detects recovery.

**Reconnect action (`Reconnect.ts`)** — extracted from the old `NetworkConnection` into a focused module. Called on recovery and app foreground to sync data via `openApp`/`reconnectApp`.

**Synchronous offline reads (`useNetwork` + `useSyncExternalStore`)** — `useNetwork` now reads `isOffline` synchronously from `NetworkState` via `useSyncExternalStore` instead of subscribing to Onyx. This eliminates the render-lag from async Onyx reads and ensures components see the correct offline state immediately.

**Offline→online queue flush (`Reconnect.ts`)** — `Reconnect.ts` subscribes to `NetworkState` and calls `SequentialQueue.flush()` on offline→online transitions. `SequentialQueue` itself reads `getIsOffline()` synchronously for its guard checks but no longer owns the transition subscription. `pause()`/`unpause()` remain for server-requested pauses and data gap resolution, but the offline/online lifecycle is now driven by `NetworkState`.

**`NetworkConnection.ts` deleted** — all its responsibilities (NetInfo configuration, Onyx subscriptions, poor connection simulation, `getDBTimeWithSkew`) are consolidated into `NetworkState.ts`, eliminating the split between the two modules.

**Background location task safety** — the background location tracking task uses `NetInfo.fetch()` directly instead of the in-memory `NetworkState.isOffline()`, so it works correctly in Android headless JS contexts where module-level state hasn't been populated.

Architecture documentation added at `contributingGuides/NETWORK_STATE_DETECTION.md`.

### Fixed Issues
$ https://github.com/Expensify/App/issues/80759
PROPOSAL:

### Tests

**1. Basic offline/online detection (Web)**
1. Open the app on web (`https://dev.new.expensify.com:8082/`)
2. Log in with any account
3. Open browser DevTools → Network tab
4. Set network to **Offline** (DevTools throttling or disconnect WiFi)
5. Verify the app shows the offline indicator within ~15 seconds
6. Verify the JS console shows `[NetworkState] Hard stop: NO_RADIO` log
7. Re-enable network
8. Verify the offline indicator disappears
9. Verify the JS console shows `[NetworkState] Internet reachability restored` log
10. Verify no errors appear in the JS console

**2. Basic offline/online detection (Mobile — iOS & Android)**
1. Open the app on a physical device or simulator
2. Log in with any account
3. Enable **Airplane Mode**
4. Verify the app shows the offline indicator
5. Disable Airplane Mode
6. Verify the offline indicator disappears and the app reconnects (data refreshes)
7. Verify no errors appear in the JS console

**3. Sustained failure detection (Web)**
1. Open the app on web, log in
2. Open DevTools → Console, watch for `[FailureTracker]` logs
3. Open the **Test Tool Menu** (Settings → Troubleshoot → Test tools or shake gesture)
4. Toggle **"Simulate failing network requests"** ON
5. Perform actions that trigger API requests (send a chat message, navigate between reports)
6. Observe `[FailureTracker] Failure #1`, `#2`, `#3` logs in the console
7. After ≥3 failures spanning ≥10 seconds, verify `[FailureTracker] Sustained failure threshold reached` appears
8. Verify the app transitions to offline state (offline indicator shows)
9. Toggle **"Simulate failing network requests"** OFF
10. Perform another action that triggers an API request
11. Verify `[FailureTracker] Success after N failures — resetting tracker` appears
12. Verify the app returns to online state

**4. Force offline toggle (Test Tool Menu)**
1. Open the app, log in
2. Open the **Test Tool Menu**
3. Toggle **"Force offline"** ON
4. Verify the app immediately shows the offline indicator
5. Verify pending requests are paused (send a chat message — it should appear optimistically but not sync)
6. Toggle **"Force offline"** OFF
7. Verify the app goes back online and the queued message syncs

**5. App foreground recovery**
1. Open the app on mobile, log in
2. Verify some data is loaded (reports list visible)
3. Background the app (press Home)
4. Wait ~30 seconds
5. Bring the app to foreground
6. Verify the console shows `[NetworkState] App became active`
7. Verify data is refreshed (any new reports/messages appear)

**6. Recovery after radio loss + restore**
1. Open the app on mobile, log in
2. Turn off WiFi (but keep cellular on, or vice versa)
3. Verify the app may briefly show offline then recover via the other radio
4. Turn off ALL radios (Airplane Mode)
5. Verify the app goes offline
6. Turn Airplane Mode off
7. Verify the app recovers: offline indicator clears, data syncs, sequential queue flushes

**7. Transient failures do NOT trigger offline**
1. Open the app on web, log in
2. Using DevTools, block a specific API endpoint briefly (e.g., block for 2-3 seconds then unblock)
3. Trigger 2-3 quick API calls that fail
4. Verify the app does NOT go offline (failures are under both the 3-count AND 10-second thresholds)
5. Verify `[FailureTracker] Failure #N` logs appear but NO `Sustained failure threshold reached`

### Offline tests

**8. Optimistic updates while offline**
1. Force the app offline (Test Tool Menu → Force offline)
2. Send a chat message in any report
3. Verify the message appears immediately (optimistic update)
4. Create an expense/money request
5. Verify it appears optimistically
6. Turn Force offline OFF
7. Verify queued requests flush and data syncs with the server

**9. Sequential queue behavior**
1. Force the app offline
2. Perform multiple write actions (send 3 messages, create an expense)
3. Verify each action appears optimistically
4. Turn Force offline OFF
5. Verify all queued actions process in order (messages appear server-side in the correct order)

### QA Steps

**10. End-to-end connectivity on staging**
1. Log in on staging on each platform (iOS native, Android native, mWeb Safari, mWeb Chrome, Desktop Chrome/Safari)
2. Navigate around the app (open reports, switch workspaces, view search)
3. Verify no unexpected offline indicators appear
4. Verify no console errors related to network state

**11. Real network interruption on staging**
1. Log in on staging
2. Disconnect network for ~20 seconds
3. Verify offline indicator appears
4. Reconnect network
5. Verify the app recovers and data syncs
6. Verify no duplicate or lost messages

**12. High traffic account regression**
1. Log in with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) on staging
2. Navigate through reports, search, workspaces
3. Verify no long loading states or unexpected offline indicators
4. Verify the app remains responsive and data loads as expected

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/Network/SequentialQueue.ts
```diff
@@ -127,11 +127,6 @@ function process(): Promise<void> {
         return Promise.resolve();
     }
 
-    if (isOffline()) {
-        Log.info('[SequentialQueue] Unable to process. We are offline.');
-        return Promise.resolve();
-    }
```
TMisiukiewicz, line 134: what if `process` is being fired when already offline, but `pause()` was not called yet? Is such scenario possible?

src/libs/FailureTracker.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    Log.info(`[FailureTracker] Failure #${failureCount} (elapsed: ${elapsed}ms, threshold: ${thresholdCount} in ${thresholdWindow}ms)`);
+
+    // Only trigger sustained failure when BOTH count AND time thresholds are met
+    if (failureCount >= thresholdCount && elapsed >= thresholdWindow) {
+        Log.info('[FailureTracker] Sustained failure threshold reached — triggering hard stop');
+        for (const cb of sustainedFailureListeners) {
+            cb(true);
+        }
+    }
```
TMisiukiewicz, line 59: We have this pretty common scenario where people use the app with spotty connection. What happens if someone gets 2 failures in 5 seconds, then a success, then another 3 failures? They would never reach the threshold even the connection is garbage.

src/libs/FailureTracker.ts
```diff
@@ -0,0 +1,86 @@
...
+    firstFailureTimestamp = 0;
+}
+
+/**
+ * Reset the tracker (for testing).
+ */
+function reset() {
+    failureCount = 0;
+    firstFailureTimestamp = 0;
+}
```
TMisiukiewicz: it could be a single function. Additionally, you never use them in the code except tests - shouldn't it reset the counter when we are back online?

contributingGuides/NETWORK_STATE_DETECTION.md
```diff
@@ -0,0 +1,180 @@
...
+- When `isConnected` returns to `true`, it calls `NetworkState.setHasRadio(true)`
+- Detects: airplane mode, WiFi disabled, no cellular signal
+- Does **not** determine actual server reachability — a device can be connected to WiFi but have no internet
+
+### Reachability tracking
+
+The NetInfo listener also tracks `isInternetReachable` transitions. When `isInternetReachable` changes from `false`/`null` to `true`, it calls `NetworkState.onReachabilityRestored()` to clear the hard stop and trigger reconnection.
+
+**Platform behavior:**
+- **Mobile**: `useNativeReachability=true` (default) — NetInfo trusts native `isInternetReachable`, JS polling never starts. Recovery is detected through native state change events.
```
TMisiukiewicz: Docs says `true` but you actually defined it as `false` in the `NetworkConnection`

src/libs/actions/Reconnect.ts
```diff
@@ -0,0 +1,46 @@
...
+        Log.info('[Reconnect] App is still loading, calling openApp');
+        App.openApp();
+    } else {
+        Log.info('[Reconnect] Calling reconnectApp');
+        App.reconnectApp(lastUpdateIDAppliedToClient);
+    }
+
+    // Flush the sequential queue to process any pending write requests
+    flush();
+}
```
TMisiukiewicz, line 55: before calling `OpenApp`/`ReconnectApp` was throttled, is it safe to skip throttling?

adhorodyski:
### Bug analysis: App stuck in offline/skeleton state despite working internet

**Reported in:** [Slack thread](https://expensify.slack.com/archives/C05LX9D6E07/p1773051770866429)

**Symptoms:** App permanently stuck showing skeletons and offline indicator. Ping endpoint returns 200 in DevTools, but the app stays offline. Resolved only by clearing cache and restart. Logs showed "a couple of network state changes which led to being stuck in the offline state."

---

**How the old system (current `main`) gets stuck:**

The old architecture has a catch-22:

1. Network flaps → `isOffline` set to `true` in Onyx via scattered code paths
2. `isOffline: true` → SequentialQueue pauses → no outgoing requests
3. `RecheckConnection` middleware only fires when requests are made — but no requests are going out because the queue is paused
4. The 60s recheck interval only calls `NetInfo.fetch()`, which on web just reads `navigator.onLine` (returns `true`) — doesn't check actual server reachability
5. **Dead end:** app thinks it's offline → doesn't send requests → can't discover it's back online

Additionally, `networkStatus` could get stuck in `UNKNOWN`, and multiple code paths could independently write `isOffline` to Onyx without coordination.

**How this PR prevents it:**

1. NetInfo's reachability polling runs **independently of our request queue** — it doesn't care if SequentialQueue is paused. Polls `api/Ping` every 5s when unreachable.
2. When Ping succeeds → `isInternetReachable` transitions to `true` → `onReachabilityRestored()` forcefully clears ALL hard stop flags, resets FailureTracker counters, unpauses queue, and triggers `reconnect()`. **Recovery within ~5s max.**
3. `isOffline` is now **only** set by `NetworkState.updateState()` — no scattered writers, no inconsistent state.
4. Even stale Onyx state from a previous session is cleared on startup: module-level state initializes clean, and the first NetInfo listener event (fires immediately on subscribe) triggers `updateState()` → `isOffline = false`.

**Verdict:** This PR breaks the catch-22 that caused this bug. The old system relied on request-driven rechecks (dead when the queue is paused). Ours uses NetInfo's independent polling that runs regardless of app state.

src/libs/Network/SequentialQueue.ts
adhorodyski: No race here — `updateState()` calls `pause()` synchronously, which sets `isQueuePaused = true` immediately. `process()` checks `isQueuePaused` (not Onyx `isOffline`) and returns early if paused.

The old `isOffline()` guard was reading from `NetworkStore.isOffline()` which uses the same async Onyx subscription — so it had the same timing characteristics. The synchronous `isQueuePaused` flag is the real guard.

contributingGuides/NETWORK_STATE_DETECTION.md
adhorodyski: Good catch — fixed in 840fe1a.

src/libs/FailureTracker.ts
adhorodyski: Merged into a single `reset()` in 62d5e81.

src/libs/FailureTracker.ts
adhorodyski: yeah this is my main, valid concern as well and what I'm trying to really nail down with this design. Working on it, will update.

adhorodyski: This is intentional — on a spotty connection where some requests succeed, the app IS working (just slowly). Showing the offline indicator would be a lie. The hard stop is for "nothing works at all" — that's when it helps the user.

In the 2-fail → success → 3-fail scenario, SequentialQueue's built-in retry with exponential backoff handles individual write failures. READ failures are retried by components. The 5s NetInfo polling catches actual outages independently.

If we made FailureTracker more aggressive (e.g., sliding window), we'd risk showing offline flicker on borderline connections and unnecessarily pausing the queue.

adhorodyski: I will still just continue testing it, because the threshold might be wrong - I have to agree with the concept though that staying 'online' for the user is key here and reliability comes from the SQ, not the online/offline tracking.

src/libs/actions/Reconnect.ts
adhorodyski: Both `reconnectApp()` and `openApp()` are WRITE commands that use `API.writeWithNoDuplicatesConflictAction`, so the SequentialQueue already deduplicates them:

1. `reconnect()` → `App.reconnectApp()` (`src/libs/actions/Reconnect.ts:38`)
2. `reconnectApp()` → `API.writeWithNoDuplicatesConflictAction(WRITE_COMMANDS.RECONNECT_APP, ...)` (`src/libs/actions/App.ts:498`)
3. `writeWithNoDuplicatesConflictAction()` wraps with `resolveDuplicationConflictAction` as the conflict resolver (`src/libs/API/index.ts:174-184`)
4. `SequentialQueue.push()` calls `checkAndFixConflictingRequest` before persisting (`src/libs/Network/SequentialQueue.ts:489-505`)
5. `resolveDuplicationConflictAction()` scans the persisted queue (`src/libs/actions/RequestConflictUtils.ts:56-72`) — if a matching command exists, it replaces it (`type: 'replace'`), keeping only one

So even if `reconnect()` is called multiple times simultaneously, only one `ReconnectApp` ever sits in the queue. No additional throttling needed.

adhorodyski: This all gets deduplicated by the SQ mechanism so it's completely safe to call it even from multiple places. Only one request will get executed.

src/libs/actions/Reconnect.ts
```diff
@@ -0,0 +1,46 @@
...
+        isLoadingApp = val;
+    },
+});
+
+/**
+ * Centralized reconnection logic.
+ * Called when recovering from a hard stop or when the app comes to foreground.
+ * Syncs app data and flushes the sequential queue.
+ */
+function reconnect() {
```
abzokhattab, line 40: The old `triggerReconnectionCallbacks` was throttled to 5 seconds with `trailing: false`. This new `reconnect()` has no throttling and is called from 3 independent sources: `onReachabilityRestored()`, `initAppForegroundListener()`, and `PusherUtils.onPusherResubscribeToPrivateUserChannel()`. When a user returns to the app after being offline, all 3 can fire within ~100ms, triggering 3 simultaneous `reconnectApp()` / `openApp()` calls 

i think we should bring the throttle back:
```js
const reconnect = throttle(() => {
    Log.info('[Reconnect] Triggering reconnection');
    if (isLoadingApp) {
        openApp();
    } else {
        reconnectApp(lastUpdateIDAppliedToClient);
    }
    flush();
}, 5000, {trailing: false});
```

what do u think?

abzokhattab (changes requested):

src/libs/actions/Reconnect.ts
adhorodyski: please refer to https://github.com/Expensify/App/pull/84760#discussion_r2918272242, this is completely safe now in the new design

abzokhattab: thanks yes that makes sense now

src/libs/NetworkConnection.ts
```diff
@@ -257,105 +124,34 @@ function subscribeToNetInfo(accountID: number | undefined): () => void {
...
-                    });
+                    .then((json: ResponseJSON) => Promise.resolve(json.jsonCode === 200))
+                    .catch(() => Promise.resolve(false));
             },
-
-            // If a check is taking longer than this time we're considered offline
             reachabilityRequestTimeout: CONST.NETWORK.MAX_PENDING_TIME_MS,
+            // Use JS fetch polling (api/Ping) on all platforms instead of native OS reachability.
+            // This aligns behavior across web and mobile: poll every 60s when reachable, 5s when unreachable.
+            useNativeReachability: false,
```
abzokhattab: The default for `useNativeReachability` is true  ...  on mobile, NetInfo uses the OS-level reachability check (instant, no network requests, no battery cost). Setting it to `false` forces JS fetch polling (api/Ping every 60s online, 5s offline) on all platforms, adding battery/data overhead for every mobile user.

The "WiFi connected but no internet" gap that native reachability misses is already covered by Layer 2 (FailureTracker). I'd suggest keeping native reachability on mobile and only disabling it on web:

```js
useNativeReachability: Platform.OS !== 'web'
```

src/libs/Network/SequentialQueue.ts
```diff
@@ -408,14 +407,22 @@ function unpause() {
...
+    // because waitForBatchedUpdates() drains microtasks before resolving,
+    // keeping test timing predictable.
+    //
+    // We pass shouldResetPromise=false to preserve the existing isReadyPromise so that
+    // pending READ requests (waiting via waitForIdle()) resolve correctly after WRITEs complete.
+    Log.info('[SequentialQueue] Deferring flush(false) to allow pending Onyx callbacks to settle', false, {
         numberOfPersistedRequests,
     });
-    flush(false);
+    Promise.resolve().then(() => flush(false));
```
abzokhattab: The microtask deferral `Promise.resolve().then(() => flush(false))` relies on an implicit assumption that pending `Onyx.set` callbacks from `save()`/`update()` during offline `push()` calls are enqueued before this microtask. If they aren't, `flush()` runs before persisted requests are updated in memory, and `process()` reads stale data .. causing duplicate API calls.
 
should we make this explicit instead? something like:

```js
waitForActivePersistedRequestWrites().then(() => {
    Log.info('[SequentialQueue] Persisted request writes settled, flushing');
    flush(false);
});
```

where `PersistedRequests` exposes a promise that resolves when its pending `Onyx.set` calls complete. This way the ordering is guaranteed rather than relying on microtask scheduling.

what do u think

src/libs/NetworkConnection.ts
adhorodyski: The design proposal explicitly states: "**The probe is only a recovery detector**" and "**Periodic lightweight probe with its own backoff**." _The JS polling is that probe_ for the FailureTracker. 

Without it on mobile, there's no reliable mechanism to exit a sustained-failure hard stop when native OS can't distinguish "our API is down" from "internet works fine."

The cost is one tiny GET every 60s when reachable, so negligible. It's close to 0 and here we want to ensure every platform uses the same detection framework. It's intentional to keep this mechanism 1:1 between the platforms.

src/libs/Network/SequentialQueue.ts
adhorodyski: Good catch on the fragility. The microtask deferral was specifically working around [Issue 4 from #80759](https://github.com/Expensify/App/issues/80759#issuecomment-4024407800) — stale `Onyx.set()` callbacks from `save()`/`update()` overwriting in-memory `persistedRequests` during `process()`, causing duplicate API calls. 

The [comment in the code](https://github.com/Expensify/App/pull/84760/changes#diff-c92534c52391b0fdb1593d4bdd47aa724137670a54783d2138ff1f8d9149f3b6R414) explicitly says *"A proper fix to PersistedRequests would make this deferral unnecessary."*

That proper fix just merged: [#84622](https://github.com/Expensify/App/pull/84622) makes in-memory state authoritative after initialization — the Onyx connect callback is now a no-op for non-null values, so stale disk callbacks can no longer overwrite the in-memory queue. 

Since the root cause is eliminated, I'll remove the microtask deferral entirely and go back to calling `flush(false)` directly.

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function updateState() {
+    const offline = isInHardStop();
+    setIsOffline(offline, `hard stop: noRadio=${!hasRadio}, sustainedFailures=${sustainedFailuresActive}, forceOffline=${shouldForceOffline}`);
+
+    if (offline) {
+        pause();
+    } else {
+        unpause();
```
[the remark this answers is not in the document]
adhorodyski: validating

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,117 @@
...
+function setHasRadio(connected: boolean) {
+    const hadRadio = hasRadio;
+    hasRadio = connected;
+
+    if (hadRadio && !hasRadio) {
+        Log.info('[NetworkState] Hard stop: NO_RADIO — OS reports no network interface');
+        updateState();
+    } else if (!hadRadio && hasRadio) {
+        Log.info('[NetworkState] NO_RADIO cleared — OS reports radio is back');
+        updateState();
```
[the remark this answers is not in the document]
adhorodyski: validating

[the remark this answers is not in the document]
adhorodyski: **By design — optimistic instant recovery is intentional.**

When the radio comes back, we clear the hard stop immediately and let requests fly. The reasoning:

1. **Common case is fast recovery.** Airplane mode off → internet works. Making the user wait up to 5s for a Ping reachability poll to confirm what's almost always true is a worse UX trade-off than the rare false-online window.

2. **Dead WiFi / captive portal is covered by Layer 2.** If the radio is back but internet isn't actually reachable, requests fail → FailureTracker hits the sustained failure threshold → hard stop re-activates. The false-online window is bounded to a few failing requests.

3. **Reachability polling still runs independently.** NetInfo polls `api/Ping` every 5s when unreachable. When it confirms connectivity, `onReachabilityRestored()` fires and triggers a full reconnect. So even if FailureTracker didn't exist, recovery would self-correct.

The layered model is: Layer 1 (radio) is optimistic and fast, Layer 2 (FailureTracker) is the safety net. Gating Layer 1 on reachability would collapse the two layers into one and add latency to the happy path.

adhorodyski:
@mjasikowski still working on a sync version of isOffline that is not tied to Onyx, will push some updates soon

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,284 @@
...
+    for (const cb of reconnectListeners) {
+        cb();
+    }
+}
+
+function updateState() {
+    const offline = isOffline();
+
+    if (offline) {
+        lastOfflineAt = new Date().toISOString();
```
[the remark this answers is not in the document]
adhorodyski: false positive, they all use `useNetwork` which got updated.

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
+/**
+ * Called when shouldForceOffline changes in Onyx (debug tool).
+ */
+function setForceOffline(force: boolean) {
+    shouldForceOffline = force;
+    Log.info(`[NetworkState] shouldForceOffline set to ${force}`);
+    updateState();
```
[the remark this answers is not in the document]
adhorodyski: Good catch on the stale-state concern. The underlying issue was that the `NetInfo` listener had an early return when `shouldForceOffline` was active, which meant `hasRadio` and `prevIsInternetReachable` stopped tracking reality while the flag was on.

Fixed by removing the early return. The listener now always updates hasRadio and reachability state regardless of `shouldForceOffline`. Only the `onReachabilityRestored()` reconnect side effect is gated behind `!shouldForceOffline`. This way when force-offline is turned off, the state is already accurate and no `NetInfo.refresh()` call is needed.

src/libs/actions/Reconnect.ts
```diff
@@ -0,0 +1,63 @@
...
+});
+
+// App came to foreground — reconnect to catch up on missed Pusher events
+AppStateMonitor.addBecameActiveListener(() => {
+    Log.info('[Reconnect] App became active');
+    if (isOffline()) {
+        refreshNetworkState();
+    }
+    // Always reconnect on foreground to catch up on missed events
+    reconnect();
```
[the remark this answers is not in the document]
adhorodyski: low harm but true, addressing

src/libs/FailureTracker.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Record a successful request outcome.
+ * Resets the failure tracker — one success proves connectivity.
+ */
+function recordSuccess() {
+    if (failureCount === 0) {
+        return;
+    }
+
+    console.debug(`[FailureTracker] Success after ${failureCount} failures — resetting tracker`);
```
TMisiukiewicz, line 25: I think we should replace all `console.debug`s with `Log.info` for observability reasons, babel removes `console` statements on prod

src/libs/actions/Reconnect.ts
```diff
@@ -0,0 +1,76 @@
...
+
+// App came to foreground — reconnect to catch up on missed Pusher events
+AppStateMonitor.addBecameActiveListener(() => {
+    Log.info('[Reconnect] App became active');
+    if (getIsOffline()) {
+        refreshNetworkState();
+    }
+    // Always reconnect on foreground to catch up on missed events
+    reconnect();
+});
```
TMisiukiewicz, line 81: Does it make sense to introduce some throttling here? If a user is switching between tabs every few seconds, triggering reconnect on each change seems unnecessary

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * Force NetInfo to re-fetch native network state.
+ * Useful when the app comes to foreground while in a hard stop —
+ * bypasses stale isInternetReachable cache (see NetInfo issue #326).
+ */
+function refresh() {
+    NetInfo.refresh();
+}
+
+export {getIsOffline, getLastOfflineAt, subscribe, onReachabilityConfirmed, setHasRadio, setSustainedFailures, setForceOffline, onReachabilityRestored, getDBTimeWithSkew, refresh};
```
TMisiukiewicz: This file seems critical for the entire mechanism - could you cover it with tests?

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,282 @@
...
+function onReachabilityRestored() {
+    Log.info('[NetworkState] Internet reachability restored — clearing hard stops');
+    hasRadio = true;
+    sustainedFailuresActive = false;
+    resetFailureCounters();
+    updateState();
+
+    // Notify reconnect listeners (Reconnect.ts will handle app data sync)
+    notifyReconnectListeners();
+}
```
TMisiukiewicz, line 201: since it's supposed to be called only by NetInfo, can we do anything to protect it from being called outside of it?

src/CONST/index.ts
```diff
@@ -2087,25 +2087,20 @@ const CONST = {
         },
         MIN_RETRY_WAIT_TIME_MS: 10,
         MAX_RANDOM_RETRY_WAIT_TIME_MS: 100,
-        MAX_RETRY_WAIT_TIME_MS: 10 * 1000,
+        MAX_RETRY_WAIT_TIME_MS: 30 * 1000,
```
TMisiukiewicz: what's the reason behind increasing this time by 3x?

contributingGuides/NETWORK_STATE_DETECTION.md
```diff
@@ -0,0 +1,187 @@
...
+## Overview
+
+The app uses a two-layer detection model to determine connectivity status. Each layer feeds into a central **hard stop** state machine that decides whether the app is offline. When a hard stop is active, the app pauses outgoing requests and shows the offline UI. Recovery relies on NetInfo's built-in reachability polling (`isInternetReachable`) and the FailureTracker detecting a successful request. Once connectivity is confirmed, the app clears the hard stop and reconnects.
+
+## Architecture Diagram
+
+```
+┌──────────────────────┐
+│ Layer 1: OS Radio    │   NetInfo isConnected
+│ (NetworkConnection)  │──────────────────────────┐
```
TMisiukiewicz: I can see the entire `NetworkConnection` module is gone, is this diagram correct?

src/libs/Network/SequentialQueue.ts
```diff
@@ -430,12 +439,21 @@ function getShouldFailAllRequests(): boolean {
...
+// This replaces the old NetworkState → unpause() → flush() coupling
+// with a cleaner dependency direction (SequentialQueue → NetworkState).
+let wasOfflineForQueue = isOfflineNetwork();
+subscribeNetworkState(() => {
+    const currentlyOffline = isOfflineNetwork();
+    if (wasOfflineForQueue && !currentlyOffline) {
+        flush();
+    }
+    wasOfflineForQueue = currentlyOffline;
+});
```
TMisiukiewicz: do we now flush twice on network state change? once here and once on `reconnect()`?

src/libs/NetworkState.ts
adhorodyski: just stopped exporting it, thanks. its an internal helper only

src/libs/actions/Reconnect.ts
adhorodyski: responded a few times now in the big thread here, this should be covered (is) by the deduplication mechanism in SQ. No harm in calling it so far.

src/libs/FailureTracker.ts
adhorodyski: This was done on purpose to fix all tests due to a circular dep issue. Let's first make sure this works correctly, this was the smaller, low harm decision I could make.

src/libs/Network/SequentialQueue.ts
adhorodyski: yes, for clarity and to avoid bugs caught during implementation. the 2nd flush is a noop because the queue is already running. these 2 invocations do not _always_ overlap so we have to cover edgecases.

src/CONST/index.ts
adhorodyski: true, reverted

TMisiukiewicz (approved):

OlimpiaZurek (approved):

src/hooks/useNetwork.ts
```diff
@@ -12,16 +9,12 @@ type UseNetwork = {isOffline: boolean; lastOfflineAt?: string};
 
 export default function useNetwork({onReconnect = () => {}}: UseNetworkProps = {}): UseNetwork {
     const callback = useRef(onReconnect);
-    callback.current = onReconnect;
-
-    const [network] = useOnyx(ONYXKEYS.NETWORK, {
-        selector: networkStatusSelector,
+    useEffect(() => {
+        callback.current = onReconnect;
     });
```
mkhutornyi, line 14: No dependency here?

src/hooks/useNetwork.ts
```diff
@@ -35,10 +28,8 @@ export default function useNetwork({onReconnect = () => {}}: UseNetworkProps = {
     }, [isOffline]);
 
     useEffect(() => {
-        // Used to store previous prop values to compare on next render
```
mkhutornyi, line 38: Restore this comment

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * Configure NetInfo with the reachability URL and subscribe to state changes.
+ * Must unsubscribe before calling configure() — configure tears down NetInfo internal state.
+ */
+function configureAndSubscribe() {
+    if (unsubscribeNetInfo) {
+        unsubscribeNetInfo();
+        unsubscribeNetInfo = null;
+    }
```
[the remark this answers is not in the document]
adhorodyski: Do nothing unless this can be proven during manual testing - duplicating reconnect is no harm because it's deduped on the SQ level.

src/setup/index.ts
```diff
@@ -47,7 +47,6 @@ export default function () {
             // Clear any loading and error messages so they do not appear on app startup
             [ONYXKEYS.SESSION]: {loading: false},
             [ONYXKEYS.ACCOUNT]: CONST.DEFAULT_ACCOUNT_DATA,
-            [ONYXKEYS.NETWORK]: CONST.DEFAULT_NETWORK_DATA,
             [ONYXKEYS.IS_SIDEBAR_LOADED]: false,
```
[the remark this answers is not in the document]
adhorodyski: `NetworkState` already handles the `!network` case by resetting `shouldForceOffline` and `simulatePoorConnection` to false.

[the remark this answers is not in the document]
adhorodyski: Bottom line is even if these are stale on sign out (can happen only for debug options) the first login fill merge a new server value in, keeping them up to date in Onyx. Do nothing.

src/hooks/useNetwork.ts
adhorodyski: no, this is to always assign but keeping it compliant with the React Compiler

adhorodyski: maps 1:1 to the previous behavior

src/hooks/useNetwork.ts
adhorodyski: done, thanks

mkhutornyi:
Still conflicts though

mkhutornyi:
Bug: offline indicator shows some time after enabling "Simulate failing network requests"


https://github.com/user-attachments/assets/3000348d-98e4-4aae-b6a5-687ab98653eb

adhorodyski:
> Bug: ReconnectApp api is not called after online back

This one is a true regression, coming from a known case I forgot to test eventually and we partially documented. Adding it now with unit tests to always assert reachability checks result in calls to Reconnect 👍🏻

mountiny:
> It's a UX tradeoff where we avoid false positives from transient failures at the cost of a brief delay before showing the offline state.


I think that is fine and better ux

mkhutornyi:
> > Bug: ReconnectApp api is not called after online back
> 
> This one is a true regression, coming from a known case I forgot to test eventually and we partially documented. Adding it now with unit tests to always assert reachability checks result in calls to Reconnect 👍🏻

Confirmed this working on real network reconnect.
But still not working on Force offline simulation

Screenshot 2026-04-10 at 8 03 49 am

So this causes regression:
(Message sent by other user is not fetched after online)

https://github.com/user-attachments/assets/6ff744d4-3c05-44fa-819d-a00e2082d0b6

mkhutornyi:
@MelvinBot can you list all PRs as much as possible where removed code in this PR was originally implemented?

Example:
https://github.com/Expensify/App/blob/7a0ee35a51409162e44cae99afdbf00a5c229624/src/libs/actions/Session/AttachmentImageReauthenticator.ts#L16-L27
https://github.com/Expensify/App/blob/7a0ee35a51409162e44cae99afdbf00a5c229624/src/Expensify.tsx#L214-L217
https://github.com/Expensify/App/blob/7a0ee35a51409162e44cae99afdbf00a5c229624/src/setup/backgroundLocationTrackingTask/index.native.ts#L18-L22

adhorodyski:
>>> Bug: ReconnectApp api is not called after online back

>> This one is a true regression, coming from a known case I forgot to test eventually and we partially documented. Adding it now with unit tests to always assert reachability checks result in calls to Reconnect 👍🏻

> Confirmed this working on real network reconnect. But still not working on Force offline simulation.


https://github.com/user-attachments/assets/3c05f4fb-f694-4884-a465-d4d59ba84eb0

repros correctly now, thanks. covered with new unit tests as well

adhorodyski:
@mkhutornyi let me know how this works for you now

cspell.json
```diff
@@ -405,6 +405,7 @@
         "lucene",
         "Luhn",
         "Maat",
+        "macrotask",
```
mkhutornyi: What's this for? No occurencies found

mkhutornyi:
Please generate new adhoc build

mkhutornyi:
While https://github.com/Expensify/App/pull/84760#issuecomment-4206843516 is not blocker, the core feature of `Simulate failing network requests` is not working at all now.


https://github.com/user-attachments/assets/1d4fa713-1f54-4b84-ba62-c171b466e3b1

mkhutornyi:
While https://github.com/Expensify/App/pull/84760#issuecomment-4206843516 is not blocker, there are still issues.

Please check this video around 00:40. Message is duplicated when disabling `Simulate failing network requests`


https://github.com/user-attachments/assets/fcb48bfc-1061-4546-84b2-cca18f94bb60

----

After enabling `Simulate failing network requests` and became "offline", it gets stuck on offline status.
Also, even after disabling `Simulate failing network requests`, still offline forever before calling any api.


https://github.com/user-attachments/assets/f8df03c3-0848-434c-a08c-e5fbe8cff948

mkhutornyi:
When network is on (local network working) but internet not reachable:

There's no offline indicator even after waiting long time

https://github.com/user-attachments/assets/4262c852-7f42-4ed0-be95-ccb4de9c0a64



production:


https://github.com/user-attachments/assets/dc705f54-853f-4343-93ab-87372410b68b

adhorodyski:
> Please check this video around 00:40. Message is duplicated when disabling Simulate failing network requests

I think this is a fundamental issue with this debug option: `shouldFailAllRequests` throws _after_ the server responds successfully, creating a split where the server has the data but the client rolls it back. On recovery, the server re-delivers it. @mjasikowski if we want to support this, we should rework it to actually fail the requests and not send them while marking them as 'failed'.

```ts
.then((response) => {
            // Test mode where all requests will succeed in the server, but fail to return a response
            if (shouldFailAllRequests || shouldForceOffline) {
                throw new HttpsError({
                    message: CONST.ERROR.FAILED_TO_FETCH,
                });
            }
```

adhorodyski:
> After enabling Simulate failing network requests and became "offline", it gets stuck on offline status.
Also, even after disabling Simulate failing network requests, still offline forever before calling any api.

Oh makes sense because NetInfo's `/api/ping` calls succeed regardless of this option and we're hitting a deadlock. Let me push a fix for this.

adhorodyski:
Ok I was able to dive into a big aspect that looks like I only glimpsed over before, which is what `useNativeReachability: false` allows us to do - now finally fully rely on `isInternetReachable` as well. I'm pushing an update now that makes this another, 5th signal that helps distinguish true online/offline changes:

With `useNativeReachability: false`, all platforms fall through to `_setExpectsConnection(state.isConnected)`, which starts JS `fetch(api/Ping)` polling.

### Result: consistent behavior across all platforms

| Platform | `isConnected` (hasRadio) | `isInternetReachable` (internetUnreachable) |
|----------|-------------------------|---------------------------------------------|
| Web | `navigator.onLine` | `fetch(api/Ping)` |
| iOS | `SCNetworkReachability` | `fetch(api/Ping)` |
| Android | `NetworkCapabilities` | `fetch(api/Ping)` |

`isInternetReachable` is determined by `fetch(api/Ping)` on every platform. That's the whole point of `useNativeReachability: false` — cross-platform consistency via a real HTTP request to your own backend. The `internetUnreachable` flag will work identically on web, iOS, and Android.

adhorodyski:
@mkhutornyi can you please re-test this one now?

> When network is on (local network working) but internet not reachable, there's no offline indicator even after waiting long time

mkhutornyi:
> When network is on (local network working) but internet not reachable

This isn't easily testable. I'll retest once this happens again for me.

Are all other comments addressed?

adhorodyski:
@mkhutornyi yes, all addressed

mountiny:
Conflicts @adhorodyski

adhorodyski:
Resolved all conflicts

mkhutornyi:
> > When network is on (local network working) but internet not reachable
> 
> This isn't easily testable. I'll retest once this happens again for me.

Retested but still offline indicator not showing

src/libs/NetworkState.ts
```diff
@@ -0,0 +1,397 @@
...
+        // Always track real radio/reachability state so values are never stale
+        // when shouldForceOffline is turned off. Only gate the reconnect side effect.
+        setHasRadio(radio);
+
+        // When isInternetReachable transitions to false, api/Ping failed — go offline.
+        // Guard against undefined (initial event) and null (indeterminate) — only
+        // react to a confirmed true→false transition.
+        if (state.isInternetReachable === false && prevIsInternetReachable === true) {
+            setInternetUnreachable(true);
+        }
```
mkhutornyi: The root cause is that `prevIsInternetReachable = null` so this condition isn't met.

This works for me:
```
        if (state.isInternetReachable === false && prevIsInternetReachable !== false) {
```

adhorodyski: ok I wasn't sure about this one before but if definitely can be a case with debug options, pushed together with tests.

src/setup/index.ts
```diff
@@ -47,7 +47,6 @@ export default function () {
             // Clear any loading and error messages so they do not appear on app startup
             [ONYXKEYS.SESSION]: {loading: false},
             [ONYXKEYS.ACCOUNT]: CONST.DEFAULT_ACCOUNT_DATA,
-            [ONYXKEYS.NETWORK]: CONST.DEFAULT_NETWORK_DATA,
             [ONYXKEYS.RAM_ONLY_IS_SIDEBAR_LOADED]: false,
```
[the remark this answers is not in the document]
adhorodyski: Invalid, `DEFAULT_NETWORK_DATA` was intentionally removed as part of the new architecture

mkhutornyi:
Is this ready for final review?

adhorodyski:
I'm just looking for the latest comments - much of them are already discussed but want to double check everything reported by codex and melvin. I will update here once done.

adhorodyski:
@mkhutornyi all ready now

mkhutornyi:
Lint failing

mkhutornyi:
@mjasikowski can you please generate new build?

mkhutornyi:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

Tested on real device

</details>

<details>

Tested on real device

</details>

<details>



https://github.com/user-attachments/assets/e7d3ad3b-31ff-46c5-a5d7-91dff8d80edb


</details>

<details>



https://github.com/user-attachments/assets/1d0e6e04-94a5-4fab-98f7-da8e70654ef1

Also tested on real device

</details>

<details>


https://github.com/user-attachments/assets/f5a4266b-e897-4bda-be26-0b5daa572b75



</details>

mjasikowski (approved):

mkhutornyi (approved):

mkhutornyi:
No more blockers found other than the reported ones already marked as nab.
Let's give this a go.

mjasikowski:
Merging, hold tight

mountiny:
😍

src/libs/actions/Reconnect.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Internet confirmed reachable — reconnect
+onNetworkReachabilityConfirmed(() => {
+    reconnect();
+});
+
+// Any offline→online transition — flush the sequential queue
+let wasOffline = getIsOffline();
+subscribeNetworkState(() => {
```
thelullabyy, line 64: They don't get loaded until the app makes its first actual API request (e.g., a user navigates to a screen that triggers a call). After we restart the App, this file is not loaded yet, which causes the listener is not registered --> Caused this issue https://github.com/Expensify/App/issues/89056

Issue #83356, opened 2026-02-24
[Due for payment 2026-04-14] [Odometer] DRY shared camera logic between IOURequestStepScan and Odometer image capture


Julesssss:
Parent issue: https://github.com/Expensify/App/issues/71841

Follow up improvement issue from https://github.com/Expensify/App/pull/82432
___________

The Odometer image capture flow (`IOURequestStepDistanceOdometerScan`) duplicates significant logic from `src/pages/iou/request/step/IOURequestStepScan/index.native.tsx`. Extract shared functions/hooks/components to be reused in both places.

**Improvements include:** (some were already applied to PR updates)
- Remove useEffect watching focusPoint - call focusCamera directly from tap
  gesture handler
- Missing return after showCameraAlert() when camera.current
- `tapGesture` / `capturePhoto` duplication
- Child components
- CONSISTENCY-6: Missing .catch() on cropImageToAspectRatio().then()
- CONSISTENCY-5: eslint-disable at line 408 lacks justification comment
- P1 (Codex): Hardcoded shouldUseTransactionDraft(CONST.IOU.ACTION.CREATE,
  CONST.IOU.TYPE.REQUEST)
- [Magic number style values](https://github.com/Expensify/App/pull/82432#discussion_r2844560329) should use named constants

<details><summary>Issue Owner</summary>Current Issue Owner: @Eskalifer1</details>

jakubkalinski0:
This topic should be taken into account and discussed when resolving this issue https://github.com/Expensify/App/pull/83236/changes/a471d44d63f70521f6b25c399bbcb1af2e24f469#r2857231314

<details>
<summary>Full context</summary>

- issue:
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/535c0001-ea5a-40e0-9fa5-74920db07866" />

- change and discussion:

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/bca64cbf-6997-443e-b0e6-94848473e768" />

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/11e4dfa4-fa53-4c04-8b24-9a325ddd64a0" />

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/81a425d0-22e8-42fc-9e6c-babd5aefc9de" />

</details>

Eskalifer1:
# BugZero Checklist:

[...]

PR #85758, opened 2026-03-19
Dry camera logic

DylanDylann:
### Explanation of Change

Extracts duplicated camera logic into two shared hooks — `useNativeCamera` and `useWebCamera` — replacing repeated implementations across IOURequestStepScan, IOURequestStepOdometerImage, and MobileWebCameraView. This reduces code duplication and makes camera behavior easier to maintain in one place.

### Fixed Issues

$ https://github.com/Expensify/App/issues/83356
PROPOSAL:

### Tests

### QA Steps

1. Go to Scan request
2. Try to make a scan request by using the camera in the app
3. Verify that the flow is working well 
4. Verify image attachment is working well for chats

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/1a149ade-43cd-4bd8-8a7a-9a3048ff1e27

</details>

<details>

https://github.com/user-attachments/assets/8164934a-4402-4196-88d2-480258e4eece

</details>

<details>

https://github.com/user-attachments/assets/a8a414fd-b0b2-4394-acfc-9dde5fd8488a

</details>

<details>

https://github.com/user-attachments/assets/bb564d70-52db-4b84-807b-ca39d6a48ec9

</details>

<details>

https://github.com/user-attachments/assets/6527dcb3-ba9c-4cf1-8003-a337d0d58c71

</details>


src/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    useEffect(() => {
+        if (!isTabActive) {
+            return;
+        }
+        queryCameraPermission()
+            .then((state) => {
+                setCameraPermissionState(state);
+                if (state === 'granted') {
+                    requestCameraPermission();
```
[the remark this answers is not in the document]
DylanDylann: This is existing behavior; I only refactored it into a hook

src/pages/iou/request/step/IOURequestStepScan/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,207 @@
...
+                            lastBackDeviceId = device.deviceId;
+                            break;
+                        }
+                    }
+                    if (!lastBackDeviceId) {
+                        setDeviceConstraints(defaultConstraints);
+                        return;
+                    }
+                    setDeviceConstraints({deviceId: lastBackDeviceId});
+                });
```
Eskalifer1: I think a `catch` block is missing here, which could result in outdated data, such as an infinite Loader
```ts
 .catch(() => {
   setVideoConstraints(defaultConstraints);
});
```

src/pages/iou/request/step/IOURequestStepOdometerImage/index.tsx
```diff
@@ -98,111 +102,16 @@ function IOURequestStepOdometerImage({
...
-
-        if ('torch' in capabilities && capabilities.torch) {
-            trackRef.current = track;
-        }
-        setIsTorchAvailable('torch' in capabilities && !!capabilities.torch);
-    };
-
-    const viewfinderLayout = useRef<LayoutRectangle>(null);
+        setIsInitialPermissionQueried(true);
+    }, [isInitialPermissionQueried]);
```
Eskalifer1: We have eslint warning here:
`Error: Calling setState synchronously within an effect can trigger cascading renders`

Let's handle it!

src/pages/iou/request/step/IOURequestStepOdometerImage/index.tsx
```diff
@@ -98,111 +102,16 @@ function IOURequestStepOdometerImage({
...
     });
 
-    /**
-     * On phones that have ultra-wide lens, react-webcam uses ultra-wide by default.
-     * The last deviceId is of regular len camera.
-     */
-    const requestCameraPermission = useCallback(() => {
-        if (!isMobile()) {
-            return;
-        }
```
Eskalifer1, line 113: We used to have a check for `isMobile` in OdometerImage to prevent this function from being called where there is clearly no camera (PC), but now that logic is missing. However, there was never any such logic in `MobileWebCameraView.tsx` to begin with.

Now it turns out that this logic is missing in both places.

It might be better to add an `enabled` prop to control when we want to execute the logic, and if this flag is set to `false`, simply skip the entire logic (camera request, useEffect, etc.). 

What do you think about this?

src/pages/iou/request/step/IOURequestStepOdometerImage/index.native.tsx
```diff
@@ -20,49 +17,29 @@ import Text from '@components/Text';
...
 import getReceiptsUploadFolderPath from '@libs/getReceiptsUploadFolderPath';
 import {shouldUseTransactionDraft} from '@libs/IOUUtils';
 import Log from '@libs/Log';
 import Navigation from '@libs/Navigation/Navigation';
-import type {SkeletonSpanReasonAttributes} from '@libs/telemetry/useSkeletonSpan';
-import CameraPermission from '@pages/iou/request/step/IOURequestStepScan/CameraPermission';
 import NavigationAwareCamera from '@pages/iou/request/step/IOURequestStepScan/components/NavigationAwareCamera/Camera';
 import {cropImageToAspectRatio} from '@pages/iou/request/step/IOURequestStepScan/cropImageToAspectRatio';
 import type {ImageObject} from '@pages/iou/request/step/IOURequestStepScan/cropImageToAspectRatio';
+import useNativeCamera from '@pages/iou/request/step/IOURequestStepScan/hooks/useNativeCamera';
```
Eskalifer1: Now that we have reusable hooks that are used identically in two places, it doesn't seem quite right to tie ourselves to a single `IOURequestStepScan` module and import it into another module, `IOURequestStepOdometerImage`

src/pages/iou/request/step/IOURequestStepOdometerImage/index.tsx
```diff
@@ -19,15 +16,15 @@ import useLocalize from '@hooks/useLocalize';
...
+import {isMobile} from '@libs/Browser';
 import {base64ToFile} from '@libs/fileDownload/FileUtils';
 import {shouldUseTransactionDraft} from '@libs/IOUUtils';
 import Log from '@libs/Log';
 import Navigation from '@libs/Navigation/Navigation';
-import type {SkeletonSpanReasonAttributes} from '@libs/telemetry/useSkeletonSpan';
 import NavigationAwareCamera from '@pages/iou/request/step/IOURequestStepScan/components/NavigationAwareCamera/WebCamera';
 import {cropImageToAspectRatio} from '@pages/iou/request/step/IOURequestStepScan/cropImageToAspectRatio';
 import type {ImageObject} from '@pages/iou/request/step/IOURequestStepScan/cropImageToAspectRatio';
+import useWebCamera from '@pages/iou/request/step/IOURequestStepScan/hooks/useWebCamera';
```
Eskalifer1: Same

src/pages/iou/request/step/IOURequestStepScan/components/MobileWebCameraView.tsx
```diff
@@ -17,15 +14,16 @@ import {useMemoizedLazyExpensifyIcons, useMemoizedLazyIllustrations} from '@hook
...
 import useTheme from '@hooks/useTheme';
 import useThemeStyles from '@hooks/useThemeStyles';
-import {isMobileWebKit} from '@libs/Browser';
 import {base64ToFile} from '@libs/fileDownload/FileUtils';
 import {cancelSpan, endSpan, getSpan, startSpan} from '@libs/telemetry/activeSpans';
 import type {SkeletonSpanReasonAttributes} from '@libs/telemetry/useSkeletonSpan';
 import {cropImageToAspectRatio} from '@pages/iou/request/step/IOURequestStepScan/cropImageToAspectRatio';
 import type {ImageObject} from '@pages/iou/request/step/IOURequestStepScan/cropImageToAspectRatio';
 import useMobileReceiptScan from '@pages/iou/request/step/IOURequestStepScan/hooks/useMobileReceiptScan';
+import useWebCamera from '@pages/iou/request/step/IOURequestStepScan/hooks/useWebCamera';
```
Eskalifer1: Same

src/pages/iou/request/step/IOURequestStepScan/index.native.tsx
```diff
@@ -26,31 +23,27 @@ import usePolicy from '@hooks/usePolicy';
...
 import ONYXKEYS from '@src/ONYXKEYS';
 import type {Route} from '@src/ROUTES';
 import ROUTES from '@src/ROUTES';
 import type {FileObject} from '@src/types/utils/Attachment';
-import {getEmptyObject} from '@src/types/utils/EmptyObject';
-import CameraPermission from './CameraPermission';
 import NavigationAwareCamera from './components/NavigationAwareCamera/Camera';
 import ReceiptPreviews from './components/ReceiptPreviews';
 import useMobileReceiptScan from './hooks/useMobileReceiptScan';
+import useNativeCamera from './hooks/useNativeCamera';
```
Eskalifer1: Same

src/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,207 @@
...
+        toggleFlashlight,
+        isTorchAvailable,
+        isQueriedPermissionState,
+        videoConstraints,
+        requestCameraPermission,
+        setupCameraPermissionsAndCapabilities,
+        clearTorchConstraints,
+        capturePhotoWithFlash,
+    };
+}
```
Eskalifer1, line 223: The hook returns `trackRef` and `clearTorchConstraints`, which appear to be more of an internal implementation of the hook and are not used externally. I think we can stop passing these values; if they are needed in the future, we can always add them back.

src/pages/iou/request/step/IOURequestStepOdometerImage/index.tsx
```diff
@@ -98,111 +102,16 @@ function IOURequestStepOdometerImage({
...
-                setCameraPermissionState('denied');
-            });
-    }, []);
-
+    // The odometer's own initial permission query for non-mobile (useWebCamera handles mobile)
     useEffect(() => {
-        if (!isMobile() || !isTabActive) {
+        if (!isMobile() || isInitialPermissionQueried) {
             return;
         }
```
Eskalifer1: Could you please explain why this logic is needed? It seems like some leftover code. We have a state called `isInitialPermissionQueried`, which is updated from `false` to `true` during the first render, depending on `isMobile`. However, this state itself isn't used anywhere.

src/hooks/useNativeCamera.ts
```diff
@@ -0,0 +1,167 @@
...
+    onFocusStart?: () => void;
+
+    /** Additional cleanup to run when focus effect cleans up */
+    onFocusCleanup?: () => void;
+};
+
+function useNativeCamera({context, onFocusStart, onFocusCleanup}: UseNativeCameraOptions) {
+    const {translate} = useLocalize();
+    const {isLoaderVisible} = useFullScreenLoaderState();
+    const {setIsLoaderVisible} = useFullScreenLoaderActions();
```
Eskalifer1, line 36: I’m not a big fan of mixing logic, since the `useNativeCamera` hook should only handle logic related to the camera. Right now, this hook also handles the logic for disabling the Loader indicator, which leads to a mix of logic. In other words, if we want to use this hook elsewhere in the future—where, for example, we don’t want to be tied to the Loader—this part of the code will simply be redundant, and in the worst case, it might even cause some regressions.

Maybe we should move the cleanup logic to `onFocusCleanup`—what do you think?

src/pages/iou/request/step/IOURequestStepScan/hooks/useNativeCamera.ts
```diff
@@ -0,0 +1,167 @@
...
+        didCapturePhoto,
+        setDidCapturePhoto,
+        isAttachmentPickerActive,
+        setIsAttachmentPickerActive,
+        isPlatformMuted,
+        askForPermissions,
+        tapGesture,
+        cameraFocusIndicatorAnimatedStyle,
+        cameraLoadingReasonAttributes,
+        setIsLoaderVisible,
```
Eskalifer1: Given the previous point, I don't think it's this hook's responsibility to return that value; it can be retrieved from another context.

src/pages/iou/request/step/IOURequestStepScan/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,207 @@
...
+                    requestCameraPermission();
+                }
+            })
+            .catch(() => {
+                setCameraPermissionState('denied');
+            })
+            .finally(() => {
+                setIsQueriedPermissionState(true);
+            });
+        // Refresh permission state whenever this tab regains focus.
```
Eskalifer1: It looks like some code is missing here
```ts
        return () => {
            setVideoConstraints(undefined);
        };
```

src/pages/iou/request/step/IOURequestStepScan/hooks/useNativeCamera.ts
Eskalifer1: MB, i'm talking only about `setIsLoaderVisible `

src/pages/iou/request/step/IOURequestStepOdometerImage/index.tsx
DylanDylann: removed

src/hooks/useNativeCamera.ts
DylanDylann: I’d prefer to keep the current implementation to avoid code duplication

Eskalifer1: Okay, let's keep it for now

Eskalifer1:
> @Eskalifer1 Thanks for investigation. I saw this difference when implementing, but I think it’s safe. Why would this cleanup cause a loader issue on your side?

Simply put, if the user has granted camera access but `videoConstraints` is empty, an `ActivityIndicator` is displayed. 

https://github.com/Expensify/App/blob/d222a21795f40431eee3168fdcf4bce70549da16/src/pages/iou/request/step/IOURequestStepScan/components/MobileWebCameraView.tsx#L365-L378 

With the new approach, we always set this value to `undefined` before switching to another tab, which means that when we return later, this value is initially `undefined`, causing the `ActivityIndicator` to appear.

Is this acceptable behavior for us, or should we fix it? What do you think?

src/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,213 @@
...
+            })
+            .catch(() => {
+                setCameraPermissionState('denied');
+            })
+            .finally(() => {
+                setIsQueriedPermissionState(true);
+            });
+        return () => {
+            setDeviceConstraints(undefined);
+        };
```
Eskalifer1, line 154: It seems that implementing this does indeed cause regressions. The camera takes longer to turn on when switching between tabs.

I think we need to fix this. One idea could be to dynamically pass a single parameter that determines whether cleanup (Odometer flow) occurs or not (Scan flow).

Eskalifer1: To see the difference in time, you can watch the comparison in the video that I recorded in the comment above, where the first point is the changes in this PR, and the second point is the behavior on the prod.

And on the prod, the camera turns on much faster than locally, precisely because we added this cleanup

Eskalifer1 (approved):

src/pages/iou/request/step/IOURequestStepScan/index.native.tsx
```diff
@@ -72,12 +66,40 @@ function IOURequestStepScan({
...
     const StyleUtils = useStyleUtils();
     const {translate} = useLocalize();
-    const {isLoaderVisible} = useFullScreenLoaderState();
+
+    // Ref for double-tap protection (doesn't trigger re-render)
+    const isCapturingPhoto = useRef(false);
+
+    const onFocusStart = useCallback(() => {
+        isCapturingPhoto.current = false;
+    }, []);
```
jakubkalinski0: **Nitpick:** React Compiler handles this memoization automatically, so we should remove the manual useCallback

src/pages/iou/request/step/IOURequestStepScan/index.native.tsx
```diff
@@ -72,12 +66,40 @@ function IOURequestStepScan({
...
+
+    const onFocusCleanup = useCallback(() => {
+        cancelSpan(CONST.TELEMETRY.SPAN_RECEIPT_CAPTURE);
+        cancelSpan(CONST.TELEMETRY.SPAN_SHUTTER_TO_CONFIRMATION);
+    }, []);
```
jakubkalinski0: **Nitpick:** Same as above. This is handled by React Compiler so we should remove the useCallback

src/pages/iou/request/step/IOURequestStepScan/index.native.tsx
```diff
@@ -94,17 +116,8 @@ function IOURequestStepScan({
     const navigateBack = useCallback(() => {
         Navigation.goBack(backTo);
     }, [backTo]);
```
jakubkalinski0: **Nitpick:** I know you didn't touch that but since we are already introducing changes here this useCallback also can be removed for the same reason as above.

src/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,211 @@
...
+                    requestCameraPermission();
+                }
+            })
+            .catch(() => {
+                setCameraPermissionState('denied');
+            })
+            .finally(() => {
+                setIsQueriedPermissionState(true);
+            });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
jakubkalinski0: **Nitpick:** we should add an explanation here for example:
`// eslint-disable-next-line react-hooks/exhaustive-deps -- requestCameraPermission is defined inline so including it would cause an infinite loop`

src/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,211 @@
...
+    }, [isTabActive]);
```
jakubkalinski0, line 156: **Nitpick:**
No cleanup is returned here, so if `isTabActive` toggles quickly we can end up with two promises in-flight at once - the older one might settle last and overwrite the newer result with stale state.

**Realistically the risk is pretty low:** `queryCameraPermission()` almost always resolves from the module-level cache (synchronously via microtask) and `useIsFocused()` shouldn't toggle that fast in practice. But the fix is cheap and in compliance with the known [pattern](https://react.dev/learn/synchronizing-with-effects#fetching-data) so worth adding imo:
```Suggestion
useEffect(() => {
        if (!isTabActive) {
            return;
        }
        let ignore = false;
        queryCameraPermission()
            .then((state) => {
                if (!ignore) {
                    setCameraPermissionState(state);
                    if (state === 'granted') {
                        requestCameraPermission();
                    }
                }
            })
            .catch(() => { if (!ignore) setCameraPermissionState('denied'); })
            .finally(() => { if (!ignore) setIsQueriedPermissionState(true); });
        return () => { ignore = true; };
        // eslint-disable-next-line react-hooks/exhaustive-deps -- requestCameraPermission is defined inline so including it would cause an infinite loop
    }, [isTabActive]);
```

src/hooks/useWebCamera.ts
```diff
@@ -0,0 +1,211 @@
...
+     */
+    const requestCameraPermission = () => {
+        if (!isMobile()) {
+            return;
+        }
+        const defaultConstraints = {facingMode: {exact: 'environment'}};
+        navigator.mediaDevices
+            .getUserMedia({video: {facingMode: {exact: 'environment'}, zoom: {ideal: 1}}})
+            .then((stream) => {
+                setCameraPermissionState('granted');
```
jakubkalinski0, line 74: **Nitpick:** When `queryCameraPermission` resolves with `'granted'`, `setCameraPermissionState('granted')` is called twice - once at [L130](https://github.com/Expensify/App/pull/85758/changes#diff-e04294f872ce84e9894b41d6154cba4270287b44d43f3705875211e816e84662R130) before `requestCameraPermission()` is invoked, and again inside its `.then() at L74 (this line). React bails out on same-value state updates so there's no extra render, but L74's call is actually the meaningful one (it confirms the real getUserMedia grant, and also handles the 'prompt' -> 'granted' transition). Imo L130's call is the redundant one in this specific branch

src/hooks/useNativeCamera.ts
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            camera.current.focus(point).catch((error: Record<string, unknown>) => {
+                if (error.message === '[unknown/unknown] Cancelled by another startFocusAndMetering()') {
+                    return;
+                }
+                Log.warn('Error focusing camera', error);
+            });
+        },
+        [camera],
```
jakubkalinski0: `useRef` always returns the same object, and `const camera = useRef<Camera>(null);` so I think we should should just have an empty array here

jakubkalinski0 (approved):
Changes looking great! Testing also went smoothly! I left a few nitpick comments here and there but overall great job! Thanks for handling that 🚀

DylanDylann:
@Julesssss Could you take a look when you have a chance?

Julesssss (approved):
Testing well for me.

Issue #66381, opened 2025-07-16
Remove `Onyx.connect()` for the key: `ONYXKEYS.COLLECTION.REPORT_ACTIONS` in `src/libs/OptionsListUtils.ts`


tgolen:
This is part of deprecating Onyx.connect. [Parent Issue](https://github.com/Expensify/Expensify/issues/522215)

## Details
- **Module**: [src/libs/OptionsListUtils.ts](https://github.com/Expensify/App/blob/main/src/libs/OptionsListUtils.ts)
- **Onyx Key**: `ONYXKEYS.COLLECTION.REPORT_ACTIONS`
- **Reference**: `Line 386: ONYXKEYS.COLLECTION.REPORT_ACTIONS`

Be sure to refactor methods using test driven development:
1. Create a unit test for methods before refactoring
2. Refactor the methods
3. Ensure the unit test still runs after refactoring
4. Reinforce the unit testing with a functional test and a QA test
5. Once you are done, decrease the number of allowable ESLint warnings [here](https://github.com/Expensify/App/blob/e7b5ac9401e2ae9b5c2c70ec513e8de6f5279d7d/package.json#L48) to equal the number of `Onyx.connect` references remaining in the code

<details><summary>Issue Owner</summary>Current Issue Owner: @DylanDylann</details>

danieldoglas:
same

jliexpensify:
Not overdue

DylanDylann:
Please assign me as the reviewer too

truph01:
@DylanDylann 3rd [PR](https://github.com/Expensify/App/pull/86643) is ready

truph01:
@DylanDylann 3rd [PR](https://github.com/Expensify/App/pull/86643) is ready

grgia:
thank you @truph01, I see review is in progress

jliexpensify:
Hi, did anyone need a new payment summary?

mallenexpensify:
@jliexpensify I'll snag this from ya
Please comment and tag me if/when payment is due, thx

jliexpensify:
Np, cheers Matt!

DylanDylann:
Not overdue

tgolen:
Not overdue, dropping to weekly.

tgolen:
This is being worked on

DylanDylann:
Not overdue. Waiting for next PRs

truph01:
[...]

PR #85783, opened 2026-03-19
Remove Onyx.connect() key ONYXKEYS.COLLECTION.REPORT_ACTIONS in src/libs/OptionsListUtils.ts - part 1

truph01:
### Explanation of Change

This PR begins the migration of `Onyx.connect()` for `ONYXKEYS.COLLECTION.REPORT_ACTIONS` in `OptionsListUtils` by introducing an OnyxDerived value and refactoring the `prepareReportOptionsForDisplay → getValidOptions` flow.
**What changed:**
1. **New OnyxDerived config (`sortedReportActions.ts`)**: Extracts the sorting/caching logic from the `Onyx.connect` callback into a derived value under `ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS`. This computes `sortedActions`, `lastActions`, and `transactionThreadIDs` from `REPORT_ACTIONS` and `REPORT` collections — the same derivation the callback performs today.
2. **Refactored** `prepareReportOptionsForDisplay → getValidOptions` flow.
3. **Deprecated module-level variables**: Renamed `allSortedReportActions`, `lastReportActions`, `cachedOneTransactionThreadReportIDs`, and `allReportActions` with a `deprecated` prefix and added `@deprecated` JSDoc tags to signal other developers to use the derived value instead.

### Fixed Issues

$ https://github.com/Expensify/App/issues/66381
PROPOSAL:

#### Test 1: 

1. Navigate to New Chat (tap the FAB → "Start Chat" or equivalent)
2. Verify the contact list and recent chats load correctly
3. Search for a contact by name or email
4. Verify results display and you can select a contact to start a chat

#### Test 2: 
1. Log in to the app with an account that has multiple chats (including DMs, group chats, and workspace rooms)
2. Tap the Search icon / open the search page
3. Verify the search options list loads and displays recent reports correctly
4. Type a partial name of a known contact or workspace in the search field
5. Verify filtered results appear correctly and match expected contacts/reports

### QA Steps

// TODO: These must be filled out, or the issue title must include ""[No QA].""

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] I verified that any new or modified comments were clear, correct English, and explained ""why"" the code was doing something instead of only explaining ""what"" the code was doing.
    - [x] I verified proper file naming conventions were followed for any new files or renamed files. All non-platform specific files are named after what they export and are not named ""index.js"". All platform-specific files are named for the platform the code supports as outlined in the README.

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/d1def9f8-690c-414f-9701-a66ac6c4d0ff

</details>


src/libs/actions/OnyxDerived/configs/sortedReportActions.ts
```diff
@@ -0,0 +1,53 @@
...
+                sortedReportActions = getCombinedReportActions(sortedReportActions, transactionThreadReportID, transactionThreadReportActionsArray, reportID);
+            }
+
+            const firstReportAction = sortedReportActions.at(0);
+            if (firstReportAction) {
+                lastActions[reportID] = firstReportAction;
+            }
+        }
+
+        return {sortedActions, lastActions, transactionThreadIDs};
```
tgolen, line 104: I couldn't find `lastActions` or `transactionThreadIDs` referenced anywhere, so I assume that those references will be added in future PRs? If so, that's fine for now.

src/hooks/useSearchSelector.base.ts
```diff
@@ -173,6 +173,7 @@ function useSearchSelectorBase({
     const [draftComments] = useOnyx(ONYXKEYS.COLLECTION.REPORT_DRAFT_COMMENT);
     const [nvpDismissedProductTraining] = useOnyx(ONYXKEYS.NVP_DISMISSED_PRODUCT_TRAINING);
     const [visibleReportActionsData] = useOnyx(ONYXKEYS.DERIVED.VISIBLE_REPORT_ACTIONS);
+    const [sortedReportActionsData] = useOnyx(ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS);
```
tgolen: It looks like only the `sortedActions` are used from this? If so, I think it would be better to destructure this so that only the `sortedActions` are retrieved and passed down (same goes for the other components).

tgolen (changes requested):
Wow, really cool approach for this! I love it.

src/libs/actions/OnyxDerived/configs/sortedReportActions.ts
truph01: They are currently derived in `Onyx.connect`:

https://github.com/Expensify/App/blob/9fcd07a9e4ff4f408829adc0c940af3acffc81bf/src/libs/OptionsListUtils/index.ts#L218-L223

So once we remove `Onyx.connect`, these should be removed as well.


>  so I assume that those references will be added in future PRs

Yes, `lastActions` and `transactionThreadIDs` will be used in future PRs.

src/hooks/useSearchSelector.base.ts
truph01: I updated PR to apply your suggestion

src/libs/actions/OnyxDerived/configs/sortedReportActions.ts
```diff
@@ -0,0 +1,53 @@
...
+            }
+
+            const reportID = key.replace(ONYXKEYS.COLLECTION.REPORT_ACTIONS, '');
+            if (!reportID) {
+                continue;
+            }
+
+            const reportActionsArray = Object.values(actions);
+            let sortedReportActions = getSortedReportActions(withDEWRoutedActionsArray(reportActionsArray), true);
+            sortedActions[reportID] = sortedReportActions;
```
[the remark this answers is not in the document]
DylanDylann: @truph01 I think this is valid. Could you take a look? We should also add some performance measurements in this PR to see if we need to optimize this

src/components/Search/SearchFiltersParticipantsSelector.tsx
```diff
@@ -97,6 +99,7 @@ function SearchFiltersParticipantsSelector({initialAccountIDs, onFiltersUpdate,
     const [nvpDismissedProductTraining] = useOnyx(ONYXKEYS.NVP_DISMISSED_PRODUCT_TRAINING);
     const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
     const [recentAttendees] = useOnyx(ONYXKEYS.NVP_RECENT_ATTENDEES);
+    const [{sortedActions} = getEmptyObject<OnyxTypes.SortedReportActionsDerivedValue>()] = useOnyx(ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS);
```
DylanDylann: I’d suggest avoiding a default value here. A safer approach is to let Onyx return undefined and handle the fallback (e.g., defaulting,..) inside the function logic itself. This prevents bugs if someone forgets to set a default in future useOnyx() calls

tgolen: Yeah, I agree with this as well.

tgolen (dismissed):
I approve, assuming the comments from @DylanDylann are addressed.

src/libs/actions/OnyxDerived/configs/sortedReportActions.ts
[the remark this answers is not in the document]
truph01: @MelvinBot To address this concern, what do you think about the following solution?

Before: ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS computed and persisted {sortedActions, lastActions, transactionThreadIDs} to disk via Onyx.set(). The sortedActions field — a full sorted copy of every report's actions — was the bulk of the data, duplicating REPORT_ACTIONS and causing expensive disk writes on every message.

After: The derived value only persists {lastActions, transactionThreadIDs} (lightweight — one entry per report). The expensive sortedActions is computed in-memory by a new useSortedReportActions hook using useMemo, with the exact same sorting logic. It lives only in RAM and is never written to disk.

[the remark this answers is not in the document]
MelvinBot: Your proposed split is a sound approach and directly addresses the P1 concern. Here's my analysis:

### Why the current approach is expensive

`setDerivedValue` in `src/libs/actions/OnyxDerived/utils.ts:23-26` calls `Onyx.set()` with `skipCacheCheck: true`, which unconditionally triggers `storage.setItem()` — every recompute writes to disk (IndexedDB/SQLite). Since `SORTED_REPORT_ACTIONS` isn't registered as a RAM-only key in `Onyx.init()` (`src/setup/index.ts`), the full `sortedActions` (`Record<string, ReportAction[]>`) — a sorted duplicate of the entire `REPORT_ACTIONS` collection — gets serialized to disk on every incoming message. The compute function also lacks incremental-update optimization (unlike `visibleReportActions.ts` which inspects `sourceValues` to only recompute changed keys), so it re-sorts **all** reports on every change.

### Your proposed split: verdict

**Good direction.** `lastActions` and `transactionThreadIDs` are genuinely lightweight (one entry per report) and appropriate for persisted derived state. Moving `sortedActions` out eliminates the biggest cost.

### Two implementation paths to consider

**Option A — `useSortedReportActions` hook (your proposal):**
- Consumers (`useSearchSelector.base.ts`, `NewChatPage.tsx`, `SearchFiltersParticipantsSelector.tsx`) already have `REPORT_ACTIONS` in Onyx's in-memory cache, so a `useMemo` sort doesn't trigger storage reads.
- Precedent exists: `usePaginatedReportActions.ts:28-41` already sorts per-report via a `useCallback` selector.
- **Important**: the hook must include `withDEWRoutedActionsArray` in the sort pipeline (as the current config does) or the ordering will differ from what the report view uses.
- Tradeoff: each mounted instance holds its own sorted copy in the JS heap, and the sort runs in the render thread. Since the current OnyxDerived config also sorts all reports on every change (no incremental optimization), computation cost is comparable — you're just trading disk I/O for render-thread CPU.

**Option B — Register as RAM-only key (simpler alternative):**
Adding `ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS` to `Onyx.init()`'s `ramOnlyKeys` in `src/setup/index.ts` would skip `storage.setItem()` entirely while keeping the centralized single-computation model. This avoids disk writes **and** avoids duplicating computation across multiple hook instances. It's a one-line change.

### Recommendation

If the primary goal is eliminating expensive disk writes, **Option B (RAM-only)** is the simplest fix with the smallest diff. If you also want to avoid the global recompute-all-reports behavior (which is a real secondary concern), then **Option A (useMemo hook)** is better long-term, especially if you add per-report memoization so that only the changed report's actions get re-sorted.

Either way, I'd also recommend adding incremental-update logic to the remaining `{lastActions, transactionThreadIDs}` OnyxDerived config, similar to how `visibleReportActions.ts` uses `sourceValues` to only recompute changed reports.

src/libs/OptionsListUtils/index.ts
```diff
@@ -2289,11 +2305,13 @@ function prepareReportOptionsForDisplay(
         let lastIOUCreationDate;
         // Add a field to sort the recent reports by the time of last IOU request for create actions
         if (preferRecentExpenseReports) {
-            const reportPreviewAction = allSortedReportActions[option.reportID]?.find((reportAction) => isActionOfType(reportAction, CONST.REPORT.ACTIONS.TYPE.REPORT_PREVIEW));
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
```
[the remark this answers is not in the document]
truph01: Fixed

src/components/Search/SearchFiltersParticipantsSelector.tsx
truph01: I removed the default value.

src/hooks/useSearchSelector.base.ts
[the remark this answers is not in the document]
truph01: @MelvinBot, I think it is the tradeoff when we decide to remove ONYXKEYS.COLLECTION.REPORT_ACTIONS in src/libs/OptionsListUtils.ts and use the value from useOnyx instead.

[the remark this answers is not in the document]
MelvinBot: Fair point — moving from `Onyx.connect()` callbacks (which update module-level variables without triggering React re-renders) to `useOnyx` hooks inherently means component re-renders on data changes. That's the tradeoff of this migration.

That said, even within the `useOnyx` paradigm, selectors can help mitigate the jank concern raised here. For example, `useSearchSelectorBase()` could use a `selector` on its `useOnyx(SORTED_REPORT_ACTIONS)` call that extracts only the specific fields it actually needs (e.g., just `lastActions` or `transactionThreadIDs`) and returns a referentially stable result when those fields haven't changed — avoiding the full re-render cascade from unrelated message changes.

So the tradeoff is real but not unavoidable at this severity level — it's a question of how surgical the selectors are.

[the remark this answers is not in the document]
truph01: @MelvinBot Could you please generate a selector for `useOnyx(SORTED_REPORT_ACTIONS)` to ensure all the necessary data is always included?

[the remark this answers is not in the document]
MelvinBot: Here's a selector following the existing `src/selectors/` convention (same pattern as `Attributes.ts` and `Todos.ts`):

**Create `src/selectors/SortedReportActions.ts`:**

```typescript
import type {OnyxEntry} from 'react-native-onyx';
import type {SortedReportActionsDerivedValue} from '@src/types/onyx/DerivedValues';

const sortedActionsSelector = (value: OnyxEntry<SortedReportActionsDerivedValue>) => value?.sortedActions;

export default sortedActionsSelector;
```

**Usage at both consumer sites** (`useSearchSelector.base.ts` and `NewChatPage.tsx`):

```typescript
import sortedActionsSelector from '@selectors/SortedReportActions';

// Replace:
const [sortedReportActionsData] = useOnyx(ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS);
const sortedActions = sortedReportActionsData?.sortedActions;

// With:
const [sortedActions] = useOnyx(ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS, {selector: sortedActionsSelector});
```

This ensures `useOnyx` only triggers re-renders when the `sortedActions` field changes (not when `lastActions` or `transactionThreadIDs` change), and keeps the extraction logic centralized.

If you later need a per-reportID selector (for a component that only cares about one report's sorted actions), you can add a curried variant:

```typescript
const sortedActionsByReportIDSelector = (reportID: string) =>
    (value: OnyxEntry<SortedReportActionsDerivedValue>) => value?.sortedActions?.[reportID];
```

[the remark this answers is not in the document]
truph01: I added selector

src/libs/actions/OnyxDerived/configs/sortedReportActions.ts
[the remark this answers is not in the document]
truph01: I updated PR to use Option B

tgolen (dismissed):

src/libs/OptionsListUtils/index.ts
```diff
@@ -2289,11 +2304,12 @@ function prepareReportOptionsForDisplay(
...
```
DylanDylann:
```suggestion
```

src/libs/OptionsListUtils/types.ts
```diff
@@ -221,6 +222,7 @@ type GetOptionsConfig = {
     countryCode?: number;
     visibleReportActionsData?: VisibleReportActionsDerivedValue;
     reportAttributesDerived?: ReportAttributesDerivedValue['reports'];
+    sortedActions?: Record<string, ReportAction[]>;
```
DylanDylann, line 226: Please add a comment noting that we’ll remove the optional operator in a follow-up PR, and include the related issue link.

src/setup/index.ts
```diff
@@ -60,6 +60,7 @@ export default function () {
         },
         skippableCollectionMemberIDs: CONST.SKIPPABLE_COLLECTION_MEMBER_IDS,
         snapshotMergeKeys: ['pendingAction', 'pendingFields'],
+        ramOnlyKeys: [ONYXKEYS.DERIVED.SORTED_REPORT_ACTIONS],
```
DylanDylann, line 63: This approach makes sense. I’m not sure why we haven’t used it before. I wonder if we could also consider adding other derived values to the ramOnlyKeys array to minimize disk writes.

src/libs/OptionsListUtils/index.ts
```diff
@@ -2277,9 +2291,10 @@ function prepareReportOptionsForDisplay(
         let isOptionUnread = option.isUnread;
         if (shouldUnreadBeBold) {
             const chatReport = allReports?.[`${ONYXKEYS.COLLECTION.REPORT}${report.chatReportID}`];
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
```
DylanDylann: Looks redundant
```suggestion
```

DylanDylann (approved):

src/setup/index.ts
truph01: > I wonder if we could also consider adding other derived values to the ramOnlyKeys array to minimize disk writes.

I think we can explore that in a follow-up. The reason it hasn't been used before is that data in ramOnlyKeys is cleared when the app closes, so every derived value needs to be fully recomputed on the next cold start.

src/libs/OptionsListUtils/index.ts
truph01: @DylanDylann I just removed it

src/setup/index.ts
DylanDylann: Yeah, I see. But I think it works for some other derived values. Anyway, let’s move forward for now, I’ll start a thread on Slack to discuss this further.

tgolen (approved):

Issue #85368, opened 2026-03-16
[Due for payment 2026-04-14] Odometer - Removed photo is shown on confirm page when it is removed from Distance page


nlemma:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.38-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** N/A - new feature, doesn't exist in prod
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://github.com/Expensify/App/pull/83556
**Email or phone of affected tester (no customers):** applausetester+sjddjois2j1d@applause.expensifail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** Mac 26.2 / Chrome
**App Component:** Money Requests

## Action Performed:
Precondition:
- Log in with Expensifail account.

1. Go to staging.new.expensify.com
2. Go to workspace chat.
3. Click + > Track distance > Odometer.
4. Enter Start and End reading.
5. Insert Start and End photo.
6. Click Next.
7. On confirm page, click Distance.
8. Click End photo.
9. Remove the photo.
10. Click RHP back button.
11. Click Distance.

## Expected Result:
Since odometer photo is removed from Distance page, receipt on confirm page will not show the removed photo.

## Actual Result:
Odometer photo is removed from Distance page, but the receipt on confirm page still shows the removed photo.

## Workaround:
Unknown

## Platforms:
- [x] Android: App
- [x] Android: mWeb Chrome
- [x] iOS: App
- [x] iOS: mWeb Safari
- [x] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/a39d3d44-8306-48bb-be61-833aec4af8a2

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @DylanDylann</details>

Julesssss:
Requested retesting [here](https://expensify.slack.com/archives/C09V78U42D8/p1774477074542509?thread_ts=1774477072.240439&cid=C09V78U42D8)

PR #85105, opened 2026-03-12
[Odometer] Move odometer image stitching to confirmation page

jakubkalinski0:
### Explanation of Change

Moves odometer image stitching from the odometer reading step to the confirmation page. Previously, stitching blocked navigation with a loading button. Now, stitching runs on the confirmation page while an `ActivityIndicator` replaces the receipt preview, and errors surface via `FormHelpMessage`. Also extracts the stitched filename prefix to a constants file and skips stitching when only one image is present.

### Fixed Issues

$ https://github.com/Expensify/App/issues/84998
PROPOSAL: N/A

### Tests

> [!IMPORTANT]  
> This PR only moves the stitching logic to the confirmation page and does't introduce the backup transaction during editing from confirmation page (this will be added here https://github.com/Expensify/App/pull/86022). Because of that when user makes any changes to the images when editing from confirmation page this change will be reflected on confirmation page as receipt will be re-stitched (no matter if he pressed `"Save"` or `<` back button)

We are testing the odometer expense creation flow which you start by doing the following:
Press `FAB` -> go to `"Track distance"` -> choose `"Odometer"` tab

**Test 1 - Loading indicator while stitching**
1. Create a new odometer distance expense.
2. Capture a start odometer photo and an end odometer photo and add both readings.
3. Navigate to the confirmation page.
4. Verify a loading indicator appears in the receipt thumbnail area while stitching runs, then the stitched combined image appears.

---

**Test 2 - Re-stitching Deleting one photo clears the stale receipt**
5. Create an odometer expense with start + end photos and navigate to the confirmation page.
6. Press `"Distance"` and replace/delete/crop/rotate one of the photos.
7. Then press `"Save"` and return to the confirmation page.
8. Verify that the receipt is re-stitched accordingly to the changes photo is adjusted to the changes (not the previously stitched version).

---

**Test 3 - Single odometer image (no stitching)**
9. Create an odometer expense but add only a start photo (skip the end photo).
10. Navigate to the confirmation page.
11. Verify the single image is shown as the receipt directly, with no loading indicator and no stitching attempt.

- [x] Verify that no errors appear in the JS console

### Offline tests

Same as **Tests**

### QA Steps

Same as **Tests**

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
[...]

PR #86022, opened 2026-03-21
[Odometer] Add backup transaction for purpose of editing from confirmation

jakubkalinski0:
### Explanation of Change

PR https://github.com/Expensify/App/pull/85105 has to be merged before we can go ahead with this one

Adds backup transaction support for editing odometer readings from the confirmation page. On entry, a fresh backup is created (bypassing AsyncStorage race conditions). Canceling restores the backup and revokes new blob URLs. Saving cleans up the backup and revokes the old URLs. Also skips redundant image re-stitching when source images are unchanged. **If user didn't press `"Save"` then we interpret it as an intention to discard the changes and stick with the original images/readings.**

### Fixed Issues
$ https://github.com/Expensify/App/issues/85368 - main issue
$ https://github.com/Expensify/App/issues/86355 - related small bug
PROPOSAL: N/A

### Tests

We are testing the odometer expense creation flow which you start by doing the following:
Press `FAB` -> go to `"Track distance"` -> choose `"Odometer"` tab

1. Create a new odometer expense, add both readings and upload start and end odometer images.
2. Press `"Next"` to reach the confirmation page and verify the stitched receipt is shown.
3. Press `"Distance"` to land on the odometer screen.
4. Replace one of the odometer images and one of the readings.
5. Tap `"Save"` and navigate back to the confirmation page.
6. Verify the stitched receipt updates to reflect the new image and distance reflects the changed reading.

---

7. Follow steps 1-4 above.
8. Tap `<`**back button** instead of `"Save"`.
9. Verify the confirmation page shows the original odometer images (not the newly selected ones) and distance.
10. Once again press `"Distace"` to land on the odometer screen and verify that the odometer images and readings are the original ones, not the changed ones that we dicarded by tapping `<` **back button**

---
This test steps below verify that this issue is fixed -> [Expense - Odometer photo is closed immediately after clicking rotate](https://github.com/Expensify/App/issues/86355)
11. Create a new odometer expense, add both readings and upload **ONLY ONE ODOMETER IMAGE** (start or end).
12. Press `"Next"` to reach the confirmation page and verify the stitched receipt is shown.
13. Press `"Distance"` to land on the odometer screen.
14. Click the uploaded photo.
15. Click `"Rotate"`.
16. Verify that the photo is rotated and you are **NOT** brought back to the Odometer Tab (thats what happened in a bug described here https://github.com/Expensify/App/issues/86355)

### Offline tests

Same as **Tests**

### QA Steps

Same as **Tests**

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

**iOS needs to be tested on a physcical device**

<details>

https://github.com/user-attachments/assets/ce238023-5d62-4ffa-8f52-55325d69bf43

</details>

<details>

https://github.com/user-attachments/assets/12d6c8a8-1dc5-43cd-ac8d-ae33108bfde7

</details>

<details>

https://github.com/user-attachments/assets/53c7d4d3-9bb5-45cb-ab5b-efe48fed0874

</details>

<details>

https://github.com/user-attachments/assets/10363460-6116-45fe-b691-b31c32c45cb7

</details>


src/libs/actions/IOU/index.ts
```diff
@@ -1685,30 +1686,15 @@ function setMoneyRequestOdometerReading(transactionID: string, startReading: num
...
 /**
  * Set odometer image for a transaction
  * @param transactionID - The transaction ID
  * @param imageType - 'start' or 'end'
  * @param file - The image file (File object on web, URI string on native)
  * @param isDraft - Whether this is a draft transaction
+ * @param shouldRevokeOldImage - Whether to revoke the previous blob URL immediately (false when a backup transaction exists making the caller responsible for revoking)
  */
-function setMoneyRequestOdometerImage(transactionID: string, imageType: OdometerImageType, file: FileObject | string, isDraft: boolean) {
+function setMoneyRequestOdometerImage(transactionID: string, imageType: OdometerImageType, file: FileObject | string, isDraft: boolean, shouldRevokeOldImage = true) {
```
DylanDylann: I think `shouldRevokeOldImage` should default to false, and we shouldn’t pass this parameter in the native file.

src/pages/iou/request/step/IOURequestStepOdometerImage/index.native.tsx
```diff
@@ -183,10 +184,7 @@ function IOURequestStepOdometerImage({
         }
 
         const file = files.at(0);
-        if (!file) {
-            return;
-        }
```
DylanDylann, line 188: Why do you remove this check? I think we should integrate this check to getOdometerImageUri function

src/pages/iou/request/step/IOURequestStepConfirmation.tsx
```diff
@@ -431,14 +438,21 @@ function IOURequestStepConfirmation({
     const odometerEndImage = transaction?.comment?.odometerEndImage;
```
DylanDylann: Please put these into useEffect

src/pages/iou/request/step/IOURequestStepDistanceOdometer.tsx
```diff
@@ -348,11 +371,14 @@ function IOURequestStepDistanceOdometer({
 
     const navigateBack = useCallback(() => {
         if (isEditingConfirmation) {
-            Navigation.goBack(confirmationRoute);
+            backupHandledManually.current = true;
+            restoreOriginalTransactionFromBackupWithImageCleanup(transactionID, isTransactionDraft, () => {
+                Navigation.goBack(confirmationRoute);
+            });
```
DylanDylann, line 378: Why do we need to call restoreOriginalTransactionFromBackupWithImageCleanup here? Should we handle this in the useEffect cleanup instead by removing the backupHandledManually early return?

src/pages/iou/request/step/IOURequestStepDistanceOdometer.tsx
```diff
@@ -231,6 +234,26 @@ function IOURequestStepDistanceOdometer({
         }
     }, [currentTransaction?.comment?.odometerStart, currentTransaction?.comment?.odometerEnd, isEditing]);
 
+    useEffect(() => {
+        if (!isEditingConfirmation) {
+            return () => {};
+        }
+        createBackupTransaction(transaction, isTransactionDraft, true);
```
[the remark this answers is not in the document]
DylanDylann: I believe IOURequestStepDistanceOdometerWithFullTransactionOrNotFound already covers this case

src/libs/actions/IOU/index.ts
jakubkalinski0: It's a sensible change -  I will adjust the code

jakubkalinski0: I see only one issue with it. When you think about it, by default we always want to remove stale data and when we replace `shouldRevokeOldImage = true` with `false` someone in the future might mess it up and forget about passing `true` which would lead to memory leak. Do we still want to go ahead with this change then @DylanDylann?

src/pages/iou/request/step/IOURequestStepOdometerImage/index.native.tsx
jakubkalinski0: I've come to the conclusion that this is essentially equivalent to the check above:
```
        if (files.length === 0) {
            return;
        }
```

jakubkalinski0: If we were to add this check directly to `getOdometerImageUri` function then it wouldn't actually change anything since we already handle `null/undefined` case in `getOdometerImageUri`. I think that we can restore the original check if we want to have 100% guarantee that this file in fact exists, what do you think about that @DylanDylann?

src/pages/iou/request/step/IOURequestStepDistanceOdometer.tsx
jakubkalinski0: We actually need this `restoreOriginalTransactionFromBackupWithImageCleanup` here. That's because the cleanup effect fires after unmount, which is after navigation - meaning the confirmation page starts rendering the new `blob URL` before it gets revoked. That causes the infinite loading spinner

The exact flow that happens if we rely on cleanup in `useEffect`:
1. `Navigation.goBack(confirmationRoute)` - confirmation page starts mounting immediately
2. The confirmation page reads the current transaction from `Onyx`, sees the new/replaced image as a `blob URL` (`blob://...xyz`) and begins rendering the stitched receipt
3. The odometer component unmounts -> cleanup effect fires -> `revokeOdometerImageUri` invalidates `blob://...xyz`
4. The confirmation page is already trying to render that URL - now it's revoked -> infinite loading spinner

The callback pattern ensures that blob URLs are revoked before `Navigation.goBack()` is called so the confirmation page always mounts with valid (original) images already in Onyx

src/pages/iou/request/step/IOURequestStepConfirmation.tsx
jakubkalinski0: Thanks for catching that!

src/libs/actions/IOU/index.ts
DylanDylann: To be honest, both approaches are fine. My initial thought is that removing the old image isn’t necessary in all cases, so we shouldn’t default it to true, for example, in native. That said, we could make this parameter required.

src/pages/iou/request/step/IOURequestStepOdometerImage/index.native.tsx
DylanDylann: Makes sense. Let’s stick with your approach.

src/pages/iou/request/step/IOURequestStepDistanceOdometer.tsx
DylanDylann: Interesting. Thanks for the explanation. It makes sense now.

DylanDylann:
@jakubkalinski0 I appreciate your [explanation](https://github.com/Expensify/App/pull/86022#discussion_r2996157130), it’s very reasonable. I think the initial idea of `transactionsBackup_` wasn’t to save the entire transaction. In other cases, we also only use part of it.

Anyway, within the scope of this PR, let’s keep it as is. If we agree to update it later, it would make sense to apply the same change to other backup transactions as well.

src/libs/actions/IOU/index.ts
jakubkalinski0: > removing the old image isn’t necessary in all cases

Yeah, I completely agree with you, that's why my initial thought was to go with your suggestion. I just brought up the alternative to discuss this issue I saw.

> we could make this parameter required

I am a big fan of this idea. I think it would very nicely solve both our points, thanks for bringing up that topic and for this short, yet productive discussion 😄

src/pages/iou/request/step/IOURequestStepOdometerImage/index.native.tsx
jakubkalinski0: Lovely, then I will restore this check then. Thank you!

DylanDylann: I’m fine with whichever you prefer

jakubkalinski0:
> The backup behaviour is only needed for the images so I think specific logic for them is acceptable

If you agree with this then it seems completely reasonable to me!

> Also it means we don't need to perform an Onyx operation for every mount/unmount of this page

Yeah, that would require some logic in `TransactionReceiptModalContent` as we would have to create the backup transaction whenever user taps on any of the `crop/rotate/replace/delete` buttons (and also verify if such backup transaction wasn't already created before as user could've already made some changes to the image before). But it's definitely doable.

> One thing I was thinking as a far simpler alternative would be to just drop the images immidiately and not care that users did not hit save... but that's not ideal

Agreed.

> I strongly agree. We should only create a backup transaction when an image is actually added or updated.

Yup, I completely agree with that too.

> I think the initial idea of transactionsBackup_ wasn’t to save the entire transaction. In other cases, we also only use part of it.

Hmm... I wasn't aware of that. If that's the case then we could stick with this key and only add new functions that handle only the images (or even maybe some more general function for backup of partial transaction which would accept transaction fields that have to be saved).

> Lets go with the current solution 👍

> Anyway, within the scope of this PR, let’s keep it as is.

Sure, sounds good to me to just move with the current solution.

> If we agree to update it later, it would make sense to apply the same change to other backup transactions as well.

That would definitely be worth exploring!

jakubkalinski0:
@DylanDylann All yours, if checks pass of course 😄

DylanDylann (approved):

Julesssss:
> In my opinion user should be able to change his mind and not save the changes he made, which is basically the point of this PR. My question is, should we show DiscardChangesConfirmation modal when user taps < back button to make it seem more like an intentional decision and not a bug?

Yeah good idea. Missed this comment yesterday! 

> In this PR, as discussed above, we have 2 improvement points... ...that we should handle this separately (at least for point 1, since it’s worth cleaning up other usages of transactionBackup as well). So I will approve this PR in advance, then you can merge it if you agree

Yeah thanks for confirming. I'm happy with how the changes look currently. To be honest I am now wondering how important these improvements are...  I might suggest we don't apply the partial transactionBackup changes to other expense edits if we can get avoid that refactor. Anyway that is a discussion for later.

Julesssss:
@jakubkalinski0 follow up is fine, assuming you don't need to make changes for the cropping issue 👍

jakubkalinski0:
The `saveCrop` function in `TransactionReceiptModalContent.tsx` passes [cropRect](https://github.com/Expensify/App/blob/71ea695235df10fb95da1aaeeffe4e17dd3f0c6f/src/pages/media/AttachmentModalScreen/routes/TransactionReceiptModalContent.tsx#L370-L373) values directly to the native image manipulator without rounding. [ReceiptCropView/index.tsx](https://github.com/Expensify/App/blob/71ea695235df10fb95da1aaeeffe4e17dd3f0c6f/src/components/ReceiptCropView/index.tsx#L314-L319) computes these via `width = displayWidth × (imageWidth / displayWidth)`. Due to `IEEE 754` floating-point arithmetic, this `a × (b/a)` operation produces `793.0000000000001` instead of `793.0` for the odometer image. It's an issue since the native `CropTransformer.kt` validates if `rect.width` <= `bitmap.width`. For the odometer we have the following `793.0000000000001` <= `793` -> `false` -> `ImageInvalidCropException`.

Julesssss (approved):

PR #86651, opened 2026-03-30
[No QA] Revert "[Perf] Scope REPORT_ACTIONS subscription to per-row level in LHNOptionsList   "

cristipaval:
Reverts Expensify/App#84262 because even if it reduces the LHN rerenders when the report action collection changes, it might introduce performance issues when scrolling the LHN. See [this](https://github.com/Expensify/App/pull/84262#issuecomment-4152973739) for more details.


srikarparsi (approved):

IuliiaHerets:
Hi @cristipaval. Can we check it off?

cc @srikarparsi

Issue #83634, opened 2026-02-26
ManualSubmitToDestinationVisible - Measure end-to-end time from expense creation submit to destination page render


mountiny:
## Overview

We need a new end-to-end performance metric that measures the time from when the user taps the **Create Expense** button (the submit/confirm action on the confirmation step) until the **destination page is fully rendered and visible** to the user.

This metric should cover **all expense types** and **all possible navigation destinations** after expense creation.

## Target

I think we should aim for `400ms`

## Why this matters

After a user confirms an expense, they experience a transition period: the confirmation modal dismisses, navigation occurs, and a destination page renders. This is a critical moment in the user experience — the user is waiting to see the result of their action. Any delay here feels like lag or broken behavior. We currently lack visibility into this end-to-end duration across all creation flows.

## Current telemetry coverage

We already have several spans measuring parts of the expense creation flow:

| Span | What it measures | Gap |
|------|-----------------|-----|
| `ManualOpenCreateExpense` | Opening the creation UI (from `startMoneyRequest()` to confirmation page layout) | None — this is a separate flow |
| `ManualShutterToConfirmation` | Camera shutter press to confirmation screen layout (scan only) | None — this is a separate flow |
| `ManualCreateExpenseSubmit` | Submit button click to API dispatch completion | Ends at API dispatch, not at destination render |
| `ManualCreateExpenseNavigation` | Post-submit navigation to destination | **Only covers the Search page destination** — all other destinations are unmeasured |

The `ManualCreateExpenseNavigation` span currently only ends when the Search list component lays out (`markNavigateAfterExpenseCreateEnd()` in `src/components/Search/index.tsx`). This means we have no measurement for the majority of post-creation navigation scenarios.

## What to measure

**Start point**: The moment the user taps the Create/Submit/Confirm button on the expense confirmation step (in `IOURequestStepConfirmation.tsx`).

**End point**: The moment the destination page is fully rendered and visible to the user (e.g., `onLayout` of the main content list/view on the destination screen).

**Duration**: The entire end-to-end time including optimistic data application, modal dismissal animation, navigation, and destination page render.

## Expense types to cover

All expense creation request types:
- **Manual** expense
- **Scan** (receipt) expense
- **Distance** expense (manual, GPS, odometer variants)
- **Per diem** expense
- **Time** expense

All IOU types:
- **Submit** (request money from someone)
- **Track** (track expense for self)
- **Split** (split bill among participants)
- **Invoice** (send invoice)
[...]

PR #86738, opened 2026-03-31
Optimize post-expense search page transition with progressive rendering

JakubKorytko:
### Explanation of Change

Optimizes the post-expense-creation search page transition on narrow (mobile) layouts. Instead of waiting for the heavy `Search` component to mount, we show lightweight static stand-ins and progressively upgrade to the full interactive UI.

**Rendering phases (submit-and-navigate flow only):**

| Phase | Trigger | List area | Chrome (tabs, header, filters) | Interactive? |
|-------|---------|-----------|-------------------------------|--------------|
| 1 - Static list | Navigation lands | `SearchStaticList` renders real data via plain `FlatList` (max 10 items) | Static pulsing stand-ins (`StaticTabSelector`, `StaticSearchPageHeader`, `StaticFiltersBar`) | Rows are tappable |
| 2 - Header interactive | `SearchStaticList.onLayout` fires | Still `SearchStaticList` (unchanged) | Pulsing stops, real chrome components mount | Header interactive, list rows tappable |
| 3 - Search replaces static | `useEffect` sets `isInteractive=true` via `startTransition` | `Search` mounts and receives `staticListContent` as `initialContent` prop - renders same static content on first frame while hooks initialize | Real chrome | Header interactive |
| 4 - Fully interactive | `Search` finishes deferred work (`shouldShowRowSkeleton` clears) | Full `FlashList` replaces `initialContent` | Real chrome | Yes (everything) |

Key design: there is **no overlay**. `SearchStaticList` is rendered alone first, then replaced by `Search` which receives the same `staticListContent` element as `initialContent`. This avoids having two list components mounted simultaneously, preventing UI thread contention.

**Other changes:**
- `skipWaitForWrites` -- fires the search API immediately when a deferred write is pending (no risk of overwriting optimistic data since search snapshots live in separate Onyx keys)
- Synchronous navigation in `handleNavigateAfterExpenseCreate` when `navigationRef` is already ready
- Cached optimistic item index so re-injection preserves sorted position instead of appending to the end
- `PulsingView` -- reusable Reanimated component for opacity pulse animation with `wrapperStyle` prop for opaque backgrounds that shouldn't pulse
- `skipSkeleton` prop on `SearchInputSelectionWrapper` to avoid double-skeleton on the static path
- `SearchStaticList` uses `useSession()` context instead of `useCurrentUserPersonalDetails()` to avoid extra Onyx subscriptions
- Single `PulsingView` wrapper around all static header components

**Performance gains (`ManualSubmitToDestinationVisible` span):**

| Platform | `main` avg (ms) | This branch avg (ms) | Improvement |
|----------|-----------------|----------------------|-------------|
| Android  | 1212            | 615                  | **-597 ms (49%)** |
| iOS      | 1736            | 1118                 | **-618 ms (36%)** |

<details>
<summary>Raw measurements (the first one is cold)</summary>

- **Android `main`** (6 runs): 1684, 1079, 1085, 1095, 1079, 1249 ms
- **Android `this branch`** (6 runs): 617, 628, 579, 575, 604, 686 ms
- **iOS `main`** (5 runs): 1804, 1758, 1590, 1863, 1664 ms
- **iOS `this branch`** (5 runs): 904, 1083, 1332, 1129, 1143 ms

</details>

### Fixed Issues

$ https://github.com/Expensify/App/issues/83634
PROPOSAL: N/A

### Tests

1. Open the app on a narrow-layout device (or resize the browser to mobile width)
2. Navigate to the expense creation flow (e.g. tap **+** -> **Submit expense**), fill in the required fields, and submit
3. Verify that after submission you are taken to the Search page and a static list of search results with pulsing tab selector, header, and filters bar appears almost instantly (no blank screen or long skeleton)
4. Wait a moment for the full interactive Search component to mount
5. Verify that the pulsing animation stops and the page becomes fully interactive (tabs are tappable, filters work, search input is focusable)

### Offline tests

1. Submit an expense while offline (airplane mode / network disabled)
2. Verify the search page still shows a pulsing skeleton placeholder for the pending expense
3. Re-enable the network and verify the optimistic expense resolves into the real search list once the API responds

### QA Steps

1. On staging, open the app on a mobile device or narrow browser viewport
2. Submit an expense and verify the transition to the Search page is fast with the static list and pulsing chrome visible immediately
3. Verify that after the full interactive components load, all search functionality works correctly (tabs, filters, search input, tapping on expense rows)

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/5185b302-b394-476a-93be-a1b68aaf81cc

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/8e196a36-1b4f-4656-8401-94397dcdd720

</details>

<details>

</details>

<details>

</details>


src/components/Search/index.tsx
```diff
@@ -238,15 +241,18 @@ function Search({
...
+    // skeleton placeholder so the UI doesn't get stuck. The stableSortedData
+    // cache (cachedOptimisticItemRef) is intentionally kept alive so the item
+    // stays visible at its sorted position until server-confirmed data arrives;
+    // clearOptimisticTracking handles that cleanup when pendingAction !== ADD.
     useEffect(() => {
         if (!hasPendingWriteOnMountRef.current || !optimisticWatchKeyRef.current) {
             return;
         }
-        const id = setTimeout(clearOptimisticTracking, OPTIMISTIC_TRACKING_TIMEOUT_MS);
+        const id = setTimeout(() => setShowPendingExpensePlaceholder(false), OPTIMISTIC_TRACKING_TIMEOUT_MS);
```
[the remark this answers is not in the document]
JakubKorytko: Added a rollback safety net for optimistic tracking. If a cached optimistic item disappears from `sortedData` (e.g. failed/rolled-back submit) and never returns with `pendingAction !== ADD`, `stableSortedData` was re-injecting it forever as a ghost row. Fix: when the cached item vanishes, a 3s grace-period timeout schedules `clearOptimisticTracking()`. If the item reappears (stale snapshot), the timeout is cancelled - happy path unchanged.

src/components/Search/SearchInputSelectionWrapper/index.native.tsx
```diff
@@ -7,8 +7,8 @@ import SearchInputSelectionSkeleton from '@components/Skeletons/SearchInputSelec
 // Once initialized, subsequent mounts are fast, so we only show the skeleton once per app session.
 let isAutocompleteInputInitialized = false;
 
-function SearchInputSelectionWrapper({ref, ...props}: SearchAutocompleteInputProps) {
-    const [showSkeleton, setShowSkeleton] = useState(!isAutocompleteInputInitialized);
+function SearchInputSelectionWrapper({ref, skipSkeleton, ...props}: SearchAutocompleteInputProps & {skipSkeleton?: boolean}) {
```
staszekscp, line 10: Can `skipSkeleton` be a part of `SearchAutocompleteInputProps`?

src/components/Search/index.tsx
```diff
@@ -89,13 +89,19 @@ type SearchProps = {
...
 
 // Max time (ms) to keep the optimistic item cache/skeleton alive before
 // clearing all tracking state. Must be longer than deferredLayoutWrite's
 // 5s safety timeout so the API.write() has time to apply optimistic data.
 const OPTIMISTIC_TRACKING_TIMEOUT_MS = 10_000;
 
+// Grace period (ms) before clearing optimistic tracking after a cached item
+// disappears from sortedData. Short enough to clean up rolled-back items,
+// long enough to survive a brief stale-snapshot gap.
+const OPTIMISTIC_ROLLBACK_GRACE_MS = 3_000;
```
staszekscp: How did you choose this value? It seems like it should be tied to the `OPTIMISTIC_TRACKING_TIMEOUS_MS`

src/components/Search/SearchStaticList.tsx
```diff
@@ -0,0 +1,256 @@
...
+    const {translate, localeCompare, formatPhoneNumber} = useLocalize();
+    const {accountID, email} = useCurrentUserPersonalDetails();
+
+    const [showPendingExpensePlaceholder] = useState(() => hasDeferredWrite(CONST.DEFERRED_LAYOUT_WRITE_KEYS.SEARCH));
+
+    const {type, status, sortBy, sortOrder, groupBy} = queryJSON;
+    const validGroupBy = getValidGroupBy(groupBy);
+    const searchData = searchResults?.data;
+
+    const sortedData = useMemo(() => {
```
staszekscp: Is the component compiled by the React Compiler? If so, we can skip `useMemo` and `useCallback` 😄

src/components/Search/SearchStaticList.tsx
```diff
@@ -0,0 +1,256 @@
...
+            currentAccountID: accountID,
+            currentUserEmail: email ?? '',
+            translate,
+            formatPhoneNumber,
+            bankAccountList: undefined,
+            allReportMetadata: undefined,
+            conciergeReportID: undefined,
+        });
+
+        return (getSortedSections(type, status, filteredData, localeCompare, translate, sortBy, sortOrder, validGroupBy) as TransactionListItemType[]).slice(0, STATIC_LIST_MAX_ITEMS);
```
staszekscp: I think this return statement requires a bit of refining to get rid of casting if possible 😄

src/components/Search/SearchStaticList.tsx
```diff
@@ -0,0 +1,256 @@
...
+                            />
+                        </View>
+                    </View>
+                </PressableWithoutFeedback>
+            );
+        },
+        [styles, onPressItem, theme.highlightBG],
+    );
+
+    const keyExtractor = useCallback((item: TransactionListItemType) => item.keyForList ?? item.transactionID, []);
```
staszekscp: Isn't `keyForList` always an oblicatory property for `TransactionListItemType`? 🤔

src/components/PulsingView.tsx
```diff
@@ -0,0 +1,97 @@
...
+function stopPulse(opacity: SharedValue<number>) {
+    'worklet';
+
+    cancelAnimation(opacity);
+    opacity.set(withTiming(1, {duration: RECOVERY_DURATION, easing: EASING_OUT}));
+}
+
+function PulsingView({shouldPulse, children, style = [], minOpacity = 0.5, needsOffscreenAlphaCompositing = false}: PulsingViewProps) {
+    const shouldPulseShared = useSharedValue(shouldPulse);
+    shouldPulseShared.set(shouldPulse);
```
staszekscp: Is it ok to set new value on every render? It seems like it should be rather wrapped in a `useEffect`

src/components/PulsingView.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    const minOpacityShared = useSharedValue(minOpacity);
+    minOpacityShared.set(minOpacity);
```
staszekscp: Same as above

src/libs/actions/IOU/index.ts
```diff
@@ -1217,13 +1217,19 @@ function handleNavigateAfterExpenseCreate({
         alreadyOnSearchRoot && isSameSearchType ? CONST.TELEMETRY.SUBMIT_FOLLOW_UP_ACTION.DISMISS_MODAL_ONLY : CONST.TELEMETRY.SUBMIT_FOLLOW_UP_ACTION.NAVIGATE_TO_SEARCH,
     );
     const queryString = buildCannedSearchQuery({type});
-    Navigation.isNavigationReady().then(() => {
+    const doNavigate = () => {
```
staszekscp: Definitely NAB, but this name seems a bit odd 😄

src/libs/actions/Search.ts
```diff
@@ -542,9 +551,9 @@ function search({
         });
     }
 
-    return waitForWrites(READ_COMMANDS.SEARCH).then(() => {
+    const fireRequest = () =>
```
staszekscp:
```suggestion
    const startRequest = () =>
```

src/pages/Search/SearchPageNarrow.tsx
```diff
@@ -169,6 +189,68 @@ function SearchPageNarrow({queryJSON, searchResults, isMobileSelectionModeEnable
 
     const isDataLoaded = shouldUseLiveData || isSearchDataLoaded(searchResults, queryJSON);
     const shouldShowLoadingState = !isOffline && (!isDataLoaded || !!metadata?.isLoading);
+    const contentContainerStyle = !isMobileSelectionModeEnabled ? styles.searchListContentContainerStyles : undefined;
+
+    const renderStaticSearchList = () => (
```
staszekscp: Could you explain in a comment what is the difference between static and standard `SearchList`? It seems like instead of standard it should be dynamic to match the naming 😄

src/pages/Search/SearchPageNarrow.tsx
```diff
@@ -150,6 +150,26 @@ function SearchPageNarrow({queryJSON, searchResults, isMobileSelectionModeEnable
         return () => removeRouteKey(route.key);
     }, [addRouteKey, removeRouteKey, route.key, searchRouterListVisible]);
 
+    const [useStaticRendering] = useState(() => !!getPendingSubmitFollowUpAction());
+    const [isInteractive, setIsInteractive] = useState(!useStaticRendering);
+    const [isSearchReady, setIsSearchReady] = useState(!useStaticRendering);
+    const [isChromeInteractive, setIsChromeInteractive] = useState(!useStaticRendering);
```
staszekscp: What do you mean by this state value? "Chrome" seems to be a bit ambiguous

staszekscp (changes requested):

src/components/Search/SearchInputSelectionWrapper/index.native.tsx
JakubKorytko: Unfortunately no - `SearchAutocompleteInput` itself doesn't use `skipSkeleton`, so adding it to `SearchAutocompleteInputProps` triggers `react/no-unused-prop-types`. It's only consumed by the wrapper layer, so the intersection type is the correct approach here.

src/components/Search/index.tsx
JakubKorytko: You're right, it should be tied. Changed it to `OPTIMISTIC_TRACKING_TIMEOUT_MS * 0.3` so it stays proportional (gives us 3s with the current 10s timeout).

src/components/Search/SearchStaticList.tsx
JakubKorytko: Yes it is! Removed all `useMemo` and `useCallback` from the component.

src/components/Search/SearchStaticList.tsx
JakubKorytko: Replaced the `as` cast with a `.filter()` type guard -> `.filter((item): item is TransactionListItemType => 'transactionID' in item)`. No more casting.

src/components/Search/SearchStaticList.tsx
JakubKorytko: Yep, it's a required `string` on the type. Removed the unnecessary `?? item.transactionID` fallback.

src/components/PulsingView.tsx
JakubKorytko: Good catch, wrapped both `.set()` calls in `useEffect` with proper deps.

src/components/PulsingView.tsx
JakubKorytko: Fixed together with the one above.

src/libs/actions/IOU/index.ts
JakubKorytko: Renamed to `navigateToSearch` - more descriptive.

src/libs/actions/Search.ts
JakubKorytko: Done!

src/pages/Search/SearchPageNarrow.tsx
JakubKorytko: Added a comment explaining the difference and renamed `renderStandardSearchList` -> `renderDynamicSearchList` to match the static/dynamic naming.

src/pages/Search/SearchPageNarrow.tsx
JakubKorytko: Fair point - renamed `isChromeInteractive` to `isHeaderInteractive`. It controls whether the surrounding UI elements (tabs, search input, filters) are interactive vs their static placeholders.

src/components/Search/SearchStaticList.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    fixedNumItems={1}
+                    containerStyle={contentContainerStyle}
+                    reasonAttributes={pendingExpenseReasonAttributes}
+                />
+            </View>
+        );
+    }
+
+    if (sortedData.length === 0) {
+        return null;
```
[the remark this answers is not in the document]
JakubKorytko: Good catch, this is a real deadlock scenario. If the deferred write flushes before mount and search data hasn't arrived yet, we get `sortedData.length === 0` with `showPendingExpensePlaceholder === false`, returning `null` and never firing `onLayout`. Fixed by returning an empty `<View onLayout={onLayout} />` instead of null so the transition always completes.

src/pages/Search/SearchPageNarrowStatics.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+    for (const policy of Object.values(policies ?? {})) {
+        if (!policy) {
+            continue;
+        }
+        if (!policy.isJoinRequestPending && shouldShowPolicy(policy, false, undefined)) {
+            workspaceCount++;
+        }
+        if (policy.type === CONST.POLICY.TYPE.TEAM || policy.type === CONST.POLICY.TYPE.CORPORATE) {
+            hasPaidGroupPolicy = true;
```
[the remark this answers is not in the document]
JakubKorytko: This is actually the same check. The static selector does `policy.type === TEAM || policy.type === CORPORATE`, and the real logic uses `isPaidGroupPolicy()` from `PolicyUtils` which does exactly the same thing: `policy?.type === CONST.POLICY.TYPE.TEAM || policy?.type === CONST.POLICY.TYPE.CORPORATE`. The Submit tab eligibility in `getSuggestedSearchesVisibility` is just `isEligibleForSubmitSuggestion = isPaidPolicy` where `isPaidPolicy = isPaidGroupPolicy(policy)` - same predicate. No mismatch here.

src/components/Search/index.tsx
```diff
@@ -1227,14 +1247,24 @@ function Search({
...
+            cachedOptimisticItemIndexRef.current = sortedData.indexOf(optimisticItem);
 
             if (optimisticItem.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.ADD) {
                 clearOptimisticTracking();
             }
+        } else if (cachedOptimisticItemRef.current && !rollbackTimeoutRef.current) {
+            rollbackTimeoutRef.current = setTimeout(() => {
+                rollbackTimeoutRef.current = undefined;
+                clearOptimisticTracking();
+            }, OPTIMISTIC_ROLLBACK_GRACE_MS);
```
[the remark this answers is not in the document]
JakubKorytko: Valid. In the rollback path, `showPendingExpensePlaceholder` is already false (set when the item first appeared), so `setShowPendingExpensePlaceholder(false)` in `clearOptimisticTracking` is a no-op and no re-render happens. The ref mutations are invisible to `stableSortedData` since its only dep is `sortedData`. Fixed by adding an `isOptimisticTrackingCleared` state that `clearOptimisticTracking` flips to true. `stableSortedData` reads it directly in its condition and includes it in deps, so the ghost row is removed on the next render.

src/components/Search/SearchStaticList.tsx
```diff
@@ -0,0 +1,257 @@
...
+    const session = useSession();
+    const accountID = session?.accountID ?? CONST.DEFAULT_NUMBER_ID;
+    const email = session?.email;
+
+    const [showPendingExpensePlaceholder] = useState(() => hasDeferredWrite(CONST.DEFERRED_LAYOUT_WRITE_KEYS.SEARCH));
+
+    const {type, status, sortBy, sortOrder, groupBy} = queryJSON;
+    const validGroupBy = getValidGroupBy(groupBy);
+    const searchData = searchResults?.data;
+
```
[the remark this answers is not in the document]
JakubKorytko: The search snapshot is keyed by query hash (`useOnyx(SNAPSHOT_{currentSearchHash})` in `SearchContext`), so the data returned is always for the current query or undefined. The Onyx race condition documented in `Search/index.tsx` is about transient stale data during an in-place hash change (e.g. switching tabs while already on search). In the submit-and-navigate flow the component mounts fresh with the target hash, so there is no cross-query stale data to guard against.

src/pages/Search/SearchPageNarrow.tsx
```diff
@@ -150,6 +151,31 @@ function SearchPageNarrow({queryJSON, searchResults, isMobileSelectionModeEnable
         return () => removeRouteKey(route.key);
     }, [addRouteKey, removeRouteKey, route.key, searchRouterListVisible]);
 
+    const [useStaticRendering] = useState(() => !!getPendingSubmitFollowUpAction());
```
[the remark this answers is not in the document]
JakubKorytko: Partially valid but low impact. When `alreadyOnSearchRoot && isSameSearchType`, `DISMISS_MODAL_ONLY` is set and `Navigation.navigate(route, {forceReplace: true})` is called. If forceReplace causes a remount, `useStaticRendering` would be `true` and briefly show pulsing static UI on an already-loaded page. In practice the impact is minimal: `SearchStaticList` renders the same Onyx data (no content flash) and `onLayout` fires immediately so the pulsing resolves in one frame. The span would not dangle either way - Search component itself also ends the span via its own `onLayout` and `useFocusEffect` (the focus effect specifically handles `DISMISS_MODAL_ONLY` re-focus). `SearchStaticList` just ends it earlier. Not worth acting on right now.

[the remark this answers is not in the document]
JakubKorytko: On second thought, since the span is safely handled by Search's `useFocusEffect` regardless, there's no reason not to gate this. Changed `useStaticRendering` to only activate for `NAVIGATE_TO_SEARCH`, so `DISMISS_MODAL_ONLY` (already on search) skips the static path entirely and avoids the unnecessary pulsing flash.

JakubKorytko:
@ShridharGoel Next time, please try to provide the minimum number of steps necessary to reproduce the issue, as the video is quite long and it will make it easier for QA's in future. 🙏 The offline + no Onyx issue is reproducible on main, though I agree it's something to look into but out of scope here. The headers themselves are regressions on this PR, so that's a good catch - thanks! I will take care of this.

Repro steps:
1. Go to Troubleshoot page
1. Clear the Onyx cache
1. Turn on the offline mode
1. Add an expense outside of the Inbox 

Video from main:

https://github.com/user-attachments/assets/76900af2-c7d9-4b24-8394-69466aef0394

ShridharGoel:
Discussed over Slack, we can ignore this since it is happening on main also

ShridharGoel (approved):

luacmartins (approved):
I'm slightly concerned with the maintainability of this architecture and the potential class of bugs it might introduce. I think we should revisit this solution once the tab navigation changes are in.

PR #87500, opened 2026-04-09
fix: keep receipt thumbnail dimensions consistent in landscape mode

jakubstec:


jakubstec:
@MelvinBot questions unrelated to this PR

1. what does agent-device exactly do in our project
2. do you have dedicated test accounts, similar to those used by QAs? Can you login to the account, do end to end tests?

jakubstec:
@MelvinBot do you run the test steps during the deploy blocker investigation to test if the issue is reproducible?

jakubstec:
@MelvinBot do you run the test steps during the deploy blocker investigation to test if the issue is reproducible?

PR #88163, opened 2026-04-16
Update README with Expensify engineer instructions

NicolasBonet:
Added instructions for Expensify engineers regarding email configuration and CORS issues.

### Explanation of Change
Points into the right direction after running the iOS/Android apps for expensify engineers pointing to a local VM regarding errors with using their email accounts to start using the app.

### Fixed Issues
$https://github.com/Expensify/App/issues/88164

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ For detailed setup instructions for each platform, see the following guides:
    Access to fetch at 'https://www.expensify.com/api/BeginSignIn' from origin 'http://localhost:8080' has been blocked by CORS policy
    ```
    You probably have a misconfigured `.env` file - remove it (`rm .env`) and try again.
-3. Expensify engineers having an error `The given email is not valid` must [follow this instructions](https://stackoverflowteams.com/c/expensify/questions/23537/23538#23538) to configure the local dev email account.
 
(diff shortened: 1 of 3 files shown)


grgia:
@NicolasBonet I can't add a review because this issue is locked. BUT

Rather than adding a note to the main doc, could we add this step in the correct place in the set up flows?

It's unclear where you saw this error in your flow, even from the linked issue

NicolasBonet:
> @NicolasBonet I can't add a review because this issue is locked. BUT
> 
> Rather than adding a note to the main doc, could we add this step in the correct place in the set up flows?
> 
> It's unclear where you saw this error in your flow, even from the linked issue

I have fixed that already, I didn't want to have it in two places, but I do agree it looks more straightforward that way!

grgia:
at what point in the setup flow do you need to do this?

NicolasBonet:
As soon as we run the app @grgia, actually this step was missing and my account wasn't working because of it.

grgia (approved):

m-natarajan:
@NicolasBonet @grgia No QA steps, OK to checkoff this one?

NicolasBonet:
@m-natarajan correct

m-natarajan:
Thanks @NicolasBonet

Issue #89255, opened 2026-04-30
[Due for payment 2026-05-11] [Due for payment 2026-05-07] Spend - Report carrousel disappears on approve


izarutskaya:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.64-23
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @JmillsExpensify 
**Slack conversation** (hyperlinked to channel name): [#expensify_bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1777533024582899)

## Action Performed:
Prerequisites: Workspace with 2 employee

1. Go to https://staging.new.expensify.com/
2. Create several reports 
3. Employee A and employee B submit to manager C expense
4. As manager C, go to the Spend page > Needs approval
5. Approve one report

## Expected Result:
User should be able to navigate to the other reports for approval using report carrousel.

## Actual Result:
Report carrousel disappears on approve page.

## Workaround:
Unknown

## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/1c8cab7e-14c2-41f2-8fd7-9d80a522baa4

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @daledah</details>

MelvinBot:
## Proposal

### What is the root cause of that problem?

[...]

PR #89287, opened 2026-04-30
Revert "Avoid mounting MoneyRequestReportNavigation when unnecessary"

arosiclair:
Reverts Expensify/App#88714

$ https://github.com/Expensify/App/issues/89255


arosiclair:
Bug is not fixed in ad-hoc

Issue #89933, opened 2026-05-07
[Due for payment 2026-06-01] [$250] Android - App crashes after tapping back from trigger a DEW submit failure


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** v9.3.68-1
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://github.com/Expensify/App/pull/89153
**Email or phone of affected tester (no customers):** applausetester+piyozx@applause.expensifail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Pull Request QA execution
**Device used:** Samsung Galaxy A52 / Android 14
**App Component:** Money Requests

## Action Performed:
Preconditions
- Logged in to applausetester+piyozx@applause.expensifail.com

Steps:
1. Toggle on the staging server then clear cache and restart in troubleshoot
2. Navigate to Inbox -> Open the workspace chat which the user is the owner of the workspace
3. Open the draft report in the workspace chat
4. Add a manual expense in the report that just opened
5. Tap submit button -> note this trigger a DEW submit failure
6. Tap back button once or twice

## Expected Result:
User navigated to workspace chat if tapping back button once, or navigated to Inbox if tapping back button twice

## Actual Result:
App crashes after tapping back from trigger a DEW submit failure (sometimes it happens with 1x tap or 2x tap back button)
Device details: Samsung Galaxy A52 - Android 14,

## Workaround:
Unknown

## Platforms:
- [x] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [ ] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

[Bug7146929_1778164043154!dew_crash_log.txt](https://github.com/user-attachments/files/27488375/Bug7146929_1778164043154.dew_crash_log.txt)

https://github.com/user-attachments/assets/efbfcb37-6803-4c7a-a1c2-e55c03437730

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)
[...]

PR #89153, opened 2026-04-29
fix: dedup RBR/Fix badge on child expense report under accessible workspace chat

yuvrajangadsingh:
### Explanation of Change

In the LHN, when a child expense report had a DEW or violation error, the `Fix` badge appeared on both the workspace chat row AND the child expense report row, and the child got force-shown in LHN with an RBR — sorted into the `errorReports` bucket so it visually appeared "pinned".

The fix dedups at the source-of-truth in `reportAttributes.ts`: when the parent workspace chat is accessible to the current user (own policy expense chat or admin), the child's `brickRoadStatus` and `actionBadge` are suppressed in the per-report pass. The parent still gets `ERROR`/`FIX` through the existing propagation loop, so the actionable indicator surfaces on the workspace chat row only.

Per @linhvovan29546's [guidance](https://github.com/Expensify/App/issues/88160#issuecomment-4297242775), this mirrors GBR's access rule (only show on the child when the user can't already see it on the parent).

To avoid breaking propagation (which historically read `brickRoadStatus` to discover children with errors), introduce a separate signal `needsParentChatErrorPropagation` set whenever `getReasonAndReportActionThatHasRedBrickRoad` returns a reason. The propagation loop reads that instead.

Mirror the suppression for two consumers that read `reportErrors` directly rather than `brickRoadStatus`:
- `shouldDisplayReportInLHN` → `hasReportErrorsOtherThanFailedReceipt`: skip child's `allReportErrors` so the row is not force-shown.
- `OptionsListUtils` brickRoadIndicator: same gate so search/share-to pickers don't promote `reportErrors` to a duplicate RBR.

Personal IOU reports (DM-based, parent is not a policy expense chat) keep their own RBR/Fix because the access gate is false.

### Fixed Issues

$ https://github.com/Expensify/App/issues/88160
PROPOSAL: https://github.com/Expensify/App/issues/88160#issuecomment-4263958508

### Tests

1. Sign in as the owner or admin of a workspace that has DEW (Manager Approval workflow) configured
2. Open or create an expense report under that workspace and trigger a DEW submit failure
3. In the LHN, verify the `Fix` badge appears only on the workspace chat row, not on the child expense report row
4. Verify the child expense report is not promoted into the `errorReports` bucket (does not appear "pinned" at the top)
5. Tap the `Fix` badge on the workspace chat row — confirm it deep-links to the violation as before
6. Open the search router (Cmd+K) and confirm the child expense report does not show a duplicate RBR there either
[...]

Issue #88160, opened 2026-04-16
[$250] [LHN] "Fix" action badge incorrectly appears on both workspace chat and individual expense report


MelvinBot:
## Bug: "Fix" action badge appears on both workspace chat and individual expense report in LHN

When there are violations on forwarded receipts (triggering the DEW flow), the `Fix` action badge/pill appears on **both** the workspace chat and the individual expense report in the LHN. It should only appear on the workspace chat, not the individual report.

### Expected behavior
The `Fix` action badge should only show on the workspace chat in the LHN, not on the individual expense report.

### Actual behavior
Both the workspace chat and the individual expense report display the `Fix` icon/pill in the LHN.

Additionally:
- The expense report appears to be pinned in the LHN
- This seems correlated with the DEW error message: _"This report can't be submitted. Please review the comments to resolve."_ — whenever that red text appears in the header, the double `Fix` pills also appear in the LHN

### How to reproduce
No reliable repro steps yet. The reporter observed this after forwarding receipts that resulted in violations. Onyx state was shared in the Slack thread for debugging.

### Related context
- This may be related to the DEW (Daily Expense Workflow) handling of violations
- Possibly related to Expensify/App#86287 (phantom GBR on workspace chat)

**Slack thread:** https://expensify.slack.com/archives/C05LX9D6E07/p1775659495167829

<details><summary>Issue Owner</summary>Current Issue Owner: @yuvrajangadsingh</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022044910272916862953</li>
        <li>Upwork Job ID: 2044910272916862953</li>
        <li>Last Price Increase: 2026-04-23</li>
<li>Automatic offers: </li>
<ul>
<li>linhvovan29546 | Reviewer | 111272259</li>
</ul></li>
    </ul>
</details>

MelvinBot:
## Proposal

### What is the root cause of that problem?

[...]

PR #89936, opened 2026-05-07
fix: avoid Android FlashList crash on DEW back press

yuvrajangadsingh:
### Explanation of Change

PR #89153 (the dedup fix for #88160) introduced a deploy blocker on Android: tapping back after a DEW submit failure crashes the app with `Error: index out of bounds, not enough layouts` (#89933).

The crash trace points at FlashList's layout cache going out of sync during a touch event:

```
getLayout@1:15168580
getLayout@1:15157101
...
dispatchEvent@1:4066331
```

Root cause: PR #89153's `getEffectiveReportErrors` returns `{}` when the child is propagating to an accessible parent. That fed `hasReportErrorsOtherThanFailedReceipt` → `shouldDisplayReportInLHN`, removing the child from the LHN FlashList. On Android, when the user taps back at the moment that data update lands, Fabric/FlashList holds a stale layout index and crashes.

This patch changes only that consumer: `hasReportErrorsOtherThanFailedReceipt` now reads raw `reportAttributes.reportErrors` directly. Child stays force-shown in the LHN, FlashList layout cache stays consistent, no crash on back press.

The badge dedup (the core bounty fix for #88160) is preserved: `brickRoadStatus`/`actionBadge` are still suppressed at the source in `reportAttributes.ts`, so the child appears in LHN with no badge while the parent workspace chat carries the Fix indicator. `OptionsListUtils.createOption` continues to use `getEffectiveReportErrors` for search/share-to picker dedup.

**Tradeoff vs the original bounty fix:** child still appears in LHN (with no badge) instead of being hidden entirely. The full-hide can be tracked as a follow-up once the FlashList layout race is handled safely (likely needs an `InteractionManager.runAfterInteractions` wrap on the data update or an upstream FlashList/Fabric fix).

### Fixed Issues

\$ https://github.com/Expensify/App/issues/89933

### Tests

1. Sign in to a workspace where you're the owner (so `isOwnPolicyExpenseChat` is true) on **Android (HybridApp or mWeb Chrome)**
2. Open the workspace chat
3. Open or create an expense report under that workspace
4. Add a manual expense in the report
5. Tap submit → trigger a DEW submit failure
6. Tap back button (1x or 2x)
7. Verify: app does NOT crash. user navigates back to workspace chat (1x) or Inbox (2x)
8. Verify in LHN: workspace chat row shows the \`Fix\` badge, child expense report row appears without a duplicate \`Fix\` badge
9. Tap the \`Fix\` badge on the workspace chat — confirm it deep-links to the violation
10. Control: open a DM-based personal IOU thread (parent is a DM) — RBR/Fix should still appear there as before

### Offline tests

Same as Tests — derived-value computation runs from in-memory Onyx state.

### QA Steps

Same as Tests, performed against staging on Android with a workspace configured for DEW (Manager Approval workflow + a transaction that triggers a submit failure).

### DIFF
--- a/src/libs/ReportUtils.ts
+++ b/src/libs/ReportUtils.ts
@@ -9474,7 +9474,13 @@ function hasReportErrorsOtherThanFailedReceipt(
     transactions: OnyxCollection<Transaction>,
     reportAttributes?: ReportAttributesDerivedValue['reports'],
 ) {
-    const allReportErrors = getEffectiveReportErrors(reportAttributes?.[report?.reportID]);
+    // Use raw reportErrors here (not getEffectiveReportErrors) so the child stays force-shown in
+    // LHN when it has errors. Removing it from the LHN data while Android Fabric/FlashList holds
+    // a stale layout index during back navigation crashes the app with "index out of bounds, not
+    // enough layouts" (see #89933). brickRoadStatus/actionBadge are still suppressed at the source
+    // in reportAttributes.ts, so the child appears with no badge while the parent workspace chat
+    // carries the Fix indicator.
+    const allReportErrors = reportAttributes?.[report?.reportID]?.reportErrors ?? {};
     const transactionReportActions = getAllReportActions(report.reportID);
     const oneTransactionThreadReportID = getOneTransactionThreadReportID(report, chatReport, transactionReportActions, undefined);
     let doesTransactionThreadReportHasViolations = false;


pecanoro:
@yuvrajangadsingh, could we get a video showing that this was actually fixed?

pecanoro (approved):

linhvovan29546:
@yuvrajangadsingh  I don't think this fix resolves the crash. getEffectiveReportErrors already returns raw reportErrors when brickRoadStatus is set, so the behavior is identical in most cases. Even when it returns {}, the child report still appears in LHN via shouldReportBeInOptionList

pecanoro:
Let's close it for now

PR #90684, opened 2026-05-14
Fix composer hidden by default in narrow layouts

steveahnahn:
Fixes a case where the composer can be hidden in narrow layouts (e.g. report threads opened in a side panel) before `ONYXKEYS.SHOULD_SHOW_COMPOSE_INPUT` is initialized.

Changes:
- Default `shouldShowComposeInput` to `true` in `ReportFooter`
- Add a UI regression test covering the default narrow-layout behavior

Tested:
- `npm test -- tests/ui/ReportFooter.test.tsx`

### DIFF
--- a/src/pages/inbox/report/ReportFooter.tsx
+++ b/src/pages/inbox/report/ReportFooter.tsx
@@ -58,7 +58,9 @@ function ReportFooter() {
     const isReportArchived = useReportIsArchived(report?.reportID);
     const {isCurrentReportLoadedFromOnyx} = useIsReportReadyToDisplay(report, reportIDFromRoute, isReportArchived);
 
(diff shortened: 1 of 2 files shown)


steveahnahn:
After reviewing the issue investigation, this PR doesn't address the reported root cause (thread report  coming from backend/sharedReports). The missing composer here is backend-driven via , not . Closing to avoid noise; happy to reopen if maintainers confirm this path is desired.

steveahnahn:
Clarification (my previous comment got mangled by shell globbing): the issue investigation points to backend-derived `report.permissions` being set to `["read"]` for some thread reports, which makes `canWriteInReport(report)` return false and hides the composer. That is different from the `SHOULD_SHOW_COMPOSE_INPUT` default, so this PR likely would not fix as reported.

Issue #91006, opened 2026-05-18
[$250] Expense - Edited expense shows the original amount instead of 0.00 amount in the table


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.75-6
**Reproducible in staging?:** Yes
**Reproducible in production?:** Yes
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://github.com/Expensify/App/pull/86824
**Email or phone of affected tester (no customers):** sdsdjoijdoisnjodis@gmail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** Mac 26.2 / Chrome
**App Component:** Money Requests

## Action Performed:
1. Go to staging.new.expensify.com
2. Go to workspace chat.
3. Create an expense with amount that is not zero.
4. Open the expense.
5. Click Amount.
6. Edit the amount to 0.00 and save it.
7. Create one more expense in the same report.

## Expected Result:
The edited expense will show 0.00 amount in the table.

## Actual Result:
The edited expense shows the original amount instead of 0.00 amount in the table.

## Workaround:
Unknown

## Platforms:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/658fef77-d32c-4d61-bf5d-59b8082e8c28

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022056448477046924432</li>
        <li>Upwork Job ID: 2056448477046924432</li>
        <li>Last Price Increase: 2026-05-18</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @robertjchen</details>

MelvinBot:
## Proposal

### What is the root cause of that problem?

[...]

PR #91011, opened 2026-05-18
fix: resolve expense /usr/bin/bash.00 display bug bounty #91006

bomlinux92-byte:
## Summary
Fix for [BOUNTY $250] Issue #91006 — Expense edited to $0.00 shows original amount instead of 0.00 in the table

## Root cause
When editing an expense amount to $0.00,  is correctly set to , but:
1.  renders the TOTAL column using  when present, bypassing the edited 
2.  in  sorts by  instead of checking for valid 

Both cases use  logic that treats  as falsy.

## Fix
1. **TransactionItemRowWide.tsx**: Added  check — when a transaction has a valid  (including ), prefer it over stale 
2. **ReportUtils.ts**: Updated  for TOTAL column to use 

$ #91006


Valforte:
Please read the repo guidelines before submitting a PR

Issue #90469, opened 2026-05-13
[Due for payment 2026-07-24] [$250] Migrate YearPickerModal to @react-navigation modal screen


mountiny:
## Context

This is a sub-issue of #53493 (Replace modal screens with modals from `@react-navigation`).

For full background and migration approach, see the [triage analysis comment](https://github.com/Expensify/App/issues/53493#issuecomment-4312826488) and the [sub-issue breakdown comment](https://github.com/Expensify/App/issues/53493#issuecomment-4439641260).

## Problem

`YearPickerModal` animates in from right-to-left like a regular screen, but uses `react-native-modal` internally. This causes inconsistent animations compared to standard `@react-navigation` screens on native.

## Migration approach (Group 2 — Lift local `useState` to Onyx)

- **File:** `src/components/DatePicker/CalendarPicker/YearPickerModal.tsx`
- **Used by:** Only `CalendarPicker/index.tsx`
- **On select:** `onYearChange(option.value)` → updates `currentDateView` state in `CalendarPicker`
- **State to lift:** The `currentDateView` year component from `CalendarPicker` → a shared Onyx key.
- **Note:** `CalendarPicker` is embedded in `DatePickerModal` and `ScheduleCallPage` and owns complex month/year/day state, making this the trickiest in Group 2.

### Steps
1. Lift the selected year state from `CalendarPicker` to an Onyx key.
2. Create a new `@react-navigation` modal route for year selection.
3. Extract `YearPickerModal` content into a standalone screen component that writes the selected year to Onyx.
4. Update `CalendarPicker` to read from Onyx instead of local state for the year.
5. Test all flows that use `DatePicker` (date of birth, IOURequestStepDate, SetDatePage, etc.) on iOS, Android, and web.

## Reference

- Reference PR for this migration pattern: https://github.com/Expensify/App/pull/88915
- Parent tracking issue: #53493
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022054512616348356516</li>
        <li>Upwork Job ID: 2054512616348356516</li>
        <li>Last Price Increase: 2026-05-13</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

nkdengineer:
@mountiny Is the open for contributor? I'd like to work on this if it opens.

yusufdeveloper2903:
## Proposal

[...]

Issue #53493, opened 2024-12-03
Replace modal screens with modals from `@react-navigation`


chrispader:
P/S Slack thread: [expensify.slack.com/archives/C01GTK53T8Q/p1733938583809829](https://expensify.slack.com/archives/C01GTK53T8Q/p1733938583809829)

### Problem

As discussed in [this issue](https://github.com/Expensify/App/pull/53475#issue-2715390810) and this [Slack thread](https://expensify.slack.com/archives/C083CKCMWRE/p1733241544952329?thread_ts=1733228986.361849&cid=C083CKCMWRE) we should replace modal screens (navigation screens/routes, that only render a modal) and modals that stay open when the containing screen gets unmounted.

The current modals use `react-native-modal` which uses RN internal Modal under the hood. Using these modals with `@react-navigation/native-stack` poses issues with animations and navigation, as these modals cannot animate in/out, due to the screen mounting/unmounting at the same time. Additionally, the animations will not look the same as the rest of the screen animations, which doesn't look very nice.

### Solution

Replace the following types of modals with modal (screens) from `@react-navigation`

- screens/routes that only render a modal
- modals on screens, that are navigated back from while the modal is still open or animating out at the same time

cc @mountiny 

---

# List of components and screens to migrate to `@react-navigation` modal screens

Tracking my progress on an exhaustive list of components and screens that can/should be migrated to modal screens.

The list of components/screens (and other files) are triaged/prioritized as follows:
(File names are listed without file ending, usually `.tsx`)

- 🔴 **HIGH**: Should definitely be migrated, due to current issues with animations and navigation.

  These modals are already causing issues with animations, because

  - they are **animating in while a new screen is still mounting**
  - they are **still open/displayed while the old screen unmounting**

  causing no or cut off animations.

- 🟡 **MEDIUM**: Not causing any issues at the moment, but still worth to migrate to modal screens since the behave "like a screen", to achieve consistent animations and generally improve navigation. (gestures, native feel)
 
  This includes modals that are either
  - **animated in from right-to-left** (like a screen) or
  - that are part of another screen with **`animation: none`**,
  - etc.

- 🟢 **LOW**: These screens are **animated in from bottom-to-top** like actual modals and are never (un-)mounted while at the same time as the screen. Migrating these modals to modal screens (and extra navigation routes) would make animations more consistent with the other screens, but it is not necessary and might not be worth the effort.

[...]

PR #91106, opened 2026-05-19
Migrate YearPickerModal to a @react-navigation modal screen

trasnake87:
### Explanation of Change

`YearPickerModal` was a `react-native-modal` (`RIGHT_DOCKED`) rendered inside `CalendarPicker`, driven by local `isYearPickerVisible`/`currentDateView` state. Because `react-native-modal` mounts and animates independently of the native stack, it animated inconsistently with the surrounding `@react-navigation` screens (the problem described in the parent migration Expensify/App#53493).

This migrates the year picker to a dynamic `@react-navigation` modal route, following the pattern established by PR [Expensify/App#88915](<https://github.com/Expensify/App/issues/88915>) (`DynamicExpenseLimitTypeSelectorPage`):

* Adds `DYNAMIC_ROUTES.YEAR_SELECTOR` and a standalone `DynamicYearSelectorPage` screen (the year list + search input, built from route query params), wired through `ROUTES`, `SCREENS`, `ModalStackNavigators`, `linkingConfig`, and navigation `types`.
* Lifts the picker result out of `CalendarPicker`'s local state into a transient Onyx key (`CALENDAR_PICKER_SELECTED_YEAR`), tagged with a per-instance `contextID`. `CalendarPicker` is rendered in multiple places, and `RangeDatePicker` mounts two instances simultaneously, so the result is matched back to the instance that opened the picker via `contextID` and cleared once consumed.
* `CalendarPicker` now navigates to the year route (via `createDynamicRoute`) and applies the returned year to the displayed date. Hosts that render `CalendarPicker` inside a popover (`DatePickerModal`, the Search date dropdown) pass `onBeforeOpenYearPicker` so the popover is dismissed before the year screen is shown, instead of stacking a screen behind a `react-native-modal`.
* Removes `YearPickerModal` and the now-unused year-list state in `CalendarPicker`.

### Fixed Issues

$ [https://github.com/Expensify/App/issues/90469](<https://github.com/Expensify/App/issues/90469>)  <br>PROPOSAL: [https://github.com/Expensify/App/issues/90469#issuecomment-4440082351](<https://github.com/Expensify/App/issues/90469#issuecomment-4440082351>)

### Tests

1. Open Settings → Profile → Date of birth (or any date field, e.g. an expense's date).
2. Tap the date field to open the calendar.
3. Tap the year in the calendar header.
4. Verify the year selector opens as a screen that slides in like other native screens (not a modal that animates independently), with a working search box.
5. Select a year and verify you return to the date flow and the calendar is showing the year you picked.
6. Repeat for a Search date filter (Advanced filters → a date filter → switch to a single date / range) and for Schedule a call, and verify the year selector works in each and updates the correct calendar (for a range, the calendar you opened it from).
7. Verify the month picker still opens and works as before.

### Offline tests

The year selector is client-side navigation and writes only to a transient local Onyx value; behavior is identical offline.

### QA Steps

Same as tests.

### PR Author Checklist

  - [X] I tested this PR with a [High Traffic account](<https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts>) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [X] I included screenshots or videos for tests on [all platforms](<https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms>)
  - [X] Android: Native
  - [X] Android: mWeb Chrome
  - [X] iOS: Native
  - [X] iOS: mWeb Safari
  - [X] MacOS: Chrome / Safari
- [X] I followed proper code patterns (see [Reviewing the code](<https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code>))
  - [X] I verified any copy / text shown in the product is localized by adding it to `src/languages/*` files and using the [translation method](<https://github.com/Expensify/App/blob/4510fc76bbf5df699a2575bfb49a276af90f3ed7/src/components/LocaleContextProvider.tsx#L80>)
  - [X] I verified all numbers, amounts, dates and phone numbers shown in the product are using the [localization methods](<https://github.com/Expensify/App/blob/4510fc76bbf5df699a2575bfb49a276af90f3ed7/src/components/LocaleContextProvider.tsx#L116-L123>)
  - [X] I verified the JSDocs style guidelines (in [`STYLE.md`](<https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md#jsdocs>)) were followed
- [X] I followed the guidelines as stated in the [Review Guidelines](<https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md>)
  - [X] The style can't be created with an existing [StyleUtils](<https://github.com/Expensify/App/blob/main/src/styles/utils/index.ts>) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG)`)
- [X] I added [unit tests](<https://github.com/Expensify/App/blob/main/tests/README.md>) for any new feature or bug fix in this PR to help automatically prevent regressions in this user flow.

### Screenshots/Videos

<details><summary>Android: Native</summary>

https://github.com/user-attachments/assets/10333be9-5ee8-45b5-b2b3-da1df4edfbc6

</details>

<details><summary>Android: mWeb Chrome</summary>

https://github.com/user-attachments/assets/e2ad6edb-6254-45ba-9601-7a090faf78f9

</details>

<details><summary>iOS: Native</summary>

https://github.com/user-attachments/assets/a77db594-b06b-4f2c-86e7-48dc677a6f8d

</details>

<details><summary>iOS: mWeb Safari</summary>

https://github.com/user-attachments/assets/55a245f7-f5c2-4652-b34f-11ebafcb093f

</details>

<details><summary>MacOS: Chrome / Safari</summary>

https://github.com/user-attachments/assets/5c921122-2e02-4da1-b26e-bc98cce460a5

</details>


trasnake87:
Good catch on the entry screens. Rather than extend the allowlist, I switched `YEAR_SELECTOR` to `entryScreens: ['*']`.

Reasoning: `CalendarPicker` is a generic component reachable from a lot of screens — direct date input fields plus `DateSelectPopup`, `RangeDatePicker`, `DatePresetFilterBase`, `ScheduleCallPage`, the company/personal card transaction start date pages, the debug datetime picker, and so on. The previous in-place `YearPickerModal` had no screen restriction, so an explicit allowlist is both a behavior change and inherently fragile — it would silently break the year picker again whenever a new date-input screen is added. The `'*'` wildcard keeps it reachable from every host and matches the existing generic flows (`KEYBOARD_SHORTCUTS`, `EXIT_SURVEY_*`).

Added a regression test in `CalendarPickerTest.tsx` asserting the route stays unrestricted so this can't quietly regress to a partial allowlist. jest and eslint are green (`83b38b6`).

parasharrajat:
Fix the checklist @trasnake87 and add vids for all platforms.

trasnake87:
Added the Android: mWeb Chrome video and ticked the offline-steps box (the section was already in the body, just hadn't been checked). iOS to follow.

src/components/DatePicker/CalendarPicker/index.tsx
```diff
@@ -84,8 +105,10 @@ function CalendarPicker({
     const pressableRef = useRef<View>(null);
     const monthPressableRef = useRef<View>(null);
     const [currentDateView, setCurrentDateView] = useState(() => getInitialCurrentDateView(value, minDate, maxDate));
-    const [isYearPickerVisible, setIsYearPickerVisible] = useState(false);
     const [isMonthPickerVisible, setIsMonthPickerVisible] = useState(false);
+    const [fallbackContextID] = useState(() => `calendarPicker-${Math.random().toString(36).slice(2)}`);
```
parasharrajat: useREf

src/components/DatePicker/CalendarPicker/index.tsx
```diff
@@ -97,27 +120,22 @@ function CalendarPicker({
...
-        requestAnimationFrame(() => setIsYearPickerVisible(false));
+    // When the year picker screen writes back a selection for this CalendarPicker instance,
+    // apply it to the displayed date and clear the transient result so it isn't re-applied.
+    useEffect(() => {
+        if (!selectedYearResult || selectedYearResult.contextID !== contextID) {
+            return;
+        }
+        const {year} = selectedYearResult;
+        clearCalendarPickerSelectedYear();
+        requestAnimationFrame(() => setCurrentDateView((prev) => setYear(new Date(prev), year)));
```
parasharrajat, line 130: why we need `requestAnimationFrame` here?

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
```diff
@@ -467,6 +472,8 @@ function DatePresetFilterBase({
                 onSelected={handleSingleDateSelected}
                 minDate={CONST.CALENDAR_PICKER.MIN_DATE}
                 maxDate={CONST.CALENDAR_PICKER.MAX_DATE}
+                pickerContextID="searchSingleDate"
```
parasharrajat, line 481: I think we can just leave these to `CalendarPicker` auto-calculated values. What is the benefit of passing static values?

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
```diff
@@ -103,6 +103,9 @@ type DatePresetFilterBaseProps = {
     /** Force vertical stacking of calendars in range picker */
     forceVerticalCalendars?: boolean;
 
+    /** Called right before navigating to the year picker screen (used by popover hosts to dismiss themselves) */
+    onBeforeOpenYearPicker?: () => void;
```
parasharrajat: Maybe instead of this, you can call `Modal.closeTop` from modal action file.

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: Good call on the unused `useState` setter. I switched the fallback identifier to `useId()` rather than `useRef`: `contextID` is read during render, and the repo's React Compiler lint rules reject reading `ref.current` during render. `useId()` gives a stable per-instance id with no unused setter, so it's the cleaner fit here. (426a1008)

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: The effect reacts to the year that the year picker screen writes into Onyx. Calling `setCurrentDateView` synchronously in the effect body trips `react-hooks/set-state-in-effect` (the cascading-render rule), so it's deferred one frame. I added a comment in 426a1008 making that explicit.

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
trasnake87: `fallbackContextID` is regenerated each time a CalendarPicker mounts. This host's popover is dismissed when the year picker opens, so the CalendarPicker unmounts and remounts fresh when it's reopened — a regenerated fallback wouldn't match the `contextID` the year picker tagged its result with in Onyx, and the selection would be dropped. A stable static id keeps that round-trip intact (and the range picker needs two distinct ids so its calendars don't pick up each other's result).

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
trasnake87: Done in 426a1008 — `CalendarPicker` now calls `Modal.closeTop()` directly instead of taking a close callback from each host. I gated it behind a boolean prop (`shouldCloseModalOnYearPickerOpen`): the wide-screen popover hosts set it, but the full-screen mobile `DateSelectPopup` overlay deliberately stays mounted while the year picker is open, so the dismissal needs to stay opt-in rather than fire unconditionally from `CalendarPicker`.

trasnake87:
@parasharrajat I've added the iOS native recording to the **Screenshots/Videos** section. The flow shows opening **Search → Date filter → Custom date** and tapping the year label, which now navigates to the migrated `@react-navigation` Year screen (full-screen page with its own header + Back, replacing the old `react-native-modal`). Selecting 2024 returns to the calendar with the new year applied.

Web/Android boxes were already ticked with their respective recordings; iOS row is now ticked too.

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
parasharrajat: Then we should remove the fallback ID as it is prone to this bug. we should make sure the contextID is provided from parent.

src/ROUTES.ts
```diff
@@ -108,6 +108,16 @@ const DYNAMIC_ROUTES = {
...
     },
+    YEAR_SELECTOR: {
+        path: 'year-selector',
+        queryParams: ['contextID', 'currentYear', 'minYear', 'maxYear'],
+        // CalendarPicker is a generic component reached from many screens (date input fields,
+        // DateSelectPopup, RangeDatePicker, DatePresetFilterBase, ScheduleCallPage, ...), and the
+        // previous in-place YearPickerModal had no screen restriction. Use '*' so the year selector
+        // remains reachable from every CalendarPicker host and doesn't silently break when new
+        // date-input screens are added (matches KEYBOARD_SHORTCUTS / EXIT_SURVEY_* generic flows).
+        entryScreens: ['*'],
```
parasharrajat, line 171: Add a getRoute handler here to add queryParams.

src/components/DatePicker/CalendarPicker/index.tsx
```diff
@@ -97,27 +121,29 @@ function CalendarPicker({
...
+    }, [selectedYearResult, contextID]);
+
+    const openYearPicker = () => {
+        // Dismiss the popover/modal hosting this CalendarPicker (if any) so the year picker
+        // screen is not rendered behind it.
+        if (shouldCloseModalOnYearPickerOpen) {
+            closeTop();
+        }
+        Navigation.navigate(
+            createDynamicRoute(`${DYNAMIC_ROUTES.YEAR_SELECTOR.path}?contextID=${encodeURIComponent(contextID)}&currentYear=${currentYearView}&minYear=${minYear}&maxYear=${maxYear}`),
```
parasharrajat: Use the getRoute handler here to create the route path

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
trasnake87: Done in cd942200 — `pickerContextID` is now required on `CalendarPicker`, the `useId()` fallback and the `pickerContextID ?? fallbackContextID` coalescence are gone, and `useEffect` / `openYearPicker` reference `pickerContextID` directly. All five product call sites (`ScheduleCallPage`, `DatePresetFilterBase`, `RangeDatePicker` × 2, `DatePickerModal`) were already passing stable ids, so this just tightens the contract. Added a small `CalendarPickerForTest` wrapper in `tests/unit/CalendarPickerTest.tsx` so existing tests that don't exercise the contextID still satisfy the required prop.

src/ROUTES.ts
trasnake87: Done in cd942200 — added a `getRoute` on `YEAR_SELECTOR` that takes `{contextID, currentYear, minYear, maxYear}` and goes through `getUrlWithParams('year-selector', …)` so encoding stays in one place and matches the declared `queryParams`.

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: Done in cd942200 — `openYearPicker` now calls `DYNAMIC_ROUTES.YEAR_SELECTOR.getRoute({contextID: pickerContextID, currentYear: currentYearView, minYear, maxYear})` and passes that into `createDynamicRoute`. Existing `CalendarPickerTest` assertions (`year-selector`, `contextID=…`, `currentYear=…`) still pass.

src/components/Search/FilterComponents/DatePresetFilterBase.tsx
trasnake87: Done in cd942200 — `pickerContextID` is now a required prop on `CalendarPicker` (no internal `useId` fallback), so every host has to provide a stable ID explicitly. The `DatePresetFilterBase` call site still passes `pickerContextID="searchSingleDate"` since that is now the explicit parent-provided ID.

src/ROUTES.ts
trasnake87: Done in cd942200 — `YEAR_SELECTOR` now has a typed `getRoute({contextID, currentYear, minYear, maxYear})` handler that builds the query params via `getUrlWithParams`.

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: Done in cd942200 — this site now uses `DYNAMIC_ROUTES.YEAR_SELECTOR.getRoute({contextID: pickerContextID, currentYear: currentYearView, minYear, maxYear})` instead of the manual string concatenation.

parasharrajat:
BUG: I can see the pages in stack remounting when we click year. 

1. Send a expense to any workspace. 
2. open that expense. 
3. Edit the date. 
4. Select the year. 


https://github.com/user-attachments/assets/0d508906-ddb9-4e88-b57f-3872c40a7581

parasharrajat:
I also have a thought that we are closing the date picker before opening the year modal. Maybe we shouldn't close it, and overlay the RHP over it. I am not sure how complicated it would be.


Can you open up a Slack discussion to confirm what should be done here before we move forward?


Current behaviour can be buggy.

parasharrajat:
BUG: Date picker closes on year selection and does not open back on spends page. 

Steps: 
1. Go to spends page. 
2. add a date filter. 
3. Open the date filter pill and try to add custom date and custom range. 



https://github.com/user-attachments/assets/d194855c-3c38-41fc-a27e-0bf228088105

trasnake87:
@parasharrajat Thanks for catching these — both are real regressions of the migration. Root cause for each:

**(1) Pages remounting on year tap from the edit-expense date flow.** `IOURequestStepDate` is in `MoneyRequestModalStackNavigator`; the new `DYNAMIC_YEAR_SELECTOR` is registered only in `SettingsModalStackNavigator`. Navigating between two different modal stack navigators forces React Navigation to unmount the previous stack — that's the visible remount when the year is tapped.

**(2) Date picker doesn't reopen on the spends page after year selection.** `DateSelectPopup` is dismissed (`closeTop()`) before the year route is pushed, and there's no caller-side logic to reopen it on `goBack`. `CalendarPicker` unmounts with the popover, so the year written to Onyx by `DynamicYearSelectorPage` is never consumed.

**(3) Overlay-the-RHP direction.** For that to behave cleanly the year route needs to live in the same modal stack as the host (so the host doesn't unmount), AND the popover needs to stay open underneath (so `CalendarPicker` is still mounted to apply the year on return). Two concrete shapes I see:

a) **Register `DYNAMIC_YEAR_SELECTOR` in every modal stack that hosts a `CalendarPicker`** (MoneyRequest, Settings, Search/Reports, etc.) and drop `shouldCloseModalOnYearPickerOpen` so the popover stays open under the in-stack RHP. The popover already passes `shouldDisplayBelowModals`; we'd want to verify the RHP-over-popover layering still holds on each platform.

b) **Scope-down: keep `YearPickerModal` (stacked `@components/Modal`) for popover-hosted callers** (`DatePickerModal`, `RangeDatePicker`/`DatePresetFilterBase`) and use the new `DynamicYearSelectorPage` only for non-popover hosts (DOB, ScheduleCall, Set Date, etc.). Smaller blast radius but it's a partial migration of the original issue's scope.

Happy to prototype either — (a) is closer to the overlay direction you suggested; (b) is the safer scope-down. Let me know which you'd prefer before I push, or if there's a third direction I'm missing.

parasharrajat:
I don't think we should add a dynamic route to all stack navigators. Can you open a discussion on Slack for this problem?

These AI suggestions are fine. What is your opinion on this? Have you verified the root cause?

trasnake87:
Agreed — I don't think we should register the year route in every stack either; that'd pollute all the navigators.

I dug into both regressions and they're really the same root cause: opening the year selector tears down the date picker that launched it. In the edit-expense flow, the year screen only lives in the Settings stack, so tapping the year swaps stacks and unmounts the host (`IOURequestStepDate`). In the search/popover flow, we call `closeTop()` before pushing the year route, so the `CalendarPicker` unmounts before it can read back the selected year — which is why it doesn't reopen.

So I'm with your instinct here: rather than closing the date picker, keep it mounted and present the year selector over it. That removes the unmount behind both bugs, and it lines up with how the old `react-native-modal` worked (the year picker just overlaid the calendar).

I'm getting set up on #expensify-open-source now and I'll start a thread there to settle the cleanest way to present it as an overlay without the cross-stack teardown.

trasnake87:
Thanks @parasharrajat! I filled out the form to get into #expensify-open-source — just waiting on the invite to come through, and I'll start the thread the second I'm added. Anything else here you'd like me to pick up while I wait on access?

flaviadefaria (dismissed):
From a product standpoint, the fix looks good to me.

trasnake87:
Pushed the approach we settled on in Slack — on web wide-screen I keep the date picker mounted and just hide it (it's a z-index 9996 portal painting over the RHP) while the year selector is open, then restore it on return; narrow screens and native keep the existing dismiss. Ready for another look when you get a chance.

trasnake87:
@parasharrajat this is green and ready for review now. The overlay approach we landed on with yehor is implemented — on web wide-screen the date popover is a z-index 9996 portal that was painting over the migrated year-selector RHP, so instead of dismissing it I keep it mounted and hide it (opacity + non-interactive) while the year selector is open, then restore on return; narrow screens and native keep the existing dismiss. I also fixed a CalendarPicker unit test that needed its mock updated for the new navigation-state read. Whenever you have a moment.

trasnake87:
@parasharrajat whenever you get a chance — this is green and implements the overlay approach we aligned on (and yehor signed off on). Ready for your re-review, and happy to address anything that comes up.

parasharrajat:
Thanks. I will check.

src/components/DatePicker/CalendarPicker/index.tsx
```diff
@@ -239,7 +250,7 @@ function CalendarPicker({
     const getAccessibilityState = useCallback((isSelected: boolean) => ({selected: isSelected}), []);
 
     return (
-        <View style={[themeStyles.pb4, themeStyles.pt1]}>
+        <View style={[themeStyles.pb4, themeStyles.pt1, isHiddenForYearPicker && themeStyles.opacity0, isHiddenForYearPicker && themeStyles.pointerEventsNone]}>
```
parasharrajat: Let's use visibility hidden as well.

src/components/DatePicker/DatePickerModal.tsx
```diff
@@ -46,6 +47,7 @@ function DatePickerModal({
     // We need to use isSmallScreenWidth instead of shouldUseNarrowLayout to distinguish RHL and narrow layout
     // eslint-disable-next-line rulesdir/prefer-shouldUseNarrowLayout-instead-of-isSmallScreenWidth
     const {isSmallScreenWidth} = useResponsiveLayout();
+    const shouldHideInPlace = Platform.OS === 'web' && !isSmallScreenWidth;
```
parasharrajat: Use Platform.OS is not not recommended directly in our repo.

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: Done — added `themeStyles.visibilityHidden` alongside the opacity/pointer-events hide.

src/components/DatePicker/DatePickerModal.tsx
trasnake87: Done — switched to `getPlatform() === CONST.PLATFORM.WEB`. Behaviorally identical here since getPlatform's web impl returns `WEB` for desktop too (electron uses the web bundle), so no platform-coverage change.

parasharrajat:
@trasnake87 Please test your changes across the app before requesting another review.

1. First, only the calendar gets hidden. 
2. Secondly, when I click on the year, the picker closes. 

image



https://github.com/user-attachments/assets/1ffa16c0-2b4e-4275-9a4a-5372bc0afdc4

parasharrajat:
BUG: Personal details. 


https://github.com/user-attachments/assets/a196063f-d82b-4653-863b-9fd94cab3268

trasnake87:
Reworked this to be comprehensive across every wide-screen date-picker host. Instead of hiding only the inner `CalendarPicker` (which left the popover frame painted under the year-selector RHP), each host now hides its whole popover frame while the year-selector route is focused, and suppresses the `goBack` popstate-close. This is centralized via a small `useIsYearSelectorOpen` hook and applied at `DatePickerModal` (DOB + form date inputs) and `DropdownButton` (the shared owner of the search filter popover — single/range/presets). The old per-calendar `shouldHideOnYearPickerOpen` mechanism is removed entirely. The selected year still survives the host unmount via the existing `pickerContextID`/Onyx write-back. tsc/eslint/prettier green, CalendarPickerTest 35/35.

src/components/Search/FilterDropdowns/DropdownButton.tsx
```diff
@@ -213,7 +219,9 @@ function DropdownButton({
             <PopoverWithMeasuredContent
                 anchorRef={triggerRef}
                 avoidKeyboard
-                isVisible={isOverlayVisible}
+                // Hide the whole popover frame (not just the inner CalendarPicker) while the year-selector RHP is focused,
```
parasharrajat: why do we need this change?

trasnake87:
@parasharrajat — quick flag that the comprehensive rework is up and all checks are green. Each wide-screen host now hides its whole popover frame while the year-selector is focused (not just the inner calendar), so both issues you caught — "only the calendar hides" and the picker closing when you tap the year — are addressed, across the search filters and the DOB/personal-details flow alike. Whenever you have a moment to re-review, happy to walk through any part.

src/components/Search/FilterDropdowns/DropdownButton.tsx
trasnake87: The search **Date** filter is hosted by this `DropdownButton` — it isn't in `FILTER_COMPONENT_MAP`, so `SearchFilterBar.tsx:113` falls through to the default `DropdownButton`, and the date popover (`DateSelectPopup → DatePresetFilterBase → RangeDatePicker → CalendarPicker`) is its `PopoverComponent`. We keep that popover mounted across the year-selector `goBack` (`shouldCloseWhenBrowserNavigationChanged={false}`, L233) so the selection round-trips back via `pickerContextID`/Onyx. But on web the `CalendarPicker` renders as a portal at z-index 9996, so a still-visible popover paints over the year-selector RHP and you can't pick a year.

So we gate `isVisible` on `!(shouldHideInPlace && isYearSelectorOpen)` to hide the whole popover frame while that route is focused, then restore it on `goBack`. It's the same host-frame-hide as `DatePickerModal.tsx` (L79) — DropdownButton just owns the search-filter popover chain.

src/components/Search/FilterDropdowns/DropdownButton.tsx
```diff
@@ -84,6 +86,10 @@ function DropdownButton({
     // We need to use isSmallScreenWidth instead of shouldUseNarrowLayout to distinguish RHL and narrow layout
     // eslint-disable-next-line rulesdir/prefer-shouldUseNarrowLayout-instead-of-isSmallScreenWidth
     const {isSmallScreenWidth} = useResponsiveLayout();
+    // While the year-selector route is focused on web wide-screen, hide the whole dropdown popover frame so the
+    // year-selector RHP isn't painted over the still-visible popover (e.g. the search date filter's CalendarPicker).
+    const shouldHideInPlace = getPlatform() === CONST.PLATFORM.WEB && !isSmallScreenWidth;
+    const isYearSelectorOpen = useIsYearSelectorOpen();
```
parasharrajat: I don't think we should pollute this generic component of this case. You should composer a new component from this if you need a specific one and render that component where needed.

src/components/Search/FilterDropdowns/DateSelectPopup/index.tsx
```diff
@@ -44,6 +44,7 @@ type DateSelectPopupProps = {
 function DateSelectPopup({label, value, presets, style, closeOverlay, onChange, setPopoverWidth}: DateSelectPopupProps) {
     // eslint-disable-next-line rulesdir/prefer-shouldUseNarrowLayout-instead-of-isSmallScreenWidth
     const {isSmallScreenWidth, isInLandscapeMode} = useResponsiveLayout();
+    const shouldHideInPlace = Platform.OS === 'web' && !isSmallScreenWidth;
```
parasharrajat: No direct use of `Platform.OS`. You can use getPlatform instead or create a lib where value is controlled via platform files such as (index.web.ts) etc.

src/components/DatePicker/DatePickerModal.tsx
```diff
@@ -46,6 +48,8 @@ function DatePickerModal({
     // We need to use isSmallScreenWidth instead of shouldUseNarrowLayout to distinguish RHL and narrow layout
     // eslint-disable-next-line rulesdir/prefer-shouldUseNarrowLayout-instead-of-isSmallScreenWidth
     const {isSmallScreenWidth} = useResponsiveLayout();
+    const shouldHideInPlace = getPlatform() === CONST.PLATFORM.WEB && !isSmallScreenWidth;
```
parasharrajat:
```suggestion
    const isDesktopWeb = getPlatform() === CONST.PLATFORM.WEB && !isSmallScreenWidth;
```

Update all places.

parasharrajat: Because the value does not indicate `hideInPlace` logic.

parasharrajat:
BUG: date filter reset back to main menu. 


https://github.com/user-attachments/assets/c3328af4-3290-4fd2-939f-c8f4a517c4af

parasharrajat:
BUG: The Picker closes abruptly on selecting a year and then shows up again. 

Expected: It was closed so it should only be shown. 


https://github.com/user-attachments/assets/30f62335-62a2-475b-85e1-ca1e3760ab86

trasnake87:
Thanks @parasharrajat 🙏 I rebased on `main`, and that actually resolves your first point for us.

**On the `DropdownButton` change (your "don't pollute the generic component" comment):** `main` has since refactored the search dropdowns — all the popover logic now lives in the new `FilterPopupButton`, which wraps the popover in an `{isFocused && …}` gate. That gate already unmounts the search-filter popover (and its `CalendarPicker`) when the year-selector RHP takes focus, so the overpaint no longer happens on the search path. I've therefore **dropped my `DropdownButton` change entirely and taken `main`'s version untouched** — no wrapper / no pollution of the generic component.

I verified this on a merged build: opening the year selector from the search Date filter removes the calendar from the DOM and the RHP renders cleanly (no overpaint).

**The `DatePickerModal` host-hide is still required**, though — that host (DOB / standalone date pickers) wasn't part of the `FilterPopupButton` refactor and has no `isFocused` gate, so without the hide the `CalendarPicker` still paints over the year-selector RHP. A negative-control test (removing the hide) confirmed the overpaint returns there.

I also applied your other two suggestions:
- `DateSelectPopup`: `Platform.OS === 'web'` → `getPlatform() === CONST.PLATFORM.WEB`
- renamed `shouldHideInPlace` → `isDesktopWeb` (updated all usages)

Happy to share the before/after recordings if useful.

trasnake87:
@parasharrajat thanks for catching both of these — you're right, they reproduce, and I traced the root cause on the rebased build:

The search Filters popover (the advanced **Filters** menu, and the Date sub-view) is hosted by the new `FilterPopupButton`, which gates its popover on `useIsFocused()`. When the year-selector screen takes focus, the Search screen blurs → the whole popover **unmounts**, and on `goBack` it **remounts fresh** — which is exactly why it resets to the top filter menu (your first comment) and visibly closes/re-opens (your second).

I'm putting up a fix that preserves the popover's view state (the selected filter + the Custom-date sub-view) across the year-selector round-trip, so you land back on the calendar with the picked year applied. I'll push shortly and tag you to re-test. Appreciate the thorough review 🙏

trasnake87:
Pushed the fix (`b1d1f4ce`) 👍

Both came from the popover host (`FilterPopupButton`) unmounting when the year-selector screen takes focus — on `goBack` it remounted fresh, resetting the advanced **Filters** menu to the default **Type** filter and the Custom-date sub-view to its top menu. The fix preserves that view state across the round-trip (persisted in Onyx alongside the existing year write-back, restored only when returning from a search calendar):

- The advanced Filters menu now reopens to the **Date** filter on return (instead of Type), with the Custom-date/range sub-view and your picked year restored.
- Same persistence for the inline Date filter chip.

I verified on a merged build: opening the year selector from the Date filter → picking a year → back now lands directly on the calendar with the selected year applied (no longer the main menu).

One honest note on your 2nd comment (the brief close/re-open): the popover genuinely unmounts while the year-selector screen is focused — that's what stops the calendar's modal frame from painting over the year-selector RHP, so it's a deliberate trade-off rather than a glitch. It now reopens to exactly the right place. If you'd rather we also eliminate that transition, we'd need to keep it mounted-but-hidden, which reintroduces the overpaint risk — happy to discuss the trade-off. Ready for re-test 🙏

trasnake87:
Update on the year-selector approach (cc @parasharrajat @yehor) — pushed at `d62b9db`.

I implemented the keep-mounted + `CalendarPicker` opacity/pointerEvents approach we discussed and verified it **live on both date-picker entry points**:

**✅ DatePickerModal (DOB / standalone date picker)** — works cleanly. The popover stays mounted and the `CalendarPicker` hides itself (`opacity:0` + `pointerEvents:none`) on wide-screen web, so its z-index-9996 portal no longer paints over the year-selector RHP, the context is preserved, and there's no close/re-open flicker. Narrow/native still dismiss via `closeTop()` + the Onyx handoff.

**⚠️ Search Advanced Filters date filter** — keep-mounted didn't transfer cleanly. That popover isn't just a calendar; it's the whole filter menu (the Type/Status/Date… list + the date sub-view). Keeping it mounted leaves the *menu* painting over the RHP — only the `CalendarPicker` can self-hide, not the surrounding menu — and the menu also remounts back to the top Type filter on return.

So for the search path I used the other option you mentioned, @yehor — **close it and pass the context via Onyx** — which restores the user to the Custom-date calendar with the picked year on return. Trade-off: that path still has a brief close/re-open, since the popover must unmount to keep the filter menu off the RHP.

**Net:** DOB → keep-mounted + opacity; search filter → close + Onyx context. Both are within the two options you laid out. If you'd prefer to handle the search filter menu differently (e.g. a way to hide the whole popover frame without unmounting), I'm happy to try it — otherwise this is ready for re-test 🙏

src/components/Search/FilterDropdowns/DropdownButton.tsx
trasnake87: Dropped this change entirely 👍 — after rebasing on `main`, `DropdownButton` is now untouched. `main` extracted the popover into the new `FilterPopupButton`, which already gates rendering on `isFocused`, so there's no longer anything to compose out of the generic component here.

src/components/Search/FilterDropdowns/DateSelectPopup/index.tsx
trasnake87: Done — switched to `getPlatform() === CONST.PLATFORM.WEB` (no direct `Platform.OS`).

src/components/DatePicker/DatePickerModal.tsx
trasnake87: Done — renamed `shouldHideInPlace` → `isDesktopWeb` across all usages (DatePickerModal + DateSelectPopup).

parasharrajat:
@trasnake87, you tagged the wrong user, I think. Please go back to the Slack thread if you have any blockers. Even I do not have a solution for this problem.

I feel like hiding and trying to show back pickers is a nasty workaround. It might do anything in the future.

trasnake87:
You're right, sorry about that — the `@yehor` I tagged is the wrong account (an empty profile, not the person I meant). The yehor I had in mind is the navigation engineer from the Slack thread, and I'll keep that side of it in Slack rather than pinging strangers here.

I hear you on the hide-and-restore being a nasty workaround — I share the worry that it's fragile and could bite us later. For context, it came out of the Slack diagnosis that `CalendarPicker` renders as a z-index-9996 web portal that paints over the year-selector RHP, where the two options on the table were (1) keep it mounted but hidden, or (2) close it and pass context back via Onyx (what the current hybrid does). Neither feels clean enough to call done.

Since you don't have a solution in mind either, let me take it back to the Slack thread and pull the nav folks in to land on something more robust than hide/restore — I'll bring whatever we settle on back here. Thanks for the patience on this 🙏

trasnake87:
Update @parasharrajat — I took this back to the Slack thread and yehor (the nav engineer) signed off on the approach 👍. The crux: `CalendarPicker` renders through `ReactDOM.createPortal` at `modalLowestZIndex` (9996), so it paints outside the nav stacking context — there's no way to put the year-selector RHP above it without raising the whole RHP layer over 9996, which would then cover other modals. So hide-and-restore is the right tool here, and the `opacity:0`/`pointerEvents:none` hide is a pure visual state (no timing, fully reversible), not a fragile workaround.

I'm rebasing on the latest `main` now (it re-architected the search date filters again, so I'm re-mapping that part) and will push a clean update. Does yehor's sign-off settle the approach for you, or is there anything else you'd want before re-review? 🙏

parasharrajat:
I am fine with hiding and showing, but we need to fix the UI bugs and glitches. That is not acceptable.

Also, you will have to extend the test cases to the whole app, which should cover all instances of date pickers, as every date picker is impacted independently.

trasnake87:
@parasharrajat — pushed a comprehensive update and rebased on `main` (CI green).

**Root cause of the un-clickable year + the glitches:** the date popovers are react-native-web `Modal`s whose full-screen portal root (`ModalAnimation` — `position:fixed`, `z-index:9996`, `pointer-events:auto`) sat over the year-selector RHP and swallowed the clicks. RNW only forwards `zIndex` to that root (no prop for its pointer-events), and dropping `coverScreen` to avoid it remounts the popover — which is what caused the reset-to-Type and the close-then-reshow flicker. **Fix:** a `shouldDisablePointerEvents` prop threaded `BaseModal → ReanimatedModal` that makes the kept-mounted modal subtree pointer-transparent while the year selector is open, so the popover stays mounted (state preserved, no flicker) and the RHP receives the clicks. Same reusable prop for the search filters and `DatePickerModal` — a ready-made pattern for future modal→dynamic-route migrations (the direction yehor suggested in Slack).

**UI bugs fixed (verified on web):**
- **Personal details / DOB:** the calendar now stays open showing the picked year. It was closing because the year-selector `goBack` fires a popstate that reaches the close listener *after* `isYearSelectorOpen` has flipped back to false (a race the existing gating couldn't catch); a short grace window absorbs it.
- **Search AdvancedFilters:** stays on **Date** (no reset to the Type menu) and applies the picked year; no close-then-reshow.

**Inline comments addressed:** `visibility:hidden` added alongside `opacity:0`; `isDesktopWeb` named everywhere; (DropdownButton untouched, `getPlatform` not `Platform.OS`, fallback contextID removed, route built via a `getRoute` handler — all in).

**Tests (cover-all-instances):** year-selector round-trip coverage across the date pickers — CalendarPicker (year applies without resetting the month/day view, contextID isolation, desktop-web hide on/off), RangeDatePicker (the two From/To pickers don't collide on the shared Onyx key), search AdvancedFilters (returns to the Date view, not the Type menu), ScheduleCall, and DatePickerModal (the DOB keep-open guard + frame-hide). All green.

Ready for another look 🙏

trasnake87:
@parasharrajat — gentle bump on this one. The 06-20 update addressed the UI glitches you flagged (root cause + fix are in that comment), the review threads are all marked done, and CI is green. Whenever you get a chance to take another look — thanks! 🙏

parasharrajat:
@trasnake87 Can you please merge main?

trasnake87:
@parasharrajat merged latest `main` and pushed the branch. The only merge conflict was in `SearchAdvancedFiltersPopup`; I kept the year-selector hide logic and dropped an unused stale import.\n\nLocal checks:\n- `npm run typecheck`\n- `npx eslint src/components/Search/FilterDropdowns/SearchAdvancedFiltersPopup/index.tsx`\n- `npx prettier --check src/components/Search/FilterDropdowns/SearchAdvancedFiltersPopup/index.tsx`\n\nReady for re-review. Thanks!

src/ONYXKEYS.ts
```diff
@@ -622,6 +622,12 @@ const ONYXKEYS = {
     /** Stores the role selected for members being imported from a spreadsheet */
     IMPORTED_SPREADSHEET_MEMBER_ROLE: 'importedSpreadsheetMemberRole',
 
+    /** Stores the year selected in the year picker so it can be read back by the CalendarPicker that opened it */
+    CALENDAR_PICKER_SELECTED_YEAR: 'calendarPickerSelectedYear',
+
+    /** Stores the active Search date-filter sub-view (Custom date/range modifier) so the date filter can restore it after the year picker screen unmounts and remounts it */
+    CALENDAR_PICKER_SELECTED_DATE_MODIFIER: 'calendarPickerSelectedDateModifier',
```
parasharrajat, line 661: Can we use one key for both as we need the other for year selector behaviour?

trasnake87: Good question — I did look at folding the modifier into `CALENDAR_PICKER_SELECTED_YEAR`, but they shouldn't share one key. They differ in scope and in how they're written:

- **Scope / dead field.** `SELECTED_YEAR` is the generic, `contextID`-routed key that *every* CalendarPicker uses — `datePicker-${inputID}` (e.g. DOB), `scheduleCall`, and the three search pickers (`searchSingleDate`/`searchRangeFrom`/`searchRangeTo`). `SELECTED_DATE_MODIFIER` is Search-only and carries no `contextID`. Merging would push a Search-only field onto the shared key — dead weight on every non-search picker, and a Search concern leaking into the generic CalendarPicker layer.

- **They're written at different times, never together.** The modifier is written the moment the user opens a Custom-date/range sub-view — *before* the year picker opens, and on flows where it's never opened (DOB/ScheduleCall write a year with no modifier at all). So there's no single point where both values are known, i.e. no natural shared write site.

- **Merging forces `Onyx.merge` on what's a clean `Onyx.set`.** `setCalendarPickerSelectedYear` does `Onyx.set(SELECTED_YEAR, {contextID, year})`, replacing the whole value. Since the modifier is written first, a merged object would clobber it on the year write — so both writers would have to move to `Onyx.merge` just to coexist. More moving parts on a key shared by five hosts, for no benefit.

So I've kept them separate, but tightened things based on your comment (pushed in 392638d):
- Added `clearCalendarPickerSelectedDateModifier` and clear the breadcrumb when the user leaves the sub-view (back/reset/apply), so it doesn't linger in storage. The year-picker round-trip doesn't touch those handlers, so restore is unaffected.
- Expanded the ONYXKEYS doc comments to make the pairing explicit — `SELECTED_YEAR` is the generic transient year-routing command; `SELECTED_DATE_MODIFIER` is its Search-only companion, only honoured while a `search*` year write-back is pending.

Happy to revisit if you'd still prefer a single key.

flaviadefaria:
No new product considerations - removing my assignment and unsubscribing.

trasnake87:
@parasharrajat — gentle bump 🙏 I answered the one-key question above (kept the two keys separate, with the reasoning) and added a clear-on-exit so the modifier doesn't linger in storage — pushed in `392638d`. `flaviadefaria` has since cleared the product-review flag too. Whenever you get a chance to take another look, thank you!

parasharrajat:
Thanks. I will check this.

src/components/DatePicker/CalendarPicker/index.tsx
```diff
@@ -243,7 +254,10 @@ function CalendarPicker({
     const getAccessibilityState = useCallback((isSelected: boolean) => ({selected: isSelected}), []);
 
     return (
-        <View style={[themeStyles.pb4, themeStyles.pt1, containerStyle]}>
+        <View
+            style={[themeStyles.pb4, themeStyles.pt1, containerStyle, shouldHideForYearSelector && {opacity: 0, visibility: 'hidden'}]}
```
parasharrajat: No inline rules. Please use styles.

parasharrajat: No inline rules. Please use styles.

src/components/DatePicker/DatePickerModal.tsx
```diff
@@ -101,12 +108,21 @@ function DatePickerModal({
...
-            shouldCloseWhenBrowserNavigationChanged={shouldCloseWhenBrowserNavigationChanged}
-            innerContainerStyle={isSmallScreenWidth ? styles.w100 : {width: CONST.POPOVER_DATE_WIDTH}}
+            // Suppress the popstate close while the year selector is open; selecting a year does a goBack (history
+            // change) and would otherwise tear this host down instead of returning to it with the new year applied.
+            shouldCloseWhenBrowserNavigationChanged={shouldCloseWhenBrowserNavigationChanged && !isYearSelectorOpen}
+            hasBackdrop={!shouldHideForYearSelector}
+            shouldDisablePointerEvents={shouldHideForYearSelector}
+            innerContainerStyle={{
+                ...(isSmallScreenWidth ? styles.w100 : {width: CONST.POPOVER_DATE_WIDTH}),
+                ...(shouldHideForYearSelector ? {opacity: 0, visibility: 'hidden', pointerEvents: 'none'} : {}),
```
parasharrajat: same

src/components/Modal/ReanimatedModal/index.tsx
```diff
@@ -182,6 +190,37 @@ function ReanimatedModal({
...
     }, [style]);
 
+    // react-native-web's <Modal> renders its own full-screen portal root (ModalAnimation: position:fixed, inset 0,
+    // z-index 9996, pointer-events:auto) plus a full-screen ModalContent View. RNW only forwards `zIndex` to that
+    // root, so there is no prop to make it pointer-transparent. When `shouldDisablePointerEvents` is set, a route/RHP
+    // (the dynamic year selector) is intentionally shown over this kept-mounted popover and must receive clicks, but
+    // those full-screen wrappers would otherwise swallow them. We keep the popover mounted (so its state survives the
+    // round-trip — no remount/reset) and instead toggle pointer-events on the portal root imperatively. The popover
+    // content is already visually hidden by the caller, so disabling pointer events on the whole subtree is safe.
+    useEffect(() => {
```
parasharrajat, line 209: Why do we need this? We should already have access to the modal element. I don't like this solution we should do it in react way.

parasharrajat:
I think we should break this PR into sub-PRs for each type of component/popup we are handling. We are doing a lot of workarounds to keep the popup state just because the calendar picker shows above the Modal.


I think using a Context provider would be cleaner here.

trasnake87:
@parasharrajat Thanks for the thorough pass 🙏 — two things:

**Context provider: agreed, and I'm on it now.** You're right that it's cleaner. I'm refactoring so the year-selector coordination flows through a provider instead of prop-threading plus the imperative portal-root workaround you flagged — hosts consume the context and react to it the React way, and the flagged inline styles move into the stylesheet as part of the same pass. I'll push the refactor on this branch.

**On splitting into sub-PRs — one concern I'd like your take on: atomicity.** The core of the PR is the migration itself (`YearPickerModal` is deleted and replaced by the `@react-navigation` route), and that switch changes behavior for *every* date picker at once — DOB, ScheduleCall, and the three search pickers. If we split per host, each intermediate PR ships a `main` where some pickers are broken until the rest land, and reverting any one piece leaves things inconsistent. My hope is that once the context refactor lands, the per-host diffs become small enough to review comfortably as one unit.

That said — if you still prefer a split after seeing the refactor, I'll do it. In that case I'd suggest two PRs rather than one per popup (core migration + DOB host, then the search hosts) so no intermediate state ships broken pickers.

Refactor + style fixes incoming.

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: Done in `1754931` — the hide now uses `styles.opacity0` + `styles.visibilityHidden` (and it only applies on the no-modal-ancestor fallback path; when the picker is inside a popover, the modal hides itself via the new `HiddenForOverlayContext`).

trasnake87: Done in `1754931` (same change — `styles.opacity0` + `styles.visibilityHidden`).

src/components/DatePicker/DatePickerModal.tsx
trasnake87: Done in `1754931` — the hide styles are gone from this host entirely (the modal hides itself via `HiddenForOverlayContext`), and the remaining width rule now uses `StyleUtils.getWidthStyle`.

src/components/Modal/ReanimatedModal/index.tsx
trasnake87: You were right — we do already have access to it. RNW forwards the `<Modal>` ref to ModalContent's outermost element, and its `parentElement` IS the portal root (ModalAnimation), so `1754931` drops the sentinel View and the DOM walk entirely: a plain ref on `<Modal>` + one `parentElement`. The one remaining imperative line is toggling `pointer-events` on that root — RNW forwards no prop for it (only `zIndex`), so there's no declarative path to that element; happy to adjust if you know one. Coordination is now a context provider (`HiddenForOverlayContext`): the CalendarPicker asks its hosting modal to hide in place, and the host prop-threading (`hasBackdrop`/`shouldDisablePointerEvents`/`innerContainerStyle`) is deleted — net −31 lines, with a dedicated test suite for the context.

trasnake87:
@parasharrajat Refactor is up in `1754931` — the Context-provider approach you suggested, and it came out well:

- **New `HiddenForOverlayContext`**, provided by `ReanimatedModal`: the `CalendarPicker` (the component that actually knows the year selector opened) asks its hosting modal to hide in place. The modal applies the hide style, drops its own backdrop, and makes the RNW portal root pointer-transparent.
- **The sentinel View + DOM walk are gone.** You were right that we already have access to the element — RNW forwards the `<Modal>` ref to ModalContent's outermost node, whose `parentElement` is the portal root.
- **Hosts got much simpler**: `DatePickerModal`, `FilterPopupButton`, and `SearchAdvancedFiltersPopup` no longer thread `hasBackdrop`/`shouldDisablePointerEvents`/hide styles — they only keep the popover mounted and suppress the selector's popstate. Net **−31 lines**.
- **Inline styles → `styles.opacity0`/`styles.visibilityHidden`/`StyleUtils.getWidthStyle`** across the flagged spots.
- **Tests**: new `ReanimatedModalHiddenForOverlayTest` covers the context mechanism directly (hide/restore/unmount-release); the host suites are updated to the new wiring. 64 year-selector tests + 137 modal/popover-adjacent tests green; tsc/eslint/prettier clean.

On the sub-PR question from my earlier comment: with the hosts this thin now, hopefully reviewing as one unit is comfortable — but the offer stands if you'd still prefer the 2-PR split.

trasnake87:
@parasharrajat Gentle bump when you get a chance — the Context-provider refactor you suggested has been up since last Tuesday (`1754931`, net −31 lines, all four inline threads answered, CI green). Happy to adjust anything that doesn't sit right.

src/components/DatePicker/CalendarPicker/index.tsx
```diff
@@ -101,27 +127,51 @@ function CalendarPicker({
...
+    useEffect(() => {
+        if (!selectedYearResult || selectedYearResult.contextID !== pickerContextID) {
+            return;
+        }
+        const {year} = selectedYearResult;
+        clearCalendarPickerSelectedYear();
+        // Defer applying the year to the next frame: this effect reacts to an Onyx-delivered
+        // selection, and updating state synchronously inside the effect triggers a cascading
+        // render (react-hooks/set-state-in-effect).
+        requestAnimationFrame(() => setCurrentDateView((prev) => setYear(new Date(prev), year)));
```
parasharrajat: Can we move all these effects related logic to a hook as I think this logic is side-effect to the main component for rendering year picker and is not related to the calender itself. Best would be the this hook exposes a method to open the year picker which this component comsumes.

src/components/DatePicker/index.tsx
```diff
@@ -111,7 +119,28 @@ function DatePicker({
...
+            return;
+        }
+        if (!yearSelectorGraceRef.current) {
+            return;
+        }
+        const graceTimeoutID = setTimeout(() => {
+            yearSelectorGraceRef.current = false;
+        }, 250);
+        return () => clearTimeout(graceTimeoutID);
+    }, [isYearSelectorOpen]);
```
parasharrajat, line 161: Why do we need this? can you add a comment here on the effect?

src/components/Modal/ReanimatedModal/index.tsx
```diff
@@ -217,10 +247,10 @@ function ReanimatedModal({
     if (!coverScreen && isVisibleState) {
         return (
             <View
-                pointerEvents="box-none"
+                pointerEvents={isHiddenForOverlay ? 'none' : 'box-none'}
```
parasharrajat: `isOverlayHidden` would be a better name. Update the naming similiarly everywhere such as Context as well.

src/components/Modal/ReanimatedModal/index.tsx
```diff
@@ -230,6 +260,11 @@ function ReanimatedModal({
     return (
         <LayoutAnimationConfig skipExiting={getPlatform() !== CONST.PLATFORM.WEB}>
             <Modal
+                // On web, RNW forwards this ref to ModalContent's outermost element (an HTMLElement); on native
+                // it resolves to the Modal instance and the instanceof guard in the effect above makes it a no-op.
+                ref={(node) => {
+                    modalContentRef.current = node;
+                }}
```
parasharrajat:
```suggestion
                ref={modalContentRef}
```

src/components/Modal/ReanimatedModal/index.tsx
```diff
@@ -230,6 +260,11 @@ function ReanimatedModal({
     return (
         <LayoutAnimationConfig skipExiting={getPlatform() !== CONST.PLATFORM.WEB}>
             <Modal
+                // On web, RNW forwards this ref to ModalContent's outermost element (an HTMLElement); on native
+                // it resolves to the Modal instance and the instanceof guard in the effect above makes it a no-op.
```
parasharrajat:
```suggestion
```

Remove it here. Too much comment pollution

src/components/Modal/ReanimatedModal/index.tsx
```diff
@@ -192,12 +222,12 @@ function ReanimatedModal({
...
             animationIn={animationIn}
             animationOut={animationOut}
-            style={style}
+            style={[style, isHiddenForOverlay && [styles.opacity0, styles.visibilityHidden]]}
             type={type}
             onSwipeComplete={onSwipeComplete}
             swipeDirection={swipeDirection}
         >
-            {children}
+            <HiddenForOverlayContext.Provider value={setIsHiddenForOverlay}>{children}</HiddenForOverlayContext.Provider>
```
parasharrajat: is this correct to render provider here. the year picker will open in its own page under the stack and will this provider wrap that if rendered here?

src/components/Search/FilterDropdowns/DateSelectPopup/index.tsx
```diff
@@ -52,6 +57,32 @@ function DateSelectPopup({label, value, presets, style, closeOverlay, onChange,
...
+        if (hasRestoredDateModifierRef.current || selectedDateModifier || !storedDateModifier || !storedYearSelection?.contextID.startsWith('search')) {
+            return;
+        }
+        const dateModifierToRestore = Object.values(CONST.SEARCH.DATE_MODIFIERS).find((modifier) => modifier === storedDateModifier);
+        if (!dateModifierToRestore) {
+            return;
+        }
+        hasRestoredDateModifierRef.current = true;
+        requestAnimationFrame(() => setSelectedDateModifier(dateModifierToRestore));
+    }, [storedDateModifier, storedYearSelection, selectedDateModifier]);
```
parasharrajat: It would be good if we moved all the logic inside the main yearpicker hook we discussed in the previous comment.


Currently, it seems that we are adding many different pieces to different components to open the year picker and manage temporary state.

It would be good to have it work like this: 
1. We create a hook with the logic related to the year picker opening as well as the internal state management required. 
2. We use this hook in the component that needs to open the year picker, and the hook controls all the side effects and internal state. 
3. it exposes necessary functions to manage internal state.

src/components/DatePicker/CalendarPicker/index.tsx
trasnake87: Done in `1c68db0` — moved all the year-picker effect logic out of CalendarPicker into a new `useYearSelector` hook. It owns the write-back subscription, the host-modal coordination, and the self-hide check, and exposes an `openYearPicker` method; the calendar component now just renders the calendar.

src/components/DatePicker/index.tsx
trasnake87: Comment added. This mirrors the selector state into a ref plus a short grace window so `closeDatePicker` can ignore the trailing popstate: selecting a year does a `goBack`, and its popstate reaches the close listener a beat after `isYearSelectorOpen` has flipped back to false — gating on the live flag alone would close the picker before the picked year is applied.

src/components/Modal/ReanimatedModal/index.tsx
trasnake87: Renamed — `isOverlayHidden` for the state and `OverlayHiddenContext` for the context, updated everywhere (imports, files, tests).

src/components/Modal/ReanimatedModal/index.tsx
trasnake87: Done — passing `modalContentRef` directly now. Typed the ref as `Modal` for the prop and read it back as `unknown` in the effect, since on web RNW actually stashes an HTMLElement there.

src/components/Modal/ReanimatedModal/index.tsx
trasnake87: Removed — kept a one-line pointer instead.

src/components/Modal/ReanimatedModal/index.tsx
trasnake87: It's intentional: the provider wraps only the modal's own `children` (the popover content, including CalendarPicker). The year selector opens as its own @react-navigation screen in the stack — not a descendant of the modal — so it isn't wrapped, and doesn't need to be: the only consumer is the in-modal CalendarPicker asking its host to hide while the selector is up.

src/components/Search/FilterDropdowns/DateSelectPopup/index.tsx
trasnake87: Done — that bookkeeping now lives in a small `useSearchYearSelector` hook layered on the same year-selector Onyx keys (persists/restores the open sub-view, exposes `clearBreadcrumb`), so DateSelectPopup no longer carries it. Kept it separate from the generic `useYearSelector` since the date-modifier state only exists for the search filter.

PR #91192, opened 2026-05-20
[codex] Fix zero amount expense table display

ndycode:
## Summary
- Make the report table TOTAL column prefer a valid edited amount over stale `convertedAmount`.
- Apply the same precedence in report table TOTAL sorting.
- Add regression coverage for `modifiedAmount: 0` with stale non-zero `convertedAmount`.

## Root cause
The expense report table displayed `convertedAmount` whenever it was present. After editing an expense amount to `0.00`, the transaction has `modifiedAmount = 0`, but the old backend-computed `convertedAmount` can remain on the row until a fresh value comes back. Because the table and sort path prioritized `convertedAmount`, they kept using the original amount instead of the explicit zero edit.

## Validation
- `npx prettier --write src/components/TransactionItemRow/TransactionItemRowWide.tsx src/libs/ReportUtils.ts tests/unit/ReportUtilsTest.ts`
- `git diff --check`
- `sh scripts/lintChanged.sh src/components/TransactionItemRow/TransactionItemRowWide.tsx src/libs/ReportUtils.ts tests/unit/ReportUtilsTest.ts`
- `npm run react-compiler-compliance-check -- check-changed`

Attempted focused Jest:
- `$env:TZ='utc'; $env:NODE_OPTIONS='--experimental-vm-modules --max_old_space_size=8192'; npx jest tests/unit/ReportUtilsTest.ts --runInBand --testNamePattern="getTransactionSortValue"`
- Failed before running tests due existing Jest/ESM mock setup issue in this checkout: `Must use import to load ES Module: @react-navigation/core`.

`npm run typecheck-tsgo` was attempted and failed on existing broad repo/dependency type errors unrelated to this change, including `TextInputPasteEventData` missing from `react-native`, `ScreenOptionsOrCallback` missing from `@react-navigation/native`, and many existing `userSelect: "contain"` style type errors.

$ #91006

### DIFF
--- a/src/components/ReportActionItem/MoneyRequestReportPreview/index.tsx
+++ b/src/components/ReportActionItem/MoneyRequestReportPreview/index.tsx
@@ -15,6 +15,7 @@ import useTransactionViolations from '@hooks/useTransactionViolations';
 import getNonEmptyStringOnyxID from '@libs/getNonEmptyStringOnyxID';
 import {getIOUActionForReportID, isSplitBillAction as isSplitBillActionReportActionsUtils, isTrackExpenseAction as isTrackExpenseActionReportActionsUtils} from '@libs/ReportActionsUtils';
 import {isIOUReport} from '@libs/ReportUtils';
+import {hasValidModifiedAmount} from '@libs/TransactionUtils';
 import {startSpan} from '@libs/telemetry/activeSpans';
 import Navigation from '@navigation/Navigation';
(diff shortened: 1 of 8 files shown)


inimaga:
Please review our [contributor guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md) on procedures to follow when contributing.

Issue #85802, opened 2026-03-19
[Due for payment 2026-06-30] Add stop address editing for GPS mileage


JmillsExpensify:
**Problem**
When an employee creates a GPS mileage expense, if the recorded trip has an inaccurate stop destination (e.g., forgot to stop), then they have to recreate the expense using manual mileage or a re-drawn map route, which is harder to reproduce accurately and provides weaker supporting documentation for approvers and auditors.

**Solution**
In one release, we will:
 - Make the stop address a push row
 - Show an “Edit stop” page, with a map and a slider below
 - Users can drag the slider to decrease the distance, which is animated on the map
 - Where ever the users stops the slider and confirms, we save that stop address when creating the expense (and updating the mileage receipt).

<img width="2410" height="1624" alt="Image" src="https://github.com/user-attachments/assets/a0b742a6-f01d-4bc4-a5e1-976b36d76669" />

**Further details**
 - We'll start saving GPS way points in the BE, which means storing the `gpsCoordinates` in the `commentNVPs` (where we currently store existing distance data like `routeImageURL` and `odometer` readings).
 - We'll also start saving any mileage decreases via `modifiedDistance`, similar to `modifiedCreated`, `modifiedAmount`, `modifiedMerchant` and `modifiedMCC`.

This is an internal initiative, so we won't open it up to proposals

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

dylanexpensify:
Moved back to weekly as we don't want this monthly

GCyganek:
I have the implementation pretty much ready, but it depends on https://github.com/Expensify/App/pull/88609 that should be  switched to review soon (when the related backend PR is deployed to prod), so I'll post a PR when it's merged. We will also have to wait for a backend update to be able to generate receipts with modified stop

GCyganek:
@Expensify/design I'm working on applying the new route line + icons designs:

1. Should I be using assets exported directly from here to get the icons? https://www.figma.com/design/yaFFb7MH1mGeOr79a1wSWH/GPS-Milage-Tracking?node-id=463-737
2. We currently use dot-indicator icon for waypoints between start & end and dot-indicator-unfilled for the start waypoint, but on the design there is only a new version for dot-indicator, so I'm wondering if you want me to use it for every waypoint apart from the end waypoint?

See what I currently have:

[...]

PR #90237, opened 2026-05-11
[CP Staging] Show Resume/Save buttons when GPS tracking is stopped immediately

MelvinBot:
### Explanation of Change

When a user starts GPS tracking and immediately taps Stop before moving 100+ meters, `stopGpsTrip` removes the only GPS segment (because it has 0 or 1 points), leaving `gpsPoints: []`. Then `isTripStopped` returns `false` (requires total points > 0), so `GPSButtons` renders the Start button instead of Resume/Save.

This PR makes two changes in `GPSDraftDetailsUtils.ts`:
1. **`stopGpsTrip`**: Only remove an empty/single-point last segment when there are multiple segments (`gpsPoints.length > 1`). When it's the only segment, keep it so the trip state is preserved.
2. **`isTripStopped`**: Check segment count (`gpsPoints.length > 0`) instead of point count, so it correctly returns `true` when segments exist but are empty.

The existing `zeroDistanceModal` in `saveGpsTrip` already handles the case where the user tries to save a zero-distance trip.

### Fixed Issues
$ https://github.com/Expensify/App/issues/90103
PROPOSAL: https://github.com/Expensify/App/issues/90103#issuecomment-4412884889

### Tests
  1. Start a distance request with GPS tracking enabled
  2. Tap Stop immediately without moving — verify the buttons show Resume and Save (not Start)
  3. Tap Resume, walk 100+ meters, tap Stop — verify Resume and Save still appear correctly
  4. Tap Save on a zero-distance trip (step 2 scenario) — verify the zero-distance modal appears
  5. Verify no console errors throughout

- [x] Verify that no errors appear in the JS console

### Offline tests
N/A — GPS tracking requires an active connection; the fix is purely state logic with no offline-specific
  behavior.

### QA Steps
  1. On a physical device, open a distance request with GPS tracking
  2. Tap Start, then immediately tap Stop without moving — confirm Resume and Save buttons are shown (this was the
  bug: Start appeared instead)
  3. Tap Resume, move for a bit, tap Stop — confirm Resume and Save still appear correctly
  4. Complete a normal full-distance trip end-to-end and save it — confirm no regression

- [ ] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
[...]

Issue #90103, opened 2026-05-09
[Due for payment 2026-05-20][$250] Track distance - GPS - Stop GPS tracking and Resume not displayed if tap on Stop GPS tracking


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.69-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://test-management.browserstack.com/projects/2219752/folder/39523254/test-cases/86236110
**Email or phone of affected tester (no customers):** applausetester+fish50826@applause.expensifail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Regression TC Execution
**Device used:** iOS
**App Component:** Other

## Action Performed:
1. Launch the app and login with expensifail account
2. Tap the Global + > Track distance
3. Select GPS
4. Tap Start > Immediately tap Stop before moving

## Expected Result:
Verify Stop GPS tracking modal displayed with Stop GPS tracking and Resume tracking buttons

## Actual Result:
Stop GPS tracking and Resume tracking buttons doesn't displayed.

## Workaround:
Unknown

## Platforms:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [x] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [ ] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/7f247bbb-0e45-4756-ad9e-8864800087de

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022053913406861052385</li>
        <li>Upwork Job ID: 2053913406861052385</li>
        <li>Last Price Increase: 2026-05-11</li>
    </ul>
</details>

MelvinBot:
## Proposal

### What is the root cause of that problem?

[...]

PR #91418, opened 2026-05-22
[GPS] Update user location on GPS screen when there is no active GPS trip

GCyganek:
### Explanation of Change

Requested [here](https://github.com/Expensify/App/pull/90169#issuecomment-4476930634)

Summary: New `GPSMapView` that uses Mapbox internal location engine via `followUserLocation` Camera prop to smoothly follow user location instead of jumping from one spot to another during GPS location updates for the GPS trip.

List of changes:

#### New GPSMapView component
**What**: New native-only map component used on the GPS distance screen instead of DistanceMapView.
**Why**: GPS tracking needs different camera behavior (follow user, permissions, custom puck, route animation) than the manual distance map.
**How**: Handles foreground location permission, followUserLocation, center/recenter, fallback marker when permission is denied, and wires up waypoints, route, and location puck.

#### @rnmapbox/maps patch — immediate follow transition
**What**: Adds opt-in followUserLocationUseImmediateTransition to Mapbox Camera (iOS + Android).
**Why**: On first follow, Mapbox animates from world view (“fly-in from space”); there is no JS-only fix when followUserLocation is true.
**How**: Uses makeImmediateViewportTransition() on initial follow (no waypoints); later re-centers use the default animated transition.

#### GPSDirection + useAnimatedTrailingDirectionCoordinate
**What**: Smoothly animates the trailing end of the GPS route line as the user moves.
**Why**: Raw GPS updates make the line jump; animation keeps movement fluid during an active trip.
**How**: Interpolates the last segment endpoint over GPS_ROUTE_ANIMATION_DURATION_MS (1s) using requestAnimationFrame.

#### GPSWaypointLayer
**What**: Renders start/stop/waypoint markers as a Mapbox SymbolLayer instead of MarkerView.
**Why**: Markers must sit below the native location puck so the puck stays on top.
**How**: Uses belowLayerID with platform-specific puck layer IDs; remounts on Android when the puck layer becomes available.

#### Custom location puck (MapCurrentLocationPuck)
**What**: Custom SVG puck for Mapbox LocationPuck, plus shared blue fill color in CONST.
**Why**: Default Mapbox puck does not match Expensify design; puck geometry is tuned for Mapbox’s anchor.
**How**: Registers puck image via Mapbox.Images; hides default UserLocation marker and listens for updates for route animation.

#### Platform-specific locationPuckLayerId
**What**: Exports the native puck layer ID per platform (puck on iOS, mapbox-location-indicator-layer on Android).
**Why**: Layer IDs differ by platform and are required for correct z-ordering.

#### Extract useAccessToken hook
**What**: Shared hook for setting the Mapbox access token.
**Why**: Deduplicates logic between MapView and GPSMapView.
**How**: MapView now uses the hook instead of an inline useEffect.

#### GPS screen integration (IOURequestStepDistanceGPS)
**What**: Swaps DistanceMapView for GPSMapView; removes manual flyTo on each GPS update and the map ref.
**Why**: Follow mode and animated route line replace imperative camera moves; tracking state is passed via isTrackingGPS.

#### GPS draft segment handling
**What**:

- removeLastSegment: clears the last segment when it is the only one (instead of removing it).
- stopGpsTrip: removes the last segment when empty/single-point even if it is the only segment.
**Why**: Avoids dropping the last segment entirely when stopping a trip with a single segment.

#### Types & constants
**What**: GPSMapViewProps, GPSDirectionProps, optional markerType on WayPoint; new WAYPOINTS layer IDs, GPS_ROUTE_ANIMATION_DURATION_MS, MAP_CURRENT_LOCATION_FILL_COLOR.
**Why**: Supports the new GPS map API and shared styling/animation values.

#### Map utils
**What**: areCoordinatesEqual, interpolateCoordinate; isSingleSegmentRoute no longer depends on is2dArray.
**Why**: Supports route animation and multi-segment GPS coordinates.

#### ESLint seatbelt
**What**: Adds GPSMapView.tsx to the seatbelt for react-hooks/set-state-in-effect.
**Why**: Keeps lint passing for intentional setState-in-effect usage.
`isSingleSegmentRoute` is fixed as it was returning `true` for `[[]]` which is a multi segment empty route

### Fixed Issues
$ https://github.com/Expensify/App/issues/85802
PROPOSAL: N/A

### Tests

Camera behaviour:

1. FAB > Track distance > GPS
2. Verify that the map shows correctly current location marker with pulsing animation and smoothly animates following user location
3. Move the map using gestures
4. Verify that it stopped following user location
5. Tap top-right map button
6. Verify it animates smoothly to user location and starts following user location again
7. Tap Start to start the GPS trip
8. Verify the recorded route smoothly follows user location marker and camera follows user location marker
9. Move the map
10. Verify that it stopped following user location marker
11. Tap top-right map button
12. Verify it animates smoothly to user location and starts following user location again
13. Stop GPS trip
14. Verify it moves the map to show the whole recorded route on the map
15. Move the map
16. Tap top-right map button
17. Verify it smoothly moves to show the whole recorded route on the map again
18. Move the map and leave Track distance screen
19. FAB > Track distance > GPS
20. Verify it shows the whole recorded route on the map
21. Discard the trip
22. Verify it transitions smoothly to show the current user location and keeps following the marker

Foreground location permissions not granted yet (granted):
Prerequisites: foreground location permissions not granted, user can ask, precise location is on, last visited Track distance screen is GPS:

1.  FAB > Track distance > GPS
2. Verify there is an os prompt asking for location permissions
3. Tap `Allow while using app` (iOS) or `While using the app` with `Precise` location (Android)
4. Verify it shows location marker that updates smoothly and that has pulsing animation

Foreground location permissions not granted yet (denied):
Prerequisites: foreground location permissions not granted, user can ask, precise location is on, last visited Track distance screen is GPS:

1.  FAB > Track distance > GPS
2. Verify there is an os prompt asking for location permissions
3. Tap `Don't allow`
4. Verify it shows location marker without pulsing animation based on the last saved location or default location (San Francisco) that is not being updated as we do not have location permissions

Foreground location permissions not granted, GPS trip saved:

1.  FAB > Track distance > GPS
2. Record a GPS trip
3. Leave Track distance screen
4. Change location permissions to `Never` (iOS) or `Don't allow` (Android) in app's settings 
5. Go back to the app
6. Go back to GPS screen
7. Verify it opens the map with the whole recorded route visible on the map
8. Move the map
9. Tap top-right map button
10. Verify it transitions smoothly to recorded route
11. Discard the GPS trip
12. Verify it transitions smoothly to current user location marker that is not being updated because of no location permissions

### Offline tests

1. FAB > Track distance > GPS
2. Verify it shows `Map pending` screen in offline mode

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

Camera behaviour: 

https://github.com/user-attachments/assets/46a9e244-50f1-4d61-b471-41c1ffd3ae45

Offline

Screenshot 2026-06-16 at 18 41 39

Foregrund location permissions:

https://github.com/user-attachments/assets/b0b02188-e380-4e38-ac75-901d35a93672

https://github.com/user-attachments/assets/88798330-4a00-4981-93c4-7e7f8d358a27

https://github.com/user-attachments/assets/b8bd8da1-281f-4032-980b-637dca7469ac

</details>

<details>

Camera behaviour: 

https://github.com/user-attachments/assets/3242b266-1f70-451c-8640-56ca7f2307e1

Offline:

Screenshot 2026-06-16 at 18 21 56

Foreground permissions

https://github.com/user-attachments/assets/db2316c7-cb73-45f1-9c07-e128c87e75a0

https://github.com/user-attachments/assets/5e872222-1fd8-418a-8269-c028d6da24d2

https://github.com/user-attachments/assets/1884f12c-d440-47f5-b046-294c38b40282

https://github.com/user-attachments/assets/f94999f8-4ac8-47b8-b210-52cf4151bd08

https://github.com/user-attachments/assets/08268558-1a8d-40db-814b-6a3debd23e94

</details>


GCyganek:
@Expensify/design @JmillsExpensify so I added user's location updates every 5 seconds even if there is no active GPS trip, but in this case I feel like it would be better to not move the map to the updated location every time and staying on the recorded route is better than this IMO: 

https://github.com/user-attachments/assets/0fe0351b-3460-4ccc-902c-8118901106be

What do you think? And is 5 seconds ok?

shawnborton:
Can you rephrase, I'm not quite following...

JmillsExpensify:
Hmm, yes I think I agree with Gabriel. Basically if we move to the updated location it's a bit jarring with the zoom. Then additionally, the blue dot sticks at the last recorded location. Is that what you mean?

GCyganek:
Sorry, what I mean is currently because we do not update current location when GPS tracking is paused/not started the map shows the whole recorded route and user can freely move the map. But now when I added current location updates every 5 seconds when the tracking is not active, the map keeps moving to show latest location on every update, so user can't freely move the map or see the recorded trip without the map flying to latest location after a few seconds as you can see on the video. 

IMO it would be better to:
- when there is a recorded GPS trip, do not move the map on the current location updates to keep the recorded route in view on the map
- or at least make sure that when user has moved the map or tapped map controls at the top right corner, the map does not move with the current location updates anymore

shawnborton:
Nice, yeah I think that makes sense then and I agree. The first bullet makes the most sense to try first, right?

GCyganek:
> Nice, yeah I think that makes sense then and I agree. The first bullet makes the most sense to try first, right?

Yep, just wanted to confirm that you want it like this

JmillsExpensify:
Nice, I agree as well. Let's try the first one and we can always re-assess later.

GCyganek:
I'll push changes in a moment, but now it will work like this:

https://github.com/user-attachments/assets/1042b6b4-ca18-4d09-9941-3f56e7407b06

See it follows users location when there is no recorded trip and while recording it, but afterwards it does not follow the user, it just keeps showing the recorded route. When user moves the map and then taps map controls it moves it back to the recorded trip

shawnborton:
I think that looks good? Happy to help test as well.

GCyganek:
@codex review

GCyganek:
@shawnborton feel free to run build here to test and see if it works good for you

src/components/MapView/MapView.tsx
```diff
@@ -212,7 +213,7 @@ function MapView({
     const centerCoordinate = useMemo(() => (currentPosition ? [currentPosition.longitude, currentPosition.latitude] : initialState?.location), [currentPosition, initialState?.location]);
 
     const waypointsBounds = useMemo(() => {
-        if (!waypoints) {
+        if (!waypoints || !waypoints.length) {
```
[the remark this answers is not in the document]
GCyganek: fixed

GCyganek:
I found a bug in one edge case where it throws an error and I just fixed it, so it shouldn't appear in the next build, sorry for that in case you encounter it

shawnborton:
I will test this tomorrow at some point I hope, sorry for the delay!

GCyganek:
> I will test this tomorrow at some point I hope, sorry for the delay! 

No problem!

shawnborton:
This is feeling really good, just tested! 

I will say that I don't think there's any perceivable changes related to this:
> Verify that the user location marker on the map updates within ~5 seconds to reflect the new position.

For example, it didn't track me going around this circle here, it shortened it quite a bit:
CleanShot 2026-05-26 at 15 04 18@2x

Everything else seems good though!

GCyganek:
> > Verify that the user location marker on the map updates within ~5 seconds to reflect the new position.
> 
> For example, it didn't track me going around this circle here, it shortened it quite a bit:

The only perceivable change on this PR is that now when the tracking is not active the current location marker is still being updated every 5 seconds. The screenshot you attached has recorded route and there were no changes according to how often we do location updates while recording the GPS trip. As Jason mentioned on related issue, we will probably experiment with location updates frequency once we will work on user's activity updates.

But I can change the frequency of user location updates while not recording a GPS trip if you want something else than 5 seconds

shawnborton:
Ah okay, apologies for the confusion! All good on my end then, looking forward to the route improvements :)

GCyganek:
> Ah okay, apologies for the confusion! All good on my end then, looking forward to the route improvements :)

No worries, will switch this for review then, thank you!

src/pages/iou/request/step/IOURequestStepDistanceGPS/index.native.tsx
```diff
@@ -174,6 +178,40 @@ function IOURequestStepDistanceGPS({
...
+        useCallback(() => {
+            // If there is an ongoing GPS trip, do not poll foreground location
+            if (gpsDraftDetails?.isTracking) {
+                return;
+            }
+
+            let cancelled = false;
+
+            const intervalId = setInterval(() => {
+                if (cancelled) {
```
dukenv0307: I think we can move `cancelled` to the .then function. Otherwise, it can be removed entirely since we already called clearInterval in the clean-up function

GCyganek: Yeah, absolutely, I planned to check it inside `.then()` huh, thanks for catching this

dukenv0307 (approved):

JmillsExpensify (dismissed):
Looks good to me as well.

dukenv0307:
@AndrewGable All yours!

JmillsExpensify (dismissed):
LGTM

dukenv0307:
@AndrewGable can you please review it?

dylanexpensify:
We need a new internal engineer, Andrew is OOO!

JmillsExpensify:
Going to assign @neil-marcellini since he offered to help out as time permits.

src/pages/iou/request/step/IOURequestStepDistanceGPS/index.native.tsx
```diff
@@ -178,6 +182,40 @@ function IOURequestStepDistanceGPS({
...
+            if (gpsDraftDetails?.isTracking) {
+                return;
+            }
+
+            let cancelled = false;
+
+            const intervalId = setInterval(() => {
+                // We do not check foreground location permissions here because
+                // it's already checked in DistanceMapView to get initial user's location
+                getCurrentPositionAsync()
```
neil-marcellini: Why did we decide to go with a polling approach vs `watchPositionAsync`? It seems like that might be better, especially for battery if not moving that much?

src/components/MapView/MapView.tsx
```diff
@@ -42,6 +42,7 @@ function MapView({
     unit,
     ref,
     shouldDisplayCurrentLocation = true,
+    shouldFollowCurrentLocationWhenRoutePresent = true,
```
neil-marcellini: I find this prop a bit confusing, and I don't think it's totally necessary, because I think we already don't follow the current location when a route is present. Plus, we have existing checks in the component for when to center on the user's location. Here's a suggestion for how we could have the same UX without this prop.

(Neil's AI agent)

**Suggestion: drop `shouldFollowCurrentLocationWhenRoutePresent` and reuse existing `shouldPanMapToCurrentPosition` logic**

Polling updates `USER_LOCATION` → `centerCoordinate` changes. `MapView` already blocks camera panning when waypoints exist via `shouldPanMapToCurrentPosition`, but the `centerCoordinate` prop on `Mapbox.Camera` (and `defaultSettings`) was not gated — that's what caused the map to jump after stopping a trip.

Instead of a new prop, apply the same condition everywhere the camera can follow the user:

```ts
const shouldFollowUserLocation = useMemo(
    () => !userInteractedWithMap && shouldDisplayCurrentLocation && (!waypoints || waypoints.length === 0),
    [userInteractedWithMap, waypoints, shouldDisplayCurrentLocation],
);

const defaultSettings = useMemo(() => {
    if (interactive && shouldFollowUserLocation) {
        return centerCoordinate ? {zoomLevel: initialState?.zoom, centerCoordinate} : undefined;
    }
    if (!waypointsBounds) {
        return undefined;
    }
    return {bounds: waypointsBounds};
}, [interactive, centerCoordinate, waypointsBounds, initialState?.zoom, shouldFollowUserLocation]);

const initCenterCoordinate = useMemo(
    () => (interactive && shouldFollowUserLocation ? centerCoordinate : undefined),
    [interactive, centerCoordinate, shouldFollowUserLocation],
);
```

Then remove `shouldFollowCurrentLocationWhenRoutePresent`, `shouldKeepCameraOnRoute`, and the prop from the GPS screen. The GPS screen would only add foreground location polling.

This should preserve the same UX:

- No trip → waypoints empty → camera follows user
- Stopped with recorded route → waypoints present → camera stays on route bounds, marker still updates
- After discard → waypoints cleared → camera follows user again

`true` as the default for the new prop doesn't enable an intentional product behavior elsewhere — it was only load-bearing once polling introduced continuous location updates. One rule ("follow user only when there are no waypoints") is simpler than a prop nothing passes explicitly.

Optional: rename `shouldPanMapToCurrentPosition` → `shouldFollowUserLocation` since it would gate more than the pan `useEffect`.

Minor unrelated: JSDoc on the prop (if kept) is inverted; also consider `watchPositionAsync` over 5s polling, and unit tests for `isSingleSegmentRoute`.

neil-marcellini (changes requested):
Looks pretty good, thanks for improving this incrementally. I think one of the props is making this confusing so that's the main reason I request changes.

Regarding the [video from this comment](https://github.com/Expensify/App/pull/91418#issuecomment-4533106244) showing the behavior, I agree it's good but I also think it could be much better.

I think we should move away form a polling approach and receive updates from the OS at a high frequency, something like once per second. Due to the low sample rate of once every 5 seconds, there is a pretty big jump between updates.

NAB: It might also be good to smoothly animate between position updates by calculating an estimated velocity and animating if it's above a certain threshold. Or maybe mapbox allows us to adjust the "fly to position" speed and acceleration so we can achieve the same effect. I seem to remember that google maps will show your location moving very smoothly across a map if you open it up in a car, for example.

We don't need to make it perfectly smooth in this PR, but maybe we can try more frequent updates.

NAB: Also, the symbol that we show at the top right is commonly used to center the map on your current location, not the route, so that's a bit unexpected for me and I would rather use a different UX.

Maybe we leave that current location centering icon button in place, change it to always center on the user, and have a separate button at the bottom left that pops up if there is a route and you are not centered on it, which just says "Re-center" and re-centers you on the route?

src/pages/iou/request/step/IOURequestStepDistanceGPS/index.native.tsx
GCyganek: I did that mainly because I thought we wanted to have regular updates based on time interval, so that even if user moves only a little during the last 5 seconds, there is still an update that user can see on the map. `watchPositionAsync` has `timeInterval` prop working only on Android, so we would have to rely on the distance interval instead. I aggree that `watchPositionAsync` seems like a better fit here, I just thought that polling is better if we want to have these regular updates every x seconds no matter if user moved 5 meters or 100 meters

src/components/MapView/MapView.tsx
GCyganek: I agree that this is not the best solution after giving it some more thought. I applied your suggested changes and added some minor improvements, like reseting `userInteractedWithMap` state to `false` when discarding the trip or when tapping the top-right button which goes back to user's current location when there is no waypoint - this way it keeps following location updates on the map until user moves the map again

GCyganek:
@shawnborton @JmillsExpensify could you confirm if this UX is fine in these scenarios? I'm curious because I noticed that after user has interacted with the map in any way, we stop following user's location updates. So in these two cases I decided to change the logic to make sure that we start following the location updates again (1. when user interacted with the map and then tapped top-right button to go back to their current location; 2. when user interacted with the map and then discarded the trip):

1. User opens GPS screen, interacts with the map => map stops following user's location updates
2. User taps top-right button to center their current location on the map
3. Map starts following user's location updates again

alternatively in point 3. the map would stay on the location from the time user tapped the button, but then updates would not have been followed 

https://github.com/user-attachments/assets/ca696770-08c1-4fe4-82d8-1683746f1e1b


1. User opens GPS screen, records GPS trip, stops it, map doesn't follow location updates as we want to be showing recorded trip
2. User interacts with the map, then discards the trip
3. Map goes back to user's current location and follows the updates

alternatively in point 3. the map would not move at all after discarding the trip 

https://github.com/user-attachments/assets/fe359083-017f-4842-9e71-da1edb0029cb

shawnborton:
I think all of that makes sense to me 👍 Thanks for thinking through those scenarios.

GCyganek:
@shawnborton @JmillsExpensify A few suggestions from @neil-marcellini here:

> NAB: Also, the symbol that we show at the top right is commonly used to center the map on your current location, not the route, so that's a bit unexpected for me and I would rather use a different UX.
> 
> Maybe we leave that current location centering icon button in place, change it to always center on the user, and have a separate button at the bottom left that pops up if there is a route and you are not centered on it, which just says "Re-center" and re-centers you on the route?


and this one suggesting more frequent updates:

> I think we should move away form a polling approach and receive updates from the OS at a high frequency, something like once per second. Due to the low sample rate of once every 5 seconds, there is a pretty big jump between updates.


and smooth animation suggestion:

> NAB: It might also be good to smoothly animate between position updates by calculating an estimated velocity and animating if it's above a certain threshold. Or maybe mapbox allows us to adjust the "fly to position" speed and acceleration so we can achieve the same effect. I seem to remember that google maps will show your location moving very smoothly across a map if you open it up in a car, for example.
> 
> We don't need to make it perfectly smooth in this PR, but maybe we can try more frequent updates.

shawnborton:
> NAB: Also, the symbol that we show at the top right is commonly used to center the map on your current location, not the route, so that's a bit unexpected for me and I would rather use a different UX.

I don't feel strongly, but it's a fair point. A re-center button is more akin to modern map apps... we could do something like this? I think we might need to use a success style button so we get separation from the background. Thoughts @Expensify/design ?

CleanShot 2026-06-10 at 11 31 17@2x


> I think we should move away form a polling approach and receive updates from the OS at a high frequency, something like once per second. Due to the low sample rate of once every 5 seconds, there is a pretty big jump between updates.

You know how I feel about this :) big yes to improving this. 

> NAB: It might also be good to smoothly animate between position updates by calculating an estimated velocity and animating if it's above a certain threshold.

No strong feelings, I like the idea, would be curious to see some POCs.

JmillsExpensify:
I think those are all good points of feedback, but I think the better solution for polling is that we actively use OS-level activity updates. That's a more efficient use of the device's battery, and equally, we have already planned to incorporate activity updates in a follow-up initiative. So my personal opinion is that we shouldn't block this PR on more frequent polling, as I think there is still an even better solution, that is also better for the user.

JmillsExpensify:
As for the rest of the NAB comments, I think those and I also agree that they are NAB. Let's tackle them down the line, since this PR is holding up stoping editing, which is the main feature users are asking for.

JmillsExpensify:
> @shawnborton @JmillsExpensify could you confirm if this UX is fine in these scenarios? I'm curious because I noticed that after user has interacted with the map in any way, we stop following user's location updates. So in these two cases I decided to change the logic to make sure that we start following the location updates again (1. when user interacted with the map and then tapped top-right button to go back to their current location; 2. when user interacted with the map and then discarded the trip):
> 
> 1. User opens GPS screen, interacts with the map => map stops following user's location updates
> 2. User taps top-right button to center their current location on the map
> 3. Map starts following user's location updates again
> 
> alternatively in point 3. the map would stay on the location from the time user tapped the button, but then updates would not have been followed
> 
>  Screen.Recording.2026-06-10.at.10.56.37.mov 
> 1. User opens GPS screen, records GPS trip, stops it, map doesn't follow location updates as we want to be showing recorded trip
> 2. User interacts with the map, then discards the trip
> 3. Map goes back to user's current location and follows the updates
> 
> alternatively in point 3. the map would not move at all after discarding the trip
> 
>  Screen.Recording.2026-06-10.at.10.57.19.mov

For this one, agree with Shawn. This all sounds great to me.

GCyganek:
Thanks for your feedback! So in case we will want to add Re-center button I'll do that in a follow-up PR, same for the location updates frequency - we want to stay with what we currently have and improve it after we start using user activity updates

As for the user location animation, I'll see what options we have and share my findings with you

shawnborton:
Sounds like a plan 👍

GCyganek:
So it looks like we have an interesting option to use here - we can easily turn on Mapbox's own native location engine to get very smooth location tracking on the map:


https://github.com/user-attachments/assets/e6cceb3e-002e-4210-9def-9adbd7455e65

And it gives us pulsing animation that can have custom radius and color basically for free! The problem is that we can't sync Mapbox location tracking with our GPS location tracking, so it would be very tricky to make the route follow the current location marker perfectly. But still, I'm curious what you think about it @shawnborton @JmillsExpensify 

If we'd like to use it, then I'd need some time to polish this solution and test it thoroughly, but it looks promising

JmillsExpensify:
I agree, it looks really promising. Down to try and make it work.

dannymcclain:
That smoothing is pretty nice!

>A re-center button is more akin to modern map apps... we could do something like this? I think we might need to use a success style button so we get separation from the background. Thoughts @Expensify/design ?

Yeah I don't mind the success version! But I think I'm a little confused on the difference between the re-center button and the crosshairs button?

GCyganek:
> But I think I'm a little confused on the difference between the re-center button and the crosshairs button?

Crosshairs button current behaviour:
- if there are waypoints and user location on the map -> center to show waypoints
- if there are only waypoints on the map -> center to show waypoints
- if there is only user location on the map -> center to show user location

shawnborton:
That does look promising indeed, I like it!

GCyganek:
Let's then try the smooth location tracking solution instead of merging the current one? I should be able to finish it tomorrow and it would replace current 5sec location polling mechanism

dannymcclain:
@shawnborton any styling tips for the little pulsing radius? It doesn't look terrible as is, but maybe we could use our brighter green or something for it?

shawnborton:
Yeah, I could see trying it in a lighter green, or trying it where it felt more like a "ring" was growing and expanding off of the marker dot?

dubielzyk-expensify:
> I don't feel strongly, but it's a fair point. A re-center button is more akin to modern map apps... we could do something like this? I think we might need to use a success style button so we get separation from the background. Thoughts @Expensify/design ?

Big fan of the right one (green)

> @shawnborton any styling tips for the little pulsing radius? It doesn't look terrible as is, but maybe we could use our brighter green or something for it?

I think we should use blue here instead of green cause the blue dot is blue. Also I think it should blink with about half the interval it does at the moment as it's very frequent. I think reducing the starting opacity a tiny bit would be good to.

shawnborton:
I like those ideas!

GCyganek:
Update: I'm still trying to make this work with Mapbox's smooth location tracking, but it's a bit painful. I had to fix a few bugs with this new approach and I'm close to the end, but I'm still working on a better solution for route drawing so that it follows the location dot smoothly. I'm also trying to find some workaround for the fly-in animation, as Mapbox by default does animation from space level to street level on first map render and I guess we do not want that and showing pending map view a bit longer would look better compared to this:


https://github.com/user-attachments/assets/f535f706-2530-4ef3-8b57-94dcd16af92d




As for the pulsing animation, Mapbox does not give many options for customization (only radius and color), so if we want different opacity and interval I'd have to implement our custom solution for this, but I haven't tried that yet to see if it would be hard. Will take a look at it after I'm done with other things mentioned above (which hopefully won't take longer than few hours tomorrow).

shawnborton:
For the pulsing thing, we could at least just try it with blue for now and see how it goes before diving in too deep. Color would be `#0185FF`

shawnborton:
> as Mapbox by default does animation from space level to street level on first map render 

I mean this is kinda cool... but I see what you mean. I don't hate it though!

JmillsExpensify:
Yeah, I don't either. It's kind of nice for perspective.

dubielzyk-expensify:
I think it's a bit too over the top (and that's me who usually love that shit haha). Either way it's fine, but I think snapping straight to ground level is less distracting imo.

GCyganek:
@Expensify/design @JmillsExpensify could you please test it on an adhoc build? With emphasis on route animation, fly-in animation (as I see it's not like you really hate it, so I just left it as is for now for you to see) on map load and overall experience after this change.


Route now should follow the user location marker smoothly:

https://github.com/user-attachments/assets/d78cee45-18b8-441b-9197-04ced3adaba1


I just want to get some feedback on this to see what else I could fix here and what would be possible to do as a follow-up to finally get this PR merged

shawnborton:
Hmm it's still not following my route very closely, it missed this little curve for instance: 
CleanShot 2026-06-12 at 14 47 07@2x


The pulsing animation is nice. And I think we prefer not to use the globe fly-in if possible.

shawnborton:
Still not handling tunnels very well 🤣 

CleanShot 2026-06-12 at 14 48 29@2x

dannymcclain:
Pulsing looks great to me, and I also would prefer not to start from space 🧑‍🚀 👍

GCyganek:
@shawnborton the problem is that there is no change to the accuracy of the trip in this PR. We still would have to tweak our location tracking settings to make it more accurate (as we currently just use 0.1mi distance interval for the location updates), but that's what we want to do with activity updates as @JmillsExpensify mentioned here before. I think we also have discussed the tunnels problem some time ago and that's a separate issue that we could take a deeper look at in the future, if it's even possible to do something about it. 

I'll try to do something about the fly-in from space animation.

shawnborton:
Cool, sounds like a plan 👍

JmillsExpensify:
Agreed, sounds great.

GCyganek:
@Expensify/design can you build and check now? If everything is ok I will switch for review

I fixed fly-in animation and now it immediately snaps to the route/current location on map load:


https://github.com/user-attachments/assets/15f5a0bf-6a98-45c3-a66e-4ccfec4d20b1

GCyganek:
@codex review please

patches/@rnmapbox/maps/@rnmapbox+maps+10.1.44+001+immediate-follow-camera-transition.patch
```diff
@@ -0,0 +1,160 @@
...
+         camera.setFollowUserMode(value.asStringOrNull())
+diff --git a/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXCamera.swift b/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXCamera.swift
+index 69c7e27..ca13fbd 100644
+--- a/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXCamera.swift
++++ b/node_modules/@rnmapbox/maps/ios/RNMBX/RNMBXCamera.swift
+@@ -142,6 +142,8 @@ open class RNMBXCamera : RNMBXMapComponentBase {
+     }
+   }
+ 
++  @objc public var followUserLocationUseImmediateTransition: Bool = false
```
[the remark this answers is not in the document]
GCyganek: I'll check but it works for me even with default `false` value

shawnborton:
Will run it now 🚀

shawnborton:
Nice, that feels pretty good to me. Let's get this into final review then?

JmillsExpensify:
Still looks good to me, so I think we're back to Neil?

GCyganek:
@JmillsExpensify Yeah, I just have been working on and off today and I need to make sure this works correctly with different location permissions, after that I will switch for review

GCyganek:
@dukenv0307 @neil-marcellini can you take a look at this again? The concept has changed so I updated the description and described all the changes, you can also check the whole discussion here for full context.

dukenv0307:
on it now

src/components/MapView/GPSDirection.tsx
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        const lastSegment = directionCoordinates.at(-1);
+        if (!lastSegment) {
+            return directionCoordinates;
+        }
+
+        if (lastSegment.length === 0) {
+            return directionCoordinates;
+        }
```
dukenv0307:
```suggestion
        if (!lastSegment?.length) {
            return directionCoordinates;
        }
```

src/components/MapView/GPSMapView.tsx
```diff
@@ -0,0 +1,285 @@
...
+        const {southWest, northEast} = utils.getBounds(
+            waypoints.map((waypoint) => waypoint.coordinate),
+            directionCoordinates,
+        );
+        cameraRef.current?.fitBounds(northEast, southWest, mapPadding, CONST.MAPBOX.ANIMATION_DURATION_ON_CENTER_ME);
+    });
+
+    const centerMap = () => {
+        const waypointCoordinates = waypoints?.map((waypoint) => waypoint.coordinate) ?? [];
+        if (!isTrackingGPS && (waypointCoordinates.length > 1 || (directionCoordinates ?? []).length > 1)) {
```
dukenv0307:
```suggestion
        if (!isTrackingGPS && (waypointCoordinates.length > 1 || directionCoordinates?.length > 1)) {
```

dukenv0307:
1. Start tracking
2. go back
3. Click `GPS tracking is progress...` button
4. Observe that the route isn't shown

https://github.com/user-attachments/assets/b9f6ccdd-0d17-42da-b3a9-7366507da949

src/components/MapView/GPSDirection.tsx
GCyganek: Makes sense

GCyganek:
@dukenv0307 should be fixed now, layering is very tricky with this LocationPuck component because its layer is not ready to be used as `belowLayerID` prop immediately, so I need to defer it for `GPSWaypointLayer` and `GPSDirections`:

https://github.com/user-attachments/assets/24089d22-afd7-464e-9796-61c49b10c532

I also added additional foreground location permissions check in `useAppFocusEvent` when user backgrounds the app and then foregrounds it again (in case they change location permissions in the meantime):

https://github.com/user-attachments/assets/9c4766a1-fb6d-49c1-8a6e-f02ff7a3f4aa

dukenv0307:
I think there's an issue on BE side, I can't login anymore... I'll try to test on Android soon

Screenshot 2026-06-18 at 10 06 08 PM

dukenv0307 (approved):

dukenv0307:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



https://github.com/user-attachments/assets/5f6bb967-7c01-4a43-9d08-968c665609b3

</details>

<details>



</details>

<details>



https://github.com/user-attachments/assets/569e5995-b5a1-4212-b036-eacb2e7f105d


</details>

<details>



</details>

<details>



</details>

dukenv0307 (approved):

JmillsExpensify:
Looks like we're back to Neil for final review.

src/pages/iou/request/step/IOURequestStepDistanceGPS/index.native.tsx
neil-marcellini: Ok cool. I do not see any reason we need to have updates every x seconds, I think we just want the map to stay centered on the user.

src/components/MapView/GPSMapView.tsx
```diff
@@ -0,0 +1,314 @@
...
+    const styles = useThemeStyles();
+    const theme = useTheme();
+    const expensifyIcons = useMemoizedLazyExpensifyIcons(['Crosshair', 'MapCurrentLocationPuck', 'MapCurrentLocation']);
+    const isAccessTokenSet = useAccessToken({accessToken});
+
+    const cameraRef = useRef<Mapbox.Camera>(null);
+
+    const [foregroundLocationPermissionsGranted, setForegroundLocationPermissionsGranted] = useState<boolean | null>(null);
+    // Request foreground location permissions if not granted yet to determine if we can use followUserLocation prop on the map camera
+    useFocusEffect(() => {
```
neil-marcellini, line 50: Please wrap this in useCallback so it doesn't run on every render while the screen is focused, as per [the docs](https://reactnavigation.org/docs/use-focus-effect/).

neil-marcellini: Also please include a blank line above the comment as per our style rules.

src/pages/iou/request/step/IOURequestStepDistanceGPS/index.native.tsx
GCyganek: @neil-marcellini it currently works this way, I changed the approach to use Mapbox's internal location engine via `followUserLocation` prop which gives us location updates every 1sec and smooth transitions between the updates. I described everything in the PR description

neil-marcellini (changes requested):
Here is a very small review from Friday. I'll review more today.

src/components/MapView/GPSMapView.tsx
GCyganek: is the `useCallback` really needed though if the component is auto-memoized by React Compiler?

See, the compiler output suggests that the callback is only recreated when `isOffline` changes which is the same if I used `useCallback` with `isOffline` in the dependency array:

Screenshot 2026-06-22 at 16 11 54

I'll add the blank line above the comment

neil-marcellini: Oh ok wow, I did not know about that. It should be fine then.

src/components/MapView/GPSMapView.tsx
```diff
@@ -0,0 +1,315 @@
...
+                            onUpdate={onUserLocationUpdate}
+                            visible={false}
+                        />
+                    </>
+                )}
+
+                <GPSWaypointLayer
+                    waypoints={waypoints}
+                    // To ensure that waypoints are shown below the location puck we need to pass belowLayerID prop
+                    // Android does not support dynamic belowLayerID prop change, so we pass key to remount this component witch belowLayerID change
```
neil-marcellini, line 279: NAB:typo: "witch" 🧙‍♀️

src/components/MapView/useAnimatedTrailingDirectionCoordinate.ts
neil-marcellini, line 1: It feels a little bit backwards to me to call this "trailing". It should be called "leading", in my opinion, because as you are tracking over time, we're recording a series of coordinates. The leading edge of those coordinates is the one at the front in the direction that we are traveling in, which is what we are animating here.

src/components/MapView/utils.ts
```diff
@@ -112,6 +115,17 @@ function closestPointOnSegment(point: LngLatLiteral, startPoint: Coordinate, end
...
 }
 
+function areCoordinatesEqual(coordinate1: Coordinate | undefined, coordinate2: Coordinate | undefined) {
+    if (!coordinate1 || !coordinate2) {
+        return false;
+    }
+    return coordinate1[0] === coordinate2[0] && coordinate1[1] === coordinate2[1];
+}
+
+function interpolateCoordinate(start: Coordinate, end: Coordinate, progress: number): Coordinate {
```
neil-marcellini, line 125: NAB: We should probably name this something like "simple interpolate coordinate" or add a comment because it is using linear interpolation rather than a [great circle](https://en.wikipedia.org/wiki/Great-circle_navigation) path. It will be fine for small distances and how we are using it today, but if you used this across large distances it would not follow a straight path since the earth is spherical.

patches/@rnmapbox/maps/details.md
```diff
@@ -0,0 +1,14 @@
...
+- Solution: Adds an opt-in `followUserLocationUseImmediateTransition` prop to `Camera`. When `true`, entering follow mode uses `makeImmediateViewportTransition()` so the camera snaps straight to the follow state. When `false` (default), upstream behavior is preserved and the default viewport transition animates. `GPSMapView` passes the prop only on initial map load when follow is active with no route; later re-engagements (center button, follow after clearing a route) pass `false` so the camera animates back to the user.
+- Files changed:
+  - `src/specs/RNMBXCameraNativeComponent.ts`, `src/components/Camera.tsx`, and `lib/typescript/src/components/Camera.d.ts` — new prop
+  - iOS — `ios/RNMBX/RNMBXCamera.swift`, `ios/RNMBX/RNMBXCameraComponentView.mm`, `ios/RNMBX/RNMBXCameraViewManager.m`
+  - Android — `android/src/main/java/com/rnmapbox/rnmbx/components/camera/RNMBXCamera.kt` and `RNMBXCameraManager.kt`
+
+- Upstream PR/issue: 🛑
```
neil-marcellini, line 12: NAB: Why no upstream PR or issue?

neil-marcellini (approved):
The code looks good. Great work! I have a few non blocking items that could be improved in a quick follow up.

unzig21:
Deploy Blocker https://github.com/Expensify/App/issues/94406 was identified to be related to this PR.

neil-marcellini:
Noting that my NAB comments were resolved here. Thanks.

- [x] https://github.com/Expensify/App/pull/94284

src/components/MapView/GPSMapView.tsx
```diff
@@ -0,0 +1,315 @@
...
+    const waypointsBounds = getWaypointBounds();
+
+    const onUserLocationUpdate = (update: Mapbox.Location) => {
+        const coords = update.coords;
+        setLastLocation({longitude: coords.longitude, latitude: coords.latitude});
+    };
+
+    // On first render LocationPuck layer is not ready to be used as belowLayerID prop
+    // for GPSDirection and GPSWaypointLayer, so we need to wait for the layer to be ready
+    const onDidFinishRenderingFrameFully = () => {
```
dukenv0307, line 169: Coming from https://github.com/Expensify/App/issues/98965, `onDidFinishRenderingFrameFully` pointed layers at a not-yet-mounted layer, which iOS Mapbox silently drops. we should use always-mounted layer

src/libs/GPSDraftDetailsUtils.ts
```diff
@@ -116,7 +116,7 @@ async function stopGpsTrip(isOffline: boolean, gpsPoints: GPSPoint[][], skipLast
         return;
     }
 
-    if (isLastSegmentEmptyOrHasOnlyOnePoint(lastSegment) && gpsPoints.length > 1) {
+    if (isLastSegmentEmptyOrHasOnlyOnePoint(lastSegment)) {
```
mkhutornyi: `&& gpsPoints.length > 1` condition was accidentally removed here, which was added in #90237 to fix [Track distance - GPS - Stop GPS tracking and Resume not displayed if tap on Stop GPS tracking](https://github.com/Expensify/App/issues/90103).

So this caused regression again - [Android - GPS - "Save" and "Resume" buttons not present after tapping "Stop" without moving](https://github.com/Expensify/App/issues/100938)

Issue #69562, opened 2025-08-29
[$250] Fix localization: dates not being localized correctly


iwiznia:
# Problem

We've identified several places where we are not using our localization properly, leading to using formats that are not used in the selected locale.

For example this is with spanish language:
<img width="314" height="55" alt="Image" src="https://github.com/user-attachments/assets/b8897a0a-f9d9-47c4-80cd-80f1e047f941" />

In spanish should be something like `19 ago, 2025 a las 12:02PM` or something like that.

# Solution

Remove any code that uses a hardcoded date format (not based on the locale) as long as the date is used to display information to users. Some examples (but please check if there are more places):
- https://github.com/Expensify/App/blob/e091028680e8e21729421dee3ec0547880042f59/src/libs/DateUtils.ts#L252-L255
- https://github.com/Expensify/App/blob/0db6a75d619ac9d6280cf28ec0d2138f2b0ef8b5/src/CONST/index.ts#L432-L453

Replace all of then with the standardized [`dateStyle` and `timeStyle`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat) (using one of the values "full", "long", "medium", and "short"), but let's inspect each case to make sure it makes sense to replace it and with which format exactly we want to replace it.

Additionally, change [this](https://github.com/Expensify/App/blob/e091028680e8e21729421dee3ec0547880042f59/src/libs/DateUtils.ts#L112) to use `getWeekInfo().firstDay` of the current locale and ensure the calendar component works properly, showing the correct start day according to the locale (FYI 99% sure [this](https://github.com/Expensify/App/blob/e091028680e8e21729421dee3ec0547880042f59/src/libs/DateUtils.ts#L121)  is wrong and 6 should be 7)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021961470094099753846</li>
        <li>Upwork Job ID: 1961470094099753846</li>
        <li>Last Price Increase: 2026-05-05</li>
<li>Automatic offers: </li>
<ul>
<li>situchan | Contributor | 110344669</li>
</ul></li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @TaduJR</details>

iwiznia:
[...]

Issue #80024, opened 2026-01-20
[Due for payment 2026-01-29] Date - Wrong date (moved back by one day) displayed in date field


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** v9.3.5-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** Exploratory around https://github.com/Expensify/App/pull/78251
**Email or phone of affected tester (no customers):** kiatest.78+ck118261@gmail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** Windows 11/Chrome, iPhone 16/iOS 18.7.1
**App Component:** User Settings

## Action Performed:
Precondition: User is located in US & current date is 01/20/2026

1. Navigate to staging.new.expensify.com
2. Go to Profile >> Status >> Under Clear after, select Custom.
3. Open the Date picker.
4. Select a different date 
5. Observe date in date field is moved back by one day
6. Scroll down to Private section >> Date of birth field
7. Select any year/date
8. Observe  date in date field is moved back by one day
9. Save it & observe incorrect date get saved
10. Repeat steps in Company cards page assigning card to user 
11. Select transaction start date & observe incorrect date get displayed in date field

## Expected Result:
Same date selected in date picker should reflect in date field. Date should not be moved back by one day in date field

## Actual Result:
Wrong date (moved back by one day) displayed in date field

## Workaround:
Unknown

## Platforms:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [x] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/f0b011ac-1198-4b32-85c0-025591b7502c

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @garrettmknight</details>

Beamanator:
[...]

Issue #80013, opened 2026-01-20
[Due for payment 2026-01-29] Expense - App crashes when reaching merge confirm page


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.5-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** No
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://github.com/Expensify/App/pull/78251
**Email or phone of affected tester (no customers):** sdjiosnodjosijnosd@gmail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** Mac 26.1 / Chrome
**App Component:** Money Requests

## Action Performed:
1. Go to staging.new.expensify.com
2. Go to workspace chat.
3. Create two expenses.
4. Open the expense report.
5. Select both expenses via checkbox.
6. Click dropdown button > Merge.
7. Select any details > Next.

## Expected Result:
App will not crash when reaching merge confirm page.

## Actual Result:
App crashes when reaching merge confirm page.

## Workaround:
Unknown

## Platforms:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/527268ca-e6b4-445f-b0f5-ee21694cc940

[logs 1.txt](https://github.com/user-attachments/files/24745428/logs.1.txt)

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @stephanieelliott</details>

iwiznia:
@sofi-a can you look into this? It seems to have been caused by your PR... otherwise I can fully revert it

Issue #80011, opened 2026-01-20
[Due for payment 2026-01-29] Bank - Date of birth field placeholder shows YYYY-MM-DD when the new format is MM/DD/YYYY


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.5-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** N/A - new feature, doesn't exist in prod
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://github.com/Expensify/App/pull/78251
**Email or phone of affected tester (no customers):** applausetester+sdjois2jd@applause.expensifail.com
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** Mac 26.1 / Chrome
**App Component:** Workspace Settings

## Action Performed:
1. Go to staging.new.expensify.com
2. Go to Workspaces > Workflows.
3. Click Add bank account.
4. Click Connect manually.
5. Proceed to Personal info - Date of birth step.
6. Select a birth date.

## Expected Result:
Date of birth field placeholder should be MM/DD/YYYY.

## Actual Result:
Date of birth field placeholder shows YYYY-MM-DD when the new format is MM/DD/YYYY.

## Workaround:
Unknown

## Platforms:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/2636ff29-4872-452e-ad0d-ee0e282250df

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @heyjennahay</details>

Beamanator:
Looks like this same as in prod now

PR #91569, opened 2026-05-25
fix: localization dates not being localized correctly

TaduJR:
### Explanation of Change

Dates were translated word by word but kept the English layout. A Spanish user saw `ago 19, 2025 a 12:02 PM`, but Spanish writes `19 ago 2025` and uses a 24-hour clock. The order, separators and clock style are part of the language, just like the month name.

Every date shown in the app is now formatted with the built-in `Intl` API for the user's language, and the language is passed in directly instead of read from a global setting. Dates saved to Onyx or sent to the API keep their fixed format, except the per diem merchant (see below).

What changes for users:

- **Dates and times** in chat, expenses, search, trips, wallet, subscription, workspace settings and schedule call follow the language's order and clock. For example `19 ago 2025 a 14:30` in Spanish, `19.08.2025` in German.
- **Relative times** read `2 hours ago` or `hace 2 horas`, and `now` for anything under a minute. iOS and Android get a polyfill, because Hermes doesn't support `Intl.RelativeTimeFormat` and these were blank.
- **The calendar** starts the week on the language's first day and shows translated weekday letters and month names. Date fields show a matching placeholder: `MM/DD/YYYY`, `DD.MM.YYYY`, `YYYY/MM/DD`.
- **Switching language** no longer shows translation keys or the old language. The splash screen waits for the language to load and falls back to English if the download fails.
- **Per diem** expenses can now be saved in German, Spanish and Japanese. The time was written with that language's AM/PM word and read back as English, so the "at least one minute in the future" check always failed.
- **The per diem merchant** is now saved in English (`Berlin, Aug 19, 2025 - Aug 20, 2025`), so every viewer's copy is the same. Each reader sees the dates in their own language. Expenses saved in another language before this change still display correctly.

The week start comes from a fixed table rather than `Intl.Locale.getWeekInfo()`, because Firefox and older Hermes builds don't have it. A test checks that the table matches `getWeekInfo()` for every supported language.

A lint rule now flags date-fns formats that render differently per language, so hardcoded display formats can't come back.

### Fixed Issues

$ https://github.com/Expensify/App/issues/69562
PROPOSAL: https://github.com/Expensify/App/issues/69562#issuecomment-4382709811

### 1. Everyday date surfaces

**1.1 Home screen greeting**
1. Open the **Home** tab and read the date above the greeting.
2. Expected:
   - English `Wednesday, September 9`
   - Spanish `miércoles, 9 de septiembre`
   - German `Mittwoch, 9. September`
3. Spanish puts the day before the month and inserts `de`.

**1.2 Chat message timestamps**
1. Open a chat with messages from today, earlier this week, and last year.
2. English: `Today at 2:30 PM`, `Aug 19 at 2:30 PM`, `Aug 19, 2025 at 2:30 PM`.
3. Spanish: `Hoy a las 14:30`, `19 ago a 14:30`, `19 ago 2025 a 14:30`.
4. German: `Heute um 14:30`, `19. Aug. um 14:30`, `19.08.2025 um 14:30`.
5. No timestamp is blank, and none ends with a dangling word (`Today at` with no time).

> The "this week" format boundary depends on the language's first day of the week, so a message near the weekend may legitimately switch between the short and full form when you change language.

**1.3 Expense list and detail**
1. Open the expense list with expenses from this year and last year.
2. Expected in the list:
   - English `Sep 3` and `Aug 19, 2025`
   - Spanish `3 sept`
   - Japanese `8月19日`
3. Open the expense. The **Date** row is localized and carries the year: English `Sep 3, 2026`, Spanish `3 sept 2026`.
4. It must **never** read `2026-09-03`.

**1.4 Relative times (must be done on iOS and Android)**
1. **Account > Security > Device management**. Rows read `2 hours ago` or `3 days ago`, never blank.
2. Expected:
   - Spanish `hace 2 horas`
   - German `vor 2 Stunden`
3. If available, **Workspace > Accounting** "Last synced …" must not be blank in either language.

**1.5 Travel and trips**
1. Open a trip in chat (flight, hotel, car or train).
2. Check departure and arrival dates, the range on the card, and any cancellation deadline.
3. No dangling dash or separator with nothing beside it (for example `- 10:30`).
4. A range inside one month reads `Aug 17-20` in English and `17-20 ago` in Spanish. The month must not sit between the two days.

**1.6 Other date surfaces.** Confirm each reads in the selected language:
1. **Account > Wallet**: card details and statements
2. **Account > Subscription**: billing and renewal dates
3. **Workspace > Workflows**: monthly submission day ("the 15th of each month")
4. **Workspace > Expensify Card > Settings**: settlement date
5. **Another user's profile**: their local time, and confirm it updates on its own after a minute
6. **Schedule a call**: available slots and the confirmation screen
7. **An expense report's next-step message**: any date inside the sentence

---

### 2. Calendar and date fields

**2.1 First day of the week**
1. **Account > Profile > Date of birth** (under the **Private** section). Tap the field to open the picker.
2. **English starts on Monday** (`M T W T F S S`). This is deliberate, not a bug.
3. Spanish, German, French: also Monday. **Japanese and Portuguese (BR): Sunday.**
4. Weekday letters are translated **and ordered from that language's first day**:
   - Spanish (Monday first): `L M X J V S D`
   - Japanese (Sunday first): `日 月 火 水 木 金 土`
   - The Spanish `D` sits at the **end**. `D` appearing first would be the bug.

**2.2 Month names**
1. In the same picker, open the month list.
2. Expected:
   - Spanish `Enero, Febrero, Marzo…`
   - French `Janvier, Février, Mars…`
   - German `Januar, Februar, März…`

**2.3 Field placeholder**
1. On an empty **Date of birth** field: English `MM/DD/YYYY`, German `DD.MM.YYYY`, Japanese `YYYY/MM/DD`, Spanish/French/Italian/Portuguese `DD/MM/YYYY`, Dutch `DD-MM-YYYY`.
2. Pick a date. The filled value uses the **same order and separators** as the placeholder did.

---

### 3. Per diem
*Needs a workspace with per diem enabled and rates uploaded.*

**3.1 A German user can save a per diem**

1. Set the language to **German**.
2. Create a per diem expense and pick any destination.
3. Set the **start date to tomorrow** and the end date to the day after.
4. Change the start time to `08:00` and the end time to `14:00`. Remember the picker uses AM/PM even in German, so choose `08:00 AM` and `02:00 PM`.
5. Save.
6. **It must save.** Failure: *"Bitte wähle eine Zeit, die mindestens eine Minute in der Zukunft liegt."*
7. Repeat in **Spanish** and **Japanese**.

**3.2 Per diem times and dates read in your language**
1. On the per diem time screen, the time rows read `12:00 AM` in English and **`00:00` in German** (24-hour).
2. The date fields read `09/09/2026` in English and **`09.09.2026` in German**.

**3.3 The merchant reads in your language**
1. In **Spanish**, look at a saved per diem expense in three places: the chat preview card, the expense list row, and the expense's own **Merchant** row.
2. All three read `Berlín, 19 ago 2025 - 20 ago 2025`, not `Aug 19, 2025 - Aug 20, 2025`.

**3.4 It survives a language switch**
1. Create a per diem expense in English, view it in Spanish, then switch back.
2. It reads correctly each time, and a location containing commas (for example `San Francisco, California, USA`) keeps all of its parts.

### Offline tests

Same as tests

### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/f78bd1d1-6372-4a63-9442-51e0287c1156

</details>

<details>

https://github.com/user-attachments/assets/2292b6c5-bde0-40f8-a693-a9c10ff737cd

</details>

<details>

https://github.com/user-attachments/assets/1faf694d-72b7-4e63-8472-a7d066e8a12b

</details>

<details>

https://github.com/user-attachments/assets/fa48dbcb-b24d-46c9-a981-0870b00928fc

</details>

<details>

https://github.com/user-attachments/assets/0a54777c-90c1-49aa-92ff-f47c0495ceb5

</details>


situchan:
@TaduJR you're missing native platform tests.

Please test various timezones (US timezone UTC-X, Asia timezone UTC+X)

situchan:
Regression test cases:
- https://github.com/Expensify/App/pull/69945#discussion_r2586355819
- https://github.com/Expensify/App/pull/69945#issuecomment-3608804531
- https://github.com/Expensify/App/issues/76643
- https://github.com/Expensify/App/issues/76646
- https://github.com/Expensify/App/issues/80011
- https://github.com/Expensify/App/issues/80013
- https://github.com/Expensify/App/issues/80024

src/components/DatePicker/CalendarPicker/generateMonthMatrix.ts
```diff
@@ -20,7 +21,7 @@ export default function generateMonthMatrix(year: number, month: number) {
     }
 
     // Get the week day for the end of week
-    const weekEndsOn = DateUtils.getWeekEndsOn();
+    const weekEndsOn = DateUtils.getWeekEndsOn(locale);
```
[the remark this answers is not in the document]
TaduJR: Invalid.

heyjennahay (commented):
Product review not required

situchan:
Please check ^

TaduJR:
@codex review

TaduJR:
@codex review

TaduJR:
@codex review

situchan:
Was [this](https://github.com/Expensify/App/pull/91569#issuecomment-4922656868) addressed too?
Typescript failing

TaduJR:
> Can we split into smaller PRs to avoid being reverted?

@situchan 

No we can't they are tightly coupled.

TaduJR:
@codex review

src/libs/DateUtils.ts
```diff
@@ -794,21 +1023,27 @@ function getLastBusinessDayOfMonth(inputDate: Date): number {
...
         // Dates are from the same day
-        return format(date1, 'MMM d', {locale: dateFnsLocale});
+        return formatIntl(locale, 'MONTH_DAY', date1);
     }
     if (isSameMonth(date1, date2)) {
         // Dates in the same month and year, differ by days
-        return `${format(date1, 'MMM d', {locale: dateFnsLocale})}-${format(date2, 'd', {locale: dateFnsLocale})}`;
+        const monthDay = formatIntl(locale, 'MONTH_DAY', date1);
+        const dayOnly = formatIntl(locale, 'DAY_ONLY', date2);
+        return monthDay && dayOnly ? `${monthDay}-${dayOnly}` : '';
```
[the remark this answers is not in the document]
TaduJR: Addressed

TaduJR:
@codex review

TaduJR:
@codex review

situchan:
@MelvinBot review

TaduJR:
@situchan 

Could you request review from Melvin for both deep review and regression review.

Thanks!

TaduJR:
Ahh mistakenly missing during git add. Re-pusing

TaduJR:
@codex review

TaduJR:
@codex review

TaduJR:
@situchan 

Ready for review. Would it be possible to push this over the finish line quickly?

I am not sure to be able to keep up with the conflicts 😢.

situchan:
@MelvinBot find regressions

src/languages/IntlStore.ts
```diff
@@ -251,39 +278,62 @@ class IntlStore {
...
         return retryDynamicImport(loaderPromise, `${LOCALE_RETRY_KEY_PREFIX}${locale}`)
             .then(() => {
-                this.currentLocale = locale;
-                // Set the default date-fns locale
-                const dateUtilsLocale = this.dateUtilsCache.get(locale);
-                if (dateUtilsLocale) {
-                    setDefaultOptions({locale: dateUtilsLocale});
+                // Superseded, so let the newer call commit the locale. Still notify: the cache grew either way.
+                if (IntlStore.loadToken !== token) {
+                    IntlStore.notifyListeners();
```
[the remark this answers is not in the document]
TaduJR: Addressed.

TaduJR:
@codex review

src/languages/IntlStore.ts
```diff
@@ -251,39 +278,62 @@ class IntlStore {
...
                 Sentry.captureException(error, {
                     fingerprint: ['locale-load-failed'],
                     extra: {locale},
                 });
+                // Only stamp the span this call started. A superseded load shares the span id, so it would blame the successor.
+                if (localeSpan && IntlStore.loadToken === token) {
+                    endSpanWithAttributes(CONST.TELEMETRY.SPAN_LOCALE.TRANSLATIONS_LOAD, {[CONST.TELEMETRY.ATTRIBUTE_FAILED]: true});
+                }
+                // The splash gate below only lifts on a non-empty cache, so without this fallback the app hangs on the boot splash forever.
+                if (IntlStore.loadToken !== token || IntlStore.cache.size > 0 || locale === LOCALES.DEFAULT) {
```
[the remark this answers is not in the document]
TaduJR: Coming from merge main. Resolved.

TaduJR:
@codex review

src/components/HTMLEngineProvider/HTMLRenderers/VictoryChartRenderer/components/VictoryChartLabel.tsx
```diff
@@ -35,7 +36,7 @@ type ProcessedLine = {
 function VictoryChartLabel({x, y, text, color, fontSize, fontWeight, fontFamily, fontStyle, lineHeight, textAnchor = 'start', verticalAnchor = 'middle', timezone}: VictoryChartLabelsProps) {
     const typefaces = useChartTypefaces();
     const theme = useTheme();
-    const displayText = getLocalizedVictoryChartLabelText(text, timezone);
+    const displayText = getLocalizedVictoryChartLabelText(text, timezone, IntlStore.getCurrentLocale());
```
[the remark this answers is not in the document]
TaduJR: Invalid. But made it better so it knows its explicit.

TaduJR:
@codex review

situchan:
Please fix conflict before going off for weekend. Then I'll review during weekend

src/libs/DateUtils.ts
```diff
@@ -224,10 +333,64 @@ function datetimeToCalendarTime(locale: Locale | undefined, datetime: string, cu
...
+        formatter = null;
+    }
+    relativeTimeFormatCache.set(locale, formatter);
+    return formatter;
+}
+
+function formatRelative(locale: Locale, date: Date, now: Date): string {
+    const rtf = getRelativeTimeFormat(locale);
+    if (!rtf) {
+        return '';
```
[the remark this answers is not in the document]
TaduJR: I don't think it can actually happen in a browser we support.

We already call `new Intl.ListFormat` on web with no polyfill and no try/catch (`src/libs/Localize/index.ts`). It shipped later than `Intl.RelativeTimeFormat` in every browser Firefox 78 vs 65, Safari 14.1 vs 14.0 so any browser missing `RelativeTimeFormat` is also missing `ListFormat`, where we throw outright instead of just showing a blank date. That's already true on main, so this PR doesn't make anything worse.

Keeping `formatDistance` as a fallback isn't free either: it only translates if you pass it `getDateFnsLocale`, so we'd have to keep the eleven date-fns locale imports and the global date-fns default that moving to `Intl.RelativeTimeFormat` lets us delete.

That said, it did point at something real but I've added a guard to the `ListFormat` call so it falls back to a comma-separated list instead of throwing, with a test for it. May be that is better 🤷‍♂️.

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot @codex review

TaduJR:
> Nit: one now-dead cache input

Nit breaks the code. Invalid suggestion.

Resolved conflicts and merge main inconsistencies.

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot @codex review

TaduJR:
@MelvinBot Please Hide all your review comments as resolved since they are all addressed by hiding them

situchan:
In review & testing

TaduJR:
@MelvinBot @codex review

situchan:
Conflicts again

TaduJR:
@MelvinBot review

src/pages/home/TimeSensitiveSection/items/PayOverdueInvoice.tsx
```diff
@@ -22,11 +22,11 @@ type PayOverdueInvoiceProps = {
...
 
 function PayOverdueInvoice({gracePeriodEndUnixSeconds, isOverdue = false}: PayOverdueInvoiceProps) {
-    const {translate, dateFnsLocale} = useLocalize();
+    const {translate, preferredLocale} = useLocalize();
     const icons = useMemoizedLazyExpensifyIcons(['InvoiceGeneric']);
 
-    // formatWithUTCTimeZone needs ISO 8601. toUTCString (RFC 1123) would silently format as empty
-    const dueDate = DateUtils.formatWithUTCTimeZone(fromUnixTime(gracePeriodEndUnixSeconds).toISOString(), CONST.DATE.MONTH_DAY_YEAR_FORMAT, dateFnsLocale);
+    // An ISO string rather than a `Date`, which `formatInUTCToMedium` would re-read as the device's local calendar day.
+    const dueDate = DateUtils.formatInUTCToMedium(fromUnixTime(gracePeriodEndUnixSeconds).toISOString(), preferredLocale);
```
situchan: NAB: **Overdue-invoice due date lost its full month name**

`main` used `formatWithUTCTimeZone(…, MONTH_DAY_YEAR_FORMAT)` = `'MMMM d, yyyy'` (full month).
The PR swapped it to `formatInUTCToMedium` (**abbreviated**). UTC anchoring is preserved — only the month presentation shrank.
Every other full-month UTC site in this PR was preserved with `formatInUTCToLong`, so this reads as an oversight.

- Before: "…pay before **January 15, 2026**" → After: "…pay before **Jan 15, 2026**"
- Fix: `formatInUTCToMedium` → `formatInUTCToLong`.
- Repro:
   * Account with an **invoice due soon** (billing grace period active, not overdue) so the Inbox "Time sensitive" card shows *Pay overdue invoice*.
   * Read the "pay before {date}" title.
   * Expected full month "January 15, 2026"; actual abbreviated "Jan 15, 2026".

src/libs/DateUtils.ts
```diff
@@ -244,33 +411,25 @@ function datetimeToRelative(locale: Locale | undefined, datetime: string, curren
...
  */
 function getZoneAbbreviation(datetime: string | Date, selectedTimezone: SelectedTimezone): string {
-    const abbreviation = formatInTimeZone(datetime, selectedTimezone, 'zzz');
+    const abbreviation = formatInTimeZoneWithFallback(datetime, selectedTimezone, 'zzz');
     if (abbreviation === 'GMT') {
-        return formatInTimeZone(datetime, selectedTimezone, 'O');
+        return formatInTimeZoneWithFallback(datetime, selectedTimezone, 'O');
     }
     return abbreviation;
 }
```
situchan: NAB: **Timezone abbreviation now follows the device language, not the app language**

`getZoneAbbreviation` calls `formatInTimeZoneWithFallback(..., 'zzz')` with no `locale` option.
date-fns-tz's `z` formatter resolves its locale as `options.locale ?? getDefaultOptions().locale` (`date-fns-tz/_lib/tzIntlTimeZoneName`), and when that is `undefined` it constructs `new Intl.DateTimeFormat(undefined, {timeZoneName: 'short'})` — the **runtime/device** locale.
Previously `setDefaultOptions` pinned it to the app language.

So in a PR whose whole point is that dates follow the app language, this one string now follows the OS instead.
Visible in [`AutoUpdateTime`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/components/AutoUpdateTime.tsx#L35), [`OnboardingHelpDropdownButton`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/components/OnboardingHelpDropdownButton.tsx) and [`ScheduleCallConfirmationPage`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/pages/ScheduleCall/ScheduleCallConfirmationPage.tsx).

**Repro:**
1. Set the **device/OS** language to German and the region to Germany; set the device timezone to Europe/Berlin.
2. In the app, set Account → Preferences → Language → **English**.
3. Open another user's profile (or your own details page) and read the **Local time** row.
4. Expected (and what `main` shows): `2:30 PM CEST`. Actual: `2:30 PM MESZ` — the abbreviation is German while the rest of the row is English.
5. Invert it: OS in English, app language German → the row reads `14:30 CEST`, where `main` gave `MESZ`.

situchan (commented):
NAB: **CSV import silently drops rows whose dates use non-English month names**

[`src/libs/CSVDateUtils.ts#L41`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/libs/CSVDateUtils.ts#L41) calls `parse(trimmedInput, dateFormat, new Date())` with no `locale`.
Four of the [`CSV_DATE_FORMATS`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/libs/CSVDateUtils.ts#L16-L19) are month-name formats (`MMM d, yyyy`, `MMMM d, yyyy`, `d MMM yyyy`, `dd MMM yyyy`).
On `main`, `setDefaultOptions({locale: de})` made those parse German month names; now they only parse English.

`parseCSVDate` returns `null`, and both consumers `continue` past the row with no error surfaced — [`ImportTransactions.ts#L161-L166`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/libs/actions/ImportTransactions.ts#L161-L166) and [`CompanyCards.ts#L142-L149`](https://github.com/Expensify/App/blob/3a5a504e66aeee4191b887f2c2f9b96173b6b46b/src/libs/actions/CompanyCards.ts#L142-L149).

`Mär`, `Okt`, `Dez`, `Mai` (de), `mrt`, `okt` (nl), `mar`, `paź`, `gru` (pl) all stop matching. Spanish/French/Italian mostly survive by accident, because their abbreviations share the English three-letter prefix.

**Repro:**
1. Account → Preferences → Language → **Deutsch**.
2. Build a CSV with a header row and one data row whose date cell is `2 Mär 2025`, plus merchant and amount columns.
3. Account → Wallet → **Import transactions** (or Workspace → Company cards → import a CSV feed).
4. Upload the file and map Date / Merchant / Amount.
5. Expected: the row imports as 2025-03-02. Actual: the row is silently missing from the imported list — no error, no count mismatch shown.
6. Repeat the same steps on `main` in German: it imports.

situchan (approved):
I performed full regression test and no major blockers found.
But let's still be on a look out for any issues QA team will be reporting.

situchan:
@TaduJR please check https://github.com/Expensify/App/pull/91569#issuecomment-5755531759, https://github.com/Expensify/App/pull/91569#pullrequestreview-5263206502 and address valid bugs.

src/languages/localeDayOfMonthMap.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Separate from `localeOrdinalMap` because a date is not a rank: Japanese dates read `15日` where the rank is `第15`.
+ * Each entry renders the day alone. Any surrounding preposition or unit belongs to the sentence in `<locale>.ts`.
+ * English is absent because its dates read as the rank (`15th`), which `toLocaleOrdinal` already renders.
+ */
+import type {Locale} from '@src/CONST/LOCALES';
+import {LOCALES} from '@src/CONST/LOCALES';
+
+type DayOfMonthRenderer = (day: number) => string;
+
+const localeDayOfMonthMap: Partial<Record<Locale, DayOfMonthRenderer>> = {
```
iwiznia, line 11: I don't get why this is here instead of in the language files

src/libs/Localize/index.ts
```diff
@@ -108,12 +128,21 @@ const memoizedGetTranslatedPhrase = memoize(getTranslatedPhrase, {
...
  *
- * @param [locale] eg 'en', 'es'
+ * @param locale eg 'en', 'es'. Always defined, since LocaleContextProvider and IntlStore.getCurrentLocale guarantee it.
  * @param [parameters] Parameters to supply if the phrase is a template literal.
  */
-function translate<TPath extends TranslationPaths>(locale: Locale | undefined, path: TPath, ...parameters: TranslationParameters<TPath>): string {
-    if (!locale) {
-        // If no language is provided, return the path as is
+function translate<TPath extends TranslationPaths>(locale: Locale, path: TPath, ...parameters: TranslationParameters<TPath>): string {
+    if (!IntlStore.hasLocale(locale)) {
```
iwiznia, line 135: In what case we would get inside this if?

src/libs/Localize/index.ts
```diff
@@ -145,20 +173,29 @@ function translateLocal<TPath extends TranslationPaths>(phrase: TPath, ...parame
     return translate(currentLocale, phrase, ...parameters);
 }
 
-function getPreferredListFormat(): Intl.ListFormat {
-    return memoizedCreateConjunctionListFormat(IntlStore.getCurrentLocale() ?? CONST.LOCALES.DEFAULT);
+/** The separator for runtimes with no `Intl.ListFormat`. Dropping the conjunction is cosmetic, dropping the items is not. */
+const LIST_FALLBACK_SEPARATOR = ', ';
```
iwiznia, line 177: Not sure I get why we need this now but not before

src/libs/DateUtils.ts
```diff
@@ -43,73 +40,187 @@ import {
...
+    return tz in timezoneNewToBackwardMap;
 }
 
-/**
- * Get the day of the week that the week starts on
- */
-function getWeekStartsOn(): WeekDay {
-    return CONST.WEEK_STARTS_ON;
+/** A Wednesday in UTC, used where only the locale's own conventions matter and the instant must not vary by run. */
+const LOCALE_PROBE_DATE = new Date(Date.UTC(2023, 0, 4));
```
iwiznia, line 69: I don't get why we need this hardcoded date

src/libs/DateUtils.ts
```diff
@@ -43,73 +40,187 @@ import {
...
+    const formatter = getIntlDateTimeFormat(locale, formatKey, timeZone);
+    return formatter?.formatToParts(date).find((part) => part.type === partType)?.value ?? '';
+}
+
+/**
+ * CLDR week starts for the shipped locales, read from here rather than `Intl.Locale.getWeekInfo` so engines without it agree
+ * with those that have it. `en` is the exception: CLDR resolves it to en-US's Sunday, which would move the calendar for every
+ * existing English user.
+ */
+const WEEK_STARTS_ON_BY_LOCALE: Readonly<Record<Locale, WeekDay>> = {
```
iwiznia, line 170: > read from here rather than `Intl.Locale.getWeekInfo` so engines without it 

Which engines do not have it?

situchan:
@TaduJR no need to care about edge cases. Fixing edge cases might cause another unexpected regressions.

Please check these potential bugs from latest commits:

**CSV rows with localized month names dropped**: fixed, but the fix rewrote the parser and introduced two new regressions, below.
`parseCSVDate` now takes the app language and translates month names to English before parsing.

---

### New 1: CSV dates with two-digit years import as year 0025

[`parseDateValue`](https://github.com/Expensify/App/blob/cae54f8a37530b4d623cfc5e4d99971ac85cff59/src/libs/CSVDateUtils.ts#L73-L86) now tries the date-fns formats first and the browser's own date parser last. `main` did it the other way round. None of the formats has a two-digit year pattern, and date-fns `yyyy` accepts `25` as a valid year 25. Checked with V8:

| Cell | `main` (browser parser first) | This PR (formats first) |
|---|---|---|
| `3/4/25` | `2025-03-04` | `0025-03-04` |
| `12/31/24` | `2024-12-31` | `0024-12-31` |

This affects every language, English included. `3/4/25` is Excel's default US short-date export. The bad value is not dropped: Import Transactions creates the expense with that date, and the company-card path [writes it back into the row](https://github.com/Expensify/App/blob/cae54f8a37530b4d623cfc5e4d99971ac85cff59/src/libs/actions/CompanyCards.ts#L155-L157) that is sent to the backend. On iOS/Android `main` already produced 0025, because Hermes' date parser rejects this format, so the regression is web, desktop and mWeb.

**Repro:**
1. Keep the app language as English, on web.
2. Create a CSV with a header row and the row `3/4/25,Coffee,4.50` (Date, Merchant, Amount).
3. Account → Wallet → **Import transactions**, upload it and map the columns.
4. Open the imported expense.
5. Expected (and what `main` does): Mar 4, 2025. Actual: the date is Mar 4, 0025.

---

### New 2: US-format CSVs get day and month swapped when the app language isn't English, Japanese or Chinese

[`getDateFormats`](https://github.com/Expensify/App/blob/cae54f8a37530b4d623cfc5e4d99971ac85cff59/src/libs/CSVDateUtils.ts#L15-L23) now tries [day-first formats first](https://github.com/Expensify/App/blob/cae54f8a37530b4d623cfc5e4d99971ac85cff59/src/libs/CSVDateUtils.ts#L15) whenever `isDayBeforeMonth(locale)` is true. That covers es, fr, de, it, nl, pl, pt-BR and el. But the order of fields in a CSV is set by the bank that exported it, not by the uploader's app language. `main` always read slash dates month-first, so a US bank file was read correctly for every user.

With this PR, a Spanish-language user in the US uploading a Chase or Amex export gets every row whose day is 12 or less silently swapped. Rows with a day above 12 still parse month-first, because the day-first read fails and it falls through. So one file ends up with mixed interpretations, and nothing warns the user.

**Repro:**
1. Account → Preferences → Language → **Español**.
2. Create a CSV with a header row and two rows: `03/04/2025,Coffee,4.50` and `03/15/2025,Lunch,12.00`.
3. Account → Wallet → **Import transactions**, upload it and map the columns.
4. Expected (what `main` does): both expenses are in March, on the 4th and the 15th. Actual: Coffee is dated **3 abr 2025** and Lunch **15 mar 2025**.
5. Switch the language to English and import the same file: both are March, which shows the reading depends on the uploader's language.

src/libs/DateUtils.ts
```diff
@@ -43,73 +40,187 @@ import {
...
     | typeof CONST.DATE.FNS_TIMEZONE_FORMAT_STRING;
-type WeekDay = 0 | 1 | 2 | 3 | 4 | 5 | 6;
 
 const TIMEZONE_UPDATE_THROTTLE_MINUTES = 5;
 
-function isDate(arg: unknown): arg is Date {
-    return Object.prototype.toString.call(arg) === '[object Date]';
+type IntlFormatKey = keyof typeof CONST.DATE.INTL_FORMATS;
+
+function isKnownTimezone(tz: string): tz is SelectedTimezone {
```
iwiznia, line 64: In what case would we have an unkown timezone?

src/libs/DateUtils.ts
```diff
@@ -43,73 +40,187 @@ import {
...
-    return CONST.WEEK_STARTS_ON;
+/** A Wednesday in UTC, used where only the locale's own conventions matter and the instant must not vary by run. */
+const LOCALE_PROBE_DATE = new Date(Date.UTC(2023, 0, 4));
+
+/** Sampled offsets cannot identify a zone: Europe/Athens and Africa/Cairo share both solstices yet differ through April. Resolving costs ~20x a format, hence the reuse window. */
+const DEVICE_TIME_ZONE_TTL_MS = 1000;
+let deviceTimeZone: string | undefined;
+/** Negative infinity so the first lookup resolves, and so a failure is remembered for the window like a success rather than rethrowing per call. */
+let deviceTimeZoneResolvedAt = Number.NEGATIVE_INFINITY;
+function getDeviceTimeZone(): string | undefined {
```
iwiznia, line 76: Why do we need this and how is it different from `getCurrentTimezone`?

src/libs/DateUtils.ts
```diff
@@ -43,73 +40,187 @@ import {
...
+    } catch {
+        deviceTimeZone = undefined;
+    }
+    return deviceTimeZone;
+}
+
+const WEEK_DAYS = [0, 1, 2, 3, 4, 5, 6] as const satisfies readonly WeekDay[];
+
+/** Hand-rolled rather than `memoize`: this runs per rendered cell, and that cache scans its entries linearly. */
+function cacheIntlDateTimeFormat(cacheKey: string, formatter: Intl.DateTimeFormat | null): void {
```
iwiznia, line 94: Does this really need a cache?

situchan:
Have you actually reproduced the potential bug I reported and tested the fix?


Another potential from last commit:

1. Build an import CSV whose date column uses two-digit years, e.g. 3/4/25.
2. On the native app, Settings → Wallet → import transactions (or Company cards → import), map the date column, import.
3. Expected date 2025-03-04. Actual on native the date is blank/missing, while the same file imports correctly on web.


Push changes only when you're able to reproduce bug and confirm the fix. Not just code assumptions

TaduJR:
@situchan 

I tried to repro using unit tests since i don't have that specific CSV. Could you upload it here

situchan:
Simply use AI to generate test csv.
@MelvinBot generate csv for https://github.com/Expensify/App/pull/91569#issuecomment-5762577894

situchan:
> Another potential from last commit:
> 
> 1. Build an import CSV whose date column uses two-digit years, e.g. 3/4/25.
> 2. On the native app, Settings → Wallet → import transactions (or Company cards → import), map the date column, import.
> 3. Expected date 2025-03-04. Actual on native the date is blank/missing, while the same file imports correctly on web.

I think this is pre-existing bug happening on `main` so out of scope

src/libs/DateUtils.ts
```diff
@@ -217,60 +336,97 @@ function datetimeToCalendarTime(locale: Locale | undefined, datetime: string, cu
...
+ * only asked for the sub-minute "now". Unbounded because the key space is the shipped locales in those two styles, but
+ * clearable for when polyfill data arrives late.
  */
-function getZoneAbbreviation(datetime: string | Date, selectedTimezone: SelectedTimezone): string {
-    const abbreviation = formatInTimeZone(datetime, selectedTimezone, 'zzz');
-    if (abbreviation === 'GMT') {
-        return formatInTimeZone(datetime, selectedTimezone, 'O');
+function getRelativeTimeFormat(locale: Locale, numeric: Intl.RelativeTimeFormatNumeric): Intl.RelativeTimeFormat | null {
+    const cacheKey = `${locale}|${numeric}`;
+    const cached = relativeTimeFormatCache.get(cacheKey);
```
iwiznia, line 362: Why does this need a cache?

src/libs/DateUtils.ts
```diff
@@ -835,29 +1032,52 @@ function getLastBusinessDayOfMonth(inputDate: Date): number {
     return getDate(currentDate);
 }
 
+/**
+ * Field order is a property of the locale, not of any date, so a fixed reference day answers it and the result caches.
+ */
+const isDayBeforeMonth = memoize(
```
iwiznia, line 1033: I don't get why we need this...

src/libs/DateUtils.ts
```diff
@@ -835,29 +1032,52 @@ function getLastBusinessDayOfMonth(inputDate: Date): number {
...
+            const placeholder = getFallbackDatePlaceholder(locale);
+            return placeholder.indexOf('DD') < placeholder.indexOf('MM');
+        }
+        return parts.findIndex((part) => part.type === 'day') < parts.findIndex((part) => part.type === 'month');
+    },
+    {maxSize: 16, equality: 'shallow'},
+);
+
+/** Both ends or neither, so a formatter failure cannot render a range as " to Mar 20". */
+function joinRange(startPart: string, endPart: string, separator: string): string {
```
iwiznia, line 1046: Is this really needed? Seems like something we can just do whanever we call this?

src/libs/DateUtils.ts
```diff
@@ -985,8 +1219,20 @@ function formatCountdownTimer(translateParam: LocaleContextProps['translate'], h
     return `${hours}${translateParam('common.hourAbbreviation')} : ${paddedMinutes}${translateParam('common.minuteAbbreviation')} : ${paddedSeconds}${translateParam('common.secondAbbreviation')}`;
 }
 
+const WIRE_YEAR_PREFIX = /^(\d{4})/;
+
 function doesDateBelongToAPastYear(date: string): boolean {
-    const transactionYear = new Date(date).getFullYear();
+    // Read the year off the wire string, so a Dec 31 transaction viewed that evening (already Jan 1 in UTC) is not
+    // suffixed with a year on what is still today's row.
+    const yearMatch = date.match(WIRE_YEAR_PREFIX);
```
iwiznia: Why are we parsing the dates like this? The previous one seemed better as it does not depend on format, no?

src/libs/MergeTransactionUtils.ts
```diff
@@ -592,6 +594,10 @@ function getDisplayValue(
     if (field === 'amount') {
         return convertToDisplayString(Number(fieldValue), getCurrency(transaction));
     }
+    if (field === 'created') {
+        // An expense date is date-only, so it renders UTC-anchored. One the formatter cannot read stays as stored, because a blank row reads as the two candidates sharing a date.
```
iwiznia: This is not fully true, it can contain time.
But also, why do we have a new field now?

src/libs/NowStore.ts
```diff
@@ -0,0 +1,93 @@
+/**
```
iwiznia, line 1: Where is this used?

src/libs/PerDiemMerchantUtils.ts
```diff
@@ -0,0 +1,67 @@
+/**
```
iwiznia, line 1: Where is this used?

TaduJR:
@situchan 

Verified, and they are fixed. But two another pre-existing on main that are related to transition so they are unrelated.

We can continue.

situchan (approved):

src/pages/ReimbursementAccount/USD/BusinessInfo/subSteps/IncorporationDateBusiness.tsx
```diff
@@ -78,7 +78,6 @@ function IncorporationDateBusiness({onNext, isEditing}: SubPageProps) {
                 InputComponent={DatePicker}
                 inputID={COMPANY_INCORPORATION_DATE_KEY}
                 label={translate('businessInfoStep.incorporationDate')}
-                placeholder={translate('businessInfoStep.incorporationDatePlaceholder')}
```
iwiznia, line 81: Why?

TaduJR:
@iwiznia 

All yours.

iwiznia (changes requested):
This PR is huge, which makes reviewing it quite a challenge, specially since even GH UI is getting laggy and breaks due yo it being so big.
Can't we do something to send this in smaller chunks?

TaduJR:
>This PR is huge, which makes reviewing it quite a challenge, specially since even GH UI is getting laggy and breaks due yo it being so big.
Can't we do something to send this in smaller chunks?

We discussed this above, but the functionalities are super sightly coupled trying to divide will cause a lot bug during previous attempt from my side

src/pages/ReimbursementAccount/USD/BusinessInfo/subSteps/IncorporationDateBusiness.tsx
TaduJR: That placeholder said `yyyy-mm-dd` in every language: `Start date (yyyy-mm-dd)` in English, `Fecha de inicio (aaaa-mm-dd)` in Spanish, `開始日 (yyyy-mm-dd)` in Japanese. The field doesn't take dates in that order, so it was telling people the wrong format which is the same bug as #80011.

Without it, the picker shows the format for the user's language: `MM/DD/YYYY` in English, `DD/MM/YYYY` in Spanish, `YYYY/MM/DD` in Japanese. The label above already says "Incorporation date", so the placeholder only needs to show the format.

This was the only place that key was used, so I removed it too.

src/libs/NowStore.ts
TaduJR: It backs the `useNow` hook, and two components use that hook: `AutoUpdateTime` and `ParticipantLocalTime`, and the two places showing a live local time.

Before, each ran its own `setInterval` at one second. Now there's one timer for the whole app: it sleeps until the next minute boundary, wakes once, and notifies both. It's only running while something is subscribed, and the last consumer to unmount clears it.

src/libs/PerDiemMerchantUtils.ts
TaduJR: Three functions, all in use:

- `getPerDiemMerchant` writes the merchant when a per diem expense is created - `actions/IOU/PerDiem.ts`, through `computePerDiemExpenseMerchant`, on both the workspace and self-DM paths.
- `getDisplayMerchant` rebuilds it in the reader's language - `TransactionUtils.getMerchantName`, `ReportUtils`, `SearchUIUtils`, `MoneyRequestView`, and the transaction preview.
- `getPerDiemDisplayParts` splits it into destination and dates for the e-receipt - `PerDiemEReceipt.tsx`.

The merchant is saved in English so every viewer stores the same value, and each reader sees the dates in their own language. The writer and the parser are in one file because the parser only matches what the writer stored, so both sides change together. It's a leaf, and it imports `DateUtils`, `Log`, `CONST` and types, nothing else.

src/libs/MergeTransactionUtils.ts
TaduJR: You're right that `created` can hold a time, but not by the time it reaches here. The merge screen reads its values through `getTransactionDetails` right?, which formats `created` to `yyyy-MM-dd` first, so there's no time left in it. I'll fix the comment to say that instead.

The case is new because `main` didn't have one, and `created` fell through to the generic string branch, so the merge screen showed `2023-07-09` while the row next to it showed `Jul 9, 2023`. Same fix as the three other places that still showed the stored value: the Date row on expense details, the `Posted` date beside it, and the distance e-receipt imo.

src/libs/DateUtils.ts
TaduJR: Yea, the regex was the wrong shape. I'll replace it with the file's own parser:

But `main`'s version does have a bug. `new Date('2026-01-01')` is UTC midnight, which is still 31 December for anyone west of UTC, so `getFullYear()` returns 2025 there. A 1 January expense is read as last year's, all year round, not just on New Year's Eve. In the transaction preview and the list rows it shows as `Jan 1, 2026` where a 2 January row shows `Jan 2`, and in Search it's worse: `shouldShowYearCreated` latches for the whole table, so one such row turns the year on for every date in the column.

`toLocalDate` reads `2026-01-01` as local midnight, so the day is read as the day it is. It also handles the `yyyy-MM-dd HH:mm:ss` values Search passes for `submitted` and `approved`, so nothing depends on the string starting with a year any more.

src/libs/DateUtils.ts
TaduJR: It's file-local and used 13 times here. The range formatters plus the compact Search labels, and three of those nest it (a range, then the year suffix on top).

Inlining means repeating `start && end ? … : ''` at each one, and that check is the point: `formatIntl` returns `''` rather than a wrong date when a date or locale is bad, so without it one failed end renders as `" - Mar 20"` or `"Mar 17 - "`, which reads like a real value. With it, the range is either whole or empty.

src/libs/DateUtils.ts
TaduJR: It decides which end of a same-month range carries the month, and that flips by language: English writes `Mar 17-20`, Spanish writes `17-20 mar`. So one side is the bare day and the other is month-and-day, and something has to answer "does this language put the day first?".

It asks Intl for the parts of a month-day format and compares their positions, falling back to the CLDR placeholder table when Intl can't report them. The answer depends only on the locale, so it's memoized and cleared with the other locale caches. Two callers: the range formatter and the compact Search range label.

src/libs/DateUtils.ts
TaduJR: Two things but more importantly, unlike `Intl.DateTimeFormat`, this constructor is polyfilled on native. `@formatjs` installs locale data asynchronously, so an instance built before that data lands formats in English for the rest of the session. The cache gives that a single place to fix: `IntlStore` calls `refreshIntlFormatterCaches()` right after it commits a locale, which clears this cache outright, and the next render builds instances that see the new data. Without a cache there's nothing to clear and nothing to rebuild.

src/libs/DateUtils.ts
TaduJR: Yes. The construction is the expensive part. `new Intl.DateTimeFormat(...)` costs ~22µs against ~0.4µs to format with one that exists, because each builds its own ICU state. And they run per cell: `DateCell` formats in its own render, and a wide report row has five date columns, so a 50-row table is 250 constructions per render without a cache.

It also remembers failures, so a bad locale or timezone isn't retried on every row, and it's bounded at 256 entries because each holds 10–50KB of ICU state.

src/libs/DateUtils.ts
TaduJR: Different questions. `getCurrentTimezone` answers "which timezone should we show this user?" It takes their saved setting and, if it's on automatic, swaps in the device's. That's the display timezone, read by `useCurrentTimezone` and when a comment saves an updated timezone back to personal details.

`getDeviceTimeZone` answers "which timezone is this formatter bound to?" and has one caller: the cache key in `getIntlDateTimeFormat`. `Intl.DateTimeFormat` resolves the timezone at construction, so a formatter built with no explicit zone keeps the zone the app started in, while `Date` getters move immediately. Keying on the resolved zone is what stops a cached formatter being reused after the device moves.

It's cached for a second because resolving costs about 20x a cached format and it runs on the same per-cell path.

src/libs/DateUtils.ts
TaduJR: "Unknown" there doesn't mean invalid, it means "has no older alias". `timezoneNewToBackwardMap` is the reverse of the renamed-zone table in `TIMEZONES.ts`: 134 old names collapsing onto 93 current ones (`Europe/Kyiv` > `Europe/Kiev`, `Asia/Kolkata` > `Asia/Calcutta`). Plenty of zones aren't in it at all `Europe/Madrid`, for one.

Both call sites use it for one thing: when a platform rejects a newer zone ID. Older iOS and macOS ICU do this with `Europe/Kyiv` we retry with the old name. The guard just narrows the string before indexing that map, because what arrives is a plain `string`: either the device zone from `Intl.DateTimeFormat().resolvedOptions()`, whatever the OS reports, or a value stored on the account.

src/libs/DateUtils.ts
TaduJR: Two cases.

On native, Hermes ships no `Intl.Locale` at all, so `getWeekInfo` doesn't exist there. We force-install `@formatjs/intl-locale`, which does implement it, so native is fine in practice.

On web it's recent: Chrome 130, Safari 17, Firefox 153, and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo) only lists it as newly available across browsers since July 2026. Some older versions exposed it as a `weekInfo` property rather than a method, so calling `getWeekInfo()` there throws. This app has no browserslist, and the oldest thing we already rely on unpolyfilled is `Intl.ListFormat` (Firefox 78, Safari 14.1), so plenty of browsers we support don't have it.

The table sidesteps all of that it's the same answer on every engine, and a test asserts it matches real `getWeekInfo` for every shipped locale, so a CLDR change fails CI rather than drifting.

src/libs/DateUtils.ts
TaduJR: Both places that use it ask the locale a question, not a date one: what its weekday letters are, and what order it puts day, month and year in. Any date gives the same answer, so the point is to pin the input — with `new Date()` there, the memoized answers and the tests would depend on when they ran.

UTC and mid-week are defensive rather than load-bearing. A local-time instant is a different calendar day either side of UTC, and a probe on a week edge lands in a different week under different week-start rules. Neither changes what comes back today, and both would make the probe's meaning depend on where it runs.

src/languages/localeDayOfMonthMap.ts
TaduJR: It sits next to `localeOrdinalMap`, which is outside the language files on `main` for a reason its own comment gives: those files are typed `TranslationDeepObject<typeof en>`, so every locale has to declare exactly the keys English declares.

This map needs the opposite. English has no entry at all an English date reads as the rank (`15th`), which `toLocaleOrdinal` already renders while ten other locales do have one. In a language file that's not expressible: English would have to carry a key it doesn't need, and the rule would be spread across 11 files instead of readable as one table.

It's also not translated text. Nothing here is a phrase a translator writes. It's the grammar of writing a day number, consumed by `LocaleDigitUtils`, not by the translation layer.

src/libs/Localize/index.ts
TaduJR: Two changes there, and only one of them is "now".

The cache registration is. `Intl.ListFormat` is polyfilled on native, and an instance resolves its locale at construction. The store starts on `en` before anything has loaded, so a list built in that window is bound to whatever data existed then and stays cached under `en` for the session. This PR adds the refresh that drops derived Intl caches when a locale commits, and this cache has to be registered with it.

The try/catch and the separator aren't. main constructs `Intl.ListFormat` bare, so on a runtime that lacks it `formatList` and `formatMessageElementList` throw rather than render. I hit that while adding the same guard for `Intl.RelativeTimeFormat` next door, and it turned out `ListFormat` is the likelier of the two to be missing: on web we polyfill only `NumberFormat`, and `ListFormat` shipped latest of the Intl constructors we rely on (Firefox 78, Safari 14.1). Losing the "and" is cosmetic, losing the names isn't.

That second part is pre-existing, so we can pull it into its own PR if you'd rather keep this one to the date work.

src/libs/Localize/index.ts
TaduJR: Three cases: a cold start before any chunk has landed, a caller that passes the locale straight from settings while that chunk is still loading, and a locale whose chunk failed, where the store has fallen back to English but the setting still holds the failed one. It returns the committed locale's string if there is one, and the path if nothing is loaded yet.

It matters because `IntlStore.get` falls back to the current locale on its own, so without this branch a Spanish request would quietly come back with the English string. `memoizedGetTranslatedPhrase` is keyed on `(locale, path)`, unbounded and never cleared, so that English string would then be cached under `es` for the rest of the session.

Issue #90477, opened 2026-05-13
Migrate CardAuthenticationModal to @react-navigation modal screen


mountiny:
## Context

This is a sub-issue of #53493 (Replace modal screens with modals from `@react-navigation`).

For full background and migration approach, see the [triage analysis comment](https://github.com/Expensify/App/issues/53493#issuecomment-4312826488) and the [sub-issue breakdown comment](https://github.com/Expensify/App/issues/53493#issuecomment-4439641260).

## Problem

`CardAuthenticationModal` handles 3D Secure (SCA) card authentication. It is web-only and its visibility is driven by an Onyx key rather than props, making it complex to migrate to `@react-navigation`.

## Migration approach (Group 4 — Complex)

- **File:** `src/pages/settings/Subscription/CardAuthenticationModal/index.tsx`
- **Used by:** `PaymentCard/index.tsx` (no `policyID`) and `WorkspaceOwnerChangeWrapperPage.tsx` (with `policyID`)
- **Complexity:**
  - Web-only: the native implementation returns `null`.
  - Visibility is driven by the `VERIFY_3DS_SUBSCRIPTION` Onyx key (not a prop).
  - Uses `window.addEventListener('message')` for the SCA 3D Secure iframe callback.
  - Different success actions depending on whether `policyID` is present.

### Suggested migration approach
1. Create a dedicated navigation page (e.g., `CardAuthenticationPage`) that reads `VERIFY_3DS_SUBSCRIPTION` from Onyx.
2. When `authenticationLink` appears in Onyx, navigate to the new page automatically (can be done in a `useEffect` in the parent or a global listener).
3. Pass `policyID` as a route param to handle the different success actions.
4. Implement the `window.addEventListener('message')` 3DS callback within the new page.
5. On web-narrow, the page should render as a modal screen; on native, the flow is a no-op.
6. Test the full 3D Secure card authentication flow on web (both with and without `policyID`).

## Reference

- Triage analysis: https://github.com/Expensify/App/issues/53493#issuecomment-4312826488
- Parent tracking issue: #53493

<details><summary>Issue Owner</summary>Current Issue Owner: @parasharrajat</details>

sharabai:
@mountiny could you assign me this one?

parasharrajat:
No update yet.

sharabai:
I've been focusing on other things for now, I'll come back to this issue soon.

sharabai:
[...]

PR #91982, opened 2026-05-28
Migrate card-auth modal component to a real modal

sharabai:
### Explanation of Change

Migrates the inline `CardAuthenticationModal` (`react-native-modal`) to a real `@react-navigation` screen (`DynamicCardAuthenticationPage`), reachable via the dynamic `card-authentication` URL suffix. Native is intentionally inert (3DS on mobile is handled by OldDot).

> [!IMPORTANT]
> Behavior changes on top of the migration (improvements, not regressions):
> * **"Authenticate payment" no longer detours through Add Payment Card.** Pre-migration the modal needed a portal host, so Subscription pushed the Add Payment Card RHP just to host the overlay. With a real screen there is no host to justify, so re-auth opens the 3DS challenge directly on top of Subscription and closes back to Subscription.
> * **3DS screen is a root-stack route presented as a centered floating modal** instead of a SETTINGS RHP sub-screen. It uses the same `transparentModal` screen options (`centeredModalNavigator`) as the other route-based centered modals, so the screen that triggered it (e.g. the Add Payment Card RHP) stays visible behind a dimmed backdrop; clicking the backdrop or pressing Esc cancels the challenge, matching the old modal's backdrop behavior.
> * **The modal card's styles are copied, not shared.** The geometry lives in a new `styles.getCardAuthenticationModalInnerView` helper whose values come from the old modal's `CENTERED_UNSWIPEABLE` container in `ModalStyleUtils` — fills the screen with a 20px inset, `componentBorderRadiusLarge` corners and a shadow on wide layouts; full-bleed with square corners on narrow (the inset is expressed as margins instead of the original `windowWidth - 40` width computation, which drops the window-dimensions dependency). Reusing `getModalStyles` directly was rejected because its output is coupled to the `react-native-modal` contract (`animationIn`/`animationOut`, `swipeDirection`, safe-area flags) that a navigation route doesn't consume, and it would couple the migrated route back to the legacy modal system it's leaving — future tweaks meant for the remaining `CENTERED_UNSWIPEABLE` consumers would silently restyle this route. The existing `getCenteredModalInnerView` wasn't a fit either: it's built for fixed-width, content-hugging dialog cards (feature training, welcome modals) with a bottom-sheet shape on narrow layouts, and can't express a screen-filling card for an iframe with no intrinsic height. This mirrors how the other migrated surfaces do it (`getTestToolsNavigatorInnerView`, `getCenteredModalInnerView` are also per-surface helpers).
> * **Currency defaults now resolve through `usePreferredCurrency`** (billing-card currency → local currency → USD) everywhere — the Add Payment Card form's selector button, the currency picker, and Change Billing Currency. Previously the button and picker derived the default differently, so the button could show one currency while the picker opened pre-selected on another.
> * **The add-card currency default respects the `EUR_BILLING` beta.** `usePreferredCurrency` can resolve to EUR (from the user's local currency) regardless of the beta, but the currency picker only offers EUR when the beta is on. So on the add-card entry points (the Add Payment Card form default and the picker's pre-selected row) a EUR default is clamped to USD when the beta is off — keeping the default in sync with the options the user can actually pick and avoiding an unintended EUR/SCA add-card for users who can't select EUR. Change Billing Currency stays ungated, since it reflects an existing card's real currency.
> * **Currency picker is scoped per flow, keeps the pick within an attempt, and resets across attempts.** Saving in the Add Payment Card flow writes the **draft** key (`ADD_PAYMENT_CARD_FORM_DRAFT`, which `FormProvider` hydrates from) so the form no longer reverts after picking EUR/GBP and returning; the Change Billing Currency flow writes its own draft. Both the read and the write are scoped by entry point, so committing in one flow no longer bleeds into the other — previously *both* keys were written on every commit, which is why Change Billing Currency carried an unmount cleanup (`setPaymentMethodCurrency(initialCurrency)`) purely to undo the add-card half; scoping the write lets that cleanup be deleted rather than kept. `clearPaymentCardFormErrorAndSubmit` (run on add-card open/close) also clears the draft currency, preserving the long-standing reset-on-open intent — a fresh attempt re-defaults through `usePreferredCurrency` rather than hydrating a currency abandoned in an earlier attempt.
>   *Note on provenance: the picker commits on **Save** rather than on row tap as of [#97075](https://github.com/Expensify/App/pull/97075), which landed on `main` while this PR was open and is merged in here. The per-flow scoping above applies to that Save handler; the deferred-commit interaction itself is not this PR's change.*
> * **Teachers Unite Add Payment Card** is registered as an entry screen for the `payment-card-currency` dynamic route, so the currency picker no longer 404s from that flow.
> * **"Authenticate payment" is gated while a request is in flight.** Rapid clicks previously fired one `VerifySetupIntent` request each (up to ~9 back-to-back). The button's `isLoading` is now bound to a new optimistic Onyx flag (`SUBSCRIPTION_VERIFY_SETUP_INTENT_PENDING`), set in the request's `optimisticData` and cleared in `successData`/`failureData` — so the button shows a spinner and additional clicks are no-ops until the response lands.
> * **The 3DS link is now scoped to the screen that triggered it.** `verify3dsSubscription` is a single global Onyx key, so previously clicking **Authenticate payment** and then opening **Change payment card** would let the in-flight auth link open the challenge iframe on top of the Add Payment Card page — unexpected, since by then the user means to start a fresh add-card flow. The producing request now records its own `route.name` in `verify3dsSubscriptionSource` via its `successData`, so the source lands in the same Onyx flush as that response's link and stays tied to it (overlapping requests can't make the source point at a different attempt than the link). The navigation hook only reacts to a link when the focused screen matches that source — so a link can no longer be consumed by an unrelated screen the user has since navigated to.
> * **Repeat 3DS attempts always reopen, even when the backend returns a byte-identical SCA link.** The navigation hook fires only on a *change* to `verify3dsSubscription`, and the backend hands back the same link on retries. Rather than clearing the link from multiple consumer spots (on close, on orphan), every link-producing action now clears it first at a single seam (`prepareCardAuthentication`, gated by the presence of a `source`) before issuing its request — so the response, new or identical, always lands as a `'' → URL` change. Stale links from an abandoned attempt or a reload are inert at mount (the mount-time run is skipped) and overwritten by the next attempt.

### Fixed Issues

$ https://github.com/Expensify/App/issues/90477

### Tests

> 🔴 The 3DS iframe only appears for **GBP or EUR** — always pick GBP/EUR in the currency selector **and press Save**. The selector commits on Save, not on row tap, so leaving via the back caret discards the pick. Use Stripe test card `4000 0000 0000 3220` (3DS Required) on a **staging** account in Stripe test mode. The challenge iframe has its own **Complete authentication** / **Fail authentication** buttons.

**Setup for Flow A / D:** the **Account → Subscription** menu item only shows if you own a paid workspace. If you don't see it, create a workspace and switch its plan to **Control**.

**Flow A — Subscription add card**
1. Go to **Account → Subscription**. If no card on file, click **Add payment card**; otherwise click the card's **⋮ menu → Change payment card**.
2. Fill the form, set currency to **GBP or EUR**, submit. Verify the challenge opens as a centered floating card (20px inset, rounded corners) over the dimmed Add Payment Card form on wide screens, and full-screen without rounding on narrow ones.
3. In the 3DS challenge, click **Complete authentication**. Verify the page auto-closes, the URL drops `/card-authentication`, and the card is added.
4. Repeat the flow but click **Fail authentication**. Verify the page stays open and the header **back** returns you to the Add Payment Card form.
5. **Cross-flow scoping check.** On the Add Payment Card form open the **Currency** row, pick **GBP**, press **Save**, then back out of the form. Now open the card's **⋮ menu → Change payment currency** and verify it shows the **card's actual currency**, not GBP. Do the reverse too: change the currency to **GBP** there, back out, then open **Change payment card** and verify the form defaults through `usePreferredCurrency` instead of showing GBP.

**Flow B — Teachers Unite add card** (requires an account with **no billing card on file** — easiest is a fresh test account)
1. Go to **Account → Save the World** and toggle **Enable Personal Karma** ON.
2. Fill the add-card form, **GBP or EUR**, submit.
3. Complete the 3DS challenge and verify the page auto-closes back to Save the World.
4. While on the form, open the **Currency** row, pick **EUR**, press **Save**, and verify the form still shows **EUR** (regression check for the draft-key fix). Reopen the row and verify **EUR** is pre-selected with **Save** disabled — the persisted pick and the selection now agree, so there is nothing to commit.

**Flow C — Workspace owner change** (needs two accounts: Owner A and Admin B)
1. As Owner A, create a workspace and invite B as **Admin**.
2. Log in as B. Open **Workspace → Members → A (the owner)** and press **Transfer owner**.
3. When the **Enter your payment card to transfer ownership** form appears, enter a **GBP/EUR** card, accept terms, submit.
4. Complete the 3DS challenge. Verify **ownership actually transfers** (not just a card add), the `/card-authentication` suffix is removed, and exiting the owner-change flow returns to the screen it was opened from without a `backTo` query parameter.

> 🔧 If B is on an active free trial, `RequestWorkspaceOwnerChange` returns success immediately and the card/3DS step is skipped. To reach the card form, expire the trial first by setting the trial NVP to a past date (in testing this was enough for the owner-change call to come back requiring a billing card). Temporary shim to drop into `WorkspaceOwnerChangeWrapperPage`:
>
> ```tsx
> // import Onyx from 'react-native-onyx';
> useEffect(() => {
>     Onyx.set(ONYXKEYS.NVP_LAST_DAY_FREE_TRIAL, '2026-02-03 13:40:26');
> }, []);
> ```

**Flow D — Re-authenticate existing billing** (only stageable when last billing failed with auth-required)
1. Go to **Account → Subscription** and press **Authenticate payment**.
2. Verify the URL goes directly to `/settings/subscription/card-authentication` (no Add Payment Card detour).
3. Complete the 3DS challenge and verify billing clears and you return to Subscription.

**Navigation edge cases (run on Flow A):**
1. Mid-challenge, hard-reload the tab. Verify the URL stays single-suffix (no `…/card-authentication/card-authentication`) and back still returns to the parent.
2. Press header **back** during a challenge. Verify it returns to the parent and the page does NOT immediately reopen. Then press **Authenticate payment** again and verify the challenge reopens (each new attempt clears the link at the producer, so a repeat click reopens even if the backend returns the same link).
3. Start a challenge, navigate away (or hard-reload back onto the parent URL), then return to the parent. Verify the challenge does **NOT** auto-reopen on its own (the mount-time run is skipped). Then trigger it again (**Authenticate payment** / **Change payment card**) and verify a fresh challenge opens normally.
4. Mid-challenge, click the dimmed area outside the card, and in a separate attempt press **Esc**. Verify both cancel the challenge and return to the parent exactly like the header back caret (the parent form keeps its filled-out values).
5. On Subscription, click **Authenticate payment** and then immediately open the card's **⋮ menu → Change payment card** before the auth link arrives. Verify the 3DS iframe does **NOT** open over the Add Payment Card form — the form stays put — because the link is scoped to the screen that triggered it. Then return to Subscription and click **Authenticate payment** again: verify a fresh challenge opens (the next attempt clears the orphaned link at the producer, so a repeat click still registers as a change).

### Offline tests

Same as tests.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/de67f5ee-06d7-494b-a693-5cabbc5572e5

https://github.com/user-attachments/assets/a977d812-e746-4b12-a682-e364dae54221

https://github.com/user-attachments/assets/bee0174d-dd5b-443d-990e-8385e0c4fda3

</details>


sharabai:
Update: I was testing, but found some visual bug that is not related to the issue directly, but decided to fix it since I'm already working on that component. 
Other than that, it's almost ready to for review.

sharabai:
Update: found another bug/issue that could be fixed that works the same on main. Working on it

sharabai:
@codex review

sharabai:
@mountiny I tried to fix as many things that looked wrong that are currently on main, but one bug isn't fixable on FE. 

Tabbing between Wallet and Subscription causes this flicker:

https://github.com/user-attachments/assets/b14b4471-666e-4d7e-9187-bc1f6d2fb190


@mountiny Is this smth that even needs to be fixed? Should I file an issue for it? 
Fixing it would require help from BE and refactoring multiple places on FE that use a certain Onyx key. 

A more concrete mechanism is this:
the `fundList` Onyx key is shared by two different backend domains
(Wallet and Subscription). Opening the Wallet causes the backend to **replace
(`set`)** `fundList` with the wallet's payment‑method view — which, for an
account whose only card is the subscription billing card, is empty (`{}`). That
momentarily deletes the billing card the Subscription page reads, so the
empty‑state CTA renders until the Subscription's own request **merges** the card
back.

sharabai:
@codex review

src/pages/workspace/members/WorkspaceOwnerChangeWrapperPage.tsx
```diff
@@ -29,6 +29,7 @@ type WorkspaceOwnerChangeWrapperPageProps = WithPolicyOnyxProps & PlatformStackS
 function WorkspaceOwnerChangeWrapperPage({route, policy, isLoadingPolicy}: WorkspaceOwnerChangeWrapperPageProps) {
     const styles = useThemeStyles();
     const {translate} = useLocalize();
+    useNavigateToCardAuthenticationOnLink();
```
parasharrajat: can we move it later where other effects are so that it is easy to spot? let's do it in each component.

sharabai: sure, not a problem. On one hand it's just a hook like any other, on the other hand, the hook inside uses `useEffect`, so moving it closer to other `useEffect`s won't hurt.

sharabai:
@parasharrajat I've applied the changes you suggested , but also changed slightly the logic. 
I'd need to do the proper testing again. For that I'd need a local backend and currently the person I was relying on is OOO and they'll be back Thursday. So till that day, I won't be able to check if the changes introduced any regressions. (They shouldn't, but it's always best to go over all the test cases).

sharabai:
@codex review

sharabai:
@codex review

src/libs/actions/PaymentMethods.ts
```diff
@@ -305,30 +320,69 @@ function clearPaymentCardFormErrorAndSubmit() {
...
+ * Begin a NEW 3DS attempt: drop any stale link so the backend's next link (even an identical one)
+ * registers as a change and reopens the challenge, and record which screen initiated it. No-ops without
+ * a source — the in-place finalize/re-verify case, which must leave the active link untouched.
+ */
+function prepareCardAuthentication(source?: string) {
+    if (!source) {
+        return;
+    }
+    clearPaymentCard3dsVerification();
+    setVerify3dsSubscriptionSource(source);
```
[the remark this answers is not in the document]
sharabai: synced source to SCA_URL

sharabai:
@codex review

src/libs/actions/PaymentMethods.ts
```diff
@@ -308,27 +323,62 @@ function clearPaymentCardFormErrorAndSubmit() {
...
 
 /**
  * Set currency for payments
  *
  */
 function setPaymentMethodCurrency(currency: ValueOf<typeof CONST.PAYMENT_CARD_CURRENCY>) {
-    Onyx.merge(ONYXKEYS.FORMS.ADD_PAYMENT_CARD_FORM, {
+    Onyx.merge(ONYXKEYS.FORMS.ADD_PAYMENT_CARD_FORM_DRAFT, {
         [INPUT_IDS.CURRENCY]: currency,
     });
```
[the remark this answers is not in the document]
sharabai: resolved with also clearing draft on clear cb

src/hooks/useNavigateToCardAuthenticationOnLink/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        if (firstRenderRef.current) {
+            firstRenderRef.current = false;
+            return;
+        }
+        // `verify3dsSubscription` is one global key shared by several mounted hooks: only the focused screen
+        // that recorded the link as its `source` may consume it, so an unrelated screen can't hijack it.
+        if (!isFocused || !authenticationLink || route.name !== source) {
+            return;
+        }
+        Navigation.navigate(createDynamicRoute(DYNAMIC_ROUTES.CARD_AUTHENTICATION.path));
```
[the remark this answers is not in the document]
sharabai: policyID isn't lost here, the card-authentication screen is a dynamic route layered on the OWNER_CHANGE_CHECK base path (workspaces/:policyID/change-owner/...), and getStateFromPath merges that base route's params (including policyID) onto the dynamic leaf, so route.params.policyID is populated for the owner-change flow and undefined for subscription, exactly as intended.

sharabai:
@codex review

sharabai:
@codex review

src/hooks/useNavigateToCardAuthenticationOnLink/index.tsx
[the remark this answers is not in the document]
sharabai: Same as [here](https://github.com/Expensify/App/pull/91982#discussion_r3499231765)

src/components/AddPaymentCard/PaymentCardForm.tsx
```diff
@@ -355,7 +357,7 @@ function PaymentCardForm({
                 {!!showCurrencyField && (
                     <View style={[styles.mt4, styles.mhn5]}>
                         <InputWrapper
-                            value={data?.currency ?? CONST.PAYMENT_CARD_CURRENCY.USD}
+                            defaultValue={preferredCurrency}
```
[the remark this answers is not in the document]
sharabai: Refactored code to gate behind a check.

sharabai:
@codex carefully review

sharabai:
@codex review

sharabai:
@codex review

sharabai:
@codex review

src/pages/workspace/members/WorkspaceOwnerPaymentCardForm.tsx
```diff
@@ -72,9 +74,9 @@ function WorkspaceOwnerPaymentCardForm({policy}: WorkspaceOwnerPaymentCardFormPr
                 addressZip: values.addressZipCode,
                 currency: values.currency,
             };
-            addBillingCardAndRequestPolicyOwnerChange(policyID, currentUserAccountID, currentUserEmail, cardData);
+            addBillingCardAndRequestPolicyOwnerChange(policyID, currentUserAccountID, currentUserEmail, cardData, route.name);
```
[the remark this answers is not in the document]
sharabai: Yeah, for that to happen, a whole chain of very unlikely events would have to line up, so I don't think it's something we need to account for. Besides, this PR already introduces gating logic, whereas the original one didn't have any.

src/libs/actions/PaymentMethods.ts
```diff
@@ -305,30 +331,70 @@ function clearPaymentCardFormErrorAndSubmit() {
         [INPUT_IDS.ACCEPT_TERMS]: '',
         [INPUT_IDS.CURRENCY]: CONST.PAYMENT_CARD_CURRENCY.USD,
     });
+    // The currency picker writes its selection to the form draft, which FormProvider hydrates the currency field from.
+    // This reset runs whenever an add-card flow opens/closes, so dropping the draft currency lets each attempt default
+    // through usePreferredCurrency rather than reusing a currency picked then abandoned in an earlier attempt.
+    Onyx.merge(ONYXKEYS.FORMS.ADD_PAYMENT_CARD_FORM_DRAFT, {[INPUT_IDS.CURRENCY]: null});
```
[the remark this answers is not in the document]
sharabai: Nah, `Onyx.merge` treats `null` as "remove the property" rather than "write null" like `Onyx.set` does. It might be worth adding a regression test in case someone changes that behavior down the road.

sharabai:
@parasharrajat
I've refactored the code based on your suggestion. I also verified that the new commits fixing the issue Codex found didn't introduce any regressions. It's back in your court now. 😄

parasharrajat:
@MelvinBot review this.

sharabai:
> ## ⚠️ Possible 3DS auto-open on hard reload (edge cases 1 & 3)
> The mount-skip in [`useNavigateToCardAuthenticationOnLink/index.tsx:27-36`](https://github.com/Expensify/App/blob/ae298c624680ea1867200c38820cd318e48b0ce1/src/hooks/useNavigateToCardAuthenticationOnLink/index.tsx#L27-L36) relies on `firstRenderRef` catching the render that carries the stale link. But on a **hard reload** the Onyx cache is cold, so `useOnyx(VERIFY_3DS_SUBSCRIPTION)` returns `undefined` on the mount render and the real stored URL arrives in a _later_ render. That means:

The "cold cache on hard reload" premise no longer holds. Since [react-native-onyx#752](https://github.com/Expensify/react-native-onyx/pull/752) (Onyx 3.0.52, in App since [#85210](https://github.com/Expensify/App/pull/85210), March 2026), Onyx.init() eagerly loads the entire database into the in-memory cache in a single batch read (initializeWithDefaultKeyStates → storage.getAll()), and every connection, including useOnyx, is gated behind deferredInitTask, which only resolves after that load completes. Since the authenticated UI itself mounts behind useOnyx-delivered session data, any screen using this hook mounts strictly after full hydration.

No undefined mount render. useOnyx(VERIFY_3DS_SUBSCRIPTION) reads synchronously from the warm cache on the very first render, so firstRenderRef is spent on the render that already carries the stale link — exactly the case it exists for. 
I verified this empirically: after a hard reload onto the entry screen with a non-empty verify3dsSubscription, the link is defined on every render (logged it) and nothing auto-opens.
No hydration orderings. verify3dsSubscription and verify3dsSubscriptionSource land in the cache together in one atomic batch before any mount, so the three delivery orders can't occur. There is no in-session undefined -> URL transition on reload.

sharabai:
@parasharrajat I've addressed the review feedback and fixed the nit MelvinBot pointed out (even though it already behaves the same way on main, btw 😄). The ball's back in your court!

sharabai:
@parasharrajat Friendly bump

sharabai:
@mountiny I've just resolved another conflict. If @parasharrajat doesn't have time to review, would you mind assigning it to someone who does?

parasharrajat:
I am working on it. I will have update shortly

parasharrajat:
@MelvinBot if we want to test this PR by mocking data on App UI, how can we do that?

parasharrajat:
I couldn't test these flows with real data. I don't see the 3DS challange modal using my personal card. The testing flows require local backend which is also not available to to me. 


I will try to test it with mocked data.

parasharrajat:
I think we should reassign who can QA this properly. May be this needs to be internal QA.  @mountiny

src/libs/Navigation/AppNavigator/AuthScreens.tsx
```diff
@@ -367,6 +369,12 @@ function AuthScreens() {
                             options={rootNavigatorScreenOptions.fullScreen}
                             component={ConnectionCompletePage}
                         />
+                        <RootStack.Screen
+                            name={SCREENS.DYNAMIC_CARD_AUTHENTICATION}
```
allgandalf, line 373: `DYNAMIC_PROFILE_AVATAR` is in `MODAL_ROUTES_TO_DISMISS`, this one isn't. `dismissModal` no-ops while the challenge is open, is that intentional ?

src/pages/settings/Subscription/PaymentCard/index.tsx
```diff
@@ -116,12 +122,17 @@ function AddPaymentCard() {
...
         <ScreenWrapper testID="AddPaymentCard">
             <FullPageNotFoundView
-                shouldShow={shouldShowBlockingView}
+                shouldShow={shouldShowBlockingView && !hasSubmittedPaymentCard}
                 onBackButtonPress={Navigation.goBack}
             >
                 <DelegateNoAccessWrapper accessDeniedVariants={[CONST.DELEGATE.DENIED_ACCESS_VARIANTS.DELEGATE]}>
                     <HeaderWithBackButton title={translate('subscription.paymentCard.addPaymentCard')} />
                     <View style={styles.containerWithSpaceBetween}>
+                        {!!isVerifyingSetupIntent && (
```
allgandalf, line 131: this key is global, so clicking Authenticate payment on Subscription and then opening Change payment card covers this form with a spinner.

src/libs/actions/PaymentMethods.ts
```diff
@@ -309,30 +361,70 @@ function clearPaymentCardFormErrorAndSubmit() {
...
     Onyx.set(ONYXKEYS.VERIFY_3DS_SUBSCRIPTION, '');
 }
 
+/**
+ * Begin a NEW 3DS attempt: drop any stale link so the backend's next link (even an identical one) registers as a
+ * change and reopens the challenge. No-ops without a source — the in-place finalize/re-verify case, which must leave
+ * the active link untouched. The screen that initiated the attempt is recorded separately via
+ * getVerify3dsSubscriptionSourceData in the request's successData, so the source stays tied to that response's link.
+ */
+function prepareCardAuthentication(source?: string) {
```
allgandalf, line 383: nothing clears `verify3dsSubscription` when the challenge closes, so the SCA url sits in storage until the next attempt. should we clear it on close too ?

allgandalf (commented):
left a few comments, mostly around the new Onyx keys being global

src/libs/actions/PaymentMethods.ts
```diff
@@ -309,30 +361,70 @@ function clearPaymentCardFormErrorAndSubmit() {
         [INPUT_IDS.ACCEPT_TERMS]: '',
         [INPUT_IDS.CURRENCY]: CONST.PAYMENT_CARD_CURRENCY.USD,
     });
+    // The currency picker writes its selection to the form draft, which FormProvider hydrates the currency field from.
+    // This reset runs whenever an add-card flow opens/closes, so dropping the draft currency lets each attempt default
+    // through usePreferredCurrency rather than reusing a currency picked then abandoned in an earlier attempt.
+    Onyx.merge(ONYXKEYS.FORMS.ADD_PAYMENT_CARD_FORM_DRAFT, {[INPUT_IDS.CURRENCY]: null});
```
allgandalf, line 367: FormProvider reads the draft on its first render, before this runs on mount, so only the unmount call clears anything. a reload brings the old currency back.

src/libs/actions/PaymentMethods.ts
```diff
@@ -309,30 +361,70 @@ function clearPaymentCardFormErrorAndSubmit() {
         [INPUT_IDS.ACCEPT_TERMS]: '',
         [INPUT_IDS.CURRENCY]: CONST.PAYMENT_CARD_CURRENCY.USD,
```
allgandalf, line 362: nothing reads `currency` off `ADD_PAYMENT_CARD_FORM` any more, the form and the selector both use the draft. can we drop this line ?

src/pages/settings/Subscription/DynamicCardAuthenticationPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+    };
+    const onClose = () => {
+        Navigation.goBack(backPath);
+    };
+
+    useKeyboardShortcut(CONST.KEYBOARD_SHORTCUTS.ESCAPE, onClose, {shouldBubble: false});
+
+    return (
+        <>
+            <CenteredModalLayoutOverlay onBackdropPress={onClose} />
```
allgandalf, line 54: `CenteredModalLayout` already does the overlay, the `box-none` view and the escape shortcut. can we reuse it and pass the card style in ?

src/libs/actions/PaymentMethods.ts
```diff
@@ -283,12 +284,63 @@ function addSubscriptionPaymentCard(
...
 
+/**
+ * Builds the `successData` entry that records which screen initiated a 3DS verification request, so only that
+ * screen's mounted useNavigateToCardAuthenticationOnLink hook reacts to the resulting link by navigating. Returning
+ * it as `successData` makes the source land in the same Onyx flush as that response's link, so it stays tied to the
+ * exact response that produced the link even when two 3DS requests are in flight. Callers pass their own `route.name`;
+ * passing nothing (e.g. the card-authentication screen re-verifying its own iframe in place) records no source and
+ * leaves the active one untouched.
+ */
+function getVerify3dsSubscriptionSourceData(source?: string): Array<OnyxUpdate<typeof ONYXKEYS.VERIFY_3DS_SUBSCRIPTION_SOURCE>> {
```
allgandalf, line 295: this one, `prepareCardAuthentication` and the `box-none` comment are all 4+ lines. can we cut each to a line and leave the rest in the PR body ?

src/libs/actions/Policy/Policy.ts
```diff
@@ -1606,7 +1607,7 @@ function addBillingCardAndRequestPolicyOwnerChange(
             currency: currency as ValueOf<typeof CONST.PAYMENT_CARD_CURRENCY>,
             isP2PDebitCard: false,
         };
-        PaymentMethods.addPaymentCardSCA(params);
+        PaymentMethods.addPaymentCardSCA(params, undefined, source);
```
allgandalf, line 1610: can the loading onyxData just be built inside `addPaymentCardSCA` so callers don't have to pass `undefined` ?

allgandalf:
Flow A — Subscription add card
1. Account → Subscription → Add payment card (or the card's ⋮ → Change payment card).
2. Fill the form. Open Currency, pick GBP, press Save.
Verify the form still shows GBP after you land back on it.
3. Submit.
Verify the challenge opens as a floating card with a 20px inset and rounded corners over a dimmed Add Payment Card form. Narrow the window and verify it goes full bleed with square corners.
Verify the URL is …/settings/subscription/add-payment-card/card-authentication.
4. Click Complete authentication. Verify the form shows a spinner, then both screens close and the card is on file.


https://github.com/user-attachments/assets/b3ba2370-2de2-4240-9df6-1de954a9a953


5. Repeat with Fail authentication. 

Verify you land back on the filled form with an error.


https://github.com/user-attachments/assets/05a5c9fa-6560-4bdd-b353-186ee7098422

sharabai:
Hi! Just a heads-up that I'm currently on sick leave, so I won't be able to look into this right away. I'll get back to you as soon as possible once I'm back. Thanks for your patience!

allgandalf:
all good, sorry i didn't know that, please take your time, get better soon ❤️

sharabai:
@allgandalf I'm back and will get back to you as quickly as possible!

src/libs/Navigation/AppNavigator/AuthScreens.tsx
sharabai: Intended. The point of the PR is that the challenge stopped being a modal, so nothing in the flow calls `dismissModal()` anymore: the back button, the backdrop and Escape all close it with `Navigation.goBack(backPath)`. It still reads modal-ish in the diff because it keeps the centered options and the styling. Screens land in that set when someone finds a `dismissModal()` call on them that no-ops, which is why the profile avatar is there.

src/libs/actions/PaymentMethods.ts
sharabai: Intended. Closing is only the happy path: a closed tab or a hand-typed URL leaves the link behind just the same, so every action that requests a new one has to clear it first regardless ([PaymentMethods.ts:383-389](https://github.com/software-mansion-labs/expensify-app-fork/blob/017b1d438f32e3e11d4e164fc1ad67f87e18b7ac/src/libs/actions/PaymentMethods.ts?plain=1#L383-L389)). That covers the close case too, so clearing on close would be a second path to the same place rather than a fix. Nothing goes wrong while the link waits: the navigation hook skips its mount run, and only the screen that recorded the link as its source may consume it ([useNavigateToCardAuthenticationOnLink/index.tsx:27-33](https://github.com/software-mansion-labs/expensify-app-fork/blob/017b1d438f32e3e11d4e164fc1ad67f87e18b7ac/src/hooks/useNavigateToCardAuthenticationOnLink/index.tsx?plain=1#L27-L33)).

src/libs/actions/PaymentMethods.ts
sharabai: Yep, dropped in [`42ae65a`](https://github.com/Expensify/App/pull/91982/commits/42ae65a01c6a14491350e4a0f4cacd362e082b2a). This PR removed the form's last read of it, so the draft reset just below is the only one left that does anything.

src/pages/settings/Subscription/PaymentCard/index.tsx
sharabai: Good catch, fixed in [`6fe9210`](https://github.com/Expensify/App/pull/91982/commits/6fe921058ac916854a80b0130921c1d9d56f1579). The spinner now also requires a submission made on this screen, so a verification started from Subscription no longer covers this form.

src/pages/settings/Subscription/DynamicCardAuthenticationPage.tsx
sharabai: Duplication's gone in [`0e473e75`](https://github.com/Expensify/App/pull/91982/commits/0e473e756b8141a3004dc95b0cf52d370aabee89): a shared `DismissibleBackdrop` owns all three, and both this screen and `CenteredModalLayout` render it.

Not `CenteredModalLayout` itself, since no prop reaches its card ([index.tsx:62](https://github.com/Expensify/App/blob/0e473e756b8141a3004dc95b0cf52d370aabee89/src/components/CenteredModalLayout/index.tsx?plain=1#L62)) and its geometry is the opposite of this screen's.

src/libs/actions/PaymentMethods.ts
sharabai: Right, fixed in [`5e3a222`](https://github.com/Expensify/App/pull/91982/commits/5e3a2221a3b).

src/libs/actions/PaymentMethods.ts
sharabai: Trimmed all three, though not quite down to a line each. I cut whatever the code, the Onyx key's doc or a neighbouring comment already says, and the history lives in the PR body. What's left is the reasoning you can't get from reading the code, like why the source goes in successData rather than optimisticData, or why clearing the link is skipped when there's no source. Moving that to the PR body would make it easy to "simplify" away later.

src/libs/actions/Policy/Policy.ts
sharabai: Yep, done in [`5e0e022`](https://github.com/Expensify/App/pull/91982/commits/5e0e022a2333c0806b55c388fade33079d2c4a34). `addPaymentCardSCA` builds the form loading data itself now, so both callers pass just the params and the source.

sharabai:
Thanks for the review @allgandalf! 🙌 It led to a lot of good improvements.

I've addressed all the comments, and the latest `main` is merged into the branch.

PR #92137, opened 2026-05-30
Bump Node.js from 20.20.0 (EOL) to 22.22.3 (Active LTS)

mountiny:
## Why

Node 20 reached **end-of-life on April 30, 2026** and no longer receives security patches. We're currently running on an unsupported runtime in CI and in local development.

Node 22 (codename **Jod**) is the current Active LTS release, supported until **April 2027**. It's the conservative choice — Node 24 is newer and riskier for a large codebase.

## What changed

- `.nvmrc`: `20.20.0` → `22.22.3`
- `package.json` `engines.node`: `20.20.0` → `22.22.3`

The `setupNode` composite action reads `.nvmrc` via `node-version-file`, so this single change propagates to every workflow automatically.

## Test plan

- [ ] CI passes (lint, typecheck, tests) — the main signal that nothing in the toolchain regressed on Node 22
- [ ] Local: `nvm install && nvm use && node --version` shows `v22.22.3`
- [ ] No new engine warnings from `npm install`

**No changes to application code.** `.nvmrc` and `package.json` engines only.

Made with [Cursor](https://cursor.com)


mountiny:
I think we might actually handle this update as part of the node updates

PR #92154, opened 2026-05-31
fix: KYC flow doesn't trigger after adding bank account when paying another user

samranahm:
### Explanation of Change

When paying another user with a personal bank account, the KYC flow wasn't triggering after adding the account. The cause was `AccountFlowEntryPoint` calling `clearPersonalBankAccount()` on mount, which wiped the `onSuccessFallbackRoute` that the pay flow sets before this screen and that `AddPersonalBankAccountPage` reads after it, so the app could no longer continue into KYC.

To fix this we:

1. Thread `onSuccessFallbackRoute` through the KYC flow, pay/search/settlement flows pass `personalBankAccountOnSuccessFallbackRoute` into `triggerKYCFlow`, and Wallet uses `getPersonalBankAccountOnSuccessFallbackRoute`, both store the route in `PERSONAL_BANK_ACCOUNT` via `openPersonalBankAccountSetupView` when the user selects "Personal bank account".

2. Preserve the route on entry, `AccountFlowEntryPoint` still clears stale `PERSONAL_BANK_ACCOUNT` state on mount (for deep links and any edge case leftover data), but passes `onSuccessFallbackRoute` into `clearPersonalBankAccount` so the KYC continuation route survives.
3. Reset cleanly between flows, `openPersonalBankAccountSetupView` now use `Onyx.set` instead of `merge`, so each new flow starts with only the fields explicitly passed.

### Fixed Issues

$ https://github.com/Expensify/App/issues/91873
PROPOSAL: https://github.com/Expensify/App/issues/91873#issuecomment-4558847650

### Tests

**Precondition:**
- The user does not have a personal bank account connected.

**Test 01**
1. Start a 1:1 conversation with another user.
2. Press the + button and select Pay [User].
3. On the amount screen, tap the currency selector and choose USD, then enter any amount and continue.
4. Verify that the payment method dropdown displays "Pay with personal account".
5. Press "Pay with personal account" and verify that the bank account connection flow starts.
6. Add a bank account using either Plaid or manual entry, then press Continue on the "Bank account added" page.
7. Verify the KYC verification flow is triggered (confirm additional details to process the payment)

**Test 02**
1. Go to Settings > Wallet and press "Add bank account".
2. Select "Get reimbursed", then choose United States as the country.
3. Enter the bank account details (Plaid or manual entry) and press Continue.
4. On the "Let's double-check..." page, press Confirm and verify the bank account is added (the "Bank account added" page is shown).
5. Press Continue and verify that the user is returned to Add bank account entry point (Streamline payments) page, the KYC flow is not triggered and the additional details page does not open.

### Offline tests

N/A – Bank account setup can only be run while online.

### QA Steps

Same as test

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/8c3d5b14-0b22-4b1a-8be3-56336a9f5e28

</details>

<details>

https://github.com/user-attachments/assets/e144c23a-a959-4acf-8fa3-33c1d22924f0

</details>

<details>

https://github.com/user-attachments/assets/c020d689-f87a-4f9c-806d-6cd32d733bcb

</details>

<details>

https://github.com/user-attachments/assets/cadf5034-d2ec-47c3-9aee-00403943fe56

</details>

<details>

https://github.com/user-attachments/assets/134f2f71-12de-4876-9cf9-4910e66e9210

</details>


src/libs/actions/BankAccounts.ts
```diff
@@ -291,9 +291,19 @@ function setPersonalBankAccountContinueKYCOnSuccess(onSuccessFallbackRoute: Rout
...
+            shouldShowSuccess: null,
+            isLoading: null,
+            errors: null,
+            errorFields: null,
+            updateError: null,
+        });
+    } else {
+        Onyx.set(ONYXKEYS.PERSONAL_BANK_ACCOUNT, null);
+    }
     Onyx.set(ONYXKEYS.FORMS.PERSONAL_BANK_ACCOUNT_FORM_DRAFT, null);
```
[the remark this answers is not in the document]
samranahm: `PERSONAL_BANK_ACCOUNT_FORM_DRAFT` holds form input fields `account number`/`routing number`, not navigation metadata. The navigation fields we're preserving live only in `PERSONAL_BANK_ACCOUNT`. Clearing the draft is always correct here since `AccountFlowEntryPoint` is the entry point before any form data is entered.

src/libs/actions/BankAccounts.ts
```diff
@@ -291,9 +291,13 @@ function setPersonalBankAccountContinueKYCOnSuccess(onSuccessFallbackRoute: Rout
...
 
-function clearPersonalBankAccount(shouldPreserveRoutingData = false) {
+/**
+ * Clears personal bank account state. Pass `shouldPreserveAccountData=true` to only clear UI/error
+ * fields while keeping routing and context fields intact.
+ */
+function clearPersonalBankAccount(shouldPreserveAccountData = false) {
     clearPlaid();
-    if (shouldPreserveRoutingData) {
+    if (shouldPreserveAccountData) {
```
Eskalifer1: I don't think we should do it this way, because we currently have cleanup logic throughout the entire project, so I think we should keep it.

I think that when we added the cleanup logic, we simply forgot that `onSuccessFallbackRoute` shouldn’t be cleaned up, so I think we should just fix that. If we leave it as it is now, we’ll definitely get regressions and incorrect behavior.

I think it should look something like this:
```ts
function clearPersonalBankAccount(preservedData?: Partial<PersonalBankAccount>) {
    clearPlaid();
    Onyx.set(ONYXKEYS.PERSONAL_BANK_ACCOUNT, preservedData ?? null);
    Onyx.set(ONYXKEYS.FORMS.PERSONAL_BANK_ACCOUNT_FORM_DRAFT, null);
    clearPersonalBankAccountSetupType();
}
```

And pass it like this:
```ts
   useEffect(() => {
        clearPersonalBankAccount(personalBankAccount?.onSuccessFallbackRoute ? {onSuccessFallbackRoute: personalBankAccount.onSuccessFallbackRoute} : undefined);
    }, []);
```

What do you think about that?

samranahm: Agreed, that's much cleaner approach. Updated `clearPersonalBankAccount` to take `preservedData?: Partial<PersonalBankAccount>` with `Onyx.set(preservedData ?? null)`, and the entry point now passes only `onSuccessFallbackRoute`. All other call sites stay as a full reset, so no behavior change of any kind there.

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
```diff
@@ -39,10 +39,12 @@ function AccountFlowEntryPoint({policyName = '', onBackButtonPress}: AccountFlow
...
 
     const [isPlaidDisabled] = useOnyx(ONYXKEYS.IS_PLAID_DISABLED);
+    const [personalBankAccount] = useOnyx(ONYXKEYS.PERSONAL_BANK_ACCOUNT);
+    const onSuccessFallbackRoute = personalBankAccount?.onSuccessFallbackRoute;
 
     useEffect(() => {
-        clearPersonalBankAccount();
-    }, []);
+        clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
+    }, [onSuccessFallbackRoute]);
```
Eskalifer1: I don't think it's the best idea to add `onSuccessFallbackRoute` to the deps; this could lead to unexpected consequences, as well as potential duplicate cleans—we only want to perform the clean on the first mount.

I suggest adding a check for `isLoadingOnyxValue` by creating a separate constant and passing it to the dependency; if this constant is `true`, then do not perform the cleanup. This way, we ensure that the `onSuccessFallbackRoute` value will be present if it was there before, and we won’t trigger unnecessary cleanups.

```ts
const [personalBankAccount, personalBankAccountResult] = useOnyx(ONYXKEYS.PERSONAL_BANK_ACCOUNT);

const isLoadingPersonalBankAccount = isLoadingOnyxValue(personalBankAccountResult)

useEffect(() => {
    if (isLoadingPersonalBankAccount) {
        return;
    }
    clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
}, [isLoadingPersonalBankAccount]);
```

Could you please check if this approach works?

Eskalifer1 (commented):
Please fill out the `Explanation of Change` section

Let's also add some information to the `Offline tests` section, specifically “N/A” and the reason why there are no offline tests here

Eskalifer1:
I also think we should update the `preconditions`, specifically to specify that both users must have currency in USD:
`Both users in USD currency default`

src/libs/actions/BankAccounts.ts
```diff
@@ -291,9 +291,13 @@ function setPersonalBankAccountContinueKYCOnSuccess(onSuccessFallbackRoute: Rout
     Onyx.merge(ONYXKEYS.PERSONAL_BANK_ACCOUNT, {onSuccessFallbackRoute});
 }
 
-function clearPersonalBankAccount() {
+/**
+ * Clears personal bank account state. Pass routing metadata to preserve when resetting the flow
+ * without losing navigation context.
+ */
```
Eskalifer1, line 323:
```suggestion
/**
 * Clears personal bank account state, Plaid data, and the form draft.
 * Pass `preservedData` to retain specific fields (e.g. routing metadata) across the reset.
 */
```

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
```diff
@@ -39,10 +39,12 @@ function AccountFlowEntryPoint({policyName = '', onBackButtonPress}: AccountFlow
     const expensifyIcons = useMemoizedLazyExpensifyIcons(['Bank', 'Connect', 'Lightbulb', 'Lock']);
 
     const [isPlaidDisabled] = useOnyx(ONYXKEYS.IS_PLAID_DISABLED);
+    const [personalBankAccount] = useOnyx(ONYXKEYS.PERSONAL_BANK_ACCOUNT);
+    const onSuccessFallbackRoute = personalBankAccount?.onSuccessFallbackRoute;
 
     useEffect(() => {
-        clearPersonalBankAccount();
-    }, []);
+        clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
```
Eskalifer1, line 54: Let's add coment why we pass `onSuccessFallbackRoute` here

samranahm:
Thanks @Eskalifer1 for taking a look, I'm looking into this and will update the PR shortly.

Eskalifer1:
Sure! Let me know when this PR will be ready for next round of reviews!

Eskalifer1:
@codex review

Eskalifer1:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/9ad53544-0449-4bfb-8774-ed8722b74e08




</details>

<details>

https://github.com/user-attachments/assets/410c3a34-75a9-4dba-8373-eb9bb8fce6fc




</details>

<details>

https://github.com/user-attachments/assets/fb2d3712-231b-49ee-83a7-24b2a45ce773




</details>

<details>

https://github.com/user-attachments/assets/cb07c660-610f-4744-909a-a64d9826bc8b




</details>

<details>

https://github.com/user-attachments/assets/00668bb9-40f0-4666-8d1d-f9b5fb611bd4




</details>

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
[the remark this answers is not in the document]
Eskalifer1: This is valid concern. I think we should find a solution to this. Although it's unlikely, it's still possible

samranahm:
@Eskalifer1 Preserving `onSuccessFallbackRoute` approach worked, but it indeed leaves an old route saved after the user leaves without finishing. This is most unlikely to happen, but it can still cause a regression, and we should consider it an edge case.

We actually merge `onSuccessFallbackRoute` in `setPersonalBankAccountContinueKYCOnSuccess`, which we call before entering the bank account flow (pressing "Add personal bank account" in our case). Then, as soon as we land on `AccountFlowEntryPoint` after country selection, the whole `PERSONAL_BANK_ACCOUNT` was wiped to start fresh, which is what removed the route too early.

**Here's what I found to be the most professional solution**

Wipe `PERSONAL_BANK_ACCOUNT` on unmount, so we keep the `onSuccessFallbackRoute` that we save before entering the `AccountFlowEntryPoint` screen, and clear everything when the user finishes setup or goes back and restarts. This is possible because `AccountFlowEntryPoint` stays mounted while the user moves forward through the rest of the flow (it's under the same navigation stack), so it only unmounts and clears when the flow closes or the user goes back that is exactly what we wanted here.

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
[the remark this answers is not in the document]
samranahm: Updated the logic to handle this edge case.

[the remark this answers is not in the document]
Eskalifer1: Hi @samranahm, I don't really think that's the approach we want to take. After all, it makes more sense to clear the previous data on entry (which is much more stable) rather than on exit—plus, that's the logic that's been working for a very long time.

Can you check how `onSuccessFallbackRoute` was being cleared before this `useEffect` was added? Maybe that will give us a little more information.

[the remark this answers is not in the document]
Eskalifer1: @MelvinBot How can this be fixed using this approach:
```ts
const [personalBankAccount, personalBankAccountResult] = useOnyx(ONYXKEYS.PERSONAL_BANK_ACCOUNT);

const isLoadingPersonalBankAccount = isLoadingOnyxValue(personalBankAccountResult)

useEffect(() => {
    if (isLoadingPersonalBankAccount) {
        return;
    }
    clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
}, [isLoadingPersonalBankAccount]);
```

How was `onSuccessFallbackRoute` cleared before this useEffect was added?

samranahm:
> I also think we should update the `preconditions`, specifically to specify that both users must have currency in USD: `Both users in USD currency default`

@Eskalifer1 This one is not needed because we only care about the `IOU` currency because of the `canUseWallet` check. The recipient's currency doesn't matter, and instead of changing default currency, the simplest and most reliable way is to select USD on the amount screen during the Pay flow.

I've updated the test step to reflect this and added a new test as well.

Eskalifer1:
@MelvinBot i was asking how we can fix [this](https://github.com/Expensify/App/pull/92154#discussion_r3348428788) and how it was working before `useEffect` with clear was added

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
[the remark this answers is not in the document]
MelvinBot: **Short version:** `onSuccessFallbackRoute` has never had a dedicated clear — it's only ever wiped as a side effect of a *full* `clearPersonalBankAccount()` (which nulls the whole `PERSONAL_BANK_ACCOUNT` key). The entry-point mount clear was one of those full wipes, and it was the **only** thing protecting this screen from a stale route. So the `isLoadingOnyxValue` version fixes your deps concern, but on its own it does **not** close the stale-route hole `chatgpt-codex-connector` flagged in this thread — you need one more piece for that.

---

### Q2 — How was `onSuccessFallbackRoute` cleared before this `useEffect`?

There is no route-specific clear anywhere in the codebase. The field's whole lifecycle:

- **Set** before a pay/KYC flow begins, via [`setPersonalBankAccountContinueKYCOnSuccess`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/libs/actions/BankAccounts.ts#L287-L292) — called from [`SettlementButton`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/components/SettlementButton/index.tsx#L565), [`PaymentUtils`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/libs/PaymentUtils.ts#L262), [`Search`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/libs/actions/Search.ts#L1335), [`WalletPage`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/pages/settings/Wallet/WalletPage/index.tsx#L814).
- **Consumed** on success by `exitFlow(true)` → [`continueSetup`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/libs/actions/PaymentMethods.ts#L39-L48) (it reads the route to navigate; it does not clear it).
- **Cleared** only by full `clearPersonalBankAccount()` calls scattered through the flow: the [`AccountFlowEntryPoint`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx#L43-L44) mount clear (the long-standing one you're referring to), plus [`AddBankAccount.tsx:73`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/pages/EnablePayments/AddBankAccount/AddBankAccount.tsx#L73), [`AddPersonalBankAccountPage.tsx:52`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/pages/AddPersonalBankAccountPage.tsx#L42-L54), [`PersonalInfo.tsx:34-37`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/pages/EnablePayments/PersonalInfo/PersonalInfo.tsx#L34-L37) (KYC continuation), [`FeesAndTerms.tsx:27`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/pages/EnablePayments/FeesAndTerms/FeesAndTerms.tsx#L27) (end of flow).

So on `main`, entering this screen always nuked the entire object. That wipe is simultaneously (a) what kept a stale route from leaking into a later flow, and (b) what wiped the route this PR needs to keep — the bug. **That's the tension: the moment you preserve the route on entry, you remove the thing that used to protect against stale routes.** Note [`openPersonalBankAccountSetupView({})`](https://github.com/Expensify/App/blob/ebebf0d94fdbc092ad0e81204a3411577b8d5cd3/src/libs/actions/BankAccounts.ts#L119-L140) — the generic Wallet → Add bank account entry — does **not** clear `PERSONAL_BANK_ACCOUNT`, which is precisely why `chatgpt-codex-connector`'s scenario is reachable.

### Q1 — Making the `isLoadingOnyxValue` approach work

It needs two pieces, not one.

<details>
<summary><b>Piece A — the entry-point effect (your version, corrected)</b></summary>

```ts
const [personalBankAccount, personalBankAccountResult] = useOnyx(ONYXKEYS.PERSONAL_BANK_ACCOUNT);
const isLoadingPersonalBankAccount = isLoadingOnyxValue(personalBankAccountResult);
const onSuccessFallbackRoute = personalBankAccount?.onSuccessFallbackRoute;

useEffect(() => {
    if (isLoadingPersonalBankAccount) {
        return;
    }
    clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
}, [isLoadingPersonalBankAccount]);
```
with `clearPersonalBankAccount(preservedData?: Partial<PersonalBankAccount>)` doing `Onyx.set(PERSONAL_BANK_ACCOUNT, preservedData ?? null)`.

Two things to get right:
- `onSuccessFallbackRoute` must be **derived** from `personalBankAccount?.onSuccessFallbackRoute` (your snippet references it but doesn't declare it).
- The `isLoadingOnyxValue` guard is **load-bearing**, not cosmetic: `useOnyx` returns `undefined` while the key hydrates. Without the guard, the first render reads `undefined` → `clearPersonalBankAccount(undefined)` → `Onyx.set(..., null)` → the route is wiped again (the original bug). Gating on hydration guarantees you read the real route before clearing, and keeping the dep array as `[isLoadingPersonalBankAccount]` only means the re-seed write won't re-trigger the effect (which is also what fixes your duplicate-clean concern). ✅
</details>

<details>
<summary><b>Piece B — close the stale-route hole (the concern this thread is about)</b></summary>

Piece A alone still preserves the route on **every** entry, so the abandon path you and `chatgpt-codex-connector` discussed is still open: set route → land here → preserve → back out (nothing clears it) → later generic flow reuses the stale route.

Since you (rightly) want to keep clear-on-entry rather than move it to unmount, the targeted fix is to **wipe fully when the flow is actually abandoned** at this entry screen. Wrap the back handler so leaving without proceeding does a full reset:

```ts
const handleBackButtonPress = () => {
    clearPersonalBankAccount(); // full wipe — abandoning the flow must not leave a stale route
    onBackButtonPress();
};
```
and pass `handleBackButtonPress` to `HeaderWithBackButton`. That preserves the route while moving *forward* through the flow, but clears it on a real abandon — satisfying both "clear on entry is stable" and "no stale route."
</details>

---

To be transparent: I verified the data flow and call sites directly from the code (links above), but I did **not** reproduce the stale-route scenario in a running app — it's plausible and you already agreed it's valid, just flagged here as analysis rather than a confirmed repro.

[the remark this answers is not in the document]
Eskalifer1: @samranahm I think we can try to use this approach

samranahm:
@Eskalifer1 PR ready for another review, please take a look. Unit test failure is not related to this PR.

Eskalifer1:
Hi @samranahm take a look [here](https://github.com/Expensify/App/pull/92154#discussion_r3351911278) pls

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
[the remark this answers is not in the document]
samranahm: @Eskalifer1 We have same approach in `AddPersonalBankAccountPage` to clear the data on unmount, this make sure we clear everything when flow finish to make sure each new flow start fresh. 

https://github.com/Expensify/App/blob/00b560618bee575adc9a1875d043e57f8073314d/src/pages/AddPersonalBankAccountPage.tsx#L56

This is already established approach through out the codebase.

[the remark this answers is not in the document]
Eskalifer1: It's already 1AM for me, so i will take a closer look tomorrow

Eskalifer1:
@codex review

samranahm:
@Eskalifer1 It's midnight for me, I'll continue working on this tomorrow.

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
```diff
@@ -40,9 +40,8 @@ function AccountFlowEntryPoint({policyName = '', onBackButtonPress}: AccountFlow
 
     const [isPlaidDisabled] = useOnyx(ONYXKEYS.IS_PLAID_DISABLED);
 
-    useEffect(() => {
-        clearPersonalBankAccount();
-    }, []);
+    // Clear on unmount so fallback route remains available during the next flow and is clear when the flow ends.
+    useEffect(() => clearPersonalBankAccount, []);
```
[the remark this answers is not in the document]
Eskalifer1: Hi @samranahm Yes, I think that's a valid concern. The problem here is that it also occurs during cleanup on mount.

I think we should revert to the logic I proposed [here](https://github.com/Expensify/App/pull/92154#discussion_r3348194221). That way, we’ll fix the bug and avoid new regressions.

I'm convinced that cleaning up on unmount is a very bad idea and will introduce a lot of new regressions, since the app has been working this way for months—cleaning up data on mount—and changing that now just because of one small bug is a bad idea.

At the very least, I can say that if the user simply closes the page, the cleanup in useEffect won’t run, so the next time the user starts the flow, all their data will be stale, not just `onSuccessFallbackRoute`—so this is a bad idea, and that’s just one example.

We need to perform cleanup on mount, as is currently the case.

Perhaps we’ll find a solution once we figure out how cleanup worked before this cleanup was added; please check how it worked back then.

Please do not use AI; it’s very noticeable, and AI might give an incorrect result, which will only confuse us further.

We can add a separate cleanup on unmount just for `onSuccessFallbackRoute`; this way, we’ll preserve the existing logic and fix the bug:
```ts
    useEffect(() => {
        if (isLoadingPersonalBankAccount) {
            return;
        }
        clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
    }, [isLoadingPersonalBankAccount]);

    useEffect(() => () => setPersonalBankAccountContinueKYCOnSuccess(null), []);
```

Let me know what do you think

Eskalifer1:
@MelvinBot Here's the bug we're dealing with:

Bug: When paying another user with a personal bank account, the KYC flow does not trigger after adding the account.

Root Cause
AccountFlowEntryPoint calls clearPersonalBankAccount() on mount, which sets ONYXKEYS.PERSONAL_BANK_ACCOUNT to null. This wipes onSuccessFallbackRoute — a value written by the pay flow (via setPersonalBankAccountContinueKYCOnSuccess) before navigating to this screen. AddPersonalBankAccountPage and AddBankAccount read onSuccessFallbackRoute after the bank is added to decide where to redirect the user (e.g. into the KYC/enable-payments flow). With it gone, the redirect never happens.

We want to fix this by clearing everything in `useEffect` except for `onSuccessFallbackRoute`, but this causes another bug:
onSuccessFallbackRoute is now preserved in Onyx indefinitely(if user doesn't finish flow). If the user abandons the flow mid-way and later opens the add-bank-account flow from an unrelated part of the app (e.g. Wallet settings), AccountFlowEntryPoint mounts again, isLoadingPersonalBankAccount transitions from true → false, and the effect reads the stale onSuccessFallbackRoute from the previous session — preserving it again instead of discarding it. The user ends up being redirected to a route that has nothing to do with their current intent.

Do you have any idea how we can fix that?

Is it good to clear only `onSuccessFallbackRoute` on unmount? Like this:
```ts
useEffect(() => () => setPersonalBankAccountContinueKYCOnSuccess(null), []);
```

Eskalifer1:
@samranahm Take a look at Option A that Melvin suggested. What do you think of it?

It seems like the best idea is to do the cleanup first and then handle the navigation, but it looks like that would require rewriting quite a bit of code.

Eskalifer1:
@MelvinBot  We have two options:
1. Remove the `useEffect` with cleanup from `AccountFlowEntryPoint` and perform the cleanup before the flow begins

2. Keep the approach I suggested (full cleanup except for `onSuccessFallbackRoute`), but add the `onSuccessFallbackRoute` cleanup specifically wherever the flow is initiated.

Which of these two options is better? And which of these options falls within the scope of this PR?

Eskalifer1:
@MelvinBot I understand that we need to add this cleanup everywhere the flow can start. Could you help identify all of those entry points?

From my understanding, the cleanup should be added not only in places where we set `onSuccessFallbackRoute`, but also in flows where it isn't set. Otherwise, if a user exits midway through a flow that has `onSuccessFallbackRoute` and then starts another flow that doesn't, the stale value could remain and the cleanup logic might not work correctly.


Link all places where we need to add this cleanup. Maybe we can do something so we don't have to repeat the same thing in every place where logic related to `AccountFlowEntryPoint` is used, but instead do it just once somewhere?

Eskalifer1:
@samranahm Please take a look at the solutions Melvin has suggested to see which one works best. Do you perhaps have another solution that would work better?

samranahm:
Looking into it, I'll explore each one and will let you know about my findings regarding these.

Eskalifer1:
@samranahm Thank you! I'll be OOO part of the day tomorrow and all next week, so I'll review everything once I'm back. If you have any higher-priority tasks, feel free to work on them now!

samranahm:
Merging the latest `main` to pick up the new `no-unsafe-type-assertion` ESLint rule ([Slack](https://expensify.slack.com/archives/C01GTK53T8Q/p1780945507720319)).

Eskalifer1:
Hi @samranahm did you able to find reliable solution to this one?

samranahm:
@Eskalifer1 Yes, I am combining the flow at a single source of truth and removing the clearing of previous state data from mount/unmount entirely. Now it lives inside the openPersonalBankAccountSetupView entry point, where we pass onSuccessFallbackRoute only where we need to continue the KYC flow. Since openPersonalBankAccountSetupView runs before each flow in the app, we always make sure the new flow starts fresh, so there's no risk of stale data that might have been left behind with the previous approach where we cleared it on unmount.

Testing it locally will push the changes shortly.

src/components/KYCWall/BaseKYCWall.tsx
```diff
@@ -255,7 +264,7 @@ function KYCWall({
      */
     const continueAction = useCallback(
         (params?: ContinueActionParams) => {
-            const {event, iouPaymentType, paymentMethod, policy, goBackRoute} = params ?? {};
+            const {event, iouPaymentType, paymentMethod, policy, goBackRoute, personalBankAccountOnSuccessFallbackRoute} = params ?? {};
```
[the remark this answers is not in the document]
samranahm: Updated, When we skip the bank-account setup and go straight to enable payments, we now save `onSuccessFallbackRoute` in Onyx so `AddBankAccount` can continue into KYC after the user adds their bank account instead of going back to Wallet page.

samranahm:
@codex review

samranahm:
@Eskalifer1 I've implemented the updated solution and tested it deeply and it's working as expected, Please take a look.

Eskalifer1:
@codex review

Eskalifer1:
@MelvinBot this branch was updated. Do you see any valid concern about new approach?

Maybe we didn't implement this logic everywhere, or are there any potential regressions? Check that as well—remember that quality is more important than speed, so take your time with your response.

src/pages/settings/Wallet/WalletPage/index.tsx
```diff
@@ -806,13 +800,9 @@ function WalletPage() {
                                     <KYCWall
                                         ref={kycWallRef}
                                         onSuccessfulKYC={(_iouPaymentType?: PaymentMethodType, source?: Source) => navigateToWalletOrTransferBalancePage(source)}
-                                        onSelectPaymentMethod={(selectedPaymentMethod: string) => {
-                                            if (hasActivatedWallet || selectedPaymentMethod !== CONST.PAYMENT_METHODS.PERSONAL_BANK_ACCOUNT) {
-                                                return;
-                                            }
-                                            // To allow upgrading to a gold wallet, continue with the KYC flow after adding a bank account
```
Eskalifer1, line 813: Let's revert this comment, since it might be usefull in the future

src/libs/actions/BankAccounts.ts
```diff
@@ -117,17 +120,34 @@ function setPlaidEvent(eventName: string | null) {
...
-            Onyx.merge(ONYXKEYS.PERSONAL_BANK_ACCOUNT, {source});
+            personalBankAccountState.source = source;
+        }
+        if (onSuccessFallbackRoute) {
+            personalBankAccountState.onSuccessFallbackRoute = onSuccessFallbackRoute;
         }
+
+        // Seed state via Onyx.set to avoid set/merge races (see resetPersonalBankAccountForUpdate).
+        Onyx.set(ONYXKEYS.PERSONAL_BANK_ACCOUNT, Object.keys(personalBankAccountState).length > 0 ? personalBankAccountState : null);
+        Onyx.set(ONYXKEYS.FORMS.PERSONAL_BANK_ACCOUNT_FORM_DRAFT, null);
```
Eskalifer1, line 152: Why do we add this?

src/components/KYCWall/BaseKYCWall.tsx
Eskalifer1, line 322: Let's add comment here why we don't pass `personalBankAccountOnSuccessFallbackRoute` here

samranahm:
@MelvinBot I intentionally removed the mount `clearPersonalBankAccount()` call because It isn’t needed anymore because every normal path to `AccountFlowEntryPoint` already goes through `openPersonalBankAccountSetupView` first, which sets up `PERSONAL_BANK_ACCOUNT` with only the fields needed for the current flow (including `onSuccessFallbackRoute`).

Eskalifer1:
Hi @samranahm, I'm reviewing this right now, so please don't push any new changes yet. I'll let you know which concerns I consider valid and which I don't, as well as what questions I have.

src/components/KYCWall/BaseKYCWall.tsx
```diff
@@ -81,6 +83,7 @@ function KYCWall({
     const transferBalanceButtonRef = useRef<HTMLDivElement | View | null>(null);
 
     const [shouldShowAddPaymentMenu, setShouldShowAddPaymentMenu] = useState(false);
+    const [pendingPersonalBankAccountOnSuccessFallbackRoute, setPendingPersonalBankAccountOnSuccessFallbackRoute] = useState<Route | undefined>();
```
Eskalifer1: Why are we using `undefined` here instead of `null`? Is there a reason for this? It seems like `null` would be a better fit here, since we’ll later set this value to empty?

I’d also suggest adding a default value to `useState` right away (`useState<Route | null>(null);`)

Eskalifer1: I think it’s worth adding a detailed comment here explaining why we need this state and which flow it’s used for.

src/pages/settings/Wallet/WalletPage/index.tsx
```diff
@@ -806,13 +800,9 @@ function WalletPage() {
...
-                                            setPersonalBankAccountContinueKYCOnSuccess(ROUTES.SETTINGS_WALLET);
-                                        }}
+                                        getPersonalBankAccountOnSuccessFallbackRoute={(selectedPaymentMethod) =>
+                                            !hasActivatedWallet && selectedPaymentMethod === CONST.PAYMENT_METHODS.PERSONAL_BANK_ACCOUNT ? ROUTES.SETTINGS_WALLET : undefined
+                                        }
```
Eskalifer1, line 808: I think we should add a detailed comment here explaining why we need this prop and why we can't use the same approach as elsewhere.

src/libs/actions/BankAccounts.ts
```diff
@@ -117,17 +120,33 @@ function setPlaidEvent(eventName: string | null) {
...
         }
         if (source) {
-            Onyx.merge(ONYXKEYS.PERSONAL_BANK_ACCOUNT, {source});
+            personalBankAccountState.source = source;
+        }
+        if (onSuccessFallbackRoute) {
+            personalBankAccountState.onSuccessFallbackRoute = onSuccessFallbackRoute;
         }
+
+        Onyx.set(ONYXKEYS.PERSONAL_BANK_ACCOUNT, Object.keys(personalBankAccountState).length > 0 ? personalBankAccountState : null);
```
Eskalifer1, line 151: Let's add a comment explaining why we're using `set` here instead of `merge`, since it's not obvious at first glance.

Eskalifer1 (commented):
Overall, everything looks very good; there are a few comments, but they aren't too critical

From what Melvin wrote:

1. Undocumented behavior change — I think we should check whether there’s any data in PERSONAL_BANK_ACCOUNT in the places where openPersonalBankAccountSetupView({}) is called, since this function used to only handle navigation but now overwrites data. Therefore, we should add this to the jsDoc description

2. Test coverage gap—this also needs to be fixed; currently, our test coverage is lower than it should be. You can check it [here](https://github.com/Expensify/App/pull/92154#issuecomment-4587468685)

3. Removing the `AccountFlowEntryPoint` mount-clear—I think this could indeed be a problem, as Melvin pointed out. We’ve now closed off all entry points where the flow is triggered using `AccountFlowEntryPoint`, so from the UI side, the user won’t be able to access `AccountFlowEntryPoint` in any way that would allow it to contain any other random data. However, none of this will work if a user has data in `PERSONAL_BANK_ACCOUNT` and navigates to the `AccountFlowEntryPoint` page not via the UI but through a link—this is an unlikely scenario, but still possible, since deep linking is an actively supported feature of the app. What do you think of the solution Melvin proposed? Clearing everything except `onSuccessFallbackRoute`? Will this break anything? It seems like it shouldn’t affect the bug fix, but it should also preserve the behavior for deep links. Could you please check this?

Eskalifer1:
Hi @samranahm. Seems like you choose a good approach, i have left few comments.

All yours! 

Let me know when i can do next round of reviews 🙏

samranahm:
Thanks a lot @Eskalifer1 for the detailed review. I'm currently working on another PR and will get to this one in about an hour or two. I'll go through all the comments shortly and update the PR accordingly.

src/pages/settings/Wallet/WalletPage/index.tsx
samranahm: Done.

src/libs/actions/BankAccounts.ts
samranahm: @Eskalifer1 Previously this function only merged specific fields, so stale data (like a leftover `onSuccessFallbackRoute` from an abandoned pay/KYC flow) could survive into a new flow. Using set replaces the whole object with only what we explicitly pass.

src/components/KYCWall/BaseKYCWall.tsx
samranahm: Done.

src/libs/actions/BankAccounts.ts
Eskalifer1: Sorry, i meant why do we add cleanup here for `PERSONAL_BANK_ACCOUNT_FORM_DRAFT `?

src/components/KYCWall/BaseKYCWall.tsx
samranahm: Good point, actually no strong reason to use `undefined` here. Updated to `useState<Route | null>(null)` and reset it with `setPendingPersonalBankAccountOnSuccessFallbackRoute(null)` everywhere we clear it. null is a better fit since we're explicitly representing "no pending route."

samranahm: Done.

src/pages/settings/Wallet/WalletPage/index.tsx
samranahm: Done

src/libs/actions/BankAccounts.ts
samranahm: Done.

src/libs/actions/BankAccounts.ts
samranahm: `PERSONAL_BANK_ACCOUNT_FORM_DRAFT` holds progress form user input. If a user abandoned a previous attempt, that draft can still be in Onyx when they start a new "add bank account" flow. So clearing it here makes sure each entry starts with a clean draft form, same as `clearPersonalBankAccount()` does on other entry paths.

samranahm:
@Eskalifer1 PR ready for another round of review.

flaviadefaria (dismissed):
From a product standpoint, the fix looks good to me.

Eskalifer1:
@codex review

Eskalifer1:
@MelvinBot we have updated PR, can you take another look, do we cover all cases? Maybe we have introduced some potential regressions, check this one please. Remember quality > time response

src/libs/actions/BankAccounts.ts
Eskalifer1: Okay, got it, you can resolve it!

src/components/KYCWall/BaseKYCWall.tsx
Eskalifer1: Good! You can resolve it!

Eskalifer1:
Hi @Expensify/design I noticed that during the flow for adding a bank account on the Wallet page, after we've entered all the information in the “Plaid” section, at the bank selection stage, we see this UI. It looks like there's some extra spacing on the sides—should this be fixed?

Знімок екрана 2026-06-18 о 17 45 36

Eskalifer1:
Looking good, completing checklist!

Eskalifer1 (approved):
LGTM!

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
```diff
@@ -39,10 +40,20 @@ function AccountFlowEntryPoint({policyName = '', onBackButtonPress}: AccountFlow
...
-    }, []);
+        if (isLoadingPersonalBankAccount) {
+            return;
+        }
+
+        // Clear stale flow state on entry while preserving onSuccessFallbackRoute if it was set before entering this screen (e.g. from a pay/KYC flow or deep link).
+        // openPersonalBankAccountSetupView also resets state, but this handles direct navigation to this screen.
+        clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- only run once after Onyx hydration, not when the route changes from our own write
+    }, [isLoadingPersonalBankAccount]);
```
chuckdries, line 56: is this eslint-disable-next-line strictly necessary? I think that `clearPersonalBankAccount` preserves the `onSuccessFallbackRoute` you passed it, so including that here shouldn't cause the effect to re-run I think

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
```diff
@@ -39,10 +40,20 @@ function AccountFlowEntryPoint({policyName = '', onBackButtonPress}: AccountFlow
...
     useEffect(() => {
-        clearPersonalBankAccount();
-    }, []);
+        if (isLoadingPersonalBankAccount) {
+            return;
+        }
+
+        // Clear stale flow state on entry while preserving onSuccessFallbackRoute if it was set before entering this screen (e.g. from a pay/KYC flow or deep link).
+        // openPersonalBankAccountSetupView also resets state, but this handles direct navigation to this screen.
+        clearPersonalBankAccount(onSuccessFallbackRoute ? {onSuccessFallbackRoute} : undefined);
```
chuckdries, line 54: The PR description says 

> To fix this we update the reset on unmount instead of mount

But this is still running on mount, right? Please ensure that the description matches the implementation in the code

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
samranahm: right, `clearPersonalBankAccount` preserves the same `onSuccessFallbackRoute` value we pass in, so including it in the dependency array doesn't cause a re-run. Updated

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
samranahm: Updated.

src/pages/settings/Wallet/InternationalDepositAccount/subPages/AccountFlowEntryPoint.tsx
Eskalifer1: I don't think this should be in `deps`. The logic is that this `useEffect` should run only once on mount, but still leave `onSuccessFallbackRoute` intact.

Previously, the data was cleared completely, but now we clear everything except `onSuccessFallbackRoute`. We intentionally add a check here for `isLoadingPersonalBankAccount` so that we can perform the cleanup once we have all the information about the bankAccount.

If we add `onSuccessFallbackRoute`, then if there’s any change to `onSuccessFallbackRoute` (which is an edge case, but still potentially possible), this `useEffect` will be triggered again and clear everything except `onSuccessFallbackRoute`.

What do you think guys @samranahm @chuckdries?

samranahm: > If we add `onSuccessFallbackRoute`, then if there’s any change to `onSuccessFallbackRoute` (which is an edge case, but still potentially possible), this `useEffect` will be triggered again and clear everything except `onSuccessFallbackRoute`.
> 

The effect here is meant to run once when the personal bank account data has finished loading, not re-run if `onSuccessFallbackRoute changes later` but I couldn't find any reliable repro steps for that edge case, but theoretically if the route changed while on this screen, cleanup would run again and wipe other `PERSONAL_BANK_ACCOUNT` fields. So to get exact outcome we are expecting here I think we should remove `onSuccessFallbackRoute` from deps to prevent any unexpected regression later. 

@chuckdries You're right that preserving the same value wouldn't cause a loop but the concern here is different we don't want the effect to re-run on any route change, not just avoid infinite loops.

chuckdries: Ok. I guess I'm still a little unclear on what route changes could trigger that, but I don't need to block on that.

chuckdries:
This looks pretty good to me. Sorry to be a stickler, but @Eskalifer1 would you mind running through the tests one more time on the various platforms to make sure that the latest change to the useEffect dependencies array in AccountFlowEntryPoint is safe? Thanks 🙏

Eskalifer1:
Hi @chuckdries Actually, we haven't really changed the logic much compared to `main`, but that's okay—I'll check again on Monday!

samranahm:
removing `onSuccessFallbackRoute` from deps.

shawnborton:
@Eskalifer1 good eye! Yes, let's please fix that spacing/padding issue.

Eskalifer1:
Hi @chuckdries, i have tested all works good(atleast i didn;t find any issue), i think we can procced with this one!

Also desgn team decided we should fix [this issue](https://github.com/Expensify/App/pull/92154#issuecomment-4743167517), can you create new issue for that pls!

I think i can work with Melvin on this one to make sure all works well and i didn't miss something!

Or we can open for external!

samranahm:
When user selects only one account in Plaid, we still show the “Choose your bank account” screen with a single predetermined option in radio list. So user is effectively confirming the same single account they already chose in Plaid, with no real decision to make.

Should we update this to when Plaid returns exactly one eligible account, skip the “Choose your bank account” screen entirely and navigate to next screen with that account. We would keep the current behavior intact when two or more accounts are returned, since that is where the user actually needs to choose.

Screenshot 2026-06-23 at 3 08 14 AM


cc: @chuckdries @shawnborton

Eskalifer1:
> When user selects only one account in Plaid, we still show the “Choose your bank account” screen with a single predetermined option in radio list. So user is effectively confirming the same single account they already chose in Plaid, with no real decision to make.
> 
> Should we update this to when Plaid returns exactly one eligible account, skip the “Choose your bank account” screen entirely and navigate to next screen with that account. We would keep the current behavior intact when two or more accounts are returned, since that is where the user actually needs to choose.

I don't think this is related to this issue and seems like change to logic that is working now. But it's just IMO

shawnborton:
I don't mind that idea, but I also don't think we need to make any changes to the flow - the extra account name confirmation isn't necessarily a bad thing.

chuckdries:
I don't think we should skip the choose account screen here, would be out of scope for this fix.

chuckdries:
Ok, just running through the test steps one last time. Test 1 passes, but for test 2, I noticed that the user is returned to the "Streamline payments" screen, even though we can see underneath that the bank account was successfully added. @samranahm @Eskalifer1 can either of you confirm whether this is a regression in this PR?

https://github.com/user-attachments/assets/10a449b4-ce31-49b6-ae58-016fdfe03a0c

Eskalifer1:
> Ok, just running through the test steps one last time. Test 1 passes, but for test 2, I noticed that the user is returned to the "Streamline payments" screen, even though we can see underneath that the bank account was successfully added. @samranahm @Eskalifer1 can either of you confirm whether this is a regression in this PR?
> 
>  kyc.PR.test.mp4

I see there is the same bahaviour on stage, so it's not regression. Let's update Test case @samranahm and resolve conflict

samranahm:
Agree with @Eskalifer1, we're not changing this behaviour in our PR. Updated the test steps.

samranahm:
Resolved the conflict, @chuckdries all yours.

chuckdries (approved):

Issue #87472, opened 2026-04-09
[$250] Tags - Can trigger delete all Tags after toggle on "Members must tag all expenses"


nlemma:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** v9.3.55-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** Yes
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** https://test-management.browserstack.com/projects/2219752/test-runs/TR-17113/folder/13176676/41237292/1317000710
**Email or phone of affected tester (no customers):** N/A
**Issue reported by:** Applause Internal Team
**Bug source:** Exploratory - Significant User Experience Deterioration
**Device used:** Windows 11/Chrome
**App Component:** Workspace Settings

## Action Performed:
1. Open the NewDot app
2. Navigate to the Workspace settings of a Control WS -> Tags
3. Click on Add tags -> enter any name -> Save
4. Click on the More -> Settings -> Toggle on  "Members must tag all expenses"
5. Click on More -> Import Spreadsheet -> Multi level tags
6. Dismiss the RHP -> More -> Settings ( observe that "Members must tag all expenses" is on and has a lock state)

## Expected Result:
Warning about the "Cannot delete or disable all tags" modal is shown.

## Actual Result:
All Tags deleted, and the "Members must tag all expenses" is on and has a lock state

## Workaround:
Unknown

## Platforms:
- [x] Android: App
- [ ] Android: mWeb Chrome
- [x] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/79d588fc-450a-45a3-863d-0b424d19698e

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022043625636678841601</li>
        <li>Upwork Job ID: 2043625636678841601</li>
[...]

PR #92338, opened 2026-06-02
Fix required tags state after switching tag levels

nabi-ebrahimi:
### Explanation of Change

When a workspace switches between single-level and multi-level tags, the backend clears the existing tags and disables `Members must tag all expenses` if keeping it enabled would leave the workspace in an invalid state.

The problem is that the frontend did not reflect that backend change immediately. This created a confusing experience where the switch flow completed successfully, but the UI still showed `Members must tag all expenses` as enabled until the user left and revisited the workspace settings. In other words, the workspace state was valid on the backend, but the frontend temporarily showed stale information.

This PR fixes that inconsistency by updating the frontend after the tag cleanup request succeeds, so users see the correct required-tags state right away and the UI stays aligned with the backend during the tag-level switching flow.
### Fixed Issues

$ #87472
PROPOSAL: [#87472 (comment)](https://github.com/Expensify/App/issues/87472#issuecomment-4312455748)

### Tests

1. Open the NewDot app.
2. Go to `Workspace Settings` for a Control Workspace → `Tags`.
3. Click `Add tags`, enter any name, and click `Save`.
4. Navigate to `More` → `Settings`, then enable `Members must tag all expenses`.
5. Go to `More` → `Import Spreadsheet` → `Multi-level tags`.
6. Verify the `Switch Tag Levels` warning modal appears, then click `Switch Tag Levels`.
7. Navigate back to `More` → `Settings`.
8. Verify `Members must tag all expenses` is automatically disabled after the switch succeeds.
9. Click Add tags, enter a tag name, and click Save.
10. Wait for the tag change to finish and verify the Members must tag all expenses toggle update automatically.
11. Go to More > Settings and confirm the toggle reflects the new state.

### Offline tests

Same as Tests.

### QA Steps

Same as Tests.

### PR Author Checklist

    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari
- [X] If the PR modifies the UI (e.g. new buttons, new UI components, changing the padding/snabi-ebrahimi:fix/delete-tags-when-toggling-required-tagspacing/sizing, moving components, etc) or modifies the form input styles:

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/965eedf9-233e-47d7-8c06-3aac1d0d7d67

</details>

<details>

>

https://github.com/user-attachments/assets/f2da3c31-a5e6-4af1-93eb-8be1f3ff4247

</details>

<details>

>

https://github.com/user-attachments/assets/29a6fcc8-087e-4239-955e-970f5be7fc21

</details>

<details>

>

https://github.com/user-attachments/assets/ff38c1a0-641f-43b1-86c6-c1fc35fd1449

</details>

<details>

>

https://github.com/user-attachments/assets/f3545646-15cc-4f14-808b-ddedf8011be5

</details>


nabi-ebrahimi:
**Root Cause of the Tags Toggle Regression:**
The root cause is that the tags settings flow was relying on stale Onyx policy data after switching tag levels, so the UI did not pick up the updated `requiresTag` and `areTagsEnabled` state until the screen was remounted or manually refreshed.

More specifically:
- `src/pages/workspace/tags/ImportTagsOptionsPage.tsx` triggers `cleanPolicyTags()` when the user confirms switching from single-level to multi-level tags.
- `cleanPolicyTags()` was clearing `requiresTag` too early, before the backend finished the cleanup, which left the policy in a half-updated state during the transition.
- `src/pages/workspace/tags/WorkspaceTagsSettingsPage.tsx` and `src/pages/workspace/tags/TagSettingsPage.tsx` were only reading from `usePolicyData(policyID)` and were not re-fetching tag data on focus or reconnect, so the screens could stay stuck on the old disabled state until refresh.

**Solution:**
The fix is to make both the backend state transition and the screen refresh explicit.

The changes are:
- In [src/libs/actions/Policy/Tag.ts](/home/h-naiby/Github/Nabi/App/src/libs/actions/Policy/Tag.ts#L849), let `cleanPolicyTags()` accept a `shouldDisableRequiresTag` flag and only clear `requiresTag` in `successData`.
- In [src/pages/workspace/tags/ImportTagsOptionsPage.tsx](/home/h-naiby/Github/Nabi/App/src/pages/workspace/tags/ImportTagsOptionsPage.tsx#L72), pass `policy?.requiresTag === true` into `cleanPolicyTags()` when switching tag levels so the cleanup can update the policy correctly.
- In [src/pages/workspace/tags/WorkspaceTagsSettingsPage.tsx](/home/h-naiby/Github/Nabi/App/src/pages/workspace/tags/WorkspaceTagsSettingsPage.tsx#L64), refetch tag data with `openPolicyTagsPage(policyID)` on focus and reconnect.
- In [src/pages/workspace/tags/TagSettingsPage.tsx](/home/h-naiby/Github/Nabi/App/src/pages/workspace/tags/TagSettingsPage.tsx#L72), do the same refetch on focus and reconnect for the tag detail screen.
- In [tests/actions/PolicyTagTest.ts](/home/h-naiby/Github/Nabi/App/tests/actions/PolicyTagTest.ts#L205), add success/failure coverage for `cleanPolicyTags()` so the `requiresTag` transition stays correct.

This keeps the normal tags flow unchanged, while ensuring the settings UI reflects the backend state without needing a hard refresh.

**Fixed Demo:**


https://github.com/user-attachments/assets/35fa2bab-fe02-470e-aabe-1856f5b79278

hoangzinh:
@nabi-ebrahimi can you quickly explain the root cause of the previous deploy blocker? And your solution to fix it?

nabi-ebrahimi:
> @nabi-ebrahimi can you quickly explain the root cause of the previous deploy blocker? And your solution to fix it?

@hoangzinh, thank you for taking the time to review this. I appreciate your feedback.

**Root Cause of The Previous Deploy Blocker:**

 the tags settings screens were reading from stale Onyx policy data and did not refresh the tag state when returning from the import flow. After the tag cleanup completed, the UI still had the old requiresTag / areTagsEnabled snapshot, so the Tags toggle stayed disabled until a manual refresh remounted the screen.

Solution: keep cleanPolicyTags() updating requiresTag only on success, and add a focus/reconnect refetch via openPolicyTagsPage(policyID) in both WorkspaceTagsSettingsPage and TagSettingsPage so the screens always pick up the latest backend state.


**Reproduced  Deploy Blocker Demo:**

https://github.com/user-attachments/assets/b87f40a8-fd2c-4a6a-85fb-7b79be15ab97

joekaufmanexpensify (dismissed):
Second attempt on the PR

nabi-ebrahimi:
@hoangzinh, friendly bump. thanks.

hoangzinh:
> After the tag cleanup completed, the UI still had the old requiresTag / areTagsEnabled snapshot, so the Tags toggle stayed disabled until a manual refresh remounted the screen.

The RCA doesn't seem strong to me. Let me check

hoangzinh:
@robertjchen Could you help me verify the value of `policy.requiredTag` in the database in the following scenario?

1. Go to a workspace and select "Tags."
2. Add at least one tag.
3. In "Tags Settings," enable "Members must tag all expenses"
4. Navigate to "More" and select "Import spreadsheet" then switch to any tag level.
5. Verify that all tags are clear.
6. Open "Tags Settings" — the FE still displays "Members must tag all expenses" as enabled

=> @robertjchen Can you check the current value of `policy.requiredTag` in the database in this step?

7. Reload the page — the FE now shows "Members must tag all expenses" as disabled.
8. Add a new tag.
9. Open "Tags Settings" again — the FE still shows "Members must tag all expenses" as disabled

=> @robertjchen Can you verify the current value of `policy.requiredTag` in the database in this step?

If in step 6, the value of `policy.requiredTag` in the database is disabled, and in step 9 it is enabled, then I believe there is a BE issue. After step 8, the BE should send a PusherEvent if `policy.requiredTag` value has changed.


https://github.com/user-attachments/assets/754e35bd-d6a8-4b70-862f-f7fb62893c82

Cc @nabi-ebrahimi, I would like to hear your thoughts too.

nabi-ebrahimi:
hi @hoangzinh and @robertjchen 
Based on the behavior I observed,` policy.requiresTag `appears to stay stale on the current page until a refresh, and it only becomes consistent afterward. This suggests the root cause may be a missing or delayed **backend** **update/Pusher** event in this flow, with the **frontend** continuing to display the previous **Onyx** state until it receives the new value.

Thank you.

**Testing Demo:**


https://github.com/user-attachments/assets/6ddda601-e5ab-489d-88f9-630f3ea58cf1


Testing in Offline Mode:


https://github.com/user-attachments/assets/860d4a5f-82c9-438e-8979-0b8d025c79ae

nabi-ebrahimi:
@hoangzinh, gentle bump. thanks.

hoangzinh:
> @hoangzinh, gentle bump. thanks.

Yeah, given we think it's a BE issue. We should wait for @robertjchen to help me confirm on this message https://github.com/Expensify/App/pull/92338#issuecomment-4632923558

robertjchen:
Will check this out over the weekend

nabi-ebrahimi:
@robertjchen, Gentle bump. thanks.

nabi-ebrahimi:
@robertjchen, Quick bump.

robertjchen:
checking this out tomorrow!

robertjchen:
@hoangzinh I just tested the workflow you've described here: https://github.com/Expensify/App/pull/92338#issuecomment-4632923558, here are the results:

After Step 6, the FE still displays "Members must tag all expenses" as enabled.

Screenshot 2026-06-23 at 8 07 45 PM

The database shows this value for the field `$.requiresTag` on the policy:
```
1
```

After refresh, the FE shows this:

Screenshot 2026-06-23 at 8 08 18 PM

I add a new tag (steps 7-9) and open the new Open "Tags Settings" again:

Screenshot 2026-06-23 at 8 08 41 PM

The database shows this value for the field `$.requiresTag` on the policy:
```
1
```

Let me know if you have any other questions, thanks!

hoangzinh:
Hi @robertjchen. I'm assuming `1 = true/enabled`

- After step 6, if `$.requiresTag` on the policy is `1`, why does the OpenPolicyTagsPage API return `$.requiresTag=false` when user reload the page

robertjchen:
Yes, so on further investigation it turns out `OpenPolicyTagsPage` returns a derived `requiresTag` that's forced to `false` when the policy has no enabled tags, so reload shows `false` even though the raw `$.requiresTag` NVP in the database stays `1` (true).

The bug is that `ClearPolicyTags` recomputes this value but only pushes it to the `policyTags` collection, never to the policy object, so clients keep the stale `requiresTag` until an `Open*` command re-derives it on reload. 🤔 

So, we'll need a backend fix to also push the recomputed `requiresTag` onto the policy object (`KEY_TAGLISTS`)

hoangzinh:
@robertjchen We're handling this by **optimistic set requiresTag of policy object to false** here https://github.com/Expensify/App/pull/92338/changes#diff-fac52412018ef10d5c847bacf499268588a23ce862851515de4d4a63610e06f7R856-R858

However, at step 9 above, when the user starts to add a first tag, it becomes an issue. As BE starts to return the correct value in the database of `requiresTag` of the policy object.

@nabi-ebrahimi proposed a solution (in this PR) that we will try to fetch policy data everytime user open Tag settings page. Do you agree with this approach, or do you think it's a workaround solution?

robertjchen:
I will research and get back this week

nabi-ebrahimi:
@robertjchen, Gentle bump. thanks.

robertjchen:
yep, still on the list!

nabi-ebrahimi:
@robertjchen, Friendly bump. thanks.

robertjchen:
backend part to be implemented

src/libs/actions/Policy/Tag.ts
```diff
@@ -119,7 +128,7 @@ function getImportMultiLevelTagsFinalModal(): ImportFinalModal {
     };
 }
 
-function createPolicyTag({
+async function createPolicyTag({
```
hoangzinh: This will violate this rule

> Action methods SHOULD not return a promise
Ref: https://github.com/Expensify/App/blob/main/contributingGuides/philosophies/DATA-BINDING.md#--action-methods-should-not-return-a-promise

src/libs/actions/Policy/Tag.ts
```diff
@@ -139,8 +148,13 @@ function createPolicyTag({
     const policyID = policy?.id;
     const policyTag = PolicyUtils.getTagLists(policyTags)?.at(0) ?? ({} as PolicyTagList);
     const newTagName = PolicyUtils.escapeTagName(tagName);
+    const requiredAfterSwitchKey = policyID ? getPolicyTagsRequiredAfterSwitchKey(policyID) : undefined;
+    // Restore the required toggle only for the first tag created after the switch-level cleanup.
+    const shouldRestoreRequiresTag = !!requiredAfterSwitchKey && (await OnyxUtils.get(requiredAfterSwitchKey)) === true && getEnabledPolicyTagsCount(policyTag) === 0;
```
hoangzinh: I don't think we're allowed to use`OnyxUtils.get(requiredAfterSwitchKey)`. Please pass the value of this key from component level

src/libs/actions/Policy/Tag.ts
```diff
@@ -166,6 +180,7 @@ function createPolicyTag({
                 key: `${ONYXKEYS.COLLECTION.POLICY_TAGS}${policyID}`,
                 value: {
                     [policyTag.name]: {
+                        ...(shouldRestoreRequiresTag ? {required: true} : {}),
```
hoangzinh, line 179: Should we put this in the `optimisticData`?

src/pages/workspace/tags/WorkspaceCreateTagPage.tsx
```diff
@@ -103,7 +104,7 @@ function WorkspaceCreateTagPage({route}: WorkspaceCreateTagPageProps) {
             setupCategoriesAndTagsParentReportAction,
             currentUserAccountID: currentUserPersonalDetails.accountID,
             policyHasCustomCategories,
-        });
+        }).catch((error: unknown) => Log.warn('[WorkspaceCreateTagPage] Failed to create policy tag', {error}));
```
hoangzinh: What is purpose of this log?

src/ONYXKEYS.ts
```diff
@@ -805,6 +805,8 @@ const ONYXKEYS = {
         POLICY_CATEGORIES_DRAFT: 'policyCategoriesDraft_',
         POLICY_RECENTLY_USED_CATEGORIES: 'policyRecentlyUsedCategories_',
         POLICY_TAGS: 'policyTags_',
+        // Client-only marker used to restore required tags after switching tag levels clears all tags.
+        POLICY_TAGS_REQUIRED_AFTER_SWITCH: 'policyTagsRequiredAfterSwitch_',
```
hoangzinh: Uhm, should we add a new client key into existing `Onyx.Policy`? I.e https://github.com/Expensify/App/blob/a7b8fea0bb20ade7bc8653c2e81295f745c4d428/src/types/onyx/Policy.ts#L251-L252

src/libs/actions/Policy/Tag.ts
tgolen: That's correct. We should never use `OnyxUtils.get()`

src/libs/actions/Policy/Tag.ts
tgolen: +1

src/pages/workspace/tags/WorkspaceCreateTagPage.tsx
tgolen: +1 Let's remove it.

src/libs/actions/Policy/Tag.ts
nabi-ebrahimi: Hi @hoangzinh , thanks for pointing this out. I addressed your feedback by making createPolicyTag synchronous again, so the action method no longer returns a Promise. The required-tags restore value is now passed into the action instead.

src/libs/actions/Policy/Tag.ts
nabi-ebrahimi: Hi @hoangzinh , thanks for the feedback. I addressed this by removing OnyxUtils.get(requiredAfterSwitchKey) from the action and reading the marker value at the component level with useOnyx instead. The value is now passed into createPolicyTag as a parameter

src/libs/actions/Policy/Tag.ts
nabi-ebrahimi: Hi @hoangzinh , thanks for the suggestion. I addressed this by making sure required: true is included in the optimistic update as well, so the UI reflects the required-tags state immediately while the request is pending and in offline mode.

src/pages/workspace/tags/WorkspaceCreateTagPage.tsx
nabi-ebrahimi: Hi @hoangzinh , thanks for the feedback. I removed this log because it was only added to handle the async action path. Since createPolicyTag is now synchronous again, the log is no longer needed

src/ONYXKEYS.ts
nabi-ebrahimi: Hi @hoangzinh , thanks for the suggestion. I addressed this by moving the client-only restore marker into the existing policy Onyx data instead of keeping it as a separate POLICY_TAGS_REQUIRED_AFTER_SWITCH collection key.

hoangzinh: can you check if you already pushed this commit @nabi-ebrahimi

hoangzinh: I'm still seeing POLICY_TAGS_REQUIRED_AFTER_SWITCH key

nabi-ebrahimi: Hi @hoangzinh , sorry for the confusion, and thanks for pointing this out. I’ve updated the implementation to remove the separate `POLICY_TAGS_REQUIRED_AFTER_SWITCH` Onyx collection and moved the client-only restore marker into the existing policy Onyx data instead. Could you please take another look when you have a chance? Thanks!

hoangzinh: cool cool. I will try to review again today

src/types/onyx/Policy.ts
```diff
@@ -2623,6 +2623,9 @@ type Policy = OnyxCommon.OnyxValueWithOfflineFeedback<
         /** Whether new transactions need to be tagged */
         requiresTag?: boolean;
 
+        /** Client-only marker used to restore required tags after switching tag levels clears all tags */
+        shouldRestoreRequiresTagAfterSwitch?: boolean | null;
```
hoangzinh: The `should`  prefix reads like a boolean action. A name describing the pending intent/state fits better than one describing an instruction. Wdyt about:
- wasRequiresTagClearedBySwitch
- pendingRequiresTagRestore

src/libs/actions/Policy/Tag.ts
```diff
@@ -134,13 +139,18 @@ function createPolicyTag({
...
     const {policy, tags: policyTags} = policyData;
     const policyID = policy?.id;
     const policyTag = PolicyUtils.getTagLists(policyTags)?.at(0) ?? ({} as PolicyTagList);
     const newTagName = PolicyUtils.escapeTagName(tagName);
+    // Restore the required toggle only for the first tag created after the switch-level cleanup.
+    const shouldRestoreRequiresTag = !!policyID && shouldRestoreRequiresTagAfterSwitch === true && getEnabledPolicyTagsCount(policyTag) === 0;
+    const policyRequiresTagOptimisticData: Partial<Policy> = shouldRestoreRequiresTag ? {requiresTag: true} : {};
     const tagListsOptimisticData = {
         [policyTag.name]: {
+            ...(shouldRestoreRequiresTag ? {required: true} : {}),
```
hoangzinh, line 153: do we need to set it here? I thought we only need to set `policy.requiresTag: true`

src/libs/actions/Policy/Tag.ts
```diff
@@ -193,7 +205,35 @@ function createPolicyTag({
...
+            value: {
+                ...policyRequiresTagOptimisticData,
+                shouldRestoreRequiresTagAfterSwitch: null,
+            },
+        });
+        onyxData.successData?.push({
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+            value: {
+                ...policyRequiresTagOptimisticData,
```
hoangzinh: We don't need to set on `successData`. It will potentially override BE response data

src/libs/actions/Policy/Tag.ts
```diff
@@ -888,7 +928,10 @@ function enablePolicyTags(policyData: PolicyData, enabled: boolean) {
     }
 }
 
-function cleanPolicyTags(policyID: string) {
+function cleanPolicyTags(policyID: string, shouldRestoreRequiresTagAfterTagCreate = false) {
+    // Remember this locally because the clean response can leave policy.requiresTag false until the next policy refresh.
+    Onyx.merge(`${ONYXKEYS.COLLECTION.POLICY}${policyID}`, {shouldRestoreRequiresTagAfterSwitch: shouldRestoreRequiresTagAfterTagCreate ? true : null});
```
hoangzinh:
```suggestion
    if (shouldRestoreRequiresTagAfterSwitch) {
      Onyx.merge(`${ONYXKEYS.COLLECTION.POLICY}${policyID}`, {shouldRestoreRequiresTagAfterSwitch: true});
    }
```

src/pages/workspace/tags/ImportTagsOptionsPage.tsx
```diff
@@ -67,6 +67,8 @@ function ImportTagsOptionsPage({route}: ImportTagsOptionsPageProps) {
         () => [getTagLists(policyTags), isMultiLevelTagsPolicyUtils(policyTags), hasDependentTagsPolicyUtils(policy, policyTags)],
         [policy, policyTags],
     );
+    // If required tags was enabled before clearing tags for a level switch, the next first tag should restore that setting.
+    const shouldRestoreRequiresTagAfterTagCreate = (policy?.requiresTag ?? false) || policyTagLists.some((policyTagList) => policyTagList.required);
```
hoangzinh, line 71: I'm curious if we need to check `policyTagLists.some((policyTagList) => policyTagList.required);`

src/libs/actions/Policy/Tag.ts
nabi-ebrahimi: @hoangzinh Thank you for the feedback.

Good question. I set the tag list `required` value here because the Tags page still relies on `policyTags[firstList].required` for the required/locked tag UI. Setting only `policy.requiresTag` updates the `Members must tag all expenses` toggle, but it can leave the Tags page temporarily inconsistent during the offline optimistic state. This also matches the existing `setPolicyRequiresTag()` flow, which updates both `policy.requiresTag` and the tag list `required` state.

src/libs/actions/Policy/Tag.ts
nabi-ebrahimi: @hoangzinh Thank you for pointing this out.
I updated the successData so it no longer sets policy.requiresTag. It now only clears the local pendingRequiresTagRestore marker, allowing the backend response to remain the source of truth after CreatePolicyTag succeeds

src/libs/actions/Policy/Tag.ts
nabi-ebrahimi: @hoangzinh Thank you for the suggestion.
I updated cleanPolicyTags() so it only sets pendingRequiresTagRestore when the restore flow is needed. This avoids writing null during the normal cleanup path and keeps the Onyx update scoped to the case that actually requires restoring the required tag state.

src/pages/workspace/tags/ImportTagsOptionsPage.tsx
nabi-ebrahimi: @hoangzinh Good question.

I kept this check intentionally because, during this flow, `policy.requiresTag` can temporarily be stale while the tag list already has `required: true`. If we only rely on `policy.requiresTag`, we may miss setting the restore marker before clearing tags, and the first tag created afterward would not restore the required tag state offline. This keeps the restore intent aligned with the tag list state before cleanup.

src/types/onyx/Policy.ts
nabi-ebrahimi: @hoangzinh Thank you for the suggestion.

I renamed the marker to `pendingRequiresTagRestore` so it describes the pending restore state more clearly.

src/libs/actions/Policy/Tag.ts
```diff
@@ -193,7 +205,34 @@ function createPolicyTag({
...
+            value: {
+                ...policyRequiresTagOptimisticData,
+                pendingRequiresTagRestore: null,
+            },
+        });
+        onyxData.successData?.push({
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+            value: {
+                pendingRequiresTagRestore: null,
```
hoangzinh: NAB: we don't need this.

nabi-ebrahimi: @hoangzinh,I have addressed this feedback. Thanks!

hoangzinh (approved):

mountiny (approved):
Thank you

robertjchen (approved):
thanks, let us know how testing goes 👍

PR #92386, opened 2026-06-02
fix: settings popover overlaps controls in Android fullscreen video (#91216)

tanmayxchoudhary:
## Description

Fixes the bug where opening the Speed/Audio settings popover in Android fullscreen video causes the popover to overlap the playback controls bar and other menu items.

## Root cause

The popover position is cached and computed against the controller layout at the time the popover was last dismissed. After the video reaches `STATE_ENDED`, the controller layout changes (play button becomes replay, progress bar stays visible) but the cached popover position is not recalculated. So the next time the popover is opened (after re-open), the anchor is wrong and the popover overlaps controls.

## Fix

Add a `Player.Listener` on `PlayerView.player` that forces a controller rebuild on `STATE_ENDED`. This forces the popover position to be recomputed against the new layout on next open.

## Reproduction

1. Open app → any chat → + → Add Attachment → video → send
2. Open the video in chat
3. Tap fullscreen
4. Tap gear icon (Speed/Audio) → popover appears overlapping the controls bar
5. After video ends and you reopen, the popover position is correct (because the listener triggered a rebuild)

With this patch, the rebuild happens proactively on `STATE_ENDED` so the popover is correctly positioned when re-opened.

## Implementation

The patch is applied via patch-package against `expo-video@55.0.3` to keep the native Android fix separate from the JS layer. The patch file is `patches/expo-video/expo-video+55.0.3+002+fix_settings_popup_position_in_fullscreen.patch` and follows the same naming convention as the existing `001` patch.

## Tests

- Manual reproduction confirms the popover position is now correct after re-open
- The patch follows the standard Android `Player.Listener` pattern
- `patches/expo-video/details.md` updated with the new patch entry

## Reference

- Issue: #91216
- Previous (failed) attempt: #91800 (the JS-only fix didn't address the native popover anchoring issue)

$ #91216


tanmayxchoudhary:
## Verification & Fix Update

I verified the patch against the actual ExoPlayer/Media3 source code (`androidx/media3/ui/PlayerView.java`). Here's what I found:

### v1 was a no-op
The original patch called `playerView.post { playerView.useController = true }` on STATE_ENDED. However, `PlayerView.setUseController(boolean)` has this guard:
```java
if (this.useController == useController) {
    return;  // <-- early return when already true
}
```
Since `useController` is already `true` in the fullscreen player, this call does absolutely nothing. The controller is not rebuilt, the popup position is not recalculated.

### Revised root cause understanding
The settings popup position IS recalculated each time it's opened — `PlayerControlView.displaySettingsWindow()` calls `PopupWindow.showAsDropDown(anchorView, xoff, yoff)` with fresh offsets every time. The issue is that after `STATE_ENDED`, the controller's internal layout state may not have settled (play→replay button transition, auto-show mode switch). This can cause stale anchor view positions or controller dimensions to feed into the offset calculation.

### v2 fix
Replaced with `playerView.post { playerView.requestLayout() }` — the standard Android mechanism to force a fresh measure/layout pass. This ensures the settings button anchor view and controller dimensions are up-to-date when the popup is next opened.

**Commit**: 0a65190 (already pushed to head branch)

PR #92695, opened 2026-06-04
[CP Staging] Revert "Remove inline-flex from comment anchors so links wrap inline"

Pujan92:
Reverts Expensify/App#92387


lakchote (approved):
LGTM

lakchote:
Not an emergency, straight revert

Issue #82319, opened 2026-02-12
[Hold E/E#596490] [CFI] NEW FEATURE: Add Expensify Card statement download to NewDot


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.18-2
**Reproducible in staging?:** Y
**Reproducible in production?:** Y
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @joekaufmanexpensify 
**Slack conversation** (hyperlinked to channel name): [#Expensify Bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1770907122177089)

## Action Performed:

1. Login to an admin account with the Expensify Card. 
2. Go to Reports > Reconciliation > Expensify Card. 
3. Set a specific date range. 
4. Select all the settlements. 
5. Open the bulk action menu and go to export.

## Expected Result:
We show the option to download a PDF Expensify Card statement for that period, like we do from the settlements reconciliation tool in OldDot. Notes:

- We can reuse the same PDF statement as OldDot, see below. We just need to update the domain field to use the workspace name when applicable.
- We should only show this option when a user selects settlements from the same Expensify Card feed. So if they have multiple and select settlements from both, we wouldn't show this export option. That seems simplest compared to creating a new PDF.
- The `Download statement PDF` option should go in the bulk actions menu as a new option, rather than inside of `Export`. 

## Actual Result:
There's no option to download the PDF Expensify Card statement in NewDot.
## Workaround:
Can the user still use Expensify without this being fixed? Have you informed them of the workaround?

## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

<img width="1874" height="1494" alt="Image" src="https://github.com/user-attachments/assets/b7489fce-06e1-4211-8c2c-9fc2cab019c2" />

[...]

Issue #54061, opened 2024-12-12
[HOLD for payment 2025-01-17] [$250] Attachment - "This image has been resized" message overlaps with Android controls


lanitochka17:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.0.75-0
**Reproducible in staging?:** Y
**Reproducible in production?:** Y
**If this was caught on HybridApp, is this reproducible on New Expensify Standalone?:** Y
**If this was caught during regression testing, add the test name, ID and link from TestRail:** https://expensify.testrail.io/index.php?/tests/view/5338217&group_by=cases:section_id&group_id=292107&group_order=asc
**Issue reported by:** Applause - Internal Team
**Component** Other

## Action Performed:
1. Open the app
2. Log in with a Gmail account
3. Open any 1:1 DM
4. Send a big picture attachment
5. Open the image

## Expected Result:
Elements shouldn't overlap

## Actual Result:
"This image has been resized" message overlaps with Android controls

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [x] Android: Standalone
- [x] Android: HybridApp
- [ ] Android: mWeb Chrome
- [ ] iOS: Standalone
- [ ] iOS: HybridApp
- [ ] iOS: mWeb Safari
- [ ] MacOS: Chrome / Safari
- [ ] MacOS: Desktop

## Screenshots/Videos

<details>
  <summary>Add any screenshot/video evidence</summary>

  
</details> 

https://github.com/user-attachments/assets/71800d1a-dd57-4e7a-a8b3-57b591f5c8c6

![Bug6692060_1734002573544!bigggg](https://github.com/user-attachments/assets/c41be38d-ccba-4dd0-8cc4-a6bfff378cd7)

![!ol](https://github.com/user-attachments/assets/2c4057a3-1062-4af7-972a-22149398b636)

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~021867655181903495230</li>
        <li>Upwork Job ID: 1867655181903495230</li>
[...]

Issue #22947, opened 2023-07-15
[Hold for PR 24041] [$1000] Dev: Web - Composer disappears while switching chats in high traffic account


kbecciv:
This is held on https://github.com/Expensify/App/pull/24041

If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
1. Switch between large chats in a high traffic account

## Expected Result:
Composer remains visible

## Actual Result:
Composer disappears, then later reappears

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [ ] Android / Chrome
- [ ] iOS / native
- [ ] iOS / Safari
- [x] MacOS / Chrome / Safari
- [ ] MacOS / Desktop

**Version Number:** Dev 1.3.40-1
**Reproducible in staging?:** n/a
**Reproducible in production?:** n/a
**If this was caught during regression testing, add the test name, ID and link from TestRail:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** Any additional supporting documentation

https://github.com/Expensify/App/assets/93399543/36ce712b-7efa-48cf-9a27-c63f8d955401

**Expensify/Expensify Issue URL:**
**Issue reported by:** Amy Evans
**Slack conversation:** https://expensify.slack.com/archives/C049HHMV9SM/p1689360443816699

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~01d355d230175a068f</li>
        <li>Upwork Job ID: 1681049473425895424</li>
        <li>Last Price Increase: 2023-07-24</li>
    </ul>
</details>

conorpendergrast:
@amyevans I wasn't able to see this on v1.3.41-2 (Staging), so outside of dev. Have you only found this on dev so far?

amyevans:
I reproduced it on prod, albeit a lot less pronounced:

https://github.com/Expensify/App/assets/7277067/a3459543-258f-43e0-885e-aca0178acc99

[...]

PR #92275, opened 2026-06-01
feat: Enable multi-select Download as PDF

truph01:
### Fixed Issues

$ https://github.com/Expensify/App/issues/90300
PROPOSAL:

### Tests

1. Open app
2. Go to Spend > Reports
3. Choose more than one report > Click on "X selected" dropdown button > Click on "Download as PDF" option
4. Verify the below modal is shown:

Screenshot 2026-06-16 at 16 34 13

5. Wait a minute until the below modal is shown:

Screenshot 2026-06-16 at 16 34 42

6. Click on "Download file" button. 

7. Verify the pdf files are downloaded.

8. Do step 1 - 4

9. Click on “Send me the file when it’s ready”

10. Wait a few minute and click Go to Concierge

11. Verify the pdf files is send by Concierge

- [x] Verify that no errors appear in the JS console

### QA Steps

1. Open app
2. Go to Spend > Reports
3. Choose more than one report > Click on "X selected" dropdown button > Click on "Download as PDF" option
4. Verify the below modal is shown:

Screenshot 2026-06-16 at 16 34 13

5. Wait a minute until the below modal is shown:

Screenshot 2026-06-16 at 16 34 42

6. Click on "Download file" button. 

7. Verify the pdf files are downloaded.

8. Do step 1 - 4

9. Click on “Send me the file when it’s ready”

10. Wait a few minute and click Go to Concierge

11. Verify the pdf files is send by Concierge

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
[...]

PR #92790, opened 2026-06-05
Add Expensify Card statement PDF export to search bulk actions

ishpaul777:
### Explanation of Change

Adds a **Download statement** bulk action in Search when an admin selects Expensify Card settlements in the reconciliation view. It generates a PDF of the selected settlements' transactions and payments.

The action shows when all of these hold:

- The search is the Expensify Card reconciliation view: `type:expense`, `withdrawal-type:expensify-card`, `groupBy:withdrawal-id`.
- Whole settlements are selected (a collapsed group row, or every transaction inside an expanded group). A partial in-group selection does not count.
- No transaction-narrowing filter (merchant, category, amount, date, etc.) is active. Those would make the PDF disagree with the rows on screen.
- The backend marks each selected settlement exportable via `canExportStatement` on the withdrawal group. The client gates the button on this flag instead of re-deriving admin access, so the button and the download endpoint authorize identically (see Auth PR below). Non-admins who can still see a settlement in search do not get the button.
- Not in "select all matching" mode. That only loads the visible rows, so the statement would be incomplete.

Scope of the PDF:

- No workspace filter: the whole settlement, covering every workspace it spans. The header is named from the feed's workspace (backend-provided `feedName`).
- Single `policyID:` filter: scoped to that workspace's transactions only. A multi-workspace or negated `policyID:` filter hides the action, since a statement scopes to one workspace or the whole settlement, not an arbitrary subset.

A statement covers one feed (identified by `fundID`, so two settlements under the same program but different feeds are distinct). If the selection spans more than one feed, the action shows "Please select settlements from one Expensify Card feed at a time." and sends no request. Failed and pending settlements are exportable; the statement labels their status.

On a valid selection: offline shows an offline message and sends nothing; online shows a generating-PDF modal, then downloads the file when ready.

### Related backend PRs (deploy first)

- Auth: [#22947](https://github.com/Expensify/Auth/pull/22947) (authorization + `canExportStatement` flag)
- Web-Expensify: [#54061](https://github.com/Expensify/Web-Expensify/pull/54061) (PDF generation)

Deploy order: Auth, then Web-Expensify, then App.

### Fixed Issues
$ https://github.com/Expensify/App/issues/82319

PROPOSAL:

### Tests

Sign in as a workspace admin of a workspace with an Expensify Card feed and at least one settlement.

1. Open Search, run the reconciliation view (`type:expense withdrawal-type:expensify-card groupBy:withdrawal-id`). Select a whole settlement, open bulk actions, tap Download statement. Confirm the generating modal appears, then the PDF downloads. It contains the settlement's transactions/payments, header named from the feed's workspace.
2. Add a single `policyID:` filter, repeat. The PDF is scoped to that workspace's transactions.
3. Select a settlement spanning multiple workspaces (no `policyID:` filter). Download statement returns the whole settlement.
4. Select a failed settlement. It exports and the statement labels the status.
5. Select settlements from two feeds (different `fundID`). Confirm the "one Expensify Card feed at a time" message appears and no request is sent.
6. Select only some transactions inside a settlement (not the whole group). Confirm Download statement does not appear.
7. Add a transaction-narrowing filter (merchant/category/amount). Confirm Download statement disappears while active.
8. Sign in as a non-admin cardholder who can see a settlement. Confirm Download statement does not appear (backend `canExportStatement` is false).
9. Confirm no `Failed to construct 'URL': Invalid URL` or `baseURL is not defined` errors in the JS console.

### Offline tests

1. Disable network, select a settlement, tap Download statement.
2. Confirm the offline modal shows and no `GetExpensifyCardStatementPDF` request is sent.
3. Re-enable network, retry. Confirm the normal generation flow works.

### QA Steps

Prerequisites: Auth [#22947](https://github.com/Expensify/Auth/pull/22947) and Web-Expensify [#54061](https://github.com/Expensify/Web-Expensify/pull/54061) on staging.

Log in as a workspace admin with an Expensify Card feed and settlements. Go to Search and open the reconciliation view (`withdrawal-type:expensify-card`, grouped by withdrawal).

Download statement should appear when you select a whole settlement:

1. A completed settlement, a pending one, and a failed one each show it (the failed one's status is labeled in the PDF).
2. A settlement spanning multiple workspaces shows it and exports the whole settlement.
3. Several settlements from the same feed show it and combine into one statement.

Shown but tapping shows a message, no download:

4. Settlements from two feeds → "Please select settlements from one Expensify Card feed at a time."

Should not appear:

5. Nothing selected, or only individual transactions inside a settlement (not the whole row).
6. A transaction-narrowing filter (merchant, category, amount) is active.
7. Signed in as a non-admin cardholder who can see the settlement (backend `canExportStatement` is false).

Scope:

8. No workspace filter → the PDF is the whole settlement, headed by the feed's workspace name.
9. Single `policyID:` filter → the PDF is scoped to that workspace's transactions.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


ishpaul777:
@codex review

@MelvinBot review

ishpaul777:
@truph01 would you like to review and test this while its still in draft ? there will be backend changes that i am parallely working on i will share ngrok pointing to my local BE env so you can test and give feedback

truph01:
> @truph01 would you like to review and test this while its still in draft ? there will be backend changes that i am parallely working on i will share ngrok pointing to my local BE env so you can test and give feedback

@ishpaul777 Sure, I will review this PR today. Could you give me the env config?

ishpaul777:
> Could you give me the env config?

I DM you the details

src/hooks/useSearchBulkActions.ts
```diff
@@ -487,6 +494,57 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
...
+        if (!feed) {
+            return;
+        }
+
+        const statementParams = getExpensifyCardStatementParamsFromFeed(feed);
+        const {entryIDs} = statementParams;
+
+        setExpensifyCardStatementPDFParams(statementParams);
+        setIsExpensifyCardStatementPDFModalVisible(true);
+        getExpensifyCardStatementPDF(statementParams.policyID, statementParams.feedCountry, entryIDs)?.then((response) => {
```
truph01: There's no `.catch.` And `getExpensifyCardStatementPDF` only defines optimisticData/failureData — failureData just flips `isGenerating: false` with no filename. In the modal:

```
const hasFinishedPDFDownload = !!statementKey && !isGeneratingPDF && typeof statementFileName === 'string' && statementFileName.length > 0;
```

On failure, `isGenerating` becomes false but `statementFileName` stays empty → `hasFinishedPDFDownload` is false forever → the modal sits on "waiting for PDF" with a spinner and only a Cancel button; the error is never surfaced. Combined with the missing .catch, a rejected request is also an unhandled promise rejection.

truph01:
@ishpaul777  I saw that this PR has change in Mobile-Expensify repo. Is is expected?

src/hooks/useSearchBulkActions.ts
ishpaul777: Good catch — fixed in cb516acadf6. Added a `.catch` and treat a missing `statementKey` the same way: both now close the loading modal and surface the existing download-error modal (same path CSV/basic export uses). Added a test for it.

ishpaul777:
@truph01 No, that was an unrelated submodule bump from a stale base — removed it when I merged main in. The Mobile-Expensify pointer now matches main, so it no longer shows in the diff.

ishpaul777:
@MelvinBot please review

ishpaul777:
@codex review

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,194 @@
...
+        }
+        if (selection.groupKey?.startsWith(CONST.SEARCH.GROUP_PREFIX)) {
+            selectedGroupKeys.add(selection.groupKey);
+        }
+    }
+
+    const settlementGroups: SearchWithdrawalIDGroup[] = [];
+    for (const groupKey of selectedGroupKeys) {
+        const group = searchData[groupKey as keyof SearchResultDataType] as SearchWithdrawalIDGroup | undefined;
+        if (!group || typeof group.entryID !== 'number' || getSettlementStatus(group.state) === CONST.SEARCH.SETTLEMENT_STATUS.FAILED) {
```
MelvinBot: **Pending settlements incorrectly qualify for PDF export.**

This filter only excludes `FAILED` settlements, so anything that isn't failed passes — including **pending** ones. Per [`getSettlementStatus`](https://github.com/Expensify/App/blob/cb516acadf607845fa1f5a592a163ecb77e3227b/src/libs/SearchUIUtils.ts#L5947-L5961), states `5,6,7` → `FAILED`, `8,9` → `CLEARED`, and **every other state → `PENDING`**. A selected settlement in any pending state therefore gets included in the export.

This contradicts the PR's stated behavior ("Only **cleared** settlements qualify (`getSettlementStatus(state) === CLEARED`)") and QA step 8 ("Select a pending/failed settlement only → verify Export as PDF is not offered").

```suggestion
        if (!group || typeof group.entryID !== 'number' || getSettlementStatus(group.state) !== CONST.SEARCH.SETTLEMENT_STATUS.CLEARED) {
```

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    const feeds = Array.from(feedsByKey.values());
+    if (feeds.length === 0 && !hasMixedWorkspaceSettlement) {
+        return undefined;
+    }
+
+    return {
+        feeds,
+        hasMultipleFeeds: feeds.length > 1 || hasMixedWorkspaceSettlement,
```
MelvinBot: **A settlement with no `policyID` poisons the whole selection / produces a dead-end menu item.**

`hasMultipleFeeds` becomes `true` whenever `hasMixedWorkspaceSettlement` is set (line 130, triggered by any selected settlement missing a `policyID`). Two problematic cases:

1. **One valid feed + one no-`policyID` settlement** → `feeds.length === 1` but `hasMultipleFeeds === true`, so a legitimate single-feed export is blocked and the user instead gets the "Please pick one Expensify Card feed at a time" alert.
2. **All selected settlements have no `policyID`** → `feeds` is empty but the early-return at line 149 (`feeds.length === 0 && !hasMixedWorkspaceSettlement`) is skipped, so a selection object is still returned. The "Export as PDF" item then **appears in the menu**, but clicking it only opens the multi-feed alert — there is no feed to export, a dead end.

If a no-`policyID` settlement should never be exportable, consider skipping it (like `FAILED`) in `getSelectedSettlementGroups` rather than flagging the entire selection as multi-feed; and gate the returned selection on `feeds.length > 0`.

src/libs/actions/CompanyCards.ts
```diff
@@ -1291,6 +1292,38 @@ function linkCardFeedToPolicy(domainAccountID: number, policyID: string, feedTyp
...
+
+    const params: GetExpensifyCardStatementPDFParams = {
+        policyID,
+        entryIDs: entryIDs.join(','),
+        ...(feedCountry ? {feedCountry} : {}),
+    };
+
+    // makeRequestWithSideEffects is used so callers can read statementKey from the response.
+    // eslint-disable-next-line rulesdir/no-api-side-effects-method
+    return API.makeRequestWithSideEffects(SIDE_EFFECT_REQUEST_COMMANDS.GET_EXPENSIFY_CARD_STATEMENT_PDF, params, {optimisticData, failureData});
```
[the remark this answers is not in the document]
ishpaul777: Good point — the server response already echoes `isGenerating: false`, so the happy path wasn't actually stuck, but relying solely on the server is fragile. Added client `successData` to clear it in 464b1fd97f9.

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Pending settlements are intentionally exportable here (there's a dedicated commit + tests for states 0 and 9) — only failed and mixed-workspace settlements are excluded. The PR description / QA step 8 was stale; fixed those to match. No code change needed.

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Valid, fixed in 464b1fd97f9. A no-`policyID` (mixed-workspace) settlement is now skipped in `getSelectedSettlementGroups` like a failed one, and the selection is gated on `feeds.length > 0`. So a valid feed is no longer blocked by a mixed-workspace settlement, and the dead-end menu item is gone. The multi-feed alert now only fires for genuinely distinct feeds.

ishpaul777:
@truph01 can you give this another review?

ishpaul777:
@MelvinBot review and leave inline comments

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    const selectedSettlementGroups = getSelectedSettlementGroups(selectedTransactions, searchData);
+    if (selectedSettlementGroups.length === 0) {
+        return undefined;
+    }
+
+    const feedsByKey = new Map<string, ExpensifyCardStatementFeed>();
+    for (const settlementGroup of selectedSettlementGroups) {
+        if (!settlementGroup.policyID) {
```
MelvinBot: This `!settlementGroup.policyID` guard is dead code: `getSelectedSettlementGroups` already skips any group without a `policyID` (the earlier `if (!value.policyID) continue;`), so every `settlementGroup` reaching this loop is guaranteed to have one. Safe to drop the check (and the `?? ''` fallback on `feedCountry` is enough to key the map).

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,204 @@
...
+        return undefined;
+    }
+
+    const feed = selection.feeds.at(0);
+    if (!feed) {
+        return undefined;
+    }
+
+    return getExpensifyCardStatementParamsFromFeed(feed);
+}
```
MelvinBot: `getExpensifyCardStatementParams` is only consumed by `ExpensifyCardStatementUtilsTest.ts` — production now uses `getExpensifyCardStatementSelection` + `getExpensifyCardStatementParamsFromFeed` directly (see `useSearchBulkActions.exportExpensifyCardStatementPDF`). So the tests exercise a function the app never calls. Consider either removing this helper, or pointing the relevant tests at the actual production path so they guard the real flow.

src/libs/actions/CompanyCards.ts
```diff
@@ -1291,6 +1292,47 @@ function linkCardFeedToPolicy(domainAccountID: number, policyID: string, feedTyp
...
+    ];
+    const successData: Array<OnyxUpdate<typeof ONYXKEYS.EXPENSIFY_CARD_STATEMENT>> = [
+        {
+            onyxMethod: Onyx.METHOD.MERGE,
+            key: ONYXKEYS.EXPENSIFY_CARD_STATEMENT,
+            value: {
+                isGenerating: false,
+            },
+        },
+    ];
```
MelvinBot, line 1451: `successData`/`failureData` only toggle `isGenerating`; the per-statement `cacheKey -> filename` entries that the backend writes into `ONYXKEYS.EXPENSIFY_CARD_STATEMENT` are never cleared, so this object grows for the lifetime of the session and keeps stale filenames around. Not a correctness bug (the modal re-requests and waits on `isGenerating`), but worth a cleanup on success/hide if statement keys can churn.

src/hooks/useSearchBulkActions.ts
```diff
@@ -511,6 +518,68 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
...
+
+        const showStatementError = () => {
+            setIsExpensifyCardStatementPDFModalVisible(false);
+            setExpensifyCardStatementPDFParams(undefined);
+            setIsDownloadErrorModalVisible(true);
+        };
+
+        setExpensifyCardStatementPDFParams(statementParams);
+        setIsExpensifyCardStatementPDFModalVisible(true);
+        getExpensifyCardStatementPDF(statementParams.policyID, statementParams.feedCountry, entryIDs)
```
MelvinBot, line 625: The `?.then(...)` implies `getExpensifyCardStatementPDF` can return `undefined`. If it ever does (or the underlying request is paused/deduped and the promise never settles), neither the `.then` nor the `.catch` runs and the modal is left stuck on the "Generating PDF" loading state with no error fallback. The offline path is handled up front, but it may be worth confirming the request promise always settles, or adding a timeout/guard so the modal can recover.

MelvinBot (commented):
Reviewed the Expensify Card statement PDF export. Overall this is clean, well-documented, and the unit tests cover the selection logic thoroughly (collapsed vs expanded rows, mixed-workspace exclusion, multi-feed, scope filters, failed/pending states). Left a few minor, non-blocking inline notes — nothing that should block once the backend PRs are deployed.

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Good catch — fixed in c1524d74261. `getSelectedSettlementGroups` now returns `ExportableSettlementGroup` (`SearchWithdrawalIDGroup & {policyID: string}`), so the type carries the guaranteed `policyID` across the function boundary and this redundant guard is gone.

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Agreed — removed `getExpensifyCardStatementParams` in c1524d74261 and pointed those tests at the production composition (`getExpensifyCardStatementSelection` + `getExpensifyCardStatementParamsFromFeed`, via a small local helper that mirrors `exportExpensifyCardStatementPDF`), so they now guard the real flow. All 13 unit tests pass.

src/libs/actions/CompanyCards.ts
ishpaul777: Fair observation, but I'm going to leave this as-is. The entries are tiny (cacheKey → filename), session-scoped (cleared on logout / Onyx clear), and — as you noted — not a correctness issue. More importantly, the modal *relies* on those keys staying present to serve a previously-generated statement from cache without re-requesting, so clearing them on success/hide would defeat the cache. The unbounded-growth concern is real for the backend NVP store and is already tracked as a server-side TTL follow-up on the Web-Expensify PR.

src/hooks/useSearchBulkActions.ts
ishpaul777: Checked this: `makeRequestWithSideEffects` returns `Promise<void | Response<TKey>>` and always settles (it goes through `processRequest` like every other request) — it never returns `undefined`, so the `.then`/`.catch` always run. The `?.` is just defensive and could be a plain `.`. The offline case is handled up front, and on a settled-but-empty response we already call `showStatementError()` (no `statementKey` → error modal), so the modal can't get stuck. No change needed.

truph01:
> @truph01 can you give this another review?

Sure, I am reviewing it

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,187 @@
...
+    const entryIDs = [...feed.entryIDs];
+    return {
+        policyID: feed.policyID,
+        feedCountry: feed.feedCountry,
+        entryIDs,
+    };
+}
+
+function downloadExpensifyCardStatementPDF(translate: LocalizedTranslate, fileName: string, statementKey: string, currentUserEmail: string, encryptedAuthToken: string): Promise<void> {
+    const baseURL = addTrailingForwardSlash(getOldDotURLFromEnvironment(environment));
```
truph01: We should use `useEnvironment()` hook to retrieve `environment` and pass `environment` as a parameter here.

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,187 @@
...
+    const downloadFileName = `Expensify_Card_Statement_${statementKey}.pdf`;
+    const pdfURL = `${baseURL}secure?secureType=pdfreport&filename=${encodeURIComponent(fileName)}&downloadName=${encodeURIComponent(downloadFileName)}&email=${encodeURIComponent(currentUserEmail)}`;
```
truph01: Once PR https://github.com/Expensify/App/pull/92275 has been merged, we can use `buildSecureDownloadURL` here.

src/types/onyx/Response.ts
```diff
@@ -121,6 +121,12 @@ type Response<TKey extends OnyxKey> = {
 
     /** Transactions pending 3DS review returned from GetTransactionsPending3DSReview */
     transactionsPending3DSReview?: TransactionsPending3DSReview;
+
+    /** PDF filename returned from GetExpensifyCardStatementPDF */
+    statementPDF?: string;
```
truph01: `statementPDF` is added to the response type but never read anywhere in the diff. Is `statementPDF` needed, or is it dead code?

tests/unit/ExpensifyCardStatementUtilsTest.ts
```diff
@@ -0,0 +1,244 @@
...
+    status: CONST.SEARCH.STATUS.EXPENSE.ALL,
+    sortBy: CONST.SEARCH.TABLE_COLUMNS.GROUP_WITHDRAWN,
+    sortOrder: CONST.SEARCH.SORT_ORDER.DESC,
+    groupBy: CONST.SEARCH.GROUP_BY.WITHDRAWAL_ID,
+    view: CONST.SEARCH.VIEW.TABLE,
+    policyID: ['policy1'],
+    filters: {operator: CONST.SEARCH.SYNTAX_OPERATORS.AND, left: 'type', right: 'expense'},
+};
+
+function makeSelectedTransaction(overrides: Partial<SelectedTransactions[string]> = {}): SelectedTransactions[string] {
```
truph01: `makeSelectedTransaction` and settlement group helpers are duplicated between `ExpensifyCardStatementUtilsTest.ts` and `useSearchBulkActionsDownloadPDFTest.ts`. These could be extracted to a shared test utility to reduce duplication.

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Done — `downloadExpensifyCardStatementPDF` now takes `environment` as a parameter, and the only caller (`ExpensifyCardStatementPDFDownloadModal`) reads it from `useEnvironment()` and passes it in. Removed the module-level `getEnvironment().then(...)` global.

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Agreed — [#92275](https://github.com/Expensify/App/pull/92275) isn't merged yet and `buildSecureDownloadURL` doesn't exist on this branch. I'll switch to it as a follow-up once that lands; tracking it.

src/types/onyx/Response.ts
ishpaul777: Good catch — it was dead. The client reads the PDF filename from the `EXPENSIFY_CARD_STATEMENT` Onyx merge keyed by `statementKey`, never from `response.statementPDF`. Removed the field.

tests/unit/ExpensifyCardStatementUtilsTest.ts
ishpaul777: Done — extracted `makeSelectedTransaction` and the settlement-group / search-data builders to `tests/utils/ExpensifyCardStatementTestUtils.ts`, and both test files now import from there. As a bonus the download-PDF test's inline settlement literals are now built through `makeSettlementGroup`, removing a lot of duplication.

ishpaul777:
@codex review

truph01:
@ishpaul777 Has the backend PR been completed yet? Or do we have ETA?

ishpaul777:
@truph01 BE just deployed to staging, can you check if it works?

src/hooks/useSearchBulkActions.ts
```diff
@@ -519,6 +526,68 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
...
+        }
+
+        const statementParams = getExpensifyCardStatementParamsFromFeed(feed);
+        const {entryIDs} = statementParams;
+
+        const showStatementError = () => {
+            setIsExpensifyCardStatementPDFModalVisible(false);
+            setExpensifyCardStatementPDFParams(undefined);
+            setIsDownloadErrorModalVisible(true);
+        };
```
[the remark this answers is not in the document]
ishpaul777: Good catch — fixed in d3d6664dca8. Each export now gets a request ID (`++expensifyCardStatementRequestIDRef.current`), and `showStatementError` only clears the modal / shows the error while that request is still the latest:

```ts
const requestID = ++expensifyCardStatementRequestIDRef.current;
const showStatementError = () => {
    if (requestID !== expensifyCardStatementRequestIDRef.current) {
        return; // a newer export has started — don't touch its modal
    }
    ...
};
```

This mirrors the `entryIDs` staleness guard already on the success path. Added a regression test ("should not let a stale failed request close the modal for a newer export") that starts export A, switches to export B, then rejects A and asserts B's modal stays open with no error.

ishpaul777:
assigned @youssef-lr for review since Backend PRs was reviewed by him

src/hooks/useSearchBulkActions.ts
```diff
@@ -519,6 +526,78 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
...
+                return;
+            }
+            setIsExpensifyCardStatementPDFModalVisible(false);
+            setExpensifyCardStatementPDFParams(undefined);
+            setIsDownloadErrorModalVisible(true);
+        };
+
+        setExpensifyCardStatementPDFParams(statementParams);
+        setIsExpensifyCardStatementPDFModalVisible(true);
+        getExpensifyCardStatementPDF(statementParams.policyID, statementParams.feedCountry, entryIDs)
```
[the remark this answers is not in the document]
ishpaul777: Fixed in 8bffcbf5748. Select-all-matching only loads the visible rows and the statement has no whole-query export path (unlike CSV/templates), so I now hide the option entirely when `areAllMatchingItemsSelected` rather than export a partial set:

```ts
if (expensifyCardStatementSelection && !areAllMatchingItemsSelected) {
    exportOptions.push({ ... });
}
```

Added a test ("should not offer Export as PDF when all matching items are selected").

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,188 @@
...
+    searchData: SearchResultDataType | undefined,
+): ExpensifyCardStatementSelection | undefined {
+    if (!isExpensifyCardStatementSearch(queryJSON) || !selectedTransactions) {
+        return undefined;
+    }
+
+    // A transaction-narrowing filter (merchant, category, amount, etc.) would make the PDF disagree with the
+    // on-screen rows, so the export is only offered for the unfiltered settlement view (workspace scope aside).
+    if (!hasOnlyStatementScopeFilters(queryJSON)) {
+        return undefined;
```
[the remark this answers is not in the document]
ishpaul777: Good catch — fixed in 5d9f285780e. `status` is a root field on `queryJSON`, not in `flatFilters`, so the narrowing guard missed it. `hasOnlyStatementScopeFilters` now also requires the expense status to be `ALL`:

```ts
if (queryJSON.status !== CONST.SEARCH.STATUS.EXPENSE.ALL) {
    return false;
}
```

So `status:unreported`/`drafts`/etc. hide the export. Added a test ("hides the export when a non-all expense status narrows the rows").

trjExpensify:
@joekaufmanexpensify I think you should take the review of this PR. 👍 

One thing immediately from the OP of the PR is that it seems to not be following the spec of the OP of the issue: 

>Add Export ▸ Export as PDF to Search bulk actions when the user selects one or more Expensify Card settlements (completed or pending) in the settlements view.

VS

>The Download statement PDF option should go in the bulk actions menu as a new option, rather than inside of Export.

So my understanding is that this is not to be nested inside "Export" and it's an option for "Download statement" or something.

joekaufmanexpensify:
Yeah, the intent was to have it be its own top-level option, as that's most consistent with how OldDot treats it. I am not super passionate on that, but I still like it most for discoverability purposes.

ishpaul777:
-- written by claude on Ishpaul's Behalf

Good catch, thanks @trjExpensify / @joekaufmanexpensify — moved it out of `Export` in 654ee43bc49. `Download statement PDF` is now its own top-level bulk-action option (sitting next to `Download as PDF`), matching the issue spec and OldDot. Renamed the string accordingly (`export.downloadStatementPDF` → "Download statement PDF") and updated the tests.

joekaufmanexpensify:
Tested, and working really well! Some feedback:

- I see the statement download option when I select settlements on different workspaces and then error message when trying to download them. Which is good. Can we update the error copy to be: "Please select settlements from one Expensify Card feed at a time."
- I'm not totally following the "this workspace's share of each settlement" line on the statement. Where did that come from?
- I thought we were going to [add the settlement state badges to the statement](https://expensify.slack.com/archives/C07NMDKEFMH/p1781027454497559?thread_ts=1780331742.072979&cid=C07NMDKEFMH)?

image

image

ishpaul777:
Thanks for testing, @joekaufmanexpensify !

1. Error copy - Will do, I'll update it to your suggested wording.

2. "this workspace's share" line - this came up as review from codex in Backend implementation, a single settlement can bundle transactions from multiple workspaces, but the statement is scoped to the one workspace named at the top so for a mixed settlement we reduce the total to just that workspace's portion, and that line is the note explaining it. Agreed it's noise in the common single-workspace case (no proration happens). I'll hide it unless a settlement was actually scoped down, so it only shows on a genuinely mixed-workspace statement Does that sound good?

3. Settlement state badges -  i filed it into a follow-up so it doesn't block this PR: [Expensify/Expensify#647943](https://github.com/Expensify/Expensify/issues/647943). On my list.

joekaufmanexpensify:
>Error copy - Will do, I'll update it to your suggested wording.

Thanks!

> Settlement state badges - i filed it into a follow-up so it doesn't block this PR: https://github.com/Expensify/Expensify/issues/647943. On my list.

Nice one!

>"this workspace's share" line - this came up as review from codex in Backend implementation, a single settlement can bundle transactions from multiple workspaces, but the statement is scoped to the one workspace named at the top so for a mixed settlement we reduce the total to just that workspace's portion, and that line is the note explaining it. Agreed it's noise in the common single-workspace case (no proration happens). I'll hide it unless a settlement was actually scoped down, so it only shows on a genuinely mixed-workspace statement Does that sound good?

Interesting. When a single settlement bundles transactions from multiple workspaces, is that referring to a scenario where an Expensify Card feed originally associated with one workspace/domain is shared across multiple workspaces?

truph01:
@ishpaul777  I encountered this BE error when trying to call `GetExpensifyCardStatementPDF`.

```
{
    "jsonCode": 404,
    "message": "User doesn't have access to policy",
    "onyxData": [],
    "requestID": "a0df65511d062f3c-HKG"
}
```


Could you check whether it is expected?

src/hooks/useSearchBulkActions.ts
```diff
@@ -519,6 +526,78 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
...
+            return;
+        }
+
+        const statementParams = getExpensifyCardStatementParamsFromFeed(feed);
+        const {entryIDs} = statementParams;
+        const requestID = ++expensifyCardStatementRequestIDRef.current;
+
+        // Only surface the failure while this is still the latest export. If the user started another one in
+        // the meantime, this older request must not close the newer modal.
+        const showStatementError = () => {
```
truph01, line 614: Should we move this function out of `exportExpensifyCardStatementPDF`?

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }, [isVisible]);
+
+    useEffect(() => {
+        if (!isVisible || !shouldAutoDownloadPDF.current || !hasFinishedPDFDownload || !statementKey || !statementFileName) {
+            return;
+        }
+
+        downloadExpensifyCardStatementPDF(translate, statementFileName, statementKey, currentUserLogin, encryptedAuthToken, environment);
+        shouldAutoDownloadPDF.current = false;
+    }, [currentUserLogin, encryptedAuthToken, environment, hasFinishedPDFDownload, isGeneratingPDF, isVisible, statementFileName, statementKey, translate]);
```
truph01: I think we can remove `currentUserLogin, encryptedAuthToken, environment, statementFileName, statementKey, translate` from the dependencies. So, the useEffect only triggers when `hasFinishedPDFDownload, isGeneratingPDF, and isVisible` change.

truph01:
@ishpaul777  Should we align the E-card statement PDF download modal with the CSV export modal?

CSV export modal's style:

Screenshot 2026-06-19 at 10 45 46

E-card statement PDF download modal's style:

Screenshot 2026-06-19 at 10 46 07

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+                    </View>
+                    <Button
+                        style={[styles.mt3, styles.noSelect]}
+                        onPress={() => {
+                            if (!hasFinishedPDFDownload || !statementKey || !statementFileName) {
+                                onClose();
+                                return;
+                            }
+
+                            downloadExpensifyCardStatementPDF(translate, statementFileName, statementKey, currentUserLogin, encryptedAuthToken, environment);
```
truph01: The modal should be closed once user clicks on "Download" button

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+                                <ActivityIndicator
+                                    size={CONST.ACTIVITY_INDICATOR_SIZE.SMALL}
+                                    color={theme.textSupporting}
+                                    style={styles.ml3}
+                                    reasonAttributes={pdfLoadingReasonAttributes}
+                                />
+                            </View>
+                        )}
+                    </View>
+                    <Button
```
truph01: Should we add style `success` to this button if `hasFinishedPDFDownload` is `true`?

ishpaul777:
> @ishpaul777 Should we align the E-card statement PDF download modal with the CSV export modal?

https://github.com/user-attachments/assets/38b7d4f3-deb4-4aa3-85a7-98a11a3869c5

i kept the styling similar to Report pdf download, since we dont yet have support to send pdf as message

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
ishpaul777: -- written by claude on Ishpaul's Behalf

Done in 10607aa5919 — `success={hasFinishedPDFDownload}` so the button turns green once the PDF is ready.

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
ishpaul777: -- written by claude on Ishpaul's Behalf

Done in 10607aa5919 — clicking Download now closes the modal.

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
ishpaul777: -- written by claude on Ishpaul's Behalf

I'd keep these. They're genuinely read inside the effect (the filename/key/translate/env are passed to `downloadExpensifyCardStatementPDF`), so trimming them triggers a `react-hooks/exhaustive-deps` error. And the extra triggers are harmless — the `shouldAutoDownloadPDF.current` ref guard means the auto-download still fires only once per modal open; any extra run just hits the early `return`. This also matches `ReportPDFDownloadModal`, which keeps the same dependency set.

src/hooks/useSearchBulkActions.ts
ishpaul777: -- written by claude on Ishpaul's Behalf

I'd prefer to keep it inline — it closes over the per-call `requestID` (the stale-request guard) plus the request-scoped setters, so it's specific to each export invocation. Pulling it out would mean threading `requestID` through as an argument for no real benefit.

ishpaul777:
> Interesting. When a single settlement bundles transactions from multiple workspaces, is that referring to a scenario where an Expensify Card feed originally associated with one workspace/domain is shared across multiple workspaces?


Close  it's at the domain level rather than a workspace's feed being shared. The Expensify Card feed and its settlements live on the domain, not a single workspace. A settlement (one ACH withdrawal / entryID) bundles all the card transactions that settled together for that domain + feed.

Within that domain, cardholders can belong to different workspaces. We determine a transaction's workspace from the report it's on (report.details.policyID). So when a domain has more than one workspace, a single settlement can contain transactions spanning those workspaces.

Since the statement is scoped to the one workspace named at the top, for such a settlement we keep only that workspace's transactions and reduce the settlement's payment total to their sum — that's the "share" line


@joekaufmanexpensify

truph01:
> @ishpaul777 I encountered this BE error when trying to call `GetExpensifyCardStatementPDF`.
> 
> ```
> {
>     "jsonCode": 404,
>     "message": "User doesn't have access to policy",
>     "onyxData": [],
>     "requestID": "a0df65511d062f3c-HKG"
> }
> ```
> 
> Could you check whether it is expected?

@ishpaul777  Could you check this comment?

ishpaul777:
> > @ishpaul777 I encountered this BE error when trying to call `GetExpensifyCardStatementPDF`.
> > ```
> > {
> >     "jsonCode": 404,
> >     "message": "User doesn't have access to policy",
> >     "onyxData": [],
> >     "requestID": "a0df65511d062f3c-HKG"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you check whether it is expected?
> 
> @ishpaul777 Could you check this comment?

ohw sorry i missed it earlier, i think i can i know whats wrong with BE raising a fix will let you know 👍

joekaufmanexpensify:
>Close it's at the domain level rather than a workspace's feed being shared. The Expensify Card feed and its settlements live on the domain, not a single workspace. A settlement (one ACH withdrawal / entryID) bundles all the card transactions that settled together for that domain + feed.
>
>Within that domain, cardholders can belong to different workspaces. We determine a transaction's workspace from the report it's on (report.details.policyID). So when a domain has more than one workspace, a single settlement can contain transactions spanning those workspaces.
>
>Since the statement is scoped to the one workspace named at the top, for such a settlement we keep only that workspace's transactions and reduce the settlement's payment total to their sum — that's the "share" line

I'm concerned that this workspace restriction, which may reduce settlement totals, is not what we want. In OldDot, the statement feature only downloads complete settlements for the selected period. This is useful because it allows the admin to reconcile the settlement with the debit to their bank account and view the underlying transactions that comprise it. If we allow them to download partial settlements on the NewDot statement, the settlement total won't match any debit in their bank account. I'm not sure it serves any purpose if that isn't possible. 

I understand that, practically speaking, limiting by workspace will often result in the full settlement. Because it's most common for all Expensify Card transactions from one feed to be reported in the same workspace. Our universal feeds design will make it more common to shift away from that model, though. 

In my opinion, we'd be better served by only allowing this PDF statement to be generated with full settlements from the same Expensify Card feed. So if the user selects multiple settlements from different Expensify Card feeds OR a subset of an Expensify Card settlement that spans multiple workspaces (with a workspace filter), we'd display an error. 

Thoughts?

joekaufmanexpensify:
I'm also going to tax @madmax330 in to review this PR. As he designed the feature for sharing Expensify Card feeds across workspaces and will likely have thoughts on this.

trjExpensify:
Maybe I'm not following, but I feel like that would be what someone wants if they filter by workspace?

- If they only want to see the transactions and settlements for a given workspace, put a workspace filter on. 
- If they want to see all the transactions settled on the feed, don't put a workspace filter on. (which is the default for the bank rec suggested search)

Reminds me of the discussion with that customer call the other week that have a number of different brands which they allocate cards to - then sometimes a specific brand will need to produce a statement for something. The limitation they have with Classic is that when they download the statement report it's all settlements and transactions for all cards. 

What am I missing?

joekaufmanexpensify:
The scenario I'm talking about is only when there is a shared Expensify Card feed across workspaces. There's only one settlement for each Expensify Card feed. We don't break them out by workspace (even if the feed is being used on multiple workspaces). If we allow them to generate a statement for one workspace in that scenario, it will show partial settlements that don't correspond to anything in their bank statement. I feel like that will be confusing. 

I guess if they don't want to use it for reconciliation purposes and just want a statement for one workspace's activity (which contains partial settlements only) it will work for that. 

I think we agree that we don't want to force limit this to a selected workspace though, yeah? Meaning, if you select a settlement from one Expensify Card feed that is split across multiple workspaces, we should allow generating a statement for that settlement. Otherwise, it wouldn't be possible to generate a statement with the complete settlements for that type of Expensify Card feed.

madmax330:
I agree with both @joekaufmanexpensify and @trjExpensify. 
I think we want to have the ability to download/export the full settlement for everyone on every policy. 
If we have that then I think if people find it useful to filter per policy I don't see why would disallow it 🤷🏽 
We might want to put something on the statement title though to say that it is filtered

trjExpensify:
>I think we agree that we don't want to force limit this to a selected workspace though, yeah? Meaning, if you select a settlement from one Expensify Card feed that is split across multiple workspaces, we should allow generating a statement for that settlement. Otherwise, it wouldn't be possible to generate a statement with the complete settlements for that type of Expensify Card feed.

For sure, but my understanding is that we aren't forcing workspace selection - and the default for this view is feed filter, no workspace filter.

joekaufmanexpensify:
Maybe I was misunderstanding, but I thought this [comment meant ](https://github.com/Expensify/App/pull/92790#issuecomment-4753336187) we were not showing complete settlements that span across multiple workspaces on the statement, even if no workspace filter was selected. Perhaps @ishpaul777 can clarify

ishpaul777:
Thanks all, let me clarify exactly where it stands today since the behavior isn't what we want yet.

### Backend (already correct)

`GetExpensifyCardStatementPDF` takes the selected `entryIDs` and an optional `policyID`.
* **No `policyID`:** Returns the complete settlement(s) — every transaction across all workspaces, full settlement total. No scoping.
* **With `policyID`:** Scopes to that workspace (keeps only its transactions, reduces the total to its share) and the statement shows a *"This workspace's share of each settlement"* note.

So the backend already supports both "full settlement" and "filtered to one workspace."

### Frontend (current PR — the part that's wrong)

Today the client drops any settlement that spans multiple workspaces and only lets you export settlements that map to a single workspace. So @joekaufmanexpensify is right: as it stands, complete cross-workspace settlements aren't exportable at all, even with no workspace filter. That's the gap.

### Proposed Fix

* Stop dropping cross-workspace settlements.
* **No workspace filter:** Send no `policyID` -> full settlement (the reconciliation default).
* **Workspace filter applied:** Send that `policyID` -> scoped statement (with the "filtered" note already on the PDF).

### Product Question

One product question before I finalize: on a full (cross-workspace) statement, what should the workspace header shows in statement

Today it's labeled `"Workspace: {name}"`, which is wrong when it's not scoped to one. Should it show the domain (or feed/program) name instead? @madmax330's point about labeling filtered statements is handled by the existing "share" note; this is the inverse labeling the unfiltered one.

trjExpensify:
>Today it's labeled "Workspace: {name}",

Isn't that the same situation as: 

1. Add a feed to workspace 1
2. Link it to workspace 2
3. Access the feed in workspace 2, isn't it "called" workspace 1?

ishpaul777:
i might be wrong but that is company-card feed sharing this statement feature is Expensify Card only (gated to withdrawal-type:expensify-card)

Expensify Card feeds work differently: they're domain-level and auto-provisioned, so there's no "originating workspace" to inherit a name from. For a settlement that spans workspaces, the frontend's search payload has no policyID and no feed/workspace name just the entryID, program (US/GB/etc.), and bank + masked account.

joekaufmanexpensify:
>Expensify Card feeds work differently: they're domain-level and auto-provisioned, so there's no "originating workspace" to inherit a name from. For a settlement that spans workspaces, the frontend's search payload has no policyID and no feed/workspace name just the entryID, program (US/GB/etc.), and bank + masked account.

My understanding is same as Tom's. Here are some example internal Expensify Card feeds. Three of them are workspace feeds (provisioned from a workspace in New Expensify). One is a domain feed from Expensify Classic. The workspace feeds show the workspace name as the feed name. 

image

trjExpensify:
Yeah cool, thought as much. 👍

ishpaul777:
Thanks for pushing on this, and apologies for the earlier misdirection I'd been treating the card program (feedCountry) as the "feed," which sent me the wrong way on naming and grouping. I've now traced the full flow and it's much clearer.

The key correction: a feed is identified by its fundID (the workspace the feed is provisioned under), not the program. One program can have several feeds, so the program alone can't identify one and that workspace is the canonical "origin workspace" name we were looking for.

Updated plan:

- Download pdf is unscoped by default — export the whole settlement; only scope to a workspace when the search is explicitly filtered to one.
- Group settlements by fundID, and show the multi-feed warning only when user selection spans more than one feed.
- Backend derives the owning workspace from the settlement itself and stamps its name, so the header reads Workspace: <name of workspace owning the feed>.

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+    }, [isVisible]);
+
+    useEffect(() => {
+        if (!isVisible || !shouldAutoDownloadPDF.current || !hasFinishedPDFDownload || !statementKey || !statementFileName) {
+            return;
+        }
+
+        downloadExpensifyCardStatementPDF(translate, statementFileName, statementKey, currentUserLogin, encryptedAuthToken, environment);
+        shouldAutoDownloadPDF.current = false;
+    }, [currentUserLogin, encryptedAuthToken, environment, hasFinishedPDFDownload, isGeneratingPDF, isVisible, statementFileName, statementKey, translate]);
```
truph01: Is `isGeneratingPDF` required in these dependencies? I ask since `isGeneratingPDF` doesn't appear in the useEffect's body.

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,207 @@
...
+    fileName: string,
+    statementKey: string,
+    currentUserEmail: string,
+    encryptedAuthToken: string,
+    environment: EnvironmentType,
+): Promise<void> {
+    const baseURL = addTrailingForwardSlash(getOldDotURLFromEnvironment(environment));
+    const downloadFileName = `Expensify_Card_Statement_${statementKey}.pdf`;
+    const pdfURL = `${baseURL}secure?secureType=pdfreport&filename=${encodeURIComponent(fileName)}&downloadName=${encodeURIComponent(downloadFileName)}&email=${encodeURIComponent(currentUserEmail)}`;
+    return fileDownload(translate, addEncryptedAuthTokenToURL(pdfURL, encryptedAuthToken, true), downloadFileName, '');
```
truph01: Do we need `return`  here? I asked since I don't see any place where we use this returned value.

truph01:
> @ishpaul777 I encountered this BE error when trying to call `GetExpensifyCardStatementPDF`.
> 
> ```
> {
>     "jsonCode": 404,
>     "message": "User doesn't have access to policy",
>     "onyxData": [],
>     "requestID": "a0df65511d062f3c-HKG"
> }
> ```
> 
> Could you check whether it is expected?

@ishpaul777 Could you share an update on this concern? I tested, but the error is still encountered.

ishpaul777:
@truph01 BE still not merged on this, i'll update once done meanwhile feel free to use my local vm via shared ngrok

truph01:
> @truph01 BE still not merged on this, i'll update once done meanwhile feel free to use my local vm via shared ngrok

Ah, thanks for your information!

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,207 @@
...
+    translate: LocalizedTranslate,
+    fileName: string,
+    statementKey: string,
+    currentUserEmail: string,
+    encryptedAuthToken: string,
+    environment: EnvironmentType,
+): Promise<void> {
+    const baseURL = addTrailingForwardSlash(getOldDotURLFromEnvironment(environment));
+    const downloadFileName = `Expensify_Card_Statement_${statementKey}.pdf`;
+    const pdfURL = `${baseURL}secure?secureType=pdfreport&filename=${encodeURIComponent(fileName)}&downloadName=${encodeURIComponent(downloadFileName)}&email=${encodeURIComponent(currentUserEmail)}`;
```
truph01: We can consider using `buildSecureDownloadURL` here

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+import Text from './Text';
+
+type ExpensifyCardStatementPDFDownloadModalProps = {
+    statementParams: ExpensifyCardStatementParams;
+    isVisible: boolean;
+    onClose: () => void;
+    onModalHide?: () => void;
+};
+
+function ExpensifyCardStatementPDFDownloadModal({statementParams, isVisible, onClose, onModalHide}: ExpensifyCardStatementPDFDownloadModalProps) {
```
truph01, line 22: The new modal (133 lines) shares ~90% of its structure with `ReportPDFDownloadModal.tsx` (135 lines): same shouldAutoDownloadPDF ref pattern, same two-useEffect auto-download logic, same Modal layout (Header + Text + ActivityIndicator + Button + close PressableWithFeedback). They differ only in the Onyx key read and the download function called. A shared generic PDFDownloadModal component parameterized by a status resolver and download callback would eliminate ~100 lines of duplication.

truph01:
> Add Export ▸ Export as PDF to Search bulk actions

> Select one settlement group (select the whole group) → open bulk actions → Export → Export as PDF.

@ishpaul777  Could you update PR description? Now, we add "Export as PDF" as a standalone top-level bulk action

truph01:
@codex review

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
ishpaul777: Done in 9c9f944 — extracted a shared `PDFDownloadModal` that both this modal and `ReportPDFDownloadModal` now use. The generic component owns the auto-download ref + effects, the modal layout, and the close button; each caller keeps only its Onyx read and download call. Report's behavior is preserved via `shouldCloseOnDownload`/`shouldUseSuccessButton` (both default to its original behavior; the card statement modal opts in).

-- written by claude on Ishpaul's Behalf

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Done in cc0f5f6 — removed the return; the caller fires-and-forgets the download, so the Promise was unused. (return type is now void.)

-- written by claude on Ishpaul's Behalf

src/components/ExpensifyCardStatementPDFDownloadModal.tsx
ishpaul777: Done in 705380e — dropped isGeneratingPDF from the effect deps; it isn't referenced in the body (hasFinishedPDFDownload, which derives from it, is what's used).

-- written by claude on Ishpaul's Behalf

src/libs/ExpensifyCardStatementUtils.ts
ishpaul777: Done in 02b3a69 — now using buildSecureDownloadURL from UrlUtils, matching the report and travel-invoice download paths.

-- written by claude on Ishpaul's Behalf

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,207 @@
...
+    if (queryJSON.type !== CONST.SEARCH.DATA_TYPES.EXPENSE) {
+        return false;
+    }
+
+    if (queryJSON.groupBy !== CONST.SEARCH.GROUP_BY.WITHDRAWAL_ID) {
+        return false;
+    }
+
+    const withdrawalTypeFilter = queryJSON.flatFilters?.find((filter) => filter.key === CONST.SEARCH.SYNTAX_FILTER_KEYS.WITHDRAWAL_TYPE);
+    return withdrawalTypeFilter?.filters?.some((filter) => filter.value === CONST.SEARCH.WITHDRAWAL_TYPE.EXPENSIFY_CARD) ?? false;
```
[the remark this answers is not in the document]
ishpaul777: Done in 91d94a5 — isExpensifyCardStatementSearch now requires the withdrawal-type filter to be exactly `withdrawal-type:expensify-card` (single value, operator EQUAL_TO). Negated (neq) and mixed filters like `reimbursement,expensify-card` no longer enable the action.

-- written by claude on Ishpaul's Behalf

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// Returns the single workspace the search is filtered to, or undefined. We scope the statement to a workspace only
+// when the user filtered by one, so the PDF matches the on-screen rows; otherwise the whole settlement is exported.
+// policyID is a root-level search key (like type/groupBy), so it lives on queryJSON.policyID, not in flatFilters.
+function getScopedPolicyID(queryJSON: SearchQueryJSON | undefined): string | undefined {
+    const policyID = queryJSON?.policyID;
+    if (typeof policyID === 'string') {
+        return policyID;
+    }
+    return policyID?.length === 1 ? policyID.at(0) : undefined;
```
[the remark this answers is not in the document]
ishpaul777: Done in d748527 — when the search is filtered to multiple workspaces (e.g. policyID:A,B), getExpensifyCardStatementSelection now returns undefined so the action is hidden, rather than falling back to an unscoped export that would include workspaces outside the filter.

-- written by claude on Ishpaul's Behalf

src/hooks/useSearchBulkActions.ts
```diff
@@ -532,6 +539,78 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
...
+                    // instead of leaving it stuck on the loading state.
+                    showStatementError();
+                    return;
+                }
+
+                // Sync the modal to the server's cache key, but only while the on-screen selection is still
+                // the one we requested. If the user changed the selection or dismissed the modal in the
+                // meantime, this response is stale and must not overwrite the current params.
+                setExpensifyCardStatementPDFParams((currentParams) =>
+                    currentParams && currentParams.entryIDs.join(',') === entryIDs.join(',') ? {...currentParams, statementKey} : currentParams,
```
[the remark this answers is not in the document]
ishpaul777: Done in c5dc8ca — the success handler now gates on the latest request id (like the failure path) instead of only comparing entryIDs, so a cancel + re-scope + re-export of the same settlement IDs can't let a stale earlier response write its statementKey into the current modal.

-- written by claude on Ishpaul's Behalf

ishpaul777:
BE [PR](https://github.com/Expensify/Auth/pull/22487) is still awaiting final review and merge

ishpaul777:
@truph01 BE PRs deployed please test and let me know if it works for you now

ishpaul777:
cc @joekaufmanexpensify incase you like to test again on adhoc build

truph01:
@codex review

src/libs/ExpensifyCardStatementUtils.ts
```diff
@@ -0,0 +1,232 @@
...
+    const selectedGroupKeys = new Set<string>();
+    for (const [key, selection] of Object.entries(selectedTransactions)) {
+        if (!selection?.isSelected) {
+            continue;
+        }
+        if (key.startsWith(CONST.SEARCH.GROUP_PREFIX)) {
+            selectedGroupKeys.add(key);
+        }
+        if (selection.groupKey?.startsWith(CONST.SEARCH.GROUP_PREFIX)) {
+            selectedGroupKeys.add(selection.groupKey);
```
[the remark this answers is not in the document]
ishpaul777: Good catch — fixed in a08e877e828.

`getSelectedSettlementGroups` no longer treats a single tagged transaction as the whole settlement. A settlement is now only included when either its row is selected directly (collapsed group key in `selectedTransactions`) or **all** of its transactions are selected (selected count for that `groupKey` ≥ the group's `count`). A single expanded child no longer enables the export.

Added coverage: "hides the export when only some of a settlement's transactions are selected" and "includes a settlement when all of its transactions are selected".

-- written by claude on Ishpaul's Behalf

joekaufmanexpensify:
Testing well! Only thing I noticed is still seeing the old error copy. Can we update this to: `Please select settlements from one Expensify Card feed at a time`?

Otherwise looks good to me 👍 

image

ishpaul777:
Updated in 21befc7959a — the alert now reads:

> Please select settlements from one Expensify Card feed at a time.

Updated the English copy and all translations. Thanks for the review! 🙏

-- written by claude on Ishpaul's Behalf

joekaufmanexpensify:
Updated error looks good.

joekaufmanexpensify (dismissed):
Good from my perspective 👍

truph01:
@ishpaul777 Could you check this `requestID: "a1858c7c69ccdd51-SJC"` when calling `GetExpensifyCardStatementPDF`, I encountered error:
```
jsonCode: 401
message: "401 Unauthorized"
```

truph01 (approved):

src/hooks/useSearchBulkActions.ts
```diff
@@ -1970,10 +2051,24 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
                     }
                     const exportID = exportReportsToPDF(selectedReportIDs);
                     trackExport(exportID);
+                    clearSelectedTransactions();
```
truph01: The fix added here doesn't seem related to the purpose of this PR. It also breaks the behavior of the `CONST.SEARCH.BULK_ACTION_TYPES.DOWNLOAD_PDF` action.

The action that should be updated is `CONST.SEARCH.BULK_ACTION_TYPES.DOWNLOAD_STATEMENT_PDF`, not `CONST.SEARCH.BULK_ACTION_TYPES.DOWNLOAD_PDF`.

ishpaul777: ah Good catch agreed. Moved it to the right place

truph01 (approved):

youssef-lr (approved):

madmax330 (approved):

PR #93070, opened 2026-06-09
Revert "fix: Copilot with limited access level can retract expenses"

lakchote:
Reverts Expensify/App#90076


youssef-lr (approved):

JmillsExpensify (commented):
Doesn't look like my review is required.

greg-schroeder:
@MelvinBot The latter situation is correct; this is a product decision and Limited Access Copilots should be allowed to retract reports they've submitted

PR #93055, opened 2026-06-09
[NO QA] refactor(mfa): remove flow engine, scaffold executeScenario to success screen

dariusz-biela:
### Explanation of Change

Removes the imperative MFA flow engine from `MultifactorAuthenticationMainContext.tsx`, the first step of the MFA -> XState state-machine migration.

- Deletes `process()`, `handleCallback()`, and the driving `useEffect`, together with the `eslint-disable react-hooks/exhaustive-deps`
- The orchestration layer is intentionally left **inert**: the split contexts (`StateContext` / `ActionsContext`) and `stateReducer` stay in place so every consumer page still compiles.

To keep the modal and outcome UI testable while the engine is gone, `executeScenario` navigates straight to the success outcome screen after `INIT`:

- Dispatches `SET_FLOW_COMPLETE` so back-press / backdrop close the modal directly instead of opening the cancel-confirm dialog.
- Wraps the navigation in `Navigation.runAfterTransition`, preserving the Android outcome-screen animation-race fix that the old `handleCallback` had.

This success-screen jump is a scaffold; the real device-check / registration / authorization states arrive with the machine slices.

### Fixed Issues

Part of https://github.com/Expensify/App/issues/81197
PROPOSAL: N/A (internal refactor, part of the MFA -> XState migration plan)

### Tests

1. Trigger any MFA scenario (e.g. Reveal PIN, or the Biometrics test tool row).
2. Verify the MFA modal opens and lands on the success outcome screen.
3. Tap the confirm / close button and verify the modal closes.
4. Re-open the flow, then press browser / Android back (or tap the backdrop on a wide layout) and verify the modal closes directly, with no cancel-confirm dialog.
5. Verify no errors appear in the JS console.

- [x] Verify that no errors appear in the JS console

### Offline tests

N/A - the flow engine is removed in this PR, so no network steps run. Offline resilience is handled in a later machine slice.

### QA Steps

[No QA] - internal integration-branch PR targeting `dariusz-81197-mfa-state-machine`; it is not shipped to staging or production. Covered by the manual Tests above.

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
[...]

Issue #81197, opened 2026-02-02
[ECUK In-App 3DS] [App]  Refactor MultifactorAuthenticationContextProvider - fix React rules violations and race conditions risks


dariusz-biela:
## Background

The `MultifactorAuthenticationContextProvider` component (`src/components/MultifactorAuthentication/Context/Main.tsx`) manages the entire Multifactor Authentication (MFA) flow using a `useEffect` that runs an async `process` function whenever state changes.

### Current Implementation

The component has a complex flow orchestration pattern:

1. `executeScenario()` is called to start the MFA flow
2. A `useEffect` watches for state changes and calls `process()`
3. `process()` is an async function that:
   - Reads current state
   - Determines the next step via if-statements
   - Performs async operations (API calls, biometric prompts)
   - Updates state to trigger the next iteration

```tsx
useEffect(() => {
    if (!state.scenario) {
        return;
    }

    process();
    // We intentionally omit `process` and `state` from dependencies.
    // Including them would cause infinite re-renders since `process` is recreated on every state change.
    // Instead, we list only the specific state fields that should trigger a re-run of the MFA flow.
    // eslint-disable-next-line react-hooks/exhaustive-deps
}, [
    state.error,
    state.scenario,
    state.softPromptApproved,
    state.validateCode,
    state.registrationChallenge,
    state.authorizationChallenge,
    state.isRegistrationComplete,
    state.isAuthorizationComplete,
    state.isFlowComplete,
]);
```

## Current Implementation - Strengths

Before diving into the problems, it's worth noting what the current implementation does well:

### 1. Readable Flow Structure

The `process()` function provides a clear, top-to-bottom view of the entire MFA flow. Each step is protected by explicit conditions (if-statements), making it easy to understand:
- What the flow looks like end-to-end
- What conditions must be met before each step executes
- How the flow progresses from one step to the next

### 2. Why It Currently Works

The current implementation functions correctly because of two key invariants:

1. **Dependency array contains only flow-controlled state**: The `useEffect` dependencies are narrowed down to state fields that only change at well-defined points in the flow - either by `process()` steps (e.g., `registrationChallenge`, `isRegistrationComplete`) or by expected user interactions that signal flow continuation (e.g., `validateCode`, `softPromptApproved`). These fields never change at unexpected times.

2. **State changes occur only at step completion**: Every execution path through `process()` dispatches state updates only at the very end of the step, after all async operations have completed. No state mutations occur mid-step while async operations are in flight.

[...]

PR #93180, opened 2026-06-10
[NO QA] refactor(mfa): add the XState foundation and move the modal lifecycle into the machine

dariusz-biela:
### PR Stack

This PR is part of the MFA -> XState state-machine migration (https://github.com/Expensify/App/issues/81197), shipped as a stacked-PR train. All work merges into the integration branch `dariusz-81197-mfa-state-machine` on `Expensify/App`; nothing reaches `main` until the final integration merge.

| # | PR | Branch | Base |
|---|----|--------|------|
| 1 | https://github.com/Expensify/App/pull/93055 (merged) | `dariusz-biela/refactor/3ds/mfa-cleanup-for-state-machine` | `dariusz-81197-mfa-state-machine` (integration) |
| 2 | **this PR** | `dariusz-biela/refactor/3ds/mfa-state-machine-setup` | `dariusz-81197-mfa-state-machine` (integration) |

This branch is stacked on top of the PR-1 branch, but that branch exists only on the SWM fork and cannot be used as a base on `Expensify/App`, so this PR targets the integration branch directly. PR-1 (https://github.com/Expensify/App/pull/93055) has already merged into the integration branch, so the diff here contains only this PR's own changes.

### Explanation of Change

Second step of the MFA -> XState migration. PR-1 removed the imperative flow engine and left the orchestration layer inert; this PR introduces the XState foundation and makes a machine skeleton the single owner of the flow.

Goals:

- Land the XState dependencies and the machine module that all later migration slices build on.
- Move ownership of the modal lifecycle (opening, visible screen, close animation, teardown and state wipe) into the machine, removing the duplicated lifecycle state previously spread across React components.
- Keep every consumer page compiling and behaving unchanged: consumers still read the legacy `State` shape (with the stored `isModalOpen` flag replaced by the machine-derived `modalState`), now derived from the machine snapshot.

What changed, at a high level:

- Added `xstate` 5.31.1 and `@xstate/react` 5.0.5.
- New `src/components/MultifactorAuthentication/machine/` module: the machine skeleton, its types, and a `snapshotToState` mapper that derives the state consumers read (the legacy `State` shape plus `modalState`) from the machine snapshot. Top-level machine states model the modal lifecycle; the children of `open` map 1:1 to the visible screens. In this slice `INIT` still short-circuits to the success outcome screen - the real device-check / registration / authorization states arrive in later slices.
- The `MultifactorAuthenticationMainContext` Provider became a thin typed facade over the machine's `send()`, with no flow logic of its own. Its API is split into two contexts: `MultifactorAuthenticationExternalApiContext` exposes only `executeScenario` (the entry point app code uses to start a flow), and `MultifactorAuthenticationInternalApiContext` exposes the machine-derived `state` plus `send()` wrappers for the screens and navigator hosting an active flow - it is deliberately not exported from the Context barrel.
- The legacy `State` type and `DEFAULT_STATE` moved to `Context/state.ts`, and `isModalOpen` is gone from the shape: modal visibility is no longer stored state, it is derived from the chart as `modalState`. The navigator no longer dispatches `RESET` either - the state wipe is machine-owned, running on every entry to `closed` (the legacy reducer still defines `RESET`; the remaining reducer actions are migrated and removed in later slices).
- New `getScenarioConfig()` accessor in `config/scenarios` centralizes scenario-config lookups as a generic, fully typed indexed access (no type assertion needed); the Provider reads its scenario config through it.
- The modal navigator no longer keeps its own phase state: it mounts, animates teardown, and unmounts based on the machine-derived `modalState`, and a machine-owned fallback guarantees the flow cannot get stranded if the close animation is cancelled mid-way.
- Fix: the modal backdrop now fades out with the screen slide-out's own duration and easing instead of being cut off a third of the way through.

image

### Fixed Issues

$ https://github.com/Expensify/App/issues/81197
PROPOSAL: N/A (internal refactor, part of the MFA -> XState migration plan)

### Tests

1. Trigger any MFA scenario (e.g. Reveal PIN, or the Biometrics test tool row).
2. Verify the MFA modal opens and lands on the success outcome screen.
3. Tap the confirm / close button and verify the modal slides out, with the backdrop fading for the whole length of the animation, and closes.
4. Re-open the flow, then press browser / Android back (or tap the backdrop on a wide layout) and verify the modal closes directly, with no cancel-confirm dialog.
5. Re-open the flow once more and verify it starts cleanly again (flow state is wiped on close).
6. Verify no errors appear in the JS console.

### Offline tests

N/A - this slice still runs no network steps (the scaffold jumps straight to the success screen). Offline resilience is handled in a later machine slice.

### QA Steps

[No QA] - internal integration-branch PR targeting `dariusz-81197-mfa-state-machine`; it is not shipped to staging or production. Covered by the manual Tests above.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

https://github.com/user-attachments/assets/72110e3d-e240-417f-aaf6-21004b7e7150


src/components/MultifactorAuthentication/Context/MultifactorAuthenticationInternalApiContext.ts
```diff
@@ -0,0 +1,47 @@
...
+ * (useMultifactorAuthentication); only flow-internal screens import this module directly.
+ */
+type MultifactorAuthenticationInternalApi = {
+    /** The current MFA state, derived from the machine snapshot. */
+    state: MfaState;
+
+    /** Close the modal overlay. */
+    closeModal: () => void;
+
+    /** Navigator's report that the close animation fully finished; re-enters idle, wiping the flow data. */
```
jnowakow: Could rephrase this comment? It's a bit not clear

src/components/MultifactorAuthentication/Context/state.ts
```diff
@@ -0,0 +1,82 @@
...
+    payload: MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined;
+
+    /** Whether registration step has been completed */
+    isRegistrationComplete: boolean;
+
+    /** Whether authorization step has been completed */
+    isAuthorizationComplete: boolean;
+
+    /** Whether the entire flow has been completed */
+    isFlowComplete: boolean;
```
jnowakow, line 34: Do we need three separate booleans instead of one `step: "registrationComplete" | "authorizationComplete" | "flowComplete"` variable?

jnowakow (commented):
Looks veeery good! 🙌

jakubkalinski0 (approved):
Very well done! 🚀

src/components/MultifactorAuthentication/Context/state.ts
dariusz-biela: I've thought about that before, too, but in this particular case, we actually need it because we use it for logging at the very end to see what flow the user went through.

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationMainContext.tsx
```diff
@@ -82,117 +58,61 @@ function MultifactorAuthenticationContextProvider({children}: MultifactorAuthent
...
-            dispatch({type: 'SET_ERROR', payload: error});
-            return;
-        }
-
-        addMFABreadcrumb('Flow cancelled', {reason: CONST.MULTIFACTOR_AUTHENTICATION.REASON.LOCAL_ERRORS.CANCELED}, 'warning');
-        dispatch({
-            type: 'SET_ERROR',
-            payload: createLocalMFAError(CONST.MULTIFACTOR_AUTHENTICATION.REASON.LOCAL_ERRORS.CANCELED, 'User cancelled the MFA flow'),
-        });
+        send({type: 'INIT', scenarioName, scenario, payload: params && Object.keys(params).length > 0 ? params : undefined});
```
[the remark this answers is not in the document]
dariusz-biela: I'll fix that in a later version when I'm working on the authorization.

config/eslint/eslint.seatbelt.tsv
```diff
@@ -290,7 +290,6 @@
 "../../src/components/MultiGestureCanvas/usePanGesture.ts"	"no-restricted-syntax"	1
 "../../src/components/MultiGestureCanvas/usePinchGesture.ts"	"react-hooks/set-state-in-effect"	1
 "../../src/components/MultiGestureCanvas/useTapGestures.ts"	"no-restricted-syntax"	1
-"../../src/components/MultifactorAuthentication/Context/stateReducer.ts"	"@typescript-eslint/no-unsafe-type-assertion"	1
```
roryabraham, line 293: Note: manually removing entries from this file (aka tightening the seatbelt) is not necessary. It happens automatically on merge to main. Allowing it to happen that way is generally better because it reduces conflicts.

src/components/MultifactorAuthentication/components/OutcomeScreen/OutcomeScreenBase.tsx
```diff
@@ -62,13 +62,9 @@ function OutcomeScreenBase({headerTitle, illustration, iconWidth, iconHeight, ti
     const {translate} = useLocalize();
     const styles = useThemeStyles();
     const {asset: icon} = useMemoizedLazyAsset(() => loadIllustration(illustration));
-    const {dispatch} = useMultifactorAuthenticationActions();
+    const {closeModal} = useMultifactorAuthenticationInternal();
```
roryabraham, line 65: Without having reviewed the full PR, it's jumping out at me that `useMultifactorAuthenticationInternal` is being used by an external consumer here. Which flags to me that perhaps it's not the best name.

src/components/MultifactorAuthentication/config/scenarios/index.ts
```diff
@@ -39,5 +43,17 @@ const Configs = {
...
 
-export default Configs;
+/**
+ * Every entry of the config record satisfies MultifactorAuthenticationScenarioConfig at definition, but the
+ * per-scenario action signatures make the record's value union non-narrowable, so the lookup needs
+ * an assertion. This accessor owns the single assertion so callers don't repeat it; params are
+ * type-guarded separately by ExecuteScenarioParams<T>.
+ */
+function getScenarioConfig(scenarioName: MultifactorAuthenticationScenario): MultifactorAuthenticationScenarioConfig {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
```
roryabraham: Let's resist the urge to suppress lint as much as we can during these refactors. I spent a bit of time noodling with the types here to avoid this suppression, and I think the end result is an improvement.

<details>
<summary>diff</summary>

```diff
diff --git a/src/components/MultifactorAuthentication/Context/state.ts b/src/components/MultifactorAuthentication/Context/state.ts
index 9dd7b200f8a6..71d3f45e6bdf 100644
--- a/src/components/MultifactorAuthentication/Context/state.ts
+++ b/src/components/MultifactorAuthentication/Context/state.ts
@@ -1,7 +1,8 @@
+import type {ValueOf} from 'type-fest';
+import type MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG from '@components/MultifactorAuthentication/config/scenarios';
 import type {
     MultifactorAuthenticationScenario,
     MultifactorAuthenticationScenarioAdditionalParams,
-    MultifactorAuthenticationScenarioConfig,
     MultifactorAuthenticationScenarioResponse,
 } from '@components/MultifactorAuthentication/config/types';
 import type {AuthenticationChallenge, RegistrationChallenge} from '@libs/MultifactorAuthentication/shared/challengeTypes';
@@ -36,7 +37,7 @@ type MultifactorAuthenticationState = {
     scenarioName: MultifactorAuthenticationScenario | undefined;
 
     /** Current scenario configuration being executed */
-    scenario: MultifactorAuthenticationScenarioConfig | undefined;
+    scenario: ValueOf<typeof MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG> | undefined;
 
     /** Additional parameters for the current scenario */
     payload: MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined;
diff --git a/src/components/MultifactorAuthentication/Context/stateReducer.ts b/src/components/MultifactorAuthentication/Context/stateReducer.ts
index b26871bf9b6c..19e527cf128d 100644
--- a/src/components/MultifactorAuthentication/Context/stateReducer.ts
+++ b/src/components/MultifactorAuthentication/Context/stateReducer.ts
@@ -39,7 +39,10 @@ function stateReducer(state: MultifactorAuthenticationState, action: Action): Mu
         case 'SET_SOFT_PROMPT_APPROVED':
             return {...state, softPromptApproved: action.payload};
         case 'SET_SCENARIO':
-            return {...state, scenario: action.payload};
+            if (action.payload === undefined) {
+                return {...state, scenarioName: undefined, scenario: undefined};
+            }
+            return {...state, scenarioName: action.payload.scenarioName, scenario: action.payload.scenario};
         case 'SET_PAYLOAD':
             return {...state, payload: action.payload};
         case 'SET_REGISTRATION_COMPLETE':
diff --git a/src/components/MultifactorAuthentication/Context/types.ts b/src/components/MultifactorAuthentication/Context/types.ts
index fcc5b17157c1..21efc451a946 100644
--- a/src/components/MultifactorAuthentication/Context/types.ts
+++ b/src/components/MultifactorAuthentication/Context/types.ts
@@ -1,7 +1,7 @@
+import type {MultifactorAuthenticationInitPayload, MultifactorAuthenticationSetScenarioPayload} from '@components/MultifactorAuthentication/config/scenarios/types';
 import type {
     MultifactorAuthenticationScenario,
     MultifactorAuthenticationScenarioAdditionalParams,
-    MultifactorAuthenticationScenarioConfig,
     MultifactorAuthenticationScenarioResponse,
 } from '@components/MultifactorAuthentication/config/types';
 import type {AuthenticationChallenge, RegistrationChallenge} from '@libs/MultifactorAuthentication/shared/challengeTypes';
@@ -9,11 +9,6 @@ import type {MFAError} from '@libs/MultifactorAuthentication/shared/MFAResult';
 import type {AuthTypeInfo} from '@libs/MultifactorAuthentication/shared/types';
 import type {MultifactorAuthenticationState} from './state';
 
-type InitPayload = {
-    scenario: MultifactorAuthenticationScenario;
-    payload: MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined;
-};
-
 type Action =
     | {type: 'SET_ERROR'; payload: MFAError | undefined}
     | {type: 'CLEAR_CONTINUABLE_ERROR'}
@@ -21,7 +16,7 @@ type Action =
     | {type: 'SET_REGISTRATION_CHALLENGE'; payload: RegistrationChallenge | undefined}
     | {type: 'SET_AUTHORIZATION_CHALLENGE'; payload: AuthenticationChallenge | undefined}
     | {type: 'SET_SOFT_PROMPT_APPROVED'; payload: boolean}
-    | {type: 'SET_SCENARIO'; payload: MultifactorAuthenticationScenarioConfig | undefined}
+    | {type: 'SET_SCENARIO'; payload: MultifactorAuthenticationSetScenarioPayload | undefined}
     | {type: 'SET_PAYLOAD'; payload: MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined}
     | {type: 'SET_REGISTRATION_COMPLETE'; payload: boolean}
     | {type: 'SET_AUTHORIZATION_COMPLETE'; payload: boolean}
@@ -29,7 +24,7 @@ type Action =
     | {type: 'SET_AUTHENTICATION_METHOD'; payload: AuthTypeInfo | undefined}
     | {type: 'SET_SCENARIO_RESPONSE'; payload: MultifactorAuthenticationScenarioResponse | undefined}
     | {type: 'SET_CANCEL_CONFIRM_VISIBLE'; payload: boolean}
-    | {type: 'INIT'; payload: InitPayload}
+    | {type: 'INIT'; payload: MultifactorAuthenticationInitPayload}
     | {type: 'REREGISTER'}
     | {type: 'CLOSE_MODAL'}
     | {type: 'RESET'};
diff --git a/src/components/MultifactorAuthentication/config/index.ts b/src/components/MultifactorAuthentication/config/index.ts
index 5a02d5db6403..d3bd9d4da8f5 100644
--- a/src/components/MultifactorAuthentication/config/index.ts
+++ b/src/components/MultifactorAuthentication/config/index.ts
@@ -6,4 +6,10 @@ import MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG from './scenarios';
 export {MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG};
 export {getScenarioConfig} from './scenarios';
 export {default as MULTIFACTOR_AUTHENTICATION_PROMPT_UI} from './scenarios/prompts';
-export type {Payloads as MultifactorAuthenticationScenarioPayload} from './scenarios';
+export type {
+    MultifactorAuthenticationInitEvent,
+    MultifactorAuthenticationInitPayload,
+    MultifactorAuthenticationScenarioConfigFor,
+    MultifactorAuthenticationSetScenarioPayload,
+    Payloads as MultifactorAuthenticationScenarioPayload,
+} from './scenarios/types';
diff --git a/src/components/MultifactorAuthentication/config/scenarios/index.ts b/src/components/MultifactorAuthentication/config/scenarios/index.ts
index d414624fde3f..0d235f02397d 100644
--- a/src/components/MultifactorAuthentication/config/scenarios/index.ts
+++ b/src/components/MultifactorAuthentication/config/scenarios/index.ts
@@ -1,35 +1,13 @@
-import type {EmptyObject} from 'type-fest';
-import type {
-    MultifactorAuthenticationScenario,
-    MultifactorAuthenticationScenarioConfig,
-    MultifactorAuthenticationScenarioConfigRecord,
-} from '@components/MultifactorAuthentication/config/types';
+import type {MultifactorAuthenticationScenario, MultifactorAuthenticationScenarioConfigRecord} from '@components/MultifactorAuthentication/config/types';
 import CONST from '@src/CONST';
-import type {Payload as AuthorizeTransactionPayload} from './AuthorizeTransaction';
 import AuthorizeTransaction from './AuthorizeTransaction';
 import BiometricsTest from './BiometricsTest';
-import type {Payload as ChangePINPayload} from './ChangePIN';
 import ChangePIN from './ChangePIN';
 import customConfig from './DefaultUserInterface';
-import type {Payload as RevealCardDetailsPayload} from './RevealCardDetails';
 import RevealCardDetails from './RevealCardDetails';
-import type {Payload as RevealPINPayload} from './RevealPIN';
 import RevealPIN from './RevealPIN';
-import type {Payload as SetPINOrderCardPayload} from './SetPINOrderCard';
 import SetPINOrderCard from './SetPINOrderCard';
-
-/**
- * Payload types for multifactor authentication scenarios.
- * Each scenario that requires additional parameters should have its payload type defined here.
- */
-type Payloads = {
-    [CONST.MULTIFACTOR_AUTHENTICATION.SCENARIO.BIOMETRICS_TEST]: EmptyObject;
-    [CONST.MULTIFACTOR_AUTHENTICATION.SCENARIO.SET_PIN_ORDER_CARD]: SetPINOrderCardPayload;
-    [CONST.MULTIFACTOR_AUTHENTICATION.SCENARIO.AUTHORIZE_TRANSACTION]: AuthorizeTransactionPayload;
-    [CONST.MULTIFACTOR_AUTHENTICATION.SCENARIO.REVEAL_PIN]: RevealPINPayload;
-    [CONST.MULTIFACTOR_AUTHENTICATION.SCENARIO.CHANGE_PIN]: ChangePINPayload;
-    [CONST.MULTIFACTOR_AUTHENTICATION.SCENARIO.REVEAL_CARD_DETAILS]: RevealCardDetailsPayload;
-};
+import type {MultifactorAuthenticationScenarioConfigFor} from './types';
 
 /**
  * Configuration records for all multifactor authentication scenarios.
@@ -44,16 +22,12 @@ const MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG = {
 } as const satisfies MultifactorAuthenticationScenarioConfigRecord;
 
 /**
- * Every entry of the config record satisfies MultifactorAuthenticationScenarioConfig at definition, but the
- * per-scenario action signatures make the record's value union non-narrowable, so the lookup needs
- * an assertion. This accessor owns the single assertion so callers don't repeat it; params are
- * type-guarded separately by ExecuteScenarioParams<T>.
+ * Returns the scenario config narrowed to the requested scenario key. Call sites that store the
+ * result alongside a dynamic scenario name should type that field as ValueOf<typeof MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG>.
  */
-function getScenarioConfig(scenarioName: MultifactorAuthenticationScenario): MultifactorAuthenticationScenarioConfig {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
-    return MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG[scenarioName] as MultifactorAuthenticationScenarioConfig;
+function getScenarioConfig<T extends MultifactorAuthenticationScenario>(scenarioName: T): MultifactorAuthenticationScenarioConfigFor<T> {
+    return MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG[scenarioName];
 }
 
 export default MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG;
 export {getScenarioConfig};
-export type {Payloads};
diff --git a/src/components/MultifactorAuthentication/config/scenarios/types.ts b/src/components/MultifactorAuthentication/config/scenarios/types.ts
new file mode 100644
index 000000000000..958735a4ddf3
--- /dev/null
+++ b/src/components/MultifactorAuthentication/config/scenarios/types.ts
@@ -0,0 +1,58 @@
+import type {EmptyObject} from 'type-fest';
+import type {MultifactorAuthenticationScenario, MultifactorAuthenticationScenarioParams} from '@components/MultifactorAuthentication/config/types';
+import type WithExactKeys from '@src/types/utils/WithExactKeys';
+import type {Payload as AuthorizeTransactionPayload} from './AuthorizeTransaction';
+import type {Payload as ChangePINPayload} from './ChangePIN';
+import type MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG from './index';
+import type {SCENARIO_NAMES} from './names';
+import type {Payload as RevealCardDetailsPayload} from './RevealCardDetails';
+import type {Payload as RevealPINPayload} from './RevealPIN';
+import type {Payload as SetPINOrderCardPayload} from './SetPINOrderCard';
+
+/**
+ * Per-scenario payload types. Must stay in lockstep with MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG keys.
+ */
+type ScenarioPayloadMap = {
+    [SCENARIO_NAMES.BIOMETRICS_TEST]: EmptyObject;
+    [SCENARIO_NAMES.SET_PIN_ORDER_CARD]: SetPINOrderCardPayload;
+    [SCENARIO_NAMES.AUTHORIZE_TRANSACTION]: AuthorizeTransactionPayload;
+    [SCENARIO_NAMES.REVEAL_PIN]: RevealPINPayload;
+    [SCENARIO_NAMES.CHANGE_PIN]: ChangePINPayload;
+    [SCENARIO_NAMES.REVEAL_CARD_DETAILS]: RevealCardDetailsPayload;
+};
+
+/** Payload types for multifactor authentication scenarios. */
+type Payloads = WithExactKeys<ScenarioPayloadMap, Record<MultifactorAuthenticationScenario, unknown>>;
+
+/** Resolved config for a specific scenario key (preserves per-scenario action payload types). */
+type MultifactorAuthenticationScenarioConfigFor<T extends MultifactorAuthenticationScenario> = (typeof MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG)[T];
+
+/** INIT payload with scenario name correlated to its params. */
+type MultifactorAuthenticationInitPayload = {
+    [K in MultifactorAuthenticationScenario]: {
+        scenario: K;
+        payload: MultifactorAuthenticationScenarioParams<K> | undefined;
+    };
+}[MultifactorAuthenticationScenario];
+
+/** SET_SCENARIO payload with scenario name correlated to its config. */
+type MultifactorAuthenticationSetScenarioPayload = {
+    [K in MultifactorAuthenticationScenario]: {
+        scenarioName: K;
+        scenario: MultifactorAuthenticationScenarioConfigFor<K>;
+    };
+}[MultifactorAuthenticationScenario];
+
+/**
+ * INIT machine event with scenario name, config, and payload correlated via generic T.
+ * A mapped union cannot be built from generic inputs without assertion; this form preserves
+ * correlation at call sites while remaining assignable to the default (all-scenarios) event type.
+ */
+type MultifactorAuthenticationInitEvent<T extends MultifactorAuthenticationScenario = MultifactorAuthenticationScenario> = {
+    type: 'INIT';
+    scenarioName: T;
+    scenario: MultifactorAuthenticationScenarioConfigFor<T>;
+    payload: MultifactorAuthenticationScenarioParams<T> | undefined;
+};
+
+export type {MultifactorAuthenticationInitEvent, MultifactorAuthenticationInitPayload, MultifactorAuthenticationScenarioConfigFor, MultifactorAuthenticationSetScenarioPayload, Payloads};
diff --git a/src/components/MultifactorAuthentication/machine/types.ts b/src/components/MultifactorAuthentication/machine/types.ts
index 4f947d1e1682..a32a25b7064b 100644
--- a/src/components/MultifactorAuthentication/machine/types.ts
+++ b/src/components/MultifactorAuthentication/machine/types.ts
@@ -1,10 +1,4 @@
-import type {
-    MultifactorAuthenticationScenario,
-    MultifactorAuthenticationScenarioAdditionalParams,
-    MultifactorAuthenticationScenarioConfig,
-} from '@components/MultifactorAuthentication/config/types';
-
-type ScenarioPayload = MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined;
+import type {MultifactorAuthenticationInitEvent} from '@components/MultifactorAuthentication/config/scenarios/types';
 
 /**
  * The modal lifecycle as seen by the view layer, derived 1:1 from the machine's top-level state:
@@ -20,9 +14,6 @@ type MfaModalPhase = 'open' | 'closing' | 'closed';
  * CLOSE_MODAL requests the close (flow -> `closing`); MODAL_CLOSED is the navigator's notification
  * that the close animation fully finished (`closing` -> `idle`, which wipes the context).
  */
-type MfaEvent =
-    | {type: 'INIT'; scenarioName: MultifactorAuthenticationScenario; scenario: MultifactorAuthenticationScenarioConfig; payload: ScenarioPayload}
-    | {type: 'CLOSE_MODAL'}
-    | {type: 'MODAL_CLOSED'};
+type MfaEvent = MultifactorAuthenticationInitEvent | {type: 'CLOSE_MODAL'} | {type: 'MODAL_CLOSED'};
 
 export type {MfaEvent, MfaModalPhase};
diff --git a/src/types/utils/WithExactKeys.ts b/src/types/utils/WithExactKeys.ts
new file mode 100644
index 000000000000..30a475731d57
--- /dev/null
+++ b/src/types/utils/WithExactKeys.ts
@@ -0,0 +1,9 @@
+import type {IsEqual} from 'type-fest';
+
+/** Resolves to true when left and right have exactly the same keys; false otherwise. */
+type AssertExactKeys<Left, Right> = IsEqual<keyof Left, keyof Right>;
+
+type WithExactKeys<Left, Right> = IsEqual<keyof Left, keyof Right> extends true ? Left : never;
+
+export type {AssertExactKeys};
+export default WithExactKeys;


```

</details>

English summary of changes:

- Added `getScenarioConfig<T>()` with a generic return type so per-scenario action signatures stay narrow without an eslint-suppressed type assertion.
- Introduced `scenarios/types.ts` for correlated types and Payloads, with `AssertExactKeys` / `WithExactKeys` so payload keys stay aligned with `MultifactorAuthenticationScenario`.
- Replaced widened `MultifactorAuthenticationScenarioConfig` in runtime state with `ValueOf<typeof MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG>` and the correlated payload/event types above.
- Made InitEvent generic so executeScenario can `send({type: 'INIT', ...})` without assertion; `InitPayload` / `SetScenarioPayload` remain discriminated unions for the reducer.

src/components/MultifactorAuthentication/machine/mfaMachine.ts
```diff
@@ -0,0 +1,111 @@
...
+ * `open` children, per-screen work as child states of its screen, and events shared by every screen
+ * (e.g. SET_ERROR) on `open` itself.
+ *
+ * No state is `final`: one long-lived actor serves every MFA flow (a top-level final state would
+ * stop it).
+ */
+const mfaMachine = setup({
+    // `{} as T` inside setup({types}) is XState v5's documented typing idiom (the values are erased
+    // at runtime and only carry types); there is no assertion-free way to express it.
+    /* eslint-disable @typescript-eslint/no-unsafe-type-assertion */
```
roryabraham, line 29: I'm ok with this one for the reason listed in the comment. 👍🏼 

Does `{} as unknown as MultifactorAuthenticationState` resolve the error without an assertion?

src/components/MultifactorAuthentication/machine/mfaMachine.ts
```diff
@@ -0,0 +1,111 @@
...
+    types: {
+        context: {} as MultifactorAuthenticationState,
+        events: {} as MfaEvent,
+    },
+    /* eslint-enable @typescript-eslint/no-unsafe-type-assertion */
+    actions: {
+        initFromEvent: assign(({event}) => {
```
roryabraham: Comments to contextualize each action would be helpful

src/components/MultifactorAuthentication/machine/mfaMachine.ts
```diff
@@ -0,0 +1,111 @@
...
+        resetNavigationBuffer: () => resetMfaNavigation(),
+    },
+    delays: {
+        // How long `closing` waits for MODAL_CLOSED before re-entering `idle` on its own; longer
+        // than any close animation can take.
+        closeFallback: CONST.MAX_TRANSITION_START_WAIT_MS + CONST.MAX_TRANSITION_DURATION_MS + CONST.ANIMATED_TRANSITION,
+    },
+}).createMachine({
+    id: 'mfa',
+    initial: 'idle',
```
roryabraham: One type safety gap I noticed here is that we can change the initial state to anything we want, and it doesn't result in a type error. To plug this gap, we should defined an `MFA_STATE` constant, enumerating all the states as strings:

<details>
<summary>diff</summary>

```diff
diff --git a/src/components/MultifactorAuthentication/machine/mfaMachine.ts b/src/components/MultifactorAuthentication/machine/mfaMachine.ts
index d00b531c8ea4..fb4b5bba7957 100644
--- a/src/components/MultifactorAuthentication/machine/mfaMachine.ts
+++ b/src/components/MultifactorAuthentication/machine/mfaMachine.ts
@@ -5,6 +5,7 @@ import {navigate as mfaNavigate, resetMfaNavigation} from '@components/Multifact
 import Navigation from '@libs/Navigation/Navigation';
 import CONST from '@src/CONST';
 import SCREENS from '@src/SCREENS';
+import MFA_STATES from './states';
 import type {MfaEvent} from './types';
 
 const DEFAULT_CONTEXT: MultifactorAuthenticationState = DEFAULT_STATE;
@@ -59,34 +60,34 @@ const mfaMachine = setup({
     },
 }).createMachine({
     id: 'mfa',
-    initial: 'idle',
+    initial: MFA_STATES.IDLE,
     context: DEFAULT_CONTEXT,
     states: {
-        idle: {
+        [MFA_STATES.IDLE]: {
             // The wipe runs on every (re)entry so no flow data (validate code, challenges, scenario
             // response) outlives the modal.
             entry: ['resetContext', 'resetNavigationBuffer'],
             on: {
                 // Accepted only here: an INIT sent while the modal is open or still closing is
                 // dropped rather than started on dirty state.
-                INIT: {target: 'open', actions: 'initFromEvent'},
+                INIT: {target: MFA_STATES.OPEN.name, actions: 'initFromEvent'},
             },
         },
-        open: {
-            initial: 'preparing',
+        [MFA_STATES.OPEN.name]: {
+            initial: MFA_STATES.OPEN.PREPARING,
             on: {
-                CLOSE_MODAL: {target: 'closing', actions: 'hideCancelConfirm'},
+                CLOSE_MODAL: {target: MFA_STATES.CLOSING, actions: 'hideCancelConfirm'},
             },
             states: {
                 // Transparent initial screen. There is no pre-screen work yet, so it falls straight through;
                 // later slices replace `always` with child states (device check, registration decision, ...).
-                preparing: {
-                    always: 'outcome',
+                [MFA_STATES.OPEN.PREPARING]: {
+                    always: MFA_STATES.OPEN.OUTCOME.name,
                 },
-                outcome: {
-                    initial: 'success',
+                [MFA_STATES.OPEN.OUTCOME.name]: {
+                    initial: MFA_STATES.OPEN.OUTCOME.SUCCESS,
                     states: {
-                        success: {
+                        [MFA_STATES.OPEN.OUTCOME.SUCCESS]: {
                             entry: ['markFlowComplete', 'navigateToSuccessOutcome'],
                         },
                     },
@@ -97,12 +98,12 @@ const mfaMachine = setup({
         // stays visible while it slides out. The navigator sends MODAL_CLOSED once the close
         // animation finishes; if it unmounts before that, the event never comes and the
         // `closeFallback` timer re-enters `idle` instead.
-        closing: {
+        [MFA_STATES.CLOSING]: {
             on: {
-                MODAL_CLOSED: 'idle',
+                MODAL_CLOSED: MFA_STATES.IDLE,
             },
             after: {
-                closeFallback: {target: 'idle'},
+                closeFallback: {target: MFA_STATES.IDLE},
             },
         },
     },
diff --git a/src/components/MultifactorAuthentication/machine/snapshotToState.ts b/src/components/MultifactorAuthentication/machine/snapshotToState.ts
index 1d08b1c9bcde..8000d0c22d1a 100644
--- a/src/components/MultifactorAuthentication/machine/snapshotToState.ts
+++ b/src/components/MultifactorAuthentication/machine/snapshotToState.ts
@@ -1,6 +1,7 @@
 import type {SnapshotFrom} from 'xstate';
 import type {MultifactorAuthenticationState} from '@components/MultifactorAuthentication/Context/state';
 import type mfaMachine from './mfaMachine';
+import MFA_STATES, {MFA_MODAL_PHASE} from './states';
 import type {MfaModalPhase} from './types';
 
 type MfaSnapshot = SnapshotFrom<typeof mfaMachine>;
@@ -9,10 +10,10 @@ type MfaSnapshot = SnapshotFrom<typeof mfaMachine>;
 type MfaState = MultifactorAuthenticationState & {modalPhase: MfaModalPhase};
 
 function getModalPhase(snapshot: MfaSnapshot): MfaModalPhase {
-    if (snapshot.matches('open')) {
-        return 'open';
+    if (snapshot.matches(MFA_STATES.OPEN.name)) {
+        return MFA_MODAL_PHASE.OPEN;
     }
-    return snapshot.matches('closing') ? 'closing' : 'closed';
+    return snapshot.matches(MFA_STATES.CLOSING) ? MFA_MODAL_PHASE.CLOSING : MFA_MODAL_PHASE.CLOSED;
 }
 
 /**
diff --git a/src/components/MultifactorAuthentication/machine/states.ts b/src/components/MultifactorAuthentication/machine/states.ts
new file mode 100644
index 000000000000..71f5f9c5e567
--- /dev/null
+++ b/src/components/MultifactorAuthentication/machine/states.ts
@@ -0,0 +1,28 @@
+function makeState<const N extends string, const T extends Record<string, string | {name: string}>>(name: N, children: T): {name: N} & T {
+    return {name, ...children};
+}
+
+/** XState state node IDs for the MFA machine, nested to mirror the chart hierarchy. */
+const MFA_STATES = {
+    IDLE: 'idle',
+    CLOSING: 'closing',
+    OPEN: makeState('open', {
+        PREPARING: 'preparing',
+        OUTCOME: makeState('outcome', {
+            SUCCESS: 'success',
+        }),
+    }),
+} as const;
+
+/**
+ * Modal lifecycle as seen by the view layer. `OPEN` and `CLOSING` match {@link MFA_STATES}; `CLOSED` is
+ * the view-layer name for {@link MFA_STATES.IDLE} (`idle`): modal unmounted, actor ready for INIT.
+ */
+const MFA_MODAL_PHASE = {
+    OPEN: MFA_STATES.OPEN.name,
+    CLOSING: MFA_STATES.CLOSING,
+    CLOSED: 'closed',
+} as const;
+
+export default MFA_STATES;
+export {makeState, MFA_MODAL_PHASE};
diff --git a/src/components/MultifactorAuthentication/machine/types.ts b/src/components/MultifactorAuthentication/machine/types.ts
index 4f947d1e1682..4b11f726153b 100644
--- a/src/components/MultifactorAuthentication/machine/types.ts
+++ b/src/components/MultifactorAuthentication/machine/types.ts
@@ -1,16 +1,8 @@
-import type {
-    MultifactorAuthenticationScenario,
-    MultifactorAuthenticationScenarioAdditionalParams,
-    MultifactorAuthenticationScenarioConfig,
-} from '@components/MultifactorAuthentication/config/types';
+import type {ValueOf} from 'type-fest';
+import type {MultifactorAuthenticationInitEvent} from '@components/MultifactorAuthentication/config/scenarios/types';
+import type {MFA_MODAL_PHASE} from './states';
 
-type ScenarioPayload = MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined;
-
-/**
- * The modal lifecycle as seen by the view layer, derived 1:1 from the machine's top-level state:
- * `open` -> 'open', `closing` -> 'closing', `idle` -> 'closed'.
- */
-type MfaModalPhase = 'open' | 'closing' | 'closed';
+type MfaModalPhase = ValueOf<typeof MFA_MODAL_PHASE>;
 
 /**
  * Events accepted by the machine. So far only the three that drive the
@@ -20,9 +12,6 @@ type MfaModalPhase = 'open' | 'closing' | 'closed';
  * CLOSE_MODAL requests the close (flow -> `closing`); MODAL_CLOSED is the navigator's notification
  * that the close animation fully finished (`closing` -> `idle`, which wipes the context).
  */
-type MfaEvent =
-    | {type: 'INIT'; scenarioName: MultifactorAuthenticationScenario; scenario: MultifactorAuthenticationScenarioConfig; payload: ScenarioPayload}
-    | {type: 'CLOSE_MODAL'}
-    | {type: 'MODAL_CLOSED'};
+type MfaEvent = MultifactorAuthenticationInitEvent | {type: 'CLOSE_MODAL'} | {type: 'MODAL_CLOSED'};
 
 export type {MfaEvent, MfaModalPhase};
```

</summary>

src/components/MultifactorAuthentication/machine/types.ts
```diff
@@ -0,0 +1,28 @@
...
+    MultifactorAuthenticationScenarioConfig,
+} from '@components/MultifactorAuthentication/config/types';
+
+type ScenarioPayload = MultifactorAuthenticationScenarioAdditionalParams<MultifactorAuthenticationScenario> | undefined;
+
+/**
+ * The modal lifecycle as seen by the view layer, derived 1:1 from the machine's top-level state:
+ * `open` -> 'open', `closing` -> 'closing', `idle` -> 'closed'.
+ */
+type MfaModalPhase = 'open' | 'closing' | 'closed';
```
roryabraham: If you do one of the refactors I did above, you could simplify to reference the machine's top-level state directly. That way you don't have a separate `'closed'` state. It's just `idle`, coming from the state machine.

roryabraham (changes requested):

config/eslint/eslint.seatbelt.tsv
dariusz-biela: Thanks for the info. Starting with the next release, I'll stop manually deleting the record from that file.

src/components/MultifactorAuthentication/machine/mfaMachine.ts
dariusz-biela: Unfortunately, that doesn't solve the problem.

The current implementation is the cleanest option I've been able to find.

src/components/MultifactorAuthentication/config/scenarios/index.ts
dariusz-biela: I did something similar to what you suggested, but using the utils `AssertTypesEqual`.

I also extended the type safety to the `executeScenario` payload because it wasn't fully secured.

src/components/MultifactorAuthentication/machine/mfaMachine.ts
dariusz-biela: I added the flat config because:
- we don't have to worry about whether to use “.name” at a given level
- we don't have to create additional utilities
- when changing the machine's structure, we don't have to modify the layout in `const`
- the hierarchy in `const` doesn't offer significantly more benefits compared to this implementation. 

```ts
MFA_STATE = {
    IDLE: 'idle',
    OPEN: 'open',
    CLOSING: 'closing',
    PREPARING: 'preparing',
    OUTCOME: 'outcome',
    SUCCESS: 'success',
} 
```

src/components/MultifactorAuthentication/machine/types.ts
dariusz-biela: I changed “idle” to “closed” because it's a better name.

src/components/MultifactorAuthentication/components/OutcomeScreen/OutcomeScreenBase.tsx
dariusz-biela: This is not an external consumer because it is contained within the MultifactorAuthentication module and is used within the MultifactorAuthentication flow. 

My main goal in separating these contexts was to narrow down the public interface, which can consist solely of `executeScenario` calls to the data and actions used by the MFA screens. 

This allows us to optimize the number of re-renders and give developers a better understanding of what they can use. 

I think the final shape of the contexts may differ significantly from what we have right now, as I will be refining their structure as we add new features.

src/components/MultifactorAuthentication/config/scenarios/index.ts
```diff
@@ -39,5 +21,27 @@ const Configs = {
...
+ * Resolved config for a single scenario key, preserving that scenario's specific action signature.
+ * This is the precise type behind {@link getScenarioConfig}'s return.
+ */
+type MultifactorAuthenticationScenarioConfigFor<T extends MultifactorAuthenticationScenario> = (typeof MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG)[T];
+
+/**
+ * Union of every resolved scenario config - the value type to store when the scenario is dynamic (not
+ * known at the type level), e.g. the `scenario` field on the runtime MFA state.
+ */
+type MultifactorAuthenticationResolvedScenarioConfig = ValueOf<typeof MULTIFACTOR_AUTHENTICATION_SCENARIO_CONFIG>;
```
roryabraham: NAB: I think it's clearer to just use `ValueOf` directly at the callsites than to create this "Resolved" type alias.

src/libs/MultifactorAuthentication/shared/VALUES.ts
```diff
@@ -214,6 +214,24 @@ const ANOMALOUS_FAILURES = new Set<ReasonValue>([
     REASON.LOCAL_ERRORS.HSM.NO_MATCHING_LOCAL_CREDENTIAL,
 ]);
 
+/**
+ * XState node IDs for the MFA machine (src/components/MultifactorAuthentication/machine). Flat: the
```
roryabraham: > Flat: the hierarchy is already visible in the machine's `states` tree, so nesting here would only duplicate it. Referencing these instead of bare strings de-duplicates the literals and guards reference typos.

Remove this - comments should not reference abandoned designs or code that's not there.

roryabraham (approved):

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationExternalApiContext.ts
```diff
@@ -0,0 +1,41 @@
...
+    Partial<MultifactorAuthenticationScenarioParams<T>> extends MultifactorAuthenticationScenarioParams<T>
+        ? [params?: MultifactorAuthenticationScenarioParams<T>]
+        : [params: MultifactorAuthenticationScenarioParams<T>];
+
+/**
+ * The public API for starting an MFA flow from app code (settings pages, test tools, 3DS entry
+ * points). Everything that happens after INIT is driven by the machine and exposed to flow-internal
+ * screens through MultifactorAuthenticationInternalApiContext.
+ */
+type MultifactorAuthenticationExternalApi = {
```
DylanDylann:
```suggestion
type MultifactorAuthenticationExternalAPI = {
```

please update in all places

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationMainContext.tsx
```diff
@@ -82,117 +58,63 @@ function MultifactorAuthenticationContextProvider({children}: MultifactorAuthent
...
-
-    if (!context) {
-        throw new Error('useMultifactorAuthentication must be used within a MultifactorAuthenticationContextProviders');
-    }
-
-    return context;
+    return (
+        <MultifactorAuthenticationExternalApiContext.Provider value={externalApi}>
+            <MultifactorAuthenticationInternalApiContext.Provider value={internalApi}>{children}</MultifactorAuthenticationInternalApiContext.Provider>
+        </MultifactorAuthenticationExternalApiContext.Provider>
```
DylanDylann: Do we still need MultifactorAuthenticationActionsContext and MultifactorAuthenticationStateContext? As I understand it, we'll be relying on the state machine's state. Is there a reason not to remove them?

src/components/MultifactorAuthentication/Context/state.ts
DylanDylann: I agree with @jnowakow on this point, the flow should expose a single atomic status rather than a set of independent boolean flags. Using isRegistrationComplete, isAuthorizationComplete, and isFlowComplete together to represent the status of flow is exactly the "boolean grab-bag" the XState migration was meant to eliminate

src/components/MultifactorAuthentication/components/OutcomeScreen/OutcomeScreenBase.tsx
```diff
@@ -62,13 +62,9 @@ function OutcomeScreenBase({headerTitle, illustration, iconWidth, iconHeight, ti
...
 
-    const onClose =
-        onCloseOverride ??
-        (() => {
-            dispatch({type: 'CLOSE_MODAL'});
-        });
+    const onClose = onCloseOverride ?? closeModal;
```
DylanDylann, line 67: I would expect that integrating xState can help remove `onCloseOverride`

src/components/MultifactorAuthentication/machine/mfaMachine.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * MFA state machine. The top level models the modal lifecycle (`closed` -> `open` -> `closing`); the
+ * child states of `open` map 1:1 to the screen the user currently sees. Later slices add screens as
+ * `open` children, per-screen work as child states of its screen, and events shared by every screen
+ * (e.g. SET_ERROR) on `open` itself.
+ *
+ * No state is `final`: one long-lived actor serves every MFA flow (a top-level final state would
+ * stop it).
+ */
+const mfaMachine = setup({
```
DylanDylann:
```suggestion
const MFAMachine = setup({
```

DylanDylann: Please fix all these typos

src/components/MultifactorAuthentication/useSyncMfaModalNavigatorWithHistory/index.ts
```diff
@@ -31,10 +34,13 @@ function getHistory(): readonly unknown[] {
  * so re-subscribing on `requestCancel` change does not toggle the history
  * marker off/on.
  *
- * @param isModalOpen Whether the MFA modal is currently mounted.
+ * @param modalState Current modal lifecycle state; the marker lives only while it is `open`.
  * @param requestCancel Called on every back press.
  */
-function useSyncMfaModalNavigatorWithHistory(isModalOpen: boolean, requestCancel: () => void): void {
+function useSyncMfaModalNavigatorWithHistory(modalState: MfaModalState, requestCancel: () => void): void {
+    // The marker tracks `open` only: `closing` already pops it so back is released while the close animation plays
```
DylanDylann: Should this comment be above the useEffect?

src/libs/Navigation/AppNavigator/Navigators/MultifactorAuthenticationModalNavigator.tsx
```diff
@@ -108,30 +99,34 @@ function MultifactorAuthenticationModalNavigator() {
...
             return;
         }
         if (mfaNavigationRef.isReady() && mfaNavigationRef.canGoBack()) {
             mfaNavigationRef.goBack();
         }
-        backdropProgress.set(withTiming(0, {duration: CONST.ANIMATED_TRANSITION}));
+        // Fade out in lockstep with the screen's slide-out - same duration and easing as the close
+        // spec of Animations.SLIDE_FROM_RIGHT (Animated.timing defaults to inOut(ease)). The navigator
+        // unmounts on transitionEnd, so any fade longer than the slide gets cut mid-animation.
+        backdropProgress.set(withTiming(0, {duration: CONST.MODAL.ANIMATION_TIMING.RHP_DURATION_OUT_WEB, easing: Easing.inOut(Easing.ease)}));
```
DylanDylann, line 115:
```suggestion
        backdropProgress.set(withTiming(0, {duration: CONST.MODAL.ANIMATION_TIMING.RHP_DURATION_OUT_WEB, easing: Easing.inOut(Easing.ease)}));
        
```

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationMainContext.tsx
```diff
@@ -82,117 +58,63 @@ function MultifactorAuthenticationContextProvider({children}: MultifactorAuthent
...
-        // The reducer's INIT case is the authoritative double-tap guard; this short-circuit only
-        // skips a redundant captureCredentialsState() native call + breadcrumb on the happy path.
+    const executeScenario = async <T extends MultifactorAuthenticationScenario>(scenarioName: T, ...args: MultifactorAuthenticationExecuteScenarioArgs<T>): Promise<void> => {
+        const [params] = args;
+
+        // Perf short-circuit: once a scenario is active the machine ignores INIT, so skip the redundant
+        // captureCredentialsState() native call + breadcrumb on the happy path.
         if (state.scenario) {
             return;
         }
```
DylanDylann: Per the comment above, checking `modalState` directly makes more sense to me

```
if (state.modalState !== MFA_STATE.CLOSED) {
    return;
}

```

src/libs/Navigation/AppNavigator/Navigators/MultifactorAuthenticationModalNavigator.tsx
DylanDylann: Add an empty line for better visibility

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationMainContext.tsx
dariusz-biela: I’m keeping it solely because we still have screens (SoftPrompt, MagicCode, etc.) that are tied to the old context, and I don’t want to refactor them right now, as that would drastically increase the number of changes.

Overall, I want to phase out the old context incrementally as I replace its functionality with a state machine.

src/components/MultifactorAuthentication/config/scenarios/index.ts
dariusz-biela: In the end, I removed that type and, in places where it was needed, I used `MultifactorAuthenticationScenarioConfigFor<MultifactorAuthenticationScenario>`, because it allows us to import only the type without having to import a variable, as is the case with `ValueOf`.

src/components/MultifactorAuthentication/components/OutcomeScreen/OutcomeScreenBase.tsx
dariusz-biela: To be honest, I'm planning to leave this code as is for now. 

I know it's not top-notch, but I haven't come up with a solution that really would be really good yet.

I plan to create a separate PR at some point to improve it, because this component has quite a few dependencies and I need to make sure they all work properly.

src/components/MultifactorAuthentication/useSyncMfaModalNavigatorWithHistory/index.ts
dariusz-biela: No, the comment refers to the `isModalOpen` variable and explains why we are only interested in the “open” state, and why the “closing” state—even though the modal is still open—is not included in this state.

src/components/MultifactorAuthentication/Context/state.ts
dariusz-biela: Okay, I know what you mean, and there are a few solutions we could implement that might replace those Booleans, but I won't do that in this PR because it would trigger a cascade of changes. 

Now I'll reduce the machine context to the absolute minimum I use. 

`
type MfaContext = Pick<MultifactorAuthenticationState, 'error' | 'scenarioName' | 'scenario' | 'payload' | 'isCancelConfirmVisible'>;
`

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationMainContext.tsx
```diff
@@ -1,62 +1,43 @@
...
 
-    /** Dismiss the cancel-confirmation modal without cancelling the flow. */
-    hideCancelConfirm: () => void;
-
-    /** Confirm cancellation — hides the modal and runs cancel(). */
-    confirmCancel: () => void;
-};
-
-const MultifactorAuthenticationContext = createContext<MultifactorAuthenticationContextValue | undefined>(undefined);
+let deviceBiometricsState: OnyxEntry<DeviceBiometrics>;
```
roryabraham: This module-level state and the `Onyx.connectWithoutView` in a `useEffect` remains a serious code smell imo.

Furthermore, only the `hasAcceptedSoftPrompt` field of the `DeviceBiometrics` payload is being read here. This is what I'd prefer to see:

```ts
const [hasAcceptedSoftPrompt = false] = useOnyx(getDeviceBiometricsOnyxKey(accountID), {selector: (deviceBiometrics) => deviceBiometrics?.hasAcceptedSoftPrompt})
```

Except with the selector extracted to `src/selectors` so it's a single stable reference across consumers (to properly leverage the internal `useOnyx` snapshot cache).

---

If that doesn't work, I'd be keen to understand why, as that feels indicative of a deeper issue.

src/components/MultifactorAuthentication/useSyncMfaModalNavigatorWithHistory/index.ts
DylanDylann: I see your point. However, your explanation describes this code
```
        if (!isModalOpen) {
            return;
        }
```

If a comment is placed above const isModalOpen = modalState === MFA_STATE.OPEN, it should explain why we determine isModalOpen in this way

src/components/MultifactorAuthentication/Context/MultifactorAuthenticationMainContext.tsx
dariusz-biela: I've changed that. For now, it's not linked to any critical code, so I don't know if it will ultimately stay this way, but that's a problem for future PRs.

In general, we'll have to figure out the most sensible way to retrieve Onyx's state in the state machine.

src/components/MultifactorAuthentication/Context/state.ts
```diff
@@ -0,0 +1,82 @@
...
+    MultifactorAuthenticationScenarioResponse,
+} from '@components/MultifactorAuthentication/config/types';
+import type {AuthenticationChallenge, RegistrationChallenge} from '@libs/MultifactorAuthentication/shared/challengeTypes';
+import type {MFAError} from '@libs/MultifactorAuthentication/shared/MFAResult';
+import type {AuthTypeInfo} from '@libs/MultifactorAuthentication/shared/types';
+
+/**
+ * The MFA state shape, shared so the machine layer can build its context from it without importing
+ * the reducer. The Provider maps the machine snapshot back to this shape so consumers keep reading
+ * `state.X` unchanged.
```
chuckdries: I'm a little confused by this comment, I'm wondering if you can clear up the intent/future plans, or if we should make it more precise. Right now, the machine maps its snapshot to `MfaState` via `snapshotToState`, which is more narrow than this. The full `MultifactorAuthenticationState` shape comes from `stateReducer` which is stored independently in a `useReducer` in `MultifactorAuthenticationStateProvider`. Two separate sources of truth. Forgive me if you've said this somewhere else, but is the plan to remove the latter? Should we do so now?

dariusz-biela: Good call, the comment is misleading and I'll fix it. Let me clarify the intent.

Short answer to your two questions: yes, the reducer is meant to go away, but incrementally rather than in one step, and not in this PR.

This is a strangler-style migration. The machine context today is only the subset already wired in (`Pick<MultifactorAuthenticationState, 'error' | 'scenarioName' | 'scenario' | 'payload' | 'isCancelConfirmVisible'>`). The reducer still holds the full shape for the fields no slice has migrated yet (`validateCode`, the challenges, the completion flags, `authenticationMethod`, `scenarioResponse`, ...), and its remaining consumers (`ValidateCodePage`, `usePromptContent`, `AuthenticationMethodDescription`) keep reading those off it. Each upcoming slice moves one more field into the machine; once the reducer owns nothing, it's deleted. So the two stores aren't meant to compete - they're a moving boundary between "already migrated" and "not yet".

state.ts is the shared contract during that transition: the machine derives its context as a Pick of it (without importing the reducer) and the reducer uses it directly, so both layers agree on field names/types and I don't fork the shape.

Why keep the bridge (`snapshotToState` -> `MfaState` -> `state.X`) instead of reading the machine directly in components: I wanted to avoid landing two unrelated changes together (the data-structure move and a read-API change), and to defer the final machine read-interface decision until the chart is fully designed. The end state I'm aiming for is components reading via useSelector from `@xstate/react`, which retires snapshotToState; I'll do that as its own PR once the machine is complete. That's a separate, later step from removing the reducer.

Why not do it all now: removing the reducer today means pulling every remaining field into the machine in this PR, which is exactly the oversized change I'm trying to avoid per slice.

I'll rewrite the comment to say this explicitly - in particular that `snapshotToState` is the narrow `Pick` + `modalState` (not the full shape), and that the reducer is the transitional home for not-yet-migrated fields. And from the next slice on I'll document this kind of intent in the code.

dariusz-biela: ### Important

Follow-up: I rewrote the comment, and made the "two sources of truth" structural rather
than a convention - `MfaContext` and `MultifactorAuthenticationState` are now disjoint
types, so a field that moves to the machine is removed from the reducer shape and can no
longer be read off the reducer (compile error, not a silent `undefined`). That already
caught a live case: ChangePIN was still reading `payload` off the reducer after it had
moved to the machine. Overall plan is unchanged - the reducer still drains field-by-field,
and the `useSelector` read-API migration stays a separate later PR.

chuckdries: Nice catch! sounds good to me

chuckdries (approved):

roryabraham (approved):

roryabraham:
don't need all tests passing since base isn't main

Issue #89767, opened 2026-05-06
[Due for payment 2026-07-08] [Split #88320] PR 6 - Replace ReportActionsView with ReportActions orchestrator


rlinoz:
Goal: Introduce src/pages/inbox/ReportActions.tsx as the orchestrator and replace ReportActionsView.

Scope:
- Add src/pages/inbox/ReportActions.tsx
- Delete src/pages/inbox/report/ReportActionsView.tsx
- Delete obsolete getReportActionsListInitialNumToRender helper and test
- Update consumers to mount ReportActions
- Rewrite UI and perf tests against the new surface

Risk: Medium. This is the main orchestration swap.

Depends on: https://github.com/Expensify/App/issues/89766

Parent issue: https://github.com/Expensify/App/issues/88320
Source plan: https://github.com/Expensify/App/issues/88320#issuecomment-4282484785

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

DylanDylann:
Not overdue

LukasMod:
Hi, I am Łukasz from Callstack, an expert agency, and I'm starting work on this issue, as PR5 is already merged https://github.com/Expensify/App/issues/89766

LukasMod:
PR ready for review https://github.com/Expensify/App/pull/93694

LukasMod:
Still working on refactoring current solution after first round of feedback. I will also try to split it into smaller PRs if possibile

LukasMod:
Ready for second round of review

mallenexpensify:
### Payment Summary
Contributor+: @DylanDylann  due $250 via NewDot

Issue #88320, opened 2026-04-20
Decompose ReportActionsList


rlinoz:
Decomposing ReportActionList seems to improve re-rendering time and performs way less commits https://github.com/Expensify/App/pull/87506

This issue tracks the split of the linked PR into multiple ones.

adhorodyski:
# Split plan — `decompose/report-actions-list` → PR series

## Dependency graph

```
PR 1  Extract presentation (header + previous-messages button)
  ↓
PR 2  Extract data hooks (useTransactionThread + useReportActionsPagination)
  ↓
PR 3  Extract useReportActionsVisibility
  ↓
PR 4  Extract marker hooks (useUnreadMarker + useMarkAsRead)
  ↓
PR 5  Extract useReportActionsScroll
  ↓
PR 6  Replace ReportActionsView with ReportActions orchestrator
  ↓
PR 7  Drop FlashList hidden-render workaround
  ↓
PR 8  Inverted-wheel handler for InvertedFlashList
  ↓
PR 9  Scope VISIBLE_REPORT_ACTIONS selector per report
  ↓
PR 10 Compiler-clean pipeline hooks
  ↓
PR 11 Simplify unread marker suppression rule
  ↓
PR 12 Flip list-ref ownership on ActionListContext
```

## Deferred / rejected splits

- **Router shim as a standalone PR.** Originally planned as PR 1. Rejected: introducing a router before its data layer is reusable would force either heavy duplication against MRR's subscriptions or a speculative shared selector hook. Router lands in PR 6 as part of the orchestrator replacement, after hooks exist.
- **MRR migration to shared router.** Out of scope for this series. See memory: `project_report_actions_list_unification.md`.

---

## PR 1 — Extract presentation components

**Goal.** Pull `ReportActionsListHeader` and `ShowPreviousMessagesButton` out of the 915-LOC monolith at `src/pages/inbox/report/ReportActionsList.tsx`. Components own their own rendering gates (CLEAN-REACT-PATTERNS-2). Call-site behavior gates stay in place — the monolith gets smaller, nothing changes user-visible.

**Files.**

New:

- `src/pages/inbox/report/ShowPreviousMessagesButton.tsx`
  - Props: `reportAction`, `hasPreviousMessages`, `showFullHistory`, `onPress`
  - Internal gates: returns null when `actionName !== CREATED`, `!hasPreviousMessages`, or `showFullHistory`
  - Renders the button + divider-line presentation when gates pass
[...]

PR #91179, opened 2026-05-20
[NoQA] decompose ReportActionsList: 4

adhorodyski:
### Explanation of Change

PR 4 of the ReportActionsList decomposition series ([tracker #88320](https://github.com/Expensify/App/issues/88320), [blueprint PR #87506](https://github.com/Expensify/App/pull/87506)). PRs 1–3 merged: #88392, #89598, #91048.

Extracts the unread-marker scan logic and the mark-as-read trigger logic out of `ReportActionsList.tsx` into two new focused hooks.

**Pipeline state after PR4.** `ReportActionsList` now consumes: `useUnreadMarker`, `useMarkAsRead`, `useReportActionsNewActionLiveTail`, `useReportUnreadMessageScrollTracking`, `useReportScrollManager`, `useScrollToEndOnNewMessageReceived`, `useReportIsArchived`, `useShouldShowComposerForActiveEditDraft`, and others. Remaining: scroll hook (PR5), then PR6 collapses the View into a `ReportActions` orchestrator.

#### ⚠️ Userspace behavior change

Extracting `useUnreadMarker` into a standalone, React Compiler–compliant hook required removing the path-dependent `prevUnreadMarkerReportActionID` state (it read a ref during render, which the compiler rejects in a new file). Dropping it changes one user-visible edge case:

- **Before:** the "NEW" marker could land on **your own just-sent message** in this corner — a marker was already shown above someone else's message, that message then became read, and the only remaining unread message was one you just sent yourself. The marker would move onto your own message.
- **After:** the marker **never** lands on your own new or just-confirmed message. In that corner it simply clears.

This is the only behavior difference. Everything else (marker placement for others' messages, offline-received messages, pagination anchoring, mark-as-read triggers) is unchanged. The change applies to both the chat report list and the money-request report list, since they share `shouldDisplayNewMarkerOnReportAction`. This matches the behavior in blueprint PR #87506.

### Fixed Issues

$ https://github.com/Expensify/App/issues/89765
PROPOSAL:

### Tests

Two hooks (`useUnreadMarker`, `useMarkAsRead`) extracted from `ReportActionsList.tsx`. Aside from the unread-marker edge case documented above (marker no longer lands on your own just-sent message), behavior is unchanged. The plan exercises every behavior the extracted code owns plus the cross-report `prevReportID` (now module-level), and step L below verifies the edge case explicitly.

**A. Unread marker — basic scan**

1. Open a chat with 5+ unread messages from another user
[...]

PR #93694, opened 2026-06-16
Decompose ReportActionsList: 6

LukasMod:
### Explanation of Change

> part of the `ReportActionsList` decomposition series ([#88320](https://github.com/Expensify/App/issues/88320))

This splits the old `ReportActionsView` into three pieces: a route-only orchestrator that decides what to render, a skeleton guard that owns the data pipeline and gates rendering, and a hook-driven body that renders the chat list. There are no user-facing behavior changes. This is a structural refactor.

### Changes

- New `src/pages/inbox/ReportActions.tsx`: a route-only orchestrator. It owns the coarse branching: the early skeleton states, the choice between the money-request table and the chat list, and mounting `UserTypingEventListener`. The app-load skeleton is hoisted here so the body's hooks never run during app boot — except for Concierge, which is excluded so the body stays mounted under that skeleton and can seed its session, matching the old `ReportActionsView`.
- New `src/hooks/useReportActionsListModel.ts`: the single read-and-derive pipeline for the list. It runs the subscriptions and derivations once (sort, paginate, visibility, concierge) and returns three projections: `readinessSignals` (kept off the context, feeds the skeleton decision), `state` (the render payload), and `actions` (the command handles: load older, load newer, show previous messages).
- New `src/pages/inbox/report/ReportActionsSkeletonGuard.tsx`: owns `useReportActionsListModel`, makes the skeleton-versus-content decision (`computeReportActionsSkeletonState`), runs the effects that must fire while the skeleton shows (copy-selection, pending-concierge, concierge-session start, cold-open telemetry), and provides `state` and `actions` to the body. It is modeled on `ReportNotFoundGuard`: it returns early with the skeleton before the body mounts, so the body's hooks and effects never run during the skeleton phase.
- New `src/pages/inbox/report/ReportActionsListContext.tsx`: splits the pipeline output into two contexts. `ReportActionsListStateContext` carries the render state that changes often, and it is the only slice that drives list re-renders. `ReportActionsListActionsContext` carries the command handles, which keep a stable identity, so a consumer that reads only the handles does not re-render when new actions arrive. This follows the State/Actions split that `ConciergeSessionContext` already uses.
- `src/pages/inbox/report/ReportActionsList.tsx` (the body) is now hook-driven. Its props shrink to `{reportID, onLayout?}`. It reads the pipeline output from the two contexts and holds only the UI-close hooks (`useUnreadMarker`, `useMarkAsRead`, `useReportActionsScroll`). Those hooks cannot run while the skeleton shows, because the guard mounts the body only once content is ready.
- New `src/hooks/useMarkOpenReportEndOnSkeleton.ts`: a shared hook that closes the open-report telemetry span as a cold open (`markOpenReportEnd(report, {warm: false})`) the moment a cold-load skeleton appears. It replaces four identical inline effects (the guard, the orchestrator, `MoneyRequestReportView`, and `MoneyRequestReportActionsList`).
- `ReportScreen` now mounts `<ReportActions/>` instead of the old wrapper.
- `MoneyRequestReportView` mounts the body (`ReportActionsList`) directly and mounts `UserTypingEventListener` itself. It does not route through the orchestrator on purpose, because its money-request branching stays where it is, per the blueprint.
- Deleted: `ReportActionsView.tsx`, the old 58-line `inbox/ReportActionsList.tsx` wrapper, and `getReportActionsListInitialNumToRender.ts` (and its test).

### Unit Tests

- `tests/ui/ReportActionsTest.tsx`: orchestrator tests for the skeleton states, the choice between the money-request table and the chat list, the typing mount, and the app-load `warm: false` telemetry.
- `tests/ui/ReportActionsListTest.tsx`: body tests for concierge and fine-grained skeleton behavior. It mounts the body directly with a seeded Onyx store.
- `tests/ui/MoneyRequestReportViewTest.tsx`: the second consumer. It mounts the body and the typing listener on the chat path, and the table view on the money-request path.
- `tests/perf-test/ReportActionsList.perf-test.tsx`: rewritten against the new body surface (see below).

### Perf: not a like-for-like regression

The render count increased from 3 to 5 and it has a concrete explanation.
With static props the component rendered straight through in ~3 passes. 
With only `reportID`, it subscribes to many `useOnyx` keys that resolve asynchronously after mount, that settles triggers a re-renders.

**Instability fixed along the way.** While reviewing the relocated pipeline I found a real instability: `useReportActionsVisibility` built its `{[reportID]: slice}` lookup as an inline computed-key object literal. The React Compiler refuses to memoize a dynamically-keyed object, and pruned memoization of the visible-actions filter that depends on it, so `sortedVisibleReportActions` got a fresh reference every render. That unstable array flowed into the list context value, re-rendering the list body on every settling subscription. Moving the wrap into `reportVisibleActionsSelector` (a plain selector `useOnyx` hands back stably) removed the computed-key literal and let the compiler memoize the filter again. This was a pre-existing instability (on main the same array was passed as an unstable prop into `memo(ReportActionsList)`), not something the decomposition introduced. It also resolves the review note about the `state`/`actions` context values repainting `ReportActionsListContent` on guard-only re-renders: `actions` was already referentially stable and the compiler already memoizes `state`; this one field was the only churn. Verified by `useReportActionsListModelStability.test.tsx` (state and actions keep the same reference across a guard re-render with unchanged data).

### Fixed Issues

$ https://github.com/Expensify/App/issues/89767
PROPOSAL:

### Tests

### Test 1: Open a regular chat

1. Open a regular chat (1:1 or group).
2. Expect: Messages render; no infinite skeleton; can scroll.

### Test 2: Deep-link to an old message

1. Deep-link to an old message (open a link to a specific reportAction).
2. Expect: Opens scrolled to that message, not the bottom.

### Test 3: Open an expense report with multiple transactions

1. Open an expense report with multiple transactions.
2. Expect: Shows the table view (`MoneyRequestReportActionsList`), not the chat list.

### Test 4: Open an expense report that renders chat-style

1. Open an expense report that renders chat-style (no table view).
2. Expect: Shows the chat list.

### Test 5: Open a single expense / transaction-thread report

1. Open a single expense / transaction-thread report (drill into one transaction inside a money report).
2. Expect: Chat thread renders below the money-request header.

### Test 6: Typing indicator in a 1:1 chat

1. In a 1:1 chat, have the other party type (or use two sessions).
2. Expect: "… is typing" appears (route path).

### Test 7: Typing indicator in a transaction-thread report

1. In a transaction-thread report, have someone type.
2. Expect: Typing indicator appears here too (only if single expense).

### Test 8: Open the Concierge DM

1. Open the Concierge DM.
2. Expect: Loads without a wrong-content flash

### Test 9: Open an unread report

1. Open an unread report.
2. Expect: It becomes read (unread badge clears in LHN).

### Test 10: Scroll up to load older messages, then back down

1. Scroll up to load older messages, then back down.
2. Expect: Older messages load; scrolling to the bottom marks the newest as read.

### QA Steps

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/0c2a7876-612e-43b6-9671-b60ede7fdde9

</details>

<details>

https://github.com/user-attachments/assets/734db7e6-f467-4fd0-bcec-7cb2bc5d7141

</details>

<details>

https://github.com/user-attachments/assets/d250925f-dd66-4774-acdd-3536733ee32e

</details>

<details>

https://github.com/user-attachments/assets/67f7092d-7be8-4a12-a437-e1a660b190db

</details>

<details>

https://github.com/user-attachments/assets/cb030a45-9a3f-484a-8503-a9f8c3b16d5f

https://github.com/user-attachments/assets/6946b0d1-95fd-41c7-b2ac-8d4fb673dd56

</details>


LukasMod:
Regardless failing perf test, check PR description section about it. I compared sending messages with profiler and spans and didn't spot difference in performance (only noise)

image

src/pages/inbox/report/ReportActionsList.tsx
```diff
@@ -184,34 +241,41 @@ function ReportActionsList({
...
-        hasOnceLoadedReportActions: !!reportLoadingState?.hasOnceLoadedReportActions,
+        hasOnceLoadedReportActions: !!hasOnceLoadedReportActions,
     });
 
     const {markNewestActionAsRead, completeSkippedMarkAsRead} = useMarkAsRead({
-        reportID: report.reportID,
+        reportID,
         report,
         transactionThreadReport,
         sortedVisibleReportActions,
```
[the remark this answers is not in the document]
LukasMod: should be fixed

LukasMod:
@codex review

src/pages/inbox/ReportActions.tsx
```diff
@@ -52,7 +71,19 @@ function ReportActionsList() {
...
-    return <ReportActionsView reportID={reportIDFromRoute} />;
+    if (shouldShowAppLoadSkeleton) {
+        return <ReportActionsSkeletonView />;
+    }
+
+    return (
+        <>
+            <ReportActionsList
+                key={report.reportID}
+                reportID={report.reportID}
```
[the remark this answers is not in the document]
LukasMod: Subscription returns either the new report or undefined, never the previous one and we keep relay on `report` around instead. I will keep it as it is

LukasMod:
@DylanDylann Could you take a look on this?

DylanDylann:
Sure, I will complete my review this week

src/hooks/useReportActionsScroll.ts
```diff
@@ -28,8 +28,11 @@ import useReportScrollManager from './useReportScrollManager';
 import useScrollToEndOnNewMessageReceived from './useScrollToEndOnNewMessageReceived';
 
 type UseReportActionsScrollParams = {
+    /** The ID of the report currently being looked at */
+    reportID: string;
```
DylanDylann, line 32: Why do we need to update this hook? Why don't we use `report.reportID`?

src/pages/inbox/report/ReportActionsList.tsx
```diff
@@ -453,29 +534,85 @@ function ReportActionsList({
...
+    }, [report, shouldShowInitialSkeleton]);
+
+    if (isLoadingOnyxValue(reportResult) || !report || !isReportReady || shouldShowInitialSkeleton) {
+        return <ReportActionsSkeletonView />;
+    }
+
+    const hasDerivedValueTimingIssue = reportActions.length > 0 && isMissingReportActions;
+    if ((hasDerivedValueTimingIssue || (!isReportTransactionThread && isMissingReportActions)) && !showConciergeSidePanelWelcome) {
+        return <ReportActionsSkeletonView shouldAnimate={false} />;
+    }
```
DylanDylann: ~Could we handle rendering the skeleton view entirely in the parent component?~

src/pages/inbox/report/ReportActionsView.tsx
```diff
@@ -1,257 +0,0 @@
...
-
-type ReportActionsViewProps = {
-    /** The ID of the report to display actions for */
-    reportID: string | undefined;
-
-    /** Callback executed on layout */
-    onLayout?: (event: LayoutChangeEvent) => void;
-};
-
-function ReportActionsView({reportID, onLayout}: ReportActionsViewProps) {
```
DylanDylann, line 33: I think this component serves as a good guard. It allows us to return early before executing a lot of the logic in ReportActionsList. Is there a particular reason why we’re removing this guard? 


Otherwise, if we remove this component, I think we should move all ReportActionsSkeletonView guards to the parent component rather than passing them down to ReportActionsList.

src/hooks/useReportActionsScroll.ts
LukasMod: Same reasoning as the useMarkAsRead [thread](https://github.com/Expensify/App/pull/91179/changes#r3343879263): we keep `reportID` as an explicit non-nullable string because report is an OnyxEntry, so `report?.reportID` could be undefined. 
Here `reportID` is used on every render (not just in callbacks) to build Onyx subscription keys and the child hooks `useReportUnreadMessageScrollTracking` / `useReportActionsNewActionLiveTail` and the scroll resetKey. The hook runs during cold load while report is still undefined, so deriving from `report.reportID` would build those subscriptions from undefined and force a re-subscribe once report resolves

src/pages/inbox/report/ReportActionsView.tsx
LukasMod: The guard isn't removed, it's split by what data the decision needs, and the early-return part actually got stronger.

 Before, `ReportActionsView` mounted, ran all its data hooks and effects, and only then short-circuited to a skeleton, so the guard still paid for the heavy logic. Now the coarse states that need none of the body's data (`!report, shouldWaitForTransactions`, `isLoadingApp && !isOffline`) are decided in the route-only `ReportActions` orchestrator, which returns the skeleton and never mounts the body. For those states the body's hooks and effects don't run at all, which is the whole perf goal here.

The skeletons left in `ReportActionsList` can't be hoisted because they're computed from data the body derives: `isLoadingInitialReportActions/isMissingReportActions (initial load), isUnreadMessagePageLoadingInitially (isReportReady), isConciergeHiddenHistory + allReportActions.length` (concierge panel), and `reportActions.length `(derived-value timing). 

All of these come from the body's own hooks (`usePaginatedReportActions`, `useReportActionsVisibility`, concierge wiring, unread marker). Evaluating them in the parent means re-running those hooks there, which re-creates the wide-subscription monolith we're trying to decompose and double-subscribes to REPORT_ACTIONS.

LukasMod:
@DylanDylann I answered to feedback and pulled main just in case 👍

DylanDylann:
Thanks. Checking again

DylanDylann:
@LukasMod The failed Reassure Performance is intentional, right?

src/pages/inbox/report/ReportActionsView.tsx
DylanDylann: Previously, some hooks like useUnreadMarker, useMarkAsRead, useReportActionsScroll, and other definitions in ReportActionsList weren't executed if the guard in ReportActionsView rendered the skeleton view. With this change, all of the above logic will be triggered even when the skeleton view is rendered. My concern is that this may change the app's behavior when the skeleton view is rendered.

DylanDylann: I agree with you that there were previously some duplicated subscriptions in ReportActionsView and ReportActionsList. However, I think it's fine to keep separate subscriptions in different wrappers; that's how we avoid passing lots of props into child components.

I would like to hear your thoughts

LukasMod: I went back and reanalyzed every effect that now runs while the skeleton is up, since the body stays mounted.
I added extra guards for risky hooks with `isContentVisible`.

Only two of them have an observable external effect, and both are now guarded with `isContentVisible`: `readNewestAction` in `useMarkAsRead`, and the live-tail subscription in `useReportActionsNewActionLiveTail` (the one that can `openReport` / navigate). Everything else is either local-only or already ran during the old skeleton. `useUnreadMarker` is just passive listeners plus pure derivation, and the rest of the scroll effects no-op because the list ref isn't attached yet.

On moving the skeleton logic up into a wrapper or keeping separate subscriptions: this comes down to the architecture we're building toward. The whole point of the decomposition is that the body is the single self-subscribing owner of its data pipeline. The fine-skeleton conditions aren't raw Onyx values, they're outputs of that pipeline (visibility-filtered actions, concierge hidden-history, oldest-unread), so "is a skeleton showing" is a property the body derives, not something a parent can know.

To decide the skeleton in a wrapper, the wrapper has to own a second copy of the pipeline, and now there are two owners of the same data that can disagree on what to render. To decide it by not mounting the body, you have to lift those derived values into the parent and pass them back down as props, which is the ~18-prop interface this PR exists to delete. Either way you give up the single-owner boundary we just established.

So I kept the gate where the data lives: a single `isContentVisible` flag derived from the same `shouldShowSkeleton` that drives the render. The body stays the one source of truth and stays self-subscribing, and the gate can't drift from the skeleton because it's the same value. I did revisit the other approaches before settling here, and this one still came out the cleanest. WDYT

LukasMod:
@DylanDylann  Yes, intentional. The count went up because the new harness measures the real list mount through the actual pipeline, where the old one fed pre-computed props. 

I isolated where they come from. I short-circuited the body to render a trivial view after all the hooks run, keeping every hook mounted but dropping the InvertedFlashList subtree:
  - full body: 6 commits
  - FlashList removed, hooks kept: 2 commits
  
So our decomposition hooks account for only 2 commits. The other 4 are the InvertedFlashList doing its own mount layout passes over 500 items, which is pre-existing behavior identical to what main's body renders. So the count jump isn't coming from the decomposition, it's more like a new test setup with better picture, so it could be now our new baseline with data populated with onyx

LukasMod:
@codex review

src/hooks/useMarkAsRead.ts
```diff
@@ -30,6 +30,11 @@ type UseMarkAsReadParams = {
     sortedVisibleReportActions: OnyxTypes.ReportAction[];
     isScrolledToEnd: boolean;
     hasNewerActions: boolean;
+    /**
+     * Whether the report is loaded enough to be displayed.
+     * The automatic mark-as-read effects are skipped until this is true
+     */
+    isContentVisible: boolean;
```
adhorodyski: just noting here such guards are a big code smell

src/pages/inbox/report/ReportActionsView.tsx
LukasMod: I will revisit the approach again and search for better solution

src/hooks/useMarkAsRead.ts
LukasMod: working on removing these

LukasMod:
@DylanDylann I will ping you when its ready. I still need to do some benchmarks with new appoach and double check tests

LukasMod:
@codex review

LukasMod:
Benchmarks:
- ManualSendMessage

Screenshot 2026-06-23 at 11 05 12

- OpenReport 

Screenshot 2026-06-23 at 11 17 03

src/hooks/useReportActionsData.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Read/derive-only data pipeline for the report-actions list.
+ *
+ * Bundles every subscription + derivation needed for both the skeleton decision and the list render, so
+ * the guard's decision and the content's render share one pipeline. Intentionally side-effect-free: NO
+ * Onyx writes, navigation, telemetry, or session-start. `ReportActionsSkeletonGuard` calls this once and
+ * provides the result via `ReportActionsDataContext`, so the content reads it instead of subscribing again.
+ */
+function useReportActionsData(reportID: string) {
```
adhorodyski: this hook is massive and still feels like is mixing concerns, I get the idea of encapsulating 'data' fields but this will only grow over time. if you had to name data for the `guard` as a separate content, it's like admitting it's a separate concerns already. does it make sense to only call these manually in a guard?

adhorodyski: another thing is it gets called as a `data` hook but encapsulates many non-data concerns like visibility/scrolling/fetching

LukasMod: The guard/content split is a render-isolation. `guardData` is deliberately kept off the context so its churn (loading flags, navigation, etc) never re-renders the list. `contentData` is the only thing on the context. The split is by who re-renders on this field. To decide skeleton-vs-content I need single place with all of these at the first place. 

Right know flow is:
1. guard subscribes once, 
2. guard decides skeleton vs content
3. guard pass contentData down via context so ReportActionsListContent renders from the same derivations without re-subscribing

If I split it, I will have to do double job at this point

Second one: 
I can work on different structure within contentData, separate the derived values from the command handles, e.g. {...derived, actions: {loadOlderChats, loadNewerChats, handleShowPreviousMessages, setTreatAsNoPaginationAnchor}}. Also maybe rename  `useReportActionsData` → something like `useReportActionsListModel`/`useReportActionsPipeline`

DylanDylann:
@LukasMod If possible, could you split this PR into a few smaller ones? I believe it would make it easier for us to move forward

src/pages/inbox/report/ReportActionsList.tsx
```diff
@@ -540,4 +536,21 @@ function ReportActionsList({
     );
 }
 
-export default memo(ReportActionsList);
+const MemoizedReportActionsListContent = memo(ReportActionsListContent);
```
adhorodyski, line 546: does the OG blueprint memo it as well? or this will be removed later?

src/pages/inbox/report/ReportActionsSkeletonGuard.tsx
```diff
@@ -0,0 +1,99 @@
...
+            return;
+        }
+        updateLoadingInitialReportAction(report?.reportID ?? reportID);
+    }, [isOffline, report?.reportID, reportID, reportActionIDFromRoute]);
+
+    useLayoutEffect(() => {
+        if (!isConciergeMainDM || !canStartConciergeSession) {
+            return;
+        }
+        startSession(oldestUnreadReportAction ? report?.lastReadTime : undefined);
```
adhorodyski: why does the guard call effects at all? they do not seem related

src/pages/inbox/report/ReportActionsList.tsx
LukasMod: OG removed that one at some point later. We can do it after PR-10 (hasHeaderRendered ref) + PR-12 probably.
RIght now its also pointing to this line so maybe even more PRs will be needed
```    
const [hasScrolledOverThreshold, setHasScrolledOverThreshold] = useState(() => scrollOffsetRef.current >= CONST.REPORT.ACTIONS.ACTION_VISIBLE_THRESHOLD);
```

src/pages/inbox/report/ReportActionsSkeletonGuard.tsx
LukasMod: I will double check these, one is based on shouldShowInitialSkeleton so probably will have to stay here (but maybe as custom hook) but others maybe can be relocated. I need to test that as there is a risk of some flickering

LukasMod:
@codex review

LukasMod:
@codex review

src/hooks/useReportActionsListModel.ts
```diff
@@ -0,0 +1,162 @@
...
+        isConciergeHiddenHistory,
+        isConciergeMainDM,
+        hasCachedReportActions,
+        showConciergeSidePanelWelcome,
+    };
+
+    // The render state slice on `ReportActionsListStateContext`; this is what drives list re-renders.
+    // Ambient state (network, route, archived, concierge session) is read locally in the content instead
+    // of carried here.
+    const state = {
```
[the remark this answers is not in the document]
LukasMod: 1. "the actions object below is recreated on every render" — false. actions was already referentially stable (React Compiler memoizes it). The stability test confirms actionsRefStable: true even before the fix.
  2. "wrap these provider values in useMemo" — wrong fix. The compiler already memoizes both state and actions. A manual useMemo on state would have busted exactly the same way, because the real cause wasn't the object literal — it was one unstable field inside it (sortedVisibleReportActions, from
  the computed-key prune). The compiler-memoized state recomputed because that field's reference changed every render. Stabilizing the field upstream made the already-memoized state stable, with no manual memoization anywhere. https://github.com/Expensify/App/pull/93694/changes/2c4a6ea5e7a81435b6ffd4fef7419bf932dcb068

LukasMod:
@codex review

src/pages/inbox/report/ReportActionsSkeletonGuard.tsx
LukasMod: I relocated it a bit, renamed and moved logic into custom hooks, but without huge changes to keep this safe as its already quite big

LukasMod:
@codex review

LukasMod:
@codex review

LukasMod:
@DylanDylann I looked into splitting this and I don't think it's worth it, split would be like 5% + 95% of changes.
The orchestrator, the skeleton guard, the context, the model hook, the body rewire, both consumers getting repointed, and deleting ReportActionsView, is one interlocking change. 

What is good, its mostly copy/pasting and unit tests. I updated PR description to make it a bit easier. Also with last commits number of changes reduced a lot so it should be a bit easier. Ready for review

DylanDylann:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari


### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/34494dc7-c251-4855-adf3-5545d2321c43



</details>

<details>


https://github.com/user-attachments/assets/2034a1db-32e3-478e-bb26-a3f5e7791202


</details>

<details>


https://github.com/user-attachments/assets/117ebb33-c1e1-4487-8484-a87ef16d9366


</details>

<details>

https://github.com/user-attachments/assets/7200157b-87fe-4435-8233-ff1ea56e27f1



</details>

<details>

https://github.com/user-attachments/assets/f982801a-6006-46d9-a4b8-5126800defb0



</details>

DylanDylann (approved):
Thanks @LukasMod. It's great job

DylanDylann:
@rlinoz Just a note: the failed [Reassure Performance Tests](https://github.com/Expensify/App/actions/runs/28170942502/job/83443824408?pr=93694) are intentional

LukasMod:
new conflicts resolved

src/hooks/useReportActionsScroll.ts
rlinoz: maybe this is worth a comment.

rlinoz (approved):

PR #93807, opened 2026-06-17
[NO QA] feat(dev): add the XState inspector with sensitive-data masking

dariusz-biela:
### PR Stack

This PR is part of the MFA -> XState state-machine migration (https://github.com/Expensify/App/issues/81197), shipped as a stacked-PR train. All work merges into the integration branch `dariusz-81197-mfa-state-machine` on `Expensify/App`; nothing reaches `main` until the final integration merge.

| # | PR | Branch | Base |
|---|----|--------|------|
| 1 | https://github.com/Expensify/App/pull/93055 (merged) | `dariusz-biela/refactor/3ds/mfa-cleanup-for-state-machine` | `dariusz-81197-mfa-state-machine` (integration) |
| 2 | https://github.com/Expensify/App/pull/93180 (merged) | `dariusz-biela/refactor/3ds/mfa-state-machine-setup` | `dariusz-81197-mfa-state-machine` (integration) |
| 3 | **this PR** | `dariusz-biela/refactor/3ds/xstate-inspector-dev-web` | `dariusz-81197-mfa-state-machine` (integration) |

This branch is stacked on top of the PR-2 branch (`dariusz-biela/refactor/3ds/mfa-state-machine-setup`), which exists only on the SWM fork and cannot be used as a base on `Expensify/App`, so this PR targets the integration branch directly. PR-2 (https://github.com/Expensify/App/pull/93180) has now merged into the integration branch, so the diff here is narrowed to only this PR's own changes.

### Explanation of Change

Developer-tooling step of the MFA -> XState migration: a dev-only way to visualize and debug the machine, with safeguards that keep it safe to point at a flow that handles secrets.

**In the app**

- Adds an `XState inspector` row to the dev menu (`Troubleshoot` > `XState inspector` > `Open`), on web dev builds only.
- Pressing `Open` launches the Stately visualizer (stately.ai/inspect) for the MFA machine, where you can watch its states, transitions, events and context update live as you drive the flow.
- Sensitive values (auth tokens, PIN, card pan/cvv/expiration, challenges, validate code, and the rest of the shared `CONST.SENSITIVE_AUTH_KEYS` denylist) show up masked as `***` - field names and shape stay visible.
- Only one MFA machine appears, with no duplicate "ghost" machines stuck in the initial state.

**In the code**

- `useInspectedMachine` - a thin `useMachine` drop-in that pre-wires the inspector; the MFA Provider now uses it.
- New `XStateInspector` lib: a web entry guarded by `__DEV__` + `require()` (so the code and `@statelyai/inspect` are dropped from production bundles) and a native no-op stub, plus the dev-menu wiring.
- `maskSensitive.ts` - a custom inspector serializer that masks every value under a sensitive key at any depth while keeping the shape.
- `filterGhostActorRegistrations.ts` - hides actors created in render passes React discarded (registered but never started), which would otherwise show as duplicate machines.
- Dev-menu labels in all `src/languages/*` locales, and unit tests for the masking serializer and the ghost-actor filter.

**Shared sensitive-key denylist**

The inspector masker, the request / response logs (`sanitizeLogParams`) and the thrown parameter errors (`requireParameters`) each kept their own list of sensitive keys. Those lists had drifted apart: the masker shared only `password` / `token` / `validateCode` with the others and let top-level auth tokens (e.g. `authToken`, `idToken`, `twoFactorAuthCode`, `partnerUserSecret`) reach stately.ai unmasked.

- Moves the credential and secret key names into a single `CONST.SENSITIVE_AUTH_KEYS` list, consumed by all three redaction sites. Adding a key there now protects every place data leaves the app at once, so the lists can no longer drift apart.
- `maskSensitive` masks exactly that shared set at any depth. `payload` / `body` are no longer masked wholesale but recursed into, so their sensitive leaves are redacted while the surrounding structure stays visible for debugging.
- As a side effect, `sanitizeLogParams` and `requireParameters` now also redact card and WebAuthn material (e.g. `pin`, `pan`, `cvv`, `otp`, `secret`, `expiration`, `keyInfo`, `challenge`).

### Fixed Issues

$ https://github.com/Expensify/App/issues/81197
PROPOSAL: N/A (internal refactor, part of the MFA -> XState migration plan)

### Tests

Run the app on **web in a dev build** (the inspector is dev + web only):

1. Trigger any MFA scenario (e.g. Reveal PIN, or the Biometrics test tool row).
2. Open the dev menu (`Troubleshoot`) and tap `XState inspector` > `Open`. Verify the Stately inspector window opens and shows the MFA machine with its states and transitions.
3. Drive the flow and verify events/transitions stream into the inspector in order, with the current state highlighted.
4. Trigger the change / set PIN flow - currently the only scenario that carries a secret through the machine, since it adds `pin` to the INIT payload - and inspect the machine's events / context in the inspector. Verify the `pin` value appears masked as `***`, while the field name and surrounding shape stay visible. (The serializer also guards card pan/cvv/expiration, challenges, validate code and tokens, but the simplified machine does not carry those yet, so they have no masked fields to show until later slices add them.)
5. Verify there is exactly one MFA machine in the inspector - no duplicate "ghost" machines stuck in the initial state - including after a Fast Refresh / hot reload.
6. Re-open the dev menu after a fresh load and confirm the row is present only on web dev builds.
7. Verify no errors appear in the JS console.

### Offline tests

N/A - this PR adds dev-only tooling with no network behavior; it is stripped from production builds and absent on native.

### QA Steps

[No QA] - internal integration-branch PR targeting `dariusz-81197-mfa-state-machine`; the inspector is a `__DEV__`-only, web-only developer tool that is dropped from production bundles and stubbed out on native, so there is nothing user-facing to validate on staging or production. Covered by the manual Tests above.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

https://github.com/user-attachments/assets/5fe20d0c-850e-4cbe-8fc6-4c8365eeceba


src/hooks/useInspectedMachine.ts
```diff
@@ -0,0 +1,20 @@
+import {useMachine} from '@xstate/react';
+import type {ActorOptions, AnyStateMachine} from 'xstate';
+import xstateInspector from '@libs/XStateInspector';
+
+/**
+ * A drop-in replacement for `useMachine` that also wires up the dev-only Stately inspector. Use it
```
jakubkalinski0, line 7: Would it be a good idea to enforce this through an `ESLint` rule? We could make sure nobody imports `useMachine` directly in the future and accidentally bypasses the inspector. WDYT?

src/CONST/index.ts
```diff
@@ -210,6 +210,42 @@ const EMAIL = {
...
 const CONST = {
+    /**
+     * Credential and secret keys that must be redacted before any user data leaves the running app.
+     * Every place where data can leave redacts them, such as the request and response logs
+     * (`sanitizeLogParams`), the thrown parameter errors (`requireParameters`), and the XState
+     * inspector stream (`XStateInspector/maskSensitive`). Keeping the keys in this single list is
+     * what stops those sites from drifting apart, so always add a new key here rather than at each
+     * call site.
+     */
+    SENSITIVE_AUTH_KEYS: [
```
jakubkalinski0, line 221: `maskInspectionEvent` redacts values under sensitive key names. But there is one case in action union that has secrete as bare value under a generic key:

https://github.com/Expensify/App/blob/eaabc7c6a611fdf39fa05a57aff05d967653427e/src/components/MultifactorAuthentication/Context/types.ts#L70

So if this event flows through the machine later, I think `{type: 'SET_VALIDATE_CODE', payload: '123456'}` streams to `stately.ai` unmasked.

In that case we should make this secret an object like others which would allow us to have it under a sensitive key (e.g. `validateCode` instead of string)? Also failure is silent so a `MfaEvent + a bare-payload` test might be worth it. WDYT?

jakubkalinski0 (approved):
LGTM 🚀 Very nice addition - some NIT comments but overall it looks good

src/libs/XStateInspector/filterGhostActorRegistrations.ts
```diff
@@ -0,0 +1,52 @@
+import type {InspectionEvent, Observer} from 'xstate';
+
+/**
+ * Wraps the inspector's observer so that each `@xstate.actor` registration is held back until a
+ * later inspection event arrives for the same session, which keeps "ghost" actors out of the
+ * inspector.
+ *
+ * A ghost appears because `useMachine` calls `createActor` during render, and XState announces the
```
roryabraham: Is there an upstream issue for this? It's effectively a patch for what's presumably a very common issue, right?

src/libs/XStateInspector/index.ts
```diff
@@ -0,0 +1,47 @@
...
+function createXStateInspector(): XStateInspector {
+    if (!__DEV__) {
+        return {inspect: undefined};
+    }
+
+    // These modules are pulled in with require() inside the dev-only branch rather than with static
+    // imports, so that webpack drops them from production bundles together with this dead branch.
+    // Because require() returns `any`, the type assertions restore each module's real shape from the
+    // type-only imports above, which webpack erases.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
```
roryabraham: annoying that `require` can't figure out types. I don't really think dynamic `import` will either though.

src/libs/XStateInspector/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    const {createBrowserInspector} = require('@statelyai/inspect') as {createBrowserInspector: typeof createBrowserInspectorFn};
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+    const {maskInspectionEvent} = require('./maskSensitive') as {maskInspectionEvent: typeof maskInspectionEventFn};
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+    const {filterGhostActorRegistrations} = require('./filterGhostActorRegistrations') as {filterGhostActorRegistrations: typeof filterGhostActorRegistrationsFn};
```
roryabraham: NAB: use dynamic `import` instead of require. I feel like this is generally more future-proof. I'm trying to get to a world without CJS, because the CJS/ESM boundary is always confusing. This also allows you to fire off all imports in parallel and use `Promise.all`, even though it doesn't matter in this case for performance.

roryabraham (approved):

src/hooks/useInspectedMachine.ts
```diff
@@ -0,0 +1,20 @@
...
+ * instead of `useMachine` so that every machine automatically appears in the inspector window
+ * (Troubleshoot > XState inspector) with its sensitive fields masked.
+ *
+ * It is declared as an overload because the options parameter of `useMachine` is a conditional tuple
+ * type that TypeScript cannot resolve for a generic machine. The implementation therefore uses
+ * `AnyStateMachine`, for which the conditional does resolve, while callers still keep their precise
+ * machine types.
```
chuckdries, line 14: nice 👍

src/libs/XStateInspector/index.ts
chuckdries: Agree

src/libs/XStateInspector/filterGhostActorRegistrations.ts
chuckdries: Also curious about this

chuckdries (commented):
I tested it, the inspector is pretty slick! Rory already asked all the questions I have. I'll approve when we rewrite the require as an import

src/libs/XStateInspector/index.ts
dariusz-biela: I'm open to this change, but I want to flag a tradeoff before we decide.
  
  `require()` is convenient here because it's synchronous: the inspector singleton is ready at module load, and both `useInspectedMachine` and the dev-menu button read inspect/start synchronously. Switching to
  `import()` makes that async, so we'd need one of two approaches:

  1. Delay the machine's initialization until the inspector has loaded, otherwise we miss the `@xstate.actor` registration event. The catch: `MultifactorAuthenticationContextProvider` is mounted app-wide in
  `AuthScreens`, so the machine is created at app startup. Delaying it would block app init in dev until the inspector chunk loads, and make the machine start later in dev than it does in prod.
  2. Wrap the inspector in a small observer of our own that buffers events and forwards them once the library has loaded. This keeps the machine fully synchronous (identical to prod) and just lets the inspector
  attach when it's ready. 

  Given (1)'s impact on app startup, I lean towards (2). That said, since this comment is non-blocking, a third option is to keep `require()` as-is given the synchronous-singleton constraint, if you're comfortable
  with the documented type assertions. Which would you prefer?
  
  
  Implementation of (2):
```ts
        let liveObserver: Observer<InspectionEvent> | undefined;
        let startInspector: (() => void) | undefined;
        let isStartRequested = false;
        const bufferedEvents: InspectionEvent[] = [];

        const inspect: Observer<InspectionEvent> = {
            next: (event) => {
                if (!liveObserver) {
                    bufferedEvents.push(event);
                    return;
                }
                liveObserver.next?.(event);
            },
            error: (err) => liveObserver?.error?.(err),
            complete: () => liveObserver?.complete?.(),
        };

        Promise.all([import('@statelyai/inspect'), import('./maskSensitive'), import('./filterGhostActorRegistrations')]).then(
            ([{createBrowserInspector}, {maskInspectionEvent}, {filterGhostActorRegistrations}]) => {
                const inspector = createBrowserInspector({
                    autoStart: false,
                    serialize: maskInspectionEvent,
                });
                liveObserver = filterGhostActorRegistrations(inspector.inspect);
                startInspector = () => inspector.start();
                for (const event of bufferedEvents) {
                    liveObserver.next?.(event);
                }
                bufferedEvents.length = 0;
                if (isStartRequested) {
                    startInspector();
                }
            },
        );

        return {
            inspect,
            start: () => {
                if (!startInspector) {
                    isStartRequested = true;
                    return;
                }
                startInspector();
            },
        };
```

src/libs/XStateInspector/filterGhostActorRegistrations.ts
dariusz-biela: The closest upstream report I found is https://github.com/statelyai/inspect/issues/44. It covers React Strict Mode and rerenders breaking the inspector and is still open. However, it doesn’t identify the never-started actor registration mechanism, and I couldn’t find a more specific issue in statelyai/xstate. Before opening a focused upstream issue, I’ll try to prepare a minimal CodeSandbox reproduction, if possible.

src/libs/XStateInspector/index.ts
chuckdries: 2 seems reasonable to me

DylanDylann: My preference is to keep the require(). Its drawbacks here are cosmetic, while keeping it remains the lower risk choice and  has served us well in this case

DylanDylann (approved):

src/libs/XStateInspector/index.ts
DylanDylann: I'm not sure if we have a plan to remove require() usage across the codebase. If we do, alternative option 2 seems fine. If not, I suggest we leave it as is

src/CONST/index.ts
dariusz-biela: Good catch, and I agree with the shape you're proposing (carry the secret as an object under a sensitive key, e.g. validateCode, plus a bare-payload MfaEvent test).
  
  One clarification on scope though: that {type: 'SET_VALIDATE_CODE'; payload: string} action only exists in the legacy reducer union (Context/types.ts), which is slated for removal. The state machine's MfaEvent
  doesn't have any secret-carrying event yet - the semantic input events (validate code, soft prompt, ...) get added by the slices that introduce their states. So nothing is actually streaming an unmasked code
  through the inspector on this branch today.

  I'd rather configure the event interface in the right place (the machine's MfaEvent) when that slice lands, and add the bare-payload masking test alongside it, instead of shaping it now against the reducer we're
  deleting. So I'll leave it out of this PR and handle it with the validate-code slice. I'll make sure the secret rides under a SENSITIVE_AUTH_KEYS field there so the masker redacts it.

src/libs/XStateInspector/filterGhostActorRegistrations.ts
dariusz-biela: I've opened an issue in the upstream repository at https://github.com/statelyai/inspect/issues/58. As soon as I hear back something from the maintainers, I'll let you know. I'll also add a comment in the code noting that this issue exists.

src/libs/XStateInspector/index.ts
dariusz-biela: Personally, I'm also leaning toward `require()`, but I can see the advantages of switching to `import()` as well.

So I'll leave the final decision up to you. As soon as things become clearer, I'll get started on the implementation.

chuckdries: One less require in the codebase feels like a win in the long term and I think the event buffer is pretty easy to parse. My only concern is I don't want to create a memory leak in cases where the user does not ever start the inspector. Your implementation sketch looks good IMO; I say we go for it

chuckdries: But if you find yourself rabbit hole-ing for some reason or another, stop and say so and I'll approve with require. I agree that it doesn't have to be a blocker

dariusz-biela: Okay, I'll try to put together an implementation using `import()` right away. If we don't like it, I'll just revert the commit, but at least we'll have a concrete implementation to discuss.

dariusz-biela: I've added the entire solution and made sure it's cleanly written and well-documented with comments. I've also added unit tests to the utils, so this solution should work and be relatively easy to maintain, despite the increased code size.

All in all, this solution using `import()`—aside from being more compliant with new standards—may be slightly better in terms of performance impact, since it isn’t blocking like `require()` is, so I think this change may ultimately work out in our favor.

chuckdries (approved):
Clean!

src/libs/XStateInspector/createDeferredStart.ts
```diff
@@ -0,0 +1,33 @@
+import type {LoadedInspector} from './types';
+
+/**
+ * Bridges the dev-menu button, which fires synchronously, to the asynchronously loaded inspector. A
```
roryabraham: I feel like a more intuitive and simpler design is to just hide the button if the inspector isn't loaded yet, and show it when it loads.

<details>
<summary>diff</summary>

```diff
diff --git a/src/components/TestToolMenu.tsx b/src/components/TestToolMenu.tsx
index 60db8c8f38c..e73153c3bf1 100644
--- a/src/components/TestToolMenu.tsx
+++ b/src/components/TestToolMenu.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {Suspense} from 'react';
 import {Platform} from 'react-native';
 import useIsAuthenticated from '@hooks/useIsAuthenticated';
 import useLocalize from '@hooks/useLocalize';
@@ -19,6 +19,7 @@ import Switch from './Switch';
 import TestCrash from './TestCrash';
 import TestToolRow from './TestToolRow';
 import Text from './Text';
+import XStateInspectorTestToolRow from './XStateInspectorTestToolRow';
 
 function TestToolMenu() {
     const [network] = useOnyx(ONYXKEYS.NETWORK);
@@ -161,15 +162,11 @@ function TestToolMenu() {
                     disabled={!!network?.shouldForceOffline || network?.shouldSimulatePoorConnection}
                 />
             </TestToolRow>
-            {/* Opens the Stately inspector window, which shows every XState machine wired to it. The row appears on web dev builds only, where the inspector can run. */}
-            {!!xstateInspector.start && (
-                <TestToolRow title={translate('initialSettingsPage.troubleshoot.xstateInspector')}>
-                    <Button
-                        small
-                        text={translate('initialSettingsPage.troubleshoot.openXstateInspector')}
-                        onPress={xstateInspector.start}
-                    />
-                </TestToolRow>
+            {/* Opens the Stately inspector window, which shows every XState machine wired to it. The row appears on web dev builds only, once the inspector chunk has loaded. */}
+            {!!xstateInspector.ready && (
+                <Suspense fallback={null}>
+                    <XStateInspectorTestToolRow ready={xstateInspector.ready} />
+                </Suspense>
             )}
             <SoftKillTestToolRow />
             <TestCrash />
diff --git a/src/components/XStateInspectorTestToolRow.tsx b/src/components/XStateInspectorTestToolRow.tsx
new file mode 100644
index 00000000000..43bd27890e9
--- /dev/null
+++ b/src/components/XStateInspectorTestToolRow.tsx
@@ -0,0 +1,30 @@
+import React, {use} from 'react';
+import useLocalize from '@hooks/useLocalize';
+import type {LoadedInspector} from '@libs/XStateInspector/types';
+import Button from './Button';
+import TestToolRow from './TestToolRow';
+
+type XStateInspectorTestToolRowProps = {
+    ready: Promise<LoadedInspector | null>;
+};
+
+function XStateInspectorTestToolRow({ready}: XStateInspectorTestToolRowProps) {
+    const {translate} = useLocalize();
+    const inspector = use(ready);
+
+    if (!inspector) {
+        return null;
+    }
+
+    return (
+        <TestToolRow title={translate('initialSettingsPage.troubleshoot.xstateInspector')}>
+            <Button
+                small
+                text={translate('initialSettingsPage.troubleshoot.openXstateInspector')}
+                onPress={inspector.start}
+            />
+        </TestToolRow>
+    );
+}
+
+export default XStateInspectorTestToolRow;
diff --git a/src/libs/XStateInspector/createDeferredStart.ts b/src/libs/XStateInspector/createDeferredStart.ts
deleted file mode 100644
index 597c99523d5..00000000000
--- a/src/libs/XStateInspector/createDeferredStart.ts
+++ /dev/null
@@ -1,33 +0,0 @@
-import type {LoadedInspector} from './types';
-
-/**
- * Bridges the dev-menu button, which fires synchronously, to the asynchronously loaded inspector. A
- * press made before the chunk resolves is remembered and replayed on load, so the window still opens.
- * If the chunk fails to load, the returned function stays a no-op.
- */
-function createDeferredStart(ready: Promise<LoadedInspector>): () => void {
-    let start: (() => void) | undefined;
-    let isStartRequested = false;
-
-    ready.then(
-        (inspector) => {
-            start = inspector.start;
-            if (isStartRequested) {
-                start();
-            }
-        },
-        () => {
-            // The inspector chunk failed to load, so leave start() as a no-op.
-        },
-    );
-
-    return () => {
-        if (start) {
-            start();
-            return;
-        }
-        isStartRequested = true;
-    };
-}
-
-export default createDeferredStart;
diff --git a/src/libs/XStateInspector/index.ts b/src/libs/XStateInspector/index.ts
index 427105b694b..24275a40cc8 100644
--- a/src/libs/XStateInspector/index.ts
+++ b/src/libs/XStateInspector/index.ts
@@ -1,6 +1,5 @@
 import type {InspectionEvent} from 'xstate';
 import createBufferedObserver from './createBufferedObserver';
-import createDeferredStart from './createDeferredStart';
 import type {LoadedInspector} from './types';
 import type XStateInspector from './types';
 
@@ -32,7 +31,8 @@ function createXStateInspector(): XStateInspector {
 
     return {
         inspect: createBufferedObserver<InspectionEvent>(ready.then((inspector) => inspector.inspect)),
-        start: createDeferredStart(ready),
+        // Resolves to null on chunk load failure so consumers can use React's `use()` without an error boundary.
+        ready: ready.catch(() => null),
     };
 }
 
diff --git a/src/libs/XStateInspector/types.ts b/src/libs/XStateInspector/types.ts
index ead067fabff..ca90a87d7b8 100644
--- a/src/libs/XStateInspector/types.ts
+++ b/src/libs/XStateInspector/types.ts
@@ -13,12 +13,12 @@ type LoadedInspector = {
 /**
  * A dev-only handle for the Stately inspector. The `inspect` observer is wired into an actor when
  * the actor is created, and it must already exist before the first event so that the full history is
- * captured. The `start` function opens the visualizer window on demand, and it is undefined wherever
+ * captured. The `ready` promise resolves once the inspector chunk has loaded and is undefined wherever
  * the inspector cannot run, such as in production, on native, or in tests.
  */
 type XStateInspector = {
     inspect: Observer<InspectionEvent> | undefined;
-    start?: () => void;
+    ready?: Promise<LoadedInspector | null>;
 };
 
 export default XStateInspector;
diff --git a/tests/unit/XStateInspectorDeferredStartTest.ts b/tests/unit/XStateInspectorDeferredStartTest.ts
deleted file mode 100644
index 84fce783e93..00000000000
--- a/tests/unit/XStateInspectorDeferredStartTest.ts
+++ /dev/null
@@ -1,116 +0,0 @@
-import type {InspectionEvent, Observer} from 'xstate';
-import createDeferredStart from '@libs/XStateInspector/createDeferredStart';
-import type {LoadedInspector} from '@libs/XStateInspector/types';
-
-type Deferred<T> = {
-    promise: Promise<T>;
-    resolve: (value: T) => void;
-    reject: (reason?: unknown) => void;
-};
-
-function createDeferred<T>(): Deferred<T> {
-    let resolve: (value: T) => void = () => {};
-    let reject: (reason?: unknown) => void = () => {};
-    const promise = new Promise<T>((resolveFn, rejectFn) => {
-        resolve = resolveFn;
-        reject = rejectFn;
-    });
-    return {promise, resolve, reject};
-}
-
-function createRecordingInspector() {
-    let startCount = 0;
-    const inspect: Observer<InspectionEvent> = {
-        next: () => {},
-        error: () => {},
-        complete: () => {},
-    };
-    const inspector: LoadedInspector = {
-        inspect,
-        start: () => {
-            startCount += 1;
-        },
-    };
-    return {getStartCount: () => startCount, inspector};
-}
-
-describe('createDeferredStart', () => {
-    it('opens the window straight away when pressed after the inspector has loaded', () => {
-        const deferred = createDeferred<LoadedInspector>();
-        const {getStartCount, inspector} = createRecordingInspector();
-        const start = createDeferredStart(deferred.promise);
-
-        deferred.resolve(inspector);
-
-        return deferred.promise.then(() => {
-            start();
-            expect(getStartCount()).toBe(1);
-        });
-    });
-
-    it('remembers a press made before the inspector loads and replays it on load', () => {
-        const deferred = createDeferred<LoadedInspector>();
-        const {getStartCount, inspector} = createRecordingInspector();
-        const start = createDeferredStart(deferred.promise);
-
-        start();
-        expect(getStartCount()).toBe(0);
-
-        deferred.resolve(inspector);
-
-        return deferred.promise.then(() => {
-            expect(getStartCount()).toBe(1);
-        });
-    });
-
-    it('replays only once no matter how many presses are made before the inspector loads', () => {
-        const deferred = createDeferred<LoadedInspector>();
-        const {getStartCount, inspector} = createRecordingInspector();
-        const start = createDeferredStart(deferred.promise);
-
-        start();
-        start();
-        start();
-
-        deferred.resolve(inspector);
-
-        return deferred.promise.then(() => {
-            expect(getStartCount()).toBe(1);
-        });
-    });
-
-    it('forwards every later press straight through once the inspector is live, after replaying the early one', () => {
-        const deferred = createDeferred<LoadedInspector>();
-        const {getStartCount, inspector} = createRecordingInspector();
-        const start = createDeferredStart(deferred.promise);
-
-        start();
-        deferred.resolve(inspector);
-
-        return deferred.promise.then(() => {
-            expect(getStartCount()).toBe(1);
-            start();
-            start();
-            expect(getStartCount()).toBe(3);
-        });
-    });
-
-    it('stays a no-op when the inspector chunk fails to load, both before and after the failure', () => {
-        const deferred = createDeferred<LoadedInspector>();
-        const {getStartCount} = createRecordingInspector();
-        const start = createDeferredStart(deferred.promise);
-
-        start();
-        deferred.reject(new Error('chunk failed to load'));
-
-        return deferred.promise.then(
-            () => {
-                throw new Error('promise should have rejected');
-            },
-            () => {
-                start();
-                expect(getStartCount()).toBe(0);
-            },
-        );
-    });
-});

```
</details>

tests/unit/XStateInspectorBufferedObserverTest.ts
```diff
@@ -0,0 +1,106 @@
+import type {Observer} from 'xstate';
+import createBufferedObserver from '@libs/XStateInspector/createBufferedObserver';
+
+type Deferred<T> = {
```
roryabraham: Rather than creating a custom `Deferred` type, can we use the built-in [Promise.withResolvers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers)? we already have a polyfill in this project. Or we could bump `.nvmrc` to use Node 24 instead of Node 20...

roryabraham (changes requested):

src/libs/XStateInspector/createDeferredStart.ts
dariusz-biela: That's a great idea 🥇

roryabraham (approved):

Issue #93208, opened 2026-06-10
[Due for payment 2026-08-07] Add filter negation (is not / doesn't have) to the Spend filter UI


JmillsExpensify:
## Background

When admins narrow down expenses on the Spend page, they might want to express an exclusion rather than an inclusion: "show me everything that *isn't* on Adventure Inc," "everything that *doesn't have* a receipt," "every merchant *except* Big Buy Supply." Today the filter UI only lets users build positive matches, so to answer an exclusion question they have to either reason in reverse (select every value they *do* want) or drop into raw search syntax. Both are slow, and the second is invisible to most users.

The good news is that negation is already supported by search. This is purely a front-end initiative: we're exposing a control that lets users build a negated filter from the UI, and reflecting that negation in the resulting filter chip. Nothing about the underlying search syntax or back end changes.

We use _is not_ for value-based filters (Merchant, Workspace, etc.) and _doesn't have_ for presence-based filters (the Has filter), because that's the language that reads naturally for each filter type — you say a merchant "is not" something, but a receipt is something an expense "doesn't have."

<img width="2800" height="1624" alt="Image" src="https://github.com/user-attachments/assets/da377653-6765-4ae8-b0cc-065c4c5d4f09" />

<img width="2800" height="1624" alt="Image" src="https://github.com/user-attachments/assets/137618c3-2fa4-4579-b070-0f34ca82ce57" />

The full Figma file for this initiative can be [found here](https://www.figma.com/design/0JppKzxgCpgGq2L6Q0Q3xS/Filters-Rewamp?node-id=1681-112708&t=ZaZljDQWaQXDISpW-1).

## Scope
- Add a negation toggle to the filter configuration UI for the supported filters only. No new search capability — negation already works in search syntax.
- This is front-end only. No changes to search syntax, query parsing, or the back end.
- Supported filters in this release: To, From, Has, Workspace, Status, Currency, Purchase currency, Merchant, Exported to. These are where we're seeing customer demand. Other filters keep their current behavior and can be added later.
- Reflect the negated state in the filter chip label.

### Negation toggle

Each supported filter's configuration surface gains a segmented toggle at the top that flips the filter between its positive and negated form:

- **Value-based filters** (To, From, Workspace, Status, Currency, Purchase currency, Merchant, Exported to) use an is / is not toggle.
- **Presence-based filters** (Has) use a has / doesn't have toggle.

[...]

PR #93909, opened 2026-06-18
Add filter negation feature

bernhardoj:
### Fixed Issues

$ https://github.com/Expensify/App/issues/93208
PROPOSAL:

### Tests

Same as QA Steps

### Offline tests

Same as QA Steps
### QA Steps

1. Go to Spend > Expenses
2. Press the Filters button
3. Open Status, Workspace, To, From, Currency, Purchase currency, Merchant, and "Exported to" filter
4. Verify it has "Filter type" with "is" and "is not" options, and "is" is initially selected
5. Open Has filter
6. Verify it has "Filter type" with "has" and "doesn't have" options, and "has" is initially selected
7. Apply any value and filter type to the filters
9. If a negative filter type is applied (is not/doesn't have), verify the filter chip/bar label is prefixed with `-`
10. Close and reopen the filter
11. Verify the filter value and type are still selected with the correct value
12. Open the save search page
13. Verify it has the same label as in the filter chip/bar

For steps 3-7, if you want more test steps, follow this:
No filter is applied yet
1. Open any filter from the step above
2. Verify the positive filter type is selected
3. Change the filter type to negative
4. Verify the filter is not applied yet (because no value is selected)
5. Apply any value
6. Verify that a negation filter is applied
7. Change the filter type back to positive
8. Verify that a positive filter is applied
9. Unapply and reapply the value
10. Verify that a positive filter is applied

For filter chip/bar:
1. Press any applied filter chip/bar
2. Verify that the filter value and type are selected with the correct value 
3. Play around with the test steps above

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/ecb2754f-82d5-46ea-82e4-2ec27d54dbda

</details>


bernhardoj:
Can I run the translation script myself?

bernhardoj:
Btw, when we apply a negated HAS filter with a Tag value, it will be applied as "Tag: No tag".

https://github.com/user-attachments/assets/781e0dbb-9f8d-45ad-85c2-494a81c97890

It's from this code. It causes weird buggy behavior with the Has filter component, as shown in the video above.

https://github.com/Expensify/App/blob/e5928edcc99c49467503bd6fc16638ecf7f8b9d8/src/libs/SearchQueryUtils.ts#L1110-L1112

Is that expected?

dubielzyk-expensify:
This is looking great already @bernhardoj.  Let's make this text `label` (13px) not `body` (15px)

CleanShot 2026-06-18 at 16 28 42@2x

dubielzyk-expensify:
I'm not convinced we should use the is/is not in the chip itself:

CleanShot 2026-06-18 at 16 29 35@2x

I have to go and will have a noodle, but I think we might want an icon or do `From: not Jon Dubby`. But even that feels weird. I don't think it's crucial for it to be written fully out, but just be able to signal between is/isn't in some way, then when you open the filter you can see it more clearly there anyways

cc @Expensify/design @trjExpensify @JmillsExpensify

bernhardoj:
> Let's make this text label (13px) not body (15px)

Done

JmillsExpensify:
> I'm not convinced we should use the is/is not in the chip itself:
> 
> CleanShot 2026-06-18 at 16 29 35@2x
> I have to go and will have a noodle, but I think we might want an icon or do `From: not Jon Dubby`. But even that feels weird. I don't think it's crucial for it to be written fully out, but just be able to signal between is/isn't in some way, then when you open the filter you can see it more clearly there anyways
> 
> cc @Expensify/design @trjExpensify @JmillsExpensify

I'm fine with what I'm seeing in the adhoc.

JmillsExpensify:
@bernhardoj Looks like we're still missing support for `Status` and `Workspace` in the adhoc. Those are mentioned in the OP. Let me know if I'm missing something!

CleanShot 2026-06-18 at 16 47 06@2x

JS00001:
Status and workspace are still having their BE PRs worked on: https://github.com/Expensify/Auth/pull/22241

bernhardoj:
Yep, still waiting for the BE

JmillsExpensify:
Ah, cool. Thanks!

dubielzyk-expensify:
> I'm fine with what I'm seeing in the adhoc.

As in you're fine with `From is not: Jon`? Curious what @trjExpensify and @Expensify/design thinks

dannymcclain:
I see what you're saying, but `From is not: Jon` doesn't really bother me. It just makes the chip really long which is a minor bummer. Definitely open to seeing some improvement ideas though! But I think we could follow up with them.

trjExpensify:
>I'm not convinced we should use the is/is not in the chip itself:

Yeah, I tend to agree with Jon it's pretty verbose and at times will read weird. I do think there's some value in having an indicator if the filter applied is negated or not without having to click into each pill to confirm that, but could we achieve that more subtly by just putting a `-` before the name in the pill? (I.e `-From: Jon`) or something? Then obviously if you click into it you have the segmented control for `Is | Is not` prominently stated to control it?

dannymcclain:
Yeah I could see doing something more compact like what you're suggesting.

shawnborton:
No strong feelings but I quite like Tom's suggestion, though it's more of a power user thing, but these kinds of negation filters are already power user features.

JmillsExpensify:
I'm fine with the minus if we think `is`/ `is not` is too verbose. I don't feel strongly on this one.

dannymcclain:
>it's more of a power user thing, but these kinds of negation filters are already power user features

Agree. I say we roll with the `-` to get started (i.e. `From: Jon` / `-From: Jon`)

Curious what @dubielzyk-expensify thinks too though.

dubielzyk-expensify:
Yeah I like what Tom suggested . Like Shawn mentions, this is a power-user thing. Most won't see it so as long as we signal it in a minimal way that solves the problem imo. 👍 🚢

src/libs/SearchUIUtils.ts
```diff
@@ -5415,20 +5464,22 @@ function mapFiltersFormToLabelValueList<T extends Record<string, unknown>>(
...
+                const extra = mapper?.(CONST.SEARCH.SYNTAX_FILTER_KEYS.REPORT_FIELD as Exclude<MappedFilterKey, S>) ?? ({} as T);
                 filters.push({key: CONST.SEARCH.SYNTAX_FILTER_KEYS.REPORT_FIELD, label: translate('workspace.common.reportField'), value, ...extra});
             }
             continue;
         }
 
         // Handle regular filters
-        const label = key in FILTER_VIEW_MAP ? FILTER_VIEW_MAP[key as keyof typeof FILTER_VIEW_MAP].labelKey : undefined;
+        const baseKey = removeNegation(key);
+        const labelKey = baseKey in FILTER_VIEW_MAP ? FILTER_VIEW_MAP[baseKey as keyof typeof FILTER_VIEW_MAP].labelKey : undefined;
```
aimane-chnaif: We should hide chips for non-negatable negated filters.

Category as an example:
`removeNegation('categoryNot') = 'category'` is in the map, so a chip now renders, but the category filter is not in `NEGATABLE_FILTERS`, so `getLabelValue` returns the plain positive label:

`getLabelValue: isFilterNegatable('categoryNot') === false → returns translate(labelKey)`

Image


And it causes duplicated filter chips bug.

https://github.com/user-attachments/assets/ec76d40f-c36d-435a-9546-4234a9e3fa67

aimane-chnaif (changes requested):

src/libs/SearchUIUtils.ts
bernhardoj: Fixed

bernhardoj: Fixed

bernhardoj:
Updated filter label to use `-` prefix.

@JS00001 how is the BE going?

Also, can you help run the translation script?

JS00001:
We've merged the first PR, waiting on another cc @c3024

c3024:
@bernhardoj The backend PR required for this PR has been deployed. Could you please check whether the backend responses are as required for this PR?

bernhardoj:
@c3024 can you share how the BE would receive the query?

Currently, this is the query built from the parser of `workspace:0142A68214A45343`, `filters` and `flatFilters` are empty.
```
{
    "type": "expense",
    "status": "",
    "sortBy": "date",
    "sortOrder": "desc",
    "view": "table",
    "policyID": [
        "0142A68214A45343"
    ],
    "filters": null,
    "inputQuery": "workspace:0142A68214A45343",
    "flatFilters": [],
    "isViewExplicitlySet": false,
    "hash": 1260596085,
    "recentSearchHash": 1540228154,
    "similarSearchHash": 260806718
}
```

 Compared to normal filters such as merchant, where the `filters` or `flatFilters` clearly defined the operator (eq or neq).

```
{
    "type": "expense",
    "status": "",
    "sortBy": "date",
    "sortOrder": "desc",
    "view": "table",
    "filters": {
        "operator": "neq",
        "left": "merchant",
        "right": "test"
    },
    "inputQuery": "type:expense sortBy:date sortOrder:desc -merchant:test",
    "flatFilters": [
        {
            "key": "merchant",
            "filters": [
                {
                    "operator": "neq",
                    "value": "test"
                }
            ]
        }
    ],
    "isViewExplicitlySet": false,
    "hash": 1455109608,
    "recentSearchHash": 814580803,
    "similarSearchHash": 1471278054,
    "rawFilterList": [
        {
            "key": "merchant",
            "operator": "neq",
            "value": "test",
            "isDefault": false
        }
    ]
}
```

If I enter `-workspace:0142A68214A45343`, it'll be treated as a keyword.

c3024:
@bernhardoj 

TL;DR: **A top-level field can't be negated**.  Negation (`neq` / "is not") only exists for leaves inside the `filters` AST. `status` and `workspace` have to move *into* the `filters` object so they can support "is not", just like `merchant` does today.

_____________

A couple of key things:

- The BE uses **`policyID`** as **`status`** as the filter keys, not `workspace`. `workspace` is just the user-facing keyword and the [parser](https://github.com/Expensify/App/blob/9f3578cea817279b9e16aca8d7622705844aecac/src/libs/SearchParser/baseRules.peggy#L67) already aliases it to `policyID` (see `baseRules.peggy`), so the leaf key sent to the BE should be `policyID`. `status` keeps its name.
- The shape the BE expects looks just like the merchant example you posted, only with a different `left`:

```json
{ "operator": "eq",  "left": "policyID", "right": "0142A68214A45343" }   // workspace:0142A68214A45343
{ "operator": "neq", "left": "policyID", "right": "0142A68214A45343" }   // -workspace:0142A68214A45343
{ "operator": "eq",  "left": "status",   "right": "drafts" }             // status:drafts
{ "operator": "neq", "left": "status",   "right": "drafts" }             // -status:drafts
```

So `workspace`/`status` need to end up in `filters` / `flatFilters` (with the operator), the same way `merchant` does, not as the top-level `policyID: [...]` / `status` fields, which are positive-only and have no operator to negate.

The reason it currently doesn't work is that in `searchParser.peggy`, `status` and `policyID` are listed under the **`defaultKey`** rule instead of the `key` rule. Default keys get written to the top-level fields and are then filtered out of the AST (`filtersWithoutDefaults`), which is why `filters`/`flatFilters` come back empty and why `-workspace:...` isn't recognized as a negatable filter (it falls through to a keyword). Moving them into the `key` rule should let them be parsed as proper filter leaves with `eq`/`neq` (neither is in `nonNegatableKeys`, so negation will work).

https://github.com/Expensify/App/blob/9f3578cea817279b9e16aca8d7622705844aecac/src/libs/SearchParser/searchParser.peggy#L221-L254

So, [`searchParser.peggy`](https://github.com/Expensify/App/blob/main/src/libs/SearchParser/searchParser.peggy#L221-L254) should be modified so that policyID and status are moved inside `filters` object.

cc: @JS00001

bernhardoj:
Thanks for the clarification. I think I'll prepare a new PR for refactoring the policyID and status filter. I'm currently doing it, but I need another clarification. Filter like policyID is used for some filters, for example, exported to filters.

Exported to filter will show the options scoped to the selected policyID (workspace).
https://github.com/Expensify/App/blob/1eee89197e1fc59fbdfdeb40540f029da6b0e448/src/hooks/useExportedToFilterOptions.ts#L51-L52

But when a user applies a negated policyID, should we "select" the non-selected policyID?

For example, we have policyID A, B, C, and D. The user applies a negated filter and chooses B. Should the exported to filter be scoped to policyID A, C, and D?

@JS00001 @JmillsExpensify

bernhardoj:
@c3024 also, maybe we should remove `status` from search snapshot data too?

image

c3024:
On the `status` in the snapshot, I think we should keep it for now. The frontend uses it to determine whether the search response that's loaded actually matches the search currently on screen.

https://github.com/Expensify/App/blob/6f69123be2da878863f32602ba694c653cdb796a/src/libs/SearchUIUtils.ts#L4688-L4696

So the backend will keep sending `status` in the snapshot. But I just realized that for the negated case this field won't help: the backend only derives `search.status` from an `eq` status filter, so for `-status:...` it comes back empty, meaning it can't distinguish the response for a negated-status search.

Looking at that data-loaded logic more closely, `status` (with `type`) is only used as an identity check to confirm the snapshot belongs to the current query. A cleaner, fully general way to do that is to compare hashes: the snapshot is already keyed by the query `hash`, and the FE already sends `hash` to the backend on every Search request, so the backend can just echo it back in the snapshot's `search` payload, and the FE can compare `searchResults.search.hash === queryJSON.hash`. That covers negated status and every other filter, and it's always present (so no risk of an empty/missing value causing a stuck loading state).

So I think the path is:

1. I'll spin up an Auth PR to include `hash` in the search snapshot so it can be used in the data-loaded check instead of `status`. Auth keeps sending `status` too for now, so production clients don't break.
2. This App PR can update the data-loaded check to compare the hash from the Auth response instead of `status`.
3. Once this FE PR is deployed to production, I'll remove `status` from the search snapshot along with the top-level `jsonQuery` `status` and `policyID` fields that Auth still temporarily accepts/sends anywhere.

What do you think @luacmartins and @JS00001 ?

JS00001:
I think updating to use the hash to compare makes sense!

bernhardoj:
@JS00001 can you also check this [comment](https://github.com/Expensify/App/pull/93909#issuecomment-4808046120), please

JS00001:
> For example, we have policyID A, B, C, and D. The user applies a negated filter and chooses B. Should the exported to filter be scoped to policyID A, C, and D?

Yeah I think we should select the correct policies that should be included

bernhardoj:
@JS00001 does QueueExportSearchToCSV and ExportSearchToCSV still require the `query` param? `query` param is basically a `status` filter. Asking to see if we can just remove it.

JS00001:
@bernhardoj where is 'query' passed? I dont see any reference to it in the BE, is it a part of the jsonQuery?

bernhardoj:
It's passed as a top-level param alongside jsonQuery.

https://github.com/Expensify/App/blob/fb967fe4715b3ccfac2d8809f0ab5f46ec235e1e/src/libs/actions/Search.ts#L1392-L1406

https://github.com/Expensify/App/blob/fb967fe4715b3ccfac2d8809f0ab5f46ec235e1e/src/libs/actions/Search.ts#L1335-L1364

JS00001:
Hmm, yeah it looks like we just use jsonQuery, so we may not need `query` any more

bernhardoj:
Cool, thanks for the confirmation!

JmillsExpensify:
Let me know when we are ready for another round of adhoc product testing.

bernhardoj:
Opened a seperate [PR](https://github.com/Expensify/App/pull/95023) (because it's quite big) to allow workspace and status to be negatable.

cc: @aimane-chnaif

JmillsExpensify:
Works for me.

src/libs/SearchUIUtils.ts
```diff
@@ -5474,52 +5483,81 @@ function getDisplayValue(
...
 
+function getLabelValue(key: SearchAdvancedFiltersKey, labelKey: TranslationPaths | undefined, translate: LocalizedTranslate) {
+    if (!labelKey) {
+        return undefined;
+    }
+
+    if (isFilterNegatable(key)) {
+        const prefix = isFilterNegated(key) ? '-' : '';
+        return `${prefix}${translate(labelKey)}`;
+    }
```
aimane-chnaif: **chip label is now a raw `-` prefix, contradicting the PR's own QA steps**

So an applied negated filter renders as `-Merchant: Uber` / `-Has: Receipt`.
This conflicts with three things:

* **The PR's own QA steps** (steps 8–9) still specify `Merchant is not: test` and `Doesn't have: Receipt`.
* **The picker UI itself** — `NegatableFilter` still labels the toggle buttons "is / is not" and "has / doesn't have", and those strings are still translated across all 9 languages. A user picks "is not" and gets back a chip reading "-Merchant", which doesn't echo the wording they chose.
* **The `-` is hardcoded and unlocalized** — it bypasses `translate` entirely, so the `search.filters.filterType.*.negative` strings are now only used by the toggle, never by the chip.

src/components/Search/FilterComponents/ListFilterContent.tsx
```diff
@@ -213,10 +244,23 @@ function FilterComponents({filterKey, value, type, policyID, selectionListTextIn
                     onChange={onChange}
                 />
             );
+            break;
+        }
         default:
-            return null;
+            break;
```
aimane-chnaif, line 250: `default:` case can return `undefined`.

The switch ends with `default: break;`, then unconditionally `returns <NegatableFilter>{content}</NegatableFilter>`. 
For an unhandled key `content` is `undefined`, and `NegatableFilter` early-returns `children` for non-negatable keys — so the component returns `undefined`.
The switch is currently exhaustive so it's latent, but the old code returned `null` here.

Suggestion: have `NegatableFilter` return `children ?? null`.

src/libs/SearchUIUtils.ts
bernhardoj: > chip label is now a raw - prefix, contradicting the PR's own QA steps

Looks like the AI still sees the old steps 😄

src/components/Search/FilterComponents/ListFilterContent.tsx
bernhardoj: Undefined also renders nothing, same as null

src/libs/SearchQueryUtils.ts
```diff
@@ -2448,6 +2450,27 @@ function isFilterNegatable(key: SearchAdvancedFiltersKey) {
...
+
+function getFilterFormValues<K extends ListFilterContentProps['baseFilterKey'] | TextInputFilterContentProps['baseFilterKey']>(
+    baseFilterKey: K,
+    value: SearchAdvancedFiltersForm[K] | undefined,
+    isNegated: boolean,
+): Partial<SearchAdvancedFiltersForm> {
+    const update: Partial<SearchAdvancedFiltersForm> = {};
+    const negatedFilterKey = `${baseFilterKey}${CONST.SEARCH.NOT_MODIFIER}` as K;
+    if (isFilterNegatable(baseFilterKey)) {
+        update[negatedFilterKey] = isNegated ? value : undefined;
```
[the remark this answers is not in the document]
bernhardoj: Make sense. Updated

src/libs/SearchUIUtils.ts
aimane-chnaif: yup.
Btw non-blocking nit: the `-` is hardcoded in `getLabelValue` rather than coming from a translation, so the chip prefix isn't localized the way the picker toggle ("is not" / "doesn't have") is.

src/components/Search/FilterComponents/AdvancedFilters/SearchAdvancedFiltersContent.tsx
```diff
@@ -6,96 +9,122 @@ import type {SearchAdvancedFiltersForm} from '@src/types/form';
...
+    return (
+        <ListFilter
+            key={baseFilterKey}
+            baseFilterKey={baseFilterKey}
+            value={value}
+            type={values?.type}
+            policyID={getFilterNegatableValue(CONST.SEARCH.SYNTAX_FILTER_KEYS.POLICY_ID, values)}
+            ready={ready}
+            isNegated={isNegated}
+            onChange={(newValue, negated) => onChange(getFilterFormValues(baseFilterKey, newValue, negated))}
```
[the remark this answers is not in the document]
bernhardoj: @JmillsExpensify @JS00001 currently, when we apply a negated Has with Tag value (`-has:tag`), it translates to `{tag: ['none']}`. Do we want to keep that behavior? (this is an existing behavior on prod)


https://github.com/user-attachments/assets/68b37f5e-7cc7-46cb-bbe7-2948787e4a04

Or should we change it so it translates to `{hasNot: tag}`?

src/libs/SearchUIUtils.ts
bernhardoj: Can `-` syntax be localized? 🤔

aimane-chnaif: No need localization but define in CONST?

bernhardoj: Ah, that makes more sense. Updated.

src/components/Search/FilterComponents/AdvancedFilters/SearchAdvancedFiltersContent.tsx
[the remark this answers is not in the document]
luacmartins: That looks correct to me. `none` is a special keyword that looks for empty tags.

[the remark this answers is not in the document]
JmillsExpensify: Agreed, looks correct to me too.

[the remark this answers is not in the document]
bernhardoj: It's not the `none` value of `tag`, but the `has` filter converted to `tag`.


https://github.com/user-attachments/assets/d3339db9-ce04-48d4-aa7c-d9b38272897e

You can see from the video above that when we apply a negated Has with `Tag` value, the filter becomes

```
Tag: No tag | -Has: Receipt, Attachment
```

instead of

```
-Has: Receipt, Attachment, Tag
```

This causes a weird interaction behavior. The `Tag` selection on the Has filter can't be saved because it's always converted to the Tag filter.

I think we should make them independent of each other. WDYT?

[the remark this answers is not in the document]
bernhardoj: @JS00001 @JmillsExpensify @luacmartins can you guys check this once again?

src/libs/SearchUIUtils.ts
```diff
@@ -5481,19 +5490,25 @@ function getDisplayValue(
...
-    filterKey: K,
+function getFilterNegatableValue<K extends ListFilterContentProps['baseFilterKey'] | TextInputFilterContentProps['baseFilterKey']>(
+    baseFilterKey: K,
     values: (Partial<SearchAdvancedFiltersForm> & Partial<Record<`${K}${typeof CONST.SEARCH.NOT_MODIFIER}`, SearchAdvancedFiltersForm[K]>>) | undefined,
 ): {
     isNegated: boolean;
     value: SearchAdvancedFiltersForm[K] | undefined;
 } {
-    const negatedFilterKey = `${filterKey}${CONST.SEARCH.NOT_MODIFIER}` as const;
-    if (!isFilterNegatable(filterKey) || !values || !hasKey(values, negatedFilterKey)) {
```
luacmartins, line 5549: Removing `!isFilterNegatable(filterKey)` can lead to a loop hole where users can type in `-category:Car` for example and land on a negated category query without any visual indicator in the pickers. 

https://github.com/user-attachments/assets/d8aa85b5-7d1a-4fda-a10b-925644984c66

src/libs/SearchQueryUtils.ts
```diff
@@ -2440,14 +2442,33 @@ function addNegation<T extends string>(filterKey: T, isNegated: boolean): T | `$
...
 
 function isFilterNegatable(key: SearchAdvancedFiltersKey) {
     return NEGATABLE_FILTERS.has(removeNegation(key) as SearchNegatableFilterKeys);
 }
 
+function isFilterNegated(filterKey: SearchAdvancedFiltersKey) {
+    return filterKey.endsWith(CONST.SEARCH.NOT_MODIFIER);
+}
+
+function getFilterFormValues<K extends ListFilterContentProps['baseFilterKey'] | TextInputFilterContentProps['baseFilterKey']>(
```
luacmartins, line 2549: Can we add a test for this function? It's pretty central to the negated util.

src/types/form/SearchAdvancedFiltersForm.ts
```diff
@@ -805,7 +805,7 @@ type SearchAdvancedFiltersForm = Form<
         [FILTER_KEYS.ACTION_NOT]: string;
 
         [FILTER_KEYS.HAS]: HasFilterValues;
-        [FILTER_KEYS.HAS_NOT]: string[];
+        [FILTER_KEYS.HAS_NOT]: HasFilterValues;
 
         [FILTER_KEYS.IS]: IsFilterValues;
         [FILTER_KEYS.IS_NOT]: string[];
```
luacmartins: Can we tighten the type for this one too?

src/libs/SearchUIUtils.ts
bernhardoj: Yeah, we discussed it [here](https://github.com/Expensify/App/pull/93909#issuecomment-4986778515) and decide to allow that

src/libs/SearchQueryUtils.ts
bernhardoj: Done

src/types/form/SearchAdvancedFiltersForm.ts
bernhardoj: Done

bernhardoj:
@JS00001 I mean that, changing the `has` filter shouldn't change the `tag` filter.

When we apply `-has:tag`, then it should be translated to `hasNot:[tag]` instead of `tag:[none]`.

@aimane-chnaif fixed

aimane-chnaif (approved):

src/components/Search/FilterComponents/AdvancedFilters/SearchAdvancedFiltersContent.tsx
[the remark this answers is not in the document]
JmillsExpensify: I agree with making them independent, for the reason you stated.

JS00001 (approved):

Issue #82013, opened 2026-02-10
[Due for payment 2026-07-31] [$250] Public Room Deeplink Does Not Open Correctly When User Is Signed Out (Android)


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

**Version Number:** 9.3.16-4
**Reproducible in staging?:** Y
**Reproducible in production?:** Y
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @elirangoshen 
**Slack conversation** (hyperlinked to channel name): [#Open source](https://expensify.slack.com/archives/C01GTK53T8Q/p1770657530367819)

## Action Performed:
1. Sign out of the app
2. Ensure the app is either:
    Running in the foreground, or
    Running in the background
3. Open a deeplink to a public room, for example:
https://staging.new.expensify.com/r/8477877870497982
4. Observe the app behavior.
## Expected Result:
The app should open and navigate directly to the public room.
## Actual Result:
Usually nothing happens, or login screen displayed or the app opens but does not navigate to the public room.
## Workaround:
Unknown
## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [x] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [ ] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/83bc6ed2-2356-4651-93fd-0f33476c31ce

https://github.com/user-attachments/assets/448452b3-1bb1-466b-8b10-a597bde417fc

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: https://www.upwork.com/jobs/~022021267387960460880</li>
[...]

PR #94570, opened 2026-06-25
Fix/82013 public room deeplink signed out

oqildev:
### Explanation of Change

When a signed-out user opened a public room deeplink while the app was already running (or cold-started from the link), the app failed to land on the room — it showed the login screen, or briefly opened the room and then bounced to Concierge / a "Hmm… it's not here" page. The failure was a chain of independent causes, each fixed below:

1. **Report data was dropped before it could be rendered.** `QueuedOnyxUpdates.flushQueue` applies a stale-data filter when there is no current account. For the signed-out deeplink, `OpenReport` returns the anonymous `SESSION` and the `report_*` data in the *same* batch, so the report data was filtered out and navigation had nothing to land on. Fix: bypass the filter only when the batch itself establishes an **anonymous** session (gated on `authTokenType === ANONYMOUS`), so the stale-data protection stays intact for every other flow.

2. **The deeplink was dispatched into a navigator that did not exist yet.** While signed out, `linkingConfig/subscribe` forwarded the report route into the authenticated navigator before it was mounted. Fix: skip that forward for a report deeplink while there is no auth token.

3. **The room lost focus to the default report after auth settled.** Once the anonymous session and `OpenApp` settled, the app re-resolved to the last-accessed report (Concierge for a fresh anonymous user) and overrode the room. Fix: preserve public rooms across `OpenApp`, keep the deeplinked room focused in `ReportsSplitNavigator`, and block the navigate-away — coordinated through a RAM-only `pending public-room deeplink` key whose writes go through an action.

4. **The onboarding guard hijacked the signed-out user.** On current `main`, `OnboardingGuard` redirected the user into the onboarding flow (ending on Concierge). Fix: an onboarding flow can only be completed by a real, authenticated account, so skip it for signed-out / anonymous users (gated on a real, non-anonymous session — real users are unaffected).

5. **The NotFound page flashed for the room during a warm transition.** On current `main`, `ReportNotFoundGuard` declares a report "not found" once loading finishes and the report is momentarily missing. During the anonymous-session/`OpenApp` settle the deeplinked report is transiently absent (a release-build timing race not seen in dev), producing "Hmm… it's not here". Fix: suppress the NotFound page only while the focused report is the pending public-room deeplink (gated on the same RAM-only key — every other report is unaffected).

Changes #4 and #5 touch shared guards but are **gated** so they only alter the signed-out public-room deeplink path; all other flows (real-user onboarding, normal not-found for deleted/inaccessible reports) keep their existing behavior.

A unit test (`tests/actions/QueuedOnyxUpdatesTest.ts`) covers the anonymous-session flushQueue bypass.

### Fixed Issues

$ https://github.com/Expensify/App/issues/82013
PROPOSAL: https://github.com/Expensify/App/issues/82013#issuecomment-4594677367

### Tests

1. Sign out of New Expensify.
2. Keep the app running in the foreground or background (warm), **or** force-stop it (cold).
3. Open a public room deeplink in a browser, e.g. `https://new.expensify.com/r/<publicRoomID>`.
4. Verify the app opens and lands on the **public room** — no login screen, no redirect to Concierge, no "Hmm… it's not here", and the room stays focused (does not bounce away after a second).
5. Repeat for both the warm and cold scenarios above.

Regression checks:
6. **Logged in:** while signed in, open a public room deeplink → it opens normally.
7. **Onboarding (shared guard):** sign up a brand-new account → onboarding is still shown.
8. **Not-found (shared guard):** while signed in, open a deleted / inaccessible report → the "Hmm… it's not here" page is still shown.

### QA Steps

1. Sign out of New Expensify.
2. Keep the app running in the foreground or background (warm), **or** force-stop it (cold).
3. Open a public room deeplink in a browser, e.g. `https://new.expensify.com/r/<publicRoomID>`.
4. Verify the app opens and lands on the **public room** — no login screen, no redirect to Concierge, no "Hmm… it's not here", and the room stays focused (does not bounce away after a second).
5. Repeat for both the warm and cold scenarios above.

Regression checks:
6. **Logged in:** while signed in, open a public room deeplink → it opens normally.
7. **Onboarding (shared guard):** sign up a brand-new account → onboarding is still shown.
8. **Not-found (shared guard):** while signed in, open a deleted / inaccessible report → the "Hmm… it's not here" page is still shown.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/4ad4b9b4-2469-42df-8d57-e4e2fc742f49

</details>

<details>

https://github.com/user-attachments/assets/b61aaf61-4c65-4147-875a-3ef1c86e5fa6

</details>

<details>

https://github.com/user-attachments/assets/cc10318e-5816-4492-9660-80eecfb30828

</details>

<details>

https://github.com/user-attachments/assets/53cf73a1-b9e5-4dbd-b5f6-028ba48fbe13

https://github.com/user-attachments/assets/4ecc8d3e-c162-45c1-b079-b143202e094e

</details>

<details>

https://github.com/user-attachments/assets/95125e93-734a-45fe-b888-eeefdf4cb159

</details>


src/libs/actions/QueuedOnyxUpdates.ts
```diff
@@ -33,25 +34,50 @@ function flushQueue(): Promise<void> {
...
+                typeof sessionValue === 'object' &&
+                sessionValue !== null &&
+                'authToken' in sessionValue &&
+                !!sessionValue.authToken &&
+                'authTokenType' in sessionValue &&
+                sessionValue.authTokenType === CONST.AUTH_TOKEN_TYPES.ANONYMOUS
+            );
+        });
+
+        if (!establishesAnonymousSession) {
```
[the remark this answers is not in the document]
oqildev: Thanks - addressed in 92342e8a8. cleanupSession() now also clears the QueuedOnyxUpdates buffer (previously it cleared persisted requests but not this buffer), so any old-account updates left buffered at sign-out are discarded before a later signed-out public-room deeplink establishes the anonymous session. That guarantees the flushQueue() anonymous-session bypass only ever applies to the freshly-queued OpenReport response, never stale cross-account data. Added a unit test covering the cleared buffer.

trjExpensify (dismissed):
👍

src/libs/actions/QueuedOnyxUpdates.ts
```diff
@@ -33,25 +34,50 @@ function flushQueue(): Promise<void> {
...
-
-        copyUpdates = copyUpdates.filter((update) => preservedKeys.has(update.key as OnyxKey));
+        // FIX #82013: If this queued batch itself establishes the ANONYMOUS session (it contains a SESSION
+        // update carrying an authToken whose authTokenType is anonymous), every other update in the batch
+        // belongs to that newly-established session and cannot be the stale cross-account replay that
+        // #48427/#52822 guard against. This is the signed-out public-room deeplink flow: OpenReport returns
+        // the anonymous SESSION and the report_* collection data in the same batch. Without this bypass the
+        // report data is filtered out and deeplink navigation hangs. We gate on the anonymous authTokenType
+        // (not just any authToken) so the stale-data protection stays intact for every other flow.
+        const establishesAnonymousSession = copyUpdates.some((update) => {
```
situchan, line 83: **The anonymous-session bypass disables the stale-data filter for the *entire* flush batch, not just SESSION/`report_*` keys.**
`establishesAnonymousSession` uses `.some()` over the whole buffer; if it's true, `copyUpdates` is applied with no `preservedKeys` filtering at all. `queueOnyxUpdates` concatenates updates from multiple sources into one buffer, so a stale/cross-account update buffered before the anonymous `OpenReport` response gets merged into the fresh anonymous session unfiltered.
`clear()` on `cleanupSession()` only covers the explicit sign-out path.
Consider keeping the filter but additionally whitelisting the session + `report_*`/`reportActions_*` keys, rather than dropping it wholesale.

**`establishesAnonymousSession` only matches when the SESSION update itself carries `authTokenType === ANONYMOUS`.**
If the server ever delivers the anonymous session as a partial merge that includes `authToken` but omits `authTokenType` (unchanged/split into a prior batch), the check evaluates false, the filter runs, and the co-batched `report_*` data is dropped — reproducing the exact hang this PR fixes.
Please confirm the anonymous `OpenReport` response always co-delivers `authTokenType` with `authToken`.


**The hand-rolled session-shape narrowing duplicates `Session.isAnonymousUser`.**
The `typeof … === 'object' && 'authTokenType' in … === ANONYMOUS` check re-implements existing anonymous-session detection (`Session/index.ts:306`, no circular-dep risk since Session doesn't import QueuedOnyxUpdates).
Only the extra `authToken`-presence check is new; wrapping the shared helper keeps the definition of "anonymous" in one place.

src/libs/Navigation/linkingConfig/subscribe.ts
```diff
@@ -1,12 +1,25 @@
 import type {LinkingOptions} from '@react-navigation/native';
 import {findFocusedRoute} from '@react-navigation/native';
 import {Linking} from 'react-native';
+import Onyx from 'react-native-onyx';
 import continuePlaidOAuth from '@libs/continuePlaidOAuth';
 import navigationRef from '@libs/Navigation/navigationRef';
 import type {RootNavigatorParamList} from '@libs/Navigation/types';
+import {getReportIDFromLink} from '@libs/ReportUtils';
```
situchan: **The new `import {getReportIDFromLink} from '@libs/ReportUtils'` introduces a circular import.** 
Confirmed chain: `ReportUtils.ts:141` imports `linkingConfig`, `linkingConfig/index.ts:9` imports `subscribe`, and `subscribe` now imports `ReportUtils`.
It's runtime-lazy (called inside the `url` handler) so it works today, but it directly contradicts this file's own comment justifying the manual SESSION subscription "to detect a signed-out user without importing the Session action module (which would create a circular dependency through the navigation layer)."
A lighter helper (e.g. reuse `getRouteFromLink`/route parsing without the full `ReportUtils` barrel) avoids the cycle.

src/libs/Navigation/AppNavigator/Navigators/ReportsSplitNavigator.tsx
```diff
@@ -51,6 +62,14 @@ function ReportsSplitNavigator({route}: PlatformStackScreenProps<TabNavigatorPar
             return '';
         }
 
+        // FIX #82013: While a signed-out public-room deeplink is being opened, keep that room focused
+        // instead of defaulting to the last-accessed report (which is Concierge for a fresh anonymous
+        // user). Without this, once OpenApp/auth settle and this navigator re-resolves without the
+        // reportID in its route params, findLastAccessedReport() picks Concierge and overrides the room.
+        if (pendingPublicRoomDeepLinkReportID) {
```
situchan, line 75: **The module-level `pendingPublicRoomDeepLinkReportID` is read synchronously in a `useState` initializer but populated by an async `Onyx.connectWithoutView` callback.**
If this navigator ever mounts before the listener has delivered its first value, the initializer falls through to `findLastAccessedReport()` (Concierge) and — because it's a `useState` initializer — never re-runs, so the wrong initial report is latched for the navigator's lifetime.
The set happens well before the auth tree mounts in the happy path, so this is a timing race rather than a guaranteed failure, but it cannot self-heal.

src/pages/inbox/ReportNavigateAwayHandler.tsx
```diff
@@ -91,6 +92,13 @@ function ReportNavigateAwayHandler() {
 
     // Navigation action that reads non-reactive context (concierge params, modal state, etc.)
     const navigateAwayFromReport = useEffectEvent((prevOnyxReportID: string | undefined, prevParentReportID: string | undefined) => {
+        // FIX #82013: never navigate away from a signed-out public-room deeplink while it is still being
+        // opened. The anonymous OpenApp settling can briefly make this report look removed/closed and
+        // spuriously trigger a navigate-away to Concierge ~tens of seconds after the room is shown
+        // ("room flashes then Concierge"). The pending key is cleared once the user signs in.
+        if (pendingPublicRoomReportID && reportIDFromRoute === pendingPublicRoomReportID) {
```
situchan: **The `reportIDFromRoute === pendingPublicRoomReportID` guard is duplicated across four sites** (twice here, plus `ReportNotFoundGuard.tsx` and `ReportFetchHandler.tsx`) with no shared selector/hook.
A future change to the matching rule (ID normalization, thread/parent handling) must be applied to every copy; missing one silently re-introduces the bug in just that guard.
A single `useIsPendingPublicRoomDeeplink(reportID)` hook would consolidate it.

situchan (commented):
Please pull main and fix conflict

src/libs/actions/Session/index.ts
```diff
@@ -1067,6 +1068,9 @@ function cleanupSession() {
     MainQueue.clear();
     HttpUtils.cancelPendingRequests();
     PersistedRequests.clear();
+    // FIX #82013: also drop any updates buffered in QueuedOnyxUpdates so old-account data cannot ride through
```
mountiny: Can you update the comment to not mention the issue and instead really focus on explaining why we need to do it. If someone is interested they can look up the Pr in blame and find more context

src/libs/actions/Link.ts
```diff
@@ -247,6 +247,11 @@ function openReportFromDeepLink(
     const reportID = getReportIDFromLink(url);
 
     if (reportID && !isAuthenticated) {
+        // FIX #82013: Remember the public-room deeplink reportID so the ReportsSplitNavigator keeps this
```
mountiny: Same

src/libs/actions/Link.ts
```diff
@@ -263,6 +268,14 @@ function openReportFromDeepLink(
             doneCheckingPublicRoom();
         }
     } else {
+        // FIX #82013: not a signed-out public-room deeplink — make sure no stale pending reportID lingers.
```
mountiny: Same

src/libs/actions/Link.ts
```diff
@@ -441,6 +454,14 @@ function getShortLivedAuthTokenURL(setupLink: string): Promise<string> {
         });
 }
 
+/**
+ * Clear the pending signed-out public-room deeplink reportID (#82013). Lives here (an action) so
```
mountiny: same

mountiny: Overall this comment is not super useful here, its established you cannot use onyx.set /merge from components directly

src/libs/actions/QueuedOnyxUpdates.ts
```diff
@@ -35,25 +36,50 @@ function flushQueue(): Promise<void> {
...
-            ONYXKEYS.HAS_LOADED_APP,
-            ONYXKEYS.CREDENTIALS,
-            ONYXKEYS.RAM_ONLY_IS_SIDEBAR_LOADED,
-            ONYXKEYS.ACCOUNT,
-            ONYXKEYS.RAM_ONLY_IS_CHECKING_PUBLIC_ROOM,
-            ONYXKEYS.MODAL,
-            ONYXKEYS.NETWORK,
-            ONYXKEYS.PRESERVED_USER_SESSION,
-        ]);
+        // FIX #82013: If this queued batch itself establishes the ANONYMOUS session (it contains a SESSION
```
mountiny: same

src/libs/actions/QueuedOnyxUpdates.ts
```diff
@@ -35,25 +36,50 @@ function flushQueue(): Promise<void> {
...
+        // update carrying an authToken whose authTokenType is anonymous), every other update in the batch
+        // belongs to that newly-established session and cannot be the stale cross-account replay that
+        // #48427/#52822 guard against. This is the signed-out public-room deeplink flow: OpenReport returns
```
mountiny: same

src/libs/actions/QueuedOnyxUpdates.ts
mountiny: @oqildev Can you make sure to address all reviewer comments please? thank you!

src/libs/actions/QueuedOnyxUpdates.ts
```diff
@@ -62,4 +88,15 @@ function isEmpty() {
     return queuedOnyxUpdates.length === 0;
 }
 
-export {queueOnyxUpdates, flushQueue, isEmpty};
+/**
+ * FIX #82013: Discard any queued updates without applying them. Called from cleanupSession() on sign-out so the
```
mountiny: same

src/libs/Navigation/AppNavigator/Navigators/ReportsSplitNavigator.tsx
mountiny: same here - have you explored this edge case?

src/libs/Navigation/guards/OnboardingGuard.ts
```diff
@@ -54,6 +60,13 @@ Onyx.connectWithoutView({
     },
 });
 
+Onyx.connectWithoutView({
+    key: ONYXKEYS.SESSION,
+    callback: (value) => {
+        session = value;
+    },
+});
```
mountiny, line 69: Can you give a clear `why` comment to explain why we need to use connectWithoutView and cannot use `useOnyx` here?

mountiny (changes requested):
Can you make sure to update all the comments so they really just focus on explaining why the code does what it does and no need to mention the issue  number

src/libs/actions/QueuedOnyxUpdates.ts
oqildev: Thanks, these were helpful. Pushed `be812d83d` for all three:

- I'm not dropping the filter for the whole batch anymore. It stays on, and I only let the `report_/reportActions_/reportMetadata_/reportNameValuePairs_/policy_` and `personalDetailsList` keys through when the batch is the one setting up the anonymous session. Everything else still gets filtered. I also moved the buffer clear into the SESSION listener in this file so it runs on any account loss, not just `cleanupSession()`, which is the gap you mentioned.
- On `authTokenType`: this flow is a fresh anonymous session coming back from `OpenReport`, so it's always a full session object with both `authToken` and `authTokenType`, never a partial merge that splits them. So the check holds here.
- I did try using `Session.isAnonymousUser`, but importing Session into this file pulls in the whole action module and breaks on load (the unit test even fails with a CONFIG error). So I kept a small inline check that does the same thing and left a comment about it. If you'd rather have it centralized I can pull the plain check into a tiny shared helper, just let me know.

src/libs/Navigation/linkingConfig/subscribe.ts
oqildev: Good catch, fixed in `9ed7334e0`. I removed the `ReportUtils` import and match the route with `CONST.REGEX.REPORT_ID_FROM_PATH` instead, so there's no cycle now.

src/libs/Navigation/AppNavigator/Navigators/ReportsSplitNavigator.tsx
oqildev: Yeah, it's a timing thing. The listener gets registered at module import, which runs before this navigator mounts, so in practice the value is already there when the initializer reads it, that's why it doesn't actually break. To make it fully self-healing I'd need either a synchronous read of the Onyx cache (we don't use `tryGetCachedValue` anywhere in the app) or a re-navigate after mount, which can cause a visible flash. Since the real-world risk is basically nil I left it as is, but I'm happy to add the sync cache read if you think it's worth it.

src/pages/inbox/ReportNavigateAwayHandler.tsx
oqildev: Agreed, done in `ed75c6d2c`. I moved it into a `useIsPendingPublicRoomDeeplink(reportID)` hook and used it in both places here and in `ReportNotFoundGuard`, so the check lives in one spot now.

src/libs/actions/Session/index.ts
oqildev: Done in `b34eacac652`. Went through all of them, dropped the issue number and kept each comment focused on the why. Also cut the redundant "route through an action" note and added a reason on `OnboardingGuard's` `connectWithoutView` (it runs outside a component so it can't use `useOnyx`).

src/libs/actions/Link.ts
oqildev: Done, dropped the issue number here.

src/libs/actions/Link.ts
oqildev: Done, dropped the issue number here.

src/libs/actions/QueuedOnyxUpdates.ts
oqildev: Done, dropped the issue number here.

src/libs/actions/QueuedOnyxUpdates.ts
oqildev: Done, dropped the issue number here.

src/libs/actions/QueuedOnyxUpdates.ts
oqildev: Done, dropped the issue number here.

src/libs/actions/Link.ts
oqildev: Good point, removed that part and just left what the function does.

src/libs/Navigation/guards/OnboardingGuard.ts
oqildev: Added a why comment, the guard's `evaluate()` runs in the navigation layer, not inside a component, so it can't use `useOnyx` and needs the module-level read.

oqildev: > **Note:** same here - have you explored this edge case?

@mountiny 

Yeah, this is basically the core of the bug. On a signed-out deeplink the room opens fine, but once OpenApp and auth settle this navigator re-resolves without the `reportID` in its route params, so `findLastAccessedReport()` falls back to Concierge and overrides the room. The pending key is what keeps it focused. I reproduced this on device and verified the fix on all platforms. There's a related timing point about reading the module var in the `useState` initializer, I replied on that thread too - the listener is registered at module import, before this navigator mounts, so it's populated in time in practice.

situchan:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



https://github.com/user-attachments/assets/14f98576-18a9-4bb5-937b-f67758d8e420


</details>

<details>



</details>

<details>


https://github.com/user-attachments/assets/64e2d0b4-6e24-4fed-9814-b47faa13ce9c



</details>

<details>



</details>

<details>


https://github.com/user-attachments/assets/544f4f77-1487-4299-a5df-13a1d6d59e51




</details>

situchan:
Please check ^

situchan:
@oqildev are you able to reproduce original bug in `main`?

oqildev:
> @oqildev are you able to reproduce original bug in `main`?

@situchan 

I'm no longer able to reproduce this bug. I tested it on the Android app.

src/libs/actions/QueuedOnyxUpdates.ts
```diff
@@ -11,10 +12,27 @@ import Onyx from 'react-native-onyx';
 let queuedOnyxUpdates: AnyOnyxUpdate[] = [];
 let currentAccountID: number | undefined;
 
+// Collections that OpenReport returns for a public room the anonymous (signed-out) deeplink is opening.
+// When the flushed batch itself establishes the anonymous session we widen the stale-data allow-list to these keys
+// (plus PERSONAL_DETAILS_LIST) so the room renders, while every unrelated collection stays filtered.
```
tgolen: It would be nice to expand this comment to explain why these keys are here. Or even, why the allow-list needs "widened". As an engineer looking at this, I really don't know why I would add or remove keys from this list.

src/libs/Navigation/AppNavigator/Navigators/ReportsSplitNavigator.tsx
```diff
@@ -18,11 +18,22 @@ import SCREENS from '@src/SCREENS';
...
 
 import React, {useState} from 'react';
+import Onyx from 'react-native-onyx';
 
 const loadReportScreen = () => require<ReactComponentModule>('@pages/inbox/ReportScreen').default;
 const loadSidebarScreen = () => require<ReactComponentModule>('@pages/inbox/sidebar/BaseSidebarScreen').default;
 const Split = createSplitNavigator<ReportsSplitNavigatorParamList>();
 
+// Track the pending signed-out public-room deeplink reportID via a module-level listener
+// (not useOnyx) so it is available synchronously inside the initialReportID useState initializer below.
```
tgolen: Why can't `useOnyx()` be used for this? (please explain that better in the comment).

tgolen (commented):
I've reviewed this PR and I am left with the general feeling that this is a complex PR with several complex fixes. The fixes are complex enough that there are pretty verbose code comments to explain why they exist, but even after trying to understand each comment, I am still not sure I get it.

This usually indicates that there is a deeper problem causing the issue, or that this PR should be split into smaller PRs.

I'm opening an internal conversation with @mountiny @luacmartins and @trjExpensify to discuss my concerns in more detail and see what the best path forward is for this PR.

src/libs/actions/QueuedOnyxUpdates.ts
oqildev: Good call - expanded the comment in 25b6bc7bb. These are the collections OpenReport returns for the room (the report, its actions/metadata/name-value-pairs, its policy, plus personalDetailsList for participants). The stale-data filter would otherwise drop them while there's no account and leave the deeplink stuck, so they're allow-listed to survive it. You'd only add a key here if OpenReport starts returning another collection the room needs to render.

src/libs/Navigation/AppNavigator/Navigators/ReportsSplitNavigator.tsx
oqildev: Expanded the comment in 25b6bc7bb. The initialReportID useState initializer runs on the first render and has to pick the initial report synchronously. useOnyx returns undefined on that first render (its value arrives a render later), which is too late - the initializer would already have fallen back to the last-accessed (Concierge) report. The module-level connectWithoutView keeps the value readable synchronously at init.

src/pages/inbox/ReportNavigateAwayHandler.tsx
```diff
@@ -98,6 +100,13 @@ function ReportNavigateAwayHandler() {
 
     // Navigation action that reads non-reactive context (concierge params, modal state, etc.)
     const navigateAwayFromReport = useEffectEvent((prevOnyxReportID: string | undefined, prevParentReportID: string | undefined) => {
+        // Never navigate away from a signed-out public-room deeplink while it is still being
+        // opened. The anonymous OpenApp settling can briefly make this report look removed/closed and
+        // spuriously trigger a navigate-away to Concierge ~tens of seconds after the room is shown
+        // ("room flashes then Concierge"). The pending key is cleared once the user signs in.
+        if (isPendingPublicRoomDeeplink) {
+            return;
+        }
```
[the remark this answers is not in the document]
oqildev: Good point. The suppression is scoped to the exact reportID of the room being opened anonymously and the key is cleared as soon as a real account signs in, so it only affects that one anonymous public-room session. An anonymous viewer can't leave/close/delete the room themselves, so the main risk is the room being deleted server-side mid-session, which is narrow. I kept the key for the whole anonymous session on purpose - clearing it once the room "settles" reintroduced the flash-then-Concierge race when OpenApp settled later. Happy to add a targeted clear on a genuine leaving/closed update if you'd prefer.

Issue #91408, opened 2026-05-22
Confirm new expenses with a “View” toast on Spend instead of scrolling to and highlighting the row


mountiny:
Coming from [here](https://expensify.slack.com/archives/C08CZDJFJ77/p1778675670519819) and also stemming from [this disucssion](https://github.com/Expensify/App/issues/85521#issuecomment-4328905739)

### Background
When a user creates an expense from the global create FAB (or any flow that lands them on the Spend page), New Expensify confirms the expense was saved by navigating to Reports > Expense, finding the new row in the sorted/grouped FlashList, scrolling to it, and running a Reanimated worklet that fades its background. Producing that visible effect chains together navigation readiness, screen mount, deferred API writes, deferred heavy sort/group work, snapshot delivery, FlashList mount, and an onLayout-driven scrollToIndex — and two unrelated 300ms timers in [useSearchHighlightAndScroll](src/hooks/useSearchHighlightAndScroll.ts) clear the orchestration state regardless of whether any of that has finished. Optimistic rows are placed at the bottom of page 1, then re-placed at their true sorted position once the server snapshot arrives, which routinely falls on a later page. The same logic produces meaningfully different observable behaviour on web, Android, and iOS, because every stage in the chain (navigation timing, InteractionManager, FlashList native engine, Reanimated scheduling) behaves differently per platform — see my case-by-case investigation (cases 1a/1b/2a/2b) in [the issue thread](https://github.com/Expensify/App/issues/85521#issuecomment-4328905739).

### Problem
When a user creates an expense from a flow that lands on the Spend page, if the new row is not present at its final sorted position in the rendered FlashList within ~300ms, then the row is neither scrolled to nor highlighted, leaving the user with no visual confirmation that the expense was created.

### Solution
After the optimistic write is queued, fire a single toast — built on the existing [GrowlNotification](src/components/GrowlNotification/index.tsx) already mounted globally — that says “Expense created.” with a trailing View link, and remove the scroll-and-highlight choreography from Spend on every platform.

[...]

PR #94865, opened 2026-06-29
[No QA] Growl: add action button and loading variant

borys3kk:
### Explanation of Change
Adds dormant capability to the Growl notification with no behavior change: `show`/`success` accept an optional `action` button and supporting `CONST.GROWL` durations, styles, and a reversed theme color. UI is extracted into `GrowlNotificationContent` so the shell only mounts content while visible. Groundwork for an upcoming "Expense added" growl with a "View" deep link.

### Fixed Issues
PROPOSAL:

### Tests
N/A - we do not use the growls anywhere else in the app (NO QA)

### Offline tests
N/A — this change only modifies the in-app notification component; it adds no API calls or Onyx operations. Existing growls behave identically offline and online.

### QA Steps
Same as tests.

// TODO: These must be filled out, or the issue title must include "[No QA]." — this PR is **[No QA]** (no user-facing change).

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/components/GrowlNotification/GrowlNotificationContainer/index.native.tsx
```diff
@@ -5,11 +5,11 @@ import useStyleUtils from '@hooks/useStyleUtils';
 import useThemeStyles from '@hooks/useThemeStyles';
 import type GrowlNotificationContainerProps from './types';
 
-function GrowlNotificationContainer({children, translateY}: GrowlNotificationContainerProps) {
+function GrowlNotificationContainer({children, progress, inactiveY}: GrowlNotificationContainerProps) {
```
JakubKorytko, line 10: native container destructures away `useBottomPosition` (it's optional in the type). does that mean bottom-positioned growls are intentionally unsupported on native? if so, what happens on a tablet in landscape (I believe we support landscape already or will support soon) - does the growl still come from the top?

src/components/GrowlNotification/GrowlNotificationContainer/index.tsx
```diff
@@ -4,10 +4,14 @@ import useResponsiveLayout from '@hooks/useResponsiveLayout';
 import useThemeStyles from '@hooks/useThemeStyles';
 import type GrowlNotificationContainerProps from './types';
 
-function GrowlNotificationContainer({children, translateY}: GrowlNotificationContainerProps) {
+function GrowlNotificationContainer({children, progress, inactiveY, useBottomPosition}: GrowlNotificationContainerProps) {
     const styles = useThemeStyles();
     const {shouldUseNarrowLayout} = useResponsiveLayout();
-    const animatedStyles = useAnimatedStyle(() => styles.growlNotificationTranslateY(translateY));
+    const animatedStyles = useAnimatedStyle(() => ({transform: [{translateY: inactiveY * (1 - progress.get())}]}), [inactiveY]);
```
JakubKorytko, line 14: the inline object `{transform: [{translateY: ...}]}` inside `useAnimatedStyle` is fine for Reanimated worklets, but `progress` isn't listed in the deps array - that's correct because it's a `SharedValue` (not a React value), but it could confuse future readers. a small comment might help.

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+import {useMemoizedLazyExpensifyIcons} from '@hooks/useLazyAsset';
+import useResponsiveLayout from '@hooks/useResponsiveLayout';
+import useTheme from '@hooks/useTheme';
+import useThemeStyles from '@hooks/useThemeStyles';
+import type {GrowlAction, GrowlType} from '@libs/Growl';
+import CONST from '@src/CONST';
+import type IconAsset from '@src/types/utils/IconAsset';
+import GrowlNotificationContainer from './GrowlNotificationContainer';
+
+const INACTIVE_OFFSET = 255;
```
JakubKorytko: where does 255 come from? could it live in `CONST.GROWL` alongside the other values, or at least have a comment explaining the choice?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+const INACTIVE_POSITION_Y = -INACTIVE_OFFSET;
+// Approximate time for the slide spring to settle. Used to delay onDismissed
+// until the slide-out animation has visibly finished.
+const SLIDE_DURATION_MS = 400;
```
JakubKorytko: seems to me like it is a guess/approximation of the spring duration - if the spring config changes, this value silently goes stale. could we use a `withSpring` callback or `onFinish` instead of a hardcoded timeout? also, should this live in `CONST` alongside other `GROWL` timing values?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    // Every non-loading growl variant must have an icon mapping. Loading is rendered separately
+    // (ActivityIndicator), so it's excluded here — keeping this map exhaustive over the rest.
+    type GrowlIconTypes = Record<
+        Exclude<GrowlType, typeof CONST.GROWL.LOADING>,
+        {
+            icon: React.FC<SvgProps> | IconAsset;
+            iconColor: string;
+        }
+    >;
```
JakubKorytko: defining a type inside the component body is unusual - it works but it gets re-evaluated every render. could this be hoisted to module scope next to the props type?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+        [CONST.GROWL.WARNING]: {
+            icon: icons.Exclamation,
+            iconColor: theme.warning,
+        },
+    };
+
+    /**
+     * Animate growl notification. `targetProgress` is 0 for offscreen, 1 for visible.
+     */
+    const fling = useCallback(
```
JakubKorytko: React Compiler is enabled and the file compiles - do we still need `useCallback` here? the `'worklet'` directive might require it, but worth double-checking as I am not sure

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+                }),
+            );
+        },
+        [progress],
+    );
+
+    /**
+     * Slide the growl off-screen and schedule its unmount via the dismissNonce effect.
+     */
+    const triggerDismiss = useCallback(() => {
```
JakubKorytko: same question - is `useCallback` needed here with React Compiler enabled?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+        }
+        const timeoutId = setTimeout(onDismissed, SLIDE_DURATION_MS);
+        dismissTimeoutRef.current = timeoutId;
+        return () => {
+            clearTimeout(timeoutId);
+            dismissTimeoutRef.current = null;
+        };
+    }, [dismissNonce, onDismissed]);
+
+    useEffect(() => {
```
JakubKorytko, line 113: this effect does 4 things at once: resets the pressed guard, cancels pending dismiss timers, snaps progress to 0, slides in, and schedules auto-dismiss. that's a lot of responsibilities for one effect - would a custom hook like `useGrowlLifecycle` make the intent clearer?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+            clearTimeout(dismissTimeoutRef.current);
+            dismissTimeoutRef.current = null;
+        }
+
+        // Snap to fully offscreen before sliding in so the slide-in direction matches the
+        // current placement (from above for top, from below for bottom).
+        progress.set(0);
+        fling(1);
+
+        if (duration <= 0) {
```
JakubKorytko: `<= 0` means any negative duration also skips auto-dismiss - is that intentional, or should this be `=== 0` to only target the loading variant?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+                .direction(useBottomPosition ? Directions.DOWN : Directions.UP)
+                .runOnJS(true)
+                .onStart(() => {
+                    triggerDismiss();
+                }),
+        [useBottomPosition, triggerDismiss],
+    );
+
+    return (
+        <View style={[styles.growlNotificationWrapper]}>
```
JakubKorytko: single-element array `[styles.growlNotificationWrapper]` - the array wrapper is unnecessary here

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+        fling(1);
+
+        if (duration <= 0) {
+            // Indefinite (loading) growl - slide in but don't auto-dismiss.
+            return;
+        }
+
+        const autoDismissTimeoutId = setTimeout(triggerDismiss, duration);
+        return () => clearTimeout(autoDismissTimeoutId);
+    }, [bodyText, type, action, duration, fling, progress, triggerDismiss]);
```
JakubKorytko: potential regression - if `show()` is called again with the SAME text/type/duration while the growl is already visible (e.g. user taps "Copy" twice quickly), none of these deps change, so the effect won't re-fire and the auto-dismiss timer won't reset. The old code handled this by clearing `duration` to undefined after dismiss, so subsequent identical calls would still trigger. Consider adding a nonce/counter to the content to guarantee re-trigger.

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                    accessibilityLabel={action.label}
+                                    sentryLabel="GrowlNotification-Action"
+                                    onPress={() => {
+                                        if (isActionPressedRef.current) {
+                                            return;
+                                        }
+                                        isActionPressedRef.current = true;
+                                        triggerDismiss();
+                                        action.onPress();
+                                    }}
```
JakubKorytko: `triggerDismiss()` is called before `action.onPress()`. if `onPress` navigates away, the component unmounts, and the `SLIDE_DURATION_MS` `setTimeout` in the `dismissNonce` effect fires on an unmounted component. the effect cleanup should handle it, but worth verifying there's no "setState on unmounted" warning.

src/components/GrowlNotification/index.tsx
```diff
@@ -1,151 +1,55 @@
...
-     * @param {String} type
-     * @param {Number} duration
-     */
-    const show = useCallback((text: string, growlType: string, growlDuration: number) => {
-        setBodyText(text);
-        setType(growlType);
-        setDuration(growlDuration);
+    const show = useCallback((text: string, growlType: GrowlType, duration: number, action?: GrowlAction) => {
+        setContent({bodyText: text, type: growlType, duration, action});
     }, []);
```
JakubKorytko: if `show()` is called twice with identical args while the growl is already visible, `setContent` creates a new object but the child receives the same primitive prop values, so the child's content-change effect won't re-fire. consider adding a monotonic counter  (e.g. `showNonce`) to the content object so each call is guaranteed unique.

Also: `useCallback` - needed with React Compiler?

src/components/GrowlNotification/index.tsx
```diff
@@ -1,151 +1,55 @@
...
-            return;
-        }
-
-        fling(0);
-        setTimeout(() => {
-            fling();
-            setDuration(undefined);
-        }, duration);
-    }, [duration, fling]);
+    const handleDismissed = useCallback(() => {
```
JakubKorytko: `useCallback` here - React Compiler should handle this automatically

src/CONST/index.ts
```diff
@@ -3707,8 +3707,11 @@ const CONST = {
         SUCCESS: 'success',
         ERROR: 'error',
         WARNING: 'warning',
+        LOADING: 'loading',
         DURATION: 2000,
         DURATION_LONG: 3500,
+        // Longer duration for growls with an actionable button (e.g. "View"), giving the user enough time to tap it.
+        DURATION_WITH_ACTION: 6000,
```
JakubKorytko, line 3763: PR description says this is groundwork for an upcoming growl - but since it's unused right now, could the growl component auto-apply this duration when an action is present, instead of relying on each caller to remember to pass it manually?

src/libs/Growl.ts
```diff
@@ -40,16 +48,24 @@ function error(bodyText: string, duration: number = CONST.GROWL.DURATION) {
 /**
  * Show success growl
  */
-function success(bodyText: string, duration: number = CONST.GROWL.DURATION) {
-    show(bodyText, CONST.GROWL.SUCCESS, duration);
+function success(bodyText: string, duration: number = CONST.GROWL.DURATION, action?: GrowlAction) {
+    show(bodyText, CONST.GROWL.SUCCESS, duration, action);
+}
```
JakubKorytko: `success()` now accepts `action` but `error()` above doesn't - is that intentional? if we ever need an error growl with an action (e.g. "Retry"), callers would have to bypass the helper and call `show()` directly. might be worth keeping the API symmetrical.

src/libs/Growl.ts
```diff
@@ -40,16 +48,24 @@ function error(bodyText: string, duration: number = CONST.GROWL.DURATION) {
...
+
+/**
+ * Show indefinite loading growl (no auto-dismiss). Call success/error/show again to replace it.
+ */
+function loading(bodyText: string) {
+    show(bodyText, CONST.GROWL.LOADING, 0);
 }
```
JakubKorytko: `loading()` hardcodes `duration=0` to mean "indefinite". that's a bit implicit - what if someone calls `show()` with `duration=0` not realizing it means "never dismiss"? also, there's no way to dismiss a loading growl programmatically other than replacing it with another growl. would an explicit `dismiss()` method be useful?

src/styles/index.ts
```diff
@@ -3357,23 +3357,56 @@ const staticStyles = (theme: ThemeColors) =>
             ...positioning.pFixed,
         },
 
+        growlNotificationContainerBottomRight: {
+            maxWidth: variables.sideBarWidth,
+            width: '100%',
+            right: 0,
+            bottom: 20,
```
JakubKorytko: magic number. should this use a spacing variable or be documented with a comment explaining where `20` comes from?

src/styles/index.ts
```diff
@@ -3357,23 +3357,56 @@ const staticStyles = (theme: ThemeColors) =>
...
             ...FontUtils.fontFamily.platform.EXP_NEUE,
-            width: '90%',
             lineHeight: variables.fontSizeNormalHeight,
             color: theme.textReversed,
-            ...spacing.ml4,
+        },
+
+        growlNotificationTextWithoutAction: {
+            width: '90%',
+        },
```
JakubKorytko: `width: '90%'` is carried over from the old code - but why 90%? it's a magic percentage that was likely meant to leave room for the icon. with the new `gap3` spacing,  wouldn't `flex: 1; minWidth: 0` (like the action variant) work better here too?

src/styles/index.ts
```diff
@@ -3357,23 +3357,56 @@ const staticStyles = (theme: ThemeColors) =>
...
             boxShadow: `${theme.shadow}`,
-            ...spacing.p5,
+            // Spec: padding-left 16px, padding-top/right/bottom 8px, 12px gap between items.
+            // The 8px top/bottom padding + the 40px-tall Medium Link action button add up to the
+            // 56px growl height from the design.
+            ...spacing.pl4,
+            ...spacing.pt2,
+            ...spacing.pr2,
+            ...spacing.pb2,
+            ...spacing.gap3,
```
JakubKorytko: the old style was `...spacing.p5` (20px all sides). this new padding (16/8/8/8) applies to ALL growl types. there are 3 existing `Growl.error()` calls (Onfido flows) that would look visually tighter now - rare, but worth a quick visual check.

JakubKorytko (commented):
nice one but some comments

src/components/GrowlNotification/GrowlNotificationContainer/index.native.tsx
borys3kk: yes, they're intentionally unsupported on the native as we agreed earlier that on narrow layout (ie native apps) we only show the growl docked at the top, on ipad in landscape mode we use narrow layout that's been adjusted to account for more space, but nonetheless it's still narrow layout so it' in line with the decision

src/components/GrowlNotification/GrowlNotificationContainer/index.tsx
borys3kk: added a comment explaining it

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: added a comment explaining it and moved it to const

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: good call - replaced the hardcoded `SLIDE_DURATION` with `withSpring` completion callback

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: good catch, moved it

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: yeah we can drop it for now, according to the reanimated docs it's not necessary (but in case it will not work in the actual pr we will bring it back in the next pr)

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: it's redundant, removed it

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: after refactor from [this](https://github.com/Expensify/App/pull/94865#discussion_r3497361240) comment this effect does is one cohesive lifecycle step - when new growl content arrives reset the guard, slide in the growl and arm auto-dismiss, so I would leave it as is

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: intentional <=0 treats any non-positive duration as "indefinite", added explanation comment as well

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: done

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: verified - after the change to use the withSpring's there's no JS dismiss timer anymore. onDismissed only sets the state on the parent GrowlNotification (app-root, never unmounts on navigation) - GrowlNotificationContent unmounting is the result of that callback, so a late fire is a harmless no-op. Keeping triggerDismiss() before onPress() so the slide-out starts before any navigation.

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: Added a nonce (just in case) to the growl content (bumped per show()) and used it as key, so identical-args re-shows remount and re-trigger

src/components/GrowlNotification/index.tsx
borys3kk: same as above with nonce, removing useCallback as it should not matter in this context but will see when we wire the functionality in the next PR

src/components/GrowlNotification/index.tsx
borys3kk: also removing it

src/CONST/index.ts
borys3kk: yeah, good idea, changed it now

src/libs/Growl.ts
borys3kk: made the api symmetrical for consistency

src/libs/Growl.ts
borys3kk: added comment explaining the "0" meaning, the dismiss() method needs real parent child wiring and loading() has no callers yet, so I'll add it in the PR that needs it.

src/styles/index.ts
borys3kk: not a magic number, it mirrors the previous top-positioned growl (previous top:20, here bottom:20)

src/styles/index.ts
borys3kk: good catch, the design changed how to growls looked a bit so divided this into two, the old growls stay the same and the new action growls will have this more compact style

src/styles/index.ts
borys3kk: changed it according to your suggestion, tested in app and it looks ok (added temporary test buttons to just test this)

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const theme = useTheme();
+    const styles = useThemeStyles();
+    const {shouldUseNarrowLayout} = useResponsiveLayout();
+    const icons = useMemoizedLazyExpensifyIcons(['Exclamation', 'Checkmark']);
+
+    // Derived live so that resizing the window flips position + slide direction during display.
+    const useBottomPosition = !!action && !shouldUseNarrowLayout;
+    const inactiveY = useBottomPosition ? INACTIVE_OFFSET : INACTIVE_POSITION_Y;
+
+    const types: GrowlIconTypes = {
```
JakubKorytko, line 69: This icon map is rebuilt every render (depends on `icons` + `theme`). React Compiler likely caches it, but since the values only change on theme switch, could this live outside the component as a factory function called once per theme to be extra safe?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+        [CONST.GROWL.WARNING]: {
+            icon: icons.Exclamation,
+            iconColor: theme.warning,
+        },
+    };
+
+    /**
+     * Animate growl notification. `targetProgress` is 0 for offscreen, 1 for visible.
+     */
+    const fling = (targetProgress = 0, onSettled?: () => void) => {
```
JakubKorytko, line 87: `fling` and `triggerDismiss` are both plain closures listed in the useEffect deps (line 119). React Compiler memoizes them, but does the `'worklet'` directive survive that transform? If the Reanimated babel plugin runs after the compiler wraps the function, the worklet annotation might be lost → silent JS-thread fallback → spring could stutter on native. Worth a quick sanity-check on a device with Reanimated's "strict mode" enabled.

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    useEffect(() => {
+        isActionPressedRef.current = false;
+
+        // Snap to fully offscreen before sliding in so the slide-in direction matches the
+        // current placement (from above for top, from below for bottom).
+        progress.set(0);
+        fling(1);
+
+        if (duration <= 0) {
```
JakubKorytko: If a loading growl is shown and the caller never follows up with `success`/`error` (e.g. a caught exception swallows the follow-up call), this growl stays on screen forever. Would a safety-net max timeout (say 30s) be worth adding as a fallback?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+        const autoDismissTimeoutId = setTimeout(triggerDismiss, duration);
+        return () => clearTimeout(autoDismissTimeoutId);
+    }, [bodyText, type, action, duration, fling, progress, triggerDismiss]);
+
+    // GestureDetector by default runs callbacks on UI thread using Reanimated. In this
+    // case we want to trigger an RN's Animated animation, which needs to be done on JS thread.
+    // Wrapped in useMemo so the React Compiler doesn't flag the gesture builder's internal
+    // mutable state as ref access during render.
+    const flingGesture = useMemo(
```
JakubKorytko: `useMemo` is used here despite React Compiler being enabled. The comment explains why (gesture builder's internal mutable state), but does the compiler's own caching not already handle this?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+        <View style={styles.growlNotificationWrapper}>
+            <GrowlNotificationContainer
+                progress={progress}
+                inactiveY={inactiveY}
+                useBottomPosition={useBottomPosition}
+            >
+                <PressableWithoutFeedback
+                    accessibilityLabel={bodyText}
+                    sentryLabel="GrowlNotification-Dismiss"
+                    onPress={() => triggerDismiss()}
```
JakubKorytko: nit: the arrow wrapper `() => triggerDismiss()` is equivalent to just `triggerDismiss` - the extra closure allocates for no reason.

src/libs/Growl.ts
```diff
@@ -21,35 +29,45 @@ function setIsReady() {
...
-function success(bodyText: string, duration: number = CONST.GROWL.DURATION) {
-    show(bodyText, CONST.GROWL.SUCCESS, duration);
+function success(bodyText: string, duration?: number, action?: GrowlAction) {
+    show(bodyText, CONST.GROWL.SUCCESS, duration, action);
+}
+
+/**
+ * Show indefinite loading growl (no auto-dismiss). Call success/error/show again to replace it.
+ */
+function loading(bodyText: string) {
```
JakubKorytko: There's no programmatic `dismiss()` API - the only way to remove a loading growl is to call `show`/`success`/`error`. If a consumer forgets (or the follow-up call is in a catch block that never fires), the growl is stuck. Would a `Growl.dismiss()` escape hatch be useful?

src/CONST/index.ts
```diff
@@ -3704,8 +3704,16 @@ const CONST = {
...
+        // Sentinel duration meaning "never auto-dismiss" (e.g. loading growls that stay until replaced).
+        DURATION_INDEFINITE: 0,
```
JakubKorytko: Using `0` as a sentinel is a bit unusual - it's a valid numeric value that could be confused with "instant dismiss" in code that does `setTimeout(fn, duration)`. The `GrowlNotificationContent` guards with `duration <= 0`, but a caller might pass `DURATION_INDEFINITE` directly to setTimeout elsewhere and get a 0ms fire. Would a negative sentinel (e.g. `-1`) or `Infinity` be less ambiguous?

src/styles/index.ts
JakubKorytko: Understood it mirrors the existing inset - but since they must stay in sync, wouldn't a shared constant make that coupling explicit rather than relying on a code comment?

JakubKorytko (commented):
second round my man, nothing serious though

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: right know it's smoke and mirrors whether it will work nicely, when this pr is merged and the second one will be under review, then we will be able to check that, I will keep that in mind though!

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: The map can't really live at module scope - both inputs are component-only : `icons` comes from the `useMemoizedLazyExpensifyIcons` hook and `iconColor` from `useTheme()`. A factory would still run every render with those as args, so it's identical work to the inline literal, just behind a call. And React Compiler already memoizes this (as you noted). I'd leave it as-is.

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: Removed it for now, if something breaks in the second PR I will bring it back

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: done

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: removing the whole loading variant as it's not used in the final PR

src/libs/Growl.ts
borys3kk: same as above (removing the loading variant entirely)

src/CONST/index.ts
borys3kk: it was connected with the loading growl variant which is being removed (so removing this dead code as well)

src/styles/index.ts
borys3kk: moved it to shared var for readbility

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+        isActionPressedRef.current = false;
+
+        // Snap to fully offscreen before sliding in so the slide-in direction matches the
+        // current placement (from above for top, from below for bottom).
+        progress.set(0);
+        fling(1);
+
+        const autoDismissTimeoutId = setTimeout(triggerDismiss, duration);
+        return () => clearTimeout(autoDismissTimeoutId);
+    }, [bodyText, type, action, duration, fling, progress, triggerDismiss]);
```
JakubKorytko: `fling` and `triggerDismiss` are recreated every render but listed in the effect deps. This component subscribes to theme/responsive hooks, so a layout or theme update while the growl is visible could re-run this effect, restart the slide-in, and reset the auto-dismiss timer. Would it be safer to drop those two from deps (with an eslint comment) or route dismiss through refs?

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+    // is still on screen. Reset whenever new growl content arrives.
+    const isActionPressedRef = useRef(false);
+
+    const theme = useTheme();
+    const styles = useThemeStyles();
+    const {shouldUseNarrowLayout} = useResponsiveLayout();
+    const icons = useMemoizedLazyExpensifyIcons(['Exclamation', 'Checkmark']);
+
+    // Derived live so that resizing the window flips position + slide direction during display.
+    const useBottomPosition = !!action && !shouldUseNarrowLayout;
```
JakubKorytko, line 66: `useBottomPosition` and `inactiveY` are derived live from `shouldUseNarrowLayout`. If the user resizes the window while a growl is on screen, could it jump between top and bottom anchoring mid-display? If placement should be fixed at show time, consider snapshotting these values when the growl mounts.

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+        });
+
+    return (
+        <View style={styles.growlNotificationWrapper}>
+            <GrowlNotificationContainer
+                progress={progress}
+                inactiveY={inactiveY}
+                useBottomPosition={useBottomPosition}
+            >
+                <PressableWithoutFeedback
```
JakubKorytko: the action `Button` is nested inside a dismiss `PressableWithoutFeedback`. `isNested` helps on native, but on mWeb could a click on the action also bubble to the outer pressable and double-fire `triggerDismiss`? Also, the outer pressable only exposes `accessibilityLabel={bodyText}`, so screen readers may not announce the action at all.

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const autoDismissTimeoutId = setTimeout(triggerDismiss, duration);
+        return () => clearTimeout(autoDismissTimeoutId);
+    }, [bodyText, type, action, duration, fling, progress, triggerDismiss]);
+
+    // GestureDetector by default runs callbacks on UI thread using Reanimated. In this
+    // case we want to trigger an RN's Animated animation, which needs to be done on JS thread.
+    const flingGesture = Gesture.Fling()
+        .direction(useBottomPosition ? Directions.DOWN : Directions.UP)
+        .runOnJS(true)
+        .onStart(() => {
```
JakubKorytko: nit: `.onStart(() => { triggerDismiss(); })` can be `.onStart(triggerDismiss)` since `runOnJS(true)` already bridges to the JS thread

JakubKorytko (approved):
tbh this reads much cleaner. A few lifecycle and interaction edge cases stood out that are worth sanity-checking before the follow-up PR wires the action growl for real - approving since these are NAB's, good job!

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: Good catch - fixed by going one step further, the effect is now mount only. The parent remounts this component via key={nonce} on every show(), so slide-in + timer arm exactly once per growl and theme/layout re-renders can't reset them. Also dropped the now-dead isActionPressedRef reset and progress.set(0) snap (fresh mount defaults).

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: Intentional — if we snapshot at mount and the user crosses the breakpoint mid-display, the growl is stuck at an anchor that's wrong for the new layout (e.g. bottom-right on narrow, where it overlaps the compose area). The progress + inactiveY normalization exists exactly so a live flip keeps the position and the slide-out direction correct. Given growls last 2–6s, a rare one-frame reposition beats a wrongly-anchored growl.

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: Restructured — the Button is now a sibling of the dismiss pressable (dismiss target covers only icon + text), so nothing can bubble and the action is independently focusable/announced by screen readers. Dropped isNested.

src/components/GrowlNotification/GrowlNotificationContent.tsx
borys3kk: applied it

situchan:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



https://github.com/user-attachments/assets/f0e42a82-e312-41be-9af1-07d9c604d0fa


</details>

<details>



</details>

<details>


https://github.com/user-attachments/assets/e1eab532-6b58-444a-97c7-01c905d412d7



</details>

<details>



</details>

<details>


https://github.com/user-attachments/assets/30349009-d8fa-43f6-9888-0078fc94baf8



</details>

JmillsExpensify (commented):
No product review required.

src/components/GrowlNotification/GrowlNotificationContainer/index.native.tsx
```diff
@@ -7,11 +7,13 @@ import Animated, {useAnimatedStyle} from 'react-native-reanimated';
...
     const styles = useThemeStyles();
     const StyleUtils = useStyleUtils();
     const insets = useSafeAreaInsets();
-    const animatedStyles = useAnimatedStyle(() => styles.growlNotificationTranslateY(translateY));
+    // `progress` is a SharedValue read via .get() inside the worklet, so it's intentionally
+    // omitted from the deps array — only the plain JS value `inactiveY` needs to invalidate the style.
+    const animatedStyles = useAnimatedStyle(() => ({transform: [{translateY: inactiveY * (1 - progress.get())}]}), [inactiveY]);
```
situchan, line 16: **native container ignores `useBottomPosition`, so an action growl never hides on native tablets.** 

The web container branches on `useBottomPosition` to anchor bottom-right, but the native container destructures only `{children, progress, inactiveY}` and always renders the top-anchored `growlNotificationContainer`.
When `useBottomPosition` is true (`!!action && !shouldUseNarrowLayout` — reachable on iPads), `inactiveY = +OFFSCREEN_OFFSET (+255)`, so the "offscreen" resting transform is `+255 * (1 - progress)` measured *down from `top: 20`* — i.e. the growl rests ~275px onscreen instead of hidden, and the fling-to-dismiss direction (`DOWN`) points the wrong way.
Since this PR is the explicit groundwork for an action/"View" growl, wiring that up will misrender on native tablets.
The native container needs the same `useBottomPosition` handling (or the type should forbid the bottom anchor on native).

src/styles/index.ts
```diff
@@ -3363,23 +3363,57 @@ const staticStyles = (theme: ThemeColors) =>
...
+        },
+
+        // Compact padding sized around the 40px-tall action button: padding-left 16px,
+        // top/right/bottom 8px (8 + 40 + 8 = 56px growl height from the design).
+        growlNotificationBoxWithAction: {
+            ...spacing.pl4,
+            ...spacing.pt2,
+            ...spacing.pr2,
+            ...spacing.pb2,
+        },
```
situchan, line 3394: **existing error growls get a 4px-narrower icon↔text gap, contradicting the "no behavior change" claim.** 

The old `growlNotificationText` used `...spacing.ml4` (marginLeft 16px) between icon and text.
The new layout drops that and relies on `...spacing.gap3` (gap 12px) on the box/pressable.
So the currently-live Onfido error growls render with icon-to-text spacing reduced 16px → 12px.
Minor, but the PR/offline notes assert existing growls "behave identically," so worth a deliberate confirm that the tighter spacing is intended.

situchan (commented):
NIT: The PR body advertises a `loading()` indefinite variant, but no `loading` function is added to `src/libs/Growl.ts` in this diff.

src/components/GrowlNotification/index.tsx
```diff
@@ -1,157 +1,62 @@
...
-        }
-
-        fling(0);
-        setTimeout(() => {
-            fling();
-            setDuration(undefined);
-        }, duration);
-    }, [duration, fling]);
+    const handleDismissed = () => {
+        setContent(null);
```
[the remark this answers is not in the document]
borys3kk: good catch, implemented

borys3kk:
> NIT: The PR body advertises a `loading()` indefinite variant, but no `loading` function is added to `src/libs/Growl.ts` in this diff.

removed it from the pr body as it's no longer implemented

situchan (approved):
Above review comments are not blockers but let's make sure to test tablet landscape in implementation PR

src/components/GrowlNotification/GrowlNotificationContainer/index.native.tsx
borys3kk: I've checked it in the simulator and the native tablets (at least the ipads simulators) use explicitly the narrow layout with some adjustments to make them easier to use (so we don't need bottomPosition there but I will keep that in mind for the implementation PR)

borys3kk:
> 🤖 **Code review**
> 
> Solid, well-commented refactor. Splitting the visible UI into `GrowlNotificationContent` so the shell no longer subscribes to theme/responsive/navigation contexts is a genuine improvement — the shell won't re-render on every screen/focus change, and the growl DOM is fully unmounted between shows. The nonce-guarded dismissal correctly handles the stale slide-out race (a replacement growl remounts with its own shared value, and the parent ignores dismissals whose nonce no longer matches). No behavior change for the existing (action-less) growls, which matches the PR's "dormant groundwork" intent.
> 
> A few things worth addressing before this lands:
> 
> **1. Native container ignores `useBottomPosition` → latent bug for action growls on wide native layouts (iPad).** [`GrowlNotificationContent.tsx:66`](https://github.com/Expensify/App/blob/d9bb03d367532cbf4457fb2dfcf26b0928ecbe88/src/components/GrowlNotification/GrowlNotificationContent.tsx#L66) derives `useBottomPosition = !!action && !shouldUseNarrowLayout` and, when true, sets `inactiveY` to the **positive** (bottom) offset. But the native container [`GrowlNotificationContainer/index.native.tsx:10`](https://github.com/Expensify/App/blob/d9bb03d367532cbf4457fb2dfcf26b0928ecbe88/src/components/GrowlNotification/GrowlNotificationContainer/index.native.tsx#L10) dropped the `useBottomPosition` param and always renders the top-anchored container. On a wide native layout (e.g. iPad, where `shouldUseNarrowLayout` can be `false`), an action growl would get `inactiveY = +255` while being top-anchored — so the "hidden" state translates _downward into the screen_ instead of off the top, and the fling direction would be `DOWN` against a top anchor. No current impact (nothing passes `action` yet), but this will bite the moment the "Expense added" action growl ships on native. Either make the native container honor `useBottomPosition` too, or explicitly force `useBottomPosition = false` on native.
> 
> **2. PR description mentions a `loading()` variant that isn't in the diff.** The description says it adds "a new `loading()` indefinite variant," but [`Growl.ts`](https://github.com/Expensify/App/blob/d9bb03d367532cbf4457fb2dfcf26b0928ecbe88/src/libs/Growl.ts#L58) only exports `show` / `error` / `success`. Either the description is stale or the code is missing — please reconcile so reviewers aren't looking for something that isn't there.
> 
> **3. Author checklist vs. tests.** The checklist has "I added unit tests" checked, but no tests are included and Codecov flags a coverage drop. The nonce-guard / stale-dismissal logic is the subtlest part of this change and is exactly what a small unit test would protect. "N/A – dormant groundwork" is a defensible call, but then that checkbox shouldn't be checked.
> 
> Minor / non-blocking
> * `useImperativeHandle(ref, () => ({show}), [show])` in `index.tsx` — `show` is no longer memoized (was `useCallback` before), so the handle is recreated every render. Harmless (functional `setContent` updater), just noting the behavior change.
> * `CONST.GROWL.OFFSCREEN_OFFSET = 255` is a magic offscreen distance; the comment acknowledges it only needs to exceed growl height + margins, so fine.
> * `linkReversed` / `buttonHoveredBGReversed` map to existing palette colors in both themes — verified they resolve.
> 
> I reviewed this at the code level against the PR head; I did **not** runtime-verify the slide animation on each platform, so please confirm the visual behavior (especially the bottom-right action layout on wide web) during QA.
> 
> Reply with `@MelvinBot <feedback>` to refine this review.

1. Replied in the comment thread
2. Removed it from the pr body
3. In the tests sectioned described that we cant test it until the second pr comes

src/components/GrowlNotification/index.tsx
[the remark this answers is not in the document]
borys3kk: already resolved

src/styles/index.ts
borys3kk: reverted back to original state

borys3kk:
all comments resolved so it's good to go

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,183 @@
...
+type GrowlNotificationContentProps = {
+    bodyText: string;
+    type: GrowlType;
+    duration: number;
+    action?: GrowlAction;
+
+    /** Identifies this growl instance; passed back through onDismissed so the parent can ignore stale dismissals. */
+    nonce: number;
+    onDismissed: (dismissedNonce: number) => void;
+};
```
mountiny, line 40: lets add docs to all type props

src/components/GrowlNotification/GrowlNotificationContent.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+// Every growl variant must have an icon mapping; keeping this map exhaustive over GrowlType.
+type GrowlIconTypes = Record<
+    GrowlType,
+    {
+        icon: React.FC<SvgProps> | IconAsset;
+        iconColor: string;
```
mountiny, line 47: docs

src/components/GrowlNotification/index.tsx
```diff
@@ -1,157 +1,65 @@
...
 
-import GrowlNotificationContainer from './GrowlNotificationContainer';
+import GrowlNotificationContent from './GrowlNotificationContent';
 
-const INACTIVE_POSITION_Y = -255;
+type GrowlContent = {
+    bodyText: string;
+    type: GrowlType;
+    duration: number;
+    action?: GrowlAction;
```
mountiny, line 15: docs

src/libs/Growl.ts
```diff
@@ -2,8 +2,16 @@ import CONST from '@src/CONST';
 
 import React from 'react';
 
+type GrowlAction = {
+    label: string;
+    onPress: () => void;
```
mountiny, line 7: docs

mountiny (approved):
Thank you! All comments are NAB

PR #94866, opened 2026-06-29
Confirm new expenses with a “View” toast on Spend instead of scrolling to and highlighting the row

borys3kk:
### Explanation of Change
Replaces the post-create Search row highlight with an "Expense added" / "Invoice sent" growl that has a "View" action deep-linking to the new expense's transaction thread RHP. The IOU actions (requestMoney/trackExpense/split/invoice/per-diem/distance) now solely own post-create navigation + feedback (batch-aware via `isLastTransactionOfBatch`), fixing double navigation/growls. Removes `useSearchHighlightAndScroll` (replaced by `useSearchAutoRefetch`, refetch-only) and the `TRANSACTION_IDS_HIGHLIGHT_ON_SEARCH_ROUTE` Onyx key.

### Fixed Issues
PROPOSAL:

### Tests
**Test 1: Expense-added growl with working "View" link (global create from Inbox)**
1. Sign in and go to the Inbox tab.
2. Press the global create (FAB) > Create expense, enter an amount, pick a workspace/participant, and submit from the confirmation page.
3. Verify that a green "Expense added" growl appears with a "View" button.
4. Press "View" and verify that the transaction thread RHP for the newly created expense opens (with the report visible underneath on wide layouts).
5. Repeat starting from the Reports (Spend) tab and verify the app lands on the Expenses search page and the same growl with a working "View" link appears there.

**Test 2: Single growl and single navigation on skip-confirm and multi-scan flows**
1. Seed the quick action: press the global create (FAB) > Create expense > Scan, upload a receipt, pick a recipient, and submit from the confirmation page. The FAB menu now shows a "Scan receipt" quick action for that recipient.
2. Press the FAB > the "Scan receipt" quick action, and take/upload 2 receipts. The submit happens immediately with no confirmation page (this is the skip-confirmation path).
3. Verify that navigation happens once (the scan modal dismisses with no flicker/double transition) and exactly ONE "Expense added" growl appears, after the last receipt of the batch.
4. Press "View" and verify it opens the thread of the last created expense.
5. Create an invoice via global create > Send invoice (requires a workspace with Invoices enabled under More features) and verify the growl says "Invoice sent" with a working "View" link.

**Test 3: In-report add highlights the row instead of a growl; Share flow stays correct**
1. Open an existing expense report and add an expense to it from within the report.
2. Verify that NO growl appears and the new row is highlighted in the report's transaction list.
3. (Native) Share an image into Expensify (Share flow > submit expense details) and verify a single "Expense added" growl appears after submission, with a working "View" link, and navigation happens once.

### Offline tests
1. Go offline, then create an expense from the global create button on the Inbox tab.
2. Verify the "Expense added" growl still appears and "View" opens the optimistic transaction thread RHP.
3. While on the Expenses search page offline, create another expense and verify the list refetches once you go back online (pending auto-refetch fires after reconnect).

### QA Steps
Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -1,36 +1,269 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-deprecated -- imperative module (not a React component); no useLocalize hook available here
+        Growl.success(growlMessage, CONST.GROWL.DURATION_WITH_ACTION, {label: translateLocal('common.view'), onPress: navigateToExpenseRHP});
+    };
+
+    // Fast path: the thread is already resolvable, so show the growl immediately instead of waiting on
+    // the reportActions subscription (which would otherwise hit the 8s safety timeout). This covers:
+    // - personal tracked expenses (unreported/self-DM): there's no iouReportID, but the thread ID is
+    //   passed in directly, so we can build from it right away;
+    // - the iouAction already being in Onyx (retry / non-deferred edge cases).
+    if (providedTransactionThreadReportID || (iouReportID && getIOUActionForReportID(iouReportID, transactionID)?.reportActionID)) {
```
JakubKorytko: this fast path now runs whenever callers pass `transactionThreadReportID`, but the new action call sites pass that ID before the deferred `API.write` has flushed. That skips the report-actions wait below, so the growl can show a `View` action before `iouAction.reportActionID` is in Onyx. Could we only fast-path `providedTransactionThreadReportID` when there is no `iouReportID`, or when the IOU action is already present?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -290,6 +290,8 @@ function SubmitDetailsPage({
                 isSelfTourViewed,
                 optimisticTransactionID,
                 currentUserLocalCurrency: currentUserPersonalDetails.localCurrencyCode ?? CONST.CURRENCY.USD,
+                // Navigation is owned by cleanupAndNavigateAfterExpenseCreate below; don't let the action navigate too.
+                shouldHandleNavigation: false,
```
JakubKorytko: `shouldHandleNavigation: false` no longer makes the action silent. `requestMoney`/`trackExpense` still call `surfaceExpenseCreatedFeedback` on that path, but this page has not navigated yet and immediately calls `cleanupAndNavigateAfterExpenseCreate` below. That can show the growl on the Share flow surface, and global-create paths could also show feedback again from the cleanup navigator. Can we add a separate `shouldShowFeedback` flag or let the action own the whole post-create flow here?

JakubKorytko (approved):
really good job! two medium-to-minor concerns, approved!

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
borys3kk: Good catch, fixed. The fast path now only takes a provided thread ID when there is no iouReportID (self-DM/track case where there is nothing to subscribe to), otherwise it requires the IOU action to already be in Onyx and falls through to the subscription wait. So View can no longer point at a thread whose optimistic data has not landed yet.

src/pages/Share/SubmitDetailsPage.tsx
borys3kk: Went with the separate flag: added shouldShowPostCreateFeedback (default true, only read when shouldHandleNavigation is false) and the Share page passes false, so the action is fully silent there and cleanupAndNavigateAfterExpenseCreate owns both navigation and feedback. Also threaded iouReportID / transactionThreadReportID through the cleanup helper so the growl's View link still resolves on that path. Didn't go with "action owns the whole flow" since the page computes its nav target with an optimisticChatReportID fallback for unknown-user share drafts that the action doesn't replicate.

borys3kk:
currently recording videos, made the pr ready for review to run the AI reviews, converted back to draft
edit: I will also check if I really need the ConnectWithoutView

mountiny:
@MelvinBot review

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -335,13 +340,19 @@ function SubmitDetailsPage({
                 betas,
                 personalDetails,
                 optimisticTransactionID,
+                // The whole post-create flow (navigation + feedback) is owned by cleanupAndNavigateAfterExpenseCreate below,
+                // which runs after this action returns - keep the action fully silent.
+                shouldHandleNavigation: false,
+                shouldShowPostCreateFeedback: false,
```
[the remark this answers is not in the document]
borys3kk: Good catch. Share drafts never set isFromGlobalCreate (initMoneyRequest is called without it), so the cleanup path took the silent non-global branch and the growl never fired. Fixed by adding an opt-in shouldAlwaysShowFeedback flag to cleanupAndNavigateAfterExpenseCreate / navigateAfterExpenseCreate that SubmitDetailsPage sets, so Share keeps its navigation to the destination report but still shows the growl once, after navigation. Didn't mark it as global create on purpose - that would reroute Share users to the Search page instead of the report hey picked. Added a unit test for the new flag.

situchan:
Please check AI review comments and fix conflict

src/libs/actions/IOU/TrackExpense.ts
```diff
@@ -1923,8 +1922,30 @@ function requestMoney(requestMoneyInformation: RequestMoneyInformation): {iouRep
...
+    if (!requestMoneyInformation.isRetry && isLastTransactionOfBatch) {
+        if (shouldHandleNavigation) {
+            const navigationReportID = backToReport ?? activeReportID;
+            handleNavigateAfterExpenseCreate({
+                activeReportID: navigationReportID,
+                iouReportID: iouReport?.reportID,
+                transactionID: transaction.transactionID,
+                transactionThreadReportID: transactionThreadReportID ?? iouAction?.childReportID,
+                isFromGlobalCreate,
+                shouldAddPendingNewTransactionIDs: isMoneyRequestReport,
```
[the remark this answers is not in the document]
borys3kk: Right, the gate was inverted relative to the consumer - the chat preview card is the thing that needs pending IDs on first mount, and isMoneyRequestReport is false exactly then. Fixed by making it isMoneyRequestReport || navigationReportID === chatReport.reportID, so chat destinations get the first-load highlight back (same as the old cleanup path) while in-report adds keep the pending-ID fallback for the 1-to-2 fresh-mount case that diff detection also misses. Applied the same fix to createDistanceRequest in Split.ts which had the identical gate.

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -1,36 +1,272 @@
...
+
+    safetyTimeoutID = setTimeout(() => {
+        if (resolved) {
+            return;
+        }
+        resolved = true;
+        Onyx.disconnect(connectionId);
+        const threadReportID = buildThreadFromOnyx();
+        showGrowl(threadReportID);
+    }, SAFETY_TIMEOUT_MS);
```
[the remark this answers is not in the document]
borys3kk: I don't think this is actually reachable - Onyx delivers the initial connect callback asynchronously (connectWithoutView goes through Promise.resolve().then and get(key).then internally), and the block between connect and the setTimeout assignment is synchronous, so no callback can run before safetyTimeoutID is set. Worst case the timer would just early-return on the resolved flag after 8s anyway. That said, the guard is cheap insurance against Onyx ever changing its delivery timing, so added it as suggested.

src/pages/iou/request/step/confirmation/useExpenseSubmission.ts
```diff
@@ -702,26 +674,15 @@ function useExpenseSubmission(params: UseExpenseSubmissionParams) {
...
-        const participant = selectedParticipantsForRequest.at(0);
-        if (!participant) {
-            return;
-        }
 
-        const {chatReportID: distanceChatReportID, transactionID: distanceTransactionID} = createDistanceRequestIOUActions({
+        createDistanceRequestIOUActions({
             report,
             participants: selectedParticipantsForRequest,
             currentUserLogin: currentUserPersonalDetails.login ?? '',
```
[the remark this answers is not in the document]
borys3kk: Good catch - main did guard this (the participant var fed performPostBatchCleanup, and when that param went away the guard went with it). Restored an explicit early return when selectedParticipantsForRequest is empty so we never hit the participants.at(0) ?? {} fallback with no participant.

src/libs/deferredLayoutWrite.ts
```diff
@@ -140,6 +146,25 @@ function flushDeferredWrite(key: string) {
...
+ * most-recently-registered active channel or immediately if nothing is pending.
+ * Used to sequence work that depends on the optimistic data existing.
+ */
+function runAfterDeferredWrite(callback: () => void) {
+    const channel = [...channels.values()].reverse().find((c) => !c.isReserved);
+    if (!channel) {
+        callback();
+        return;
+    }
+    channel.onFlush = [...(channel.onFlush ?? []), callback];
```
JakubKorytko, line 172: This looks good to me as an interim step. `runAfterDeferredWrite()` without a key is not fully bulletproof as a generic primitive, since it relies on the latest non-reserved channel being the intended one. In this flow that should be OK because it is called synchronously right after the deferred write is registered.

I would avoid spreading this unkeyed pattern further, since the longer-term direction is to move this into an API-owned `API.writeWhenReady(...)` model: https://github.com/Expensify/App/issues/96010

So I’m OK with this PR as long as this stays scoped to the current flow.

Issue #94821, opened 2026-06-29
[Due for payment 2026-08-03] [$250] Ensure force touch creation options land the member in Spend > Expenses


JmillsExpensify:
I just noticed this inconsistency taking expense creation in the app: If you create an expense with one of these options, we take you to the Inbox. 

<img width="864" height="1413" alt="Image" src="https://github.com/user-attachments/assets/cf8afa13-2702-44ae-86d9-dd8c2cfc7264" />

However, the existing platform standard is that creation flows outside of the Inbox take you to the Spend > Expenses page. Accordingly, this fix this inconsistency and ensure that the user is taken to Spend > Expenses anytime they choose `Track distance`, `Create expense` or `Scan receipt` in the "force touch" flow. 
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: <a href="https://www.upwork.com/jobs/~022071548366745225254">https://www.upwork.com/jobs/~022071548366745225254</a></li>
        <li>Upwork Job ID: 2071548366745225254</li>
        <li>Last Price Increase: 2026-06-29</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

MelvinBot:
## Proposal

### What is the root cause of that problem?

Post-creation navigation is centralized in [`navigateAfterExpenseCreate`](https://github.com/Expensify/App/blob/4de5999438332878e3b9f371a451674db230f22f/src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts#L41-L107). For a global-create ("force touch") flow, it decides between opening the chat in the Inbox vs. navigating to **Spend > Expenses** based on **where the user currently is** — not on the fact that the flow was started from the FAB.

The deciding line is the short-circuit at [`navigateAfterExpenseCreate.ts:55`](https://github.com/Expensify/App/blob/4de5999438332878e3b9f371a451674db230f22f/src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts#L50-L63):

```ts
const isUserOnInbox = isReportTopmostSplitNavigator();
if (!isFromGlobalCreate || isUserOnInbox || !transactionID) {
    // -> dismissModalAndOpenReportInInboxTab(...)  (stays in Inbox)
}
// otherwise -> ROUTES.SEARCH_ROOT with a canned EXPENSE query (Spend > Expenses)
```

[...]

PR #94952, opened 2026-06-30
fix: force touch creation options land in Spend > Expenses

daledah:
### Fixed Issues

$ https://github.com/Expensify/App/issues/94821
PROPOSAL:

### Tests

1. Long press on App icon
2. Choose any options ('create expense', 'track distance', 'scan receipt')
3. Complete the flow
4. Verify that: user land in `Spend > Expenses` page

### Offline tests

Same as tests
### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/d953f6e3-3098-4d67-924c-aa68d2d6ea05

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/134fe484-bebd-4471-93a2-36cd4f00b58f

</details>

<details>

</details>

<details>

</details>


marufsharifi:
@daledah, could you please check the AI feedback and failed actions? thanks.

daledah:
@marufsharifi i fixed

src/pages/iou/request/IOURequestStartPage.tsx
```diff
@@ -189,6 +190,16 @@ function IOURequestStartPage({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    useEffect(() => {
+        // Native home-screen shortcuts bypass startMoneyRequest (which sets isFromFloatingActionButton),
+        // so if this is a global create without that flag, it must be a native shortcut.
+        if (!isFromGlobalCreate || transaction?.isFromFloatingActionButton) {
+            return;
+        }
+        setNativeShortcutFlag(route.params.transactionID);
```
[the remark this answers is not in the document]
marufsharifi: What do you think about this?

marufsharifi:
@daledah, could you please merge the main? thanks.

marufsharifi:
I will complete my review here tomorrow.

marufsharifi:
@MelvinBot, Review

marufsharifi:
@daledah, it doesn't work on Android. Could you please take a look? thanks. 


https://github.com/user-attachments/assets/b81bfaca-e2e5-4a61-ad59-bf813b3c28b1

daledah:
@MelvinBot to clarify the changes in `SubmitExpenseOrchestrator`. The core issue is that `SubmitExpenseOrchestrator` (via `getSubmitHandler`) is the primary navigation controller for the confirmation page submit flow. When a native shortcut is initiated from Inbox, the orchestrator pre-inserts a Report route and `getSubmitHandler` selects `REPORT_PRE_INSERT` -> user lands in Inbox instead of Spend > Expenses. This is why changes in this file are required, not out-of-scope

@MelvinBot please investigate and find out why it only works on iOS and not on Android as this [comment](https://github.com/Expensify/App/pull/94952#issuecomment-4881080747)

daledah:
> @daledah, it doesn't work on Android. Could you please take a look? thanks.


@marufsharifi i fixed, could you please check again?

marufsharifi:
@MelvinBot review

marufsharifi:
@daledah, could you please check this? It still doesn't work


https://github.com/user-attachments/assets/d23460d7-cf76-4693-8108-e131900a78cf

marufsharifi:
@daledah, could you please check this? 



https://github.com/user-attachments/assets/ce8de9ba-b1bf-4873-a0a6-25722413ba0b

marufsharifi:
@daledah, this flow doesn't work as well. 



https://github.com/user-attachments/assets/3af90b03-fcf1-406c-bcc2-49601cc72aa6

daledah:
@marufsharifi After investigating, I found the issue: the shortcut detection lived in a `useEffect` on the start pages and raced with the tab-switch draft reset - the draft got rebuilt before the effect ran, so the flag was never set (that's why it only worked intermittently).

I've updated the approach to detect the shortcut at the deeplink layer instead (a module-level marker set when the incoming URL matches the shortcut deeplink), which is race-free since native shortcuts are the only create-expense entry arriving via a `Linking` URL

Could you please check again?

marufsharifi:
@daledah, could you please check these?



https://github.com/user-attachments/assets/7a566504-6dd7-4281-9fc3-db1cec288b96

https://github.com/user-attachments/assets/cd49a375-1fd4-4706-a4c2-77c69ddf8d83

daledah:
> RF-2.mp4

@marufsharifi i can't reproduce this bug 

https://github.com/user-attachments/assets/6f8039ee-f0c1-49d2-bc4d-8a1f9c9dbde1

> RF-1.mp4

I think this is intentional, clicking "save for later" will hide the modal

marufsharifi:
> > RF-2.mp4
> 
> @marufsharifi i can't reproduce this bug
> 
>  android.mov 
> > RF-1.mp4
> 
> I think this is intentional, clicking "save for later" will hide the modal

@daledah, could you please cold-start and then call create expense?

marufsharifi:
@daledah, Gentle bump. thanks.

daledah:
@marufsharifi  i fixed 

https://github.com/user-attachments/assets/705e87cb-8d40-4286-9598-74d1af8fa7f3

src/libs/NativeShortcutFlow.ts
```diff
@@ -0,0 +1,39 @@
...
+function markNativeShortcutFlowIfNeeded(url: string | null | undefined): void {
+    if (!url || !NATIVE_SHORTCUT_DEEPLINK_PATTERN.test(url)) {
+        return;
+    }
+    isNativeShortcutFlow = true;
+}
+
+function endNativeShortcutFlow(): void {
+    isNativeShortcutFlow = false;
+}
```
marufsharifi, line 23: Could we add focused tests for `markNativeShortcutFlowIfNeeded()`, `endNativeShortcutFlow()`, and the exact shortcut URL shapes? This marker is now the load-bearing fix for both cold and warm starts.

src/pages/Search/SearchPageNarrow/StaticSearchTypeMenu.tsx
```diff
@@ -43,8 +43,8 @@ function StaticSearchTypeMenu({queryJSON}: {queryJSON: SearchQueryJSON}) {
     const submitSearch = suggestedSearches[CONST.SEARCH.SEARCH_KEYS.SUBMIT];
 
     const tabs: TabSelectorBaseItem[] = [
-        {key: reportsSearch.key, icon: expensifyIcons.Document, title: translate(reportsSearch.translationPath)},
         {key: expensesSearch.key, icon: expensifyIcons.Receipt, title: translate(expensesSearch.translationPath)},
+        {key: reportsSearch.key, icon: expensifyIcons.Document, title: translate(reportsSearch.translationPath)},
```
marufsharifi, line 74: Why do we need this change?

src/pages/iou/request/DistanceRequestStartPage.tsx
```diff
@@ -51,7 +51,12 @@ function DistanceRequestStartPage({
     const {translate} = useLocalize();
     const [report] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT}${reportID}`);
     const [transaction] = useOnyx(`${ONYXKEYS.COLLECTION.TRANSACTION_DRAFT}${getNonEmptyStringOnyxID(route?.params.transactionID)}`);
-    const {policy} = usePolicyForTransaction({transaction, reportPolicyID: report?.policyID, action, iouType});
+    const {policy} = usePolicyForTransaction({
+        transaction,
+        reportPolicyID: report?.policyID,
+        action,
```
marufsharifi: **NAB:** These changes appear to be formatting-only and are unrelated to the final deeplink-layer approach. Since they do not affect behavior, could we remove them to reduce review noise, unless they are required by the formatter?

src/pages/iou/request/IOURequestStartPage.tsx
```diff
@@ -246,7 +246,13 @@ function IOURequestStartPage({
         // receipt) until the tab-switch reset rebuilds it as manual. Mounting the embedded confirmation against that
         // stale scan draft does throwaway work (scan loader, reading the receipt blob and a heavy first render) that
         // is immediately discarded once the reset lands. Wait for the reset so the manual confirmation mounts once.
-        manualTabContent = <FullScreenLoadingIndicator reasonAttributes={{context: 'IOURequestStartPage.manualTabPendingReset'}} />;
+        manualTabContent = (
+            <FullScreenLoadingIndicator
+                reasonAttributes={{
+                    context: 'IOURequestStartPage.manualTabPendingReset',
+                }}
+            />
```
marufsharifi: These changes appear to be formatting-only and are unrelated to the final deeplink-layer approach. Since they do not affect behavior, could we remove them to reduce review noise, unless they are required by the formatter?

tests/unit/getSubmitHandlerTest.ts
```diff
@@ -12,6 +12,7 @@ const BASE_SNAPSHOT: SubmitNavigationSnapshot = {
     isReportTopmostSplit: false,
     isSearchTopmostFullScreen: false,
     isDestinationReportLoaded: false,
+    isFromNativeShortcut: false,
```
marufsharifi, line 15: Could we add a focused test confirming that `isFromNativeShortcut && isFromGlobalCreate` takes priority over competing report or dismiss handlers and returns `SEARCH_PRE_INSERT`? The fixture now includes `isFromNativeShortcut: false`, but this precedence is not currently asserted.

src/pages/iou/request/step/confirmation/getSubmitHandler.ts
```diff
@@ -76,6 +79,11 @@ function getSubmitHandler(snapshot: SubmitNavigationSnapshot): SubmitHandler {
     if (snapshot.isPreInserted && !snapshot.isReportPreInserted) {
         return SUBMIT_HANDLER.SEARCH_PRE_INSERT;
     }
+    // Native shortcut flows should always land on Spend > Expenses regardless of
+    // what was pre-inserted or whether the pre-insert timer has fired yet.
+    if (snapshot.isFromNativeShortcut && snapshot.isFromGlobalCreate) {
```
marufsharifi, line 83: **Blocking:** Could we add focused tests confirming that `isFromNativeShortcut && isFromGlobalCreate` forces `SEARCH_PRE_INSERT` even when report pre-insert or dismiss fast-path conditions are also true? This is the core deterministic routing rule for the fix, and its priority is not currently covered.

src/libs/NativeShortcutFlow.ts
```diff
@@ -0,0 +1,39 @@
...
+ *    quickActionCallback, which generates `/create/create/start/1/{reportID}/scan|manual|distance-new`)
+ *  - consumed by `useResetIOUType` so the rebuilt draft carries `isFromNativeShortcut`
+ *  - cleared whenever an in-app flow starts a new money request, so FAB and other in-app flows
+ *    keep their original post-submit navigation
+ */
+let isNativeShortcutFlow = false;
+
+// Intentionally does not match the GPS trip deeplink (`.../distance-new/distance-gps`), which is
+// sent by the GPS tracking notification, not a home-screen shortcut.
+const NATIVE_SHORTCUT_DEEPLINK_PATTERN = /\/create\/create\/start\/[^/]+\/[^/]+\/(?:scan|manual|distance-new)\/?(?:\?.*)?$/;
```
marufsharifi: This hard-coded regex duplicates the native shortcut route shapes already defined in `ROUTES`, `linkingConfig`, and the mobile quick-action deeplinks. Could we derive or centralize this matcher near those route definitions, or at least wrap it in a narrow helper with focused tests, so future route changes do not cause the marker logic to drift?

marufsharifi:
@daledah, thanks, I am reviewing the code.

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -44,17 +45,19 @@ function navigateAfterExpenseCreate({
...
 }: NavigateAfterExpenseCreateParams) {
     const isUserOnInbox = isReportTopmostSplitNavigator();
 
-    // If the expense is not created from global create or is currently on the inbox tab,
+    // If the expense is not created from global create or is currently on the inbox tab
+    // (unless it's from a native shortcut, which should always go to Spend > Expenses),
     // we just need to dismiss the money request flow screens
     // and open the report chat containing the IOU report
-    if (!isFromGlobalCreate || isUserOnInbox || !transactionID) {
+    if (!isFromGlobalCreate || (isUserOnInbox && !isFromNativeShortcut) || !transactionID) {
```
marufsharifi: **Blocking:** Could we add a test covering the native-shortcut flow from Inbox? With `isReportTopmostSplitNavigator`, `isFromGlobalCreate`, and `isFromNativeShortcut` all set to `true`, we should assert navigation to `ROUTES.SEARCH_ROOT` rather than calling `dismissModalWithReport`.

src/libs/NativeShortcutFlow.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Tracks whether the current money-request create flow was opened from a native home-screen
+ * shortcut (force touch / long-press quick action) deeplink, e.g. "Scan receipt".
```
marufsharifi: **Blocking:** Could we add tests for the new native-shortcut deep-link matcher and marker lifecycle? Please cover matching `scan`, `manual`, and `distance-new`; non-matching URLs such as `distance-gps`; and verify that `endNativeShortcutFlow()` clears the marker.

src/pages/iou/request/step/confirmation/useExpenseSubmission.ts
```diff
@@ -347,6 +347,7 @@ function useExpenseSubmission(params: UseExpenseSubmissionParams) {
             draftTransactionIDs,
             transactionID: lastTransactionID,
             isFromGlobalCreate: getIsFromGlobalCreate(lastTransaction),
+            isFromNativeShortcut: !!lastTransaction?.isFromNativeShortcut,
```
marufsharifi, line 370: This new routing depends on `lastTransaction.isFromNativeShortcut`, but multi-receipt scan creates additional draft transactions through `buildOptimisticTransactionAndCreateDraft`, which currently only copies `isFromGlobalCreate` and `isFromFloatingActionButton`. So the last transaction can lose this new flag and route back to Inbox. Can we preserve `isFromNativeShortcut` there too?

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
marufsharifi: The inline comment below captures the native shortcut exception, but can we also update this docblock? It still says global create from Inbox opens the chat report, while this PR adds an exception for native shortcut flows that should go to `Spend > Expenses` even when `Inbox` is topmost.

marufsharifi (changes requested):

marufsharifi:
@daledah, could you please review the comments above and the failed actions? thanks.

src/pages/Search/SearchPageNarrow/StaticSearchTypeMenu.tsx
daledah: It is used to fix this [bug](https://github.com/Expensify/App/pull/94952#issuecomment-4912176046)

src/pages/iou/request/step/confirmation/SubmitExpenseOrchestrator.tsx
```diff
@@ -226,15 +232,19 @@ function SubmitExpenseOrchestrator({
...
+            setIsConfirming(false);
+        };
+
+        if (isFromNativeShortcutOnTransaction) {
+            // Native shortcuts may have a report pre-inserted, so replace it with the search route.
+            const searchRoute = ROUTES.SEARCH_ROOT.getRoute({query: buildCannedSearchQuery({type: searchDataType})});
+            Navigation.revealRouteBeforeDismissingModal(searchRoute, {afterTransition});
+        } else {
+            Navigation.dismissModal({afterTransition});
+        }
```
marufsharifi: Could you clarify why this shared wide-layout dismiss behavior needs to change for force-touch routing? This condition also affects ordinary global-create submit flows, not just native shortcuts, and the PR doesn’t make that dependency clear.

Please also confirm this has been tested for wide-layout FAB create, scan, and distance flows to rule out regressions.

src/libs/Navigation/linkingConfig/subscribe.ts
```diff
@@ -30,6 +31,9 @@ const skipRules: ReadonlyArray<{urlMatcher: RegExp; focusedScreens: readonly str
 
 const subscribe: LinkingOptions<RootNavigatorParamList>['subscribe'] = (listener) => {
     const subscription = Linking.addEventListener('url', ({url}: {url: string}) => {
+        // A native home-screen shortcut deeplink marks the create flow before navigation mounts it.
+        markNativeShortcutFlowIfNeeded(url);
+
```
marufsharifi: `markNativeShortcutFlowIfNeeded(url)` runs before the skip rules are evaluated, so a deeplink that the linking layer intentionally drops can still mutate module-level shortcut state.

This can leak navigation origin into later flows: `useResetIOUType` may then read `isNativeShortcutFlowActive()` and apply `isFromNativeShortcut` to an unrelated draft.

Could we defer setting the marker until after the skip/Plaid guards, immediately before the direct or deferred `listener(url)` call? Alternatively, return shortcut metadata from the helper and consume it only on the accepted deeplink path. This keeps dropped URLs side-effect free and scopes the state to a route that is actually opened.

src/libs/NativeShortcutFlow.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Native shortcuts are the only create-expense entry point that reaches the app through a
+ * `Linking` URL (all in-app entry points navigate via `Navigation.navigate` and go through
+ * `startMoneyRequest`/`startDistanceRequest`). Detecting the shortcut at the deeplink layer is
+ * therefore unambiguous and avoids racing against the draft rebuilds (`initMoneyRequest`) that
+ * happen while the create flow mounts.
+ *
+ * The marker is:
+ *  - set when a received URL matches the shortcut deeplink shape (see Mobile-Expensify's
+ *    quickActionCallback, which generates `/create/create/start/1/{reportID}/scan|manual|distance-new`)
+ *  - consumed by `useResetIOUType` so the rebuilt draft carries `isFromNativeShortcut`
```
marufsharifi: **NAB:** The module comment says this marker is “consumed” by `useResetIOUType`, but that hook only reads `isNativeShortcutFlowActive()` and does not clear it. The marker is cleared later by `startMoneyRequest` or `startDistanceRequest`.

“Consumed” usually implies read-and-clear, so this wording could mislead future maintainers into assuming the marker cannot remain active after it has been applied to a draft. Could we either update the comment to say it is “read by `useResetIOUType`” or add a `consumeNativeShortcutFlow()` helper that returns the current value and clears it?

marufsharifi (changes requested):

marufsharifi:
@codex review

src/pages/iou/request/step/confirmation/SubmitExpenseOrchestrator.tsx
daledah: The `if (isFromNativeShortcutOnTransaction)` guard ensures only native shortcuts use `revealRouteBeforeDismissingModal` - ordinary global create flows always hit the else branch `(Navigation.dismissModal)`, which is identical to the previous behavior.

Confirmed: i tested wide-layout FAB create, scan, and distance flows - all land correctly on Spend > Expenses with no regressions.

JmillsExpensify (dismissed):
The HL flow tests great to me. Let's address C+ feedback and get this merged.

src/libs/Navigation/linkingConfig/index.ts
```diff
@@ -1,19 +1,42 @@
...
+ * native-shortcut flow before the URL is handed to navigation — guaranteeing the marker is set
+ * before the create screen mounts, even on HybridApp cold starts where DeepLinkHandler runs late.
+ */
+function getInitialURL(): Promise<string | null | undefined> {
+    return Promise.race([
+        Linking.getInitialURL(),
+        new Promise<undefined>((resolve) => {
+            setTimeout(resolve, CONST.TIMING.REACT_NAVIGATION_GET_INITIAL_URL_TIMEOUT);
+        }),
+    ]).then((url) => {
```
marufsharifi: **NAB:** Prefer `async/await` over `.then()`

Per `STYLE.md`, we use `async/await` rather than `.then/.catch`; raw promises are reserved for wrapping callbacks / deferred signals. The `Promise.race([...])` wrapping the `setTimeout` is fine (that's the deferred-signal case), but the trailing `.then((url) => { ... })` is a plain result transformation — exactly what `async/await` is for.

```ts
async function getInitialURL(): Promise<string | null | undefined> {
    const url = await Promise.race([
        Linking.getInitialURL(),
        new Promise<undefined>((resolve) => {
            setTimeout(resolve, CONST.TIMING.REACT_NAVIGATION_GET_INITIAL_URL_TIMEOUT);
        }),
    ]);
    markNativeShortcutFlowIfNeeded(url);
    return url;
}
```

src/pages/iou/request/step/confirmation/SubmitExpenseOrchestrator.tsx
```diff
@@ -331,8 +340,8 @@ function SubmitExpenseOrchestrator({
             });
         };
 
-        if (shouldNavigateToSearch && !isNarrow) {
-            dismissWideToNewSearchType(searchType, runAfterDismiss);
+        if (!isSameType && !isNarrow && shouldNavigateToSearch) {
```
marufsharifi: **Out of scope (medium)** — the `handleSearchDismiss` condition gained an `!isSameType` guard:

```tsx
// before: if (shouldNavigateToSearch && !isNarrow)
if (!isSameType && !isNarrow && shouldNavigateToSearch) {
    dismissWideToNewSearchType(searchDataType, runAfterDismiss);
    return;
}
```

`handleSearchDismiss` is the `SEARCH_DISMISS` path, which the native-shortcut flow **does not use** — shortcuts are forced onto `SEARCH_PRE_INSERT` in `getSubmitHandler`. So this changes wide-layout dismiss behavior for ordinary global-create flows, which is adjacent logic not required by this fix.

**Suggestion:** keep only if this is a deliberate, tested part of the fix; otherwise split it out.

src/libs/Navigation/helpers/submitWithDismissFirst.ts
```diff
@@ -24,6 +24,8 @@ type SubmitWithDismissFirstParams = {
     destinationReportID: string | undefined;
     /** Telemetry metadata for the submit-expense performance span. */
     telemetryContext: SubmitExpenseContext;
+    /** Whether the flow was opened from a native home-screen shortcut (skips the report fast paths). */
+    isFromNativeShortcut?: boolean;
```
marufsharifi: **Naming: the value passed here doesn't match the prop name.**

Every call site passes the *conjunction* `isFromNativeShortcut && isFromGlobalCreate` into this prop, not a bare "is from native shortcut" flag:

- `src/libs/IOUAmountSubmission.ts:398`
- `src/pages/iou/request/step/IOURequestStepDistance/handleMoneyRequestStepDistanceNavigation.ts:112` and `:127`
- `src/pages/iou/request/step/IOURequestStepScan/components/ScanSkipConfirmation.tsx:154` and `:349`

So a reader sees `isFromNativeShortcut: transactionIsFromNativeShortcut && !!transactionIsFromGlobalCreate` — a name that understates what the value encodes — and the same expression is duplicated in 5 places. `getSubmitHandler.ts:297` also does this same `&&`, but *internally*, so the two paths express the same concept inconsistently.

Suggest one of:
- rename the prop to reflect its real meaning, e.g. `shouldForceSpendExpensesLanding`, **or**
- accept both `isFromNativeShortcut` and `isFromGlobalCreate` here and AND them internally (mirroring `getSubmitHandler`), so call sites stop repeating the conjunction under a misleading label.

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -34,27 +35,25 @@ function getNavigateAfterCreateSearchNavigatorState() {
...
 }: NavigateAfterExpenseCreateParams) {
     const isUserOnInbox = isReportTopmostSplitNavigator();
 
-    // If the expense is not created from global create or is currently on the inbox tab,
-    // we just need to dismiss the money request flow screens
-    // and open the report chat containing the IOU report
-    if (!isFromGlobalCreate || isUserOnInbox || !transactionID) {
+    // Dismiss to the report chat unless the flow must land on Spend > Expenses
+    // (global create outside the inbox tab, or any native shortcut).
+    if (!isFromGlobalCreate || (isUserOnInbox && !isFromNativeShortcut) || !transactionID) {
```
marufsharifi: **Dense boolean — worth naming the intent.**

This condition packs three distinct exit reasons plus a nested `&&` into one expression. The comment above helps, but verifying the native-shortcut carve-out still requires careful parsing. Extracting a named positive condition reads more directly:

```ts
const shouldLandOnSpendExpenses = isFromGlobalCreate && !!transactionID && (!isUserOnInbox || isFromNativeShortcut);
if (!shouldLandOnSpendExpenses) {
    // dismiss to the report chat
    ...
}
```

src/hooks/useResetIOUType.ts
```diff
@@ -106,6 +107,11 @@ function useResetIOUType({
 
     const resetIOUTypeIfChanged = (newIOUType: IOURequestType) => {
         if (transaction?.iouRequestType === newIOUType) {
+            // The draft won't be rebuilt, so stamp the native-shortcut marker onto it directly.
```
marufsharifi: **Two branches solve the same problem two different ways.**

This early-return branch marks + ends the native-shortcut flow via `setNativeShortcutFlag` + `endNativeShortcutFlow` guarded by `isNativeShortcutFlowActive() && !transaction.isFromNativeShortcut`, while the rebuild path (line 131 `shouldMarkNativeShortcut = transaction?.isFromNativeShortcut ?? isNativeShortcutFlowActive()`, then the `initMoneyRequest` param + the `endNativeShortcutFlow()` at line ~155) does it through a different mechanism and a different guard.

The asymmetry is intentional (draft not rebuilt vs. rebuilt), but a reader has to hold both branches in their head to see they're the same operation. A one-line comment tying them together would orient the reader — e.g. *"both paths copy the marker onto the draft then clear the flow; they differ only in how the draft is written (Onyx.merge here vs. initMoneyRequest param below)."* No structural change needed.

src/libs/Navigation/helpers/submitWithDismissFirst.ts
```diff
@@ -54,7 +56,19 @@ function startDismissFirstTracking(
  *
  * Must not be called from `src/libs/actions/` — view-layer only.
  */
-function submitWithDismissFirst({executeWrite, destinationReportID, telemetryContext}: SubmitWithDismissFirstParams): void {
+function submitWithDismissFirst({executeWrite, destinationReportID, telemetryContext, isFromNativeShortcut = false}: SubmitWithDismissFirstParams): void {
+    // Native shortcuts always land on Spend > Expenses, so skip the report fast paths and
+    // let the write's cleanup run `navigateAfterExpenseCreate`.
+    if (isFromNativeShortcut) {
```
marufsharifi: **High — this new branch is untested.** It short-circuits the entire orchestrator (clears the pre-inserted flag, `startTracking(..., {skipSubmitExpenseSpan: true})`, `setFastPath(DEFAULT)`, `executeWrite({shouldHandleNavigation: true})`), but `tests/unit/submitWithDismissFirstTest.ts` never sets `isFromNativeShortcut`, so neither the happy nor the regression path is exercised.

Please add a `describe('Native shortcut branch')` to `tests/unit/submitWithDismissFirstTest.ts`:

- `it('skips report fast paths and calls executeWrite with shouldHandleNavigation=true')` — assert `executeWrite` is called with `{shouldHandleNavigation: true}`, `setFastPath` with `FAST_PATH_HANDLER.DEFAULT`, and that neither the Search-topmost nor destination-report path runs.
- `it('clears the fullscreen pre-inserted flag when it is set')` — mock `Navigation.getIsFullscreenPreInsertedUnderRHP()` → `true`, assert `clearFullscreenPreInsertedFlag` was called.
- Regression: `it('with isFromNativeShortcut=false still runs the destination-report branch')`.

src/hooks/useResetIOUType.ts
```diff
@@ -106,6 +107,11 @@ function useResetIOUType({
...
+            if (isNativeShortcutFlowActive() && !transaction.isFromNativeShortcut) {
```
marufsharifi, line 113: **High — marker-consumption logic has no test.** This hook is the glue that turns the module-level marker into the persisted `isFromNativeShortcut` draft flag; the whole feature depends on it, yet there is no `useResetIOUType` test at all. Two untested branches:

1. **Same-type path** (line 109, `iouRequestType === newIOUType`): stamps `setNativeShortcutFlag` + `endNativeShortcutFlow` only when `isNativeShortcutFlowActive() && !transaction.isFromNativeShortcut`.
2. **Rebuild path** (line 131 → 155): `shouldMarkNativeShortcut = transaction?.isFromNativeShortcut ?? isNativeShortcutFlowActive()` is passed to `initMoneyRequest` and the marker is cleared afterward.

Please add `tests/unit/useResetIOUTypeTest.ts` (renderHook) asserting: marker active + no existing flag → `setNativeShortcutFlag` then `endNativeShortcutFlow`; existing draft flag already set → not re-stamped; **no marker & no flag → neither called** (negative path). If rendering the hook is impractical, extract the marker decision into a pure helper and unit-test that.

src/libs/actions/IOU/MoneyRequest.ts
```diff
@@ -430,6 +436,12 @@ function createDraftTransaction(transaction: Transaction) {
     Onyx.set(`${ONYXKEYS.COLLECTION.TRANSACTION_DRAFT}${transaction.transactionID}`, newTransaction);
 }
 
+function setNativeShortcutFlag(transactionID: string) {
```
marufsharifi, line 441: **Medium — `setNativeShortcutFlag` and the `initMoneyRequest` flag spread are untested.** `tests/actions/IOU/MoneyRequestSettersTest.ts` exercises `initMoneyRequest` but never asserts `isFromNativeShortcut`, and `setNativeShortcutFlag` is imported by no test.

Please extend `MoneyRequestSettersTest.ts`:

- `it('setNativeShortcutFlag merges isFromNativeShortcut:true onto the draft')`.
- `it('initMoneyRequest persists isFromNativeShortcut when passed')` — covers the truthy spread at line 331 / 414.
- `it('initMoneyRequest omits isFromNativeShortcut when not passed')` — covers the `...(isFromNativeShortcut ? {...} : {})` **false** branch.

src/libs/actions/TransactionEdit.ts
```diff
@@ -187,6 +187,7 @@ function buildOptimisticTransactionAndCreateDraft({initialTransaction, currentUs
         transactionID: newTransactionID,
         isFromGlobalCreate,
         isFromFloatingActionButton,
+        isFromNativeShortcut,
```
marufsharifi, line 190: **Medium — flag propagation is untested.** `buildOptimisticTransactionAndCreateDraft` now copies `isFromNativeShortcut` from `initialTransaction` (destructured at line 179, written at line 190), but no test covers this function.

Please add a test asserting the created draft carries `isFromNativeShortcut` when the initial transaction has it, and leaves it `undefined` when absent.

src/pages/iou/request/step/confirmation/SubmitExpenseOrchestrator.tsx
```diff
@@ -226,15 +232,19 @@ function SubmitExpenseOrchestrator({
...
-                setIsConfirming(false);
-            },
-        });
+
+        const afterTransition = () => {
+            createTransaction(locationPermissionGranted, false);
+            setIsConfirming(false);
+        };
+
+        if (isFromNativeShortcutOnTransaction) {
```
marufsharifi: **Medium — the reveal-vs-dismiss decision here is untested.** `getSubmitHandler` picks `SEARCH_PRE_INSERT`, but the actual navigation choice in `handleSearchPreInsert` — `revealRouteBeforeDismissingModal(searchRoute)` for a native shortcut vs `dismissModal` otherwise (lines 241–246) — is separate component logic that no test reaches.

Please cover it: either extract the "which navigation call" decision into a testable helper, or add a render test asserting a native-shortcut transaction triggers `revealRouteBeforeDismissingModal` with the canned search route (and the non-shortcut case still calls `dismissModal`).

src/libs/Navigation/linkingConfig/index.ts
```diff
@@ -1,19 +1,42 @@
...
 import prefixes from './prefixes';
 import subscribe from './subscribe';
 
+/**
+ * React Navigation's default `getInitialURL` (incl. its timeout workaround), plus marking the
+ * native-shortcut flow before the URL is handed to navigation — guaranteeing the marker is set
+ * before the create screen mounts, even on HybridApp cold starts where DeepLinkHandler runs late.
+ */
+function getInitialURL(): Promise<string | null | undefined> {
+    return Promise.race([
```
marufsharifi: **Duplicated `getInitialURL`-with-timeout pattern** (medium)

This `Promise.race([Linking.getInitialURL(), new Promise(resolve => setTimeout(resolve, …))])` guard already exists in `src/DeepLinkHandler.tsx:72`, which races the same call against `CONST.TIMING.GET_INITIAL_URL_TIMEOUT`. We now have two hand-rolled versions of the same primitive, differing only in the timeout (`REACT_NAVIGATION_GET_INITIAL_URL_TIMEOUT: 150` vs `GET_INITIAL_URL_TIMEOUT: 10000`) and the fallback value (`undefined` vs `null`).

**Suggestion:** extract a shared helper — e.g. `getInitialURLWithTimeout(timeoutMs, fallback)` in `src/libs/Navigation/helpers/` — and have both this file and `DeepLinkHandler.tsx` call it. The `.then(markNativeShortcutFlowIfNeeded)` here composes cleanly on top of that helper.

src/libs/IOUAmountSubmission.ts
```diff
@@ -386,6 +386,7 @@ function submitSkipConfirmationExpense(args: SubmitAmountArgs, ctx: SubmitAmount
             draftTransactionIDs: draftTransactionIDsList,
             transactionID: existingTransactionID ?? optimisticTransactionID,
             isFromGlobalCreate: getIsFromGlobalCreate(transaction),
+            isFromNativeShortcut: !!transaction?.isFromNativeShortcut,
```
marufsharifi, line 429: **Repeated "native shortcut AND global create" expression** (low)

`!!transaction?.isFromNativeShortcut && !!getIsFromGlobalCreate(transaction)` is copy-derived at several call sites that all feed `submitWithDismissFirst`'s `isFromNativeShortcut` arg:

- `handleMoneyRequestStepDistanceNavigation.ts:352` and `:429`
- `ScanSkipConfirmation.tsx:254` and `:352`
- and re-encoded in `getSubmitHandler.ts:83` as `snapshot.isFromNativeShortcut && snapshot.isFromGlobalCreate`

**Suggestion (optional):** add a `getIsFromNativeShortcut(transaction)` helper next to `getIsFromGlobalCreate` in `IOUUtils` returning `!!transaction?.isFromNativeShortcut && !!getIsFromGlobalCreate(transaction)`, and reuse it at these sites. Low priority since it's a two-term boolean.

src/libs/NativeShortcutFlow.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Tracks whether the current create flow was opened from a native home-screen shortcut deeplink.
+ * Shortcuts are the only create-expense entry point that arrives via a `Linking` URL, so detecting
+ * them here is unambiguous and free of races with draft rebuilds. The marker is set on a matching
+ * URL, consumed (read-and-cleared) by `useResetIOUType`, and also cleared by any in-app entry point
+ * (`startMoneyRequest`/`startDistanceRequest`) so FAB flows keep their original navigation.
+ */
+let isNativeShortcutFlow = false;
```
marufsharifi, line 8: **Medium: Module-level mutable singleton for cross-flow state**

This is process-global mutable state driving navigation across separate create flows. We normally reach for Onyx (or route/draft params) for anything that outlives a single call, because a global is order-dependent and prone to stale-`true` leakage: any future create entry point that neither routes through `startMoneyRequest`/`startDistanceRequest` nor `useResetIOUType` will observe whatever the last flow left behind.

The containment here is genuinely well thought out (header doc, the in-app-entry-point clears, the read-and-clear in `useResetIOUType`, plus the dedicated leakage tests), so this is a design question rather than a bug. Since the intent already ends up persisted on the draft as `isFromNativeShortcut`, could the linking layer stamp that onto the pending draft / route params directly and drop the singleton? If we keep the singleton, please keep the leakage tests — they're doing real work.

src/libs/NativeShortcutFlow.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// Matches full deeplinks and the bare route form HybridApp can deliver (no leading slash).
+// Intentionally excludes the GPS trip deeplink (`.../distance-gps`), which is not a shortcut.
+const NATIVE_SHORTCUT_DEEPLINK_PATTERN = /(?:^|\/)create\/create\/start\/[^/]+\/[^/]+\/(?:scan|manual|distance-new)\/?(?:\?.*)?$/;
```
marufsharifi, line 12: **Low: Regex is the brittle coupling point**

This hardcodes the `create/create/start/…/{scan|manual|distance-new}` shape emitted by Mobile-Expensify's `quickActionCallback`, with no type-level link to that source. If the OldDot route shape changes, detection breaks silently. The `it.each` coverage (incl. the `distance-gps` exclusion and partial-match negatives) is excellent, so this is fine to ship — but please add a comment pointing at the Mobile-Expensify code that generates these URLs so the coupling is discoverable.

src/libs/Navigation/linkingConfig/index.ts
```diff
@@ -1,19 +1,42 @@
...
 import {config} from './config';
 import prefixes from './prefixes';
 import subscribe from './subscribe';
 
+/**
+ * React Navigation's default `getInitialURL` (incl. its timeout workaround), plus marking the
+ * native-shortcut flow before the URL is handed to navigation — guaranteeing the marker is set
+ * before the create screen mounts, even on HybridApp cold starts where DeepLinkHandler runs late.
+ */
+function getInitialURL(): Promise<string | null | undefined> {
```
marufsharifi: **Low: Custom `getInitialURL` now owns behavior RN used to provide by default**

Previously no `getInitialURL` was supplied, so React Navigation's internal default (which itself does the 150ms race) was used. Overriding it means we now own that timing. It looks correct and the constant + RN-issue reference are the right call — just please verify on a real HybridApp **cold start** that the shortcut still opens, since that's exactly the case the 150ms timeout can drop (real `getInitialURL()` loses the race → resolves `undefined` → deeplink dropped).

marufsharifi (changes requested):

PR #95257, opened 2026-07-03
[No QA] Migrate GitHub Action bundling from ncc to esbuild

roryabraham:
### Explanation of Change
This PR replaces [`@vercel/ncc`](https://github.com/vercel/ncc) with [`esbuild`](https://esbuild.github.io/) for bundling our JavaScript GitHub Actions, and switches the compiled bundles from CommonJS to native ESM.

#### Why
- esbuild is ~18x faster than ncc (see benchmarks below).
- esbuild is more actively maintained.
- it lets us drop the two custom patches we carry for ncc, rather than adding a third.
- switching to ESM output lets us upgrade `@actions/core`/`@actions/github` past their CJS-only majors (both are ESM-only as of the versions used here), and fixes a real `import.meta.url` crash in `checkSVGCompression` (its `svgo`/`css-tree` dependency is pure ESM and can't be safely bundled into CJS).

**What changed:**

*Bundler swap*
- Replaced `.github/scripts/buildActions.sh` with `.github/scripts/buildActions.ts`, run directly via `bun`. The script ends up much simpler — no manual background-PID juggling to get parallelism.
- Fully removed `@vercel/ncc`: uninstalled the dependency, deleted `patches/@vercel+ncc/` (both patches plus `details.md`), removed it from `knip.json`, and scrubbed the remaining references in `verifyActions.sh` and the workflows README.
- Added an `esbuild` `devDependency` plus an `overrides` pin, and a `lefthook` `gh-actions-build` pre-commit hook — now that the build takes under a second, it's cheap enough to run on every commit that touches `.github/**/*.ts`.
- `buildActions.ts` builds **24** actions. `buildActions.sh` listed only 23 — it was missing `generateHelpPreviewComment`, so that action's committed bundle was never being rebuilt.
- Replaced `googleapis` with `@googleapis/androidpublisher` in the two Android actions. `googleapis` carries every Google API (212 MB unpacked) and is CJS, so esbuild can't tree-shake it; those two actions need only `google.auth.GoogleAuth` and `google.androidpublisher` v3. The per-API package is 2.5 MB unpacked and its `auth` export is the same `AuthPlus` instance, so the call sites are otherwise unchanged. `checkAndroidStatus` drops from 771,910 lines to 52,874 (-93%) and `getAndroidRolloutPercentage` from 764,891 to 45,855 (-94%) — both were large enough to break GitHub's diff viewer.

*ESM output*
- Switched esbuild's output `format` from `cjs` to `esm`, scoped via a new `.github/actions/javascript/package.json` (`{"type": "module"}`) so it applies to the bundles and nothing else.
- Added a `createRequire` banner so bundled CJS dependencies (`googleapis`, `lodash`, `semver`, `@babel/parser`, `@babel/traverse` — none of which ship an ESM build) can still call `require()` on Node builtins.
- Marked `svgo` `external` (not bundled) for `checkSVGCompression` only, so Node's own ESM resolution gives its `css-tree` dependency a real `import.meta.url` instead of esbuild's CJS-mode stub — this was crashing with `ERR_INVALID_ARG_VALUE` before.
- Converted all 18 action-level `if (require.main === module)` entry guards (plus `scripts/createOrUpdateDeployChecklist.ts`) to the ESM idiom (`import.meta.url === pathToFileURL(process.argv.at(1) ?? '').href`).
- Added a `no-restricted-globals` ESLint rule over `.github/actions/**` and `.github/libs/**` — the sources esbuild bundles — banning `module`, `__dirname` and `__filename`. esbuild leaves these identifiers untouched instead of failing, so a CJS idiom in a bundled source builds fine and then throws `ReferenceError: module is not defined in ES module scope` when the action runs in CI. `proposalPoliceComment` landed on main after this branch's original guard conversion and hit exactly this — its bundle threw on load, so the action would have been dead in CI. The rule is there to catch the next one at lint time. `.github/scripts/` is excluded — it runs directly under Bun, which does provide these.
- Fixed a bug this surfaced: `checkSVGCompression.ts`/`scripts/compressSvg.ts` and `bumpVersion.ts`/`scripts/bumpVersion.ts` each had their own `if (require.main === module)` guard, but esbuild flattens bundled files into one shared module scope, so **both** guards fired in each other's bundles — for `bumpVersion` this silently executed `scripts/bumpVersion.ts`'s real version-bump logic as a side effect of building/testing `checkSVGCompression`, and vice versa (confirmed via isolated repro, unrelated to CJS vs ESM). Fixed by splitting both into guard-free libraries (imported by their action + tests) plus new, never-imported `scripts/compressSvg-cli.ts` / `scripts/bumpVersion-cli.ts` entry points for CLI use. This pattern is now documented in the workflows README.

*Dependency upgrades*
- `@actions/core` 1.10.0 → 3.0.1, `@actions/github` 5.1.1 → 9.1.1, and the `@octokit/*` chain aligned with what `@actions/github` v9 bundles: `core` ^7, `plugin-paginate-rest` ^14, `plugin-throttling` ^11, `request-error` ^7, `webhooks-types` ^7.6.1, plus `plugin-rest-endpoint-methods` ^17 now declared explicitly.
- No breaking changes to `getInput`/`setOutput`/`setFailed`/`getOctokit`/`context`. The API breakages were:
  - `@octokit/plugin-throttling`'s `onAbuseLimit` renamed to `onSecondaryRateLimit`, with `retryCount` moving from `options.request.retryCount` to its own handler argument (`GithubUtils.ts`, `scripts/aggregateGitHubDataFromUpwork.ts`).
  - `@octokit/plugin-rest-endpoint-methods` no longer exports `RestEndpointMethods` directly, only `Api` (which wraps it as `{rest: RestEndpointMethods}`), so it's derived as `Api['rest']` (`GithubUtils.ts`, `isTeamMember.ts`).
- Stopped importing `@octokit/openapi-types` directly in `GithubUtils.ts`/`DeployChecklistUtils.ts`. It's only a transitive dependency, so the bare specifier resolved to whatever npm hoisted (v28) rather than the copy the endpoint methods were generated against (v27) — which made otherwise identical types mutually unassignable (`issue.id` is `number` in one and `number | bigint` in the other). These aliases now derive from `RestEndpointMethodTypes`, which always matches the installed endpoint methods.

*Test adaptations for `@octokit` v7*
These are the reason the dependency upgrade couldn't just be a version bump. All of these tests pass on main and broke silently on the upgrade — they kept passing type checks while firing real HTTP requests at `api.github.com`.
- `@octokit/plugin-rest-endpoint-methods` v17 exposes each REST namespace (`octokit.rest.issues`, `.pulls`, `.actions`, `.teams`, `.repos`) as a **`Proxy`** that creates endpoint methods on first access. Bun's `jest.spyOn` silently no-ops on a Proxy-backed property — it returns a mock, but the property keeps its original value — so mocks never intercepted. That showed up as `HttpError: Bad credentials` (401) in several files, and as an outright hang in `DeployChecklistUtils` once the real retry backoff started running. Added `tests/utils/materializeOctokitNamespace.ts` to copy a namespace into a plain object first, and applied it at each spy site.
- `@octokit/request` v10 reads a response body with `text()` and parses it itself rather than calling `json()`, so `postOrReplaceComment`'s stubbed `Response` returned no comments. It now stubs `text()`.

*Also in here*
- `newComponentCategory.ts` derives `WebhookPayload` from `context.payload`: `@actions/github` v9 dropped the `./lib/interfaces` deep import path from its exports map, so the type can't be imported directly any more.

#### Benchmarks

Measured back to back on the same machine today, so the two columns are directly comparable (`main` at `0932767` for the ncc baseline, this branch for esbuild):

| Metric | ncc (main) | esbuild (this PR) | Delta |
|--------|-----------|-------------------|-------|
| Median wall time (5 runs) | 13.34s | **0.45s** | **-96.6%** (~30x faster) |
| Best run (min) | 13.13s | 0.31s | -97.6% |
| Worst run (max) | 14.20s | 0.56s | -96.1% |
| Actions built | 23 | 24 | +1 (`generateHelpPreviewComment`) |
| Total `index.js` size | 83.57 MB | **29.88 MB** | **-64.2%** |
| Largest single bundle | 771,910 lines | 69,843 lines | -91.0% |
| Machine | macOS arm64, Apple M5 Max, Node 26.5.0 | (same) | |

Most of the size and a good chunk of the time come from dropping the `googleapis` mega-package rather than from the bundler swap itself. On esbuild with `googleapis` still in place the totals were 83.13 MB and 0.75s median, so the bundler change alone is roughly size-neutral against ncc; the per-API import is what makes it smaller.

Two notes on reading this table honestly: esbuild builds one more action than ncc did, which very slightly flatters it; and rebuilding with ncc on `main` leaves `isDeployChecklistLocked/index.js` and `proposalPoliceComment/index.js` dirty, i.e. those two committed bundles are already stale on main today. Neither affects the conclusion.

### Fixed Issues
$
PROPOSAL:

### Tests

Requires Node 26.5.0 / npm 11.17.0 (per `engines`).

1. Run `npm install` in a clean checkout.
2. Run `npm run gh-actions-build` and confirm all 24 actions build without error in well under two seconds.
3. Run `./.github/scripts/verifyActions.sh` and confirm it reports "Github Actions are up to date!".
4. Confirm every directory under `.github/actions/javascript/` contains exactly one `index.js` and no extra chunk files:
   ```sh
   find .github/actions/javascript -mindepth 2 -name '*.js' ! -name index.js   # expect no output
   ```
5. Confirm no bundle throws at module scope (this is the failure mode a stray CJS idiom causes — the action dies before `run()` is ever reached):
   ```sh
   for d in .github/actions/javascript/*/; do
       out=$(env -u GITHUB_TOKEN -u SEMVER_LEVEL node "$d/index.js" 2>&1)
       echo "$out" | grep -qE 'ReferenceError|SyntaxError|ERR_MODULE_NOT_FOUND|ERR_INVALID_ARG_VALUE' && echo "FAIL: $d"
   done
   ```
   Expect no `FAIL:` lines. Each action should instead fail on a missing input, which means it loaded and ran.
6. Confirm `node .github/actions/javascript/checkSVGCompression/index.js` with no `GITHUB_TOKEN` set fails with `Error: Input required and not supplied: GITHUB_TOKEN` (not a crash).
7. Confirm `node .github/actions/javascript/bumpVersion/index.js` with no `SEMVER_LEVEL` set fails with `Error: Input required and not supplied: SEMVER_LEVEL`, and that `git status` shows no unexpected file changes afterwards (i.e. it didn't actually bump the version as a side effect).
8. Run `npm run compress-svg` and confirm it runs to completion and reports compression results (rather than erroring or silently doing nothing).
9. Run `npm run test:bun` and confirm it passes. The `tests/tooling/` portion is what covers `.github/` and `scripts/`; note these files depend on per-file isolation, so run them via that script (`--parallel`) rather than pointing `bun test` at the directory yourself.
10. Run `npm run typecheck` and `npx eslint .github/ scripts/ tests/tooling/` and confirm both are clean.
11. Confirm the new ESLint guard works: append `if (require.main === module) { run(); }` to any file under `.github/actions/javascript/`, run `npx eslint` on it, and confirm it reports a `no-restricted-globals` error pointing at the ESM idiom. Revert afterwards.

Note that several of these bundled actions are exercised on this PR itself before merge, at least:

- `isAuthorizedContributor`
- `verifySignedCommits`
- `authorChecklist`

and `Validate GitHub Actions` runs the build step in CI.

### Offline tests
N/A - this PR only changes the GitHub Actions build tooling, not app code.

### QA Steps
N/A - this PR only changes internal CI/build tooling; there is no end-user-facing or staging/production behavior to QA.

### PR Author Checklist
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

N/A - build tooling change only, no app UI affected.

</details>

<details>

N/A - build tooling change only, no app UI affected.

</details>

<details>

N/A - build tooling change only, no app UI affected.

</details>

<details>

N/A - build tooling change only, no app UI affected.

</details>

<details>

N/A - build tooling change only, no app UI affected.

</details>


roryabraham:
converting to draft while I look into the codex comment

roryabraham:
@codex review

roryabraham:
@MelvinBot review

package.json
```diff
@@ -223,8 +224,8 @@
     "xlsx": "^0.18.5"
   },
   "devDependencies": {
-    "@actions/core": "1.10.0",
-    "@actions/github": "5.1.1",
+    "@actions/core": "3.0.1",
+    "@actions/github": "9.1.1",
```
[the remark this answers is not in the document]
roryabraham: Fixed in 12ca584f82566b8b03da3352994572dce742a241: rather than routing this through the new ESM loader, `createOrUpdateDeployChecklist.ts` now runs directly under Bun (already a pinned local dependency used elsewhere as a script runner, e.g. `gh-actions-build`), which loads ESM-only packages natively. Updated the shebang to `#!/usr/bin/env bun`, swapped the `require.main`/`module` entry guard for the `import.meta.url === pathToFileURL(...)` pattern already used by `awaitStagingDeploys.ts`, and switched `createDeployChecklist.yml`'s `run:` step from `npx ts-node` to `npx bun`. Verified locally that the script now gets past the import stage and actually invokes `run()` (fails only on the network call with an invalid token).

roryabraham:
@MelvinBot Addressed all three points in 12ca584f82566b8b03da3352994572dce742a241:

**1. Jest shims going stale** — rather than adding a CI staleness guard, fully eliminated the shims instead, since removing them entirely was a core goal of the Jest→Bun migration and a guard would just enshrine keeping them forever. Found and removed the last three Jest-side consumers that transitively pulled in `@actions/core`/`@actions/github`/`@octokit/*`:
- Extracted the pure `@babel/parser`/`@babel/traverse` AST logic (`detectReactComponent`) out of `newComponentCategory.ts` into its own shim-free module — `detectReactComponent.test.ts` no longer touches `@actions/github` at all.
- Extracted the pure `getMergedPR` helper out of `failureNotifier.ts` for the same reason.
- Migrated `tests/unit/GitUtilsTest.ts` to `bun:test` (it has zero `jest.mock`/`jest.spyOn` usage, so this was a low-risk move), since `GitUtils` transitively imports `@actions/core` via `GithubUtils`.
- Dropped the now-dead `@actions/core` import + 5 `jest.spyOn` calls from `jest/setup.ts`.

With those gone, deleted `jest/shims/*.cjs`, `jest/buildEsmShims.ts`, the `build-jest-shims` script, the 6 shim `moduleNameMapper` entries in `jest.config.js`, and the related `knip.json`/`.oxfmtrc.json`/`config/eslint/eslint.config.mjs` ignore entries. Verified with a full Jest run, the full `test:bun:scripts` suite, `npm run typecheck`, `npx oxfmt --check .`, and `npx knip` — no stale bundles, no new unused-file warnings, and nothing in the Jest suite calls `@actions/core` anymore.

**2. `esbuild` in both `devDependencies` and `overrides`** — confirmed intentional: it's pinned in `overrides` specifically to dedupe the version across the transitive tree (webpack/storybook/vite tooling that each bundle their own `esbuild`), so everything resolves to the same version we test against.

**3. Stale `buildActions.ts` comment** — fixed; it now only references `scripts/compressSvg.mts`.

roryabraham:
@codex review

roryabraham:
@codex review

roryabraham:
@codex review

roryabraham:
@Julesssss opening this up for review.

.github/actions/javascript/checkAndroidStatus/checkAndroidStatus.ts
```diff
@@ -2,20 +2,23 @@ import CONST from '@github/libs/CONST';
 import GithubUtils from '@github/libs/GithubUtils';
 
 import * as core from '@actions/core';
-import {google} from 'googleapis';
+// Import just the one Google API we use rather than the `googleapis` mega-package, which carries every
+// Google API and bundles to ~770k lines per action. The `auth` export is the same AuthPlus instance that
+// the mega-package exposes under its `google` namespace.
+import {androidpublisher, auth as googleAuth} from '@googleapis/androidpublisher';
```
Julesssss, line 8: Bonus fix right?

.github/actions/javascript/authorChecklist/authorChecklist.ts
```diff
@@ -158,7 +159,7 @@ async function generateDynamicChecksAndCheckForCompletion() {
     }
 }
 
-if (require.main === module) {
+if (import.meta.url === pathToFileURL(process.argv.at(1) ?? '').href) {
```
Julesssss: Maybe this could be extracted and reused as its a bit more complex than `require.main === module` now

.github/scripts/buildActions.ts
```diff
@@ -0,0 +1,98 @@
...
+import {build} from 'esbuild';
+import {readFile, writeFile} from 'fs/promises';
+import path from 'path';
+
+const REPO_ROOT = path.resolve(__dirname, '..', '..');
+const ACTIONS_DIR = path.join(REPO_ROOT, '.github', 'actions', 'javascript');
+const TSCONFIG = path.join(REPO_ROOT, '.github', 'tsconfig.json');
+
+// List of paths to all JS files that implement our GH Actions
+const GITHUB_ACTIONS = [
```
Julesssss: This seems kinda hidden right? Is there another way to register GH Actions so we don't have to remember to declare them here?

Julesssss (commented):
Good improvement, some minor comments

.github/actions/javascript/checkAndroidStatus/checkAndroidStatus.ts
roryabraham: yes and no. Yes in that it could have been done separately on top of ncc. No in that without this change, GitHub couldn't handle the diff and the PR viewer was totally broken. I guess ~`+750,000 -2,000,000` is ok, but ~`+2,000,000 -2,000,000` is not

.github/scripts/buildActions.ts
roryabraham: This is how it worked with the script before, but now that we're using TS it should be more ergonomic to do action discovery.

roryabraham: Done — `buildActions.ts` now discovers every `.github/actions/javascript/<name>/<name>.ts` entry instead of keeping a hard-coded list.

.github/actions/javascript/authorChecklist/authorChecklist.ts
roryabraham: Actions already run on `node24`, so I switched the ESM entry guards to `import.meta.main` instead of extracting a helper. Same idea as `require.main === module`, just the Node 24 ESM equivalent.

.github/actions/javascript/checkAndroidStatus/checkAndroidStatus.ts
Julesssss: Oh fair enough then

.github/scripts/buildActions.ts
Julesssss: Nice ty

Julesssss (approved):

Issue #83071, opened 2026-02-20
InteractionManager migration follow-up


blazejkustra:
- Remove patch for `InteractionManager`
- Verify that no other libraries use `InteractionManager`
- Verify there are no other `InteractionManager` usages
- Update/Remove INTERACTION_MANAGER.md document

collectioneur:
Copying this here for visibility:

@mrejdak analyzed and summarized the remaining usages in the codebase (which were added after this issue was created). We will prepare PRs and pin them to the [InteractionManager migration follow-up](https://github.com/Expensify/App/issues/83071) issue. Once all remaining usages have been migrated, we can close this issue 🙂 

<details>
<summary><b>InteractionManager.runAfterInteraction usages</b></summary>


| # | File | Line | Code / usage | Why it was added | PR that first introduced it |
|---|------|------|--------------|------------------|-----------------------------|
| 1 | [src/components/ContactPermissionModal/index.native.tsx](src/components/ContactPermissionModal/index.native.tsx#L41) | 41 | `InteractionManager.runAfterInteractions(() => {` | Defers requesting the OS contact permission (`onGrant`) until after the soft-permission modal finishes closing. | [#51018](https://github.com/Expensify/App/pull/51018) — Contacts import |
| 2 | [src/components/ContactPermissionModal/index.native.tsx](src/components/ContactPermissionModal/index.native.tsx#L58) | 58 | `InteractionManager.runAfterInteractions(() => {` | Defers re-focusing the text input until the modal has finished closing (input otherwise focuses without the keyboard appearing). | [#64207](https://github.com/Expensify/App/pull/64207) — fix: contact list search field not auto focused |
| 3 | [src/components/Reactions/ReportActionItemEmojiReactions.tsx](src/components/Reactions/ReportActionItemEmojiReactions.tsx#L86) | 86 | `InteractionManager.runAfterInteractions(() => {` | Defers signing out / redirecting anonymous users to sign-in until the context menu is hidden and interactions settle, avoiding UI jank. | [#18293](https://github.com/Expensify/App/pull/18293) — Open report for anonymous users |
[...]

PR #95391, opened 2026-07-06
Migrate useSingleExecution off InteractionManager

collectioneur:
### Explanation of Change

Migrates native `useSingleExecution` off `InteractionManager.runAfterInteractions` onto a new heuristic helper, `runAfterPredictedTransition`, and fixes two `TransitionTracker` / `ScreenLayout` bugs that made that wait unreliable.

**Bug fixes in `transitionStart` / `transitionEnd` tracking**
1. Tab transitions were unreliable on Android for every tab, and on iOS for the Home tab specifically. Besides Home, the other bottom-tab routes host nested split navigators whose screens register via stack `screenLayout`, but that registration relies on `transitionStart/transitionEnd`, which `react-native-screens` only fires when tied to an actual view animation. Since tab switches use animation: 'none', no such animation ever runs on Android, so the nested-stack signal never fires there either, it just happened to look fine on iOS, where `viewWillAppear` fires on any visibility change regardless of animation. Home was simply the most visible case on any platform, since it has no nested navigator and thus no `screenLayout` wiring at all.
Fixed by wiring `bottomTabScreenLayoutWrapper` directly into `TabNavigator`'s own `screenLayout`, so tracking now comes from bottom-tabs' own `transitionStart/transitionEnd`, which fires consistently on both platforms regardless of the animation option.

4. **Unmount can leave a stuck open transition.** A screen may emit `transitionStart` and then unmount before `transitionEnd`; cleanup unsubscribes the listeners, so `transitionEnd` never arrives and queued callbacks wait up to `MAX_TRANSITION_DURATION_MS` (1s). Fixed by calling `TransitionTracker.endTransition` on unmount whenever a handle is still open.

**`runAfterPredictedTransition` (prediction mechanism)**
Naive `TransitionTracker.runAfterTransitions()` right after a press usually sees zero active transitions (registration gap) and unlocks too early. Always using `waitForUpcomingTransition: true` would make non-navigating presses wait up to 1s. The new helper predicts whether the current press will cause a visual transition by watching `navigationRef` `__unsafe_action__` + `state`: if a navigable action is followed by a focused-route change within `NAVIGATION_PREDICTION_WINDOW_MS` (150ms), it waits for that upcoming transition to finish; otherwise it settles immediately. `useSingleExecution` then re-enables the button after the predicted (or already-active) transition ends - or right away when no transition is predicted.

### Fixed Issues

$ https://github.com/Expensify/App/issues/71913
$ https://github.com/Expensify/App/issues/83071

### Tests

Tests (on iOS and Android):

1. Go to Settings. You'll see a list of settings items.
2. As fast as you can, tap one of the list buttons, then immediately tap any other one.
5. Verify that no double navigation happens, only the first navigation that occurred goes through.

---

1. Create a workspace.
2. Go to the workspace, you'll see another list of buttons leading to different workspace pages.
3. Double-tap them the same way as above.
6. Verify that only the first navigation triggers and everything works as expected.

---

1. Go to any chat and send a message.
2. Start spamming reactions on that message: quickly add a reaction, reopen the reaction modal, add the next one, and repeat.
3. Verify that emojis are applied correctly, i.e. the reaction picker isn't left blocked when you reopen the modal.

---

1. Find any back button (top-left).
2. Double-tap that button.
3. Verify that only a single back navigation happens.

### Offline tests
N/A 

### QA Steps
Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/7caad1c0-dca5-4785-b9cb-16cb1f82b556

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/7a30882d-11c2-45ba-961d-f6ee75aa53d9

</details>

<details>

</details>

<details>

</details>


src/hooks/useSingleExecution/index.native.ts
```diff
@@ -24,15 +26,19 @@ export default function useSingleExecution() {
...
+                                setIsExecuting(false);
+                                return;
+                            }
+                            execution.finally(() => {
+                                setIsExecuting(false);
+                            });
+                        },
                     });
-                });
+                }, CONST.TIMING.SINGLE_EXECUTION_DEBOUNCE_TIME);
```
WojtekBoman: NAB: Add here a comment explaining why we need to add it

src/hooks/useSingleExecution/index.native.ts
```diff
@@ -24,15 +26,19 @@ export default function useSingleExecution() {
...
 
                 const execution = action(...params);
-                InteractionManager.runAfterInteractions(() => {
-                    if (!(execution instanceof Promise)) {
-                        setIsExecuting(false);
-                        return;
-                    }
-                    execution.finally(() => {
-                        setIsExecuting(false);
+                setTimeout(() => {
```
WojtekBoman: Nice cleanup replacing `InteractionManager.runAfterInteractions` here 👏  

One small gap worth closing while you're in this code: this `setTimeout` never gets its ID stored, so if the component unmounts before the 500ms fires, the timer still runs and calls `setIsExecuting(false)` on an unmounted component. It worked the same way in the previous version, but since you're already touching this code, it'd be worth storing the timer ID in `useRef` and clearing it in a `useEffect` cleanup on unmount.

collectioneur: Addressed, thanks!

roryabraham:
posted a discussion on an alternate solution: https://expensify.slack.com/archives/C01GTK53T8Q/p1783451929635129

JmillsExpensify (commented):
No product review required.

src/hooks/useSingleExecution/index.native.ts
```diff
@@ -24,15 +39,21 @@ export default function useSingleExecution() {
...
+                                setIsExecuting(false);
+                                return;
+                            }
+                            execution.finally(() => {
+                                setIsExecuting(false);
+                            });
+                        },
                     });
-                });
+                }, CONST.TIMING.SINGLE_EXECUTION_DEBOUNCE_TIME);
```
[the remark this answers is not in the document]
Pujan92: Valid point, I think pressing buttons quickly in the BigNumberPad won't be considered all pressed keys exactly due to the 500 ms lockout period.

src/hooks/useSingleExecution/index.native.ts
```diff
@@ -24,15 +39,21 @@ export default function useSingleExecution() {
...
-                        setIsExecuting(false);
-                        return;
-                    }
-                    execution.finally(() => {
-                        setIsExecuting(false);
+                // The timeout is a minimum debounce applied to every press; checking TransitionTracker afterwards
+                // is an extra safety net in case a transition is still in progress once the debounce ends.
+                timeoutRef.current = setTimeout(() => {
+                    transitionHandleRef.current = TransitionTracker.runAfterTransitions({
+                        callback: () => {
```
Pujan92:
```suggestion

                  transitionHandleRef.current = TransitionTracker.runAfterTransitions({
                  waitForUpcomingTransition: true,
                  upcomingTransitionTimeoutMs: CONST.TIMING.SINGLE_EXECUTION_TRANSITION_WAIT_TIME, // ~150 for non-nav taps
                      callback: () => {
```

shall we use waitForUpcomingTransition and new prop with a lower timeout value?

collectioneur: I looked into that, but unfortunately, it won't quite work. On slower devices, the transition can sometimes take more than 150ms to register, which would bring back the exact regression that `useSingleExecution` was originally meant to fix.

Instead, I went with a slightly different approach: I added a new prop to `GenericPressable` that disables `useSingleExecution` altogether, and I've applied it to all buttons where users might need to tap rapidly. I feel this is a cleaner solution because it explicitly separates rapid-tap buttons from regular ones, while still safely waiting for transitions when needed. It seems like a reasonable trade-off.

As for using the new prop, that's a great suggestion, thanks! I've added the updated `TransitionTracker` code from this [PR](https://github.com/Expensify/App/pull/95397) and applied the changes (removed `setTimeout` and added a timeout prop instead).

The PR is ready for another look 🙌

roryabraham (changes requested):
Let's step back ... it seems like we are headed towards a more robust solution in [this thread](https://expensify.slack.com/archives/C01GTK53T8Q/p1783451929635129), without any delay/throttling issues, or needing to manually track the buttons for which a double-tap might trigger a double-navigation.

Meanwhile, the current solution using the (albeit deprecated) InteractionManager doesn't have these pitfalls either.

So I suggest that we close this pull request, and instead pursue the better solution described [here](https://expensify.slack.com/archives/C01GTK53T8Q/p1783451929635129), keeping the deprecated `InteractionManager-based` solution in the meantime.

collectioneur:
@roryabraham Yeah, I agree with you, it’s not a clean migration. I’ve tried to add `unsafe_event` first, but it turns out, that the commit adding `unsafe_event` (https://github.com/react-navigation/react-navigation/commit/d22044d849a984b61a28282a575e54b3ea4b88a7) was actually backported last month to `@react-navigation/core@7.21.0`.

Since we are currently on version `7.16.1`, we'd need to update react-navigation first to use this in `TransitionTracker`. This would require a fair amount of effort, as we'll have to manually resolve some of the patches due to underlying logic changes before we can implement the event logic itself.

I think it might be a good idea to spin up a quick test app first, just to make sure this approach works for us. If it looks good, we can go ahead with the update.

I have some availability right now, so I’d be happy to take this on and start the update if we want to go that route 🙌

src/CONST/index.ts
```diff
@@ -260,6 +260,7 @@ const CONST = {
     COMPOSER_FOCUS_DELAY: 150,
     MAX_TRANSITION_DURATION_MS: 1000,
     MAX_TRANSITION_START_WAIT_MS: 1000,
+    NAVIGATION_PREDICTION_WINDOW_MS: 150,
```
war-in, line 263: I'd verify if 150ms is enough even on low-end android devices

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,122 @@
...
+import {findFocusedRoute} from '@react-navigation/native';
+
+import type {CancelHandle} from './TransitionTracker';
+
+import navigationRef from './navigationRef';
+import TransitionTracker from './TransitionTracker';
+
+// Action types to ignore when predicting an upcoming visual transition. SET_PARAMS/REPLACE_PARAMS/PRELOAD
+// never move the focused route. JUMP_TO can, but only tab navigators use it and they do not emit
+// transitionStart/transitionEnd, so waiting on TransitionTracker would never settle.
```
war-in: I wonder if we should try to support `JUMP_TO` in TransitionTracker 🤔 From what the comment says, it's not an issue at the moment, but we could be future-proof and handle it right away
Of course, we should do it in a separate issue

Let me know if that's worth investigating @roryabraham @collectioneur

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,122 @@
...
+const ACTION_TYPES_WITHOUT_TRANSITION: ReadonlySet<string> = new Set(['SET_PARAMS', 'REPLACE_PARAMS', 'PRELOAD', 'JUMP_TO']);
+
+let isTransitionPending = false;
+let pendingClearTimeout: ReturnType<typeof setTimeout> | null = null;
+let lastFocusedRouteKey: string | undefined;
+let hasConfirmedFocusMove = false;
+let settleWaiters: Array<(shouldWait: boolean) => void> = [];
```
war-in, line 44: I'd add descriptions to those variables and functions below to make their purpose clear for reviewers. Actually, I think this file should be thoroughly explained in comments

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (hasConfirmedFocusMove) {
+        onSettled(true);
+        return;
+    }
+    settleWaiters.push(onSettled);
+}
+
+navigationRef.addListener('__unsafe_action__', (event) => {
+    if (event.data.noop || ACTION_TYPES_WITHOUT_TRANSITION.has(event.data.action.type)) {
+        return;
```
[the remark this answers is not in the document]
collectioneur: The tab-level navigator and the nested stack navigator are two different navigators. In my tests, the nested navigator always emits transitionStart and transitionEnd later than the tab navigator. That's why I disabled all tabs except for Home

war-in (commented):
Changes lgtm, left a few comments

src/libs/Navigation/runAfterPredictedTransition.ts
collectioneur: Added explanatory comments throughout, thanks!

src/CONST/index.ts
collectioneur: I’ve investigated `runAfterPredictedTransition` on a lower-end device (Motorola Moto G15), and the results are pretty interesting.

The registered 150ms safety timeout is blocked and does not run while the js thread is busy with react’s state update and commit. After react commits, that overdue timeout races the state listener: react navigation emits the new state from `useEffect` (after commit), so once the thread is free again the timeout can fire either before or after the listener sees the update.

I fixed this by adding an extra check when the 150ms timeout runs: before clearing the prediction, it checks whether the navigation ref already has the new focused route. That check is reliable for normal navigations, because the ref store is updated during react navigation `setState`, before commit, so the new route is already visible even if the state listener has not run yet.

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,208 @@
...
+});
+
+// Phase 3: a real transition started - prediction state is obsolete.
+TransitionTracker.onTransitionStart(clearPending);
+
+/**
+ * Heuristic-aware drop-in for `TransitionTracker.runAfterTransitions({callback, waitForUpcomingTransition: true})`.
+ * Call this right after or right before the navigation call whose transition should be waited on.
+ *
+ * ponytail: best-effort guess based on actions dispatched through `navigationRef` shortly around this call,
```
JakubKorytko: "ponytail" looks like a typo - shouldn't "note" or "caveat" be here instead?

src/libs/Navigation/AppNavigator/Navigators/TabNavigator.native.tsx
```diff
@@ -95,6 +96,7 @@ function TabNavigator() {
             backBehavior="fullHistory"
             tabBar={renderTabBar}
             screenOptions={screenOptions}
+            screenLayout={bottomTabScreenLayoutWrapper}
```
JakubKorytko, line 99: the PR description mentions wiring this with a focus guard so nested tab routes do not double-register, but `screenLayout` is applied on `Tab.Navigator` for every tab. Nested tabs already get stack `screenLayout`, and tab switches use `animation: 'none'`, so it is unclear what extra transitions this adds outside Home. If Home is the only tab that mounts a screen directly, consider scoping `bottomTabScreenLayoutWrapper` to the Home `Tab.Screen` instead of the whole navigator

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,208 @@
...
+ * not a guarantee. A transition from a disconnected call stack (native gesture, unrelated code) is invisible
+ * here, same as calling `TransitionTracker.runAfterTransitions` with no `waitForUpcomingTransition`.
+ */
+export default function runAfterPredictedTransition(callback: () => void | Promise<void>): CancelHandle {
+    let cancelled = false;
+    let innerHandle: CancelHandle | null = null;
+
+    // Yield one macrotask so a navigation dispatch in the same press handler can register first.
+    setTimeout(() => {
```
JakubKorytko: `cancel()` sets a flag and cancels the inner `TransitionTracker` handle, but the outer `setTimeout(..., 0)` is never cleared. It is harmless today because the callback no-ops when `cancelled` is true, but storing the timeout id and calling `clearTimeout` in `cancel()` would make the CancelHandle fully abort work (and avoids running `whenPredictionSettled` after unmount)

src/hooks/useSingleExecution/index.native.ts
```diff
@@ -24,7 +33,9 @@ export default function useSingleExecution() {
                 isExecutingRef.current = true;
 
                 const execution = action(...params);
-                InteractionManager.runAfterInteractions(() => {
+                // Re-enables the button once the predicted (or actual) transition triggered by this press
+                // ends - or immediately, if the press wasn't predicted to cause one.
+                transitionHandleRef.current = runAfterPredictedTransition(() => {
```
JakubKorytko, line 38: unmount cleanup cancels the transition handle, but if `action()` returned a Promise, `execution.finally` inside this callback can still call `setIsExecuting(false)` afterward. That is the same class of issue as the old uncancelled debounce timer. A small mounted/ref guard around the `setIsExecuting(false)` paths would prevent a post-unmount state update when the async action settles late

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,208 @@
...
+// Phase 2: confirm or deny the pending prediction by comparing focused route keys.
+navigationRef.addListener('state', () => {
+    const focusedRouteKey = getHydratedFocusedRouteKey();
+
+    // Missing keys are inconclusive - do not confirm a focus move or clear a pending prediction.
+    if (focusedRouteKey === undefined) {
+        return;
+    }
+
+    const didFocusedRouteChange = focusedRouteKey !== lastFocusedRouteKey;
```
JakubKorytko, line 166: when `lastFocusedRouteKey` is still `undefined`, the first hydrated key always looks like a focus change (`undefined !== 'route-a'`). That is probably fine once the app has emitted a baseline state event, but on a cold start where the first `state` event arrives after a press, this can confirm a focus move even if the action did not actually move focus yet. 

Worth a test (or an explicit "baseline captured" flag) so we do not regress the double-tap fix on slow initial hydration

tests/unit/Navigation/runAfterPredictedTransitionTest.ts
```diff
@@ -0,0 +1,258 @@
...
+    it('runs the callback immediately when no navigation action was dispatched', () => {
+        const callback = jest.fn();
+        runAfterPredictedTransition(callback);
+        flushPredictionTick();
+
+        expect(runAfterTransitionsSpy).toHaveBeenCalledWith(expect.objectContaining({callback, waitForUpcomingTransition: false}));
+        expect(callback).toHaveBeenCalledTimes(1);
+    });
+
+    it('waits for an upcoming transition when an action is followed by a focus change', async () => {
```
JakubKorytko, line 105: this covers the happy path where a focus change is followed by a real `transitionStart`. There is still no test for a focus change with no upcoming transition (for example a tab switch that changes the focused leaf route but never emits `transitionStart`). That path falls through to `waitForUpcomingTransition: true` and then the 1s TransitionTracker timeout. A test here would document the expected behavior and catch accidental 1s button lockouts

jest/setupAfterEnv.ts
```diff
@@ -6,13 +6,24 @@ import type {KeyboardEventName} from 'react-native';
...
 jest.mock('@hooks/useAIFeaturesPromoModal', () => ({
     __esModule: true,
     default: jest.fn(),
 }));
 
+// The real hook relies on real navigation transitions to settle `isExecuting` (see
+// `runAfterPredictedTransition`/`TransitionTracker`), which don't happen in unit tests and would
+// otherwise leave buttons stuck disabled. Mocked globally so no test needs to import the navigation
+// listener machinery just because it renders a Pressable-based component (Button, MenuItem, etc.).
+jest.mock('@hooks/useSingleExecution', () => ({
```
JakubKorytko, line 22: globally mocking `useSingleExecution` makes sense for unit tests, but it means nothing in Jest exercises the native hook wiring end to end. The new `runAfterPredictedTransition` tests are thorough; a small native-hook test (even with mocked navigation listeners) would give extra confidence that `isExecuting` clears on both sync and async actions

src/libs/Navigation/TransitionTracker.ts
```diff
@@ -184,10 +190,22 @@ function runAfterTransitions({
...
+ * Returns an unsubscribe function.
+ */
+function onTransitionStart(listener: () => void): () => void {
+    transitionStartListeners.add(listener);
+    return () => {
+        transitionStartListeners.delete(listener);
+    };
+}
+
 const TransitionTracker = {
```
JakubKorytko, line 236: new `onTransitionStart` API drives prediction clearing in `runAfterPredictedTransition`, but `TransitionTrackerTest` does not cover listener notification or error isolation through this path (only the renamed `invokeSafely` messages changed). A small test that `startTransition` invokes subscribers would protect this hook from regressions

JakubKorytko (commented):
nice direction overall! most of my comments are around scoping tab-level tracking, a couple edge cases in the prediction helper, and test coverage gaps

src/hooks/useSingleExecution/index.native.ts
collectioneur: I looked into this but I don't think a mounted guard buys us anything here. We're on React 19, where calling a state setter on an unmounted component is a no-op

src/libs/Navigation/AppNavigator/Navigators/TabNavigator.native.tsx
collectioneur: It turns out that we need to track all tabs. Fixed the description and explained why, thanks!

src/libs/Navigation/runAfterPredictedTransition.ts
collectioneur: I’ve investigated this for a bit and confirmed that it’s possible to track tab transitions. Although we don’t need to remove `JUMP_TO` from here since `startTransition` for tabs will be registered synchronously with dispatch. I’ve opened a [draft](https://github.com/Expensify/App/pull/96476) so we can add this once this PR gets merged 🙂

src/libs/Navigation/runAfterPredictedTransition.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (!didFocusedRouteChange) {
+        clearPending();
+        return;
+    }
+
+    confirmFocusMove(focusedRouteKey);
+});
+
+// Phase 3: a real transition started - prediction state is obsolete.
+TransitionTracker.onTransitionStart(clearPending);
```
Pujan92: Isn't onTransitionStart (or clearPending here) gated to only 'navigation'-kind starts?

JakubKorytko: It wasn't, good catch. `onTransitionStart` notified on every startTransition, so an 'other' start (keyboard, modal, layout) could clear a pending prediction before the screen transition registered. The waiter then settled with shouldWait: false and got flushed as soon as that unrelated animation ended, which is exactly the early unlock we're trying to avoid. Fixed it: `onTransitionStart` now passes the kind to listeners and we only clearPending on 'navigation' starts. Added tests for the kind arg in `TransitionTrackerTest` and for 'a non-navigation start keeps the prediction pending' in `runAfterPredictedTransitionTest`.

roryabraham (approved):

Pujan92 (approved):
LGTM!

abzokhattab:
**BugZero Comment**

This PR migrated `useSingleExecution` from `InteractionManager.runAfterInteractions` to `TransitionTracker`-based `runAfterPredictedTransition`. The new mechanism sets `isExecuting: true` for the duration of screen transitions, which propagates through `disabled` props to menu rows — causing them to grey out during the iOS swipe-back gesture.

The previous `InteractionManager` approach resolved quickly enough that no visible dimming occurred. With `TransitionTracker`, the transition window is longer, and settings/workspace sidebar menus that fold `isExecuting` into `disabled` now visibly grey out every row during the push/swipe animation.

Issue: https://github.com/Expensify/App/issues/98493
Fix: https://github.com/Expensify/App/pull/99169 and https://github.com/Expensify/App/pull/99405

src/hooks/useSingleExecution/index.native.ts
```diff
@@ -10,9 +11,17 @@ type Action<T extends unknown[]> = (...params: T) => void | Promise<void>;
 export default function useSingleExecution() {
     const [isExecuting, setIsExecuting] = useState(false);
     const isExecutingRef = useRef<boolean | undefined>(undefined);
+    const transitionHandleRef = useRef<CancelHandle | null>(null);
 
     isExecutingRef.current = isExecuting;
 
+    useEffect(
+        () => () => {
+            transitionHandleRef.current?.cancel();
```
shubham1206agra, line 20: You forgot to reset the execution state on unmount which caused https://github.com/Expensify/App/issues/99203

Issue #95705, opened 2026-07-09
Sign in - Single button press sends 4-5 magic code emails


grgia:
**Version Number:** 9.4.32-0
**Reproducible in staging?:** Yes
**Reproducible in production?:** Unknown, believed to be a staging regression
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:** N/A
**Email or phone of affected tester (no customers):** georgia+merge5@expensifail.com
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @grgia
**Slack conversation:** [#fireroom-2026-07-08-notifications-broken](https://expensify.slack.com/archives/fireroom-2026-07-08-notifications-broken)

## Action Performed:
1. Go to staging.new.expensify.com while signed out
2. Enter an email address and press the sign in button once
3. Check the inbox for that account

## Expected Result:
One magic code email is sent

## Actual Result:
4-5 magic code emails arrive for the single button press. Logs confirm the client sends one `BeginSignIn` followed ~1 second later by 4-5 parallel `RequestNewValidateCode` API calls (within 5ms of each other, same deviceID). The backend correctly sends one email per request, so the duplication is client-side. The `resendValidateCode` throttle (maxAttempts 5) is the only thing capping the email count.

## Workaround:
Yes, sign in works normally with the most recent code. The extra emails are noise.

## Platforms:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [ ] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/38a22388-7b0e-4308-8818-c439904bfb12

MelvinBot:
## Proposal

[...]

PR #95148, opened 2026-07-02
[No QA] Fix leader-election deadlock in ActiveClientManager (SequentialQueue unable to flush)

elirangoshen:
### Explanation of Change

The web/PWA app could intermittently become unresponsive after focusing the window: the Inbox loading bar loops forever, no new data arrives, and only a full sign-out/sign-in recovers it (`Clear cache and restart` does not help).

Root cause is a leader-election deadlock in `ActiveClientManager`. Leadership is decided purely by "last GUID in the shared `ACTIVE_CLIENTS` list wins". The only cleanup path is a `beforeunload` handler, which doesn't fire on crashes, force-quit, or mobile/PWA background kills — so a **stale/ghost GUID** from a dead tab can remain in the persisted list. When that ghost sits last, the only live tab believes it is not the leader, so `SequentialQueue.flush()` bails and `OpenApp`/`ReconnectApp`/`PusherPing` pile up forever with no leader to flush them. `ACTIVE_CLIENTS` is persisted, which is why cache-clear doesn't recover but sign-in (which resets the list) does.

This PR fixes it by never dropping the live client (`ActiveClientManager`): a late disk-hydration event on boot can overwrite the list and strip the live client out (the Safari `setTimeout(init, 400)` widens this window). The `ACTIVE_CLIENTS` connect callback now re-appends our own GUID if it's missing while we're still alive, so a ghost can't take over leadership from us on startup.

### Fixed Issues

$ https://github.com/Expensify/App/issues/94963
PROPOSAL: N/A — internal bug fix.

### Tests

**Primary validation is the automated tests.** The production bug is an intermittent lifecycle/timing race, and it can't be held open by hand: the fix self-heals on the next event, and `ActiveClientManager.init()` re-runs on focus/visibility changes (`Expensify.tsx`), which re-adds this client and masks any injected stuck state. The unit tests use deterministic control (direct state injection) to exercise both the deadlock condition and the recovery.

Run:

```
npx jest tests/unit/ActiveClientManagerTest.ts
```

- `tests/unit/ActiveClientManagerTest.ts` — leader registration; the boot-race re-add guard (a stale hydration that drops the live client is repaired).

Best-effort manual check (web) — note it may self-heal before you observe anything, which is itself the fix working:

1. `npm run web`, open the dev URL in Chrome with DevTools, sign in.
[...]

PR #95718, opened 2026-07-09
[CP Staging] Revert "[No QA] Fix leader-election deadlock in ActiveClientManager (SequentialQueue unable to flush)"

grgia:
Reverts Expensify/App#95148

$ https://github.com/Expensify/App/issues/95705

> #95148 added:  if a tab sees the list without its own GUID, it re-appends itself. Sign-in resets ACTIVE_CLIENTS, so every open tab re-appends simultaneously. Each append puts that tab last in the list, so while the writes race, multiple tabs simultaneously pass the isClientTheLeader() check.
> 
> RequestNewValidateCode is one persisted request in the shared queue. Each self-believed leader flushes it before seeing another tab's dequeue, so it's sent once per tab: 7 tabs, 7 API calls, 7 emails. BeginSignIn wasn't duplicated because it bypasses the queue. Any queued write is affected, which is why SignUpUser duplicated too.


grgia:
@pecanoro added the team from offending PR, no worries

pecanoro (approved):

grgia:
tested, confirmed fix

PR #95812, opened 2026-07-10
Hide device management & merge accounts and repurpose close account when copiloting into an agent

MelvinBot:
### Explanation of Change

When a user copilots into an **agent** account (acting as a delegate for an `agent_*@expensify.ai` account), the Security settings page previously showed options that don't apply to an agent and let the user run the normal "close account" flow against the agent.

This PR adjusts the Security page for the "copiloting into an agent" state (`isAgentAccount && isActingAsDelegate`):

- **Hides "Device management"** — it manages a real user's own logged-in devices, which is meaningless for an agent.
- **Hides "Merge accounts"** — merging doesn't apply to agent accounts.
- **Repurposes "Close account"** — instead of navigating to the usual close-account flow, it now shows a delete confirmation, invokes the `DELETE_AGENT` command for the agent you're copiloting into, and then ends the copilot session (`disconnect`), returning you to your own account.

Details:
- `src/pages/settings/Security/SecuritySettingsPage.tsx` — adds the `isCopilotingIntoAgent` guard, conditionally hides the two menu items, and swaps the close-account action for the delete-agent + disconnect flow. The confirmation modal reuses the existing `editAgentPage.deleteAgentTitle`/`deleteAgentMessage` copy (same UX as deleting an agent from the Edit agent page).
- `src/libs/actions/Agent.ts` — `deleteAgent` gains an optional `shouldNavigateBack` param (default `true`, so existing callers are unaffected). The copilot flow passes `false` because the delegate transition resets navigation on its own.

> **Note for reviewer:** The "copiloting into an agent" condition is gated on both `isAgentAccount` and `isActingAsDelegate` so the delete-agent behavior only applies when a copilot session actually exists (ending the session requires the stashed delegate credentials). Let me know if you'd rather key it on `isAgentAccount` alone.

### Fixed Issues
$ https://github.com/Expensify/App/issues/95726

### Tests
1. Sign in as a user that has at least one agent set up (Settings > Agents).
2. Go to Settings > Agents, open an agent, and tap **Copilot into account** to start copiloting into the agent.
3. Navigate to Settings > Security.
4. Verify **Device management** is **not** shown.
5. Verify **Merge accounts** is **not** shown.
6. Verify **Close account** is still shown.
7. Tap **Close account** and verify a **Delete agent?** confirmation modal appears (instead of the usual close-account page).
8. Tap **Delete** and verify:
    - the agent is deleted (it no longer appears under Settings > Agents), and
    - the copilot session ends and you are returned to your own account.
9. Repeat step 7 but tap **Cancel** and verify nothing is deleted and the copilot session stays active.
10. **Regression:** Sign in as a normal (non-copiloting) user, go to Settings > Security, and verify **Device management**, **Merge accounts**, and **Close account** all appear and behave as before (Close account opens the standard close-account flow).
11. Verify that no errors appear in the JS console.

### Offline tests
Same as tests, except for the copilot "Close account" flow: ending a copilot session requires a live network call (`DisconnectAsDelegate` is a side-effect request that can't be queued offline), so when copiloting into an agent and offline, tapping **Close account** shows a "You appear to be offline" modal and does **not** delete the agent or end the copilot session. Verify the modal appears and nothing else happens; the delete flow itself can only be tested online.

### QA Steps
Same as tests.

### PR Author Checklist
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/9c4e03b5-db65-4b5b-b3fd-437da82486db

</details>


ikevin127 (changes requested):
### 🐞 Two bugs in the "copiloting into an agent → Close account → Delete" flow

Tested this flow and hit two separate issues. Both are fixed by the DIFF below.

---

#### Bug 1 — Delete fails with a `666` permission error; the agent is never deleted

The copilot session ends (you return to your own account) but the agent still appears under Settings > Agents and can still be copiloted into. The API responds with:

```json
{
    "jsonCode": 666,
    "title": "Cannot complete action",
    "message": "You do not have the permission to do the requested action."
}
```

**Root cause:** `DELETE_AGENT` must be issued by the agent's **owner**, but the current ordering fires it while the session is still authenticated **as the agent**. Requests get their auth token at send time via `enhanceParameters()` → `getAuthToken()` (mirrors `ONYXKEYS.SESSION`), and `enhanceParameters` also appends the `delegate` param while a copilot session is active. So the delete flushes under the agent's identity → backend sees the agent trying to delete itself → `666`. This is the same command `EditAgentPage` runs successfully; the only difference is the identity it's issued under.

The current commit (`Delete agent within copilot session, then disconnect after the delete flushes`) regressed this — the earlier `Delete agent as owner after ending copilot session, not as the agent` ordering was correct. The `Delegate.ts` comment ("the promise is returned so callers can run follow-up work … once the original session is restored") already describes the intended disconnect-first design; the code just stopped following it.

<details>
<summary>Bug Video</summary>

https://github.com/user-attachments/assets/2bf227cc-9545-4661-9934-e08fc2f229d2

</details>

**Fix:** end the copilot session first (restoring the owner's session + auth token), then delete the agent as the owner. Capture the agent's `accountID` before disconnecting, and `await disconnect()` so the request goes out under the owner's token. Also **remove the now-unused `waitForIdle` import** from `@libs/Network/SequentialQueue`. The `Delegate.ts` change (returning the `disconnect` promise) is now correct as-used since we `await` it.

---

#### Bug 2 — Confirming the delete while offline hangs silently

Repro: copilot into an agent → turn off the network → tap Close account → Delete. Nothing happens in the UI; the network tab shows a **pending `DisconnectAsDelegate`** request that never resolves. Going back online (`ReconnectApp`, `AuthenticatePusher`, `SearchForTodos` all fire) does **not** un-stick it — `DisconnectAsDelegate` stays pending and the agent is never deleted.

**Root cause:** `disconnect()` uses `API.makeRequestWithSideEffects` (a side-effect request), which is **not persisted to the SequentialQueue**, so offline it's never queued/retried and `await disconnect()` never resolves — the whole flow stalls before `deleteAgent()` is ever reached.

**Fix:** this flow fundamentally requires a network round-trip, so guard on `isOffline` and show the standard offline modal (same pattern as `DomainMembersPage`) instead of starting the flow.

---

#### Combined diff (`src/pages/settings/Security/SecuritySettingsPage.tsx`)

```tsx
// imports
import useNetwork from '@hooks/useNetwork';
// remove: import {waitForIdle} from '@libs/Network/SequentialQueue';

// inside the component
const {isOffline} = useNetwork();

// close-account action, copilot branch:
if (isCopilotingIntoAgent) {
    // Deleting the agent ends the copilot session via DisconnectAsDelegate, which is a network
    // side-effect request that can't be queued offline (it never resolves, leaving the flow stuck).
    // So require a connection before starting the flow.
    if (isOffline) {
        showConfirmModal({
            title: translate('common.youAppearToBeOffline'),
            prompt: translate('common.thisFeatureRequiresInternet'),
            confirmText: translate('common.buttonConfirm'),
            shouldShowCancelButton: false,
        });
        return;
    }
    const result = await showConfirmModal({
        title: translate('editAgentPage.deleteAgentTitle'),
        prompt: translate('editAgentPage.deleteAgentMessage'),
        confirmText: translate('common.delete'),
        cancelText: translate('common.cancel'),
        danger: true,
    });
    if (result.action !== ModalActions.CONFIRM) {
        return;
    }
    // The DeleteAgent command must be issued by the agent's owner, but while copiloting the session is
    // authenticated as the agent itself (deleting as the agent fails with a 666 permission error). So end
    // the copilot session first (which restores the owner's session and auth token) and only then delete
    // the agent, now that we're acting as the owner.
    const agentAccountID = session?.accountID ?? CONST.DEFAULT_NUMBER_ID;
    await disconnect({stashedCredentials, stashedSession});
    deleteAgent(agentAccountID, undefined, undefined, false);
    return;
}
```

Remember to add `isOffline` to the `useMemo` dependency array for the menu items.

I verified both fixes locally (typecheck passes): disconnect-first deletes the agent successfully as the owner, and the offline guard surfaces a clear "You appear to be offline" modal instead of hanging.

---

#### Suggested tests to add

No tests ship with this PR, and the two bugs above are exactly the kind of thing a review can't catch by eye — the failure lives in *ordering* and *offline transport*, not in the visible UI. Some coverage worth adding:

1. **Ordering / identity (guards Bug 1):** unit test the copilot close-account branch asserting `disconnect()` is called **before** `deleteAgent()`, and that `deleteAgent` receives the agent's `accountID` captured pre-disconnect. *Why:* this is the whole fix — a future refactor could silently flip the order back (as happened here) and reintroduce the `666`. A test pins the invariant that the delete runs as the owner.
2. **Offline guard (guards Bug 2):** render the page with `isOffline = true`, trigger the action, and assert the "You appear to be offline" modal shows and neither `disconnect()` nor `deleteAgent()` is called. *Why:* prevents anyone from later "simplifying" the guard away and re-stranding the flow on a `DisconnectAsDelegate` that can't resolve.
3. **Menu-item gating (guards the core PR change):** with `isCopilotingIntoAgent = true`, assert Device management and Merge accounts are **absent** and Close account is **present**; with it `false`, assert all three render and Close account navigates to `SETTINGS_CLOSE`. *Why:* this is the primary behavior of the PR and the most likely thing to regress when the security menu is edited later.
4. **Cancel path:** confirm modal → Cancel resolves to a no-op (no `disconnect`/`deleteAgent`). *Why:* a destructive action must be safely abortable; cheap to assert and easy to break.

---

#### Note on the offline modal + PR description

The blocking modal is needed **specifically because of `DisconnectAsDelegate`**: unlike `DELETE_AGENT` (a normal `API.write()` that the SequentialQueue persists and flushes on reconnect), `disconnect()` is a `makeRequestWithSideEffects` call that bypasses the queue — it's fire-once, never persisted, never retried. So the usual "queue it and let it run when back online" behavior doesn't apply, and the flow genuinely can't complete offline. That's why we block up front instead of optimistically queuing.

Because of this, the **offline behavior here differs from online** and the PR's **Offline tests** section shouldn't just say "Same as tests." It should call out that, when copiloting into an agent, tapping Close account **while offline** surfaces the offline modal and does **not** delete the agent or end the copilot session (that requires a connection) — so QA verifies the guard rather than the delete.

cc @NicolasBonet MelvinBot will need your nudge to apply the fixes I mentioned in this comment, due to permission issues.

ikevin127 (approved):
🟢 LGTM - All issues were addressed and tests well

JmillsExpensify (dismissed):
LGTM

puneetlath (approved):

PR #95873, opened 2026-07-10
Fix invoice Pay from Search transaction rows (resolve chatReport in PayActionCell)

MelvinBot:
### Explanation of Change

When searching `type:invoice` with a custom `columns:...` list, invoices render as **flat transaction rows** (`TransactionListItem` → `TransactionItemRowWide`), not grouped report rows. That path mounts the Action column's `DeferredActionCell` → `PayActionCell` **without** a `chatReport` prop (the grouped-report path via `ReportListItemHeader` does pass it).

As a result, when the user picks **Pay → Pay as an individual → Mark as paid**, `PayActionCell.confirmPayment` hits its guard `if (!type || !reportID || !hash || !amount || !chatReport)` and returns immediately because `chatReport` is `undefined`. `payInvoice()` is never called — the reported "nothing happens" symptom. (`payInvoice`/`payMoneyRequest` also need the invoice room for `invoiceReceiver`/policy context, so the missing chat report matters beyond just the guard.)

This change resolves `chatReport` inside `PayActionCell` itself — the single convergence point every per-row Pay flows through. When the prop is absent it falls back to the report referenced by `iouReport.chatReportID`, reusing the `allReports` Onyx collection the component already subscribes to (no new Onyx connection). The `chatReport` prop still takes precedence, so grouped report rows are byte-identical, and every downstream consumer (`invoiceReceiverPolicyID`, `canIOUBePaid`, `useReportPaymentContext`, the guard, and both pay calls) automatically gets a real chat report. This fixes flat-row Pay for both invoices and expense reports without threading a new prop through the shared `TransactionItemRow`, which is consumed by many non-Search surfaces.

Added `tests/ui/components/PayActionCellTest.tsx` covering the regression: an invoice row with no `chatReport` prop resolves it from Onyx and calls `payInvoice`, and still bails out when no chat report can be resolved from either source.

### Fixed Issues

$ https://github.com/Expensify/App/issues/95422
PROPOSAL: https://github.com/Expensify/App/issues/95422#issuecomment-4896354437

### Tests

// TODO: **The human co-author** must fill out the tests you ran before marking this PR as "ready for review".

Suggested manual test:
1. As a user who has received an invoice, go to the Search page.
2. Search `type:invoice columns:receipt,date,status,merchant,from,category,tag,amount,action`.
3. Click the row's **Pay** dropdown → **Pay as an individual** → **Mark as paid**.
4. Verify the invoice is marked as paid (optimistic paid/loading state updates) instead of nothing happening.
5. Verify grouped invoice/expense-report Search rows and non-invoice Pay actions still work unchanged.

### Offline tests

Same as tests.

### QA Steps

// TODO: **The human co-author** must fill out the QA tests you ran before marking this PR as "ready for review".

### PR Author Checklist

- [x] I included screenshots or videos for tests on all platforms
- [x] I ran the tests on all platforms & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
- [x] I verified there are no console errors related to changes in this PR
- [x] I tested other components that can be affected by my changes (i.e. changing the props of a reusable component to fix a bug should be tested against the usages of the reusable component)
- [x] I verified all code is DRY (the DRY principle is explained [here](https://web.archive.org/web/20220407053143/https://wiki.c2.com/?DontRepeatYourself))
- [x] If the PR modifies a generated file (`src/libs/SearchParser/searchParser.js`, `src/libs/Pusher/pusher.js`, etc.), I verified that the changes are made to the source file and the generated file is regenerated
- [x] If the PR modifies a file used for translations (e.g. `en.ts`, `es.ts`), I verified that any changes made to translations are also reflected in the other translation files
- [x] If the PR modifies the search input, I verified that the search input is not debounced and that the search results are updated in real-time
- [x] I verified that all the inputs inside a form are aligned with the design system, i.e. `TextInput` for text inputs, `AmountForm` for amount inputs, etc.


src/components/Search/SearchList/ListItem/TransactionListItem/index.tsx
```diff
@@ -138,7 +138,7 @@ function TransactionListItemInner<TItem extends ListItem>({
         transactionItem,
     ]);
     const currentUserDetails = useCurrentUserPersonalDetails();
-    const [parentChatReport] = originalUseOnyx(`${ONYXKEYS.COLLECTION.REPORT}${getNonEmptyStringOnyxID(snapshotReport?.chatReportID)}`);
+    const [parentChatReport] = originalUseOnyx(`${ONYXKEYS.COLLECTION.REPORT}${getNonEmptyStringOnyxID(snapshotReport?.chatReportID ?? snapshotReport?.parentReportID)}`);
```
Pujan92: @MelvinBot do we need this change?

MelvinBot: Not strictly for the invoice bug — it's a parity/robustness change. Invoice reports carry a `chatReportID` (the invoice room), so for the flow this PR fixes the `?? parentReportID` branch effectively never fires. I added it so the flat-row resolution matches the two places that already resolve the chat report this way:

- the grouped-report path in [`ReportListItemHeader.tsx:276`](https://github.com/Expensify/App/blob/b7311166408043646e90292cdc5b9a5814b4a992/src/components/Search/SearchList/ListItem/ReportListItemHeader.tsx#L276) — `snapshotReport?.chatReportID ?? reportItem.parentReportID`
- the adjacent `chatReportActions` line right below it in this same file, which already keys on `chatReportID ?? parentReportID`

So it keeps `parentChatReport` consistent with its neighbor and the grouped path, and covers report types where `chatReportID` is empty but `parentReportID` is set. If you'd rather keep the diff minimal, I'm happy to revert line 141 to key on `chatReportID` only — the invoice fix works either way. Let me know which you prefer.

src/components/TransactionItemRow/types.ts
```diff
@@ -56,6 +56,8 @@ type TransactionWithOptionalSearchFields = TransactionWithOptionalHighlight & {
 type TransactionItemRowProps = {
     transactionItem: TransactionWithOptionalSearchFields;
     report?: Report;
+    /** Chat report the transaction's report belongs to. Passed to the Pay action so flat transaction rows resolve the invoice room the same way grouped report rows do. */
```
Pujan92: @MelvinBot I don't think we need comment here or [here](https://github.com/Expensify/App/pull/95873/changes#diff-42a1cabf0b54d74cef23b74a0b7d8466653f7b0b3b2e35acd2d3182f1cc92aeaR36) too

MelvinBot: Agreed — the sibling props (`report`, `policy`, etc.) don't carry doc comments, so `chatReport` shouldn't be the odd one out. Removed both:

- `src/components/TransactionItemRow/types.ts`
- `src/components/Search/SearchList/ListItem/TransactionListItem/types.ts`

Pushed in 6b9d7d5.

Pujan92 (approved):
LGTM!

https://github.com/user-attachments/assets/86191e41-24c4-4c41-b6a7-299b8dd7a5e3

joekaufmanexpensify (dismissed):
Good for product.

src/components/Search/SearchList/ListItem/TransactionListItem/index.tsx
```diff
@@ -260,6 +260,7 @@ function TransactionListItemInner<TItem extends ListItem>({
         policyTagLists,
         nonPersonalAndWorkspaceCards,
         isAttendeesEnabledForMovingPolicy,
+        chatReport: parentChatReport,
```
lakchote: Can we preserve the Search-snapshot fallback here? 

Like we do already for Search:

https://github.com/Expensify/App/blob/ff207e497c8181cf8b03995cc65529b793066547/src/components/Search/SearchList/ListItem/ReportListItemHeader.tsx#L278

Pujan92: Yes, @MelvinBot apply the above suggestion. I think also need a fallback [here](https://github.com/Expensify/App/blob/4bafe132d8f2d923b4f32dc980f40d365200c05a/src/components/Search/SearchList/ListItem/TransactionListItem/index.tsx#L146) too.

MelvinBot: Done — pushed in 77080c7.

I resolved the chat report to a single `chatReport` that prefers live Onyx and **falls back to the Search snapshot**, mirroring the grouped-report path in [`ReportListItemHeader.tsx:277-278`](https://github.com/Expensify/App/blob/77080c7f21956d180f50be6af17f5b99c10273cd/src/components/Search/SearchList/ListItem/ReportListItemHeader.tsx#L277-L278):

- [`TransactionListItem/index.tsx:142-149`](https://github.com/Expensify/App/blob/77080c7f21956d180f50be6af17f5b99c10273cd/src/components/Search/SearchList/ListItem/TransactionListItem/index.tsx#L142-L149) — `snapshotChatReport` reads from the search snapshot and `const chatReport = parentChatReport ?? snapshotChatReport;`.
- The snapshot fallback now flows through all consumers, including the second spot you flagged — [`chatReportPolicyID`](https://github.com/Expensify/App/blob/77080c7f21956d180f50be6af17f5b99c10273cd/src/components/Search/SearchList/ListItem/TransactionListItem/index.tsx#L149), the `handleActionButtonPress` Pay path, and the `chatReport` shared prop.

`chatReportActions` now keys on `chatReport?.reportID ?? chatReportID` too, matching the header path.

Ran `fmt`, `lint-changed`, `typecheck-tsgo`, and the React Compiler compliance check — all pass.

lakchote (approved):
LGTM

Issue #95750, opened 2026-07-09
[CFI] [$500] Allow any workspace admin to pay reports


joekaufmanexpensify:
Coming from [this Slack thread](https://expensify.slack.com/archives/C07NMDKEFMH/p1783471123993169), @sakluger flagged a classic/NewDot parity gap regarding payments. 
 
### Details
- When no bank account is connected to a workspace, any workspace already sees the option to pay/mark as paid in NewDot. This part works correctly today and needs no changes.
- The gap: when a bank account **is** connected to a workspace, NewDot only shows the pay button to the admin set as the workspace payer. Classic shows it to any workspace admin. 
- Proposal: any workspace admin should see the option to pay (via any VBA they have access to) or manually mark as paid, regardless of whether a bank account is connected to the workspace. 
    - In the event the workspace doesn't have a VBA, they should be prompted to add one in situ when they try to pay via business account (the existing payment button logic). 
- An admin adding/paying with a bank account that's not the current workspace bank account shouldn't change the workspace bank account, same as now.
- We should **not** actively prompt admins that aren't the workspace payer to pay via badges/GBRs. That stays as-is (per https://github.com/Expensify/Expensify/issues/637864).
- One note: an admin that isn't the workspace payer who tries to pay a report via ACH in Classic sees an error, despite being able to mark as paid. I don't think we should go this direction in NewDot, since we already allow the admin set as the workspace payer to select the bank account to use at the time of payment in both NewDot and Classic. It feels more consistent to also allow the non-payer admin to select the bank account they want to use or mark as paid. Curious if anyone disagrees?
- I don't think this issue should require any backend changes. Let me know if anyone feels differently. 
- This issue is only about who can see and use the pay button. It shouldn't change which options appear or how it behaves for a given report (e.g., non-reimbursable-only reports showing "Mark as Paid" only in the report detail view, not the report preview or the Spend page).

### Non-payer workspace admin view when workspace has bank account - NewDot 

<img width="1507" height="784" alt="Image" src="https://github.com/user-attachments/assets/d5b46757-9133-434d-b812-453286308364" />

### Non-payer workspace admin view when workspace has bank account - OldDot 

<img width="1508" height="782" alt="Image" src="https://github.com/user-attachments/assets/d5e5ae28-46b1-4e27-9069-ebbc9c3b44f6" />

### Non-payer workspace admin view when workspace has no bank account - NewDot 

[...]

PR #95925, opened 2026-07-13
Allow any workspace admin to pay reports

lorretheboy:
### Fixed Issues

$ https://github.com/Expensify/App/issues/95750
PROPOSAL: https://github.com/Expensify/App/issues/95750#issuecomment-4928987073

### Tests

Precondition

1. Account A: workspace admin, owner of a Control/Collect workspace with a verified business bank account connected (reimbursementChoice = reimburseYes).
2. In Workspace > Workflows > Make or track payments, set Account A as the payer (reimburser).
3. Account B: invite as a second admin on the same workspace (B is not the payer and has no access to the workspace bank account).
4. Account C: a regular member (employee) on the workspace.
5. As Account C, submit an expense report with a reimbursable expense, and get it approved (or use Submit & Close so it's finished).

Test 1 — Report preview (chat)

1. Sign in as Account B (non-payer admin).
2. Open the workspace chat with Account C and find the approved report's preview.
3. Verify the preview shows the Pay button (previously it showed nothing for B when a bank account was connected).

Test 2 — Report detail view (primary action)

1. Still as Account B, open the expense report itself.
2. Verify the green Pay primary button appears in the header.
3. Click it and verify the payment options open (pay with a business bank account B has access to / add one via the KYC wall, or Pay elsewhere / Mark as paid).
4. Verify paying with a different bank account does not change the workspace's connected bank account.

Test 3 — Spend / Reports page

1. As Account B, go to Reports > Expense reports (Search) and locate the approved report.
2. Verify the row/bulk Pay option is available for that report.

### QA Steps

Precondition

1. Account A: workspace admin, owner of a Control/Collect workspace with a verified business bank account connected (reimbursementChoice = reimburseYes).
2. In Workspace > Workflows > Make or track payments, set Account A as the payer (reimburser).
3. Account B: invite as a second admin on the same workspace (B is not the payer and has no access to the workspace bank account).
4. Account C: a regular member (employee) on the workspace.
5. As Account C, submit an expense report with a reimbursable expense, and get it approved (or use Submit & Close so it's finished).

Test 1 — Report preview (chat)

1. Sign in as Account B (non-payer admin).
2. Open the workspace chat with Account C and find the approved report's preview.
3. Verify the preview shows the Pay button (previously it showed nothing for B when a bank account was connected).

Test 2 — Report detail view (primary action)

1. Still as Account B, open the expense report itself.
2. Verify the green Pay primary button appears in the header.
3. Click it and verify the payment options open (pay with a business bank account B has access to / add one via the KYC wall, or Pay elsewhere / Mark as paid).
4. Verify paying with a different bank account does not change the workspace's connected bank account.

Test 3 — Spend / Reports page

1. As Account B, go to Reports > Expense reports (Search) and locate the approved report.
2. Verify the row/bulk Pay option is available for that report.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/bf30a8ff-429c-4c5d-b856-7d4956140b5b

</details>

<details>

https://github.com/user-attachments/assets/c27f5d51-1ff3-45a1-96ab-b5cb57bbdebf

</details>

<details>

https://github.com/user-attachments/assets/aa344281-9d30-4261-9456-6da371749896

</details>

<details>

https://github.com/user-attachments/assets/a32a915d-4cae-447d-8875-d3c90ed0b2fd

</details>

<details>

https://github.com/user-attachments/assets/e010e627-2dfe-4bf7-824f-ccbd9d9bda87

</details>


src/libs/PolicyUtils.ts
```diff
@@ -1483,7 +1492,7 @@ function isSubmitAndClose(policy: OnyxInputOrEntry<Policy>): boolean {
     return policy?.approvalMode === CONST.POLICY.APPROVAL_MODE.OPTIONAL;
 }
 
-function arePaymentsEnabled(policy: OnyxEntry<Policy>): boolean {
+function arePaymentsEnabled(policy: OnyxInputOrEntry<Policy>): boolean {
```
ikevin127, line 1603: **🟡 Behavior broadens to `undefined` `reimbursementChoice` — confirm intent (Low/Medium)**

`arePaymentsEnabled` returns `policy?.reimbursementChoice !== REIMBURSEMENT_NO`. That is `true` not only for `MANUAL` and `YES`, but also when `reimbursementChoice` is `undefined`:

The old logic was strictly `=== REIMBURSEMENT_MANUAL`, so a policy with an unset `reimbursementChoice` previously did **not** grant a non-payer admin the **Pay** action; now it does ⚠️ 

**Why it matters**: if a policy can legitimately exist with `reimbursementChoice` `undefined` at the point these utils run (e.g. mid-onboarding, or before the payments feature is fully configured), a non-payer admin would now see a **Pay** button that wasn't there before. The PR test only covers the `REIMBURSEMENT_YES` case, so this branch is unverified 🧪 

This is likely fine (it's consistent with how `arePaymentsEnabled` is used elsewhere in the very same functions via `isPaymentsEnabled`), but since it's a semantic widening beyond what the issue strictly asks for (Manual + VBA), please confirm the `undefined` case is intended and can't surface a **Pay** button on an unconfigured policy.

tests/unit/ReportPrimaryActionUtilsTest.ts
```diff
@@ -640,6 +640,53 @@ describe('getPrimaryAction', () => {
         ).toBe(CONST.REPORT.PRIMARY_ACTIONS.PAY);
     });
```
ikevin127, line 810: 🟠 **Test coverage is partial — 1 of the 4 touched pay surfaces (Medium)**

The diff changes pay-availability in four places but adds a test for only one:

| Surface | File | Test added? |
|---|---|---|
| Report detail primary action | `ReportPrimaryActionUtils.isPrimaryPayAction` | ✅ new test |
| Report preview (chat) | `ReportPreviewActionUtils.canPay` | ❌ |
| Spend/Search bulk pay | `ReportWorkflow.canIOUBePaid` | ❌ |
| Secondary action (held/re-pay) | `ReportSecondaryActionUtils` (`canNonPayerAdminPay: true`) | ❌ |

The three PR manual tests map exactly to preview / detail / spend surfaces, but only the detail path has an automated guard. Given the whole point of `canAdminPayReport` is to keep these four call sites in lockstep, I'd add at least one canPay (preview) test and one `canIOUBePaid` test asserting a non-payer admin gets **Pay** under `REIMBURSEMENT_YES`. Otherwise a future edit to `canPay/canIOUBePaid` could silently regress the exact behavior this PR ships, and the suite wouldn't catch it.

**Rough coverage estimate provided currently**: `~25%` of the modified pay-availability surfaces have direct automated coverage (`canAdminPayReport` itself is exercised transitively, but only through the primary-action path).

src/libs/ReportPrimaryActionUtils.ts
```diff
@@ -225,11 +225,7 @@ function isPrimaryPayAction({
         return false;
     }
     const isReportPayer = isPayer(currentUserAccountID, currentUserLogin, report, bankAccountList, policy, false);
-    const canPayReport =
-        isReportPayer ||
-        (canNonPayerAdminPay &&
-            policy?.reimbursementChoice === CONST.POLICY.REIMBURSEMENT_CHOICES.REIMBURSEMENT_MANUAL &&
-            canMemberWrite(policy, currentUserLogin, CONST.POLICY.POLICY_FEATURE.WORKFLOWS_PAYMENTS));
+    const canPayReport = isReportPayer || (!!canNonPayerAdminPay && canAdminPayReport(policy, currentUserLogin));
```
ikevin127, line 225: 🟢 **Asymmetric `canNonPayerAdminPay` gating — worth a one-line note (Low, likely intentional)**

The three consumers gate differently now:

- `ReportPrimaryActionUtils`: `isReportPayer || (!!canNonPayerAdminPay && canAdminPayReport(...))` — gated by the flag.
- `ReportPreviewActionUtils.canPay` and `ReportWorkflow.canIOUBePaid`: `isReportPayer || canAdminPayReport(...)` — **not** gated.

This mirrors the pre-existing structure (only `isPrimaryPayAction` ever had the `canNonPayerAdminPay` parameter), so it's not a regression. But note the pre-existing consequence, now reachable by more users: in `getReportPrimaryAction`, the *first* `isPrimaryPayAction` call (line ~500, feeding `isPayActionWithAllExpensesHeld` → `REMOVE_HOLD`) does **not** pass `canNonPayerAdminPay`, while the PAY call (line ~539) does.

So a non-payer admin viewing a report where *all expenses are held* gets neither PAY (guarded by `!allExpensesHeld`) nor REMOVE_HOLD via the pay path. Before this PR that combination was mostly unreachable for non-payer admins (`REIMBURSEMENT_MANUAL`-only, narrow); after it, more non-payer admins hit it. Not blocking, but it's an edge case the PR's manual tests don't touch — worth a quick manual check.

tests/unit/ReportPrimaryActionUtilsTest.ts
lorretheboy: Added more tests

src/libs/PolicyUtils.ts
lorretheboy: I updated to match how isPolicyPayer already gates

ikevin127 (approved):
🟢 LGTM - Tests well

##

Re-reviewed the latest diff, all my prior comments are addressed and, with the payment-admin scope update, the role gate is correct as-is 👍

**Addressed:**
- 🟢 `canAdminPayReport` now explicitly checks `reimbursementChoice === REIMBURSEMENT_YES || === REIMBURSEMENT_MANUAL` (no longer `arePaymentsEnabled`), so an unset `reimbursementChoice` no longer surfaces Pay — and there's now a negative test for it in both `ReportWorkflowTest` and `ReportPreviewActionUtilsTest`.
- 🟢 Test coverage now spans all three surfaces: `canIOUBePaid`, `canPay` (preview), and `isPrimaryPayAction`, each with the positive `reimburseYes` case and the negative unconfigured case, asserting `isPayer(...) === false` first.
- 🟢 The `isPayActionWithAllExpensesHeld` call in `getReportPrimaryAction` now passes `canNonPayerAdminPay: true`, so the REMOVE_HOLD / all-held path is consistent with the PAY path.

**On the role gate (`canMemberWrite(policy, login, WORKFLOWS_PAYMENTS)`):** keep as-is. Per the updated scope, payment admins *should* be able to pay reports they already have access to, and `canMemberWrite(..., WORKFLOWS_PAYMENTS)` correctly covers `ADMIN` + `PAYMENTS_ADMIN` (Control policies). These utils only gate whether the Pay button *shows* on a report the user already sees — they don't expand report access — so this matches the scope exactly. Restricting to `isPolicyAdmin` would regress the payment-admin case.

**🟢 Minor/optional**: `EDITOR` also has `WORKFLOWS_PAYMENTS: WRITE` via `EDITOR_POLICY_FEATURES`, so it'd technically get the button too. `EDITOR` has no assignment action (no `MAKE_EDITOR`) so it's effectively unreachable in the UI, and this role selection is unchanged from before — noting for completeness only.

ikevin127 (changes requested):
## Review summary

| Bug | Status | Why |
|---|---|---|
| **Bug 1**: workspace bank account last-4 shown in pay button to admin who can't access it | 🟠 **Fixed for the non-payer admin, but introduced a new regression** | The `canAccessPolicyBankAccount` guard correctly hides the workspace account from non-payer admins, but it is also applied to `hasIntentToPay` and the display branch, so it now hides the last-4 from the **payer/owner too** (see Bug 1b) |
| **Bug 1b** (new): workspace bank account last-4 no longer shows for the payer/owner, where it should | 🔴 **Regression** | `canAccessPolicyBankAccount = !!policyBankAccountID && !!bankAccountList?.[policyBankAccountID]` is the wrong signal for "the payer can use the workspace account". It is false for the payer whenever `policy.achAccount.bankAccountID` is absent (the code guards this as common, e.g. `SettlementButton/index.tsx:238`) or the workspace VBBA is not in the viewer's `BANK_ACCOUNT_LIST` |
| **Bug 2**: report comment shows a different bank account to different viewers | 🔴 **Still present** | The fix reads `originalMessage.accountNumber`, but that field is **never written** onto the PAY action: `buildOptimisticIOUReportAction` (`ReportUtils.ts:7415`) sets only `bankAccountID`, and nothing in `PayMoneyRequest.ts`/`src/libs/actions/` sets `accountNumber`. So `if (accountNumber)` is always false and rendering falls back to viewer-dependent lookup (`bankAccountList[bankAccountID]` / `reportPolicy.achAccount`) |
| **Bug 3**: main pay button prompts to link bank to workspace; dropdown just pays | 🔴 **Still present** | Split-button main-press (`ButtonWithDropdownMenu.handlePress` → `onPress(event, selectedItem.value)`, no `methodID`) routes through `selectPaymentMethod` → `triggerKYCFlow`, while the dropdown item pays directly via its own `onSelected` → `onPress({methodID})`. The PR didn't unify the two paths |

##

Re-reviewed the latest build against the reported bugs. Bug 1 is fixed for the non-payer admin, but the same change introduced a new regression for the payer/owner (Bug 1b), and Bug 2 and Bug 3 are both still present. Details below.

### 🟠➡️🔴 **Bug 1b (new regression): the workspace bank account last-4 no longer shows for the payer/owner, where it should.**
The fix gates both `hasIntentToPay` and the secondary-text branch on `canAccessPolicyBankAccount = !!policyBankAccountID && !!bankAccountList?.[policyBankAccountID]`. That is the wrong signaworkspace account", so it also suppresses the digits for thelegitimate payer. `canAccessPolicyBankAccount` is false for the payer/owner whenever either of these holds, and both are common at runtime:
1. `policy.achAccount.bankAccountID` is absent. The type declares the code treats it as frequently missing, including here at`SettlementButton/index.tsx:238` (`if (policy?.achAccount?.bankAccountID === undefined) return;`) and in `SearchUIUtils`, `useSearchBulkActions`,
`duplicate/utils`, etc. If it is undefined, `policyBankAccountID`
2. The workspace VBBA is not enumerated in the viewer's `BANK_ACCOUNT_LIST` (a separate, independently-synced Onyx key from `policy.achAccount`).

Result for the payer:
- First-time pay (no `lastPaymentMethod`): `hasIntentToPay` goes fToPay && isExpenseReport` guard skips the bank-account textentirely. No digits.
- Repeat pay (`lastPaymentMethod === VBBA`): the `if (canAccessPoler)` branch is skipped and the `else if (bankAccountToDisplay...)`has nothing to show. No digits.

Before the fix the payer saw the last-4 straight from `policy.achAccount.accountNumber` (reliably present), independent of `bankAccountID`/`bankAccountList`.

**➡️ Fix**: only suppress the workspace account digits for a non-payerot for the payer. Gate on `(isReportPayer ||canAccessPolicyBankAccount)` for both `hasIntentToPay` and the display branch, so the payer keeps seeing `policy.achAccount.accountNumber` while a non-payer admin still does not see an account they can't use.

### 🔴 **Bug 2 (conflicting bank account in report comment): the fix is inert.**
`getBankAccountLastFourDigits` now prefers `originalMessage.accountNumber`, but that field is never populated on the PAY action. `buildOptimisticIO(`ReportUtils.ts:7415`) writes only `bankAccountID` into `originalMoneyRequest.ts` (or anywhere in `src/libs/actions/`) sets`accountNumber`. Since the issue precludes backend changes, `if (accountNumber)` is always false for VBBA payments, so every render site falls backviewer-dependent resolution (`bankAccountList[bankAccountID]` → on; `reportPolicy.achAccount` → the workspace account), which isexactly why different viewers see different accounts.

**➡️ Fix:** populate `originalMessage.accountNumber` optimistically at pay time from the selected method (`bankAccountList[methodID].accountData.accountNumber`), so the actual masked account is stored on the action and synced to everyone. The read side is already wired up; only the write is missing.       

### 🔴 **Bug 3 (main button prompts to link bank; dropdown just pays): dispatch asymmetry.**

The dropdown business-account item pays directly via its own `onSepe: VBBA, methodID: account.methodID, ...})`(`SettlementButton/index.tsx:347`), bypassing KYC. The main split-button press instead goes `ButtonWithDropdownMenu.handlePress` → `onPress(event, selectedItem.value)` (value is the generic `'businessBankAccount'`mentSelection` → `selectPaymentMethod` → `triggerKYCFlow(...)`,which for a non-payer admin without a workspace bank account opens the "link bank to workspace" flow. The Bug 1 fix only changed `hasIntentToPay`/secondary text, not this path.

**➡️ Fix**: make the main-button press resolve to the same direct payment as the defaulted dropdown item (carry the selected `methodID` and skip the workspace-link KYC when paying with an account the admin already has).

> [!note]
> **Also**: the new unit tests only cover the `canAdminPayReport` gatingettlementButton` display/dispatch or the `accountNumber`preference, so please add guards once these are fixed.

**Manual tests that must pass before merge:**
1. **Payer/owner (Bug 1b):** as the workspace owner/designated payer on a workspace with a connected bank account, open a report pending payment and confirm the Pay button shows the workspace bank account's last 4 digits, both on first-time pay and after having paid before. For Workspace Admin side, the 4 digits should not show though (current behaviour is correct, except the WS Owner regression).
2. **Non-payer admin (Bug 3):** non-payer admin (User B) with their own bank account (BA2) taps the **main** Pay button and pays with BA2 directly. It must NOT prompt to link BA2 to the workspace.
3. **Report comment (Bug 2):** after that payment, the report comment shows the **same** account (the one actually paid with) to **both** User A and User B, in the chat message, report name, and preview.

joekaufmanexpensify (dismissed):
Good from my perspective. 👍

lorretheboy:
> @lorretheboy Did you create it fresh directly from the workspace workflows page as a different user than the owner?

Actually, I created it from workspace owner. However, let me try with different user today and get back to you

lorretheboy:
@chuckdries Hi, I can confirm that it works as expected - no issue in case I create it fresh directly from the workspace workflows page as a different user than the owner

https://github.com/user-attachments/assets/49565696-0f68-4af2-b802-cde5eab0521f

chuckdries:
@lorretheboy Hey, sorry to drag this along so long, but I want to make sure the customer never sees one bank account on the button, then a different one actually gets debited. Can you confirm, when you're logged in as the User A that reproduces the bug (workspace owner and payer, right?), do you see the workspace bank account in your `bankAccountList` in Onyx? 

If not, I think that's actually good news. The backend behavior is because the workspace account is not shared with that user, so it means the fix here is "only show the workspace bank account on the Pay button if it's in the user's bankAccountList".

lorretheboy:
@chuckdries Please give me sometimes to check that case, I will try to give you outcome today. Sorry if there is any delay may happen these days, I'm not really good in health so that my working time is limited

lorretheboy:
@chuckdries I can confirm that there is no `111` account in `bankAccountList` of the workspace owner/payer

image

image

chuckdries:
Ok, sweet. Let's solve the conflicts, and make sure that we only show the workspace account as an option when it is present in that user's bankAccountList. Sound good? @lorretheboy @ikevin127

lorretheboy:
In progress

lorretheboy:
@ikevin127 @chuckdries I pushed the fix according above discussion. Please review

ikevin127 (approved):
> @ikevin127 I pushed the fix according above discussion. Please review

### What changed since the "3 cases" discussion

The BE question got resolved as an FE-only fix, no backend dependency remains. The thread settled it:

- <ins>@**joekaufmanexpensify**</ins> **confirmed the product rule**: you can only pay with a bank account that is actually shared with you. Being the workspace payer (or owner) does not by itself grant access to the workspace account
- <ins>@**chuckdries**</ins> confirmed the workspace account (1111) was not in User A's `bankAccountList`, and the "shows 1111, debits 0000" was the backend falling back to an old workspace-associated withdrawal account. **The agreed fix**: only show or use the workspace account on the **Pay** button when it is present in the user's `bankAccountList`

Commit `dbefac84b` implements exactly that.

##

<details>
<summary>🟢 Review of the latest change (dbefac84b)</summary>

<br>

**Verdict**: sound, and it correctly reverses my earlier Bug 1b suggestion. Basing access on isPolicyPayer was wrong, because a payer can lack the actual share. The new gate keys off the real signal (membership in bankAccountList).

- 🟢 `getAccessiblePolicyBankAccount(policy, bankAccountList)` returns the resolved account or `undefined`, and `canAccessPolicyBankAccount` is now a thin wrapper on it. The `isPolicyPayer ||` short-circuit is gone. All callers (**SettlementButton**, `useSearchBulkActions`, `Search.getPayActionCallback`, **PaymentUtils**) are updated to the 2-arg signature. No dangling `currentUserLogin` (still used at Search.ts:664)

- 🟢 The button now reads the last-4 from the resolved account (`policyBankAccount.accountData.accountNumber`), not the stale `policy.achAccount.accountNumber`. The comments explain the "stale number vs live bankAccountID" trap well. `getBankAccountLastFourDigits` got the same treatment for the report-comment path

- 🟢 New `onPrimaryPress` on **ButtonWithDropdownMenu** closes a real gap in the Bug 3 fix: when the button face names the workspace account (which has no menu option, since the menu is built from `bankAccountList`), the main press now pays VBBA without `methodID` so the debited account matches the label, instead of running `options[0].onSelected` and paying a different account. It routes through `runPaymentAction`, so the `checkForNecessaryAction` guards (locked account, delegate, validation, grace period) are preserved, no bypass. It is opt-in (only **SettlementButton** passes it), so other **ButtonWithDropdownMenu** consumers are unaffected

- 🟢 **Test coverage grew substantially**: new `PaymentUtilsTest.ts`, **PolicyUtilsTest** cases for `getAccessiblePolicyBankAccount` / `canAccessPolicyBankAccount`, plus **ReportPrimaryActionUtilsTest**, **ReportPreviewActionUtilsTest**, **ReportWorkflowTest**

</details>

@lorretheboy **Two things to confirm before merge, not code defects:**

- 🟡 A payer whose workspace account is not shared with them will no longer see the workspace last-4 on the button and cannot one-tap or bulk pay with it (they pick their own account). Per **Joe** this is the correct product behavior, just make sure the team is aligned that this "unshared payer" edge now shows their own account rather than the workspace one

- 🟡 `onPrimaryPress` pays **VBBA** with no `methodID` and relies on the backend defaulting to the workspace account (the comment notes naming it explicitly is rejected). Since it only fires when the account is in `bankAccountList` (shared), the default should match, but it is worth one manual confirmation that the main-button workspace payment debits the account shown

chuckdries:
Sorry for the delay, I will re-review today. In the mean time please solve conflicts

src/components/SettlementButton/index.tsx
```diff
@@ -166,10 +166,15 @@ function SettlementButton({
     const hasSinglePolicy = !isExpenseReport && activeAdminPolicies.length === 1;
     const hasMultiplePolicies = !isExpenseReport && activeAdminPolicies.length > 1;
     const formattedPaymentMethods = formatPaymentMethods(bankAccountList ?? {}, fundList ?? {}, styles, translate);
+    // Any workspace admin can pay, but the workspace account may only be shown to the members it is shared with — being
+    // the designated payer is not enough. Without this, the button advertises the workspace account to someone who cannot
+    // use it, and the payment then debits a different account. The dropdown already scopes itself to `bankAccountList`,
+    // which is why it stays correct.
```
chuckdries: Sorry, this is a NIT but I don't want it to come back to bite us later: setting the workspace payer is supposed to share the workspace bank account with them. Your test account is in a bugged state that we need to defend against, but I don't want future maintainers to believe it is intentional


```suggestion
    // Only show the workspace account when it is present in the current user's bankAccountList.
    // The normal ShareBankAccountAndUpdatePolicyReimburser flow creates a payer-owned copy 
    // and updates the policy to use it, but we've observed accounts where the payer doesn't have
    // access to the workspace account for some reason.
```

chuckdries (commented):
One last comment then I think we're good to go on this!

src/libs/actions/Search.ts
chuckdries, line 684: I think I have found a very niche edge case, but it should be easy to address so I figure we might as well solve it. Earlier in the function, we use `canAccessPolicyBankAccount(policy ?? snapshotPolicy...`, but down here, we use `policy: snapshotPolicy ?? policy`, and `methodID: ...snapshotPolicy?achAccount?.bankAccountID`. My concern is: what if someone else changes the workspace bank account while you're looking at the search page? I think we may end up doing the wrong thing if `policy` and `snapshotPolicy` end up being different.

I think for consistency, we should define something like `const paymentPolicy = policy ?? snapshotPolicy` above line 647, then use `paymenyPolicy` in all 3 places

chuckdries (commented):
I noticed in testing that the reportAction preview has the wrong wording for the user who made the payment. See on the right user's LHN, "paid $9.00 with personal account 6281" rather than "paid with bank account 6281"

Image

src/libs/actions/Search.ts
lorretheboy: Hmm interesting. I'm taking a look on this @chuckdries

lorretheboy:
@chuckdries You can re-review now. Thanks

src/libs/ReportUtils.ts
```diff
@@ -6228,7 +6234,11 @@ function getReportPreviewReportActionMessage(
             report.isWaitingOnBankAccount
         ) {
             translatePhraseKey = 'iou.paidWithExpensify';
-            const isFromInvoice = !!originalMessage?.bankAccountID;
+            // A VBBA payment is always described as "paid with bank account 1234", so it has to win over the personal/
+            // business account wording below — that's the order `PaymentContent` renders the action itself in. Any
+            // payment can name an account now that a paying admin picks one, so `bankAccountID` alone no longer
+            // implies an invoice paid outside VBBA, and keying off it would label a workspace payment as personal.
```
chuckdries: Can we rephrase this comment? We prefer comments to not describe how this code used to work. Just explain the context that's important to know now

chuckdries (commented):
We're sooooo close, I have one more NIT then we have to solve conflicts then we're golden!

trjExpensify:
@lorretheboy looks like we have conflicts now too.

lorretheboy:
@chuckdries I think it is ready now. Please note that the failed test is latest main issue, it doesn't related to our PR

cc @ikevin127

chuckdries (approved):

chuckdries:
@lorretheboy failing unit tests should be fixed in main now by https://github.com/Expensify/App/pull/100205, please merge one last time

lorretheboy:
@chuckdries I merged. Please check

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100353 was identified to be related to this PR.

lanitochka17:
Deploy Blocker https://github.com/Expensify/App/issues/100358 was identified to be related to this PR

Issue #95820, opened 2026-07-10
[Due for payment 2026-08-25] [$250] Support bulk Mark as Exported for accounting integrations in New Expensify


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding. To join our Slack channel, fill out [this form](https://forms.gle/Q7hnhUJPnQCK7Fe56).
___

**Version Number:**
**Reproducible in staging?:** Needs Reproduction
**Reproducible in production?:** Needs Reproduction
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @Getrudygidashang 
**Slack conversation** (hyperlinked to channel name): [#Expensify bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1783521908274499)

## Action Performed:

1. Connect to any accounting integrations (Netsuite here)
2. Have 50+ reports ready for export
3. Navigate to Reports > select all
4. Click All matching items selected dropdown
5. Observe the options

## Expected Result:
New Expensify should provide a bulk action to manually mark selected reports or expenses as exported to an accounting integration, consistent with the existing Classic Expensify functionality.
## Actual Result:
bulk manually marking reports or expenses as exported to an accounting integration is not displayed in All matching items selected dropdown

## Workaround:
Unknown
## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

<img width="805" height="575" alt="Image" src="https://github.com/user-attachments/assets/3a880686-06ab-4826-8b25-c08a05f9230a" />

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
[...]

PR #93297, opened 2026-06-11
Fix: show Search bulk export options without opening a report first

MelvinBot:
### Explanation of Change

On the Search > Reports page, the bulk-select export options ("Export to NetSuite" / "Mark as manually exported") did not appear on a fresh load (after clearing cache) until the user opened a report first.

**Root cause:** the bulk-export visibility gate in `useSearchBulkActions` reads the report from **live Onyx** (`allReports`). `SearchBulkActionsButton` is rendered *outside* `SearchScopeProvider`, so the `useOnyx` wrapper does **not** redirect this read to the Search snapshot. On a fresh load the live `report_<id>` is missing entirely, so `canReportBeExported` bails early at `if (!completeReport) return false` and neither export option is offered. Opening a report fires `OpenReport`, which populates the live `report_<id>` — that is exactly the "open it first" workaround.

**Fix:** resolve the report through the shared [`getReportFromSearchSnapshot`](https://github.com/Expensify/App/blob/24f30c6c618832a6d08029d639abf067ca83b0f3/src/libs/actions/Search.ts#L109) helper, which prefers the Search snapshot (`currentSearchResults.data`) and falls back to live Onyx. The snapshot already carries the full report on a fresh load — the same data the per-row Search export action relies on — so the bulk menu now shows the options without opening a report.

Two related fallbacks were added so the existing "export again" flow keeps working with snapshot-sourced data:
- The [export-action loop](https://github.com/Expensify/App/blob/24f30c6c618832a6d08029d639abf067ca83b0f3/src/hooks/useSearchBulkActions.ts#L1317-L1318) falls back to the snapshot report when the live `report_<id>` is missing.
- The [report name](https://github.com/Expensify/App/blob/24f30c6c618832a6d08029d639abf067ca83b0f3/src/hooks/useSearchBulkActions.ts#L1326-L1328) used in the export-again modal falls back to the snapshot when the live record is an incomplete optimistic entry that lacks `reportName`.

The policy reads remain on live Onyx (no snapshot lookup), matching the rest of the export-options logic.

Additionally, the "export again" confirmation modal now supports a scrollable prompt (new `shouldEnablePromptScroll` prop on `ConfirmModal`) so that a large list of already-exported reports scrolls within the modal instead of extending beyond the screen, and the "Are you sure you want to export them again?" question now precedes the report list.

### Fixed Issues

$ https://github.com/Expensify/App/issues/92976
PROPOSAL: https://github.com/Expensify/App/issues/92976#issuecomment-4654097274

[...]

PR #95988, opened 2026-07-13
Fix missing bulk export-to-integration options on fresh load (Search Reports)

MelvinBot:
### Explanation of Change

On the Search **Reports** tab, the bulk-action **Export** dropdown was missing the **Export to `<integration>`** and **Mark as exported** options on a fresh load / after a cache clear.

`SearchBulkActionsButton` renders **outside** `SearchScopeProvider`, so `useOnyx` reads are **not** auto-redirected to the Search snapshot. On a fresh load the selected policy may only exist in the Search snapshot and not yet in live Onyx. PR #93297 already fixed the equivalent problem for the **report** lookup by falling back to `getReportFromSearchSnapshot(...)`, but deliberately left the two **policy** reads in `getExportOptions()` on live Onyx. When the policy is absent from live Onyx, `getConnectedIntegration(policy)` returns `undefined`, the whole `if (connectedIntegration)` block is skipped, and both integration options disappear.

This PR mirrors that snapshot fallback for the policy reads in [`src/hooks/useSearchBulkActions.ts`](https://github.com/Expensify/App/blob/becf8124589/src/hooks/useSearchBulkActions.ts):

- The single-policy `policy` read now uses `getPolicyFromSearchSnapshot(...)`.
- The per-report `reportPolicy` read inside `canReportBeExported` now uses `getPolicyFromSearchSnapshot(...)`.

`getPolicyFromSearchSnapshot` prefers live Onyx and falls back to the snapshot, so current workspace settings still win when available — the only behavioral change is that the integration gate no longer requires the policy to already be present in live Onyx.

**Tests:** the existing `useSearchBulkActionsExportTest.ts` mocked `getPolicyFromSearchSnapshot` as a no-op; it is now a faithful mock (live-first, snapshot-fallback), and the same faithful mock was added to the sibling suites that mock `@libs/actions/Search`. A new test covers the fresh-load case where the policy exists **only** in the search snapshot and verifies both integration options appear.

### Fixed Issues
$ https://github.com/Expensify/App/issues/95820
PROPOSAL:

### Tests

Preconditions

 - You have a workspace connected to an accounting integration (e.g., NetSuite).
 - The workspace contains a list of reports with a status of Approved or Paid.

1. Go to Troubleshoot > Clear cache and restart.
2. Navigate to Spend > Reports.
3. Select Report with `Approved` status.
4. Verify that both export options (Export to NetSuite and Mark as Exported) are available in the bulk action menu without opening any report first.

### Offline tests

N/A.

### QA Steps

Same as Tests step.

### PR Author Checklist
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/hooks/useSearchBulkActions.ts
```diff
@@ -1468,7 +1468,9 @@ function useSearchBulkActions({queryJSON}: UseSearchBulkActionsParams) {
 
             const includeReportLevelExport = ((isExpenseReportType || typeInvoice) && areFullReportsSelected) || (typeExpense && !isExpenseReportType && isAllOneTransactionReport);
 
-            const policy = selectedPolicyIDs.length === 1 ? policies?.[`${ONYXKEYS.COLLECTION.POLICY}${selectedPolicyIDs.at(0)}`] : undefined;
+            // Read the policy through the Search-snapshot fallback: this hook renders outside SearchScopeProvider, so on a
+            // fresh load / cache clear the policy may only exist in the search snapshot and not yet in live Onyx.
+            const policy = selectedPolicyIDs.length === 1 ? getPolicyFromSearchSnapshot(selectedPolicyIDs.at(0), currentSearchResults?.data, policies) : undefined;
```
ahmedGaber93: @MelvinBot Can we use `currentSearchResults?.data?.` directly instead of getPolicyFromSearchSnapshot since `getPolicyFromSearchSnapshot` prefer policy from live onyx 

https://github.com/Expensify/App/blob/1e4e79e63788e8ed54cd74c97a9449c88b374856/src/libs/actions/Search.ts#L139-L147

MelvinBot: We could, but there's a real behavioral difference — both approaches fix the reported bug, they only differ in the tie-break when the policy exists in **both** live Onyx and the snapshot.

- [`getPolicyFromSearchSnapshot`](https://github.com/Expensify/App/blob/1e4e79e63788e8ed54cd74c97a9449c88b374856/src/libs/actions/Search.ts#L138-L147) is **live-first**: `policies?.[policy_<id>] ?? snapshotPolicy`.
- Reading [`currentSearchResults?.data?.[`policy_<id>`]`](https://github.com/Expensify/App/blob/2d5e41baa69a4c3425752de752e4aa1a5c5e78f8/src/hooks/useSearchBulkActions.ts#L1473) directly is **snapshot-only**.

Both resolve the fresh-load case (policy missing from live Onyx → fall back to / read from snapshot), so the reported options-disappearing bug is fixed either way. The trade-off:

- **Live-first (current PR):** if the user just changed their workspace's accounting connection, `getConnectedIntegration` reflects that immediately. It's also consistent with the report read right below it, which routes through the sibling `getReportFromSearchSnapshot` helper, and reuses the shared helper instead of duplicating the key construction inline.
- **Snapshot-only (your suggestion):** more robust *only if* live Onyx can hold a **partial** policy that's missing `connections` — in that case live-first would return the incomplete live policy and re-hide the options. If that scenario can't happen on this path, the two are equivalent aside from staleness.

My recommendation is to keep the helper: it gives current workspace settings when available, keeps the two reads consistent, and avoids inlining the collection-key string. The only reason to switch to snapshot-direct is if we expect the live policy to be present-but-incomplete here.

If you'd prefer snapshot-direct, I'm happy to make the change — just confirm and I'll push it (and update the faithful test mocks accordingly).

ahmedGaber93: @MelvinBot Please fix this ^

ahmedGaber93: @MelvinBot The issue is on the first load live policy come without `policy.connections`, so we need to use the snapshotPolicy instead

MelvinBot: Done — pushed [`972d7fb`](https://github.com/Expensify/App/commit/972d7fba2fb81d4fb0fcc0a2c646f6c2b35fb783). Both policy reads in the export block now read from the Search snapshot instead of the live-first helper:

- [`useSearchBulkActions.ts:1473`](https://github.com/Expensify/App/blob/972d7fba2fb81d4fb0fcc0a2c646f6c2b35fb783/src/hooks/useSearchBulkActions.ts#L1473) (single-policy `policy`)
- [`useSearchBulkActions.ts:1492`](https://github.com/Expensify/App/blob/972d7fba2fb81d4fb0fcc0a2c646f6c2b35fb783/src/hooks/useSearchBulkActions.ts#L1492) (`reportPolicy`)

**One deliberate deviation:** I read `currentSearchResults?.data?.[policy_<id>] ?? policies?.[policy_<id>]` — snapshot-**first with a live fallback** — rather than snapshot-only. A pure snapshot-only read regresses the existing test [`useSearchBulkActionsExportTest.ts:377`](https://github.com/Expensify/App/blob/972d7fba2fb81d4fb0fcc0a2c646f6c2b35fb783/tests/unit/hooks/useSearchBulkActionsExportTest.ts#L377): there the report arrives fresh from search (snapshot only) while the policy is already loaded in **live Onyx** and never written to the snapshot, so a snapshot-only read returns `undefined` and re-hides the integration options. Snapshot-first fixes your concern (the snapshot now wins over live) while keeping that case working, and it mirrors the sibling [`getReportFromSearchSnapshot`](https://github.com/Expensify/App/blob/1e4e79e63788e8ed54cd74c97a9449c88b374856/src/libs/actions/Search.ts#L128-L136) which is already `snapshot ?? live`.

Types are unchanged: this is the same expression `getPolicyFromSearchSnapshot` uses internally (just snapshot-first), and it matches the existing direct snapshot read at [`useSearchBulkActions.ts:504`](https://github.com/Expensify/App/blob/972d7fba2fb81d4fb0fcc0a2c646f6c2b35fb783/src/hooks/useSearchBulkActions.ts#L504). I couldn't run the local typecheck/lint/test tooling in this environment, so I verified statically against those precedents — CI will run the full checks.

If you truly want snapshot-**only** (dropping the live fallback), let me know and I'll adjust that test to match, but flagging it because it would drop the policy-already-in-live-Onyx case.

MelvinBot: Exactly right — that's the case I was unsure about, and you've confirmed it: on first load the live policy can be present but without its `connections` yet, so a live-first read (`getPolicyFromSearchSnapshot`) would return the connection-less live policy and drop the whole integration block.

The snapshot-first read is already in [`972d7fb`](https://github.com/Expensify/App/commit/972d7fba2fb81d4fb0fcc0a2c646f6c2b35fb783) (`currentSearchResults?.data?.[key] ?? policies?.[key]`), so the snapshot policy — which does carry `connections` — now wins.

To lock that behavior in, I just pushed [`11bd281`](https://github.com/Expensify/App/commit/11bd28154) adding a regression test for this precise scenario: a live policy of `{id}` **with no connections** plus a fully-connected snapshot policy. It fails under a live-first read (options disappear) and passes with the snapshot-first read. Full suite:

```
✓ offers the integration export options when the report is only in the search snapshot (fresh load)
✓ offers the integration export options when the policy is only in the search snapshot (fresh load)
✓ prefers the snapshot policy when the live policy is present but has not loaded its connections (fresh load)
✓ does NOT offer the integration export options when the report is absent from both snapshot and live Onyx
Tests: 4 passed, 4 total
```

ahmedGaber93:
> I'll pause work on this for now because I'm unable to reproduce the original bug.

@MelvinBot please close this PR

PR #96040, opened 2026-07-14
refactor: extract queue-derived pending-request state into PendingRequests module

BartekObudzinski:
### Explanation of Change

Loading indicators that read from a stored boolean flag can get stuck on forever if the write that clears the flag is lost, for example when the app closes or reloads in the gap between a request finishing and the flag being cleared. The LoadingBar avoids this today by deriving its visibility straight from the request queue instead: a request is always removed from the queue when it resolves, so there is no separate flag that can go stale.

This PR extracts that pattern into `src/libs/PendingRequests`, a module that answers &#34;is a request of this kind currently in flight?&#34; by reading the two Onyx keys behind the queue (`PERSISTED_REQUESTS` and `PERSISTED_ONGOING_REQUESTS`). It defines three command groups (app load, report load scoped by `reportID`, and the LoadingBar&#39;s own command set) and exposes one hook per group: `useIsAppLoadPending()`, `useIsReportLoadPending(reportID)`, and `useIsLoadingBarPending()`. No new Onyx writes or connections are added.

`useLoadingBarVisibility` is refactored to call `useIsLoadingBarPending()` instead of keeping its own copy of the selectors. The behavior is unchanged, including the asymmetric offline filter (a persisted request initiated while offline does not count as pending, but the ongoing request always does).

The other two hooks are not consumed yet. They exist so later PRs can move loading and skeleton gates that currently rely on stored flags onto the same queue-derived pattern, one call site at a time.

### Fixed Issues

$ https://github.com/Expensify/App/issues/96060
PROPOSAL:

### Tests

1. Online, open any report (or pull to refresh) so an OpenApp/ReconnectApp/OpenReport/ReadNewestAction request fires. Verify the thin LoadingBar under the header appears while the request is in flight and disappears once it resolves, exactly as on main.
2. Go offline and send a message in any report so a write command queues without sending. Verify the LoadingBar stays hidden for the queued request.
3. Come back online with that request still queued. Verify the LoadingBar appears while the request is actively sending, then disappears.

### Offline tests

1. Go offline and trigger a command from the LoadingBar set (e.g. send a message so a write command queues). Verify the LoadingBar stays hidden while the request sits in the offline queue.
2. Come back online. Verify the LoadingBar appears once the queued request starts actively sending and disappears when it resolves.

### QA Steps

Same as tests

### PR Author Checklist

- [x] I ran the tests on **all platforms** &amp; verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] I verified there are no console errors (if there&#39;s a console error not related to the PR, report it or open an issue for it to be fixed)
    - [x] I verified that any new or modified comments were clear, correct English, and explained &#34;why&#34; the code was doing something instead of only explaining &#34;what&#34; the code was doing.
    - [x] A similar style doesn&#39;t already exist
    - [x] The style can&#39;t be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/styles/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG)`)

### Screenshots/Videos

Android: Native

Android: mWeb Chrome

iOS: Native

iOS: mWeb Safari

MacOS: Chrome / Safari


src/hooks/useInFlightRequests.ts
adhorodyski, line 1: isn't it overlapping with the SQ module that is named similar?

adhorodyski: It was super weird to me to stumble upon this given I know SQ already. If you're only exporting hooks from there, should we name it better for this context?

src/hooks/useInFlightRequests.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * Whether an OpenReport request for the given report is currently in the queue.
+ *
+ * Do not call this inside list-item render paths (e.g. per row in a list): every call opens two Onyx
+ * subscriptions. Lift it to the screen level and pass the result down instead.
+ */
+function useIsReportLoadPending(reportID: string): boolean {
+    return useIsPendingInternal('reportLoad', reportID);
```
adhorodyski, line 110: I know this is type safe (and I personally prefer string literals for this) but I'm not sure we follow this pattern anywhere else in EApp (we're rather relying on strict constants usage). I appreciate it though, love `type PendingRequestGroup = keyof typeof PENDING_REQUEST_GROUPS;`!

adhorodyski (changes requested):
I think it's not really NIT but the naming could be improved for the pending stuff module.

src/hooks/useInFlightRequests.ts
BartekObudzinski: Fair point, the name sat too close to PersistedRequests. Moved the whole thing to src/hooks/useInFlightRequests.ts since it only exports hooks anyway. The exported hook names stay the same.

src/hooks/useInFlightRequests.ts
BartekObudzinski: These ids never leave the module. They are typed with keyof typeof over the registry, so a typo does not compile, which gives the same guarantee CONST would. For keys internal to one file, adding them to CONST.ts felt like overhead. Happy to move them if you feel strongly about it.

src/hooks/useInFlightRequests.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+function useLoadingBarVisibility(): boolean {
+    const hasPendingLoadingBarRequest = useIsPendingInternal('loadingBar');
+    const {isOffline} = useNetwork();
+
+    // Don't show loading bar if currently offline
+    return !isOffline && hasPendingLoadingBarRequest;
+}
+
+export {useIsAppLoadPending, useIsReportLoadPending, useIsLoadingBarPending, useLoadingBarVisibility};
```
DylanDylann, line 129: @BartekObudzinski Looks good overall! I'm just curious how you're planning to use these hooks in the upcoming PRs?

BartekObudzinski: Thanks! The plan is to move the loading gates off stored flags and onto these hooks, one screen at a time. First the app-load surfaces (e.g. WorkspacesListPage, which today gates on IS_LOADING_APP) switch to useIsAppLoadPending, then the per report surfaces switch to useIsReportLoadPending scoped by reportID. Each screen keeps its own data and offline checks, the hook only answers whether a matching request is in flight right now. Once a family of screens is migrated and verified, we delete the old stored flag and its recovery timer. This PR just lands the mechanism and moves LoadingBar onto it so nothing changes behavior yet.

BartekObudzinski: @DylanDylann

src/hooks/useInFlightRequests.ts
adhorodyski: love it now

src/hooks/useInFlightRequests.ts
adhorodyski: yeah i totally get the type guarantee, fine with me 👍🏼

src/hooks/useLoadingBarVisibility.ts
```diff
@@ -1,33 +1,3 @@
...
-    const {isOffline} = useNetwork();
-
-    // Don't show loading bar if currently offline
-    if (isOffline) {
-        return false;
-    }
-
-    return !!hasRelevantPersistedRequests || !!hasRelevantOngoingRequest;
-}
+export default useLoadingBarVisibility;
```
adhorodyski: I would not reexport, but either put the implementation (that uses internals) directly there OR use the module itself. This file seems redundant then.

tests/unit/useInFlightRequestsTest.ts
```diff
@@ -0,0 +1,142 @@
...
+            const {result: matching} = renderHook(() => useIsReportLoadPending('1234'));
+            await waitFor(() => expect(matching.current).toBe(true));
+
+            const {result: nonMatching} = renderHook(() => useIsReportLoadPending('5678'));
+            await act(() => waitForBatchedUpdates());
+            expect(nonMatching.current).toBe(false);
+        });
+    });
+
+    describe('useIsLoadingBarPending', () => {
```
adhorodyski, line 116: I think these test describes guide themselves into 'we want a separate file' :P but that's just NIT

adhorodyski (approved):
NIT only, this is fine I think

src/hooks/useLoadingBarVisibility.ts
DylanDylann: agree

BartekObudzinski: Good call. Removed the re-export file and pointed the four consumers straight at the module (import {useLoadingBarVisibility} from '@hooks/useInFlightRequests'). Putting the implementation back in its own file is what reintroduced the compiler memoization divergence, so keeping it in the module is the way.

tests/unit/useInFlightRequestsTest.ts
BartekObudzinski: The three describes just mirror the public hooks, but they all share one registry and the internal generic, so a single module with a single test file is deliberate. Kept it as one.

DylanDylann (approved):

mountiny (approved):
Thanks

PR #96042, opened 2026-07-14
Add top-level navigation suggestions to Search Router

nabi-ebrahimi:
### Explanation of Change

This PR introduces the core Search Router navigation infrastructure and adds navigation suggestions for the five top-level destinations:

- Home
- Inbox
- Spend
- Workspaces
- Account

Navigation suggestions are rendered inside the existing `SelectionListWithSections` flow, preserving keyboard focus, Enter selection, and existing Search Router row ordering.

The suggestions:

- Support plain queries and case-insensitive `go` / `go to` prefixes.
- Require at least three destination characters, except for the intentional bare `go` and `go to` queries.
- Use accent-insensitive matching.
- Sort results alphabetically and limit them to eight.
- Display localized `Go to {destination}` labels.
- Close Search Router before executing navigation to avoid modal and route-transition conflicts.
- Preserve the existing last Spend search.
- Open the Workspaces list instead of restoring a previously visited workspace subpage.

This is the first focused part of #92752. It intentionally does not add Spend submenu, Account settings, Workspace settings, Domain, or Create/FAB suggestions.

### Fixed Issues

### Tests

1. **Plain navigation query**
   - Open Search Router.
   - Search separately for `home`, `inbox`, `spend`, `workspaces`, and `account`.
   - Select each navigation row.
   - **Expected:** The matching `Go to ...` row appears and opens the correct destination.

2. **Navigation intent prefixes**
   - Search for `go inbox` and `go to inbox`.
   - Repeat with mixed capitalization, such as `GO TO INBOX`.
   - **Expected:** All queries show `Go to Inbox` and selecting it opens Inbox.

3. **Minimum query length**
   - Search separately for `a`, `ab`, and `acc`.
   - **Expected:** No navigation row appears for `a` or `ab`. `acc` shows `Go to Account`.

4. **Bare navigation intent and ordering**
   - Search for `go`, then `go to`.
   - **Expected:** Both queries show the five top-level destinations, alphabetically ordered and limited to the navigation results supported by this PR.

5. **Keyboard selection**
   - Search for `inbox`.
   - Focus `Go to Inbox` using Arrow Up/Down and press Enter.
   - **Expected:** Search Router closes and Inbox opens once.

6. **Spend navigation**
   - Open Spend and enter a recognizable search.
   - Navigate elsewhere.
   - Use Search Router to select `Go to Spend`.
   - **Expected:** Spend opens using its existing last-search behavior.

7. **Workspaces root navigation**
   - Open a workspace subpage such as Members.
   - Navigate to Home.
   - Use Search Router to select `Go to Workspaces`.
   - **Expected:** The Workspaces list opens instead of the previously visited workspace subpage.

8. **Modal close before navigation**
   - Select any `Go to ...` row.
   - Reopen Search Router after arriving at the destination.
   - **Expected:** Search Router closes before navigation, the destination opens once, and no stale modal or delayed navigation remains.

9. **Localized navigation text**
   - Switch to one non-English language.
   - Search for a localized top-level destination.
   - **Expected:** The localized `Go to {destination}` row appears and navigates correctly.

### Offline tests
1. Log in and allow the app to finish loading.
2. Disconnect from the network.
3. Open Search Router.
4. Search for Home, Inbox, Spend, Workspaces, and Account.
5. Select each destination.
6. Reconnect to the network.

**Expected:** Cached navigation suggestions remain available, local navigation does not crash, and the app recovers normally after reconnecting.

### QA Steps
Same as Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/4055021a-9fa2-4f18-9de8-0d8c0c8077e9

https://github.com/user-attachments/assets/580eee44-c0b4-454f-8d54-9239ecd08b18

</details>

<details>

>

https://github.com/user-attachments/assets/ad5c1e41-c378-4e3e-ab7c-1283ae9aecbb

https://github.com/user-attachments/assets/74b01073-cc91-47ef-9bd9-87ab80067b3d

</details>

<details>

>

https://github.com/user-attachments/assets/0487359d-cba1-4f40-910f-52c7809d6391

https://github.com/user-attachments/assets/8db08069-9c7c-4941-9476-e1dbff492a26

</details>

<details>

>

https://github.com/user-attachments/assets/800c3204-b781-47f5-be08-0cc64c2fa173

https://github.com/user-attachments/assets/520ed3fd-6434-4953-91ef-ed833c1f1b18

</details>

<details>

>
01

https://github.com/user-attachments/assets/1b3aeee3-595e-4cbc-90b0-b42f843eb68d

02

https://github.com/user-attachments/assets/f414486c-6f01-4bd7-bf4e-38e926d152c8

03

https://github.com/user-attachments/assets/d009adb5-77bc-48fe-b065-df3d0aa304b5

04

https://github.com/user-attachments/assets/114dffd5-7b03-4a44-ab98-5f67a006ac33

05

https://github.com/user-attachments/assets/c4a25b35-33a9-4cdc-ace6-5950a9af93c5

06

https://github.com/user-attachments/assets/15782f0e-2dc9-4e95-83ea-47278defc1f7

07

https://github.com/user-attachments/assets/e5a2ec4c-77bf-4c40-bd01-80ba1a8acbe9

08

https://github.com/user-attachments/assets/98c75fd9-6766-4973-a416-4ca2a28e2c5e

09

https://github.com/user-attachments/assets/aa7b730a-eb5d-4de6-bdba-6320502d9e62

</details>


src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,89 @@
...
+    const isNavigationIntentOnly = isNavigationIntentOnlyQuery(trimmedQuery);
+    const matchQuery = stripNavigationIntentPrefix(trimmedQuery) || trimmedQuery;
+    const shouldMatchExactDestination = matchQuery.length < MIN_NAVIGATION_QUERY_LENGTH && !isNavigationIntentOnly;
+
+    return sources
+        .flatMap((source) => sortNavigationSuggestionItems(source, localeCompare))
+        .filter(
+            (item) =>
+                isNavigationIntentOnly ||
+                (shouldMatchExactDestination ? matchesNavigationQueryExactly(matchQuery, ...(item.matchTerms ?? [])) : matchesNavigationQuery(matchQuery, ...(item.matchTerms ?? []))),
```
marufsharifi, line 86: **Small perf nit:** could we return `[]` before this chain when `!matchQuery && !isNavigationIntentOnly`? Empty queries still run through `flatMap`, sorting, filtering, and mapping despite producing no navigation rows. This preserves `go` and short localized-destination behavior while avoiding unnecessary work on empty-input renders.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -7,6 +7,9 @@ import React from 'react';
 function SearchRouterPage() {
     const {closeSearchRouter} = useSearchRouterActions();
     const {isSearchRouterDisplayed} = useSearchRouterState();
+    const closeSearchRouterAndRunAction = (afterClose?: () => void) => {
+        closeSearchRouter(afterClose);
+    };
```
marufsharifi: **NAB:** `closeSearchRouterAndRunAction` only forwards `afterClose` to `closeSearchRouter`, so the wrapper doesn’t add behavior. Could we pass `closeSearchRouter` directly instead?

```tsx
<SearchRouter onRouterClose={closeSearchRouter} />
```

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function getGoToText(translate: LocaleContextProps['translate'], destination: string) {
+    return translate('search.goTo', {destination});
+}
+
+function buildNavigationSuggestions(query: string, sources: NavigationSuggestionSourceItem[][], localeCompare: LocaleContextProps['localeCompare']): SearchQueryItem[] {
+    const trimmedQuery = query.trim();
+    const isNavigationIntentOnly = isNavigationIntentOnlyQuery(trimmedQuery);
+    const matchQuery = stripNavigationIntentPrefix(trimmedQuery) || trimmedQuery;
```
marufsharifi: **NAB:** This predicate combines three matching modes—bare intent, short exact destination matching, and longer fuzzy matching—so the localized short-destination exception is easy to miss. Could we extract a `doesNavigationSuggestionMatch(...)` helper, or add a short comment documenting those modes?

marufsharifi (changes requested):

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,89 @@
...
+    matchTerms?: string[];
+};
+
+const MAX_NAVIGATION_SUGGESTIONS = 8;
+const MIN_NAVIGATION_QUERY_LENGTH = 3;
+
+function stripNavigationIntentPrefix(query: string) {
+    const trimmedQuery = query.trim();
+    if (/^go to\s+/i.test(trimmedQuery)) {
+        return trimmedQuery.replace(/^go to\s+/i, '').trim();
```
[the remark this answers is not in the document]
marufsharifi: @nabi-ebrahimi, this seems valid to me. What do you think? 


https://github.com/user-attachments/assets/063f4b6f-f9a6-46b3-af0c-85684b56081c

src/components/Search/SearchRouter/SearchRouterContext.tsx
```diff
@@ -33,7 +33,7 @@ type SearchRouterStateContextType = {
 
 type SearchRouterActionsContextType = {
     openSearchRouter: (query?: string, isFromSearchPage?: boolean) => void;
-    closeSearchRouter: () => void;
+    closeSearchRouter: (afterTransition?: () => void) => void;
```
marufsharifi, line 40: 🟡 **`afterTransition` is silently dropped on web — document (or unify) the two close mechanisms.**

This new parameter is only honored on **native** (forwarded to `Navigation.dismissModal({afterTransition})` in `toggleSearch/index.native.tsx`). The **web** `closeSearch` in `toggleSearch/index.tsx` ignores it entirely (`return setSearchState(false)`); on web the callback is instead run via `SearchRouterModal`'s `actionAfterModalHideRef` + `onModalHide`.

That works for today's single caller, but the context API now advertises a param that does nothing on web unless the call routes through `closeSearchRouterAfterModalHide`. A future `closeSearchRouter(() => …)` from anywhere else will have its callback **silently dropped on web only** — a platform-divergent bug that survives single-platform local testing.

```suggestion
    /**
     * Closes the Search Router. `afterTransition` runs after the close animation completes.
     * NOTE: on web this is honored ONLY when closing via SearchRouterModal (run in `onModalHide`);
     * calling it directly on web drops the callback. On native it is forwarded to
     * `Navigation.dismissModal({afterTransition})`.
     */
    closeSearchRouter: (afterTransition?: () => void) => void;
```

src/components/Search/SearchRouter/SearchRouterModal/index.tsx
```diff
@@ -28,14 +29,25 @@ function SearchRouterModal() {
...
+    const closeSearchRouterAfterModalHide = (afterClose?: () => void) => {
+        actionAfterModalHideRef.current = afterClose;
+        closeSearchRouter();
+    };
+
+    const handleModalHide = () => {
+        setShouldHideInputCaret(isMobileWebIOS);
+        actionAfterModalHideRef.current?.();
+        actionAfterModalHideRef.current = undefined;
+    };
```
marufsharifi, line 54: 🟡 **Stale-callback footgun: `actionAfterModalHideRef` only clears inside `onModalHide`.**

```ts
const closeSearchRouterAfterModalHide = (afterClose?: () => void) => {
    actionAfterModalHideRef.current = afterClose;   // armed unconditionally
    closeSearchRouter();
};
const handleModalHide = () => {                     // the only place it fires + clears
    setShouldHideInputCaret(isMobileWebIOS);
    actionAfterModalHideRef.current?.();
    actionAfterModalHideRef.current = undefined;
};
```

If `closeSearchRouterAfterModalHide(action)` is ever invoked while the modal is **already hidden**, `closeSearchRouter()` is a state no-op, `onModalHide` won't fire, and the ref stays armed — then the **next** modal hide fires that stale action and navigates the user somewhere unexpected.

Not reachable in today's flow (nav rows are only pressable while the modal is shown), so not a confirmed live bug — but it's a fragile invariant one refactor away from breaking. Consider arming the ref only when a hide will actually occur (guard on `isSearchRouterDisplayed`), or clearing it if `closeSearchRouter` runs while already hidden.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -0,0 +1,101 @@
+/**
+ * Builds the top-level navigation suggestions shown in the Search Router.
```
marufsharifi, line 2: 🔵 **Nit: this file has no JSX — rename to `useNavigationSuggestions.ts`.**

It only exports a hook + `buildTopLevelNavigationItems`, both returning plain objects. Its sibling `SearchRouterHelpers.ts` is already `.ts`, and the test declares `mockIcon: IconAsset = () => null` in plain `.ts`. Using `.ts` keeps the convention consistent and avoids the unnecessary TSX parse.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+        .filter(
+            (item) =>
+                // Bare intents show every destination, short localized labels require exact matches, and longer queries allow partial matches.
+                isNavigationIntentOnly ||
+                (shouldMatchExactDestination ? matchesNavigationQueryExactly(matchQuery, ...(item.matchTerms ?? [])) : matchesNavigationQuery(matchQuery, ...(item.matchTerms ?? []))),
+        )
+        .map((item) => ({
+            ...item,
+            searchItemType: CONST.SEARCH.SEARCH_ROUTER_ITEM_TYPE.NAVIGATE,
+        }))
```
marufsharifi: 🔵 **Minor: `matchTerms` leaks into the returned `SearchQueryItem[]`.**

`matchTerms` is only on `NavigationSuggestionSourceItem`, not on `SearchQueryItem` (see `SearchQueryListItem.tsx`), but spreading `...item` carries it into the final rendered items. Harmless (nothing reads it downstream) but untidy — drop it in the map so the returned type is exact:

```suggestion
        .map(({matchTerms, ...item}) => ({
            ...item,
            searchItemType: CONST.SEARCH.SEARCH_ROUTER_ITEM_TYPE.NAVIGATE,
        }))
```

(If `@typescript-eslint/no-unused-vars` flags the omitted `matchTerms`, this repo likely has `ignoreRestSiblings` enabled; otherwise use an explicit omit helper.)

src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -381,6 +386,13 @@ function SearchRouter({onRouterClose, shouldHideInputCaret, isSearchRouterDispla
             };
 
             if (isSearchQueryItem(item)) {
+                if (item.searchItemType === CONST.SEARCH.SEARCH_ROUTER_ITEM_TYPE.NAVIGATE && item.action) {
+                    backHistory(() => {
+                        onRouterClose(item.action);
+                    });
+                    return;
+                }
```
marufsharifi, line 394: 🔵 **A `NAVIGATE` item missing `action` fails silently.**

```ts
if (item.searchItemType === CONST.SEARCH.SEARCH_ROUTER_ITEM_TYPE.NAVIGATE && item.action) {
    backHistory(() => { onRouterClose(item.action); });
    return;
}
if (!item.searchQuery) {
    return;   // a NAVIGATE row with no `action` lands here and does nothing
}
```

Reasonable defensive handling, but a malformed nav item (type `NAVIGATE`, no `action`) silently no-ops and presents a dead row, which could mask a construction bug. Not blocking — consider asserting/logging in `__DEV__` so it surfaces during development.

src/components/Navigation/NavigationTabBar/getSearchTabRoute.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import type {NavigationState} from '@react-navigation/native';
+import type {OnyxEntry} from 'react-native-onyx';
+
+import getLastRoute from './getLastRoute';
+
+function getSearchTabRoute(rootState: NavigationState, lastSearchParams: OnyxEntry<LastSearchParams>) {
+    const lastSearchRoute = getLastRoute(rootState, NAVIGATORS.SEARCH_FULLSCREEN_NAVIGATOR, SCREENS.SEARCH.ROOT);
+
+    if (isSearchRootParams(lastSearchRoute?.params)) {
```
marufsharifi, line 20: 👍 Replacing the old `params as SearchFullscreenNavigatorParamList[…]` cast with the `isSearchRootParams()` type guard removes a forbidden type cast **and** validates `q` is a string before use — cleaner and safer than the original `SearchTabButton` code. Nice extraction into a pure, testable function.

src/components/Search/SearchRouter/SearchRouterModal/index.tsx
```diff
@@ -19,6 +19,7 @@ function SearchRouterModal() {
     const {shouldUseNarrowLayout} = useResponsiveLayout();
     const {isSearchRouterDisplayed} = useSearchRouterState();
     const {closeSearchRouter} = useSearchRouterActions();
+    const actionAfterModalHideRef = useRef<() => void>(undefined);
```
marufsharifi, line 22: 👍 Good catch wrapping this. Now that `closeSearchRouter` takes `afterTransition` as its first arg, passing it directly to `addEventListener` would have handed the `{window, screen}` resize payload in as `afterTransition` — a latent crash on native once that value reached `dismissModal`.

marufsharifi (changes requested):
Code looks good and tested, but there are a few minor feedback. thanks.

src/components/Search/SearchRouter/SearchRouterContext.tsx
nabi-ebrahimi: Addressed

src/components/Search/SearchRouter/SearchRouterModal/index.tsx
nabi-ebrahimi: Addressed

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Thanks for the review, This is safe as-is. The upcoming Spend navigation follow-up adds JSX for the right-side context in this file, so renaming it to `.ts` now would require renaming it back to `.tsx` immediately and create unnecessary churn in the stacked PR.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Addressed

src/components/Search/SearchRouter/SearchRouter.tsx
nabi-ebrahimi: Thanks for flagging this. All `NAVIGATE` rows are created by controlled builders that always provide an `action`, and the destination actions are covered by unit tests. Since this is not currently reachable and additional logging or broader type changes would only add defensive complexity, I suggest keeping the tested implementation unchanged in this PR.

src/components/Navigation/NavigationTabBar/getSearchTabRoute.ts
nabi-ebrahimi: Thanks for confirming!

src/components/Search/SearchRouter/SearchRouterModal/index.tsx
nabi-ebrahimi: Thanks for confirming!

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -0,0 +1,101 @@
...
+            spend: translate('common.spend'),
+            workspaces: translate('common.workspacesTabTitle'),
+            account: translate('initialSettingsPage.account'),
+        },
+        icons,
+        getSpendRoute: () => getSearchTabRoute(navigationRef.getRootState(), lastSearchParams),
+        getDestinationText: (destination) => getGoToText(translate, destination),
+    });
+
+    return buildNavigationSuggestions(query, [topLevelItems], localeCompare);
```
marufsharifi, line 98: **⚡ Performance — medium (conditional on React Compiler)**

`buildNavigationSuggestions` allocates a **new array on every call** (`.slice(...)`, and a fresh `[]` for the common no-match case), so this hook returns a new array identity on every render. That value is added to the `getAdditionalSections` `useCallback` deps in `SearchRouter.tsx`, and `getAdditionalSections` is itself a dependency of the `sections` `useMemo` in `SearchAutocompleteList.tsx` (L611) — which maps over `recentReportsOptions` and runs `Parser.htmlToText()` per recent report (L510–524).

If this hook is **not** memoized by React Compiler on *both* pipelines (Babel for native/Jest, OXC for web), every SearchRouter re-render that doesn't change the debounced query — arrow-key list navigation, `setSelection`, autocomplete-substitution updates — will rebuild the whole section list and re-parse HTML for every recent chat. That's wasted work on an interactive path, and it scales with the user's recent-report count.

**First, verify** the compiler already handles it: `npm run react-compiler-compliance-check check-changed`. React Compiler keys on the icon *member reads* and the query, so if it compiles here, no change is needed.

**If it isn't compiled**, stabilize manually — note you must key on the individual icon values, not the `icons` object (that object is intentionally a fresh reference each render, see `useLazyAsset.ts` L273–277, so `[icons]` would defeat the memo):

```tsx
return useMemo(
    () => buildNavigationSuggestions(
        query,
        [buildTopLevelNavigationItems({labels: {/* … */}, icons, getSpendRoute, getDestinationText})],
        localeCompare,
    ),
    [query, translate, localeCompare, lastSearchParams, icons.Home, icons.Inbox, icons.ReceiptMultiple, icons.Building, icons.Gear],
);
```

Also consider returning a shared `EMPTY` constant instead of `[]` from the early return in `SearchRouterHelpers.ts:73`, so the no-suggestion case keeps a stable identity too.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+    return trimmedQuery;
+}
+
+function isNavigationIntentOnlyQuery(query: string) {
+    return /^go(?:\s+to)?$/i.test(query.trim());
+}
+
+function matchesNavigationQuery(query: string, ...values: Array<string | undefined>) {
+    const normalizedQuery = StringUtils.normalizeAccents(query).toLowerCase();
```
marufsharifi, line 36: **⚡ Performance — low / informational**

`matchesNavigationQuery` recomputes `StringUtils.normalizeAccents(query).toLowerCase()` on every call, and it's called once per item inside `buildNavigationSuggestions`'s `.filter(...)` — so the same query string is normalized N times per build (`normalizeAccents` = `.normalize('NFD')` + regex replace, not free). Same pattern in `matchesNavigationQueryExactly`.

At today's **fixed 5 destinations** this is negligible — no change required now. Flagging only because these helpers are written generically over `sources: [][]`: if the suggestion set ever grows (workspace/report lists, etc.), hoist the normalized query out of the per-item loop and pass it in, e.g. `matchesNavigationQuery(normalizedQuery, ...values)` with the normalization done once in `buildNavigationSuggestions`.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+function matchesNavigationQueryExactly(query: string, ...values: Array<string | undefined>) {
+    const normalizedQuery = StringUtils.normalizeAccents(query).toLowerCase();
+    return values.some((value) => StringUtils.normalizeAccents(value ?? '').toLowerCase() === normalizedQuery);
+}
+
+function sortNavigationSuggestionItems<T extends NavigationSuggestionSourceItem>(items: T[], localeCompare: LocaleContextProps['localeCompare']): T[] {
+    return [...items].sort((firstItem, secondItem) => {
+        const firstText = StringUtils.normalizeAccents(firstItem.text ?? '').toLowerCase();
+        const secondText = StringUtils.normalizeAccents(secondItem.text ?? '').toLowerCase();
+        const textComparison = localeCompare(firstText, secondText);
```
marufsharifi, line 57: **⚡ Performance — low / informational**

The comparator normalizes both items' text (`StringUtils.normalizeAccents(...).toLowerCase()`) on **every comparison**, so sorting does O(N log N) `normalizeAccents` calls instead of O(N).

Negligible for the current **5 items** — no change needed now. If `sources` ever grows, use a decorate-sort-undecorate pass: precompute each item's normalized text once, sort on the precomputed key, then drop it.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -0,0 +1,101 @@
...
+        {
+            text: getDestinationText(labels.spend),
+            singleIcon: icons.ReceiptMultiple,
+            action: () => Navigation.navigate(getSpendRoute()),
+            keyForList: 'topLevelSpend',
+            matchTerms: [labels.spend],
+        },
+        {
+            text: getDestinationText(labels.workspaces),
+            singleIcon: icons.Building,
```
marufsharifi, line 65: NAB (consistency): This uses the `Building` (singular) icon for the Workspaces destination, but the canonical Workspaces tab in the bottom bar uses `Buildings` (plural) — see [`WorkspacesTabButton.tsx:26`](https://github.com/Expensify/App/blob/main/src/components/Navigation/NavigationTabBar/WorkspacesTabButton.tsx#L26) (`useMemoizedLazyExpensifyIcons(['Buildings'])`). The same destination would render a different icon depending on where it's shown.

Suggest switching to `Buildings` for consistency. Note this also requires updating the two other spots that name the icon: the `TopLevelNavigationIcons` type on **line 24** and the `useMemoizedLazyExpensifyIcons([...])` array on **line 81**.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -0,0 +1,101 @@
...
+        account: string;
+    };
+    icons: TopLevelNavigationIcons;
+    getSpendRoute: () => Route;
+    getDestinationText: (destination: string) => string;
+};
+
+function buildTopLevelNavigationItems({labels, icons, getSpendRoute, getDestinationText}: BuildTopLevelNavigationItemsParams): NavigationSuggestionSourceItem[] {
+    return [
+        {
```
marufsharifi, line 42: NAB (DRY / source-of-truth): `buildTopLevelNavigationItems` enumerates the five top-level destinations and their icons/routes a second time. The bottom tab bar already encodes the same destinations — `ROUTES.HOME`/`ROUTES.SETTINGS` in [`NavigationTabBar/index.tsx`](https://github.com/Expensify/App/blob/main/src/components/Navigation/NavigationTabBar/index.tsx), `ROUTES.INBOX` in [`InboxTabButton.tsx`](https://github.com/Expensify/App/blob/main/src/components/Navigation/NavigationTabBar/InboxTabButton.tsx), Workspaces in [`WorkspacesTabButton.tsx`](https://github.com/Expensify/App/blob/main/src/components/Navigation/NavigationTabBar/WorkspacesTabButton.tsx), and the tab↔route map in [`ROUTE_TO_NAVIGATION_TAB.ts`](https://github.com/Expensify/App/blob/main/src/components/Navigation/NavigationTabBar/ROUTE_TO_NAVIGATION_TAB.ts). A destination's route or icon now has to be changed in two places.

Not a blocker: there's no single data-driven destinations structure to reuse today (the tab bar is bespoke button components with genuinely different behavior — e.g. Inbox's last-report logic and the avatar-based Account tab), and the one truly reusable route (Spend's last search) was correctly extracted into `getSearchTabRoute`. Consolidating to a shared source-of-truth would be its own refactor. Flagging so we can either add a comment noting the intentional divergence or track consolidation as a follow-up.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
marufsharifi, line 36: NAB (minor DRY): `StringUtils.normalizeAccents(x).toLowerCase()` is repeated across six call sites in this file — lines 34, 40–42, 47, 48, 53 and 54. A small local helper would DRY these up:

```ts
const normalizeForMatch = (value?: string) => StringUtils.normalizeAccents(value ?? '').toLowerCase();
```

Genuinely optional though: the codebase deliberately inlines this idiom elsewhere (`shouldHighlight` in `SearchQueryUtils.ts`, `filterReports` in `OptionsListUtils/index.ts`), so leaving it inline is consistent with existing style. Take it or leave it.

src/components/Navigation/NavigationTabBar/SearchTabButton.tsx
```diff
@@ -54,27 +49,7 @@ function SearchTabButton({selectedTab, isWideLayout}: SearchTabButtonProps) {
...
-                    );
-                    return;
-                }
-            }
-
-            const lastQueryJSON = lastSearchParams?.queryJSON;
-            const lastQueryFromOnyx = lastQueryJSON ? buildSearchQueryString(lastQueryJSON) : undefined;
-            const defaultSearchQuery = buildCannedSearchQuery({type: CONST.SEARCH.DATA_TYPES.EXPENSE});
-            Navigation.navigate(ROUTES.SEARCH_ROOT.getRoute({query: lastQueryFromOnyx ?? defaultSearchQuery}));
+            Navigation.navigate(getSearchTabRoute(navigationRef.getRootState(), lastSearchParams));
```
marufsharifi, line 52: **Scope — low / non-blocking.** This PR is scoped to the Search Router top-level navigation feature, but this line refactors a *pre-existing, separate* entry point — the bottom-tab Search button — by replacing its inline last-search logic with the newly-extracted `getSearchTabRoute()` helper. Per Expensify's guidance that a needed refactor should ideally land as its own change first, I'm calling it out.

That said, it's justified: the new "Go to Spend" suggestion reuses this same helper (`useNavigationSuggestions.tsx:93`) so both entry points stay DRY, and "preserve the existing last Spend search" is a stated goal of this PR. My recommendation is **keep** — just flagging that the diff reaches outside the Search Router surface.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+function stripNavigationIntentPrefix(query: string) {
+    const trimmedQuery = query.trim();
+    if (/^go\s+to\s+/i.test(trimmedQuery)) {
+        return trimmedQuery.replace(/^go\s+to\s+/i, '').trim();
+    }
+    if (/^go\s+/i.test(trimmedQuery)) {
+        return trimmedQuery.replace(/^go\s+/i, '').trim();
+    }
+    return trimmedQuery;
+}
```
marufsharifi, line 29: **[low] readability**

In `stripNavigationIntentPrefix`, each regex literal (`/^go\s+to\s+/i`, `/^go\s+/i`) appears twice — once in `.test()`, once in `.replace()` — so the two copies have to be kept in sync by eye. Naming them once (they're stateless without the `g` flag, so reuse is safe) is clearer:

```suggestion
const MAX_NAVIGATION_SUGGESTIONS = 8;
const MIN_NAVIGATION_QUERY_LENGTH = 3;
const GO_TO_PREFIX = /^go\s+to\s+/i;
const GO_PREFIX = /^go\s+/i;

function stripNavigationIntentPrefix(query: string) {
    const trimmedQuery = query.trim();
    if (GO_TO_PREFIX.test(trimmedQuery)) {
        return trimmedQuery.replace(GO_TO_PREFIX, '').trim();
    }
    if (GO_PREFIX.test(trimmedQuery)) {
        return trimmedQuery.replace(GO_PREFIX, '').trim();
    }
    return trimmedQuery;
}
```

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+
+function getGoToText(translate: LocaleContextProps['translate'], destination: string) {
+    return translate('search.goTo', {destination});
+}
+
+function buildNavigationSuggestions(query: string, sources: NavigationSuggestionSourceItem[][], localeCompare: LocaleContextProps['localeCompare']): SearchQueryItem[] {
+    const trimmedQuery = query.trim();
+    const isNavigationIntentOnly = isNavigationIntentOnlyQuery(trimmedQuery);
+    const matchQuery = stripNavigationIntentPrefix(trimmedQuery) || trimmedQuery;
```
marufsharifi: **[low] readability**

The `|| trimmedQuery` fallback is non-obvious — it's hard to tell from the call site when `stripNavigationIntentPrefix` returns something falsy that we'd want to fall back on. A one-line "why" comment removes the guesswork:

```suggestion
    // Strip a leading "go"/"go to" intent; fall back to the raw query when stripping leaves nothing to match on.
    const matchQuery = stripNavigationIntentPrefix(trimmedQuery) || trimmedQuery;
```

src/components/Search/SearchRouter/toggleSearch/index.tsx
```diff
@@ -2,7 +2,8 @@ function openSearch(setSearchState: React.Dispatch<React.SetStateAction<boolean>
     return setSearchState(true);
 }
 
-function closeSearch(setSearchState: React.Dispatch<React.SetStateAction<boolean>>) {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars -- afterTransition is used on native only
+function closeSearch(setSearchState: React.Dispatch<React.SetStateAction<boolean>>, afterTransition?: () => void) {
```
marufsharifi, line 6: **[low] readability**

This file silences the unused parameter with an `eslint-disable-next-line`, while its native counterpart `toggleSearch/index.native.tsx:9` `_`-prefixes the unused arg (`_setIsSearchRouterDisplayed`). Same problem, two idioms — reads as accidental. Prefer the `_`-prefix here too, which satisfies the rule and drops the disable comment:

```suggestion
function closeSearch(setSearchState: React.Dispatch<React.SetStateAction<boolean>>, _afterTransition?: () => void) {
```

marufsharifi (changes requested):

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Thanks for highlighting this. I reran `npm run react-compiler-compliance-check check-changed`, and the hook passes with both Babel and OXC compilers, so the result is already memoized on both pipelines. Per your recommendation, no manual memoization is needed here.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Thanks for noting this. I agree the normalization should be hoisted once the navigation sources grow significantly. This PR only processes five fixed destinations, so I’ll keep the core change minimal and revisit it in the follow-up that introduces larger dynamic sources.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Thanks for noting this. Since this PR sorts only five fixed destinations, I’ll keep the implementation minimal. We can precompute normalized sort keys in the follow-up that introduces a larger dynamic source if profiling shows it is needed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: This row navigates directly to `WORKSPACES_LIST`, whose existing inner tab uses the singular `Building` icon. The plural `Buildings` icon represents the outer Workspaces tab containing both Workspaces and Domains. Since this action targets the Workspaces list specifically, I believe the current `Building` icon is intentional and consistent with the destination.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Thanks for the suggestion. Since the inline normalization is consistent with existing repository patterns and a helper would not change behavior or performance, I’ll keep the current implementation minimal.

src/components/Navigation/NavigationTabBar/SearchTabButton.tsx
nabi-ebrahimi: Thanks for confirming. I agree this extraction is justified because both entry points must preserve the same last Spend search behavior, and sharing the helper prevents their route logic from drifting.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: addressed.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/toggleSearch/index.tsx
nabi-ebrahimi: I tested the suggested `_afterTransition` naming, but this repository’s ESLint configuration still reports underscore-prefixed parameters as unused. The existing targeted disable is therefore required for the platform-specific signature, so I kept it unchanged.

marufsharifi (approved):
LGTM

carlosmiceli (approved):

trjExpensify (approved):
First PR of several to roll out navigational search 👍

PR #96394, opened 2026-07-17
feat: derive app-load skeleton gates from the request queue

BartekObudzinski:
### Explanation of Change

Several full-page and skeleton loaders decide whether the app is still doing its initial load by reading the stored `IS_LOADING_APP` Onyx flag. That flag is a shadow of the request queue: it is set optimistically when the initial `OpenApp` request is enqueued and cleared when the request leaves the queue, but a reload or dropped response in between can strand it, leaving the loader up forever with no request running.

This change moves those consumers onto the queue-derived `useIsAppLoadPending()` hook, which reads directly whether an `OpenApp` request is currently in the SequentialQueue. The flag was already set and cleared atomically with that request, so healthy-path behavior is identical — the only difference is that a stranded flag can no longer pin a loader.

Scope is the consumers that read the flag as a plain loading gate. Consumers with different semantics (a `true`/`false` default that treats &#34;unknown&#34; as loading, completion checks, latches, or non-loader logic) are intentionally left on the flag. The flag, its writers, and the recovery watchdog are untouched — their removal is a separate change. Telemetry `reasonAttributes` keep their existing `isLoadingApp` key name (now sourced from the hook) for continuity in Sentry.

Migrated: WorkspaceSelector, DomainsListPage, DynamicNewReportWorkspaceSelectionPage, DynamicReportChangeWorkspacePage, SetDefaultWorkspacePage, ReportActions, MoneyRequestReportView, useReportActionsListModel, ProfilePage, WalletPage.

### Fixed Issues
$ https://github.com/Expensify/App/issues/96487
PROPOSAL:

### Tests

1. Cold-start the app. Verify the initial full-page / skeleton loaders show, then the content renders.
2. Open each migrated screen (Profile, Wallet, Domains list, a workspace selector) during and after the initial load. Verify no endless loader.
3. Reload the app mid initial-load, then reopen a migrated screen. Verify the loader resolves once `OpenApp` completes (no stuck loader).

### Offline tests

1. Go offline, then open the migrated screens. Verify they do not sit on an endless skeleton and offline handling matches today.

### QA Steps

Same as tests.

### PR Author Checklist

- [x] I ran the tests on **all platforms** &amp; verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] I verified there are no console errors (if there&#39;s a console error not related to the PR, report it or open an issue for it to be fixed)
    - [x] I verified that any new or modified comments were clear, correct English, and explained &#34;why&#34; the code was doing something instead of only explaining &#34;what&#34; the code was doing.
    - [x] A similar style doesn&#39;t already exist
    - [x] The style can&#39;t be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/styles/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG)`)

### Screenshots/Videos

Android: Native

Android: mWeb Chrome

iOS: Native

iOS: mWeb Safari

MacOS: Chrome / Safari

### DIFF
--- a/src/hooks/useInFlightRequests.ts
+++ b/src/hooks/useInFlightRequests.ts
@@ -6,6 +6,8 @@ import type {AnyRequest} from '@src/types/onyx';
 
 import type {OnyxEntry} from 'react-native-onyx';
 
+import {useEffect} from 'react';
+
 import useNetwork from './useNetwork';
 import useOnyx from './useOnyx';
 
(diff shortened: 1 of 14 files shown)


src/hooks/useInFlightRequests.ts
```diff
@@ -99,9 +99,15 @@ function useIsPendingInternal(group: PendingRequestGroup, scopeKey?: string | nu
...
+/** Whether an OpenApp request or its deferred Onyx updates are pending. */
 function useIsAppLoadPending(): boolean {
-    return useIsPendingInternal('appLoad');
+    const hasPendingOpenApp = useIsPendingInternal('appLoad');
+    const [isLoadingApp] = useOnyx(ONYXKEYS.IS_LOADING_APP);
+    const [hasLoadedApp] = useOnyx(ONYXKEYS.HAS_LOADED_APP);
+
+    // HAS_LOADED_APP is committed only after deferred OpenApp updates flush. Combining both existing
+    // markers covers that gap without trusting an IS_LOADING_APP value stranded after a completed load.
+    return hasPendingOpenApp || (isLoadingApp === true && hasLoadedApp !== true);
```
DylanDylann: @BartekObudzinski I understand this gap. My question is whether we actually need to handle it at all. Could we drop the IS_LOADING_APP/HAS_LOADED_APP fallback and rely on hasPendingOpenApp only? Do you see any problem with that?

DylanDylann: Honestly, I'd like to remove isLoadingApp completely

BartekObudzinski: We can't drop it fully yet, but the reliance is minimal and self limiting.

hasPendingOpenApp goes false the moment OpenApp leaves the queue, which is before its finallyData flushes (the queue holds those writes until it drains). In that window the app data and HAS_LOADED_APP are not committed yet, so a pure hasPendingOpenApp gate would flash cleared or stale data on the migrated screens. That is the P1 Codex flagged.

The `isLoadingApp === true && hasLoadedApp !== true` term exists only to cover that first load flush window, and it is self healing for the stranded flag case: HAS_LOADED_APP is a permanent latch set only after the flush, so once the app has loaded it stays true and the term is false regardless of a stranded IS_LOADING_APP. A later reload runs ReconnectApp, not OpenApp, so nothing is left pending to strand.

Removing isLoadingApp completely is exactly the goal of the dedicated cleanup PR that deletes the flag and its writers. It can't come out here because these screens still need the flush window cover and other consumers still read it. If you would rather not lean on the markers, we can pair the removal with a different flush signal in that PR.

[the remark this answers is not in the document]
DylanDylann: I think that's @BartekObudzinski's intention now as explained above

DylanDylann (approved):

src/hooks/useInFlightRequests.ts
BartekObudzinski: Correction to my note above: the marker was not fully self healing. An account switch preserves HAS_LOADED_APP true while firing a real OpenApp, and its data plus the IS_LOADING_APP clear are deferred to the queue drain, so gating on HAS_LOADED_APP dropped the skeleton mid flush (the P2 Codex flagged). I replaced the marker with a process session latch keyed on an observed OpenApp instead, which covers that window and still ignores a stranded flag on reload, since a reload runs ReconnectApp not OpenApp so the latch is never set.

DylanDylann (approved):

mountiny (approved):
Are there no docs that need to be updated?

PR #96473, opened 2026-07-20
[HOLD] Resolve optimistic agent edit route after creation

nabi-ebrahimi:
### Explanation of Change

Opening a newly created agent before the server finishes creating it could land on a "Hmm... it's not here" page, because the page was still using the temporary ID assigned before saving instead of the real one.

This change resolves that temporary ID to the real one once the server responds, so the page keeps showing the agent instead of erroring out — across all the agent screens (edit, name, prompt, avatar), not just the main one. The temporary-to-real mapping is now saved to disk (not just kept in memory) so this also works after a refresh, and old entries are automatically cleaned up over time so it doesn't grow unbounded.

### Fixed Issues

$ #93272
PROPOSAL: [#93272 comment](https://github.com/Expensify/App/issues/93272#issuecomment-5006077398)

### Tests

**Precondition:**
- The internet connection speed must be set to 3G.

1. Go to `Account > Agents`.
2. Click `New agent`.
3. Create a new agent and save it.
4. As soon as the new agent appears in the list, open the agent edit page before creation fully finishes.
5. Wait for `CreateAgent` to finish.
6. Verify the edit page continues showing the created agent.
8. Verify the “Hmm... it’s not here” page does not appear.

### QA Steps

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/714679e7-7103-48c0-88cb-b81c75ae9e2f

</details>

<details>

>

https://github.com/user-attachments/assets/fae1112f-abc9-461b-9ee6-09e46f70d998

</details>

<details>

>

https://github.com/user-attachments/assets/adcef877-f9cb-48d7-9c16-b77d883db8af

</details>

<details>

>

https://github.com/user-attachments/assets/b7077c38-5016-4f7b-86b1-3ab4d07069ac

</details>

<details>

>

https://github.com/user-attachments/assets/9c7f1c14-51db-45ee-88dc-cb6d5e76373f

</details>


trjExpensify (changes requested):
I can't reproduce this bug on staging. 

https://github.com/user-attachments/assets/221aac99-069a-477a-83a3-90a6801f40cd

trjExpensify (dismissed):
Thanks for updating the tests. 👍

src/components/Tables/AgentsTable/AgentsTableRow.tsx
```diff
@@ -79,12 +80,18 @@ export default function AgentsTableRow({item, rowIndex, shouldUseNarrowTableLayo
                             <TextWithTooltip
                                 shouldShowTooltip
                                 text={item.displayName}
-                                style={[styles.optionDisplayName, styles.pre, isPendingDeletion && styles.offlineFeedbackDeleted]}
+                                style={[styles.optionDisplayName, styles.pre, isPendingDeletion && styles.offlineFeedbackDeleted, isPendingAdd && styles.offlineFeedbackPending]}
```
marufsharifi: **Double-dimmed text when offline / on add-error.**

`Table.Row` already wraps children in `OfflineWithFeedback` with `pendingAction: item.pendingAction`. In `OfflineWithFeedback.tsx:107`, `needsOpacity` becomes `true` whenever we're **offline with a pending action** *or* the **add has an error**, applying `styles.offlineFeedbackPending` (opacity `0.5`) to the whole row. Adding `offlineFeedbackPending` here stacks a second `0.5` on the text — RN opacity nests multiplicatively, so the name renders at ~`0.25` while the avatar/arrow sit at `0.5`.

It looks fine in the videos only because the repro is **online** (3G), where `isOffline === false` → no wrapper dimming. The regression appears offline or on a failed creation.

Recommended fix — let the wrapper own the pending look (dims the whole row once, online *and* offline), and revert this style. Update the `offlineWithFeedback` prop on `Table.Row`:

```tsx
offlineWithFeedback={{
    errors: item.errors,
    pendingAction: item.pendingAction,
    onClose: item.dismissError,
    shouldHideOnDelete: false,
    shouldForceOpacity: isPendingAdd, // uniform dim for pending-add, online or offline
}}
```

Then revert this line:

```suggestion
                                style={[styles.optionDisplayName, styles.pre, isPendingDeletion && styles.offlineFeedbackDeleted]}
```

src/components/Tables/AgentsTable/AgentsTableRow.tsx
```diff
@@ -79,12 +80,18 @@ export default function AgentsTableRow({item, rowIndex, shouldUseNarrowTableLayo
...
                                 shouldShowTooltip
                                 text={item.login}
-                                style={[styles.textLabelSupporting, styles.lh16, styles.pre, isPendingDeletion && styles.offlineFeedbackDeleted]}
+                                style={[
+                                    styles.textLabelSupporting,
+                                    styles.lh16,
+                                    styles.pre,
+                                    isPendingDeletion && styles.offlineFeedbackDeleted,
+                                    isPendingAdd && styles.offlineFeedbackPending,
+                                ]}
```
marufsharifi: **Same double-opacity issue as the `displayName` above, and it also exposes an inconsistency.** With the manual `offlineFeedbackPending` only on the text, during online pending-add the **text** dims and the **arrow** dims (it keys off `item.disabled`), but `ReportActionAvatars` stays at full opacity — the row reads as "half pending." The `shouldForceOpacity: isPendingAdd` fix in the comment above dims the whole row (avatar included) uniformly and matches every other offline-feedback row in the app. Revert this to a single line:

```suggestion
                                style={[styles.textLabelSupporting, styles.lh16, styles.pre, isPendingDeletion && styles.offlineFeedbackDeleted]}
```

src/components/Tables/AgentsTable/AgentsTableRow.tsx
```diff
@@ -115,7 +122,7 @@ export default function AgentsTableRow({item, rowIndex, shouldUseNarrowTableLayo
                                 <Button
                                     size={CONST.BUTTON_SIZE.SMALL}
                                     onPress={item.action}
-                                    isDisabled={isPendingDeletion}
+                                    isDisabled={isPendingAddOrDelete || !item.action}
```
marufsharifi: **Redundant `!item.action` term.** `item.action` is `undefined` **only** when `isPendingAdd`, and `isPendingAdd ⊂ isPendingAddOrDelete`, so `!item.action` can never be the deciding condition — `isPendingAddOrDelete` alone is equivalent. Not a bug, but it reads as if it guards a case that doesn't exist. Either simplify:

```suggestion
                                    isDisabled={isPendingAddOrDelete}
```

…or, if you want to keep it defensive against future callers, add a one-line comment explaining why (`// action is undefined for pending-add rows`).

src/pages/settings/Agents/AgentsPage.tsx
```diff
@@ -108,8 +110,8 @@ function AgentsPage() {
                 login: details.login ?? '',
                 pendingAction,
                 errors: Object.keys(rowErrors).length > 0 ? rowErrors : undefined,
-                disabled: isPendingDeletion,
-                action: () => Navigation.navigate(ROUTES.SETTINGS_AGENTS_EDIT.getRoute(accountID)),
+                disabled: isPendingAddOrDelete,
+                action: isPendingAdd ? undefined : () => Navigation.navigate(ROUTES.SETTINGS_AGENTS_EDIT.getRoute(accountID)),
```
marufsharifi: Core fix looks correct: setting `action` to `undefined` **and** `disabled: isPendingAddOrDelete` blocks both entry points (row `onPress` + Edit button) for the optimistic row, so navigation with the temporary `accountID` can't happen. ✅

This state transition already regressed once and is very unit-testable — please add coverage (the PR checklist ticks "I added unit tests" but the diff has none):
- given a row with `pendingAction: ADD` → assert `action === undefined` and `disabled === true`;
- given a settled row → assert `action` navigates to `SETTINGS_AGENTS_EDIT` and `disabled === false`;
- (optional) `pendingAction: DELETE` → `disabled === true`, behavior unchanged.

marufsharifi (changes requested):

src/pages/settings/Agents/EditAgentPage.tsx
```diff
@@ -37,7 +37,9 @@ function EditAgentPage({route}: EditAgentPageProps) {
     const {translate} = useLocalize();
     const styles = useThemeStyles();
     const icons = useMemoizedLazyExpensifyIcons(['Trashcan', 'ChatBubble', 'Users']);
-    const accountID = route.params.accountID;
+    const routeAccountID = route.params.accountID;
+    const [optimisticAgentAccountIDMapping] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING);
+    const accountID = optimisticAgentAccountIDMapping?.[routeAccountID] ?? routeAccountID;
```
marufsharifi: 🔴 **This resolve step is a no-op — nothing ever writes to `OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING`.**

```tsx
const routeAccountID = route.params.accountID;
const [optimisticAgentAccountIDMapping] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING);
const accountID = optimisticAgentAccountIDMapping?.[routeAccountID] ?? routeAccountID;
```

Across the whole diff, this key is **defined** (`ONYXKEYS.ts`), **read** (here), **allowlisted** (`ExportOnyxState/common.ts`) and **mocked** (test) — but never written. The only place that could populate it is `createAgent` in `src/libs/actions/Agent.ts`, which isn't part of this PR; its `successData` only nulls the optimistic entries.

Because the map is always empty, `optimisticAgentAccountIDMapping?.[routeAccountID]` is always `undefined` and `accountID` always falls back to `routeAccountID`. Behavior is identical to `main`, so the "Hmm… it's not here" bug from #93272 is **not fixed**.

Note there's also a design gap even with a writer: the real accountID is server-assigned by `CREATE_AGENT`, but `successData` is static optimistic data and doesn't have access to it — populating this map requires the backend response to return the real accountID and NewDot to merge it here. Please add (and test) the write path before this can work.

src/pages/settings/Agents/EditAgentPage.tsx
```diff
@@ -37,7 +37,9 @@ function EditAgentPage({route}: EditAgentPageProps) {
     const {translate} = useLocalize();
     const styles = useThemeStyles();
     const icons = useMemoizedLazyExpensifyIcons(['Trashcan', 'ChatBubble', 'Users']);
-    const accountID = route.params.accountID;
+    const routeAccountID = route.params.accountID;
+    const [optimisticAgentAccountIDMapping] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING);
```
marufsharifi: 🟡 **Use a scoped selector.** Subscribing to the whole mapping object re-renders every mounted `EditAgentPage` whenever *any* agent is created. Scope it to this page's entry:

```tsx
const [resolvedAccountID] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING, {
    selector: (m) => m?.[route.params.accountID] ?? route.params.accountID,
});
```

This is the established pattern for keyed lookups and avoids unnecessary re-renders.

src/pages/settings/Agents/EditAgentPage.tsx
marufsharifi: 🟢 **NAB type note (no change strictly required).** `route.params.accountID` is typed `number` and the map is `Record<string, number>`. Indexing a string-keyed record with a `number` compiles fine and the `?? routeAccountID` fallback keeps `accountID` a `number`, so `${...}${accountID}` and `accountID <= 0` downstream still behave. No type cast introduced — 👍. Just flagging that on browser refresh/deeplink the param arrives as a string at runtime (pre-existing), and the lookup works either way.

src/ONYXKEYS.ts
```diff
@@ -83,6 +83,9 @@ const ONYXKEYS = {
     /** Contains all the personalDetails the user has access to, keyed by accountID */
     PERSONAL_DETAILS_LIST: 'personalDetailsList',
 
+    /** Maps optimistic agent accountIDs to the real accountIDs returned by CreateAgent */
+    OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING: 'optimisticAgentAccountIDMapping',
```
marufsharifi: 🟡 **This map grows unbounded and is never cleaned up.** Optimistic IDs come from `generateReportID()`, so they're effectively random and never reused — every agent creation would add a permanently-stale `optimistic → real` entry. Combined with it being added to `safeOnyxKeys` (persisted + exported), this only ever grows.

Please add cleanup once resolution completes / on `CREATE_AGENT` success, e.g.:

```ts
Onyx.merge(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING, {[optimisticAccountID]: null});
```

Silent ever-growing persisted state is a classic slow-burn regression on high-traffic accounts.

src/libs/ExportOnyxState/common.ts
```diff
@@ -261,6 +261,7 @@ const safeOnyxKeys = new Set<string>([
     ONYXKEYS.ONBOARDING_POLICY_ID,
     ONYXKEYS.ONBOARDING_PURPOSE_SELECTED,
     ONYXKEYS.ONBOARDING_USER_REPORTED_INTEGRATION,
+    ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING,
```
marufsharifi: 🟡 **Confirm this key should be exported/persisted at all.** It's purely a transient client-side reconciliation map. If it's kept here, make sure stale entries are pruned (see the cleanup comment on `ONYXKEYS.ts`) — otherwise every exported Onyx state snapshot carries an ever-growing list of dead optimistic→real mappings.

marufsharifi: isn't it `RAM_ONLY` key? If so, could we prepend `RAM_ONLY` to the name?

tests/unit/pages/settings/EditAgentPageTest.tsx
```diff
@@ -277,4 +278,31 @@ describe('EditAgentPage', () => {
 
         expect(JSON.stringify(toJSON())).not.toContain('notFound.notHere');
     });
+
+    it('renders the real agent data when the route accountID is optimistic', () => {
```
marufsharifi: 🟡 **This test passes while the feature is broken.** It fully mocks `useOnyx` and hand-feeds `{12345: 67890}` into the mapping — a value production never produces (see the "never written" comment on `EditAgentPage.tsx`). So it green-lights a dead code path and doesn't exercise the actual bug from #93272.

Please add an integration-style test covering the real flow: create agent (pending) → navigate to edit → `CREATE_AGENT` success reconciliation → page resolves to the real agent instead of NotFound, and the optimistic map entry is cleaned up.

marufsharifi (changes requested):
🔴 **The title/description don't match the implementation.**

- Title: *"Disable edit navigation for pending agents"*
- Body: *"…temporarily non-interactive… Verify the `Edit` button is disabled while creation is still pending."*

But the diff does neither — it does accountID remapping, and `AgentsPage.tsx:112` still navigates unconditionally:

```tsx
action: () => Navigation.navigate(ROUTES.SETTINGS_AGENTS_EDIT.getRoute(accountID)),
```

Test steps 4–6 (row "temporarily unavailable", Edit button disabled while pending) can't pass against this diff because that gating code doesn't exist. Please either (a) rewrite the description to document the ID-remapping approach, or (b) if you pivoted to the "disable while pending" approach, add that logic to `AgentsPage`/`EditAgentPage`. This PR currently looks like it's missing the files that implement the described behavior.

src/pages/settings/Agents/EditAgentPage.tsx
nabi-ebrahimi: I confirmed from the actual `CreateAgent` response that this mapping is backend-provided via `onyxData`, so it does not need a frontend writer in `createAgent()`.

Example response:

```json
{
  "key": "optimisticAgentAccountIDMapping",
  "onyxMethod": "merge",
  "value": {
    "8162506105255733": 22658809
  }
}
```
The same response also includes `sharedNVP_agentPrompt_22658809` and `personalDetailsList.22658809`, so the edit page can resolve the optimistic route `accountID` to the real server [accountID](url).

tests/unit/pages/settings/EditAgentPageTest.tsx
```diff
@@ -144,18 +147,26 @@ jest.mock('@pages/ErrorPage/NotFoundPage', () => {
...
 type EditAgentPageRoute = PlatformStackScreenProps<SettingsNavigatorParamList, typeof SCREENS.SETTINGS.AGENTS.EDIT>['route'];
 type EditAgentPageNavigation = PlatformStackScreenProps<SettingsNavigatorParamList, typeof SCREENS.SETTINGS.AGENTS.EDIT>['navigation'];
 
-const mockRoute = {params: {accountID: TEST_ACCOUNT_ID}} as EditAgentPageRoute;
+const mockRoute = {key: 'Settings_Agents_Edit-test', params: {accountID: TEST_ACCOUNT_ID}} as EditAgentPageRoute;
 const mockNavigation = {} as EditAgentPageNavigation;
+const getSelectedAccountID = (options: Parameters<typeof useOnyx>[1], mapping: Record<string, number> | undefined, fallbackAccountID: number) => {
+    options?.selector?.(mapping);
+    return mapping?.[fallbackAccountID] ?? fallbackAccountID;
+};
```
marufsharifi: **The test re-implements the selector instead of exercising the real one.**

Because `useOnyx` is fully mocked, the production selector
`(mapping) => mapping?.[routeAccountID] ?? routeAccountID` never runs — this helper duplicates its logic and discards the real return value:

```ts
const getSelectedAccountID = (options, mapping, fallbackAccountID) => {
    options?.selector?.(mapping);                              // called, result ignored
    return mapping?.[fallbackAccountID] ?? fallbackAccountID;  // logic duplicated here
};
```

So the assertions would stay green even if the real selector used the wrong key or fallback. Please assert on the selector's actual output, e.g.:

```ts
expect(options.selector?.({[TEST_ACCOUNT_ID]: TEST_REAL_ACCOUNT_ID})).toBe(TEST_REAL_ACCOUNT_ID);
expect(options.selector?.(undefined)).toBe(TEST_ACCOUNT_ID);
```

so the function under test is genuinely covered.

src/pages/settings/Agents/EditAgentPage.tsx
```diff
@@ -39,7 +40,21 @@ function EditAgentPage({route}: EditAgentPageProps) {
...
+        selector: (mapping: Record<string, number> | null | undefined) => mapping?.[routeAccountID] ?? routeAccountID,
+    });
+    const accountID = resolvedAccountID ?? routeAccountID;
+
+    useEffect(() => {
+        if (!isFocused || accountID === routeAccountID) {
+            return;
+        }
+
+        Navigation.setParams({accountID}, route.key);
```
marufsharifi: This "resolve the route ID → `Navigation.setParams` to swap it in place" shape already exists in [`WorkspaceEditTaxPage.tsx#L60-L66`](https://github.com/Expensify/App/blob/main/src/pages/workspace/taxes/WorkspaceEditTaxPage.tsx#L60-L66):

```tsx
useEffect(() => {
    if (currentTaxID === taxID || !currentTaxID) {
        return;
    }
    Navigation.setParams({taxID: currentTaxID});
}, [taxID, currentTaxID]);
```

Same guarded, effect-driven route-param swap — just a different resolution source (`getCurrentTaxID` there vs. the RAM-only mapping selector here) and this one adds an `isFocused` guard + `route.key`.

There's no shared hook for this today (nothing in `src/hooks/` wraps `Navigation.setParams` for route-param resolution), so I'm **not** blocking on extraction for the 2nd occurrence. Flagging so it's on record: if a third case appears, this should become a shared hook (e.g. `useResolvedRouteParam(routeKey, paramName, resolvedValue, {onlyWhenFocused})`) and both pages migrate to it.

src/pages/settings/Agents/EditAgentPage.tsx
```diff
@@ -39,7 +40,21 @@ function EditAgentPage({route}: EditAgentPageProps) {
     const {translate} = useLocalize();
     const styles = useThemeStyles();
     const icons = useMemoizedLazyExpensifyIcons(['Trashcan', 'ChatBubble', 'Users']);
-    const accountID = route.params.accountID;
+    const isFocused = useIsFocused();
+    const routeAccountID = route.params.accountID;
+    const [resolvedAccountID] = useOnyx(ONYXKEYS.RAM_ONLY_OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING, {
```
marufsharifi, line 48: This introduces a **second, different pattern** for reconciling an optimistic entity ID with its real one. The app already has a server-driven mechanism for this in the report flow — [`HandleUnusedOptimisticID.ts#L94`](https://github.com/Expensify/App/blob/main/src/libs/Middleware/HandleUnusedOptimisticID.ts#L94) + [`replaceOptimisticReportWithActualReport.ts`](https://github.com/Expensify/App/blob/main/src/libs/actions/replaceOptimisticReportWithActualReport.ts) — which uses `preexistingReportID` to move the optimistic Onyx data onto the real key so consumers "just work" without a read-site lookup.

Here we instead keep a RAM-only ID→ID map and resolve it at the component read-site. That's exactly the kind of divergence the checklist item *"If a new code pattern is added I verified it was agreed to be used by multiple Expensify engineers"* is meant to catch.

I can't point at a drop-in util to reuse (the report mechanism is report-collection-specific — `parentReportID`, report merging), so this isn't a hard blocker. Two questions for the reviewer:

1. Was the read-site-mapping approach chosen deliberately over reconciling the agent's Onyx data the way reports do?
2. If the backend already returns the mapping, could it instead move the optimistic `SHARED_NVP_AGENT_PROMPT` / personal-details data onto the real `accountID`, eliminating the per-read lookup entirely?

src/pages/settings/Agents/EditAgentPage.tsx
```diff
@@ -39,7 +40,21 @@ function EditAgentPage({route}: EditAgentPageProps) {
...
+        selector: (mapping: Record<string, number> | null | undefined) => mapping?.[routeAccountID] ?? routeAccountID,
+    });
+    const accountID = resolvedAccountID ?? routeAccountID;
```
marufsharifi, line 51: The selector already ends in `?? routeAccountID`, so `resolvedAccountID` is never nullish once loaded — this second `?? routeAccountID` repeats a fallback the selector already guarantees, which makes a reader hunt for the case it's protecting against.

If it's only guarding the pre-load `undefined`, a short comment would make that explicit; otherwise pick a single place to default (either drop this `??` and use `resolvedAccountID` directly, or drop the `?? routeAccountID` inside the selector).

src/pages/settings/Agents/EditAgentPage.tsx
```diff
@@ -39,7 +40,21 @@ function EditAgentPage({route}: EditAgentPageProps) {
     const {translate} = useLocalize();
     const styles = useThemeStyles();
     const icons = useMemoizedLazyExpensifyIcons(['Trashcan', 'ChatBubble', 'Users']);
-    const accountID = route.params.accountID;
+    const isFocused = useIsFocused();
+    const routeAccountID = route.params.accountID;
+    const [resolvedAccountID] = useOnyx(ONYXKEYS.RAM_ONLY_OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING, {
+        selector: (mapping: Record<string, number> | null | undefined) => mapping?.[routeAccountID] ?? routeAccountID,
```
marufsharifi: The `Record<string, number>` mapping shape is inlined here and also repeated in `ONYXKEYS.ts` and the test file. Extracting a named type would keep the three in sync and shorten the selector:

```ts
type OptimisticAgentAccountIDMapping = Record<string, number>;
```

Then reuse it in the Onyx value mapping and this selector's param.

marufsharifi (changes requested):

src/pages/settings/Agents/EditAgentPage.tsx
nabi-ebrahimi: I agree this has the same general shape as `WorkspaceEditTaxPage`, but since this is only the second usage and there is no shared hook today, I’ll keep it inline for now

src/pages/settings/Agents/EditAgentPage.tsx
nabi-ebrahimi: I agree the report flow has a dedicated reconciliation mechanism, but that path is report-specific and tied to report collection behavior. For agents, the backend already returns the real agent data under the real `accountID` in the `CreateAgent` response, along with the optimistic-to-real mapping.

So this PR uses the mapping only to bridge an already-mounted optimistic edit route to the real agent keys. It avoids `moving/merging` optimistic agent data client-side and keeps the change scoped to the page that can be opened with the temporary `accountID.`

src/pages/settings/Agents/EditAgentPage.tsx
nabi-ebrahimi: Addressed: The selector now only returns the mapped real `accountID`:

src/ONYXKEYS.ts
```diff
@@ -87,6 +87,9 @@ const ONYXKEYS = {
     /** Contains all the personalDetails the user has access to, keyed by accountID */
     PERSONAL_DETAILS_LIST: 'personalDetailsList',
 
+    /** Maps optimistic agent accountIDs to the real accountIDs returned by CreateAgent */
+    RAM_ONLY_OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING: 'optimisticAgentAccountIDMapping',
```
marufsharifi, line 91: @nabi-ebrahimi, if the data of this key comes from BE, why the name is RAM_ONLY? could you please check this.

marufsharifi (changes requested):

src/ONYXKEYS.ts
nabi-ebrahimi: Thanks for the review! 

`RAM_ONLY` refers to **persistence**, not **provenance** — it means the value is never written to storage and is gone on reload, regardless of who writes it. It's the naming convention for keys registered in `ramOnlyKeys` in `src/setup/index.ts`.

We already have backend-populated RAM-only keys: `RAM_ONLY_PLAID_LINK_TOKEN` is the same shape (value comes from the server's `onyxData`, client only clears it), as is `RAM_ONLY_WALLET_ONFIDO`.

It's also the right fit here — the optimistic `accountID` can't be referenced after a reload, so RAM-only keeps the map bounded.

marufsharifi (approved):
LGTM

NikkiWines (changes requested):
@nabi-ebrahimi this seems like a different mechanism than your [proposal](https://github.com/Expensify/App/issues/93272#issuecomment-5006077398) - as caught by the bot as well [here](https://github.com/Expensify/App/pull/96473#pullrequestreview-4761856703) 

Can you elaborate on why we're not going with the route initally proposed?

nabi-ebrahimi:
@NikkiWines @marufsharifi Following up after digging into this.

### What I changed based on [this feedback](https://github.com/Expensify/App/pull/96473#issuecomment-5283357288)

- **The mapping is no longer RAM-only.** [`OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING`](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/ONYXKEYS.ts#L91) is now a **persisted** key, since the backend only sends it once — a `CreateAgent` replayed after a reload comes back `400 Unique Constraints Violation` with empty `onyxData`. That closes the "works only for the current session" gap (not-found on refresh / browser-back).
- **Resolution now lives in one shared place** — [`useResolvedAgentAccountID`](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/hooks/useResolvedAgentAccountID.ts) — used by **all four** agent screens: [EditAgentPage](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/pages/settings/Agents/EditAgentPage.tsx#L45), [EditNamePage](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/pages/settings/Agents/Fields/EditNamePage.tsx#L33), [EditPromptPage](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/pages/settings/Agents/Fields/EditPromptPage.tsx#L37) and [EditAgentAvatarPage](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/pages/settings/Agents/Fields/EditAgentAvatarPage.tsx#L233) — not just `EditAgentPage.tsx`. That directly addresses the gap you pointed out with `EditNamePage` / `EditPromptPage`.

### Why I didn't do a literal data-layer relocation like the report pattern

I looked closely at [`replaceOptimisticReportWithActualReport`](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/libs/actions/replaceOptimisticReportWithActualReport.ts) as a precedent, but it solves a different problem: it only runs when the server returns `preexistingReportID`, meaning the optimistic report the client created collided with a report that **already existed** on the server. In that case the client tears down its own optimistic report and moves its local state onto the pre-existing report's key.

**Agents are different:** by the time `CreateAgent` resolves, the backend has already written the real `personalDetails`/prompt data directly under the real accountID, and the `CreateAgent` [`successData` already clears the optimistic entries](https://github.com/Expensify/App/blob/75fac6e19623d89886df65950dbafb9cb39c7510/src/libs/actions/Agent.ts#L115). There's no leftover entity data at the optimistic key to relocate — what's actually stale is just the optimistic ID sitting in a route param.

So a single shared resolver that every consumer goes through gives the same **"one point of truth"** principle you were asking for, without forcing a data move that doesn't apply here.

Would appreciate another look when you get a chance, thanks!

nabi-ebrahimi:
@NikkiWines, could you please take a look at this [comment](https://github.com/Expensify/App/pull/96473#issuecomment-5313597353) when you get a chance. thanks.

NikkiWines:
Sorry for the delay @nabi-ebrahimi. This looks like a good approach overall. Just a couple of things worth noting 

- This approach means every created agent adds a permanent entry with no cleanup. Can we add a time-based cleanup (e.g. prune entries older than N days on app boot) so that `OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING` doesn't grow forever?
- We need to adjust and add some unit tests to cover the new approach (I understand you might've been holding off on those until getting sign off on this 👍)

nabi-ebrahimi:
@marufsharifi, I’ve addressed the latest feedback. Could you take another look?

NikkiWines:
Thanks @nabi-ebrahimi I'll take a look today. @marufsharifi are you able to re-review this as well please, since it's changed significantly since your last approval 🙇

marufsharifi:
I will do another review here. thanks.

src/pages/settings/Agents/AgentsPage.tsx
```diff
@@ -65,14 +65,19 @@ function AgentsPage() {
 
     const [agentPrompts] = useOnyx(ONYXKEYS.COLLECTION.SHARED_NVP_AGENT_PROMPT);
     const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
+    // Read here (rather than inside openAgentsPage()) so the action stays a pure function of its params.
```
NikkiWines:
```suggestion

    // Read here (rather than inside openAgentsPage()) so the action stays a pure function of its params.
```

src/pages/settings/Agents/AgentsPage.tsx
```diff
@@ -65,14 +65,19 @@ function AgentsPage() {
...
     const personalDetailsList = usePersonalDetails();
     const canSelectMultiple = shouldUseNarrowLayout ? isMobileSelectionModeEnabled : true;
 
     useEffect(() => {
         if (!isCustomAgentEnabled) {
             return;
         }
-        openAgentsPage();
+        openAgentsPage(optimisticAgentAccountIDMappingCreatedAt);
+        // Only re-fire OPEN_AGENTS_PAGE when isCustomAgentEnabled changes, not on every mapping update
```
NikkiWines:
```suggestion

        // Only re-fire OPEN_AGENTS_PAGE when isCustomAgentEnabled changes, not on every mapping update
```

NikkiWines: (same for all other comments in this PR following this pattern)

src/pages/settings/Agents/AgentsPage.tsx
NikkiWines, line 1: I think we also need to update `openAgentsPage()` in `src/pages/ProfilePage.tsx` to pass in the mapping data.

src/libs/actions/Agent.ts
```diff
@@ -12,25 +12,48 @@ import type {AvatarSource} from '@libs/UserAvatarUtils';
...
-function openAgentsPage() {
-    const finallyData: Array<OnyxUpdate<typeof ONYXKEYS.ARE_AGENTS_LOADED>> = [
+const OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_MS = 30 * 24 * 60 * 60 * 1000;
+
+// Called from both createAgent() and openAgentsPage() (the two points that have this data on hand) instead of a
+// boot-time sweep, to avoid adding a new Onyx.connectWithoutView subscription.
+function getStaleOptimisticAccountIDMappingUpdates(existingOptimisticAccountIDMappingCreatedAt: OnyxEntry<OptimisticAgentAccountIDMappingCreatedAt>): AnyOnyxUpdate[] {
+    const now = Date.now();
+    const staleOptimisticAccountIDs = Object.entries(existingOptimisticAccountIDMappingCreatedAt ?? {})
+        .filter(([, createdAt]) => now - createdAt > OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_MS)
```
NikkiWines: This only looks at the `createdAt` timestamps, but the mapping itself can arrive on another device or browser tab through the server response without that device ever writing a timestamp for it. 

I think that's something we should fix on the backend so we include the timestamp in the mapping data. But for now, we could backfill a timestamp locally the first time any device notices a mapping entry that's missing one.

src/pages/settings/Agents/AgentsPage.tsx
```diff
@@ -65,14 +65,19 @@ function AgentsPage() {
 
     const [agentPrompts] = useOnyx(ONYXKEYS.COLLECTION.SHARED_NVP_AGENT_PROMPT);
     const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
+    // Read here (rather than inside openAgentsPage()) so the action stays a pure function of its params.
+    const [optimisticAgentAccountIDMappingCreatedAt] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT);
```
NikkiWines: Do we need to subscribe to this, or can we just read this only when it's actually needed, instead of on every render?

src/ONYXKEYS.ts
```diff
@@ -87,6 +87,12 @@ const ONYXKEYS = {
     /** Contains all the personalDetails the user has access to, keyed by accountID */
     PERSONAL_DETAILS_LIST: 'personalDetailsList',
 
+    /** Maps an agent's optimistic accountID to the real one CreateAgent assigns, so an agent screen opened on the optimistic ID resolves to the real agent. Persisted so it survives a reload (the mapping is sent only once). */
+    OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING: 'optimisticAgentAccountIDMapping',
+
+    /** When each OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING entry was written locally, so stale entries can be pruned. */
+    OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT: 'optimisticAgentAccountIDMappingCreatedAt',
```
NikkiWines, line 97: maybe we should store these together so they can't get out of sync

NikkiWines (changes requested):
A couple of things!  Also i think the PR description needs to be updated. 

I'll be OOO next week, @NicolasBonet I see you in the blame for the backend code for this (as well as some front end reviews) Any chance you could take a look in my absence please🙏

src/pages/settings/Agents/AgentsPage.tsx
nabi-ebrahimi: Addressed, thanks

src/pages/settings/Agents/AgentsPage.tsx
nabi-ebrahimi: @NikkiWines, Good catch — you're right, this re-renders the page on every change to that key, not just when the effect actually needs it.

The reason: `useOnyx` doesn't have a "read once" mode, it always subscribes reactively. To read it only when needed (inside the effect, without re-rendering on every change), we'd need `Onyx.connectWithoutView` instead. Want me to switch it to that, or do you have any other suggestions? Thanks!

src/libs/actions/Agent.ts
nabi-ebrahimi: fixed. `useResolvedAgentAccountID` now backfills a `createdAt` the first time any device notices a mapping entry without one — covers the cross-device/tab case.

src/ONYXKEYS.ts
nabi-ebrahimi: Done — combined them into a derived value (`OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_ENTRIES`). Consumers read one synced entry instead of two parallel keys.

src/pages/settings/Agents/AgentsPage.tsx
marufsharifi: @NikkiWines, could you please share your thoughts here. thanks.

nabi-ebrahimi: >I'll be OOO next week, @NicolasBonet I see you in the blame for the backend code for this (as well as some front end reviews) Any chance you could take a look in my absence please🙏

@NicolasBonet, could you take a look here while NikkiWines is OOO?

src/hooks/useResolvedAgentAccountID.ts
```diff
@@ -0,0 +1,35 @@
...
+ * always-in-sync source), so an agent screen opened on the optimistic accountID (even after a reload) shows the
+ * real agent instead of "Hmm... it's not here".
+ *
+ * Also backfills a missing `createdAt` for entries this device notices without one — a mapping entry can arrive
+ * via sync from another device/tab that resolved it first, so this device never got the chance to stamp it
+ * itself. Without a timestamp an entry is invisible to createAgent()'s pruning and never expires.
+ *
+ * @returns `[resolvedAccountID, isMappingLoaded]` - a not-found screen should wait for `isMappingLoaded` to avoid a
+ *          brief not-found flash while the mapping loads. No-op (returns the input) when there's no mapping entry.
+ */
```
NicolasBonet, line 19: Explain only the why, not what it does as the code itself shows the backfilling.

src/libs/actions/Agent.ts
```diff
@@ -12,25 +12,48 @@ import type {AvatarSource} from '@libs/UserAvatarUtils';
...
 import type {AnyOnyxUpdate} from '@src/types/onyx/Request';
 
-import type {OnyxCollection, OnyxCollectionInputValue, OnyxUpdate} from 'react-native-onyx';
+import type {OnyxCollection, OnyxCollectionInputValue, OnyxEntry, OnyxUpdate} from 'react-native-onyx';
 
 import Onyx from 'react-native-onyx';
 
-function openAgentsPage() {
-    const finallyData: Array<OnyxUpdate<typeof ONYXKEYS.ARE_AGENTS_LOADED>> = [
+const OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_MS = 30 * 24 * 60 * 60 * 1000;
```
NicolasBonet: It it more insightful if you add something like 

```
const OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_IN_DAYS= 30
..... * OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_IN_DAYS
```

I also think this might be too much time?

src/pages/settings/Agents/AgentsPage.tsx
NicolasBonet: Let's do that @nabi-ebrahimi

NicolasBonet (changes requested):

src/hooks/useResolvedAgentAccountID.ts
nabi-ebrahimi: Addressed, thanks

src/libs/actions/Agent.ts
nabi-ebrahimi: Good point — what number of days would you like instead of 30?

NicolasBonet: Something like 7 or even less would be fine

src/libs/actions/Agent.ts
```diff
@@ -106,6 +137,7 @@ function createAgent(
             key: `${ONYXKEYS.COLLECTION.REPORT_METADATA}${optimisticReportID}`,
             value: {isOptimisticReport: true},
         },
+        ...getStaleOptimisticAccountIDMappingUpdates(),
```
tgolen: Hm, but if this is called in `createAgent()` then it is triggered from the UI, so I think it's wrong to use `connectWithoutView()`. Why is it triggered from `createAgent()`? Won't it trigger naturally by virtue of the merge in `successData` for `ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT`?

src/libs/actions/Agent.ts
```diff
@@ -12,25 +12,56 @@ import type {AvatarSource} from '@libs/UserAvatarUtils';
...
+import type {OnyxCollection, OnyxCollectionInputValue, OnyxEntry, OnyxUpdate} from 'react-native-onyx';
 
 import Onyx from 'react-native-onyx';
 
+const OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_IN_DAYS = 30;
+const OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_MS = OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_IN_DAYS * 24 * 60 * 60 * 1000;
+
+// Non-React file, so it reads its own data here instead of via useOnyx + a param (same pattern as HandleUnusedOptimisticID.ts).
+let optimisticAccountIDMappingCreatedAt: OnyxEntry<OptimisticAgentAccountIDMappingCreatedAt>;
+Onyx.connectWithoutView({
```
tgolen: Could you please move this reference and the associated function to it's own file? That will help isolate it and make it even more clear that the exception to use it is warrented.

src/libs/actions/Agent.ts
```diff
@@ -12,25 +12,56 @@ import type {AvatarSource} from '@libs/UserAvatarUtils';
...
+    key: ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT,
+    callback: (value) => {
+        optimisticAccountIDMappingCreatedAt = value;
+    },
+});
+
+function getStaleOptimisticAccountIDMappingUpdates(): AnyOnyxUpdate[] {
```
tgolen: Can you please add code comments to explain why this function is necessary?

tgolen (changes requested):

src/libs/actions/Agent.ts
nabi-ebrahimi: Addressed — removed the explicit call from `createAgent()`. Pruning now runs reactively from inside the `connectWithoutView` callback, so it fires automatically whenever `successData`'s stamp (or the backfill) writes to the key.

nabi-ebrahimi:
@tgolen, addressed your feedback. Could you take another look? Thanks

src/libs/actions/pruneOptimisticAgentAccountIDMapping.ts
```diff
@@ -0,0 +1,50 @@
...
+ * OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT is client-only bookkeeping for
+ * OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING (the backend never writes it), so nothing else prunes stale entries for us.
+ * This module watches the createdAt key directly and re-checks for staleness every time it changes — from
+ * createAgent()'s own stamp, from the cross-device backfill, or from this prune's own writes — so no caller has
+ * to trigger it explicitly. Onyx.connectWithoutView is used because this runs outside any component's render; no
+ * UI subscribes to it.
+ *
+ * The one caller that does still trigger it explicitly is openAgentsPage() (Agent.ts): an entry can go stale
+ * purely from time passing with no new write to fire this callback, and opening the agents list is the only
+ * realistic moment to re-check for that.
```
tgolen: This still feels like it's somewhat in the same boat as previously (where it is triggered by something in the UI). How come the entries can go stale from time passing? Maybe the comment can be improved to explain why that happens. 

If the time component is real for them going stale, wouldn't it be better to use an interval timer to re-run this check on a regular basis, rather than wait for a UI to trigger it?

tgolen (changes requested):
Thanks for the changes! I think we're getting there, but I'm still not fully understanding how the data can go stale.

src/libs/actions/pruneOptimisticAgentAccountIDMapping.ts
nabi-ebrahimi: Thanks. Improved the comment.

On the interval timer — checked the codebase for precedent first: [`checkForUpdates.ts`](https://github.com/Expensify/App/blob/cc94f8134ac831e741b417053c18448b6f14871c/src/libs/checkForUpdates.ts#L10-L13) is the only periodic-check pattern here, and it's deliberately a no-op on native ([`platformSetup/index.native.ts`](https://github.com/Expensify/App/blob/cc94f8134ac831e741b417053c18448b6f14871c/src/setup/platformSetup/index.native.ts#L1)), so a raw interval isn't something this app relies on for mobile.

Instead, added [`AppStateMonitor.addBecameActiveListener()`](https://github.com/Expensify/App/blob/cc94f8134ac831e741b417053c18448b6f14871c/src/libs/AppStateMonitor/index.ts#L13) alongside `openAgentsPage()` — it re-checks whenever the app resumes from background, which is the moment the most time is likely to have silently passed, using a pattern already proven on mobile elsewhere in this codebase.

src/hooks/useResolvedAgentAccountID.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Without a stamped `createdAt`, an entry is invisible to createAgent()'s pruning and never expires — this can
+ * happen when the mapping arrives via sync from another device/tab that resolved it first.
+ *
+ * @returns `[resolvedAccountID, isMappingLoaded]` - a not-found screen should wait for `isMappingLoaded` to avoid a
+ *          brief not-found flash while the mapping loads. No-op (returns the input) when there's no mapping entry.
+ */
+function useResolvedAgentAccountID(routeAccountID: number): [number, boolean] {
+    const [realAccountID, mappingMetadata] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING, {selector: (mapping) => mapping?.[routeAccountID]});
+    const [createdAt, createdAtMetadata] = useOnyx(ONYXKEYS.DERIVED.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_ENTRIES, {selector: (entries) => entries?.[routeAccountID]?.createdAt});
```
marufsharifi: **Read `createdAt` from the raw key and drop the derived value**

This hook is the only consumer of `OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_ENTRIES`, and resolution already reads the mapping from the **raw** key on line 25 — the derived value is used *only* to read `createdAt` here. That timestamp lives on the raw `OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT` key (the prune module already reads it that way), so we can drop the derived value entirely:

```suggestion
    const [createdAt, createdAtMetadata] = useOnyx(ONYXKEYS.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT, {selector: (createdAtByAccountID) => createdAtByAccountID?.[routeAccountID]});
```

That lets us delete `configs/optimisticAgentAccountIDMappingEntries.ts`, its registration in `ONYX_DERIVED_VALUES.ts`, the `DERIVED.OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_ENTRIES` key, the `OptimisticAgentAccountIDMappingEntry`/`…EntriesDerivedValue` types, and `optimisticAgentAccountIDMappingEntriesTest.ts`.

**Why:** the derived value's stated benefit ("consumers read a single always-in-sync source instead of two parallel keys") isn't actually realized — this consumer reads mapping-raw + createdAt-derived, i.e. two sources anyway — and it forces the cold-cache lag caveat in the JSDoc above (lines 14–16) plus the overwrite bug flagged below. Raw-key reads have neither problem.

src/hooks/useResolvedAgentAccountID.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    useEffect(() => {
+        if (realAccountID === undefined || createdAt !== undefined || createdAtMetadata.status !== 'loaded') {
```
marufsharifi, line 25: **Backfill can overwrite a valid `createdAt` and reset the 7-day prune clock**

`createdAt` here comes from the **persisted** derived value. After a reload, `useOnyx` can rehydrate a stale stored entry (`{realAccountID, createdAt: undefined}` from a session where the stamp hadn't landed) and report `status: 'loaded'` with `createdAt === undefined`, while the raw `…CREATED_AT` key already holds a real timestamp. This guard then passes and `backfillOptimisticAccountIDMappingCreatedAt` overwrites the real timestamp with `Date.now()`, silently resetting the entry's age and delaying its pruning by up to a full 7-day window.

Reading `createdAt` from the raw `OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT` key (see comment on line 26) removes this entirely, since the raw key is never a stale projection of itself. If the derived value is kept, please add a regression test: *"rehydrated-stale-derived must not overwrite an existing raw createdAt."*

src/libs/actions/pruneOptimisticAgentAccountIDMapping.ts
```diff
@@ -0,0 +1,58 @@
...
+    },
+});
+
+AppStateMonitor.addBecameActiveListener(() => pruneStaleOptimisticAccountIDMappingEntries());
+
+function pruneStaleOptimisticAccountIDMappingEntries() {
+    const now = Date.now();
+    const staleOptimisticAccountIDs = Object.entries(optimisticAccountIDMappingCreatedAt ?? {})
+        .filter(([, createdAt]) => now - createdAt > OPTIMISTIC_ACCOUNT_ID_MAPPING_MAX_AGE_MS)
+        .map(([staleOptimisticAccountID]) => staleOptimisticAccountID);
```
marufsharifi, line 63: **Mapping entries that never receive a `createdAt` are never pruned**

Pruning only iterates `optimisticAccountIDMappingCreatedAt`. But `createdAt` is stamped only (a) by `createAgent` `successData` on the *creating* device, or (b) by the backfill when *that specific agent's page* is opened. A mapping entry that syncs to a device where the user never opens that agent's page keeps no `createdAt`, so it's permanently invisible here and stays in `OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING` forever.

That contradicts the PR's "old entries are automatically cleaned up so it doesn't grow unbounded." Consider stamping `now` for any *mapping* key lacking a `createdAt` during the prune (iterate the mapping key set, not the createdAt set), so every entry becomes prunable within one window even if never viewed. NAB — each record is tiny — but it closes the unbounded-growth gap honestly.

src/libs/actions/OnyxDerived/configs/optimisticAgentAccountIDMappingEntries.ts
```diff
@@ -0,0 +1,28 @@
...
+    compute: ([mapping, createdAtByOptimisticAccountID]) => {
+        if (!mapping) {
+            return {};
+        }
+
+        const entries: OptimisticAgentAccountIDMappingEntriesDerivedValue = {};
+        for (const [optimisticAccountID, realAccountID] of Object.entries(mapping)) {
+            if (realAccountID === undefined) {
+                continue;
+            }
```
marufsharifi: **Unreachable defensive branch**

Given `OptimisticAgentAccountIDMapping = Record<string, number>` and that pruning deletes keys via `merge(…, null)` (deletion, not `undefined` assignment), `realAccountID` is never `undefined` here — this branch is dead. Either drop it, or add a one-line note on the runtime state that could actually make a value `undefined`.

marufsharifi (changes requested):

src/hooks/useResolvedAgentAccountID.ts
nabi-ebrahimi: Addressed — dropped the derived value entirely; createdAt is now read straight from the raw OPTIMISTIC_AGENT_ACCOUNT_ID_MAPPING_CREATED_AT key, so there's no stale rehydrated projection to overwrite it with.

src/libs/actions/pruneOptimisticAgentAccountIDMapping.ts
nabi-ebrahimi: Addressed — the prune now also watches the mapping key and stamps now for any entry missing a `createdAt`, so cross-device entries become prunable within one window even if the page is never opened locally.

nabi-ebrahimi:
@marufsharifi, I’ve addressed your latest feedback. Could you take another look?

tgolen:
I've had some concerns with the solution in this PR and was discussing some alternatives with @NicolasBonet. I'm going to place a HOLD on this PR until we able to resolve it.

NicolasBonet:
I believe we can close this one in favor of https://github.com/Expensify/App/pull/100068

PR #96490, opened 2026-07-20
[No QA] [Performance] Add `API.writeWhenReady` for deferred optimistic writes

JakubKorytko:
### Explanation of Change

Adds `API.writeWhenReady(command, params, onyxData, barrier?)`: it defers a write - including its optimistic Onyx updates - until a readiness signal fires (default: the navigation transition), then delegates to the normal `write()` pipeline, leaving optimistic/success/failure handling, retries, and ordering unchanged. Foundational primitive; no call sites are migrated yet. Includes a safety timeout, an app-background flush, and unit tests.

### Fixed Issues

$ https://github.com/Expensify/App/issues/96010
PROPOSAL: N/A

### Tests

This PR adds an API primitive with no call sites yet, so there is no user-facing flow to exercise; correctness is covered by unit tests.

### Offline tests

N/A - no user-facing flow invokes this primitive yet. The deferral and app-background flush behavior (which persists a pending write before the process can be killed) is covered by the unit tests.

### QA Steps

[No QA] - no code path invokes `writeWhenReady` yet, so there is nothing to verify on staging. Correctness is covered by the unit tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/API/index.ts
```diff
@@ -199,6 +201,236 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+            // Everything here runs defensively: a synchronous throw (from cancelBarrier(), write(), or
+            // prepareRequest) must settle the returned promise via reject rather than leave it pending
+            // forever - and, on the background path, must not escape into the flush loop and strand the
+            // other pending writes.
+            try {
+                clearTimeout(safetyTimeoutId);
+                pendingWriteWhenReadyFlushes.delete(flushOnBackground);
+                // Release the barrier if it is still pending (e.g. released early via timeout/background), so
+                // it doesn't leave a dangling registration.
+                cancelBarrier();
```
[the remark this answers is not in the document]
JakubKorytko: Good catch, agreed. Changed the implementation to only call cancelBarrier() on the early-release paths (safetyTimeout / appBackground). On the barrier / barrierRejected paths the barrier has already settled so there is nothing to cancel, and skipping it means a custom barrier whose cancel() is unsafe post-settle can no longer throw and drop the write. This also makes the docs accurate.

tests/unit/APIWriteWhenReadyTest.ts
```diff
@@ -0,0 +1,375 @@
...
+import {AppState} from 'react-native';
+import Onyx from 'react-native-onyx';
+
+jest.mock('@libs/Network/SequentialQueue', () => ({
+    push: jest.fn(() => Promise.resolve()),
+    waitForIdle: jest.fn(() => Promise.resolve()),
+    // Called by the network layer on init; stub so advancing fake timers doesn't hit a missing export.
+    flush: jest.fn(),
+}));
+jest.mock('@libs/Navigation/TransitionTracker');
```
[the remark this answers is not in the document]
JakubKorytko: This is a false positive. jest.mock('@libs/Navigation/TransitionTracker') auto-mocks the object-literal default export, so runAfterTransitions is a jest.fn() (returning undefined), not undefined. The suite exercises it via jest.mocked(TransitionTracker.runAfterTransitions) and all 17 tests pass, so no explicit factory is needed.

src/libs/API/index.ts
```diff
@@ -199,6 +201,236 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+ * `cancel` is optional - a plain promise-like works too.
+ */
+type CancelableBarrier = PromiseLike<unknown> & {cancel: () => void};
+
+/**
+ * A readiness signal for `writeWhenReady`: a promise-like that resolves once it is safe to apply the
+ * write's optimistic data (e.g. after a navigation transition finishes), or a function returning one
+ * (invoked when the write is queued). Either form may instead be a `CancelableBarrier`, whose `cancel`
+ * is invoked if the write fires before the barrier settles. A rejection is treated the same as
+ * resolving - the write executes anyway.
```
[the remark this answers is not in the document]
JakubKorytko: Fixed by the implementation change: cancelBarrier() is now only called when the write fires before the barrier settles (safetyTimeout / appBackground), so this doc line is now accurate as written.

src/libs/API/index.ts
```diff
@@ -199,6 +201,236 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+
+            // Everything here runs defensively: a synchronous throw (from cancelBarrier(), write(), or
+            // prepareRequest) must settle the returned promise via reject rather than leave it pending
+            // forever - and, on the background path, must not escape into the flush loop and strand the
+            // other pending writes.
+            try {
+                clearTimeout(safetyTimeoutId);
+                pendingWriteWhenReadyFlushes.delete(flushOnBackground);
+                // Release the barrier if it is still pending (e.g. released early via timeout/background), so
+                // it doesn't leave a dangling registration.
```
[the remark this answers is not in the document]
JakubKorytko: Fixed. Made the code match the comment: cancelBarrier() now only runs on the early-release paths (safetyTimeout / appBackground), and I updated the inline comment to spell that out.

src/libs/API/index.ts
```diff
@@ -199,6 +201,242 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+ * they mean the intended barrier did not release the write. A caller whose custom barrier can
+ * legitimately take longer than the default timeout should raise `safetyTimeoutMs` accordingly.
+ *
+ * @returns A promise that resolves with the underlying `write()` result once the write executes.
+ */
+function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
+    command: TCommand,
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData: OnyxData<TKey> = {},
+    barrier: WriteReadyBarrier = waitForNavigationTransition,
```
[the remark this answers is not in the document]
JakubKorytko: Checked this empirically and TypeScript does not accept it - the premise does not hold. Because OnyxData has no property a barrier would share (then/cancel/optimisticData/successData/etc), the weak-type check rejects every barrier shape in the onyxData slot: a function barrier and a plain Promise both error with TS2559 'has no properties in common with type OnyxData', a {then, cancel} object literal errors with TS2353 (excess property), and a value typed as WriteReadyBarrier errors with TS2345 'not assignable to OnyxData | undefined'. Verified all four with tsgo against the project config, so the barrier can never silently land in the onyxData slot and nothing leaks into the persisted request. Given that, I would rather not churn the signature (it intentionally mirrors write()'s positional onyxData). If we ever want belt-and-suspenders for untyped JS callers we could move the trailing optionals into an options object, but the type system already closes this off, so I left it as is.

src/libs/API/index.ts
```diff
@@ -199,6 +201,242 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+            try {
+                clearTimeout(safetyTimeoutId);
+                pendingWriteWhenReadyFlushes.delete(flushOnBackground);
+                // Release the barrier only on the early-release paths (safety timeout / app background),
+                // where the barrier may still be pending and would otherwise leave a dangling registration.
+                // On the 'barrier'/'barrierRejected' paths the barrier has already settled, so there is
+                // nothing to cancel - and skipping the call means a custom barrier whose cancel() is unsafe
+                // after settling can never throw here and drop the write.
+                if (reason === 'safetyTimeout' || reason === 'appBackground') {
+                    cancelBarrier();
```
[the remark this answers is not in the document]
JakubKorytko: Agreed, good catch - this became reachable with the previous change that moved cancelBarrier() onto the force paths. Fixed by isolating cancelBarrier() in its own try/catch on the safetyTimeout/appBackground paths: a throwing cancel() is now logged and swallowed, then write() runs regardless, so the forced-release paths always queue the write. Updated the existing background-flush test to assert the throwing-cancel write resolves (not rejects) and that all writes flush, and added a test covering the same on the safety-timeout path.

src/libs/API/index.ts
```diff
@@ -199,6 +201,247 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
     return processRequest(request, CONST.API_REQUEST_TYPE.WRITE);
 }
```
ikevin127: 🟢 `src/libs/API/index.ts`: ordering among multiple deferred writes is unspecified. The JSDoc carefully covers ordering of a deferred write relative to writes dispatched immediately after it, but not ordering between two `writeWhenReady` calls.

Because each call races its own independent barrier, a later call with a fast barrier can hit `write()` (and thus the `SequentialQueue`) before an earlier call with a slow barrier. Today that is harmless since nothing is migrated, but the first migration that fires two deferred optimistic writes with a data dependency (for example an optimistic parent then child) could see them enqueued out of call order.

🔄 Worth one sentence in the JSDoc so migrators do not assume call-order is preserved across deferred writes.

src/libs/API/index.ts
```diff
@@ -199,6 +201,247 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+ * they mean the intended barrier did not release the write. A caller whose custom barrier can
+ * legitimately take longer than the default timeout should raise `safetyTimeoutMs` accordingly.
+ *
+ * @returns A promise that resolves with the underlying `write()` result once the write executes.
+ */
+function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
+    command: TCommand,
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData: OnyxData<TKey> = {},
+    barrier: WriteReadyBarrier = waitForNavigationTransition,
```
ikevin127: 🟢 `src/libs/API/index.ts:362`: `waitForNavigationTransition` leaves an internally-pending promise on the forced-release paths. When the write fires via `safetyTimeout` / `appBackground`, `cancelBarrier()` cancels the `TransitionTracker` registration, so the internal `promise`'s `resolve` is never called and it stays pending forever with the `Promise.resolve(barrierValue).then(...)` handler attached.

It is harmless in practice (the `hasExecuted` guard makes the eventual `.then` a no-op, and the chain is GC-eligible once `writeWhenReady`'s promise settles and nothing else references it), but it is a subtle enough behavior that a one-line comment on the `cancel` handle noting "the underlying promise is intentionally left pending;

`hasExecuted` makes any late resolution a no-op" would save the next reader the trace-through.

tests/unit/APIWriteWhenReadyTest.ts
```diff
@@ -0,0 +1,402 @@
...
+// A never-settling cancelable barrier whose cancel() throws, to exercise the throw-safe execute path.
+function makeThrowingCancelBarrier() {
+    const promise = new Promise<void>(() => {});
+    const cancel = jest.fn(() => {
+        throw new Error('cancel boom');
+    });
+    const barrier: WriteReadyBarrier = {then: promise.then.bind(promise), cancel};
+    return {barrier, cancel};
+}
+
```
ikevin127, line 94: 🟢 `tests/unit/APIWriteWhenReadyTest.ts`: a few branches in the new function are not directly exercised:
- The `catch` around the thunk invocation (`barrierValue = Promise.reject(error)` at `index.ts:249`). The "executes anyway when the barrier rejects" test returns `Promise.reject(...)` from the thunk rather than throwing synchronously, so the synchronous-throw-from-thunk path is untested.
- The outer `try/catch` in `execute` that rejects the returned promise when `write()` / `prepareRequest` throws synchronously (`index.ts:232`).
- The `Log.warn('[API] writeWhenReady background flush threw', ...)` defense-in-depth catch in the flush loop (`index.ts:100`), which cannot currently trigger because `execute` is throw-safe.

✅ These are edge branches, not core behavior, but adding a "thunk throws synchronously" case and a "`write` throws synchronously → returned promise rejects" case would close the two reachable gaps cheaply.

ikevin127 (approved):
### 3 safe to merge tests passed
Since no flow invokes this yet, verify via the suite plus a throwaway call site:
1. **Unit suite (online path):** `npm run test -- APIWriteWhenReadyTest` passes green. This is the primary correctness gate for this PR ✅ 
2. **Deferral applies optimistic data after readiness (online):** temporarily wire `API.writeWhenReady` behind a navigate-away with `optimisticData`, confirm the optimistic Onyx update and re-render happen only after the transition completes (not during it), and that `success`/`failure` data still apply normally on server response ✅ 
3. **Offline persistence:** with the device offline, trigger a deferred write, let the barrier release, and confirm the request is persisted to the `SequentialQueue` and sent on reconnect exactly like a normal `API.write()`, with no duplicate on a single trigger ✅ 

##

🟢 **LGTM**. The PR is safe to merge pending the small documentation additions in the comments above (deferred-write ordering caveat, `waitForNavigationTransition` pending-promise note), none of the findings are blocking.

ikevin127 (approved):

roryabraham (changes requested):
`API.writeWhenReady()` defers a write's optimistic Onyx update until a readiness signal ("barrier") fires. To support early release (safety timeout, app backgrounding), this PR introduces a custom type:

```ts
type CancelableBarrier = PromiseLike<unknown> & {cancel: () => void};
```

The "barrier" is a promise-like that may also expose `cancel()`. `writeWhenReady` duck-types this at runtime and calls `cancel()` only on the early-release paths so a still-pending barrier doesn't leak.

My hunch: this "promise + bolted-on cancel method" is exactly the kind of problem `AbortController`/`AbortSignal` already standardizes. `AbortController`/`AbortSignal` is the idiomatic primitive for "let a caller cancel an in-flight async operation," and using it will simplify things.

Specifically:

- `waitForNavigationTransition()` should take an `AbortSignal` and call `handle.cancel()` from a `signal.addEventListener('abort', ...)` instead of returning a custom `{then, cancel}` object.
- `writeWhenReady` should own an `AbortController`, pass its `signal` to the barrier factory, and call `controller.abort()` on the early-release paths
- `AbortSignal.timeout(safetyTimeoutMs)` combined with `AbortSignal.any([...])` can merge the safety timeout and an app-background abort into a single "give up waiting" signal, replacing the manual `setTimeout`/`clearTimeout`/`pendingWriteWhenReadyFlushes` bookkeeping with a single signal the barrier race listens to.

Net effect: drops the custom `CancelableBarrier` type, the `isCancelableBarrier` type guard, and the manual cancel-closure tracking, in favor of a signal-based cancellation that's more self-documenting and more idiomatic in modern JS

<details>
<summary>diff</summary>

```diff
diff --git a/src/libs/API/index.ts b/src/libs/API/index.ts
index 8c8c505666f..a70c01c0145 100644
--- a/src/libs/API/index.ts
+++ b/src/libs/API/index.ts
@@ -202,26 +202,14 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
 }
 
 /**
- * A barrier that also exposes a `cancel` to release whatever it is waiting on. `writeWhenReady` calls
- * this only when it executes the write before the barrier has settled (released early via the safety
- * timeout, or because the app backgrounds), so a still-pending barrier doesn't leave a dangling
- * registration. When the barrier itself releases the write, it has already settled and `cancel` is not
- * called. Exposing `cancel` is optional - a plain promise-like works too.
+ * A readiness signal for `writeWhenReady`: a function, invoked when the write is queued, that returns a
+ * promise-like resolving once it is safe to apply the write's optimistic data (e.g. after a navigation
+ * transition finishes). It is given an `AbortSignal` that fires if the write executes before the barrier
+ * settles (released early via the safety timeout, or because the app backgrounds), so a barrier waiting on
+ * something cancelable can stop waiting instead of leaving a dangling registration. A rejection is treated
+ * the same as resolving - the write executes anyway.
  */
-type CancelableBarrier = PromiseLike<unknown> & {cancel: () => void};
-
-/**
- * A readiness signal for `writeWhenReady`: a promise-like that resolves once it is safe to apply the
- * write's optimistic data (e.g. after a navigation transition finishes), or a function returning one
- * (invoked when the write is queued). Either form may instead be a `CancelableBarrier`, whose `cancel`
- * is invoked if the write fires before the barrier settles. A rejection is treated the same as
- * resolving - the write executes anyway.
- */
-type WriteReadyBarrier = PromiseLike<unknown> | CancelableBarrier | (() => PromiseLike<unknown> | CancelableBarrier);
-
-function isCancelableBarrier(value: unknown): value is CancelableBarrier {
-    return typeof value === 'object' && value !== null && 'cancel' in value && typeof value.cancel === 'function';
-}
+type WriteReadyBarrier = (signal: AbortSignal) => PromiseLike<unknown>;
 
 /**
  * Why a deferred write was released. `barrier` is the happy path (the barrier resolved); the others all
@@ -287,23 +275,17 @@ function registerBackgroundFlushListener() {
  * Default `writeWhenReady` barrier: resolves once the current or upcoming navigation transition
  * completes. Bounded by TransitionTracker (it stops waiting for an upcoming transition after
  * CONST.MAX_TRANSITION_START_WAIT_MS and auto-ends transitions after CONST.MAX_TRANSITION_DURATION_MS),
- * so it always resolves. Exposes `cancel` so `writeWhenReady` can drop the TransitionTracker
- * registration if the write is released before the transition finishes.
+ * so it always resolves. Drops the TransitionTracker registration via `signal` if the write is released
+ * before the transition finishes.
  */
-function waitForNavigationTransition(): CancelableBarrier {
-    let handle: {cancel: () => void} | undefined;
-    const promise = new Promise<void>((resolve) => {
-        // The executor runs synchronously, so `handle` is assigned before this function returns.
-        handle = TransitionTracker.runAfterTransitions({
+function waitForNavigationTransition(signal: AbortSignal): Promise<void> {
+    return new Promise((resolve) => {
+        const handle = TransitionTracker.runAfterTransitions({
             callback: () => resolve(),
             waitForUpcomingTransition: true,
         });
+        signal.addEventListener('abort', () => handle.cancel());
     });
-    // Return a thin thenable rather than mutating the native Promise with a `cancel` property.
-    return {
-        then: (onFulfilled, onRejected) => promise.then(onFulfilled, onRejected),
-        cancel: () => handle?.cancel(),
-    };
 }
 
 function writeWhenReady<TCommand extends WriteCommand>(command: TCommand, apiCommandParameters: ApiRequestCommandParameters[TCommand]): Promise<void | Response<never>>;
@@ -320,8 +302,8 @@ function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
  * Like `write()`, but defers the entire write - including its optimistic Onyx updates - until a
  * readiness signal fires. By default it waits for the navigation transition to complete, so the
  * expensive optimistic re-render doesn't compete with the transition animation for the main thread.
- * Pass a custom `barrier` (a promise-like, or a function returning one) to wait for a different
- * signal instead.
+ * Pass a custom `barrier` (a function taking an `AbortSignal` and returning a promise-like) to wait for a
+ * different signal instead.
  *
  * Once ready, it delegates to the normal `write()` pipeline, so optimistic/success/failure handling,
  * retries, and queue ordering are all unchanged. (Request de-duplication via `write()`'s
@@ -366,11 +348,11 @@ function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
 
     return new Promise((resolve, reject) => {
         let hasExecuted = false;
-        // These three are read by `execute` but assigned below it; `let` closures capture the binding, not
-        // the value, so the placeholders are safe until the real values are wired up on the next lines.
+        // Read by `execute` but assigned below it; `let` closures capture the binding, not the value, so
+        // this placeholder is safe until the real value is wired up on the next line.
         let flushOnBackground: () => void = () => {};
         let safetyTimeoutId: ReturnType<typeof setTimeout> | undefined;
-        let cancelBarrier: () => void = () => {};
+        const abortController = new AbortController();
         let barrierError: unknown;
 
         const execute = (reason: WriteWhenReadyReleaseReason) => {
@@ -379,25 +361,20 @@ function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
             }
             hasExecuted = true;
 
-            // Everything here runs defensively: a synchronous throw (from write() or prepareRequest) must
-            // settle the returned promise via reject rather than leave it pending forever - and, on the
-            // background path, must not escape into the flush loop and strand the other pending writes. (A
-            // throw from cancelBarrier() is handled separately below so it never drops the write.)
+            // A synchronous throw (from write() or prepareRequest) must settle the returned promise via
+            // reject rather than leave it pending forever - and, on the background path, must not escape
+            // into the flush loop and strand the other pending writes.
             try {
                 clearTimeout(safetyTimeoutId);
                 pendingWriteWhenReadyFlushes.delete(flushOnBackground);
-                // Release the barrier only on the early-release paths (safety timeout / app background),
-                // where the barrier may still be pending and would otherwise leave a dangling registration.
-                // On the 'barrier'/'barrierRejected' paths the barrier has already settled, so there is
-                // nothing to cancel. cancel() is best-effort cleanup and these paths are meant to force the
-                // write through, so isolate a throwing cancel() here - letting it reject would drop the very
-                // write this path exists to guarantee.
+                // Abort only on the early-release paths (safety timeout / app background), where the
+                // barrier may still be pending and would otherwise leave a dangling registration - e.g. the
+                // default barrier's TransitionTracker registration. On the 'barrier'/'barrierRejected' paths
+                // the barrier has already settled, so there is nothing to release. (A throwing abort
+                // listener can't drop the write or escape here: AbortSignal's dispatch isolates listener
+                // errors the same way TransitionTracker isolates its own callbacks.)
                 if (reason === 'safetyTimeout' || reason === 'appBackground') {
-                    try {
-                        cancelBarrier();
-                    } catch (error) {
-                        Log.warn('[API] writeWhenReady barrier cancel() threw during forced release - proceeding with the write', {command, error});
-                    }
+                    abortController.abort();
                 }
 
                 if (reason !== 'barrier') {
@@ -419,19 +396,14 @@ function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
 
         safetyTimeoutId = setTimeout(() => execute('safetyTimeout'), Math.max(0, safetyTimeoutMs));
 
-        // Resolve the barrier to its value once - invoking a thunk inside a try/catch so a synchronously-
-        // thrown error is funneled into the rejection path instead of escaping - then capture its `cancel`
-        // handle (if any) so a still-pending barrier can be released once the write has executed.
-        let barrierValue: unknown;
+        // Invoke the barrier inside a try/catch so a synchronously-thrown error is funneled into the
+        // rejection path instead of escaping.
+        let barrierValue: PromiseLike<unknown>;
         try {
-            barrierValue = typeof barrier === 'function' ? barrier() : barrier;
+            barrierValue = barrier(abortController.signal);
         } catch (error) {
             barrierValue = Promise.reject(error);
         }
-        if (isCancelableBarrier(barrierValue)) {
-            const cancelableBarrier = barrierValue;
-            cancelBarrier = () => cancelableBarrier.cancel();
-        }
         Promise.resolve(barrierValue).then(
             () => execute('barrier'),
             (error: unknown) => {
diff --git a/tests/unit/APIWriteWhenReadyTest.ts b/tests/unit/APIWriteWhenReadyTest.ts
index 78b5ac47c04..203927b690f 100644
--- a/tests/unit/APIWriteWhenReadyTest.ts
+++ b/tests/unit/APIWriteWhenReadyTest.ts
@@ -57,26 +57,32 @@ async function flushMicrotasks(until: () => boolean = () => false, maxIterations
 // The awaited effect these tests key off is the push to the SequentialQueue.
 const pushHappened = () => mockPush.mock.calls.length > 0;
 
-// A cancelable barrier (promise-like + `cancel`) that stays pending until `release()` is called, so a
-// test can control exactly when (or whether) it settles and assert on its cancel handle.
-function makeCancelableBarrier() {
+// A barrier that stays pending until `release()` is called, and records whether the `AbortSignal` passed
+// to it was aborted - so a test can control exactly when (or whether) it settles and assert on abort.
+function makeAbortableBarrier() {
     let release: () => void = () => {};
-    const promise = new Promise<void>((resolve) => {
-        release = resolve;
-    });
-    const cancel = jest.fn();
-    const barrier: WriteReadyBarrier = {then: promise.then.bind(promise), cancel};
-    return {barrier, release, cancel};
+    const onAbort = jest.fn();
+    const barrier: WriteReadyBarrier = (signal: AbortSignal) => {
+        signal.addEventListener('abort', onAbort);
+        return new Promise<void>((resolve) => {
+            release = resolve;
+        });
+    };
+    return {barrier, release, onAbort};
 }
 
-// A never-settling cancelable barrier whose cancel() throws, to exercise the throw-safe execute path.
-function makeThrowingCancelBarrier() {
-    const promise = new Promise<void>(() => {});
-    const cancel = jest.fn(() => {
-        throw new Error('cancel boom');
+// A never-settling barrier whose abort listener throws, to prove a throwing listener can't drop the write
+// or break the background flush loop - this is guaranteed by AbortSignal's event dispatch (which isolates
+// listener errors), not by writeWhenReady's own code, but the invariant is worth pinning here.
+function makeThrowingAbortBarrier() {
+    const onAbort = jest.fn(() => {
+        throw new Error('abort boom');
     });
-    const barrier: WriteReadyBarrier = {then: promise.then.bind(promise), cancel};
-    return {barrier, cancel};
+    const barrier: WriteReadyBarrier = (signal: AbortSignal) => {
+        signal.addEventListener('abort', onAbort);
+        return new Promise<void>(() => {});
+    };
+    return {barrier, onAbort};
 }
 
 describe('API.writeWhenReady', () => {
@@ -131,22 +137,6 @@ describe('API.writeWhenReady', () => {
         expect(mockPush).toHaveBeenCalledTimes(1);
     });
 
-    it('accepts a promise-like barrier directly, not only a thunk', async () => {
-        let releaseBarrier: () => void = () => {};
-        const barrierPromise = new Promise<void>((resolve) => {
-            releaseBarrier = resolve;
-        });
-
-        deferWrite(barrierPromise);
-        await flushMicrotasks();
-        expect(mockPush).not.toHaveBeenCalled();
-
-        releaseBarrier();
-        await flushMicrotasks(pushHappened);
-
-        expect(mockPush).toHaveBeenCalledTimes(1);
-    });
-
     it('flushes a pending write when the app goes to the background', async () => {
         const barrier = () => new Promise<void>(() => {});
 
@@ -231,41 +221,41 @@ describe('API.writeWhenReady', () => {
         }
     });
 
-    it('cancels a still-pending cancelable barrier when released early via the safety timeout', async () => {
+    it("aborts a still-pending barrier's signal when released early via the safety timeout", async () => {
         jest.useFakeTimers();
         try {
-            const {barrier, cancel} = makeCancelableBarrier(); // never released
+            const {barrier, onAbort} = makeAbortableBarrier(); // never released
 
             deferWrite(barrier);
             await flushMicrotasks();
-            expect(cancel).not.toHaveBeenCalled();
+            expect(onAbort).not.toHaveBeenCalled();
 
             await jest.advanceTimersByTimeAsync(SAFETY_TIMEOUT_MS);
             await flushMicrotasks(pushHappened);
 
             expect(mockPush).toHaveBeenCalledTimes(1);
-            expect(cancel).toHaveBeenCalledTimes(1);
+            expect(onAbort).toHaveBeenCalledTimes(1);
         } finally {
             jest.useRealTimers();
         }
     });
 
-    it('cancels a still-pending cancelable barrier when flushed on background', async () => {
-        const {barrier, cancel} = makeCancelableBarrier(); // never released
+    it("aborts a still-pending barrier's signal when flushed on background", async () => {
+        const {barrier, onAbort} = makeAbortableBarrier(); // never released
 
         deferWrite(barrier);
         await flushMicrotasks();
-        expect(cancel).not.toHaveBeenCalled();
+        expect(onAbort).not.toHaveBeenCalled();
 
         emitAppState('background');
         await flushMicrotasks(pushHappened);
 
         expect(mockPush).toHaveBeenCalledTimes(1);
-        expect(cancel).toHaveBeenCalledTimes(1);
+        expect(onAbort).toHaveBeenCalledTimes(1);
     });
 
-    it('does not cancel a cancelable barrier that resolves normally', async () => {
-        const {barrier, release, cancel} = makeCancelableBarrier();
+    it("does not abort a barrier's signal that resolves normally", async () => {
+        const {barrier, release, onAbort} = makeAbortableBarrier();
 
         deferWrite(barrier);
         await flushMicrotasks();
@@ -275,8 +265,8 @@ describe('API.writeWhenReady', () => {
         await flushMicrotasks(pushHappened);
 
         expect(mockPush).toHaveBeenCalledTimes(1);
-        // The barrier itself released the write, so it has already settled - there is nothing to cancel.
-        expect(cancel).not.toHaveBeenCalled();
+        // The barrier itself released the write, so it has already settled - there is nothing to abort.
+        expect(onAbort).not.toHaveBeenCalled();
     });
 
     it('honors a custom safetyTimeoutMs', async () => {
@@ -313,8 +303,8 @@ describe('API.writeWhenReady', () => {
         expect(mockPush).toHaveBeenCalledTimes(3);
     });
 
-    it('does not drop a write whose barrier cancel() throws during background flush, and still flushes the others', async () => {
-        const {barrier: throwingBarrier, cancel} = makeThrowingCancelBarrier();
+    it('does not drop a write whose barrier abort listener throws during background flush, and still flushes the others', async () => {
+        const {barrier: throwingBarrier, onAbort} = makeThrowingAbortBarrier();
         // Track that the throwing write settles (does not hang) - and, per the isolation, resolves.
         const throwingOutcome = deferWrite(throwingBarrier).then(
             () => 'resolved',
@@ -324,20 +314,20 @@ describe('API.writeWhenReady', () => {
         deferWrite(neverSettlingBarrier());
         await flushMicrotasks();
 
-        // A throwing cancel() is best-effort cleanup on the forced-release path: it must neither abort the
-        // loop flushing the other two, nor drop its own write. All three writes should flush.
+        // A throwing abort listener is isolated by AbortSignal's dispatch: it must neither abort the loop
+        // flushing the other two, nor drop its own write. All three writes should flush.
         expect(() => emitAppState('background')).not.toThrow();
         await flushMicrotasks(() => mockPush.mock.calls.length >= 3);
 
         expect(mockPush).toHaveBeenCalledTimes(3);
-        expect(cancel).toHaveBeenCalledTimes(1);
+        expect(onAbort).toHaveBeenCalledTimes(1);
         await expect(throwingOutcome).resolves.toBe('resolved');
     });
 
-    it('does not drop the write when the barrier cancel() throws on the safety timeout path', async () => {
+    it('does not drop the write when the barrier abort listener throws on the safety timeout path', async () => {
         jest.useFakeTimers();
         try {
-            const {barrier: throwingBarrier, cancel} = makeThrowingCancelBarrier(); // never settles
+            const {barrier: throwingBarrier, onAbort} = makeThrowingAbortBarrier(); // never settles
             const outcome = deferWrite(throwingBarrier).then(
                 () => 'resolved',
                 () => 'rejected',
@@ -347,9 +337,9 @@ describe('API.writeWhenReady', () => {
             await jest.advanceTimersByTimeAsync(SAFETY_TIMEOUT_MS);
             await flushMicrotasks(pushHappened);
 
-            // The safety timeout must force the write through even though cancel() threw during cleanup.
+            // The safety timeout must force the write through even though the abort listener threw.
             expect(mockPush).toHaveBeenCalledTimes(1);
-            expect(cancel).toHaveBeenCalledTimes(1);
+            expect(onAbort).toHaveBeenCalledTimes(1);
             await expect(outcome).resolves.toBe('resolved');
         } finally {
             jest.useRealTimers();
```
</details>

src/libs/API/index.ts
JakubKorytko: Good point, added a paragraph to the writeWhenReady JSDoc spelling out that call order is NOT preserved across deferred writes (each races its own barrier), and telling migrators to gate a dependent deferred write on the first ones barrier rather than assuming order.

src/libs/API/index.ts
JakubKorytko: Ended up refactoring this whole path to AbortSignal, so the custom {then, cancel} object is gone: waitForNavigationTransition now takes an AbortSignal and drops the TransitionTracker registration from an abort listener. The internal promise is still intentionally left pending on the early-release paths and I kept a comment saying so - hasExecuted makes any late resolution a no-op, and the chain is GC-eligible once the writeWhenReady promise settles.

tests/unit/APIWriteWhenReadyTest.ts
JakubKorytko: Added both cheap cases: a thunk-throws-synchronously test (covers the try/catch around the thunk invocation) and a write-throws-synchronously test that spies on Onyx.update to throw during prepareRequest and asserts the returned promise rejects. Left the flush-loop Log.warn catch alone since its unreachable while execute stays throw-safe.

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+    // suspension.
+    AppState.addEventListener('change', (nextState) => {
+        if (nextState !== CONST.APP_STATE.BACKGROUND || pendingWriteWhenReadyFlushes.size === 0) {
+            return;
+        }
+        Log.info(`[API] App going to "${nextState}" - flushing ${pendingWriteWhenReadyFlushes.size} pending writeWhenReady write(s)`, false);
+        // Isolate each flush (like TransitionTracker does its callbacks) so one throwing write can't abort
+        // the loop and strand the remaining pending writes right before the process is suspended. `execute`
+        // is already throw-safe; this is defense-in-depth.
+        for (const flush of [...pendingWriteWhenReadyFlushes]) {
```
roryabraham: NAB: is this shallow copy necessary?

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+            waitForUpcomingTransition: true,
+        });
+        // On abort, drop the TransitionTracker registration. The promise is intentionally left pending (its
+        // `resolve` is never called on this path): writeWhenReady's `hasExecuted` guard makes any late
+        // resolution a no-op, and the chain is GC-eligible once the writeWhenReady promise settles.
+        signal.addEventListener('abort', () => handle.cancel());
+    });
+}
+
+function writeWhenReady<TCommand extends WriteCommand>(command: TCommand, apiCommandParameters: ApiRequestCommandParameters[TCommand]): Promise<void | Response<never>>;
```
roryabraham: I'd greatly prefer if all the overloads are stacked together:

```ts
...
...
 * @returns A promise that resolves with the underlying `write()` result once the write executes.
 */
function writeWhenReady<TCommand extends WriteCommand>(command: TCommand, apiCommandParameters: ApiRequestCommandParameters[TCommand]): Promise<void | Response<never>>;
function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
    command: TCommand,
    apiCommandParameters: ApiRequestCommandParameters[TCommand],
    onyxData?: OnyxData<TKey>,
    barrier?: WriteReadyBarrier,
    safetyTimeoutMs?: number,
): Promise<void | Response<TKey>>;
function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
    command: TCommand,
    apiCommandParameters: ApiRequestCommandParameters[TCommand],
    onyxData: OnyxData<TKey> = {},
    barrier: WriteReadyBarrier = waitForNavigationTransition,
    safetyTimeoutMs: number = WRITE_WHEN_READY_SAFETY_TIMEOUT_MS,
): Promise<void | Response<TKey>> {
...
...
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+ * the same as resolving - the write executes anyway.
+ */
+type WriteReadyBarrier = (signal: AbortSignal) => PromiseLike<unknown>;
+
+/**
+ * Why a deferred write was released. `barrier` is the happy path (the barrier resolved); the others all
+ * mean the barrier did not cleanly release the write - it rejected, timed out, or the app backgrounded
+ * first - and are logged distinctly.
+ */
+type WriteWhenReadyReleaseReason = 'barrier' | 'barrierRejected' | 'safetyTimeout' | 'appBackground';
```
roryabraham: I think we can clean this up and make it more self-documenting using:

```ts
type WriteWhenReadyReleaseReason = 'success' | 'rejected' | 'safetyTimeout' | 'appBackground';
```

Though, if the write happens in both the success and rejected case, I'm not sure what we gain by separating those out.

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
     return processRequest(request, CONST.API_REQUEST_TYPE.WRITE);
 }
 
+/**
```
roryabraham: Let's distill this comment:

```
/**
 * A readiness check used by `writeWhenReady`.
 * This function is called when the write is queued, and the write will only happen when either:
 * - The returned promise settles (resolves or rejects), OR
 * - The AbortSignal is called (so a cancelled promise doesn't delay the write forever)
 */
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+type WriteReadyBarrier = (signal: AbortSignal) => PromiseLike<unknown>;
+
+/**
+ * Why a deferred write was released. `barrier` is the happy path (the barrier resolved); the others all
+ * mean the barrier did not cleanly release the write - it rejected, timed out, or the app backgrounded
+ * first - and are logged distinctly.
+ */
+type WriteWhenReadyReleaseReason = 'barrier' | 'barrierRejected' | 'safetyTimeout' | 'appBackground';
+
+// Default upper bound on how long writeWhenReady waits for a barrier before executing regardless, so
```
roryabraham: Let's distill this comment:

```ts
// How long writeWhenReady waits for the barrier before executing, so the write happens even if the promise never settles.
// Should be at least longer than the duration of the default barrier (TransitionTracker.runAfterTransitions)
const WRITE_WHEN_READY_SAFETY_TIMEOUT_MS = 5 * CONST.MAX_TRANSITION_DURATION_MS;
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+// duration: long enough that the bounded default barrier always wins the race, short enough to still
+// bound a stuck custom barrier. Callers with a legitimately slower custom barrier can raise it via the
+// `safetyTimeoutMs` argument.
+// NOTE: the default barrier's worst case is MAX_TRANSITION_START_WAIT_MS + MAX_TRANSITION_DURATION_MS,
+// so this timeout must stay above that sum for the "default barrier always wins" invariant to hold. The
+// x5 margin covers it while both constants are ~equal; a unit test pins the invariant against drift.
+const SAFETY_TIMEOUT_TRANSITION_MULTIPLIER = 5;
+const WRITE_WHEN_READY_SAFETY_TIMEOUT_MS = CONST.MAX_TRANSITION_DURATION_MS * SAFETY_TIMEOUT_TRANSITION_MULTIPLIER;
+
+// Deferred writes still waiting on their barrier. Tracked so they can be force-flushed when the app
```
roryabraham: Let's distill the comment:

```ts
// Pending deferred writes. Tracked so they can be force-flushed when the app backgrounds.
// Because JS timeers are suspended in the background, we flush before the OS suspends the process.
// That ensures the optimistic data + queued request are not lost.
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+// backgrounds: JS timers are suspended in the background, so the per-write safety timeout cannot be
+// relied on to fire before the OS suspends/kills the process, and the optimistic data + queued
+// request would otherwise be lost.
+const pendingWriteWhenReadyFlushes = new Set<() => void>();
```
roryabraham: Also, can we simplify this variable name to just `pendingWrites`?

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+
+let hasRegisteredBackgroundFlushListener = false;
+
+// Subscribe to AppState lazily, on the first deferred write, so importers that never call
+// writeWhenReady don't pay for the subscription. Once registered the listener is intentionally never
+// removed - it lives as long as the app (matching other module singletons) and guards every later
+// deferred write.
+function registerBackgroundFlushListener() {
```
roryabraham: Let's simplify the comment:

```
/**
 * Subscribe to AppState lazily on the first deferred write,
 * so importers that never call writeWhenReady don't pay for the subscription.
 */
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+    if (hasRegisteredBackgroundFlushListener) {
+        return;
+    }
+    hasRegisteredBackgroundFlushListener = true;
+
+    // Flush only on a full `background` transition, not the transient `inactive` state (Control Center,
```
roryabraham: Distill comment:

```
// Flush only on a full `background` transition, not the transient `inactive` state (Control Center, the app switcher, permission/biometric prompts).
// `background` is the last event before the OS can suspend the process, so it's our best opportuntity for a last-minute, best-effort flush.
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+    // by running the optimistic re-render during a blip the user immediately returns from. `background`
+    // is the last event before the OS can suspend the process, so flushing here applies the optimistic
+    // data and enqueues the request for persistence (best-effort - Onyx disk writes are async) before
+    // suspension.
+    AppState.addEventListener('change', (nextState) => {
+        if (nextState !== CONST.APP_STATE.BACKGROUND || pendingWriteWhenReadyFlushes.size === 0) {
+            return;
+        }
+        Log.info(`[API] App going to "${nextState}" - flushing ${pendingWriteWhenReadyFlushes.size} pending writeWhenReady write(s)`, false);
+        // Isolate each flush (like TransitionTracker does its callbacks) so one throwing write can't abort
```
roryabraham: Distill the comment:

```
// Isolate each flush so one throwing write can't abort the loop and strand the remaining pending writes
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+            try {
+                flush();
+            } catch (error) {
+                Log.warn('[API] writeWhenReady background flush threw', {error});
+            }
+        }
+    });
+}
+
+/**
```
roryabraham: Distill the comment:

```
/**
 * Default `writeWhenReady` barrier: resolves once the current or upcoming navigation transition completes.
 */
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+ * so it always resolves. Drops the TransitionTracker registration via `signal` if the write is released
+ * before the transition finishes.
+ */
+function waitForNavigationTransition(signal: AbortSignal): Promise<void> {
+    return new Promise((resolve) => {
+        const handle = TransitionTracker.runAfterTransitions({
+            callback: () => resolve(),
+            waitForUpcomingTransition: true,
+        });
+        // On abort, drop the TransitionTracker registration. The promise is intentionally left pending (its
```
roryabraham: Distill the comment:

```
// On abort, drop the TransitionTracker registration.
// The promise is intentionally left pending because resolving or rejecting would execute the write that was cancelled via the signal
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+
+function writeWhenReady<TCommand extends WriteCommand, TKey extends OnyxKey>(
+    command: TCommand,
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData?: OnyxData<TKey>,
+    barrier?: WriteReadyBarrier,
+    safetyTimeoutMs?: number,
+): Promise<void | Response<TKey>>;
+
+/**
```
roryabraham: Let's distill the comment:

```
/**
 * Like `write()`, but deferred until a readiness signal (barrier) is fired.
 *   - Optimistic updates are intetionally deferred, as well as the API request itself. 
 *   - Once ready, it delegates to the normal `write()` pipeline.
 *   - The write is executed after `safetyTimeoutMs`, regardless of whether the barrier promise has settled.
 *
 * Caveats:
 *   - This currently does not support the `conflictResolver` param of `write()`.
 *     Any request queued via `writeWhenReady` will not enter the sequential queue or be considered in conflict resolution until the deferred execution happens
 *   - Call order is NOT preserved across multiple deferred writes. If two indepenent `writeWhenReady` calls are queued, the two barriers race eachother.
 *     If you need to preserve call order, it must be managed by callers: the later barrier must await reoslution of the former.
 *
 * Caution:
 *   - By default, it waits for the navigation transition to complete, so re-renders from optimistic writes doen't compete with the transition animation.
 *     If no navigation happens, then the write is deferred by roughly `CONST.MAX_TRANSITION_START_WAIT_MS + CONST.MAX_TRANSITION_DURATION_MS` (~2s).
 *     The default behavior can be overridden by passing a custom `barrier`
 *   - In the edge case where a user backgrounds or hard-kills the app before the write finishes, we make a best-effort to flush pending writes.
 *     This is not guaranteed to succeed, so critical writes (such as ones that move money) should not be deferred.
 *
 * @returns A promise that resolves with the underlying `write()` result once the write executes.
 */
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+    apiCommandParameters: ApiRequestCommandParameters[TCommand],
+    onyxData: OnyxData<TKey> = {},
+    barrier: WriteReadyBarrier = waitForNavigationTransition,
+    safetyTimeoutMs: number = WRITE_WHEN_READY_SAFETY_TIMEOUT_MS,
+): Promise<void | Response<TKey>> {
+    Log.info('[API] Called API writeWhenReady', false, buildLogParams(command, apiCommandParameters ?? {}));
+
+    return new Promise((resolve, reject) => {
+        let hasExecuted = false;
+        // Read by `execute` but assigned below it; `let` closures capture the binding, not the value, so
```
roryabraham: I have read this comment a few times and it's pretty confusing. I kind of think we should just remove it

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+        // this placeholder is safe until the real value is wired up on the next line.
+        let flushOnBackground: () => void = () => {};
+        let safetyTimeoutId: ReturnType<typeof setTimeout> | undefined;
```
roryabraham: nit code style:

```suggestion
        let safetyTimeoutID: ReturnType<typeof setTimeout> | undefined;
```

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+        const abortController = new AbortController();
+        let barrierError: unknown;
+
+        const execute = (reason: WriteWhenReadyReleaseReason) => {
+            if (hasExecuted) {
+                return;
+            }
+            hasExecuted = true;
+
+            // A synchronous throw (from write() or prepareRequest) must settle the returned promise via
```
roryabraham: I think this comment should be removed as well. It's just explaining why we have a `try`/`catch` and why we reject the promise in the catch. I think it's pretty self-explanatory that we would reject the promise on some error in the write path.

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+            // reject rather than leave it pending forever - and, on the background path, must not escape
+            // into the flush loop and strand the other pending writes.
+            try {
+                clearTimeout(safetyTimeoutId);
+                pendingWriteWhenReadyFlushes.delete(flushOnBackground);
+                // Abort only on the early-release paths (safety timeout / app background), where the barrier
```
roryabraham: Let's distill the comment:

```
// Abort on the early-release paths so that pending barriers can cleanup if they desire.
// On the 'success'/'rejected' paths, the barrier has already settled, so there's nothing to release.
// Barriers whose cleanup can throw should handle it in their own abort listener.
```

roryabraham (changes requested):
A few high-level comments:

1. I'm not sure it's intuitive to flush the write when the barrier rejects. I kind of think that should kill the write (just like signalling the AbortController does).
2. I wouldn't mind splitting up src/libs/API/index.ts a bit to keep each file more focused and easy to read. I was thinking something like:
    - write, read, waitForWrites, and makeRequestWithSideEffects, pagination, and writeWhenReady can each go in their own files
    - writeWithNoDuplicatesConflictAction and the other conflict ones should really just be `libs/API/ConflictResolvers.ts` with some canned conflict resolvers that can be passed to `API.write` at the callsites. We don't need wrapper functions.
    - index.ts is just a barrel for organization

    Even if we just put writeWhenReady and its helpers in its own file, I think that would simplify things. Both by making the file easier to follow, and by simplifying module-private variable names (i.e: `WriteWhenReadyReleaseReason` -> `ReleaseReason`, `WRITE_WHEN_READY_SAFETY_TIMEOUT_MS` -> `SAFETY_TIMEOUT_MS`, etc...)

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+    // Flush only on a full `background` transition, not the transient `inactive` state (Control Center,
+    // the app switcher, permission/biometric prompts) - flushing on those would defeat the perf point
+    // by running the optimistic re-render during a blip the user immediately returns from. `background`
+    // is the last event before the OS can suspend the process, so flushing here applies the optimistic
+    // data and enqueues the request for persistence (best-effort - Onyx disk writes are async) before
+    // suspension.
+    AppState.addEventListener('change', (nextState) => {
+        if (nextState !== CONST.APP_STATE.BACKGROUND || pendingWriteWhenReadyFlushes.size === 0) {
+            return;
+        }
```
Julesssss: Should we flush here if the app is currently backgrounded? This catches new background events, but what about the edge case where we defer the write AFTER app goes to background (for example push notification). Not sure if we catch that edge case elsewhere

src/libs/API/index.ts
```diff
@@ -199,6 +201,229 @@ function write<TCommand extends WriteCommand, TKey extends OnyxKey>(
...
+ * completes. Bounded by TransitionTracker (it stops waiting for an upcoming transition after
+ * CONST.MAX_TRANSITION_START_WAIT_MS and auto-ends transitions after CONST.MAX_TRANSITION_DURATION_MS),
+ * so it always resolves. Drops the TransitionTracker registration via `signal` if the write is released
+ * before the transition finishes.
+ */
+function waitForNavigationTransition(signal: AbortSignal): Promise<void> {
+    return new Promise((resolve) => {
+        const handle = TransitionTracker.runAfterTransitions({
+            callback: () => resolve(),
+            waitForUpcomingTransition: true,
```
Julesssss: Is this any transition? Could we further restrict to `navigation` here?

src/libs/API/index.ts
JakubKorytko: kept it, but I tightened the comment to say why. `execute` deleting its own entry mid-iteration is safe without the copy, so that's not the reason. The copy is for the case where a flush synchronously queues a new deferred write: without it that new write lands in the same loop and gets flushed immediately, losing its deferral.

src/libs/API/index.ts
JakubKorytko: done, JSDoc moved above the first overload and all three signatures are adjacent now.

src/libs/API/index.ts
JakubKorytko: renamed to `'success' | 'rejected' | 'safetyTimeout' | 'appBackground'` and dropped the JSDoc since the names carry it now. Note it ended up as just `ReleaseReason` rather than `WriteWhenReadyReleaseReason` - it now lives in its own module (per your high-level comment), so the prefix was redundant.

I kept success/rejected separate because the logging differs: 'rejected' warns and attaches the barrier error, 'success' doesn't log at all. Merging them would either lose the error or warn on the happy path.

src/libs/API/index.ts
JakubKorytko: applied, with one wording tweak. The signal doesn't cause the write, it fires because the write is already being released early (timeout/background), so I reworded that bullet so it doesn't read as cause and effect.

src/libs/API/index.ts
JakubKorytko: done. Also dropped the SAFETY_TIMEOUT_TRANSITION_MULTIPLIER constant so the 5x is inline like your version, and shortened the name to `SAFETY_TIMEOUT_MS` now that it lives in writeWhenReady.ts.

src/libs/API/index.ts
JakubKorytko: done

src/libs/API/index.ts
JakubKorytko: renamed to `pendingWrites`

src/libs/API/index.ts
JakubKorytko: done

src/libs/API/index.ts
JakubKorytko: done

src/libs/API/index.ts
JakubKorytko: done, plus one extra line explaining the copy

src/libs/API/index.ts
JakubKorytko: partly done. That function no longer exists in this form: to answer Jules' question on the same block I turned it into `createTransitionBarrier(waitFor)`, so there's now a param that needs documenting and a one-liner won't cover it. The doc is still much shorter than what was there before, but it isn't the single line you suggested. Happy to trim further if you think it's still too much.

src/libs/API/index.ts
JakubKorytko: done

src/libs/API/index.ts
JakubKorytko: done, applied as written

src/libs/API/index.ts
JakubKorytko: removed

src/libs/API/index.ts
JakubKorytko: done

src/libs/API/index.ts
JakubKorytko: removed

src/libs/API/index.ts
JakubKorytko: done

src/libs/API/index.ts
JakubKorytko: good catch, that case wasn't covered anywhere. Added an `AppState.currentState === CONST.APP_STATE.BACKGROUND` check up front in writeWhenReady: if we're already backgrounded we execute right away instead of setting up a barrier and a timer that a suspended process may never fire. Added a test for it.

src/libs/API/index.ts
JakubKorytko: yeah, any transition. Worth knowing only half of it is kind-aware though: `waitForUpcomingTransition` picks which transition *opens the gate*, but once it's open `runAfterTransitions` recurses without a kind and waits on `activeTransitions.size === 0`, which counts every kind. So we already wait for whatever is animating regardless.

I left the default as any-transition on purpose: the point of deferring is to keep the optimistic re-render off the main thread while something is animating, and a modal or the keyboard contends for it the same as a screen push does. Restricting the default would also mean sitting on MAX_TRANSITION_START_WAIT_MS (~1s) any time no screen transition fires.

That said your instinct is right that some callers want the stricter gate, so I made it a parameter instead: `createTransitionBarrier('navigation')` is exported for callers that know they navigate away and don't want a stray keyboard dismiss releasing the write early. Also renamed the default barrier from `waitForNavigationTransition` to `waitForTransition`, since the old name is what made this look nav-only in the first place. Tests for both.

roryabraham (dismissed):

tests/unit/APIWriteWhenReadyTest.ts
```diff
@@ -0,0 +1,469 @@
...
+                data: expect.objectContaining({value: CONST.LOCALES.EN}),
+                successData: onyxData?.successData,
+            }),
+        );
+        await expect(promise).resolves.toBeUndefined();
+    });
+
+    it('keeps the safety timeout above the default barrier worst case (guards constant drift)', () => {
+        expect(SAFETY_TIMEOUT_MS).toBeGreaterThan(CONST.MAX_TRANSITION_START_WAIT_MS + CONST.MAX_TRANSITION_DURATION_MS);
+    });
```
Julesssss, line 476: This test is only checking the const defined within the test, right? Can we use the real const, that would be a better test I think.

Line :28 `
const SAFETY_TIMEOUT_MS = CONST.MAX_TRANSITION_DURATION_MS * 5;`

tests/unit/APIWriteWhenReadyTest.ts
```diff
@@ -0,0 +1,469 @@
...
+            await jest.advanceTimersByTimeAsync(1);
+            await flushMicrotasks(pushHappened);
+
+            expect(mockPush).toHaveBeenCalledTimes(1);
+        } finally {
+            jest.useRealTimers();
+        }
+    });
+
+    it('flushes every pending write on a single background event', async () => {
```
Julesssss, line 369: So this asserts the flush, but could we also assert that `AppState.addEventListener` doesn't get called again? It would help prevent regression in the future in case the `hasRegisteredBackgroundFlushListener ` guard changes.

tests/unit/APIWriteWhenReadyTest.ts
JakubKorytko: good catch, it was only pinning a copy of the formula. Exported SAFETY_TIMEOUT_MS from writeWhenReady.ts and the test now imports the real one, so the drift guard (and every advanceTimersByTime in the file) keys off the actual value. Verified it fails if the constant drops below the default barrier's worst case.

tests/unit/APIWriteWhenReadyTest.ts
JakubKorytko: added. The tricky part is that the subscription is created once per module lifetime, so an absolute count would depend on test order. So it now snapshots addEventListener's call count after the first deferred write and asserts the next two add nothing on top. Sanity checked it by disabling the hasRegisteredBackgroundFlushListener guard: the test fails with 3 registrations instead of 1, so it is not vacuous.

roryabraham (approved):

Julesssss (approved):

PR #96496, opened 2026-07-20
Confirm new expenses with a “View” toast instead of scrolling to and highlighting the row

mhawryluk:
### Explanation of Change

Shows an "Expense added" growl (with a "View" action) after an expense is created from FAB outside of its report‚ as well as when duplicating and splitting expenses on Spend. Removes the highlighting of new expenses in the Spend tab tables completely. 

It's driven declaratively by a single _ExpenseAddedGrowl_ component, which watches an Onyx signal that the creation flows write to (same as they did for highlighting). In the [previous approach](https://github.com/Expensify/App/pull/94866) each creation flow triggered the growl and its navigation imperatively, passing values that could go stale before the user tapped "View". Centralizing it in one Onyx-driven component removes that duplication and the stale-data problem.

### Fixed Issues

PROPOSAL: N/A

### Tests

1. _Growl + "View" from global create (Inbox tab)_
    - Go to the Inbox tab.
    - Tap the green + (FAB) button → Create expense.
    - Enter an amount → tap Next → select a workspace/participant → tap Next.
    - On the confirmation page, tap Create expense.
    - Verify a green "Expense added" growl with a "View" button appears.
    - Repeat the steps above for each expense type in the Create expense flow - Manual, Scan (upload a receipt), Distance (add waypoints), and Per diem (on a workspace with per-diem rates enabled).

2. _Growl in the Spend tab_
    - Go to a different tab than Inbox.
    - Tap the + (FAB) → Create expense → enter an amount → Next → select a workspace/participant → Next → Create expense.
    - Verify the app lands on the Spend tab, and the same growl + working View appears there. The rows in the table should not be highlighted anymore.

3. _"View" opens the correct destination for the current surface_
    - Spend (or any non-Inbox) tab: the new expense's details open in a right-hand panel over the Spend view.
    - Inbox, narrow layout: opens the thread as a full-screen report.
    - Inbox, wide, report with a single expense: the new expense's details open directly in the right-hand panel.
    - Inbox, wide, report with multiple expenses: opens the expense report super-wide with the new expense's thread RHP stacked on top.
    - Inbox, wide, tracked/self-DM expense (no expense report): opens the thread as a full-screen report.
    - With an expense/report already open in the RHP: View replaces it (does not stack a second RHP).
    - After the growl appears, switch to a different tab, then tap View: it navigates per the tab you're on now.

4. _Single growl/single navigation on a batch (skip-confirm multi-scan)_
    - First, create one Scan expense for a recipient (FAB → Create expense → Scan → upload a receipt → Next → pick the recipient → Create expense) so the FAB shows a "Scan receipt" quick action for them.
    - Tap the FAB → Scan receipt quick action → add/upload 2 receipts and submit (this path has no confirmation page).
    - Verify navigation happens once (no flicker/double transition) and exactly one "Expense added" growl appears, for the last receipt of the batch.
    - Tap View and verify it opens the last created expense.

5. _Invoice_
    - Tap the FAB → Send invoice (requires a workspace with Invoices enabled under _More features_).
    - Enter an amount → Next → pick a recipient → Next → Send invoice.
    - Verify the growl reads "Invoice sent" and its View button works.

6. _Split an existing expense into multiple_
    - Open an existing expense → tap the ⋮ (more) menu → Split expense.
    - Split it into 2+ expenses and confirm.
    - Verify one "Expense added" growl appears (for the last split transaction); View opens that expense.

7. _Bulk duplicate expenses_
    - From the Spend table: select 2+ transactions (checkboxes) → "n selected" dropdown → Duplicate expenses.
    - Verify one growl appears; View opens the (last) duplicated expense.

8. _Self-DM expense_
    - Tap the FAB → Create expense → enter an amount → Next → select your own account (self-DM) as the recipient → Create expense.
    - Verify the growl appears; View opens the transaction thread as a full-screen report (not an RHP).
    - Afterward, open your self-DM chat, tap the same expense preview, and verify it opens correctly.

9. _Share flow (native only)_
    - From another app, Share an image into Expensify → choose submit expense → fill details → submit.
    - Verify a single "Expense added" growl appears after submission with a working View, and navigation happens once.

### Offline tests

- Go offline, then create an expense from the FAB on the Inbox tab (Create expense → amount → Next → participant → Next → Create expense).
- Verify the "Expense added" growl still appears and View opens the optimistic transaction thread.
- While offline on the Spend tab, create another expense; verify the list refetches once after you go back online.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/b93612f6-1111-4aa6-9eee-a44b481d474c

https://github.com/user-attachments/assets/7b5c52bb-a628-4fa4-a8c7-c40a0803f5fb

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/320d0c5e-52d2-4ec7-b596-77e188de759b

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/3d72e816-84fe-43c3-8610-e413d2fe8695

https://github.com/user-attachments/assets/75160d7a-03a9-4ce9-9a23-63393a170809

</details>


src/components/Search/index.tsx
```diff
@@ -14,7 +13,7 @@ import usePolicyForMovingExpenses from '@hooks/usePolicyForMovingExpenses';
 import usePrevious from '@hooks/usePrevious';
 import useResponsiveLayout from '@hooks/useResponsiveLayout';
 import useSaveSortedReportIDs from '@hooks/useSaveSortedReportIDs';
-import useSearchHighlightAndScroll from '@hooks/useSearchHighlightAndScroll';
+import useSearchAutoRefetch from '@hooks/useSearchHighlightAndScroll';
```
JakubKorytko: the hook is now `useSearchAutoRefetch`, but the file (and its test) are still named `useSearchHighlightAndScroll`. Could we rename the file to `useSearchAutoRefetch.ts` so it matches the hook and is easier to find?

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const [introSelected] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED);
+    const transactionID = active?.transactionID ?? Object.keys(signal ?? {}).at(-1);
+    const [transaction] = useOnyx(`${ONYXKEYS.COLLECTION.TRANSACTION}${transactionID}`);
+    const reportID = transaction?.reportID;
+    const [report] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT}${reportID}`);
+    // A tracked/unreported expense's transaction.reportID is UNREPORTED_REPORT_ID, and its IOU/track action lives
+    // on the self-DM report - not on transaction.reportID. Read actions from the report that actually hosts the
+    // action so "View" resolves the real transaction thread (the action's childReportID) instead of fabricating
+    // a mismatched optimistic one.
+    const hostReportID = !reportID || reportID === CONST.REPORT.UNREPORTED_REPORT_ID ? findSelfDMReportID() : reportID;
```
JakubKorytko: this component is always mounted (via GlobalModals), so `findSelfDMReportID()` runs on every render even when idle (no signal, no `reportID`), right? It is cached after the first resolve, but we could skip the call until there is actually a pending/active transaction, e.g. only compute `hostReportID` when `transactionID` is set

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const [reportActions] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${hostReportID}`);
+    // Only IOU/expense/invoice reports have the expense-report RHP that "View" opens. A tracked/unreported
+    // (self-DM) expense lives in a chat, so it has no such report - leaving iouReportID undefined makes
+    // navigateToCreatedExpense open the transaction thread directly instead of a super-wide report RHP it lacks.
+    const iouReport = isMoneyRequestReport(report) || isInvoiceReport(report) ? report : undefined;
+    const iouReportID = iouReport?.reportID;
+
+    useEffect(() => {
```
JakubKorytko, line 97: this effect picks the latest pending signal and then locks (once `active` is set the guard at the top short-circuits). If a batch like multi-scan writes its signals across separate renders and the first transaction's optimistic data lands before the later signals arrive, we would show and lock the growl on the first expense, then show a second growl for the rest after dismiss, instead of one growl for the last. Worth confirming the batch always writes all IDs synchronously so the "for the last expense of the batch" test holds

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,134 @@
...
+        if (active) {
+            return;
+        }
+        const pendingTransactionIDs = Object.keys(signal ?? {});
+        const latestTransactionID = pendingTransactionIDs.at(-1);
```
JakubKorytko, line 104: `.at(-1)` relies on object key insertion order to pick the last-created transaction. It works here because transaction IDs are large enough not to be treated as array-index keys, but it is a bit fragile. A short comment noting that assumption (or tracking order explicitly) would make the intent clearer.

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,134 @@
...
+        }
+        // Wait for the created expense's optimistic data to land before acting. The create's API.write is
+        // usually deferred (the deferred-for-search pattern) until Search's onLayout flushes it, and the whole
+        // optimistic dataset (transaction + IOU report/action + thread) is applied atomically - so the
+        // transaction appearing in Onyx means "View" can resolve AND its report is known for the check below.
+        if (!transaction) {
+            return;
+        }
+
+        mergeExpenseAddedGrowlTransactionIDs(Object.fromEntries(pendingTransactionIDs.map((id) => [id, null])));
```
JakubKorytko, line 119: this is the only path that clears the signal, and it only runs once the transaction has landed. Two edge cases worth a thought: if the created transaction is rolled back and never lands, its entry stays in `EXPENSE_ADDED_GROWL_TRANSACTION_IDS` forever (harmless, but it lingers); and if the app is force-quit in the window between writing the signal and consuming it, the key persists, so on next launch we could surface an "Expense added" growl for a previous-session expense. The old highlight approach also cleared the key when leaving the search type, whereas now the always-mounted component is the only consumer. Maybe worth an expiry or give-up path, e.g. dropping entries whose transaction is missing on mount?

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,134 @@
...
+        }
+        if (!threadReportID) {
+            Log.warn('[ExpenseAddedGrowl] Unable to resolve transaction thread reportID on View press.');
+            return;
+        }
+        navigateToCreatedExpense({threadReportID, transactionID: active.transactionID, iouReportID});
+    };
+
+    return (
+        <GrowlNotificationContent
```
JakubKorytko, line 166: GrowlNotificationContent's own comment says the parent remounts it per show, but there is no `key` here. Today it only remounts because `active` briefly becomes null between growls. If it ever went growl to growl directly, the mount-only slide-in/auto-dismiss timer would not re-arm. Adding `key={active.nonce}` would make that contract explicit and robust

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -70,6 +140,9 @@ function navigateAfterExpenseCreate({
 
     const type = isInvoice ? CONST.SEARCH.DATA_TYPES.INVOICE : CONST.SEARCH.DATA_TYPES.EXPENSE;
 
+    // Signal that an expense was added so we can surface the "Expense added" growl.
+    mergeExpenseAddedGrowlTransactionIDs({[transactionID]: type});
```
JakubKorytko: the IOU actions (requestMoney, trackExpense, sendInvoice, etc.) already call `signalExpenseAddedGrowl` for the same transaction, so on the Search-tab path the same signal is merged twice. It is idempotent, but having two sources emit the growl signal is easy to lose track of. Could this live in a single place?

src/components/Search/SearchList/ListItem/ChatListItem.tsx
```diff
@@ -45,7 +45,7 @@ function ChatListItem<TItem extends ListItem>({
     const {isSelected} = useRowSelection(item.keyForList);
```
JakubKorytko, line 44: now that the post-create highlight is removed, `shouldHighlight` is hardcoded to `false`, which makes the `useAnimatedHighlightStyle` call effectively dead (same in TransactionGroupListItem). Could we drop the hook usage rather than pass a constant `false`, or is it kept on purpose for a future highlight source?

JakubKorytko (commented):
Swapping the imperative per-flow growl calls for a single declarative `ExpenseAddedGrowl` component that subscribes to Onyx is a solid improvement over the old approach! the comments are mostly nits. The batch-ordering one and the signal-lifecycle one are the two I would actually confirm before merge

src/components/Search/index.tsx
mhawryluk: yes, definitely, but I'll rename it after the review, so the diff is easier to read. github shows the files as deleted and newly created after rename

src/components/ExpenseAddedGrowl.tsx
mhawryluk: in multiple-transaction creation flows we signal just one transaction, the last one, so it's fine, at least for now

src/components/ExpenseAddedGrowl.tsx
mhawryluk: fair, I added the comment

src/components/ExpenseAddedGrowl.tsx
mhawryluk: > if the created transaction is rolled back and never lands, its entry stays in EXPENSE_ADDED_GROWL_TRANSACTION_IDS forever (harmless, but it lingers)

it will be cleared once we successfully create a new expense. and before that, since it is harmless, I don't think it's worth complicating the code to remove it

> if the app is force-quit in the window between writing the signal and consuming it, the key persists, so on next launch we could surface an "Expense added" growl for a previous-session expense

I added clearing `ONYXKEYS.EXPENSE_ADDED_GROWL_TRANSACTION_IDS` when we initialize the session

src/components/ExpenseAddedGrowl.tsx
mhawryluk: done!

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
mhawryluk: removed this line, I think indeed all the callers invoke it already

src/components/Search/SearchList/ListItem/ChatListItem.tsx
mhawryluk: I kept it because it applies styles also for the non-highlighted state, it was the easiest way to ensure the rows still look the same

mhawryluk: but I can replace it with inline styles for a tiny performance gain

mhawryluk: replaced

JmillsExpensify (dismissed):
BIg fan of this. LGTM

src/hooks/useSearchAutoRefetch.ts
```diff
@@ -36,9 +31,11 @@ type UseSearchHighlightAndScroll = {
 };
 
 /**
- * Hook used to trigger a search when a new transaction or report action is added and handle highlighting and scrolling.
+ * Hook used to trigger a search when a new transaction or report action is added, so the Search snapshot
+ * reflects freshly-created entries. Also returns the newly-added transactions, which the grouped views use to
+ * refetch each expanded group's child snapshot.
  */
-function useSearchHighlightAndScroll({
+function useSearchAutoRefetch({
```
situchan, line 57: File name doesn't match renamed function

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const [active, setActive] = useState<ActiveGrowl | null>(null);
+    const nonceRef = useRef(0);
+
+    const {translate} = useLocalize();
+    const currentUserPersonalDetails = useCurrentUserPersonalDetails();
+    const [signal] = useOnyx(ONYXKEYS.EXPENSE_ADDED_GROWL_TRANSACTION_IDS);
+    const [betas] = useOnyx(ONYXKEYS.BETAS);
+    const [introSelected] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED);
+    const transactionID = active?.transactionID ?? Object.keys(signal ?? {}).at(-1);
+    const [transaction] = useOnyx(`${ONYXKEYS.COLLECTION.TRANSACTION}${transactionID}`);
```
situchan: This can subscribe to `...undefined` key when idle.
When there's no active growl and no pending signal, this subscribes to `transactions_undefined`.
Harmless functionally, but this component is mounted globally in GlobalModals the entire app lifetime, so it holds three dead Onyx connections at all times.

src/hooks/useSearchAutoRefetch.ts
mhawryluk: indeed, was waiting for after the review to make the diff more readable. I renamed this file, along with the test file

src/components/ExpenseAddedGrowl.tsx
mhawryluk: agreed. I separated out an `ExpenseAddedGrowlContent` component (housing the onyx subscriptions), so that it's mounted only when `transactionID` is defined

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -33,12 +37,77 @@ function getNavigateAfterCreateSearchNavigatorState() {
...
+ * Navigates to a just-created expense, choosing the destination from the surface the user
+ * is currently looking at (they may have switched tabs while the growl was up)
+ * - Spend tab: the transaction thread RHP within Spend (report shown underneath via the wide RHP)
+ * - Inbox tab, narrow layout: the transaction thread as a full report view
+ * - Inbox tab, wide layout, with an expense report: the expense report - a multi-transaction report opens super
+ *   wide with the specific thread RHP stacked on top; a single-transaction report collapses to the thread itself
+ * - Inbox tab, wide layout, tracked/unreported expense (no expense report): the transaction thread RHP directly
+ */
+function navigateToCreatedExpense({threadReportID, transactionID, iouReportID}: NavigateToCreatedExpenseParams) {
+    const backTo = Navigation.getActiveRoute();
```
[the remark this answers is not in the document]
mhawryluk: makes sense. done

src/GlobalModals.tsx
```diff
@@ -50,6 +51,7 @@ function GlobalModals() {
     return (
         <>
             <GrowlNotification ref={growlRef} />
+            <ExpenseAddedGrowl />
```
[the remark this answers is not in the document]
mhawryluk: it's technically true, but in practice this will not happen currently - we very rarely show the imperative growl errors, and not when creating new expenses. and I don't think it's worth adding synchronization of the two growl mechanisms at this moment, I think this can be handled if we do end up using growls in more scenarios, which would cause the overlap. right now it would complicate the code and fix an issue that is not present and might never be

situchan:
This is edge case but growl doesn't disappear after logout. When click View, nothing happens. Maybe not bug


https://github.com/user-attachments/assets/3a222d7d-d033-496a-9c61-b7655028e7bb

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
```diff
@@ -33,12 +50,82 @@ function getNavigateAfterCreateSearchNavigatorState() {
...
+    const openOnInbox = isReportTopmostSplitNavigator() && !isSearchTopmostFullScreenRoute();
+
+    // When a report/expense is already open in the RHP the app's convention is to replace it rather than stack a second
+    // report RHP on top of it.
+    const forceReplace = isReportOpenInRHP(navigationRef.getRootState());
+    const currentRouteParams = navigationRef.current?.getCurrentRoute()?.params;
+    const currentRouteBackTo =
+        typeof currentRouteParams === 'object' && currentRouteParams !== null && 'backTo' in currentRouteParams && typeof currentRouteParams.backTo === 'string'
+            ? currentRouteParams.backTo
+            : undefined;
```
situchan: A four-clause type guard inline in a navigation function is hard to scan and likely duplicates existing route-param plumbing.
Consider extracting into a named `getCurrentRouteBackTo()` helper (or reuse an existing one) so the branch logic below reads cleanly.

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,180 @@
...
+ * Reads the candidate transaction and its report/actions and renders the growl. Split out from
+ * `ExpenseAddedGrowl` so these Onyx connections only exist while there is a pending signal or a visible growl -
+ * the outer component is mounted for the whole app lifetime and would otherwise hold them permanently while idle.
+ */
+function ExpenseAddedGrowlContent({transactionID, signal, active, setActive}: ExpenseAddedGrowlContentProps) {
+    const nonceRef = useRef(0);
+
+    const {translate} = useLocalize();
+    const currentUserPersonalDetails = useCurrentUserPersonalDetails();
+    const [personalDetails] = useOnyx(ONYXKEYS.PERSONAL_DETAILS_LIST);
```
situchan, line 73: The app already exposes `usePersonalDetails()` from `OnyxListItemProvider` (used two files over in `Search/index.tsx`).
Bounded by the short inner-mount lifecycle, so cost is small, but the context read is the idiomatic, zero-subscription choice.

src/components/Search/SearchList/ListItem/GroupHeader.tsx
```diff
@@ -409,7 +401,7 @@ function GroupHeader({
                 ]}
                 wrapperStyle={[
                     styles.mh5,
-                    animatedHighlightStyle,
+                    {backgroundColor: isItemSelected ? theme.activeComponentBG : theme.highlightBG},
```
situchan: The removed `useAnimatedHighlightStyle({backgroundColor: isSelected ? active : highlight})` centralized one decision; it's now inlined across ~6 Search list items:

```
{backgroundColor: isItemSelected ? theme.activeComponentBG : theme.highlightBG}          // GroupHeader, TransactionGroupListItem
{backgroundColor: isSelected ? theme.activeComponentBG : theme.highlightBG}              // TransactionListItemWide, GroupChildrenContainer
{backgroundColor: isSelected ? theme.activeComponentBG : theme.highlightBG, borderRadius: 0}   // TransactionListItemNarrow
{backgroundColor: isSelected ? theme.activeComponentBG : theme.highlightBG, ...(!isLargeScreenWidth && {borderRadius: 0})}  // ExpenseReportListItem
```

A small `StyleUtils.getSearchRowBackgroundStyle(isSelected, theme, {borderRadius?})` would restore the single source of truth and read better than six literal objects.

situchan (commented):
A few NAB comments:

config/eslint/eslint.seatbelt.tsv
```diff
@@ -485,8 +485,8 @@
 "../../src/hooks/useRestoreWorkspacesTabOnNavigate.ts"	"@typescript-eslint/no-unsafe-type-assertion"	1
 "../../src/hooks/useReviewDuplicatesNavigation.tsx"	"@typescript-eslint/no-unsafe-type-assertion"	1
 "../../src/hooks/useReviewDuplicatesNavigation.tsx"	"react-hooks/set-state-in-effect"	1
+"../../src/hooks/useSearchAutoRefetch.ts"	"@typescript-eslint/no-unsafe-type-assertion"	4
```
situchan: Can we avoid type assertion in the first place?
I think we should try not to increase disable eslint anymore.

src/libs/Navigation/helpers/navigateAfterExpenseCreate.ts
mhawryluk: done! added the `getCurrentRouteBackTo` util

src/components/Search/SearchList/ListItem/GroupHeader.tsx
mhawryluk: yeah, great call

config/eslint/eslint.seatbelt.tsv
mhawryluk: it's not new assertions, we renamed a file. but I looked into it and managed to remove those 4 existing ones

mhawryluk:
> This is edge case but growl doesn't disappear after logout. When click View, nothing happens. Maybe not bug
> 

I made it so we unmount the global growl component when we're logged out, so we don't show it then and it also doesn't appear again after logging back in

situchan:
On Spend > Expenses page, after deleting one expense, the row is removed, followed by weird animation. This doesn't happen on `main`.
Can you please double check on your side and confirm it's caused by this PR? If so, we should fix it.

this branch:

https://github.com/user-attachments/assets/7380380d-0788-4f75-a5dc-285bacb6be4d

main:

https://github.com/user-attachments/assets/71835484-5c50-4e3a-b49d-e0eae42d935c

src/components/ExpenseAddedGrowl.tsx
mhawryluk: I'm not sure, the comment states that we should only use it in lists. though there are existing usages outside of that. but I think it doesn't make a big difference here either way

https://github.com/Expensify/App/blob/cc6fec5f3dfc0dec1650a4fa0fc4a75eb88bb294/src/components/OnyxListItemProvider.tsx#L9-L14

situchan: Agree. Skip it

mhawryluk:
> On Spend > Expenses page, after deleting one expense, the row is removed, followed by weird animation. This doesn't happen on `main`. Can you please double check on your side and confirm it's caused by this PR? If so, we should fix it.

I've just tested on main and it's the same as on this branch, same on staging

https://github.com/user-attachments/assets/f0bae2d4-ee22-4330-84c0-694afcf87553

situchan:
Bug?: New row appears in Spend with no highlight (and no growl)

Repro step:
1. Log into the same account in two devices (A and B).
2. In A, open Spend → Expenses. Leave it visible.
2. In B, create an expense (FAB → Create expense → submit) that will match A's current filter.
3. Watch A as the new row syncs in.
   - Before: the new row flashed a brief highlight animation as it appeared.
   - After (this PR): the row just appears — no highlight, and no growl (the growl signal is local to B).

this branch:

https://github.com/user-attachments/assets/e383e4a9-1904-45fa-9e36-67794facef96

main:


https://github.com/user-attachments/assets/a04809c1-9b46-45e7-a794-103875a2e1fa

mhawryluk:
> Bug?: New row appears in Spend with no highlight (and no growl)
> 
> Repro step:
> 
> 1. Log into the same account in two devices (A and B).
> 2. In A, open Spend → Expenses. Leave it visible.
> 3. In B, create an expense (FAB → Create expense → submit) that will match A's current filter.
> 4. Watch A as the new row syncs in.
>    
>    * Before: the new row flashed a brief highlight animation as it appeared.
>    * After (this PR): the row just appears — no highlight, and no growl (the growl signal is local to B).
> 

I raised this issue doing implementation and we agreed that it's fine: https://github.com/Expensify/App/issues/91408#issuecomment-5004970700

situchan:
I think this is expected. Please confirm:

No scroll/position cue after creating on Spend

Repro:

1. Open Spend → Expenses with enough rows to scroll, and set the sort so a new expense will not land at the top, e.g. sort by Amount ascending, or oldest-date-first.
2. Scroll down partway through the list.
3. Tap the FAB → Create expense, enter an amount that sorts off-screen (e.g. a large amount with amount-ascending sort), submit.
   - Before: the list scrolled to bring the new row into view and highlighted it.
   - After: an "Expense added" growl appears; the list stays put. The only way to reach the new row is the growl's View — the list gives no indication of where it landed.

situchan (approved):
I performed full regression test as many cases as possible.
Hopefully no major blockers will be reported by staging QA

mhawryluk:
> I think this is expected. Please confirm:
> 
> No scroll/position cue after creating on Spend
> 
> Repro:
> 
> 1. Open Spend → Expenses with enough rows to scroll, and set the sort so a new expense will not land at the top, e.g. sort by Amount ascending, or oldest-date-first.
> 2. Scroll down partway through the list.
> 3. Tap the FAB → Create expense, enter an amount that sorts off-screen (e.g. a large amount with amount-ascending sort), submit.
>    
>    * Before: the list scrolled to bring the new row into view and highlighted it.
>    * After: an "Expense added" growl appears; the list stays put. The only way to reach the new row is the growl's View — the list gives no indication of where it landed.

yes, I believe this is expected. this issue came about because the highlight and scroll mechanism was found unreliable, so we intentionally remove it (both highlight and scroll) and introduce the growls instead

mountiny:
conflicts

mountiny:
Tested on ios and web and looks fine

mountiny:
Asked Design team for a look https://expensify.slack.com/archives/C03TME82F/p1787219889742919

shawnborton:
Quick test and it feels good to me 👍

tests/unit/ExpenseAddedGrowlTest.tsx
mountiny, line 1: Can you make sure to follow the Given, When, Then comments structure?

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const [report] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT}${reportID}`);
+    // A tracked/unreported expense's transaction.reportID is UNREPORTED_REPORT_ID, and its IOU/track action lives
+    // on the self-DM report - not on transaction.reportID. Read actions from the report that actually hosts the
+    // action so "View" resolves the real transaction thread (the action's childReportID) instead of fabricating
+    // a mismatched optimistic one.
+    const isUnreportedExpense = !reportID || reportID === CONST.REPORT.UNREPORTED_REPORT_ID;
+    const selfDMReportID = isUnreportedExpense ? findSelfDMReportID() : undefined;
+    const hostReportID = isUnreportedExpense ? selfDMReportID : reportID;
+    const [reportActions] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${hostReportID}`);
+    // Only IOU/expense/invoice reports have the expense-report RHP that "View" opens. A tracked/unreported
```
mountiny:
```suggestion

    // Only IOU/expense/invoice reports have the expense-report RHP that "View" opens. A tracked/unreported
```

src/components/ExpenseAddedGrowl.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    const {translate} = useLocalize();
+    const currentUserPersonalDetails = useCurrentUserPersonalDetails();
+    const [personalDetails] = useOnyx(ONYXKEYS.PERSONAL_DETAILS_LIST);
+    const [betas] = useOnyx(ONYXKEYS.BETAS);
+    const [introSelected] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED);
+    const [transaction] = useOnyx(`${ONYXKEYS.COLLECTION.TRANSACTION}${transactionID}`);
+    const reportID = transaction?.reportID;
+    const [report] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT}${reportID}`);
+    // A tracked/unreported expense's transaction.reportID is UNREPORTED_REPORT_ID, and its IOU/track action lives
```
mountiny:
```suggestion

    // A tracked/unreported expense's transaction.reportID is UNREPORTED_REPORT_ID, and its IOU/track action lives
```

mountiny (changes requested):
Can you please update the test comments and also comments throughout the code, I think rarely they should require more than 2 lines and if its longer than that I think it suggests over complicating something, using plain english and explaining why we are doing what we are doing should be good

tests/unit/ExpenseAddedGrowlTest.tsx
mhawryluk: sure!

mhawryluk:
@mountiny I updated the comments ✅

mountiny (approved):
thanks for addressing that

Issue #94739, opened 2026-06-26
Systematically reduce `@typescript-eslint/no-unsafe-type-assertion` violations


blimpich:
If you run `npm run eslint-report` from App you can see our repo's eslint violations. The _vast_ majority of these violations come from `@typescript-eslint/no-unsafe-type-assertion`.

<img width="1254" height="559" alt="Image" src="https://github.com/user-attachments/assets/c2c7e4eb-d00f-4777-bfe9-729064d47be3" />

<img width="1216" height="262" alt="Image" src="https://github.com/user-attachments/assets/62dc66fa-67be-415d-b73f-e7b4eb8f9494" />

About half the violations are in test suites vs production files:

```
┌────────────┬───────┬────────────┐
│            │ Files │ Violations │
├────────────┼───────┼────────────┤
│ Test files │ 485   │ 3,380      │
├────────────┼───────┼────────────┤
│ Prod files │ 861   │ 2,212      │
├────────────┼───────┼────────────┤
│ Total      │ 1,346 │  5,592     │
└────────────┴───────┴────────────┘
```

The ease of fixing these violations varies. Some are extremely simple and others are complex. We fixed some of the worst offending test suites in these two PRs ([1](https://github.com/Expensify/App/pull/94623), [2](https://github.com/Expensify/App/pull/92979)) by beginning to use a `PartialDeep` [helper method](https://github.com/Expensify/App/blob/main/tests/utils/createMock.ts) for creating mock objects, but this is only for test files – and many violations still remain. The job of this issue is to get these number of violations down 📉

Changing one type can have cascading effects on the rest of the codebase. Careful use of [Type Narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) will be needed, and business logic will have to be preserved.

However, we can get good bang-for-our-buck here. A small number of files account for a disproportionate % of violations:

```
┌─────────┬───────┬────────────┬────────────┬────────────┐
│  Scope  │ Files │ % of files │ Violations │ % of total │
├─────────┼───────┼────────────┼────────────┼────────────┤
│ Top 5   │ 5     │ 0.4%       │ 547        │ 10%        │
├─────────┼───────┼────────────┼────────────┼────────────┤
│ Top 10  │ 10    │ 0.7%       │ 872        │ 16%        │
├─────────┼───────┼────────────┼────────────┼────────────┤
│ Top 20  │ 20    │ 1.5%       │ 1,344      │ 24%        │
[...]

PR #96513, opened 2026-07-20
[No QA] Remove unsafe type assertions from PolicyTest.ts tests

KJ21-ENG:
### Explanation of Change

Removed `@typescript-eslint/no-unsafe-type-assertion` violations from `tests/actions/PolicyTest.ts` as part of the cleanup for Expensify/App issue Expensify/App#94739.

What changed:

* Replaced unsafe test assertions with fail-fast structural narrowing, reused ObjectUtils.isRecord and TestHelper.createGlobalFetchMock, and removed redundant target-local helpers.

Why this approach is safe:

* Production behavior is unchanged; protected test cases, fixtures, assertion identity, and failure-sensitive validation are preserved while repository-standard utilities replace local duplicates.

Reviewer focus:

1. Confirm the existing ObjectUtils.isRecord and TestHelper.createGlobalFetchMock utilities preserve the original Policy action test behavior without target-local helper duplication.

Safety checks:

* The Seatbelt count decreased from 74 to 0, all 208 focused tests passed, independent review cleared the exact revision, and trusted Fork CI passed on the publication head.

### Fixed Issues

$ [https://github.com/Expensify/App/issues/94739](<https://github.com/Expensify/App/issues/94739>)<br>PROPOSAL: [https://github.com/Expensify/App/issues/94739#issuecomment-4882049995](<https://github.com/Expensify/App/issues/94739#issuecomment-4882049995>)

### Count change

Current baseline source:

* `config/eslint/eslint.seatbelt.tsv`

Before:

* `tests/actions/PolicyTest.ts`: 74 violations

After:

* `tests/actions/PolicyTest.ts`: 0 violations

Net reduction:

* 74 fewer `@typescript-eslint/no-unsafe-type-assertion` violations.

TSV evidence:

* Writable lint removed the target row when the count reached zero, or updated it to the exact remaining count.
* The generated TSV was restored byte-for-byte and is intentionally not included in this PR because Expensify/App post-merge automation tightens the baseline on `main`.

### Tests

1. Run:

   ```
   npm run lint -- --show-warnings tests/actions/PolicyTest.ts
   ```

2. Run:

   ```
   CI=true npm run lint -- --no-cache --show-warnings tests/actions/PolicyTest.ts
   ```

3. Run:

   ```
   npm run test -- tests/actions/PolicyTest.ts
   ```

4. Verification result: Focused PolicyTest verification passed all 208 tests, focused lint passed with zero target-rule violations, and trusted Fork CI passed.

### Offline tests

Not applicable because this is a test-only type-safety cleanup with no production network behavior.

### QA Steps

Not applicable because production behavior and user-interface code are unchanged.

### PR Author Checklist

  - [X] I tested this PR with a [High Traffic account](<https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts>) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [X] I included screenshots or videos for tests on [all platforms](<https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms>)
  - [X] Android: Native
  - [X] Android: mWeb Chrome
  - [X] iOS: Native
  - [X] iOS: mWeb Safari
  - [X] MacOS: Chrome / Safari
- [X] I followed proper code patterns (see [Reviewing the code](<https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code>))
- [X] I followed the guidelines as stated in the [Review Guidelines](<https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md>)
  - [X] The style can't be created with an existing [StyleUtils](<https://github.com/Expensify/App/blob/main/src/styles/utils/index.ts>) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG)`)
- [X] I added [unit tests](<https://github.com/Expensify/App/blob/main/tests/README.md>) for any new feature or bug fix in this PR to help automatically prevent regressions in this user flow.

### Screenshots/Videos

Android: Native

Not applicable on Android Native, Android mWeb Chrome, iOS Native, iOS mWeb Safari, and MacOS Chrome or Safari because only a Jest test implementation changed.

Not applicable because this change has no user-interface or visual output.

Android: mWeb Chrome

Not applicable on Android Native, Android mWeb Chrome, iOS Native, iOS mWeb Safari, and MacOS Chrome or Safari because only a Jest test implementation changed.

Not applicable because this change has no user-interface or visual output.

iOS: Native

Not applicable on Android Native, Android mWeb Chrome, iOS Native, iOS mWeb Safari, and MacOS Chrome or Safari because only a Jest test implementation changed.

Not applicable because this change has no user-interface or visual output.

iOS: mWeb Safari

Not applicable on Android Native, Android mWeb Chrome, iOS Native, iOS mWeb Safari, and MacOS Chrome or Safari because only a Jest test implementation changed.

Not applicable because this change has no user-interface or visual output.

MacOS: Chrome / Safari

Not applicable on Android Native, Android mWeb Chrome, iOS Native, iOS mWeb Safari, and MacOS Chrome or Safari because only a Jest test implementation changed.

Not applicable because this change has no user-interface or visual output.

### DIFF
--- a/tests/actions/PolicyTest.ts
+++ b/tests/actions/PolicyTest.ts
@@ -1,11 +1,14 @@
+import * as APIModule from '@libs/API';
 import {WRITE_COMMANDS} from '@libs/API/types';
 import GoogleTagManager from '@libs/GoogleTagManager';
+import {isRecord} from '@libs/ObjectUtils';
 import * as PersonalDetailsUtils from '@libs/PersonalDetailsUtils';
 import {isPolicyPayer} from '@libs/PolicyUtils';
 import * as ReportUtils from '@libs/ReportUtils';
 
 import CONST from '@src/CONST';
 import IntlStore from '@src/languages/IntlStore';
+import type {TranslationParameters, TranslationPaths} from '@src/languages/types';
 import OnyxUpdateManager from '@src/libs/actions/OnyxUpdateManager';
 import {askToJoinPolicy, joinAccessiblePolicy} from '@src/libs/actions/Policy/Member';
 import * as Policy from '@src/libs/actions/Policy/Policy';
@@ -29,8 +32,6 @@ import type {OnyxCollection, OnyxEntry} from 'react-native-onyx';
 import {Str} from 'expensify-common';
 import Onyx from 'react-native-onyx';
 
-import type {MockFetch} from '../utils/TestHelper';
-
 import createRandomPolicy from '../utils/collections/policies';
 import {createRandomReport} from '../utils/collections/reports';
 import createRandomTransaction from '../utils/collections/transaction';
@@ -53,6 +54,90 @@ const TEST_PHONE_NUMBER = '1234567890';
 const TEST_NON_PUBLIC_DOMAIN_EMAIL = 'esh@example.com';
 const TEST_SMS_DOMAIN_EMAIL = 'esh@expensify.sms';
 
+type UnknownRecord = Record<string, unknown>;
+type GuidedSetupItem = {task?: string; completedTaskReportActionID?: string};
+
+function requireRecord(value: unknown, label = 'value'): UnknownRecord {
+    if (!isRecord(value)) {
+        throw new Error(`Expected ${label} to be an object`);
+    }
+    return value;
+}
+
+function readProperty(value: unknown, key: string): unknown {
+    return requireRecord(value, `container for ${key}`)[key];
+}
+
+function requireStringProperty(value: unknown, key: string): string {
+    const property = readProperty(value, key);
+    if (typeof property !== 'string') {
+        throw new Error(`Expected ${key} to be a string`);
+    }
+    return property;
+}
+
+function getOptionalNumberProperty(value: unknown, key: string): number | undefined {
+    const property = readProperty(value, key);
+    if (property !== undefined && typeof property !== 'number') {
+        throw new Error(`Expected ${key} to be a number when present`);
+    }
+    return property;
+}
+
+function requireRecordArrayProperty(value: unknown, key: string): UnknownRecord[] {
+    const property = readProperty(value, key);
+    if (!Array.isArray(property) || !property.every(isRecord)) {
+        throw new Error(`Expected ${key} to be an array of objects`);
+    }
+    return property;
+}
+
+function requireCallArgument(call: unknown, index: number): unknown {
+    if (!Array.isArray(call) || !(index in call)) {
+        throw new Error(`Expected API call argument ${index}`);
+    }
+    return call.at(index);
+}
+
+function parseJSONRecord(value: unknown): UnknownRecord {
+    if (typeof value !== 'string') {
+        throw new Error('Expected JSON payload to be a string');
+    }
+    const parsed: unknown = JSON.parse(value);
+    if (!isRecord(parsed)) {
+        throw new Error('Expected JSON object');
+    }
+    return parsed;
+}
+
+function isGuidedSetupItem(value: unknown): value is GuidedSetupItem {
+    return (
+        isRecord(value) &&
+        (value.task === undefined || typeof value.task === 'string') &&
+        (value.completedTaskReportActionID === undefined || typeof value.completedTaskReportActionID === 'string')
+    );
+}
+
+function parseGuidedSetupData(value: unknown): GuidedSetupItem[] {
+    if (typeof value !== 'string') {
+        throw new Error('Expected guidedSetupData to be a string');
+    }
+    const parsed: unknown = JSON.parse(value);
+    if (!Array.isArray(parsed) || !parsed.every(isGuidedSetupItem)) {
+        throw new Error('Expected guidedSetupData to contain onboarding task objects');
+    }
+    return parsed;
+}
+
+function anyStringMatcher(): unknown {
+    return expect.any(String);
+}
+
+function mockTranslate<TPath extends TranslationPaths>(key: TPath, ...parameters: TranslationParameters<TPath>): string;
+function mockTranslate(key: TranslationPaths): string {
+    return key;
+}
+
 jest.mock('@libs/GoogleTagManager');
 
 OnyxUpdateManager();
@@ -63,10 +148,10 @@ describe('actions/Policy', () => {
         });
     });
 
-    let mockFetch: MockFetch;
+    let mockFetch: ReturnType<typeof TestHelper.createGlobalFetchMock>;
     beforeEach(() => {
-        global.fetch = TestHelper.getGlobalFetchMock();
-        mockFetch = fetch as MockFetch;
+        mockFetch = TestHelper.createGlobalFetchMock();
+        global.fetch = mockFetch;
         IntlStore.load(CONST.LOCALES.EN);
         jest.clearAllMocks();
         return Onyx.clear().then(waitForBatchedUpdates);
@@ -78,7 +163,7 @@ describe('actions/Policy', () => {
         });
 
         it('creates a new workspace', async () => {
-            (fetch as MockFetch)?.pause?.();
+            mockFetch?.pause?.();
             await Onyx.set(ONYXKEYS.SESSION, {email: ESH_EMAIL, accountID: ESH_ACCOUNT_ID});
             const fakePolicy = createRandomPolicy(0, CONST.POLICY.TYPE.PERSONAL);
             await Onyx.set(`${ONYXKEYS.COLLECTION.POLICY}${fakePolicy.id}`, fakePolicy);
@@ -242,7 +327,7 @@ describe('actions/Policy', () => {
             expect(manageTeamTasksCount).toBe(expectedManageTeamDefaultTasksCount);
 
             // Check for success data
-            (fetch as MockFetch)?.resume?.();
+            mockFetch?.resume?.();
             await waitForBatchedUpdates();
 
             policy = await new Promise((resolve) => {
@@ -297,7 +382,7 @@ describe('actions/Policy', () => {
         });
 
         it('duplicate workspace', async () => {
-            (fetch as MockFetch)?.pause?.();
+            mockFetch?.pause?.();
             await Onyx.set(ONYXKEYS.SESSION, {email: ESH_EMAIL, accountID: ESH_ACCOUNT_ID});
             const fakePolicy = {
                 ...createRandomPolicy(10, CONST.POLICY.TYPE.PERSONAL),
@@ -445,7 +530,7 @@ describe('actions/Policy', () => {
             expect(reportActionsOfTypeCreatedCount).toBe(expectedReportActionsOfTypeCreatedCount);
 
             // Check for success data
-            (fetch as MockFetch)?.resume?.();
+            mockFetch?.resume?.();
             await waitForBatchedUpdates();
 
             policy = await new Promise((resolve) => {
@@ -1035,12 +1120,12 @@ describe('actions/Policy', () => {
         });
 
         it('create a new workspace fails will reset hasCompletedGuidedSetupFlow to the correct value', async () => {
-            (fetch as MockFetch)?.pause?.();
+            mockFetch?.pause?.();
             await Onyx.set(ONYXKEYS.SESSION, {email: ESH_EMAIL, accountID: ESH_ACCOUNT_ID});
             await Onyx.set(ONYXKEYS.NVP_ONBOARDING, {hasCompletedGuidedSetupFlow: true, chatReportID: '12345'});
             await waitForBatchedUpdates();
 
-            (fetch as MockFetch)?.fail?.();
+            mockFetch?.fail?.();
             Policy.createWorkspace({
                 policyOwnerEmail: ESH_EMAIL,
                 makeMeAdmin: true,
@@ -1058,7 +1143,7 @@ describe('actions/Policy', () => {
             });
             await waitForBatchedUpdates();
 
-            (fetch as MockFetch)?.resume?.();
+            mockFetch?.resume?.();
             await waitForBatchedUpdates();
 
             let onboarding: OnyxEntry<Onboarding>;
@@ -1327,7 +1412,7 @@ describe('actions/Policy', () => {
             await Onyx.set(ONYXKEYS.SESSION, {email: ESH_EMAIL, accountID: ESH_ACCOUNT_ID});
             await waitForBatchedUpdates();
 
-            const apiWriteSpy = jest.spyOn(require('@libs/API'), 'write').mockImplementation(() => Promise.resolve());
+            const apiWriteSpy = jest.spyOn(APIModule, 'write').mockImplementation(() => Promise.resolve());
             const policyID = Policy.generatePolicyID();
 
             // When creating a workspace with distance rates feature enabled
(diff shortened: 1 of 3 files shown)


tests/actions/PolicyTest.ts
```diff
@@ -53,6 +54,90 @@ const TEST_PHONE_NUMBER = '1234567890';
...
 
+type UnknownRecord = Record<string, unknown>;
+type GuidedSetupItem = {task?: string; completedTaskReportActionID?: string};
+
+function requireRecord(value: unknown, label = 'value'): UnknownRecord {
+    if (!isRecord(value)) {
+        throw new Error(`Expected ${label} to be an object`);
+    }
+    return value;
+}
```
blimpich: This has been used in multiple of your PRs I've reviewed, can we please extract this out?

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +54,90 @@ const TEST_PHONE_NUMBER = '1234567890';
 const TEST_NON_PUBLIC_DOMAIN_EMAIL = 'esh@example.com';
 const TEST_SMS_DOMAIN_EMAIL = 'esh@expensify.sms';
 
+type UnknownRecord = Record<string, unknown>;
+type GuidedSetupItem = {task?: string; completedTaskReportActionID?: string};
```
blimpich: Same here, I've seen this used before. If we have generic type guards that we're going to use over and over they should be declared in one place rather than many

blimpich (dismissed):

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 4ec72b6d831.

I moved requireRecord and the related reusable record-validation helpers into the shared tests/utils/typeGuards.ts utility.

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 4ec72b6d831. 

The generic record-validation helpers are now centralized in tests/utils/typeGuards.ts. Policy-specific and mock-call helpers remain local to avoid expanding the shared API unnecessarily.

ikevin127 (approved):
## Review — Type-safety cleanup (`PolicyTest.ts` + shared `typeGuards.ts`)

Reviewed the full diff and verified the shared helpers against source. This is a strong cleanup and a clear step up from the sibling PRs: the removed assertions are replaced with genuine fail-fast runtime narrowing, and the nastiest casts (`as unknown as ...`) are gone entirely. No skirting: no `as any`, no `as unknown as`, no `@ts-ignore`, no inline rule disables.

### Findings summary

| Severity | Finding | Location | Blocker? |
| --- | --- | --- | --- |
| 🟠 Medium-High | Test changed call-selection logic (`.at(0)` to `.find(command)`), not just types | `should not build next steps ...` (DISABLE_POLICY_APPROVALS) | Confirm intent |
| 🟡 Medium | Predicates now throw on updates with missing/non-string `key` | `leaveWorkspace` / next-step tests | No |
| 🟢 Low | Failure presentation shifts from matcher diff to thrown `Error` | multiple `readProperty` sites | No |
| 🟢 Low | `isObject` (array-permissive) and `isRecord` helpers now coexist in `typeGuards.ts` | `tests/utils/typeGuards.ts` | No |
| 🟢 Low | `anyStringMatcher()` returns `unknown`, hides matcher intent | `leaveWorkspace` archived test | No |

### Confirmed against source

- The new record helpers build on `isRecord` (`src/libs/ObjectUtils.ts:41`), which correctly excludes arrays (`!Array.isArray(value)`). Good choice over the array-permissive `isObject` already in `typeGuards.ts`: it means `requireRecord` / `requireRecordArrayProperty` will not silently accept an array where an object is expected.
- The `mockTranslate` overload replaces `((key: string) => key) as unknown as Parameters<...>['localeTranslate']` and is assignable to `LocalizedTranslate` (`src/libs/actions/Policy/Policy.ts:225`). Removing an `as unknown as` on a function type is exactly the kind of change this issue wants.
- `body instanceof FormData ? Object.fromEntries(body.entries()) : {}` is type-safe: `readProperty` returns `unknown`, the `instanceof` narrows to the global `FormData`, and `.entries()` exists on both the DOM type and the TestHelper mock (`tests/utils/TestHelper.ts:46`). This removes the `as unknown as Iterable<[string, string]>` cast cleanly.
- `import * as APIModule from '@libs/API'` plus `jest.spyOn(APIModule, 'write')` removes the untyped `jest.spyOn(require('@libs/API'), 'write')`. The `require('@libs/NextStepUtils')` spies were intentionally left (not type assertions, out of scope), which is fine.

##

### 🟠 1. One test changed its call-selection logic, not just its types (please confirm intentional)

In the `DISABLE_POLICY_APPROVALS` test the inspected call changed from positional to command-matched:

```ts
// before
const writeOptions = apiWriteSpy.mock.calls.at(0)?.at(2) as {...};
// after
const apiCallArgs = apiWriteSpy.mock.calls.find((call) => call.at(0) === WRITE_COMMANDS.DISABLE_POLICY_APPROVALS);
const writeOptions = requireRecord(requireCallArgument(apiCallArgs, 2), 'DISABLE_POLICY_APPROVALS write options');
```

This is more robust (targets the right command instead of assuming it is the first `write`), so I am not asking you to revert it. But it is a behavior change beyond cast removal: if the action emits multiple `write` calls, the test now asserts against a different one than before. Please confirm this was deliberate and that the DISABLE_POLICY_APPROVALS call is the one intended. Everywhere else you kept `.at(0)` / `.find(CREATE_WORKSPACE)` semantics identical, so this is the one spot that stands out.

### 🟡 2. Predicates got stricter about `key` and will now throw on malformed updates

Several `.find()` predicates went from tolerant to fail-fast:

```ts
// before: missing key tolerated
.find((update) => (update.key ?? '').startsWith(ONYXKEYS.COLLECTION.REPORT_METADATA))
// after: throws if ANY scanned update lacks a string key
.find((update) => requireStringProperty(update, 'key').startsWith(ONYXKEYS.COLLECTION.REPORT_METADATA))
```

Because `.find` evaluates the predicate against every element until a match, this throws if any earlier `optimisticData` entry has a non-string `key`. Onyx write updates always carry a string `key`, so risk is low, but it is a stricter runtime assumption than the code it replaces. Worth a conscious ack that this can never trip on the real update shape.

### 🟢 3. Failure presentation shifted from assertion-diff to thrown `Error` (informational)

Many assertions previously used optional chaining, for example:

```ts
expect((optimisticOldChatUpdate?.value as {iouReportID?: string | null})?.iouReportID).toBeNull();
```

became:

```ts
expect(readProperty(readProperty(optimisticOldChatUpdate, 'value'), 'iouReportID')).toBeNull();
```

If `optimisticOldChatUpdate` is missing, the old code produced a clean `expect(undefined).toBeNull()` failure; the new code throws inside `readProperty` first. You do guard most of these with a preceding `expect(...).toBeDefined()`, so this is fine and arguably better (fail-fast). Just flagging that the PR wording "assertion identity ... preserved" is true for the passing path, but the failure path now surfaces as a thrown `Error` rather than a matcher diff.

### 🟢 4. Two similar guards now coexist, worth a one-line comment

`tests/utils/typeGuards.ts` now exports both `isObject` (includes arrays) and the `isRecord`-backed helpers (exclude arrays). Both are correct for their call sites, but the divergence is a footgun for the next contributor. Consider a short doc comment on `isObject` noting it is intentionally array-permissive versus the `isRecord` helpers, so nobody "fixes" one to match the other.

### 🟢 5. `anyStringMatcher()` returns `unknown` (cosmetic)

```ts
function anyStringMatcher(): unknown { return expect.any(String); }
```

This cleanly removes `expect.any(String) as unknown as string`, but hides an asymmetric matcher behind an opaque name. Minor, fine to keep. Inlining `expect.any(String)` directly in the object also works if you prefer fewer indirections.

##

@KJ21-ENG 

🟢 **Overall**: LGTM once the author confirms note 1 (the call-selection change) was intentional. This PR removes real casts, including `as unknown as` on a function type, and uses array-safe guards - no type safety was traded away.

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +55,54 @@ const TEST_PHONE_NUMBER = '1234567890';
 const TEST_NON_PUBLIC_DOMAIN_EMAIL = 'esh@example.com';
 const TEST_SMS_DOMAIN_EMAIL = 'esh@expensify.sms';
 
+type GuidedSetupItem = {task?: string; completedTaskReportActionID?: string};
```
blimpich: Why do we need a new type in a test file? Shouldn't this be re-using a type from production? cc: @ikevin127 we should not need to create new types in order to test production files

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +55,54 @@ const TEST_PHONE_NUMBER = '1234567890';
...
+    return parsed;
+}
+
+function isGuidedSetupItem(value: unknown): value is GuidedSetupItem {
+    return (
+        isRecord(value) &&
+        (value.task === undefined || typeof value.task === 'string') &&
+        (value.completedTaskReportActionID === undefined || typeof value.completedTaskReportActionID === 'string')
+    );
+}
```
blimpich: should this not be in the typeGuard.ts file?

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +55,54 @@ const TEST_PHONE_NUMBER = '1234567890';
...
+function parseGuidedSetupData(value: unknown): GuidedSetupItem[] {
+    if (typeof value !== 'string') {
+        throw new Error('Expected guidedSetupData to be a string');
+    }
+    const parsed: unknown = JSON.parse(value);
+    if (!Array.isArray(parsed) || !parsed.every(isGuidedSetupItem)) {
+        throw new Error('Expected guidedSetupData to contain onboarding task objects');
+    }
+    return parsed;
+}
```
blimpich: again should this be in some sort of generic file instead of the file specific test file PolicyTest.ts?

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +55,54 @@ const TEST_PHONE_NUMBER = '1234567890';
...
+
+function anyStringMatcher(): unknown {
+    return expect.any(String);
+}
```
blimpich: Same here, this is extraordinarily generic, why is it in this test file?

blimpich (changes requested):

tests/actions/PolicyTest.ts
ikevin127: Good catch, we should not define a new type here.

`TaskForParameters` is already exported from `@libs/actions/Report`, and `tests/unit/ReportUtilsTest.ts` already uses `Extract<TaskForParameters, {type: 'task'}>` for exactly this guidedSetupData shape (see line ~1075).

Let's reuse the production type instead of `GuidedSetupItem`.

tests/actions/PolicyTest.ts
ikevin127: My take: Once we reuse `TaskForParameters`, most of this guard disappears. It becomes a discriminated-union check on `item.type === 'task'`, same as `ReportUtilsTest.ts` does today.

Whatever narrowing we still need after `JSON.parse` should sit in a shared test util, not in `PolicyTest.ts`.

tests/actions/PolicyTest.ts
ikevin127: **My take**: Same direction, the generic half (`parseJSONRecord`, string to record) is reusable and should live in `tests/utils/typeGuards.ts`. The guided-setup specific parsing can build on top of that plus the production `TaskForParameters` type, so we are not redefining the shape per test file.

tests/actions/PolicyTest.ts
ikevin127: **My take**: Agreed, this is too generic to name here, it is a one-liner, so simplest is to inline `expect.any(String)` at the call site. If the original type friction comes back, we move it to shared test utils rather than keeping it in `PolicyTest.ts`.

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 47776ee0072: the task shape now derives from production TaskForParameters via Extract and Pick.

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 89ae38855ee: reusable JSON parsing is shared in tests/utils/typeGuards.ts; only the test-specific predicate remains local.

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 89ae38855ee: parseJSONRecord and parseJSONArray now live in tests/utils/typeGuards.ts.

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 89ae38855ee: removed the generic matcher and directly verifies that private_isArchived is a string.

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +55,25 @@ const TEST_PHONE_NUMBER = '1234567890';
 const TEST_NON_PUBLIC_DOMAIN_EMAIL = 'esh@example.com';
 const TEST_SMS_DOMAIN_EMAIL = 'esh@expensify.sms';
 
+type GuidedSetupTask = Extract<TaskForParameters, {type: 'task'}>;
```
blimpich: I understand we did this to get the specific type we needed for testing without changing production code, but I think this points out an issue with how we've typed the production code that we shouldn't be working around by creating the type in the test.

**Main point:** the test files exist to test the behavior of the production code and types. If we have to create types in the test file, that is a sign that either we're doing something wrong or the production code lacks the appropriate typing.

So for this case, can we update `src/libs/actions/Report/index.ts` to have the following

```typescript
  type GuidedSetupTask = {
            type: 'task';
            task: string;
            taskReportID: string;
            parentReportID: string;
            parentReportActionID: string;
            assigneeChatReportID?: string;
            createdTaskReportActionID: string;
            completedTaskReportActionID?: string;
            title: string;
            description: string;
  };

  type TaskForParameters =
      | GuidedSetupTask
      | ({
            type: 'message';
        } & TaskMessage);
```

and then import the type here?

tests/actions/PolicyTest.ts
```diff
@@ -53,6 +55,25 @@ const TEST_PHONE_NUMBER = '1234567890';
...
+}
+
+function isGuidedSetupTask(value: unknown): value is GuidedSetupTaskFields {
+    return (
+        isRecord(value) &&
+        value.type === 'task' &&
+        typeof value.task === 'string' &&
+        (value.completedTaskReportActionID === undefined || typeof value.completedTaskReportActionID === 'string')
+    );
+}
```
blimpich: type guard should be in typeGuards.ts

blimpich (changes requested):

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 5cb5e8ec81100abcebac8dd01f43ea0585f7a4e4.
I extracted the anonymous task variant into the exported GuidedSetupTask production type and rebuilt TaskForParameters from it without changing its structure or runtime behavior. PolicyTest.ts now consumes the production type directly.

tests/actions/PolicyTest.ts
KJ21-ENG: Fixed in 5cb5e8ec81100abcebac8dd01f43ea0585f7a4e4.

I moved isGuidedSetupTask into tests/utils/typeGuards.ts and made it validate the complete production shape, including every required string field and both optional fields. PolicyTest.ts now imports the shared guard instead of defining a partial local guard.

blimpich (approved):

PR #9348, opened 2022-06-07
Unnecessary focus on the Add payment method button

allroundexperts:
### Details

The css selector :focus-visible is responsible for showing up the outline. When the selector :focus-visible is applied to an element is largely dependent browser implementation. As always, the problematic browser is Safari. Safari restores focus on an element even if an event is a click. On Chrome and other browser, focus is restored if event type is keypress ie closing modal by keyboard.

### Fixed Issues

$ https://github.com/Expensify/App/issues/9248

### Tests

1. Go to Settings
2. Go to Payments
3. Click on the Add payment method button
4. Click anywhere on the Payments page

Make sure that there is no focus on the Add payment method button ie
image

- [x] Verify that no errors appear in the JS console

#### Contributor (PR Author) Checklist
- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [ ] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] iOS / native
    - [x] Android / native
    - [x] iOS / Safari
    - [x] Android / Chrome
    - [x] MacOS / Chrome
    - [x] MacOS / Desktop
- [x] I verified there are no console errors (if there’s a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [x] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`)
    - [x] I verified that comments were added to code that is not self explanatory
    - [x] I verified that any new or modified comments were clear, correct English, and explained “why” the code was doing something instead of only explaining “what” the code was doing.
    - [ ] I verified any copy / text shown in the product was added in all `src/languages/*` files
[...]

PR #96650, opened 2026-07-21
refactor: MenuItem decomposition into a compound component

war-in:
### Explanation of Change

Decomposes the monolithic `MenuItem` (~135 props) into a composable compound component. Consumers now assemble a row from sub-components (`MenuItem.Row`, `.Icon`, `.Avatar`, `.Content`, `.Title`, `.Description`, `.Trailing`, `.Chevron`, `.BrickRoadIndicator`, etc.), with interaction state (hover/press/focus/disabled) shared via `MenuItemContext`. Moves `MenuItem` into its own directory and migrates `SettingsMenuItem` to the new pattern as the first consumer.

### Fixed Issues

$ https://github.com/Expensify/App/issues/96202
PROPOSAL:

### Tests

1. Open the app and navigate to **Settings** (bottom tab / account menu).
2. Verify the settings menu rows render correctly: leading avatar/icon, title, description, trailing chevron, and any brick-road indicators or badges are aligned and styled as before.
3. Hover over a settings row (web) and verify hover styling applies (icon fill, chevron reveal, row background).
4. Press a settings row and verify it navigates to the expected screen and the pressed styling appears.
5. Verify a row that shows an error/brick-road indicator still displays the red/green dot and any error text.

### Offline tests

N/A

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/components/MenuItem/compound.tsx
```diff
@@ -0,0 +1,117 @@
...
+ *             <MenuItem.BrickRoadIndicator status={CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR} />
+ *             <MenuItem.Chevron />
+ *         </MenuItem.Trailing>
+ *     </MenuItem.Row>
+ *     <MenuItem.Error>{errorText}</MenuItem.Error>
+ * </MenuItem>
+ * <MenuItem.HelperText>{helperText}</MenuItem.HelperText>
+ * ```
+ *
+ * See README.md in this directory for the full design notes and the legacy-prop → composition mapping.
```
Guccio163: I don't see no README.md file, guessing by the '→' just a claude aftermath 🧹

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+                onPress(event);
+            }),
+        )();
+    };
+
+    return (
+        <View onBlur={onBlur}>
+            <Hoverable isFocused={isScreenFocused}>
+                {(isHovered) => (
+                    <PressableWithSecondaryInteraction
```
Guccio163, line 168: onPressIn, onPressOut, disabledStyle, accessible props with MenuItem's props that they are using are missing

src/pages/settings/SettingsMenuItem.tsx
```diff
@@ -49,35 +54,48 @@ function SettingsMenuItem({item, isFocused, keyTitle, isExecuting, isScreenFocus
...
+            disabled={isExecuting}
             focused={isFocused}
             role={CONST.ROLE.TAB}
-            isPaneMenu
+            accessibilityLabel={accessibilityLabel}
+            accessibilityHint={accessibilityHint}
             sentryLabel={item.sentryLabel}
-            iconRight={item.iconRight}
-            shouldShowRightIcon={item.shouldShowRightIcon}
-            shouldIconUseAutoWidthStyle
```
Guccio163, line 79: This prop added styles.wAuto to the icon and Icon doesn't seem to be addressing it neither through prop nor some inline style

src/components/MenuItem/MenuItem.tsx
dariusz-biela, line 1: Moving `MenuItem.tsx` into `MenuItem/` without an `index.ts` breaks every `@components/MenuItem` / `./MenuItem` import (~120 files). Local `tsc` shows 238 errors; CI typecheck, all 8 test shards, and knip are red for the same reason. Please add `src/components/MenuItem/index.ts` re-exporting the legacy default + types, or leave the legacy file in place and move it in a follow-up.

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+type MenuItemRootProps = WithSentryLabel & {
+    /** Sub-components composing the row. Children stack vertically: place a `MenuItem.Row` for the main
+     * line, and full-width blocks (`MenuItem.Label`, `MenuItem.Error`, `MenuItem.Hint`) before/after it. */
+    children: ReactNode;
+
+    /** Reference to the pressable element */
+    ref?: PressableRef | Ref<View>;
+
+    /** Function to fire when the row is pressed */
+    onPress?: (event: GestureResponderEvent | KeyboardEvent) => void | Promise<void>;
```
dariusz-biela: Does this component even have the right not to have an `onPress` event?

dariusz-biela: The AI suggested something like this to me. You might want to think it over.

> A row without `onPress` is still a focusable, hoverable `button` that does nothing. Suggest deriving it: `const isInteractive = interactive ?? !!onPress;`.

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+            }),
+        )();
+    };
+
+    return (
+        <View onBlur={onBlur}>
+            <Hoverable isFocused={isScreenFocused}>
+                {(isHovered) => (
+                    <PressableWithSecondaryInteraction
+                        onPress={shouldCheckActionAllowedOnPress ? callFunctionIfActionIsAllowed(onPressAction, isAnonymousAction) : onPressAction}
```
dariusz-biela: `isAnonymousAction` / `shouldCheckActionAllowedOnPress` and the `@userActions/Session` import are application policy, not row UI. Suggest keeping auth-gating in callers so Root does not start life with 20+ props.

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+     * (horizontal padding + negative margins), replacing the classic `wrapperStyle={styles.sectionMenuItemTopDescription}`. */
+    variant?: MenuItemVariant;
+
+    /**
+     * Any additional styles to apply on the pressable element.
+     */
+    style?: StyleProp<ViewStyle>;
+
+    /** Wrapper styles */
+    wrapperStyle?: StyleProp<AnimatedStyle<ViewStyle>>;
```
dariusz-biela: Are we sure we want to allow styles to be edited via props? 

This may cause problems in the future when updating the component's appearance, due to custom styling of certain components that use this component.

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+    onBlur?: () => void;
+
+    /** Whether the row should be interactive at all */
+    interactive?: boolean;
+
+    /** Should we disable this row? */
+    disabled?: boolean;
+
+    /** Whether the row is focused or active (selected-row styling) */
+    focused?: boolean;
```
dariusz-biela: `focused` here means selected/active-row styling, while `onFocus`/`onBlur` next to it mean real focus. Suggest renaming to `active`/`isActive`.

src/components/MenuItem/compound.tsx
dariusz-biela: `README.md` does not exist in this directory.

src/components/MenuItem/compound.tsx
```diff
@@ -0,0 +1,117 @@
...
+    CopyButton: MenuItemCopyButton,
+
+    /** Error message rendered under the main line (inside the pressable) */
+    Error: MenuItemError,
+
+    /** Hint message rendered under the main line (inside the pressable) */
+    Hint: MenuItemHint,
+
+    /** Non-interactive helper text — place it AFTER the root, outside the pressable */
+    HelperText: MenuItemHelperText,
```
dariusz-biela: If this component is to be placed outside the Root, then perhaps we shouldn't attach it to the Root component.

dariusz-biela: I don't have a strong opinion on this

src/components/MenuItem/layout/MenuItemTrailing.tsx
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * The right-side cell of a `MenuItem.Row`. A horizontal cluster for indicators and actions —
+ * the consumer decides which to include and in what order.
+ */
+function MenuItemTrailing({children, style}: MenuItemTrailingProps) {
+    const styles = useThemeStyles();
+    const StyleUtils = useStyleUtils();
+    const {isCompact} = useMenuItemState();
+
+    return <View style={[styles.flexRow, styles.mlAuto, styles.alignItemsCenter, styles.gap2, StyleUtils.getMenuItemTextContainerStyle(isCompact), style]}>{children}</View>;
```
Guccio163, line 29: Where does `styles.mlAuto, styles.alignItemsCenter, styles.gap2` come from?

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    /** A boolean flag that gives the row (and its icon) a green fill if true */
+    success?: boolean;
```
dariusz-biela: Shouldn't we change this to “variant”?

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+type MenuItemTitleProps = {
+    /** The title text. Strings are converted to LTR; custom nodes are rendered as-is. */
+    children?: ReactNode;
+
+    /** Used to truncate the text with an ellipsis after computing the text layout */
+    numberOfLines?: number;
+
+    /** Font weight variant. `strong` (default) is bold; `normal` uses the default weight. */
+    variant?: MenuItemTitleVariant;
```
dariusz-biela: This makes me wonder about two things:
1. Don't these values depend on some global variant from Root?
2. If not, I'd still change the name from “variant” to “weight.”

src/components/MenuItem/leaves/trailing/MenuItemBadge.tsx
```diff
@@ -0,0 +1,29 @@
...
+import useThemeStyles from '@hooks/useThemeStyles';
+
+import React from 'react';
+
+/**
+ * A `Badge` that follows the menu item's focused state. Placement is up to the consumer:
+ * inside `MenuItem.Trailing` (right), inside `MenuItem.Content` (below the title), or as a
+ * direct child of the Root (its own row).
+ */
+function MenuItemBadge({badgeStyles, pressable, ...rest}: BadgeProps) {
```
dariusz-biela, line 14: Are you sure you don't want to narrow down your BadgeProps in some way?

src/components/MenuItem/leaves/trailing/MenuItemCopyButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+type MenuItemCopyButtonProps = {
+    /** The value to copy to the clipboard */
+    value: string;
+};
+
+/**
+ * A hover-revealed copy-to-clipboard button for `MenuItem.Trailing` on devices with hover support.
+ * On touch devices, wire copying through `onSecondaryInteraction` on the Root instead.
+ */
+function MenuItemCopyButton({value}: MenuItemCopyButtonProps) {
```
dariusz-biela, line 23: On the web, the implementation is fine. 

As for the mobile version, `onSecondaryInteraction` is rather counterintuitive, and personally, I’d be surprised if double-tapping a button in the app copied its text.

I think we should review how this component is being used, and if there’s no evidence that `onSecondaryInteraction` serves any purpose, we should remove it.

src/components/MenuItem/MenuItemContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+    isDisabled: boolean;
+
+    /** Whether the row responds to interactions */
+    isInteractive: boolean;
+
+    /** Whether the row uses success (green) styling */
+    isSuccess: boolean;
+
+    /** Whether the row is rendered inside a compact popover menu */
+    isCompact: boolean;
```
jmusial: this sounds like a sth that should trigger decomposition into `MenuItem` and `CompactMenuItem` rather than a context value

src/components/MenuItem/compound.tsx
```diff
@@ -0,0 +1,117 @@
+/**
+ * MenuItem (compound)
+ *
+ * A composable decomposition of the classic `MenuItem`, following the composition-over-configuration
+ * pattern: instead of ~135 props configuring one monolith, the consumer assembles the row from
```
jmusial: I'd shorten this descption. Personally find those "historical" comments by Claude not helpful

src/components/MenuItem/compound.tsx
jmusial: can't find `README.md` file ?

src/components/MenuItem/layout/MenuItemContent.tsx
```diff
@@ -0,0 +1,36 @@
...
+import type {ReactNode} from 'react';
+import type {StyleProp, ViewStyle} from 'react-native';
+
+import React from 'react';
+import {View} from 'react-native';
+
+type MenuItemContentProps = {
+    /** The text blocks of the row (`MenuItem.Title`, `MenuItem.Description`, or any custom content),
+     * stacked vertically in the order they are declared */
+    children: ReactNode;
```
jmusial: can use `PropsWithChildren`

jmusial: can use `PropsWithChildren`

src/components/MenuItem/layout/MenuItemContent.tsx
```diff
@@ -0,0 +1,36 @@
...
+ */
+function MenuItemContent({children, style}: MenuItemContentProps) {
+    const styles = useThemeStyles();
+    const StyleUtils = useStyleUtils();
+    const {isCompact} = useMenuItemState();
+
+    return <View style={[styles.justifyContentCenter, styles.flex1, styles.gap1, StyleUtils.getMenuItemTextContainerStyle(isCompact), style]}>{children}</View>;
+}
+
+MenuItemContent.displayName = 'MenuItemContent';
```
jmusial: `displayName`  redundant

src/components/MenuItem/layout/MenuItemContent.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export type {MenuItemContentProps};
```
jmusial: `export type {MenuItemContentProps};` redundant

src/components/MenuItem/leaves/messages/MenuItemError.tsx
```diff
@@ -0,0 +1,28 @@
...
+import type {MenuItemFormMessageProps} from './MenuItemFormMessage';
+
+import MenuItemFormMessage from './MenuItemFormMessage';
+
+type MenuItemErrorProps = Omit<MenuItemFormMessageProps, 'isError'>;
+
+/**
+ * An error message block. Place it inside the Root, after `MenuItem.Row`, to render it under the main line.
+ */
+function MenuItemError({children, shouldShowRedDotIndicator = false, shouldRenderAsHTML, style}: MenuItemErrorProps) {
```
Guccio163: This and MenuItemHint - I would consider if adding components like this, essentially only wrappers on MenuItemFormMessage fixing the need for `isError`, when MenuItemFormMessage itself also adds only `styles.menuItemError`, is necessary. Theoretically it reduces the API to be used when migrating, but if these two are rarely used then at least I would stick to using only MenuItemFormMessage, defaulting to more common isError value and specifying the other inline when used. In case they would be rarely used then I would consider simply padding FormHelpMessage component as a child

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+    }, [interactive]);
+
+    const onPressAction = (event: GestureResponderEvent | KeyboardEvent | undefined) => {
+        if (disabled || !interactive) {
+            return;
+        }
+
+        if (event?.type === 'click' && typeof HTMLElement !== 'undefined' && event.currentTarget instanceof HTMLElement) {
+            event.currentTarget.blur();
+        }
```
dariusz-biela, line 147: https://github.com/Expensify/App/pull/9348

```suggestion
        // Prevent clicked menu items from retaining an unwanted focus outline on web, especially in Safari.
        if (event?.type === 'click' && typeof HTMLElement !== 'undefined' && event.currentTarget instanceof HTMLElement) {
            event.currentTarget.blur();
        }
```

src/components/MenuItem/layout/MenuItemContent.tsx
jmusial: and in all other components

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+    onFocus?: () => void;
+
+    /** Handles what to do when the row loses focus */
+    onBlur?: () => void;
+
+    /** Whether the row should be interactive at all */
+    interactive?: boolean;
+
+    /** Should we disable this row? */
+    disabled?: boolean;
```
jmusial:
```suggestion
    isDisabled?: boolean;
```

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+    onSecondaryInteraction?: (event: GestureResponderEvent | MouseEvent) => void;
+
+    /** Handles what to do when the row is focused */
+    onFocus?: () => void;
+
+    /** Handles what to do when the row loses focus */
+    onBlur?: () => void;
+
+    /** Whether the row should be interactive at all */
+    interactive?: boolean;
```
jmusial:
```suggestion
    isInteractive?: boolean;
```

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    /** Should we disable this row? */
```
jmusial:
```suggestion
    /** Weather the the menu Item is disabled? */
```

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
```diff
@@ -0,0 +1,61 @@
...
+                styles.flexShrink1,
+                styles.popoverMenuText,
+                variant === CONST.MENU_ITEM.TITLE_VARIANT.STRONG && styles.textStrong,
+                numberOfLines !== 1 ? styles.preWrap : styles.pre,
+                isInteractive && isDisabled && styles.userSelectNone,
+                styles.ltr,
+                styles.mw100,
+                style,
+            ]}
+            numberOfLines={numberOfLines || undefined}
```
Guccio163, line 51: Can't we just pass numberOfLines? If it isn't specified then should evaluate to undefined

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+    success?: boolean;
+
+    /** Whether the screen containing the row is focused (forwarded to Hoverable) */
+    isScreenFocused?: boolean;
+
+    /** The action accepts anonymous users or not */
+    isAnonymousAction?: boolean;
+
+    /** Should check anonymous user in onPress function */
+    shouldCheckActionAllowedOnPress?: boolean;
```
jmusial: should this be concern of `MenuItem` ?

src/pages/settings/SettingsMenuItem.tsx
```diff
@@ -21,6 +25,7 @@ type SettingsMenuItemProps = {
 };
 
 function SettingsMenuItem({item, isFocused, keyTitle, isExecuting, isScreenFocused, onPress, wrapperStyle}: SettingsMenuItemProps) {
```
dariusz-biela, line 23: Unless I’m missing something, SettingsMenuItem looks like a good example of the standard MenuItem composition. The issue also explicitly calls out extracting a MenuItemStandard preset to cover most call sites.
                                                                                                                                                                                                               
Would this be a good opportunity to introduce that preset and build SettingsMenuItem on top of it? The compound API would remain available as an escape hatch for specialized layouts, while common consumers  would avoid repeating the Row / Content / Trailing structure.                                                                                                                                                    
                                                                                                                                                                                                             
For example, roughly:                                                                                                                                                                                            
                                                                                                                                                                                                                
type MenuItemStandardProps = Omit<MenuItemRootProps, 'children' | 'accessibilityLabel'> & {                                                                                                                      
      title: string;                                                                                                                                                                                               
     description?: string;                                                                                                                                                                                        
     icon?: IconAsset;                                                                                                                                                                                            
     iconFill?: MenuItemIconProps['fill'];                                                                                                                                                                        
     iconRight?: IconAsset;                                                                                                                                                                                       
     shouldShowRightIcon?: boolean;                                                                                                                                                                               
                                                                                                                                                                                                                   
     brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>;                                                                                                                                      
     errorText?: ReactNode;                                                                                                                                                                                       
     hintText?: ReactNode;                                                                                                                                                                                        
                                                                                                                                                                                                                  
     /** Derived from title/description when omitted */                                                                                                                                                           
     accessibilityLabel?: string;                                                                                                                                                                                 
                                                                                                                                                                                                                  
     badgeText?: string;                                                                                                                                                                                          
     rightLabel?: string;                                                                                                                                                                                         
     numberOfLinesTitle?: number;                                                                                                                                                                                 
 };

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+                                    isHovered,
+                                    isPressed: pressed,
+                                    isFocused: focused,
+                                    isDisabled: disabled,
+                                    isInteractive: interactive,
+                                    isSuccess: success,
+                                    isCompact,
+                                }}
+                            >
+                                <View style={styles.flex1}>{children}</View>
```
dariusz-biela: I'm wondering if we can get rid of this <View>. I know it's a relatively minor optimization, but from a React Native perspective, rendering each <View> takes some time.

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -0,0 +1,229 @@
...
+    return (
+        <View onBlur={onBlur}>
+            <Hoverable isFocused={isScreenFocused}>
+                {(isHovered) => (
+                    <PressableWithSecondaryInteraction
+                        onPress={shouldCheckActionAllowedOnPress ? callFunctionIfActionIsAllowed(onPressAction, isAnonymousAction) : onPressAction}
+                        onSecondaryInteraction={onSecondaryInteraction}
+                        wrapperStyle={wrapperStyle}
+                        activeOpacity={!interactive ? 1 : variables.pressDimValue}
+                        opacityAnimationDuration={0}
```
jmusial: I belive he have consts for this

src/components/MenuItem/leaves/messages/MenuItemHelperText.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+    /** Any additional styles to apply to the helper text */
+    style?: StyleProp<TextStyle>;
+};
+
+/**
+ * Non-interactive helper text. Place it as a sibling AFTER the Root — it lives outside of the
+ * pressable/hoverable area and therefore does not consume the menu item context.
+ */
+function MenuItemHelperText({children, style}: MenuItemHelperTextProps) {
```
Guccio163: No road for shouldParseHelperText={true}

src/components/MenuItem/leaves/leading/MenuItemIcon.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+type MenuItemIconVariant = (typeof CONST.MENU_ITEM.ICON_VARIANT)[keyof typeof CONST.MENU_ITEM.ICON_VARIANT];
+
+type MenuItemIconProps = {
+    /** Icon to display */
+    src: IconAsset;
+
+    /** The fill color to pass into the icon. When omitted, the fill follows the row's interaction state
+     * (hovered/pressed/focused/disabled), matching the classic MenuItem behavior. */
+    fill?: string | ((isHovered: boolean) => string);
```
dariusz-biela:
```suggestion
    fill?: string | ((state: MenuItemState) => string);
```

src/components/MenuItem/layout/MenuItemRow.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * The main horizontal line of a menu item. A plain flex row — it never inspects its children;
+ * the consumer decides which cells to include and in what order.
+ */
+function MenuItemRow({children, style}: MenuItemRowProps) {
+    const styles = useThemeStyles();
+    const {isDisabled, isCompact} = useMenuItemState();
+
+    return <View style={[styles.flexRow, styles.pointerEventsAuto, styles.gap3, isDisabled && styles.cursorDisabled, isCompact && styles.alignItemsCenter, style]}>{children}</View>;
```
jmusial, line 29: Create named styles for componenets please - will look cleaner

src/components/MenuItem/layout/MenuItemTrailing.tsx
```diff
@@ -0,0 +1,36 @@
...
+import React from 'react';
+import {View} from 'react-native';
+
+type MenuItemTrailingProps = {
+    /** The right-aligned cells (`MenuItem.Badge`, `MenuItem.RightLabel`, `MenuItem.BrickRoadIndicator`,
+     * `MenuItem.Chevron`, `MenuItem.CopyButton`, or any custom content), laid out horizontally */
+    children: ReactNode;
+
+    /** Any additional styles to apply to the trailing container */
+    style?: StyleProp<ViewStyle>;
```
jmusial, line 15: Do we have to pass `style` to every component ? Does not used for now.

jmusial (commented):
This is just a first batch - will finish tomorrow

src/components/MenuItem/leaves/trailing/MenuItemRightLabel.tsx
```diff
@@ -0,0 +1,42 @@
...
+    /** Style variant. `label` (default) is the classic `rightLabel` look; `subtitle` is the classic `subtitle` look. */
+    variant?: MenuItemRightLabelVariant;
+
+    /** Any additional styles to apply to the label */
+    style?: StyleProp<TextStyle>;
+};
+
+/**
+ * A right-aligned supporting text cell for `MenuItem.Trailing` — covers both the classic
+ * `rightLabel` and `subtitle` use cases.
```
Guccio163, line 25: Hard to argue, but I guess it doesn't cover the `rightLabelIcon` use-case 😳

src/CONST/index.ts
```diff
@@ -9229,6 +9216,38 @@ const CONST = {
...
+
+    MENU_ITEM: {
+        BADGE_POSITION: {
+            RIGHT: 'right',
+            BELOW_TITLE: 'belowTitle',
+            SEPARATE_ROW: 'separateRow',
+        },
+        TITLE_FORMAT: {
+            MARKDOWN: 'markdown',
+            HTML: 'html',
```
dariusz-biela, line 9262: `MENU_ITEM.BADGE_POSITION` and `MENU_ITEM.TITLE_FORMAT` are used nowhere.

src/components/MenuItem/leaves/trailing/MenuItemCopyButton.tsx
dariusz-biela, line 23: The AI found this regression, but I'm not sure if it's intentional:
> Context exposes `isInteractive` but it is not checked, so a pressable copy button can legally render inside a pressable Root (legacy only showed it when `!interactive`). Suggest returning null (or warning in dev) when the row is interactive. Same concern applies to pressable `MenuItem.Badge`.

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
```diff
@@ -0,0 +1,61 @@
...
+    variant?: MenuItemTitleVariant;
+
+    /** Any additional styles to apply to the title (e.g. pass a non-bold text style for a basic title) */
+    style?: StyleProp<TextStyle>;
+};
+
+/**
+ * The title block of a `MenuItem.Content`. Bold by default, single line by default.
+ */
+function MenuItemTitle({children, numberOfLines = 1, variant = CONST.MENU_ITEM.TITLE_VARIANT.STRONG, style}: MenuItemTitleProps) {
```
Guccio163: I guess that this is in `renderTitleContent`'s place which uses DisplayNames and TextWithEmojiFragment, are we sure that this covers all the possibilities?

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    /** Any additional styles to apply to the description */
+    style?: StyleProp<TextStyle>;
+};
+
+/**
+ * The supporting text block of a `MenuItem.Content`. Place it before or after `MenuItem.Title`
+ * to render the description above or below the title.
+ */
+function MenuItemDescription({children, numberOfLines = 2, variant = CONST.MENU_ITEM.DESCRIPTION_VARIANT.SUPPORTING, style}: MenuItemDescriptionProps) {
```
Guccio163, line 32: No `isDeleted` styles, based on `style` passed to the core component, probably should be passed through context, eventually passed inline when being used, to be decided

src/components/MenuItem/leaves/leading/MenuItemIcon.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    /** Icon should be displayed in its own color (e.g. multi-color illustrations) */
+    displayInDefaultIconColor?: boolean;
+
+    /** Determines how the icon should be resized to fit its container */
+    contentFit?: ImageContentFit;
+
+    /** Container variant. `default` reserves the classic fixed-width cell; `compact` hugs the icon.
+     * Rows inside a compact menu are compact regardless of this prop. */
+    variant?: MenuItemIconVariant;
```
dariusz-biela, line 46: Maybe “size” would be a better name

src/components/MenuItem/leaves/leading/MenuItemIcon.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    /** Any additional styles to apply to the icon container */
+    style?: StyleProp<ViewStyle>;
+
+    /** Additional styles to pass to the icon itself */
+    iconStyle?: StyleProp<ViewStyle>;
```
dariusz-biela, line 52: Honestly, personally, I would remove all style props and restore them only as a last resort and with a narrowed-down type.

Guccio163 (commented):
A lot of work done, looks like we have a lot to go through; Here are some things that I've found, I'm open for a discussion.

BTW I feel like if you'd introduced some MenuItem-copy component that would in some way show which MenuItem fragments are being replaced with which primitives it would make comparison that easier, moreover we'd see in the migrated component which parts are still being used and need resolution. I would see it as a (f.ex.) nowhere-used MenuItemOld.tsx deprecated file next to the MenuItem.tsx, pushed before so it shows all changes, with code fragments replaced with exact primitives. LMKWYT 👀

src/components/MenuItem/compound.tsx
JakubKorytko: there is no README in this directory?

src/CONST/index.ts
```diff
@@ -9216,6 +9216,38 @@ const CONST = {
         NEWDOT: 83,
         OAUTH: 86,
     },
+
+    MENU_ITEM: {
+        BADGE_POSITION: {
```
JakubKorytko, line 9255: `BADGE_POSITION` (here) and `TITLE_FORMAT` (below) are never referenced anywhere. Unlike the other sub-enums, no component derives a variant type from them or reads their values, so they are dead code. Looks like leftovers from an earlier design where the badge had a position and the title supported markdown/html. Perhaps let's drop both blocks until a consumer actually needs them?

src/pages/settings/InitialSettingsPage.tsx
```diff
@@ -94,19 +94,10 @@ type MenuData = WithSentryLabel & {
     brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>;
     action: () => void;
     link?: string | (() => Promise<string>);
-    iconType?: typeof CONST.ICON_TYPE_ICON | typeof CONST.ICON_TYPE_AVATAR | typeof CONST.ICON_TYPE_WORKSPACE;
-    iconStyles?: StyleProp<ViewStyle>;
-    fallbackIcon?: IconAsset;
-    shouldStackHorizontally?: boolean;
-    avatarSize?: ValueOf<typeof CONST.AVATAR_SIZE>;
-    floatRightAvatars?: TIcon[];
```
JakubKorytko, line 102: Removing `floatRightAvatars?: TIcon[]` leaves the `Icon as TIcon` import (top of the file) with no remaining usages, so lint/typecheck will flag it. Worth removing that import in the same change.

JakubKorytko (commented):
nice decomposition, the leaf split reads well and the compiled files pass the React Compiler compliance check. A couple of small cleanups below that are additive to the feedback already left by other reviewers (I skipped the points already raised: the missing `index.ts`, and the many-props-on-Root discussion).

src/components/MenuItem/leaves/leading/MenuItemAvatar.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import React from 'react';
+import {View} from 'react-native';
+
+type MenuItemAvatarProps = {
+    /** The source of the avatar image */
+    source: AvatarSource;
+
+    /** Denotes whether it is an avatar or a workspace avatar */
+    type?: typeof CONST.ICON_TYPE_AVATAR | typeof CONST.ICON_TYPE_WORKSPACE;
```
jmusial, line 23: we have a type for this in `OnyxCommon`

src/components/MenuItem/leaves/leading/MenuItemAvatar.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * represents a user or a workspace.
+ */
+function MenuItemAvatar({source, type = CONST.ICON_TYPE_AVATAR, avatarID, fallbackIcon, size = CONST.AVATAR_SIZE.DEFAULT, name, style}: MenuItemAvatarProps) {
+    const icons = useMemoizedLazyExpensifyIcons(['FallbackAvatar']);
+    const styles = useThemeStyles();
+    const StyleUtils = useStyleUtils();
+
+    return (
+        <View style={[styles.popoverMenuIcon, StyleUtils.getAvatarWidthStyle(size), style]}>
+            <Avatar
```
jmusial, line 52: `Avatar` will be phased off in favour of `WorspaceAvatar` & `UserAvatar` where possible. Worth keeping in mind here

src/components/MenuItem/leaves/messages/MenuItemHint.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import type {MenuItemFormMessageProps} from './MenuItemFormMessage';
+
+import MenuItemFormMessage from './MenuItemFormMessage';
+
+type MenuItemHintProps = Omit<MenuItemFormMessageProps, 'isError' | 'shouldShowRedDotIndicator'>;
```
jmusial: Having `Omit` here makes the props obscured, I'd rather have explicit type here and then union with additional in `MenuItemFormMessageProps`

src/components/MenuItem/leaves/trailing/MenuItemBadge.tsx
jmusial: +1, and avoid spreading plz. It hides unnecessary props

jmusial (commented):
I would suggest a different, more step by step approach to this refactor.
Right now half of the new components have no call sites, but bloat out the PR making it harder to review and will lengthen whole process.

How about @ first implementing only the components used in a particular migration use case and then adding them progressively while migrating following call sites ? 

Would make PRs smaller and redundant props easier to spot

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
war-in: The base value in the old `MenuItem` is 1, so I replicated that behaviour. If we pass undefined as a default, all call sites that require one line (the majority) would have to define it

src/components/MenuItem/layout/MenuItemTrailing.tsx
war-in: I aggregated a few styles from the main MenuItem component to replicate the behaviour. The code is not 1:1

src/components/MenuItem/MenuItem.tsx
```diff
@@ -1,3 +1,27 @@
+import ActivityIndicator from '@components/ActivityIndicator';
```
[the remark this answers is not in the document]
war-in: fixed

src/pages/settings/SettingsMenuItem.tsx
```diff
@@ -49,35 +54,48 @@ function SettingsMenuItem({item, isFocused, keyTitle, isExecuting, isScreenFocus
           }
         : undefined;
 
+    const combinedAccessibilityLabel = [keyTitle, item.brickRoadIndicator ? translate('common.yourReviewIsRequired') : ''].filter(Boolean).join('. ');
```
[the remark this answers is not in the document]
war-in: I added MenuItemAccessibilityContext to handle that case

src/components/MenuItem/MenuItemContext.tsx
war-in: isCompact is derived from publicly available hooks
- useIsCompactMenu
- useResponsiveLayout

so I'll remove the prop and use the mentioned hooks where needed 👍

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: Nope, I removed it entirely and now callers can wrap onPress when necessary

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: Actually, we haven't but I added one 😄

src/components/MenuItem/layout/MenuItemTrailing.tsx
war-in: I'm worried that they'll be required one day, but I started with a simpler, more split approach [here](https://github.com/Expensify/App/pull/97339) and style props are not required yet there

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: There are 11 usages of MenuItem without onPress and with interactive={false}. I decided to derive the isInteractive value from onPress

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: Agree, I did exactly that

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: https://github.com/Expensify/App/pull/96650#discussion_r3675397082

src/components/MenuItem/compound.tsx
war-in: You're right. I decided to remove HelperText, Error and Hint components because all of them can be easily implemented using `FormHelpMessage`

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: Reasonable, let's add the variant once we're sure it's needed during the migration

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
war-in: Changed to weight. We sometimes need a compact icon outside the context menu (when root's `isCompact` is set)

src/components/MenuItem/leaves/trailing/MenuItemBadge.tsx
war-in: I'll do that during the migration when we migrate the real usages of badge

src/components/MenuItem/leaves/trailing/MenuItemCopyButton.tsx
war-in: We do need `onSecondaryInteraction` mostly for copying MenuItems. The interaction is actually long press which is correct and expected behaviour on mobile

src/pages/settings/SettingsMenuItem.tsx
war-in: I found an even simpler example and defined Standard preset here - https://github.com/Expensify/App/pull/97339

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: Unfortunately, it has to be placed here. I tried moving flex elsewhere, but the layout broke. I agree that it's a relatively minor optimisation

src/CONST/index.ts
war-in: True, those are leftovers from the debugging and will be needed in some point (at least should). This PR will serve as a reference for future, more focused ones, so I'll leave it for now

src/components/MenuItem/leaves/trailing/MenuItemCopyButton.tsx
war-in: fixed, thank you!

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: I restored `onPressIn` and `onPressOut` support. `disabledStyle` will be covered by style prop

src/pages/settings/SettingsMenuItem.tsx
war-in: MenuItemIcon has `variant` prop that adds `wAuto` style

src/components/MenuItem/leaves/messages/MenuItemError.tsx
war-in: Agree, I removed the mentioned components and `FormHelpMessage` can be used instead

src/components/MenuItem/leaves/trailing/MenuItemRightLabel.tsx
war-in: True, but it can be passed as a sibling of MenuItemRightLabel

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
war-in: Yes, it should be passed inline in styles

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
war-in: `DisplayNames` and `TextWithEmojiFragment ` can always be passed as children of `MenuItemTitle`

war-in:
Closing in favour of ongoing migration - https://github.com/Expensify/App/issues/96202

PR #96748, opened 2026-07-22
Optimize top-N report selection in option list building

jmgraa:
### Explanation of Change

Optimizes report selection in `createFilteredOptionList` for accounts with large report lists.

`optionsOrderBy` and `optionsOrderAndGroupBy` now decorate each option with its pre-computed sort key (shared `DecoratedOptionHeap` helper), so the comparator runs exactly once per option instead of on every O(log n) heap comparison and every eviction check, cutting redundant work when picking the top N reports.

Top-N selection happens directly in `createFilteredOptionList` via `optionsOrderBy` with `reportSortComparator`, keeping the same rules: self-DM first, then non-archived, then most recent; search mode still returns all reports. `optionsOrderBy` itself is now implemented as `optionsOrderAndGroupBy` with no separators, removing the duplicated heap logic between the two functions.

Two fixes along the way:

- `optionsOrderAndGroupBy` now respects `reversed: true` when a heap is at capacity - previously the eviction check always compared with `>`, so reversed callers kept the newest options instead of the oldest. Covered by a new unit test.
- The `limit <= 0` early return in `optionsOrderAndGroupBy` returned a sparse array (`Array(n).map()` skips empty slots), which could crash callers destructuring the groups; it now returns N+1 real empty arrays via `Array.from`.

Aside from the `reversed: true` fix, no user-facing behavior change - same reports, same order, less work during option-list building. Unit tests cover comparator call counts, grouping, limits, and reversed ordering.

#### createFilteredOptionList (tests/perf-test/OptionsListUtils.perf-test.ts)

| Before | After |
| :--- | :--- |
| 280ms   | 173ms |
| 280ms | 174ms |
| 270ms   | 172ms |
| 285ms    | 170ms |

### Fixed Issues

$ https://github.com/Expensify/App/issues/95335
PROPOSAL:

### Tests

1. Open **New chat** (+) and confirm recent chats load and are ordered with the most recent at the top.
2. Confirm your **self-DM** (Notes) appears near the top even if it has not been updated recently.
3. Open **Search**, type a report or user name, and confirm matching results appear (including older or archived chats if they match).
4. Clear the search query and confirm recent results still load correctly.
5. Share content into Expensify (or open **Share** from the app) and confirm destination chats/users load and you can pick one.
6. Scroll/paginate on a screen that loads more reports (e.g. **New chat** with many reports) and confirm additional chats appear.

In addition, the component should behave just as it does in production

### Offline tests

Same as tests

### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/1fc7c8d2-f189-41c4-be96-153da8db21b0

</details>

<details>

https://github.com/user-attachments/assets/58f623d2-a810-474c-83ed-18be238fca41

</details>

<details>

https://github.com/user-attachments/assets/6edb32d3-789c-4b9e-9bf3-b9c3a0242d01

</details>


src/libs/OptionsListUtils/index.ts
```diff
@@ -1915,33 +1915,40 @@ function optionsOrderBy<T = SearchOptionData | PersonalDetailOptionData>(
...
+    // on every O(log n) heap comparison, so `comparator` is evaluated exactly once per option.
+    type Decorated = {key: number | string; option: T};
+    const getKey = (decorated: Decorated) => decorated.key;
+    const heap = reversed ? new MaxHeap<Decorated>(getKey) : new MinHeap<Decorated>(getKey);
+
     for (const option of options) {
         if (filter && !filter(option)) {
             continue;
         }
+        const decorated: Decorated = {key: comparator(option), option};
```
dariusz-biela: Could we add a test asserting that the comparator is called exactly once per option that passes the filter? The output tests protect behavior, but they would not catch a regression of the optimization introduced by this PR.

src/libs/OptionsListUtils/index.ts
```diff
@@ -1915,33 +1915,40 @@ function optionsOrderBy<T = SearchOptionData | PersonalDetailOptionData>(
...
     if (limit !== undefined && limit <= 0) {
         return {options: [], hasMore};
     }
 
+    // Decorate each option with its comparator key once.
+    // The heap then compares precomputed primitives instead of re-running `comparator`
+    // on every O(log n) heap comparison, so `comparator` is evaluated exactly once per option.
+    type Decorated = {key: number | string; option: T};
+    const getKey = (decorated: Decorated) => decorated.key;
+    const heap = reversed ? new MaxHeap<Decorated>(getKey) : new MinHeap<Decorated>(getKey);
```
dariusz-biela: Could we extract the key decoration and bounded heap logic into a small shared helper? This would keep `optionsOrderBy()` and `optionsOrderAndGroupBy()` consistent and avoid duplicating the limit, eviction, `reversed`, and result-ordering logic.

Please note that this code was written by an AI, and I'm posting it here for illustrative purposes only:
```tsx
// Options are decorated with their comparator key once before entering a heap.
// The heap then compares precomputed primitives instead of re-running the comparator
// on every heap comparison, so the comparator is evaluated exactly once per option that passes the filter.
type KeyedOption<T> = {key: number | string; option: T};

type BoundedKeyedHeap<T> = {
    push: (item: KeyedOption<T>) => boolean;
    drain: () => T[];
};

/**
 * Create a heap of key-decorated options bounded to `limit` items.
 * `push` evicts the current worst item when the heap is full and the candidate outranks it,
 * and returns true when the heap was already full, which callers surface as the `hasMore` flag.
 * `drain` empties the heap into an array ordered best-first and strips the decoration.
 */
function createBoundedKeyedHeap<T>(reversed: boolean, limit?: number): BoundedKeyedHeap<T> {
    const getKey = (item: KeyedOption<T>) => item.key;
    const heap = reversed ? new MaxHeap<KeyedOption<T>>(getKey) : new MinHeap<KeyedOption<T>>(getKey);
    return {
        push(item: KeyedOption<T>) {
            if (limit === undefined || heap.size() < limit) {
                heap.push(item);
                return false;
            }
            const peekedValue = heap.peek();
            if (!peekedValue) {
                throw new Error('Heap is empty, cannot peek value');
            }
            if (reversed ? item.key < peekedValue.key : item.key > peekedValue.key) {
                heap.pop();
                heap.push(item);
            }
            return true;
        },
        drain() {
            return [...heap].reverse().map((entry) => entry.option);
        },
    };
}
```

dariusz-biela (commented):
This looks good. I think we should also apply the precomputed-key optimization to `optionsOrderAndGroupBy()`, which is used by `getValidOptions()` in the Share flow. It currently recalculates `recentReportComparator` during heap comparisons, while both helpers solve the same bounded top-N problem.

src/libs/OptionsListUtils/index.ts
dariusz-biela, line 1996: The `limit <= 0` early return returns a sparse array rather than `N + 1` empty arrays. `Array(n)` creates empty slots, and `.map()` skips them, so with two separators the result has length 3 but `result.options[0]` is `undefined`.

  This can cause `getValidOptions()` to crash after destructuring the result and accessing `selfDMChats.length`.

  Although this issue already exists on `main`, this PR adds a test for exactly this path (`does not call comparator when limit is 0`), but the test only checks comparator calls and therefore does not catch the invalid result.

  Could we change this to:

```ts
  return {options: Array.from({length: separators.length + 1}, () => []), hasMore};
```
  and strengthen the test with:
```ts
  expect(result.options).toStrictEqual([[], [], []]);
```

dariusz-biela, line 1: Could you please update the PR description so it reflects the current implementation and scope?

- The PR description still references the removed selectTopReportsForOptionList helper and should be updated to reflect the current implementation.

- The PR fixes optionsOrderAndGroupBy() behavior for reversed=true, so this semantic change should be documented and covered by a test.

- Once you've fixed `return {options: Array(separators.length + 1).map(() => []), hasMore};`, please mention this in the PR description as well.

src/libs/OptionsListUtils/index.ts
```diff
@@ -1904,6 +1903,56 @@ const recentReportComparator = (option: SearchOptionData) => {
...
+                throw new Error('Heap is empty, cannot peek value');
+            }
+
+            if (reversed ? decoratedOption.key < peekedValue.key : decoratedOption.key > peekedValue.key) {
+                heap.pop();
+                heap.push(decoratedOption);
+            }
+
+            return true;
+        },
```
dariusz-biela: Could we make the boolean contract of `pushDecoratedOption()` more explicit? It returns true when the heap was already at capacity, rather than when the option was pushed, so a name such as `pushAndCheckHasMore()` or a named `{hasMore}` result would make the call site easier to understand.

dariusz-biela (approved):
LGTM, nice work!

src/libs/OptionsListUtils/index.ts
```diff
@@ -1904,6 +1903,57 @@ const recentReportComparator = (option: SearchOptionData) => {
...
+        pushAndCheckHasMore(decoratedOption) {
+            if (limit === undefined || heap.size() < limit) {
+                heap.push(decoratedOption);
+                return false;
+            }
+
+            const peekedValue = heap.peek();
+            if (!peekedValue) {
+                throw new Error('Heap is empty, cannot peek value');
+            }
```
ikevin127: 🟡 **Dead `throw` in `pushAndCheckHasMore`**

The guard at `1930` already returns when `size < limit`, so reaching `peek()` implies `size >= limit >= 1` and the heap is never empty.

The `throw new Error('Heap is empty, cannot peek value')` is unreachable defensive code carried over from the original. Fine to leave, but no test can exercise it, so we do not count it toward coverage. 

Optionally drop it since `peek()` already returns `null`-safe.

tests/unit/OptionsListUtilsTest.tsx
```diff
@@ -4151,6 +4152,179 @@ describe('OptionsListUtils', () => {
         });
     });
 
+    describe('optionsOrderBy()', () => {
```
ikevin127, line 4155: 🟡 **`optionsOrderBy()` suite never asserts `hasMore`**

Both production callers discard it (`ShareTab.tsx:96`, `index.ts:1726` read only `.options`), so low risk, but a one-line `expect(result.hasMore).toBe(...)` on a `limit < input` and a `limit >= input` case would lock the contract for future consumers.

ikevin127 (approved):
**🟢 LGTM**

Only have **two** 🟡 non-blocker comments (see above) and **one** ⚠️ warning regarding merging main branch to avoid a now-fixed crash during manual testing in Spend page (not related to this PRs changes).

##

### 🧪 Tests I ran to validate PRs changes:

✅ 1. **Recency and self-DM ordering (online):** Open New chat (+) on a high-traffic account. Confirm most-recent chats are at the top and your self-DM (Notes) sits near the top even if stale. This exercises the top-N heap selection and `reportSortComparator` priority.
✅ 2. **Search returns all matches (online):** Open Search, type a name that matches an older or archived chat, confirm it appears; clear the query and confirm the recent list restores. This validates the `isSearching` branch that bypasses the cap.
✅ 3. **Offline parity + Share flow (offline):** Turn off network, repeat New chat and Share-into-Expensify, confirm the same destination chats load, are pickable, and no console error fires. This covers `ShareTab`'s `optionsOrderBy` call and the `Array.from` empty-group path.

### Reassure perf: baseline (`main`) vs this PR

**Headline:** `createFilteredOptionList`, the function this PR optimizes, drops **44.4  ms → 23.6 ms (−46.8%)**, and its search-mode variant **17.0 ms → 11.9 ms (−30.0%)**. This corroborates the PR description's numbers (different absolute ms because of machine/fixture, same direction and magnitude).

- `measureFunction` mean, single run each. Negative delta = faster

| Scenario | Baseline | This PR | Δ | % | |
| :--- | ---: | ---: | ---: | ---: | :---: |
| `createFilteredOptionList` | 44.4 ms | 23.6 ms | −20.8 ms | −46.8% | 🟢 |
| `createFilteredOptionList` with `isSearching: true` | 17.0 ms | 11.9 ms | −5.1 ms | −30.0% | 🟢 |
| `getSearchOptions` with `isSearching: true` | 1.2 ms | 1.0 ms | −0.2 ms | −16.7% | 🟢 |
| `getSearchOptions` | 6.8 ms | 5.8 ms | −1.0 ms | −14.7% | 🟢 |
| `getShareDestinationOptions` | 6.6 ms | 5.8 ms | −0.8 ms | −12.1% | 🟢 |
| empty search term with selected options + mocked personal details | 1.7 ms | 1.6 ms | −0.1 ms | −5.9% | ⚪ |
| `getFilteredOptions` with empty search value | 0.4 ms | 0.4 ms | 0.0 ms | 0.0% | ⚪ |
| `getFilteredOptions` with search value | 1.9 ms | 2.0 ms | +0.1 ms | +5.3% | ⚪ |
| worst case: search term matching subset of selected / recent reports / personal details | 13.4 ms | 14.5 ms | +1.1 ms | +8.2% | ⚪ |

src/libs/OptionsListUtils/index.ts
jmgraa: You're right that the throw was unreachable, so I removed it. However, I kept the null check, but it now returns true instead of throwing. That way we avoid needing an eslint-disable for a non-null assertion.

tests/unit/OptionsListUtilsTest.tsx
jmgraa: Added two tests in the `optionsOrderBy()` suite to cover `hasMore`:

- `limit` < `options.length` - `hasMore` is true
- `limit` >= `options.length` - `hasMore` is false

mjasikowski (approved):

PR #54798, opened 2025-01-04
Implemented International Bank Account flow

shubham1206agra:
### Fixed Issues

$ https://github.com/Expensify/App/issues/51082
$ https://github.com/Expensify/App/issues/54743

### Tests

https://docs.google.com/document/d/1H1hMnTcMrR8MT6BXz3fCE99eVIurgrHDtaxIH5aW8Ug/edit#heading=h.nhtsmauf72jq

- [x] Verify that no errors appear in the JS console

### Offline tests

Same as Tests

### QA Steps

Same as Tests

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [x] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`)
    - [x] I verified that comments were added to code that is not self explanatory
[...]

PR #23258, opened 2023-07-20
Fix focus after picking an emoji in edit mode

kosmydel:
### Details

This PR addresses an issue, where the input bar wasn't focused after selecting an emoji from the modal in the edit mode.

To make it work, I add a short delay before calling `focus()` on the input bar. A similar approach is used in the `ReportActionCompose`. They explain it in the comment:
```javascript
// Keyboard is not opened after Emoji Picker is closed
// SetTimeout is used as a workaround
// https://github.com/react-native-modal/react-native-modal/issues/114
// We carefully choose a delay. 100ms is found enough for the keyboard to open.
setTimeout(() => this.textInput.focus(), 100);
```

This doesn't fix anything on the `mWeb / Safari`, because Safari blocks focusing here.

### Fixed Issues

$ https://github.com/Expensify/App/issues/18001
$ PROPOSAL: https://github.com/Expensify/App/issues/18001#issuecomment-1642157870

### Tests

- [x] Verify that no errors appear in the JS console
- [x] Select a report
- [x] Type a short message in the input bar
- [x] Press the emoji button
- [x] Select a random emoji
- [x] Check if the focus is back in the input bar
- [x] Send the message
- [x] Press the edit button on the message that was sent (on native/mWeb long-press the message, then press Edit).
- [x] Make changes to the message (add/remove some characters)
- [x] Press the emoji button and select an emoji.
- [x] Verify that the focus is back in the edited box (on mWeb Chrome/native the keyboard should display).

### Offline tests

This PR doesn't affect any offline behavior.

### QA Steps

**Note:** On `mWeb/Safari` this PR doesn't do anything, because Safari doesn't let developers focus any input programmatically.

- [ ] Verify that no errors appear in the JS console
- [ ] Do the test from above
- [ ] Check focus after selecting emoji in the edit mode several times
   - [ ] Select a message to edit
   - [ ] Modify it
   - [ ] Add emoji using the emoji button
   - [ ] Confirm, that the focus is back in the correct input bar
- [ ] Check focus with multiple edit boxes
   - [ ] Open edit mode in several messages
   - [ ] Add emoji to one of the messages
   - [ ] Confirm that the emoji is added to the appropriate message, and everything works as expected
- [ ] Put the cursor in the middle of the message, add an emoji using a custom picker, and check if the cursor moved correctly.

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
[...]

PR #96901, opened 2026-07-23
[Payment due @mananjadhav] Show credited amount and both accounts for cross-border FX reimbursements

ishpaul777:
### Explanation of Change

Part of Release 1 of "Allow Companies to Pay FX Costs for Cross-Border Reimbursements", the NewDot side of #655929.

On a cross-border reimbursement the employee is credited an amount in their own deposit currency that differs from the report total, so the payment comment has to name that amount plus the company and employee accounts. Auth records those values on the action that announced the payment, so everything here reads them from that action's `originalMessage`.

- Adds the four FX fields to `OriginalMessageIOU`, and the two amount fields to `OriginalMessageReimbursed` (which already had the last-4s).
- Adds an `iou.reimbursedCrossBorder` string ("paid {amount} from account {debit} to account {credit}") across all languages.
- Renders it when `creditedAmount` and `creditedCurrency` are both present, wherever a VBBA payment is described, falling through to the existing wording otherwise: the IOU pay action NewDot shows for in-app payments, the `REIMBURSED` action OldDot payments produce, and the LHN and parent chat previews.
- Keeps `isSubmitterAddingBankAccount` ahead of the cross-border branch, since a payment retried after the submitter adds a deposit account carries both and the hold release is the useful thing to announce. Same precedence as OldDot.

No `Report` Onyx type changes: an earlier revision put these fields on the report, but a report can be paid more than once and report-level fields are shared by every attempt, so the newest payment rewrote earlier comments.

Companion PRs: Auth #23119, Auth #23258, Web-Expensify #54798.

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/655929
PROPOSAL:

### Tests

1. On a workspace with global reimbursements enabled, submit and pay a cross-border expense report (the employee's deposit currency differs from the workspace currency).
2. Verify the payment message on the report reads "paid {credited amount in the employee's currency} from account {company account last 4} to account {employee account last 4}" instead of the report total, and that the LHN and parent workspace chat previews match.
3. Pay a same-currency report and verify its message is unchanged ("paid with bank account ...").

Unit tests: `ReportUtilsTest` covers `getIOUReportActionDisplayMessage`; `ReportActionsUtilsTest` covers `getReimbursedMessage` for the cross-border wording, the submitter-added-bank-account precedence, and the fallback when an amount arrives without its currency.

### Offline tests

Same as tests. The message renders from Onyx data already on the report action.

### QA Steps

Same as Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/ed05fbea-e6b2-41e1-a72a-b551bdb5ab69

</details>


src/libs/ReportUtils.ts
```diff
@@ -5845,6 +5846,11 @@ function getReportPreviewMessage(translate: LocalizedTranslate, params: GetRepor
         actualPayerName = actualPayerName && isForListPreview && !isPreviewMessageForParentChatReport ? `${actualPayerName}:` : actualPayerName;
         const payerDisplayName = isPreviewMessageForParentChatReport ? payerName : actualPayerName;
         if (translatePhraseKey === 'iou.businessBankAccount') {
+            // Cross-border FX reimbursements report the amount credited to the employee (in their deposit currency)
+            // plus both account last-4s, since the company and employee move different currencies.
+            if (originalMessage?.creditedAmount) {
+                return getCrossBorderReimbursedMessage(translate, originalMessage, originalMessage?.accountNumber?.slice(-4));
```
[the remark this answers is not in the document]
ishpaul777: The control flow observation is right - `isPreviewMessageForParentChatReport` short-circuits to `iou.payerPaidAmount`, so this branch is not reached in that context. I don't think we should change it there though.

`payerPaidAmount` renders `formattedAmount`, which is `convertToDisplayString(totalAmount, report.currency)` - the report total in the report's own currency. A cross-border report is still denominated in the settlement currency and the company is still debited in that currency, so "paid $16.94" is accurate as a report-level summary. Substituting the employee's credited amount would make the chat row show GBP for a USD report while the report header, search, and the expense amounts all still show USD, which reads worse than what we have.

The `businessBankAccount` branch is different: it is the same wording as the payment system message ("paid with bank account 1234", no amount), so mirroring the cross-border wording there keeps the preview and the action in sync. The design doc and the issue both scope `creditedAmount` to that payment comment rather than to the report-total summary, so leaving `payerPaidAmount` alone is deliberate.

JmillsExpensify (commented):
No product review required on this PR

src/libs/ReportUtils.ts
mananjadhav, line 5824: **question:** When `isPreviewMessageForParentChatReport` is true we always force `iou.payerPaidAmount`, so the cross-border branch under `iou.businessBankAccount` (L5848–5850) never runs. Is that expected?

mananjadhav, line 6042: **question:** Same as isPreviewMessageForParentChatReport comment raised before.

src/libs/ReportUtils.ts
```diff
@@ -5845,6 +5846,9 @@ function getReportPreviewMessage(translate: LocalizedTranslate, params: GetRepor
         actualPayerName = actualPayerName && isForListPreview && !isPreviewMessageForParentChatReport ? `${actualPayerName}:` : actualPayerName;
         const payerDisplayName = isPreviewMessageForParentChatReport ? payerName : actualPayerName;
         if (translatePhraseKey === 'iou.businessBankAccount') {
+            if (originalMessage?.creditedAmount && originalMessage.creditedCurrency) {
+                return getCrossBorderReimbursedMessage(translate, originalMessage, originalMessage?.accountNumber?.slice(-4));
```
mananjadhav: **question:** The cross-border path only falls back to `originalMessage?.accountNumber?.slice(-4)`, but the non-FX path on the next line also falls back to `reportPolicy?.achAccount?.accountNumber?.slice(-4)`.

src/libs/ReportUtils.ts
```diff
@@ -6059,6 +6063,12 @@ function getReportPreviewReportActionMessage(params: GetReportPreviewMessageBase
         const payerDisplayName = isPreviewMessageForParentChatReport ? payerName : actualPayerName;
         if (translatePhraseKey === 'iou.businessBankAccount') {
             const last4Digits = originalMessage?.accountNumber?.slice(-4) ?? reportPolicy?.achAccount?.accountNumber?.slice(-4) ?? '';
+
+            // This variant returns raw English to match the surrounding non-localized preview strings.
+            if (originalMessage?.creditedAmount && originalMessage.creditedCurrency) {
+                const creditedAmountDisplay = convertToDisplayString(originalMessage.creditedAmount, originalMessage.creditedCurrency);
+                return `paid ${creditedAmountDisplay} from account ${originalMessage.debitBankAccountLast4 ?? last4Digits} to account ${originalMessage.creditBankAccountLast4 ?? ''}`;
```
mananjadhav, line 6080: **suggestion:** This string almost matches the `reimbursedCrossBorder` from en. Can we somehow push this to a helper to avoid any future drift?

src/libs/ReportUtils.ts
mananjadhav, line 10994: **question:** I migh not have full context but still validating.  PaymentContent only renders the cross-border message for `CONST.IOU.PAYMENT_TYPE.VBBA`, but here the same check runs for both `EXPENSIFY` and `VBBA`. Is that expected?

src/pages/inbox/report/actionContents/PaymentContent.tsx
```diff
@@ -40,6 +40,9 @@ function PaymentContent({action, policyID}: PaymentContentProps) {
 
     if (paymentType === CONST.IOU.PAYMENT_TYPE.VBBA) {
         const last4Digits = getBankAccountLastFourDigits(originalMessage.bankAccountID, bankAccountList, policy);
+        if (originalMessage.creditedAmount && originalMessage.creditedCurrency) {
+            return <ReportActionItemBasicMessage message={getCrossBorderReimbursedMessage(translate, originalMessage, last4Digits)} />;
```
mananjadhav: **question:** This returns the FX message before the wasAutoPaid branch, so an automatic VBBA payment with creditedAmount/creditedCurrency will not mention workspace rules.

Is dropping the auto-pay wording intentional for FX, or should we compose both (FX amount/accounts + automation note)?

src/libs/ReportActionsUtils.ts
```diff
@@ -467,6 +467,22 @@ function getElsewherePaymentReportActionMessage(translate: LocalizedTranslate, o
     return translate('iou.paidElsewhere', {payer, comment: originalMessage?.comment?.trim()});
 }
 
+/**
+ * Builds the payment message for a cross-border FX reimbursement, which reports the amount credited to the employee
+ * in their deposit currency plus both account last-4s, since the company and employee move different currencies.
+ */
+function getCrossBorderReimbursedMessage(
```
mananjadhav, line 482: **suggestion:** Callers already guard with `creditedAmount && creditedCurrency` before invoking this method. Yet we still set types both as optional, and convertToDisplayString will default missing values to 0 / USD.

tests/unit/ReportUtilsTest.ts
```diff
@@ -571,6 +571,30 @@ describe('ReportUtils', () => {
             expect(getIOUReportActionDisplayMessage(translateLocal, actionWithAccountNumber, undefined, iouReport)).toBe(paidSystemMessage);
         });
 
+        it('should show the cross-border FX message with the credited amount and both account last-4s', async () => {
```
mananjadhav, line 577: **suggestion:** Can we add tests for.

1. `getReportPreviewMessage` / `getReportPreviewReportActionMessage` with FX fields and `isPreviewMessageForParentChatReport: true` (should return iou.reimbursedCrossBorder, not iou.payerPaidAmount)
2. Localized preview when only the policy ACH last-4 is available (no debitBankAccountLast4 / accountNumber)

src/libs/ReportActionsUtils.ts
mananjadhav, line 521: **question:** If Pusher has method but not paymentMethod / creditBankAccountLast4, we return OldDot fragments before reading creditedAmount. Structured FX never runs on that update.

src/libs/ReportActionsUtils.ts
```diff
@@ -538,6 +554,15 @@ function getReimbursedMessage(
         return `${translate('iou.reimbursedSubmitterAddedBankAccount', submitterLogin)}${paymentSuffix}`;
     }
 
+    // The employee is credited in their own currency, so name that amount and both accounts, not the report total.
+    if (originalMessage.creditedAmount && originalMessage.creditedCurrency) {
```
mananjadhav: **question:** `paymentSuffix` (Fast_ACH ETA, StripeConnect timing, ACH expected date, “via check”) is computed above, then unused when `creditedAmount`/`creditedCurrency` are set. Is that expected?

src/pages/inbox/report/actionContents/PaymentContent.tsx
```diff
@@ -40,6 +40,9 @@ function PaymentContent({action, policyID}: PaymentContentProps) {
 
     if (paymentType === CONST.IOU.PAYMENT_TYPE.VBBA) {
         const last4Digits = getBankAccountLastFourDigits(originalMessage.bankAccountID, bankAccountList, policy);
+        if (originalMessage.creditedAmount && originalMessage.creditedCurrency) {
```
mananjadhav: **suggestion:** Can the creditedAmount ever be 0? Do we need to explicitly handle?

src/libs/ReportActionsUtils.ts
```diff
@@ -538,6 +554,15 @@ function getReimbursedMessage(
         return `${translate('iou.reimbursedSubmitterAddedBankAccount', submitterLogin)}${paymentSuffix}`;
```
mananjadhav, line 567: For a cross-border pay retried after the submitter adds a deposit account, NewDot may show FX “paid …” instead of the hold-release copy OldDot shows. Is the assumption correct and most importantly valid?

mananjadhav (commented):
Left a few comments. But quickly wanted to check if we're going to need any changes in the threads namely `ReportNameUtils` ?


Second, pptimistic Pay is built without `creditedAmount`/`creditedCurrency`, and `getIOUReportActionMessage` formats Pay as `paid {reportTotal} with Expensify`. That text is written to chat + expense report `lastMessageText`. Do we need to update anything on the optimisc pay front?

src/libs/ReportUtils.ts
ishpaul777: Expected. `payerPaidAmount` renders the report total in the report's own currency, and a cross-border report is still denominated in that currency. The credited amount is the only figure in the employee's deposit currency, since Auth books the withdrawal in the settlement currency. Substituting it here would make the chat row read GBP while the report header, the expense amounts and Search all still read USD.

The `businessBankAccount` branch is different: it is the same sentence as the payment system message, so mirroring the cross-border wording there keeps the preview and the action in sync. I added a test that pins this, `still names the report total in the parent chat preview`.

ishpaul777: Same answer as on the localized function. That one mirrors this branch for branch, so it stays on `payerPaidAmount` for the parent chat preview for the same reason: the parent chat row summarises the report, which is denominated in the report currency, not the employee's.

src/libs/ReportUtils.ts
ishpaul777: Good catch, fixed in e7e3748. The last-4 that the non-converted branch resolves (`originalMessage.accountNumber`, then the policy ACH account) is now hoisted and passed into the cross-border helper, so both branches name the debited account the same way. Covered by `names the credited amount, falling back to the policy default for the debited account`.

src/libs/ReportUtils.ts
ishpaul777: Keeping the raw string here. This function's contract is to hardcode English, because the text it returns is stored on a report action, and its docblock states both halves of that: hardcode the English strings, and keep them in sync with the `iou.*` entries in `en.ts`. Routing this one branch through `translate` would store locale-dependent text on the action, which is the thing the function exists to avoid.

To cover the drift you are worried about, I added an assertion that the stored English equals the localized output for the cross-border case, so CI fails if the two ever diverge.

src/libs/ReportUtils.ts
ishpaul777: The guard is on the data rather than the payment type. Only a Corpay cross-border settlement records `creditedAmount`/`creditedCurrency`: Auth writes them from `ProcessReimbursementWithCorpay`, onto the action that announced a withdrawal from the company's verified bank account, so the `EXPENSIFY` arm of that shared case never carries them.

`PaymentContent` only needs the `VBBA` branch because that is where an expense report payment renders. Keeping the check data-driven means the message stays correct if a conversion is ever booked under another payment type, instead of quietly falling back to the report total.

src/pages/inbox/report/actionContents/PaymentContent.tsx
ishpaul777: Fixed in 39021f4. The FX sentence still wins over the `automaticallyPaidWith...` copy, since combining the two would need new copy, but an auto-paid payment now keeps its muted rendering, so the automation signal is not lost. That also matches `getIOUReportActionDisplayMessage`, which returns the cross-border text ahead of its `automaticAction` branch.

src/libs/ReportActionsUtils.ts
ishpaul777: Moved the rule into the helper in 3be2b9c. It now returns `string | undefined` and owns the check, so the call sites fall through when there is nothing to render and `convertToDisplayString` can never be reached without an amount and a currency.

The parameter types stay optional because they mirror the Onyx types, and TypeScript does not carry a caller's truthiness narrowing into an object parameter.

tests/unit/ReportUtilsTest.ts
ishpaul777: Both added to `ReportUtilsTest`:

1. `still names the report total in the parent chat preview` covers the preview with the FX fields and `isPreviewMessageForParentChatReport: true`.
2. `names the credited amount, falling back to the policy default for the debited account` covers the localized preview when only the policy ACH last-4 is available, which is also the fix for your comment on :5850.

There is a third one alongside them asserting that the English stored on the report action equals the localized output, so the two copies cannot drift.

src/libs/ReportActionsUtils.ts
ishpaul777: That fallback cannot swallow a cross-border payment. Auth writes all four fields in a single `setMessage` call on the action that announced the payment, and that same call is what pushes the action to connected clients, so `creditBankAccountLast4` is always present on the update and the last condition of the guard is false.

Even if we did fall back, nothing would be lost: the server fragments already read "paid <credited amount> from account X to account Y", because OldDot's `Report_Action_Reimbursed::getMessage` renders the cross-border sentence as well.

src/libs/ReportActionsUtils.ts
ishpaul777: Expected, and it matches OldDot. `Report_Action_Reimbursed::getMessage` returns from its cross-border branch before appending the payment suffix (the Fast_ACH estimate, the check note, the StripeConnect timing, the ACH expected date), so a converted payment names the amount and the two accounts and stops there. Dropping the suffix in the same branch keeps both clients on one sentence.

src/pages/inbox/report/actionContents/PaymentContent.tsx
ishpaul777: Auth only records the pair when the payment actually converted (`if (creditedAmount && reportID)` in `ProcessReimbursementWithCorpay`), so a zero is never written. It is still handled: zero is falsy, so the helper returns undefined and we render the usual "paid with bank account ..." wording rather than "paid $0.00". That rule now lives in one place inside the helper.

src/libs/ReportActionsUtils.ts
ishpaul777: The assumption holds, and NewDot follows OldDot exactly. `Report_Action_Reimbursed::getMessage` checks `isSubmitterAddingBankAccount` first and only then the credited amount, so a cross-border payment retried after the submitter adds a deposit account shows the hold-release copy on both clients. That ordering is what f704aee changed, and `keeps the hold-release wording when the submitter adds a bank account for a cross-border payment` pins it.

mananjadhav (approved):
Covered the review code thoroughly but used dummy data for report actions. @ishpaul777 Can you please resolve the conflicts?

madmax330 (approved):

PR #97086, opened 2026-07-27
[Search Router] Add Create actions to navigation suggestions

nabi-ebrahimi:
### Explanation of Change

Adds Create/FAB actions to Search Router navigation suggestions:

- Create expense
- Create report
- Track distance
- New chat
- Send invoice
- New workspace

The suggestions reuse the existing FAB actions, icons, visibility rules, saved distance type, and Create Report branching logic. Search Router closes before executing an action, and an existing RHP is dismissed before opening the new flow to prevent modal stacking.

Navigation matching now ignores repeated whitespace, so queries such as `go to create     expense` match correctly.

Quick Action and Book Travel are intentionally excluded from this focused PR.

### Fixed Issues

### Tests

**Create action matching**

1. Sign in and open Search Router.
2. Search for `create expense`.
3. Repeat with `expense`, `go to create expense`, `go to create     expense`, and `CREATE EXPENSE`.
4. Verify `Create expense` appears with the correct icon and without a `Go to` prefix.

**Create expense**

1. Search for `create expense`.
2. Select the result.
3. Verify Search Router closes and the normal Create expense flow opens.

**Track distance**

1. Open Track distance from the FAB and select a distance type such as Map or Odometer.
2. Close the flow.
3. Open Search Router and select `Track distance`.
4. Verify the distance flow opens using the previously selected distance type.

**Create report**

1. Search for `create report` and select it.
2. Test with an account that has one eligible workspace.
3. Test with an account that has multiple eligible workspaces.
4. Verify the existing report confirmation, workspace selection, or report creation flow opens as appropriate, matching the FAB behavior.

**New chat**

1. Search for `new chat`.
2. Select the result.
3. Verify Search Router closes and the normal new-chat flow opens.

**Send invoice visibility**

1. Use an account with an invoice-enabled admin workspace.
2. Search for `invoice`.
3. Verify `Send invoice` appears and opens the invoice flow.
4. Repeat with an account without invoice permission.
5. Verify the navigation suggestion is not shown.

**New workspace visibility**

1. Use an account without a visible workspace.
2. Search for `new workspace`.
3. Verify the suggestion appears and opens workspace creation.
4. Repeat with an account where the FAB does not show New workspace.
5. Verify Search Router also hides the suggestion.

**Existing RHP replacement**

1. Open an Account or Workspace RHP.
2. Open Search Router while the RHP is open.
3. Select an available Create action.
4. Verify the previous RHP closes before the Create flow opens.
5. Verify the modals do not stack and one Back action closes the Create flow.

**Keyboard navigation**

1. Search for a Create action.
2. Select it using the arrow keys and Enter.
3. Repeat using a mouse or tap.
4. Verify both methods open the same flow.

**Excluded actions**

1. Search for `quick action`.
2. Search for `book travel`.
3. Verify no Create navigation suggestion appears for either action.

**Existing Search Router regression**

1. Search for `inbox` and select `Go to Inbox`.
2. Run a normal text search.
3. Use Ask Concierge.
4. Verify existing navigation, regular search, and Concierge flows continue working.

### Offline tests
Same as Tests.

### QA Steps
Same as Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/37d6b2c2-998c-410a-be95-bed55ff5bc4c

</details>

<details>

>

https://github.com/user-attachments/assets/67577d65-cf26-48c7-bc45-5aacfadc28a4

</details>

<details>

>

https://github.com/user-attachments/assets/220027ef-df05-4699-a774-1cf4a79fd935

</details>

<details>

>

https://github.com/user-attachments/assets/07ca9df3-e632-4e1c-8536-edeb1608fcbf

</details>

<details>

>

https://github.com/user-attachments/assets/931255c9-3cfe-47d8-85a7-14928b07bb44

https://github.com/user-attachments/assets/170fd4fa-9518-4e3a-9f72-64d1ae182cc3

https://github.com/user-attachments/assets/32cde233-c68f-4541-960c-762294863233

https://github.com/user-attachments/assets/645ec082-34b6-4163-9513-05978f510704

https://github.com/user-attachments/assets/49a4e076-002e-44e8-87ec-2ab50c233c9d

https://github.com/user-attachments/assets/62180b7b-388f-4de3-9ee6-8ff95b01635b

https://github.com/user-attachments/assets/111c0ee4-6133-47b8-91c0-41f2fea99d9b

https://github.com/user-attachments/assets/da04ffae-747a-45e7-a241-dadad93f5a17

</details>


src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,202 @@
...
+    const [draftTransactionIDs] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_DRAFT, {selector: validTransactionDraftIDsSelector});
+    const [lastDistanceExpenseType] = useOnyx(ONYXKEYS.NVP_LAST_DISTANCE_EXPENSE_TYPE);
+    const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
+    const [session] = useOnyx(ONYXKEYS.SESSION, {selector: sessionEmailAndAccountIDSelector});
+    const [allBetas] = useOnyx(ONYXKEYS.BETAS);
+    const [transactionViolations] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS);
+    const [activePolicyID] = useOnyx(ONYXKEYS.NVP_ACTIVE_POLICY_ID);
+    const [activePolicy] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${activePolicyID}`);
+    const [groupPoliciesWithChatEnabled = getEmptyArray<OnyxTypes.Policy>()] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {
+        selector: (policies: OnyxCollection<OnyxTypes.Policy>) => getGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, session?.email),
```
marufsharifi: 🔴 This feeds `useCreateReport` a **different** policy set than the FAB does, which contradicts the "matching the FAB behavior" claim:

- **FAB** (`CreateReportMenuItem.tsx`) → `chatEnabledPaidGroupPoliciesSelector`:
  `isPolicyExpenseChatEnabled && isGroupPolicy(policy) && !isJoinRequestPending && shouldShowPolicy(...)`
- **Here** → `getGroupPoliciesWhereReportCanBeCreated`:
  `!isJoinRequestPending && (isPaidGroupPolicy(policy) || canAccessSubmitWorkspaceFeatures(policy, isSubmit2026BetaEnabled)) && shouldShowPolicy(...)`

Since `useCreateReport.isVisible` is just `arePoliciesLoaded`, the row always shows — but the **post-click branching** (upgrade path when `length === 0`, workspace selector when `> 1`, and `getDefaultChatEnabledPolicy` resolution) is driven entirely by this set.

**Bug this can cause:** a paid group policy with expense-chat disabled, or a Submit-2026-beta workspace, qualifies here but not in the FAB. Selecting "Create report" from the FAB vs. the Search Router would then land the same user on *different* screens (direct create vs. workspace selector vs. upgrade).

This may be intentional (the beta-aware selector is arguably more correct). If so, please add a one-line comment noting the Search Router deliberately uses `getGroupPoliciesWhereReportCanBeCreated` rather than the FAB's older selector — otherwise this reads as copy-paste drift.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,202 @@
...
+                currentUserPersonalDetails,
+                hasViolations,
+                isBetaEnabled(CONST.BETAS.ASAP_SUBMIT),
+                defaultChatEnabledPolicy,
+                allBetas,
+                isTrackIntentUser,
+                false,
+                shouldDismissEmptyReportsConfirmation,
+            );
+            Navigation.navigate(getReportsRootRoute(), {forceReplace: isReportInSearch});
```
marufsharifi, line 135: This entire callback — the `clearLastSearchParams()` guard, `createNewReport(...)`, and the two-step `getReportsRootRoute()` → microtask → `getCreateReportRoute()` navigation — is copied line-for-line from `CreateReportMenuItem.handleCreateWorkspaceReport`. `useCreateReport` was built as the shared abstraction, yet this navigation tail now lives in two places and must be kept in lockstep.

Also, the explanatory comment from the FAB was dropped:
```ts
// Navigate to the Reports page first so getCreateReportRoute() resolves against
// the Search/Reports fullscreen context before opening the created report modal.
```
The non-obvious ordering (navigate to root *before* the modal, via a microtask) is now uncommented. Suggest extracting a shared `createReportAndNavigate(...)` helper next to `getCreateReportRoute`, or at minimum restoring that comment.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+function replaceTopmostModalWithAction(action: () => void) {
+    if (!Navigation.isTopmostRouteModalScreen()) {
+        action();
+        return;
+    }
+
+    Navigation.dismissModal({afterTransition: action});
+}
```
marufsharifi, line 71: This is the crux of the "modals don't stack" behavior and the most subtle piece of the PR, but it has no comment. A reader can't tell that (a) the Search Router is a `react-native-modal` already dismissed by the time `action()` runs, so this only ever targets an *underlying* RHP, and (b) `afterTransition` exists to avoid opening the new flow mid-dismiss-animation (which would re-introduce the stacking this prevents).

Please add a short comment explaining the rationale, e.g.:
```ts
// The Search Router modal is already hidden here, so a topmost modal route means an
// underlying RHP. Dismiss it first (and run the action only after the transition) so
// the new Create flow replaces it instead of stacking on top.
function replaceTopmostModalWithAction(action: () => void) {
```

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,202 @@
...
+    const [transactionViolations] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS);
+    const [activePolicyID] = useOnyx(ONYXKEYS.NVP_ACTIVE_POLICY_ID);
+    const [activePolicy] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${activePolicyID}`);
+    const [groupPoliciesWithChatEnabled = getEmptyArray<OnyxTypes.Policy>()] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {
+        selector: (policies: OnyxCollection<OnyxTypes.Policy>) => getGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, session?.email),
+    });
+    const [isTrackIntentUser] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED, {selector: isTrackIntentUserSelector});
+    const [isLoading = false] = useOnyx(ONYXKEYS.IS_LOADING_APP);
+
+    const defaultChatEnabledPolicy = getDefaultChatEnabledPolicy([...groupPoliciesWithChatEnabled], activePolicy);
```
marufsharifi, line 95: `[...groupPoliciesWithChatEnabled]` allocates a fresh array every render, but `getDefaultChatEnabledPolicy` only reads it — passing `groupPoliciesWithChatEnabled` directly is equivalent:

```ts
const defaultChatEnabledPolicy = getDefaultChatEnabledPolicy(groupPoliciesWithChatEnabled, activePolicy);
```

(Minor: `defaultChatEnabledPolicy` is also recomputed from the same inputs inside `useCreateReport` — harmless, just redundant.)

tests/unit/SearchRouterNavigationTest.ts
```diff
@@ -37,6 +38,12 @@ jest.mock('@components/Search/SearchContext', () => ({
     useSearchSelectionActions: () => ({clearSelectedTransactions: mockClearSelectedTransactions}),
 }));
 
+jest.mock('@components/Search/SearchRouter/useCreateNavigationSuggestions', () => ({
+    __esModule: true,
+    ...jest.requireActual<typeof CreateNavigationSuggestions>('@components/Search/SearchRouter/useCreateNavigationSuggestions'),
+    default: () => [],
+}));
```
marufsharifi, line 47: Mocking the hook's default export to `[]` means the bulk of the new logic in `useCreateNavigationSuggestions` never runs in tests: the visibility computations (`isInvoiceVisible`, `shouldShowNewWorkspaceButton`, `isCreateReportVisible`), the `startMoneyRequest` / `startDistanceRequest` argument wiring, `reportID` reuse, and the `onCreateReport` navigation sequence.

The added tests cover only `buildCreateNavigationItems` and `replaceTopmostModalWithAction` (~25–30% of the new code). Given the selector divergence flagged in `useCreateNavigationSuggestions.ts:93`, a rendered-hook test asserting which items appear for representative policy fixtures would be the highest-value addition here.

marufsharifi (changes requested):

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed by documenting why Search Router intentionally uses the shared beta-aware report eligibility rules. Thanks for the review!

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Keeping this as-is because the spread safely handles the shared helper’s readonly empty result. Removing it requires an out-of-scope shared utility type change.

tests/unit/SearchRouterNavigationTest.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,207 @@
...
+    // Use the shared report eligibility rules so Submit workspaces are only included for beta users.
+    const [groupPoliciesWithChatEnabled = getEmptyArray<OnyxTypes.Policy>()] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {
+        selector: (policies: OnyxCollection<OnyxTypes.Policy>) => getGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, session?.email),
+    });
+    const [isTrackIntentUser] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED, {selector: isTrackIntentUserSelector});
+    const [isLoading = false] = useOnyx(ONYXKEYS.IS_LOADING_APP);
+
+    const defaultChatEnabledPolicy = getDefaultChatEnabledPolicy([...groupPoliciesWithChatEnabled], activePolicy);
+    const hasViolations = hasViolationsReportUtils(undefined, transactionViolations, session?.accountID ?? CONST.DEFAULT_NUMBER_ID, session?.email ?? '');
+    const isReportInSearch = isOnSearchMoneyRequestReportPage();
```
marufsharifi: 🔴 **Blocker — this is read at render time, which makes it always `false` on native.**

The helper reads the **last root route** (`isOnSearchMoneyRequestReportPage.ts:18-20`):

```ts
const lastRootRoute = rootState.routes.at(-1);
const lastNestedRoute = lastRootRoute?.state?.routes?.at(-1);
return lastNestedRoute?.name === SCREENS.SEARCH.MONEY_REQUEST_REPORT;
```

On **native**, the Search Router is not a JS modal — it's a `RootStack.Screen` (`AuthScreens.tsx:282-291`) pushed by `openSearch` in `toggleSearch/index.native.tsx`. So while this hook renders, `routes.at(-1).name === SCREENS.SEARCH_ROUTER.ROOT` — a leaf screen with **no nested `state`** → `lastNestedRoute` is `undefined` → this returns **`false` unconditionally on every native invocation**.

This line works in `CreateReportMenuItem` because the FAB renders while the report screen is still topmost. Copied into a surface that *is* a route, it silently dies.

**Bug (iOS/Android native):** on a Search money-request report, open Search Router → `Create report`. `clearLastSearchParams()` never fires (line 112-114) and both navigations use `forceReplace: false` (lines 128, 130, 134), so the stale search params survive and the Reports root is **pushed instead of replaced**. The created report opens on top of an orphaned route: Back lands the user on the previous search report instead of the reports list, and the search view can re-hydrate the stale query. Web is correct, so this ships as a native-only regression that mWeb testing won't catch.

**Suggested fix** — evaluate it inside the action. After the router closes (route popped on native, modal hidden on web) the topmost route *is* the underlying page, so a lazy read is correct on both platforms and also removes the stale-render-value hazard:

```ts
const {createReport, isVisible: isCreateReportVisible} = useCreateReport({
    onCreateReport: (shouldDismissEmptyReportsConfirmation?: boolean) => {
        if (!defaultChatEnabledPolicy?.id) {
            return;
        }

        // Read the navigation state when the action actually runs: at render time the Search Router
        // itself is the topmost route on native, which would make this always false.
        const isReportInSearch = isOnSearchMoneyRequestReportPage();
        if (isReportInSearch) {
            clearLastSearchParams();
        }
        // …
        Navigation.navigate(getReportsRootRoute(), {forceReplace: isReportInSearch});
        Navigation.setNavigationActionToMicrotaskQueue(() => {
            Navigation.navigate(getCreateReportRoute({reportID: createdReportID}), {forceReplace: isReportInSearch});
        });
    },
    groupPoliciesWithChatEnabled,
    onNavigateToWorkspaceSelection: () => navigateToCreateReportWorkspaceSelection({forceReplace: isOnSearchMoneyRequestReportPage()}),
    shouldHandleNavigationBack: false,
});
```

This branch currently has **0% coverage** — see my comments on `useCreateNavigationSuggestionsTest.ts:141` and `:236`.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,207 @@
...
+
+function useCreateNavigationSuggestions(): SearchQueryItem[] {
+    const {translate} = useLocalize();
+    const icons = useMemoizedLazyExpensifyIcons(['Coins', 'Receipt', 'Cash', 'Transfer', 'MoneyCircle', 'Location', 'Document', 'ChatBubble', 'InvoiceGeneric', 'NewWorkspace']);
+    const currentUserPersonalDetails = useCurrentUserPersonalDetails();
+    const {isBetaEnabled} = usePermissions();
+    const isSubmit2026BetaEnabled = isBetaEnabled(CONST.BETAS.SUBMIT_2026);
+    const {isOffline} = useNetwork();
+    const {isRestrictedPolicyCreation} = usePreferredPolicy();
+    const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
```
marufsharifi, line 81: 🔴 **Blocker (perf) — unselected full POLICY collection on the Search Router's per-keystroke render path.**

`useNavigationSuggestions(textInputValue, …)` is called from `SearchRouter.tsx:168` with the **live input value**, so this whole hook tree re-runs on every keystroke.

Both consumers of `allPolicies` need exactly 8 fields:
- line 104 → `canSendInvoice(allPolicies ?? null, sessionEmail)`
- line 139 → `Object.values(allPolicies ?? {}).every((policy) => !shouldShowPolicy(policy, !!isOffline, sessionEmail))`

Which is precisely why the FAB uses `useMappedPolicies(policyMapper)` for these two items (`InvoiceMenuItem.tsx:34`, `NewWorkspaceMenuItem.tsx:40`), where `policyMapper` (`src/selectors/Policy.ts:198-208`) projects only `{type, role, id, isPolicyExpenseChatEnabled, pendingAction, avatarURL, name, areInvoicesEnabled}`. `useMappedPolicies`' own docblock spells out the tradeoff.

Subscribing raw means **any** write to **any** policy — `employeeList`, `customUnits`, `tagList`, `rules`, `connections` — re-renders the Search Router and rebuilds the entire suggestion list.

**Bug:** typing in the Search Router on an account with many workspaces produces dropped keystrokes and visible suggestion-list flicker, because the list re-derives on unrelated background Onyx writes rather than on input. Per `CLAUDE.md` this is `@frontend-performance` territory for a hot render path.

**Suggested fix:**

```ts
const [mappedPolicies] = useMappedPolicies(policyMapper);

const isInvoiceVisible = canSendInvoice(mappedPolicies as OnyxCollection<OnyxTypes.Policy>, sessionEmail);
const shouldShowNewWorkspaceButton =
    !isRestrictedPolicyCreation && !isLoading && Object.values(mappedPolicies ?? {}).every((policy) => !shouldShowPolicy(policy, !!isOffline, sessionEmail));
```

Counting this hook plus `useCreateReport` → `useShouldShowEmptyReportConfirmation` / `useCreateEmptyReportConfirmation`, this PR adds roughly **20 Onyx subscriptions** to the router's render path, including **3 separate subscriptions to `ONYXKEYS.COLLECTION.POLICY`** (raw here, selector at line 95, metadata inside `useCreateReport`) and **2 to `ONYXKEYS.SESSION`** (lines 88-89). Worth a consolidation pass.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,207 @@
...
+    const [activePolicy] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${activePolicyID}`);
+    // Use the shared report eligibility rules so Submit workspaces are only included for beta users.
+    const [groupPoliciesWithChatEnabled = getEmptyArray<OnyxTypes.Policy>()] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {
+        selector: (policies: OnyxCollection<OnyxTypes.Policy>) => getGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, session?.email),
+    });
+    const [isTrackIntentUser] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED, {selector: isTrackIntentUserSelector});
+    const [isLoading = false] = useOnyx(ONYXKEYS.IS_LOADING_APP);
+
+    const defaultChatEnabledPolicy = getDefaultChatEnabledPolicy([...groupPoliciesWithChatEnabled], activePolicy);
+    const hasViolations = hasViolationsReportUtils(undefined, transactionViolations, session?.accountID ?? CONST.DEFAULT_NUMBER_ID, session?.email ?? '');
```
marufsharifi: 🔴 **Blocker (perf) — a full-collection subscription to compute a value that is provably always `false`.**

`hasViolationsReportUtils(undefined, …)` calls `getReportTransactions(undefined)`, which short-circuits at `ReportUtils.ts:1279-1281`:

```ts
function getReportTransactions(reportID: string | undefined, …): Transaction[] {
    if (!reportID) {
        return [];
    }
    …
}
```

→ `[].some(…)` → **`hasViolations` is always `false`.** So the hook pays an unselected subscription to `ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS` — which on a high-traffic account churns continuously as violations stream in over Pusher — purely to hand a literal `false` to `createNewReport` at line 118.

Combined with the per-keystroke render path (see my comment on line 84), every inbound violation update re-renders the Search Router while the user is typing.

This is copied verbatim from `CreateReportMenuItem.tsx:70`; the copy is wrong there too, but importing it into the search input path is where it starts costing renders.

**Suggested fix** — drop the subscription:

```ts
// hasViolations is always false here: the draft report has no transactions yet, so
// hasViolationsReportUtils(undefined, …) short-circuits on getReportTransactions(undefined) === [].
const hasViolations = false;
```

If you'd rather not hardcode it, keep the call but pass `undefined` for `transactionViolations` — identical result, and the intent is clearer than a collection read that cannot affect the outcome.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,207 @@
...
+    const {isBetaEnabled} = usePermissions();
+    const isSubmit2026BetaEnabled = isBetaEnabled(CONST.BETAS.SUBMIT_2026);
+    const {isOffline} = useNetwork();
+    const {isRestrictedPolicyCreation} = usePreferredPolicy();
+    const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
+    const [reportID] = useState(() => generateReportID());
+    const [draftTransactionIDs] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_DRAFT, {selector: validTransactionDraftIDsSelector});
+    const [lastDistanceExpenseType] = useOnyx(ONYXKEYS.NVP_LAST_DISTANCE_EXPENSE_TYPE);
+    const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
+    const [session] = useOnyx(ONYXKEYS.SESSION, {selector: sessionEmailAndAccountIDSelector});
```
marufsharifi: 🟡 Two subscriptions to the same Onyx key:

```ts
const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
const [session] = useOnyx(ONYXKEYS.SESSION, {selector: sessionEmailAndAccountIDSelector});
```

`sessionEmailAndAccountIDSelector` already yields `email`, so `session?.email` **is** `sessionEmail`. On a hook that re-runs per keystroke, this is a second subscription + a second selector invocation for a value you already hold.

```ts
const [session] = useOnyx(ONYXKEYS.SESSION, {selector: sessionEmailAndAccountIDSelector});
const sessionEmail = session?.email;
```

Then drop the `emailSelector` import at line 33.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -33,7 +33,10 @@ function isNavigationIntentOnlyQuery(query: string) {
 }
 
 function toMatchKey(value: string | undefined) {
-    return StringUtils.normalizeForMatch(value ?? '').toLowerCase();
+    return StringUtils.normalizeForMatch(value ?? '')
+        .toLowerCase()
+        .replaceAll(/\s+/g, ' ')
+        .trim();
```
marufsharifi: 🟡 **This change is a no-op for the behavior the PR description attributes to it.**

The PR body says: *"Navigation matching now ignores repeated whitespace, so queries such as `go to create     expense` match correctly."* That already works on `main` — `buildNavigationSuggestions` normalizes the query **before** it ever reaches `toMatchKey`:

```ts
const matchQuery = stripNavigationIntentPrefix(trimmedQuery);
// stripNavigationIntentPrefix → query.trim().replaceAll(/\s+/g, ' ')   ← line 21 of this file
```

So `"go to create     expense"` → `"go to create expense"` → `"create expense"` before any matching happens. And `matchesNavigationQuery` / `matchesNavigationQueryExactly` have **no other callers in `src/`** — only `buildNavigationSuggestions` and `SearchRouterNavigationTest`.

On the value side, `normalizeForMatch` is `removeZeroWidthCharacters(normalizeAccents(text))` (`StringUtils/index.ts:122-124`) — it doesn't introduce whitespace, and the translated labels have single spaces.

Not harmful, but it makes a shared helper look load-bearing when it isn't, and the changelog claim will mislead whoever next debugs matching. Please either:
- revert it and confirm `go to create     expense` still matches (it should), **or**
- keep it and add a comment naming the actual input it guards against.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,207 @@
...
+function replaceTopmostModalWithAction(action: () => void) {
+    if (!Navigation.isTopmostRouteModalScreen()) {
+        action();
+        return;
+    }
+
+    Navigation.dismissModal({afterTransition: action});
+}
+
+function useCreateNavigationSuggestions(): SearchQueryItem[] {
```
marufsharifi: 🟡 **The declared return type erases `matchTerms`, which is the field that makes these rows findable.**

`buildCreateNavigationItems` (line 53) correctly returns `NavigationSuggestionSourceItem[]`, and every item it produces carries `matchTerms: [text]` (line 61). Widening to `SearchQueryItem[]` compiles — `matchTerms?` is optional, so it's assignable in both directions — but throws away the guarantee: if a refactor drops `matchTerms`, TypeScript stays silent and every Create row becomes unsearchable. That's exactly the failure the compiler should be catching.

The sibling builders in `useNavigationSuggestions.tsx` (`buildTopLevelNavigationItems`, `buildSpendNavigationItems`) are both typed `NavigationSuggestionSourceItem[]`.

```ts
function useCreateNavigationSuggestions(): NavigationSuggestionSourceItem[] {
```

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,207 @@
...
+    const [lastDistanceExpenseType] = useOnyx(ONYXKEYS.NVP_LAST_DISTANCE_EXPENSE_TYPE);
+    const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
+    const [session] = useOnyx(ONYXKEYS.SESSION, {selector: sessionEmailAndAccountIDSelector});
+    const [allBetas] = useOnyx(ONYXKEYS.BETAS);
+    const [transactionViolations] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS);
+    const [activePolicyID] = useOnyx(ONYXKEYS.NVP_ACTIVE_POLICY_ID);
+    const [activePolicy] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${activePolicyID}`);
+    // Use the shared report eligibility rules so Submit workspaces are only included for beta users.
+    const [groupPoliciesWithChatEnabled = getEmptyArray<OnyxTypes.Policy>()] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {
+        selector: (policies: OnyxCollection<OnyxTypes.Policy>) => getGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, session?.email),
```
marufsharifi: 🟡 **`Create report` eligibility diverges from the FAB, which contradicts the PR description.**

The PR states the suggestions *"reuse the existing FAB actions, icons, visibility rules … and Create Report branching logic."* For Create report they don't.

The FAB's `CreateReportMenuItem` uses a local selector requiring `isPolicyExpenseChatEnabled` (`CreateReportMenuItem.tsx:35-53`):

```ts
if (!policy?.isPolicyExpenseChatEnabled || policy?.isJoinRequestPending || !isGroupPolicy(policy) || !shouldShowPolicy(policy, false, currentUserLogin)) { continue; }
```

`getGroupPoliciesWhereReportCanBeCreated` **does not check `isPolicyExpenseChatEnabled`** (`PolicyUtils.ts:2704-2715`):

```ts
!policy.isJoinRequestPending && (isPaidGroupPolicy(policy) || canAccessSubmitWorkspaceFeatures(policy, isSubmit2026BetaEnabled)) && shouldShowPolicy(policy, false, currentUserLogin)
```

**Bug:** for a paid group workspace with expense chat disabled, the FAB correctly routes to workspace selection / upgrade, while the Search Router treats that workspace as the eligible default and hands it to `createNewReport` — creating a report against a workspace with no expense chat. Two entry points to the same flow, different outcomes.

This is defensible as **deliberate** — it matches `QuickCreationActionsBar/index.tsx:67` and `SearchActionsBarCreateButton.tsx:59`, and it's what earns Submit-beta support — and your comment gestures at it. But then the PR body is wrong and the reviewer will test the wrong thing. Please either:
- extend the comment and the description to name the divergence explicitly (*"intentionally uses the shared helper rather than the FAB's stricter `isPolicyExpenseChatEnabled` selector, matching QuickCreationActionsBar/SearchActionsBarCreateButton"*), **or**
- align both surfaces on one helper in a follow-up and link it here.

tests/unit/useCreateNavigationSuggestionsTest.ts
```diff
@@ -0,0 +1,240 @@
...
+}));
+
+jest.mock('@libs/ReportUtils', () => ({
+    generateReportID: jest.fn(() => 'draft-report'),
+    hasViolations: jest.fn(() => false),
+}));
+
+jest.mock('@navigation/helpers/isOnSearchMoneyRequestReportPage', () => ({
+    __esModule: true,
+    default: () => false,
```
marufsharifi: 🔴 Hard-mocking `isOnSearchMoneyRequestReportPage` to `false` leaves the entire search-report branch of `onCreateReport` at **0% coverage** — the `clearLastSearchParams()` call and both `{forceReplace: isReportInSearch}` navigations (`useCreateNavigationSuggestions.ts:112-114, 128, 130, 134`).

That's the branch that is broken on native (see my comment on `useCreateNavigationSuggestions.ts:103`). A test with this returning `true` would have caught it:

```ts
const mockIsOnSearchMoneyRequestReportPage = jest.fn(() => false);

jest.mock('@navigation/helpers/isOnSearchMoneyRequestReportPage', () => ({
    __esModule: true,
    default: () => mockIsOnSearchMoneyRequestReportPage(),
}));
```

…then assert the positive case (see my comment on line 236).

tests/unit/SearchRouterNavigationTest.ts
```diff
@@ -37,6 +38,12 @@ jest.mock('@components/Search/SearchContext', () => ({
     useSearchSelectionActions: () => ({clearSelectedTransactions: mockClearSelectedTransactions}),
 }));
 
+jest.mock('@components/Search/SearchRouter/useCreateNavigationSuggestions', () => ({
+    __esModule: true,
+    ...jest.requireActual<typeof CreateNavigationSuggestions>('@components/Search/SearchRouter/useCreateNavigationSuggestions'),
+    default: () => [],
```
marufsharifi: 🟡 Mocking the hook to `() => []` means the **three-source integration is never exercised** — which is exactly where the truncation bug I flagged on `useNavigationSuggestions.tsx:166` lives.

The mock is reasonable for isolating the top-level/spend suites, but please add one case that runs the real pipeline with all three sources:

```ts
it('keeps Create rows reachable when top-level and Spend sources are also present', () => {
    const createItems = CreateNavigationSuggestions.buildCreateNavigationItems(/* visible rows */);
    const suggestions = buildNavigationSuggestions('go to', [topLevelItems, spendItems, createItems], localeCompare);

    expect(suggestions.some((item) => item.keyForList?.startsWith('create_'))).toBe(true);
});
```

Without this, the suite is green while the feature is invisible for the bare `go to` query.

tests/unit/SearchRouterNavigationTest.ts
```diff
@@ -288,6 +297,63 @@ describe('top-level Search Router navigation source', () => {
...
+        expect(items.some((item) => item.keyForList === 'create_invoice' || item.keyForList === 'create_workspace')).toBe(false);
+        expect(items.some((item) => item.keyForList === 'create_travel' || item.keyForList === 'create_quickAction')).toBe(false);
+    });
+
+    it('matches Create rows through the existing navigation suggestion pipeline', () => {
+        const items = CreateNavigationSuggestions.buildCreateNavigationItems(createItems);
+
+        expect(buildNavigationSuggestions('expense', [items], localeCompare).map((item) => item.keyForList)).toEqual(['create_expense']);
+        expect(buildNavigationSuggestions('go to track distance', [items], localeCompare).map((item) => item.keyForList)).toEqual(['create_trackDistance']);
+        expect(buildNavigationSuggestions('go to create     expense', [items], localeCompare).map((item) => item.keyForList)).toEqual(['create_expense']);
```
marufsharifi: 🟡 Two problems with this block:

1. **Single-source only.** Passing `[items]` means the `slice(0, MAX_NAVIGATION_SUGGESTIONS)` interaction with `topLevelItems` + `spendItems` is never hit — see my comment on line 44.

2. **Line 332 is a tautology w.r.t. the `toMatchKey` change.** `buildNavigationSuggestions` already collapses the query via `stripNavigationIntentPrefix` (`SearchRouterHelpers.ts:21`) before `toMatchKey` runs, so `'go to create     expense'` passes both with and without this PR's `.replaceAll(/\s+/g, ' ')` addition. It doesn't pin the new behavior.

If the `toMatchKey` change is kept, please test it where it actually applies — against a **value** with irregular whitespace:

```ts
expect(matchesNavigationQueryExactly('create expense', 'Create  expense')).toBe(true);
```

tests/unit/useCreateNavigationSuggestionsTest.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    it('creates a report and navigates through the Reports root', () => {
+        mockGetGroupPoliciesWhereReportCanBeCreated.mockReturnValue([submitPolicy]);
+        renderHook(() => useCreateNavigationSuggestions());
+
+        const onCreateReport = mockUseCreateReport.mock.calls.at(0)?.at(0)?.onCreateReport;
+        act(() => onCreateReport?.(true));
+
+        expect(createNewReport).toHaveBeenCalledWith(expect.anything(), false, true, submitPolicy, [], false, false, true);
+        expect(clearLastSearchParams).not.toHaveBeenCalled();
```
marufsharifi, line 278: 🟡 This asserts only the **negative**. Please add the mirror test so the search-report path is actually pinned:

```ts
it('clears stale search params and replaces routes when creating from a search report', () => {
    mockIsOnSearchMoneyRequestReportPage.mockReturnValue(true);
    mockGetGroupPoliciesWhereReportCanBeCreated.mockReturnValue([submitPolicy]);
    renderHook(() => useCreateNavigationSuggestions());

    const onCreateReport = mockUseCreateReport.mock.calls.at(0)?.at(0)?.onCreateReport;
    act(() => onCreateReport?.(true));

    expect(clearLastSearchParams).toHaveBeenCalledTimes(1);
    expect(Navigation.navigate).toHaveBeenNthCalledWith(1, 'reports', {forceReplace: true});
    expect(Navigation.navigate).toHaveBeenNthCalledWith(2, 'report/created-report', {forceReplace: true});
});
```

Note this test will only be meaningful once the `isOnSearchMoneyRequestReportPage()` call moves inside the action — right now the render-time read makes the native behavior untestable by construction.

tests/perf-test/SearchRouter.perf-test.tsx
```diff
@@ -81,6 +84,7 @@ jest.mock('@react-navigation/native', () => {
             isReady: () => jest.fn(),
             getCurrentRoute: () => jest.fn(),
             getState: () => jest.fn(),
+            getRootState: () => undefined,
```
marufsharifi, line 87: 🟡 This mock addition is needed because the new hook reaches into the navigation state — which is a good signal that this perf test is now measuring a materially heavier `SearchRouter`.

Given the PR adds ~20 Onyx subscriptions to the router's render path (including an unselected `COLLECTION.POLICY` and an unselected `COLLECTION.TRANSACTION_VIOLATIONS` — see my comments on `useCreateNavigationSuggestions.ts:84` and `:91-102`), **please paste the Reassure before/after for `SearchRouter` into the PR description.** This file exists to gate exactly this regression, and right now the diff only updates mock plumbing without reporting the numbers.

Also worth noting: `getRootState: () => undefined` means `isOnSearchMoneyRequestReportPage()` short-circuits in the perf run, so the measurement is slightly optimistic versus production.

marufsharifi (changes requested):

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Thanks for flagging this. The performance concern is valid, but `useMappedPolicies()` still subscribes to the raw policy collection and maps it afterward, so it would not prevent rerenders from unrelated policy updates. The suggested casts also treat partial mapped policies as complete `Policy` objects, while `shouldShowPolicy()` can read fields excluded by `policyMapper`. I suggest keeping the current behavior here and handling policy subscription consolidation separately with a properly typed selector.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

tests/unit/useCreateNavigationSuggestionsTest.ts
nabi-ebrahimi: Addressed.

tests/unit/SearchRouterNavigationTest.ts
nabi-ebrahimi: Addressed.

tests/unit/SearchRouterNavigationTest.ts
nabi-ebrahimi: Addressed.

tests/unit/useCreateNavigationSuggestionsTest.ts
nabi-ebrahimi: Addressed.

tests/perf-test/SearchRouter.perf-test.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,204 @@
...
+                replaceTopmostModalWithAction(() => {
+                    interceptAnonymousUser(() => {
+                        startDistanceRequest(CONST.IOU.TYPE.CREATE, reportID, draftTransactionIDs, lastDistanceExpenseType, undefined, undefined, true);
+                    });
+                }),
+            keyForList: 'create_trackDistance',
+        },
+        {
+            visible: true,
+            text: translate('sidebarScreen.fabNewChat'),
```
marufsharifi, line 186: 🔴 **Blocking — this row is unreachable by its documented search term.**

`sidebarScreen.fabNewChat` resolves to **`'Start chat'`**, not `'New chat'`:

https://github.com/Expensify/App/blob/main/src/languages/en.ts#L1001

`buildCreateNavigationItems` sets `matchTerms: [text]`, and matching is a plain substring test on that single term:

```ts
// SearchRouterHelpers.ts
return values.some((value) => toMatchKey(value).includes(normalizedQuery));
```

So `'start chat'.includes('new chat')` → **`false`**.

**Impact:** the PR's own test step — *"**New chat** — 1. Search for `new chat`. 2. Select the result."* — cannot pass. The unit tests don't catch it because `translate` is mocked to return the key and the fixture hardcodes a string production never emits (see my comment on `SearchRouterNavigationTest.ts:308`).

The same gap applies to every other row: `create workspace` won't match "New workspace", `new expense` / `add expense` won't match "Create expense".

**Suggested fix** — make aliases first-class rather than deriving them from the label:

```ts
type CreateNavigationItem = {
    visible: boolean;
    text: string;
    icon: IconAsset;
    action: () => void;
    keyForList: string;
    /** Extra localized terms users are likely to type. Defaults to [text]. */
    matchTerms?: string[];
};

function buildCreateNavigationItems(items: CreateNavigationItem[]): NavigationSuggestionSourceItem[] {
    return items
        .filter((item) => item.visible)
        .map(({text, icon, action, keyForList, matchTerms}) => ({
            text,
            singleIcon: icon,
            action,
            keyForList,
            matchTerms: matchTerms ?? [text],
        }));
}
```

Aliases **must** come from translation keys, not inline English literals — otherwise the router breaks for every non-EN locale. If no suitable second key exists, this needs a copy request rather than a hardcoded string.

tests/unit/SearchRouterNavigationTest.ts
```diff
@@ -288,9 +299,67 @@ describe('top-level Search Router navigation source', () => {
...
+    ];
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+    });
+
+    it('builds visible Create rows with direct action labels and excludes unavailable items', () => {
+        const items = CreateNavigationSuggestions.buildCreateNavigationItems(createItems);
+
+        expect(items.map((item) => item.text)).toEqual(['Create expense', 'Create report', 'Track distance', 'New chat']);
```
marufsharifi: 🔴 **These fixtures assert against a label production never produces**, which is what let the `New chat` bug through.

Line 308 declares `text: 'New chat'` and line 320 asserts `.toEqual([..., 'New chat'])`, but `translate('sidebarScreen.fabNewChat')` returns **`'Start chat'`**. Combined with `translate: (key) => key` in `useCreateNavigationSuggestionsTest.ts:63`, no test in this PR ever observes a real label.

Please seed the fixtures from the actual translations so label/matching regressions are visible:

```ts
import * as Localize from '@libs/Localize';

const createItems: CreateNavigationSuggestions.CreateNavigationItem[] = [
    {visible: true, text: Localize.translateLocal('iou.createExpense'), icon: mockIcon, action: createAction, keyForList: 'create_expense'},
    {visible: true, text: Localize.translateLocal('report.newReport.createReport'), ...},
    {visible: true, text: Localize.translateLocal('iou.trackDistance'), ...},
    {visible: true, text: Localize.translateLocal('sidebarScreen.fabNewChat'), ...},
    // ...
];
```

and add the case that currently fails:

```ts
it('matches the New chat row by the terms users actually type', () => {
    const items = CreateNavigationSuggestions.buildCreateNavigationItems(createItems);
    expect(buildNavigationSuggestions('new chat', [items], localeCompare).map((i) => i.keyForList)).toEqual(['create_chat']);
});
```

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,204 @@
...
+
+            // Read this after Search Router closes so native resolves the underlying route instead of the router screen itself.
+            const isReportInSearch = isOnSearchMoneyRequestReportPage();
+            if (isReportInSearch) {
+                clearLastSearchParams();
+            }
+
+            const {reportID: createdReportID} = createNewReport(
+                currentUserPersonalDetails,
+                false,
```
marufsharifi, line 125: 🔴 **Blocking — `hasViolations` is hardcoded to `false`; every other call site computes it.**

This argument is not decorative. `createNewReport` threads it into `buildNewReportOptimisticData` → `buildOptimisticNextStep({hasViolations})`, which renders the new report's **next-step banner**:

https://github.com/Expensify/App/blob/main/src/libs/actions/Report/index.ts#L4348-L4372

All three sibling call sites compute it:

- `src/pages/inbox/sidebar/FABPopoverContent/menuItems/CreateReportMenuItem.tsx`
- `src/components/Search/SearchPageHeader/SearchActionsBarCreateButton.tsx`
- `src/components/Navigation/QuickCreationActionsBar/index.tsx`

**Concrete bug:** a user with outstanding violations types `create report` here and gets a next-step banner computed as if they had none (e.g. *"Waiting for you to submit"* instead of the fix-violations guidance). The identical action from the FAB two seconds earlier renders the correct banner. **Offline the wrong banner persists** until reconnect, since there's no server response to correct the optimistic value. That's a state divergence driven purely by which entry point the user picked.

**Fix** — mirror `CreateReportMenuItem`:

```ts
import {hasViolations as hasViolationsReportUtils} from '@libs/ReportUtils';
import {sessionEmailAndAccountIDSelector} from '@src/selectors/Session';

const [session] = useOnyx(ONYXKEYS.SESSION, {selector: sessionEmailAndAccountIDSelector});
const [transactionViolations] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_VIOLATIONS);
const hasViolations = hasViolationsReportUtils(undefined, transactionViolations, session?.accountID ?? CONST.DEFAULT_NUMBER_ID, session?.email ?? '');
```

then pass `hasViolations` here instead of `false`. Bonus: `sessionEmailAndAccountIDSelector` covers both fields, so the separate `emailSelector` subscription on line 87 can be dropped.

tests/unit/useCreateNavigationSuggestionsTest.ts
```diff
@@ -0,0 +1,260 @@
...
+    canSendInvoice: (...args: unknown[]) => mockCanSendInvoice(...args),
+    getDefaultChatEnabledPolicy: (policies: unknown[]) => mockGetDefaultChatEnabledPolicy(policies),
+    getGroupPoliciesWhereReportCanBeCreated: (policies: unknown, isSubmit2026BetaEnabled: boolean, currentUserLogin?: string) =>
+        mockGetGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, currentUserLogin),
+    shouldShowPolicy: (...args: unknown[]) => mockShouldShowPolicy(...args),
+}));
+
+jest.mock('@libs/ReportUtils', () => ({
+    generateReportID: jest.fn(() => 'draft-report'),
+    hasViolations: jest.fn(() => false),
```
marufsharifi: 🔴 This mock is orphaned — `useCreateNavigationSuggestions.ts` never imports `hasViolations`. It looks like the wiring was started and dropped, which is exactly the bug flagged on `useCreateNavigationSuggestions.ts:115`.

Once `hasViolations` is actually wired up, please assert it rather than just mocking it. The existing assertion on line 238 currently locks in the wrong value:

```ts
expect(createNewReport).toHaveBeenCalledWith(expect.anything(), false, true, submitPolicy, [], false, false, true);
//                                                              ^^^^^ asserts the bug
```

Suggested:

```ts
it('forwards the account violation state when creating a report', () => {
    mockHasViolations.mockReturnValue(true);
    mockGetGroupPoliciesWhereReportCanBeCreated.mockReturnValue([submitPolicy]);
    renderHook(() => useCreateNavigationSuggestions());

    act(() => mockUseCreateReport.mock.calls.at(0)?.at(0)?.onCreateReport?.(true));

    expect(createNewReport).toHaveBeenCalledWith(expect.anything(), true, expect.anything(), submitPolicy, [], false, false, true);
});
```

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,204 @@
...
+            // the Search/Reports fullscreen context before opening the created report modal.
+            Navigation.navigate(getReportsRootRoute(), {forceReplace: isReportInSearch});
+            Navigation.setNavigationActionToMicrotaskQueue(() => {
+                Navigation.navigate(getCreateReportRoute({reportID: createdReportID}), {forceReplace: isReportInSearch});
+            });
+        },
+        groupPoliciesWithChatEnabled,
+        onNavigateToWorkspaceSelection: () => navigateToCreateReportWorkspaceSelection({forceReplace: isOnSearchMoneyRequestReportPage()}),
+        shouldHandleNavigationBack: false,
+    });
```
marufsharifi, line 144: 🟠 **Perf: this pulls whole-collection Onyx selectors into the Search Router's render path for its entire lifetime.**

This hook is called unconditionally from `useNavigationSuggestions` (line 132 of that file), which runs on every Search Router render regardless of the query. `useCreateReport` transitively mounts [`useShouldShowEmptyReportConfirmation`](https://github.com/Expensify/App/blob/main/src/hooks/useShouldShowEmptyReportConfirmation.ts):

```ts
const [reportIDsWithActiveTransactions] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION, {selector: reportIDsWithActiveTransactionsSelector});
const [hasEmptyReport = false] = useOnyx(ONYXKEYS.COLLECTION.REPORT, {
    selector: (reports) => hasEmptyReportsForPolicy(reports, policyID, reportIDsWithActiveTransactions ?? {}, accountID),
});
```

`reportIDsWithActiveTransactionsSelector` allocates a `Record<string, boolean>` over **every transaction in Onyx**, and `useOnyx` deep-equals selector results on each store update — the exact cost `useMappedPolicies` exists to warn about:

> *Unlike passing a selector directly to `useOnyx` (which triggers expensive `deepEqual` comparisons on the entire mapped collection)…*

**Why it matters on this surface specifically:** the Search Router open path is instrumented with dedicated Sentry spans (`SPAN_OPEN_SEARCH_ROUTER`, `SPAN_SEARCH_ROUTER_LIST_RENDER`, `SPAN_SEARCH_ROUTER_MODAL_CLOSE_WAIT` in `SearchRouterContext.tsx`) precisely because it's latency-sensitive. On a High Traffic account we now scan the full report **and** transaction collections on router open, and re-run them on every Pusher-driven transaction update **while the user is typing** — to decide whether one row out of eight should render.

**Ask:** either defer the report-creation machinery until the query plausibly matches, or split so the empty-report-confirmation wiring resolves at action time rather than render time. If neither is practical, please get `@frontend-performance` sign-off and link it in the PR description per `CLAUDE.md`, and post before/after numbers from `tests/perf-test/SearchRouter.perf-test.tsx` (you already touched that file).

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -161,7 +163,7 @@ function useNavigationSuggestions(query: string, shouldWatchForApprovals = true)
         onSelect: (searchQuery) => navigateToCannedSpendSearch(searchQuery, clearSelectedTransactions),
     });
 
-    return buildNavigationSuggestions(query, [topLevelItems, spendItems], localeCompare);
+    return buildNavigationSuggestions(query, [topLevelItems, spendItems, createItems], localeCompare);
```
marufsharifi, line 166: 🟠 **Create rows are silently truncated away for bare `go` / `go to`.**

`buildNavigationSuggestions` flattens sources **in order** and hard-caps the result:

```ts
const MAX_NAVIGATION_SUGGESTIONS = 8;   // SearchRouterHelpers.ts:15
...
return sources
    .flatMap((source) => sortNavigationSuggestionItems(source, localeCompare))
    .filter((item) => isNavigationIntentOnly || ...)
    ...
    .slice(0, MAX_NAVIGATION_SUGGESTIONS);
```

`isNavigationIntentOnly` (query is exactly `go` or `go to`) is the documented *"show every destination"* case. `topLevelItems` is always 5, and `spendItems` is ≥ 2–3 for any account — considerably more for an admin with the accounting section (`SearchUIUtils.ts:4717+`). Appending `createItems` last means the 8-slot budget is exhausted before a single Create row is reached.

**Bug in the app:** a user types `go to`, sees the full destination list, and concludes Create actions aren't in the router at all.

**Fix** — either order Create ahead of Spend:

```ts
return buildNavigationSuggestions(query, [topLevelItems, createItems, spendItems], localeCompare);
```

or guarantee per-source representation for intent-only queries inside `buildNavigationSuggestions`:

```ts
// Intent-only queries advertise what's available, so every source must be represented.
const budgetPerSource = Math.max(1, Math.floor(MAX_NAVIGATION_SUGGESTIONS / sources.length));
```

Either way please add a test asserting at least one `create_*` key survives `buildNavigationSuggestions('go to', [topLevel, spend, create], localeCompare)`.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,204 @@
...
+    const [activePolicy] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${activePolicyID}`);
+    // Match Quick Creation and Search actions by using shared report eligibility rules. Unlike the FAB's
+    // expense-chat filter, this intentionally includes Submit workspaces only when the beta is enabled.
+    const [groupPoliciesWithChatEnabled = getEmptyArray<OnyxTypes.Policy>()] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {
+        selector: (policies: OnyxCollection<OnyxTypes.Policy>) => getGroupPoliciesWhereReportCanBeCreated(policies, isSubmit2026BetaEnabled, sessionEmail),
+    });
+    const [isTrackIntentUser] = useOnyx(ONYXKEYS.NVP_INTRO_SELECTED, {selector: isTrackIntentUserSelector});
+    const [isLoading = false] = useOnyx(ONYXKEYS.IS_LOADING_APP);
+
+    const defaultChatEnabledPolicy = getDefaultChatEnabledPolicy([...groupPoliciesWithChatEnabled], activePolicy);
```
marufsharifi: 🟠 **Second subscription to the same key — this value is derivable from `allPolicies` on line 82.**

You already hold the raw collection. The selector variant adds another subscription plus a `deepEqual` over the returned policy array on every policy write, for a value that's free to compute. `EmptySearchView.tsx:90` already does it the cheap way:

```ts
const groupPoliciesWithChatEnabled = getGroupPoliciesWhereReportCanBeCreated(allPolicies, isBetaEnabled(CONST.BETAS.SUBMIT_2026));
```

**Suggested:**

```ts
const groupPoliciesWithChatEnabled = getGroupPoliciesWhereReportCanBeCreated(allPolicies ?? null, isSubmit2026BetaEnabled, sessionEmail);
```

React Compiler memoizes this on `allPolicies` identity, so `useCreateReport`'s `useCallback` deps stay stable. This also lets you drop the `[...groupPoliciesWithChatEnabled]` spread on line 98 — one fewer array allocated per render.

👍 Separately: `getEmptyArray<OnyxTypes.Policy>()` here is a genuine improvement over the `CONST.EMPTY_ARRAY` + `as Array<OnyxEntry<Policy>>` cast used by the three sibling call sites. Worth propagating.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
```diff
@@ -33,7 +33,9 @@ function isNavigationIntentOnlyQuery(query: string) {
 }
 
 function toMatchKey(value: string | undefined) {
-    return StringUtils.normalizeForMatch(value ?? '').toLowerCase();
+    return StringUtils.normalizeForMatch(value ?? '')
+        .toLowerCase()
+        .trim();
```
marufsharifi: 🟠 **This doesn't do what the PR description claims, and nothing tests it.**

From the PR body:

> *Navigation matching now ignores repeated whitespace, so queries such as `go to create     expense` match correctly.*

That was already true on `main`. `stripNavigationIntentPrefix` collapses internal whitespace on line 21 of this same file:

```ts
const trimmedQuery = query.trim().replaceAll(/\s+/g, ' ');
```

So `'go to create     expense'` → `'create expense'` **before** `toMatchKey` is ever reached. `.trim()` collapses nothing internal — and on the value side it's a no-op too, since none of the six new labels has leading or trailing whitespace (`normalizeForMatch` only strips diacritics and zero-width chars).

**Ask:** Add a case that fails without it. Reviewers use the description to decide what to regression-test; as written it points at the wrong risk area.

tests/unit/SearchRouterNavigationTest.ts
```diff
@@ -288,9 +299,67 @@ describe('top-level Search Router navigation source', () => {
...
+        expect(items.some((item) => item.keyForList === 'create_invoice' || item.keyForList === 'create_workspace')).toBe(false);
+        expect(items.some((item) => item.keyForList === 'create_travel' || item.keyForList === 'create_quickAction')).toBe(false);
+    });
+
+    it('matches Create rows through the existing navigation suggestion pipeline', () => {
+        const items = CreateNavigationSuggestions.buildCreateNavigationItems(createItems);
+
+        expect(buildNavigationSuggestions('expense', [items], localeCompare).map((item) => item.keyForList)).toEqual(['create_expense']);
+        expect(buildNavigationSuggestions('go to track distance', [items], localeCompare).map((item) => item.keyForList)).toEqual(['create_trackDistance']);
+        expect(buildNavigationSuggestions('go to create     expense', [items], localeCompare).map((item) => item.keyForList)).toEqual(['create_expense']);
```
marufsharifi: 🟠 This assertion passes identically with the `.trim()` change on `SearchRouterHelpers.ts:36-38` reverted — `stripNavigationIntentPrefix` (line 21 of that file) already runs `.replaceAll(/\s+/g, ' ')` on the query before matching.

It's a green test that guards nothing. Either drop it along with the `.trim()`, or replace it with a case that genuinely exercises the new code path.

marufsharifi (changes requested):

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

tests/unit/SearchRouterNavigationTest.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: I don’t think we should restore this subscription here. The sibling call sites compute `hasViolations`, but they pass `undefined` as the reportID, so `getReportTransactions(undefined)` returns `[]` and `hasViolationsReportUtils(...)` always returns `false`. Passing `false` preserves the current runtime result while avoiding a full `TRANSACTION_VIOLATIONS` subscription on the Search Router typing path. If the sibling call sites need different behavior, that should be fixed separately with a real report/transaction source rather than copying the no-op subscription here.

tests/unit/useCreateNavigationSuggestionsTest.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: I don’t think we should change this in this PR without product/design confirmation. The current 8-result cap and source ordering were already part of the Search Router behavior, and I don’t see a confirmed requirement that bare `go` / `go to` must include every source or always include a Create row. This PR keeps Create actions reachable by their actual terms (`create expense`, `new chat`, `add expense`, etc.). If the expected priority for broad intent queries should change, I think we should handle that in the ordering/priority follow-up with clear confirmation.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterHelpers.ts
nabi-ebrahimi: Addressed.

tests/unit/SearchRouterNavigationTest.ts
nabi-ebrahimi: Addressed.

marufsharifi (approved):
LGTM

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,224 @@
...
+import {isNavigationIntentOnlyQuery, matchesNavigationQuery, stripNavigationIntentPrefix} from './SearchRouterHelpers';
+
+type CreateNavigationItem = {
+    visible: boolean;
+    text: string;
+    icon: IconAsset;
+    action: () => void;
+    keyForList: string;
+    matchTerms?: string[];
+};
```
mountiny, line 50: can you please add docs

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
```diff
@@ -0,0 +1,224 @@
...
+    const {isRestrictedPolicyCreation} = usePreferredPolicy();
+    const [allPolicies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
+    const [reportID] = useState(() => generateReportID());
+    const [draftTransactionIDs] = useOnyx(ONYXKEYS.COLLECTION.TRANSACTION_DRAFT, {selector: validTransactionDraftIDsSelector});
+    const [lastDistanceExpenseType] = useOnyx(ONYXKEYS.NVP_LAST_DISTANCE_EXPENSE_TYPE);
+    const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
+    const [allBetas] = useOnyx(ONYXKEYS.BETAS);
+    const [activePolicyID] = useOnyx(ONYXKEYS.NVP_ACTIVE_POLICY_ID);
+    const [activePolicy] = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${activePolicyID}`);
+    // Match Quick Creation and Search actions by using shared report eligibility rules. Unlike the FAB's
```
mountiny, line 93:
```suggestion

    // Match Quick Creation and Search actions by using shared report eligibility rules. Unlike the FAB's
```

tests/unit/useCreateNavigationSuggestionsTest.ts
mountiny, line 1: @MelvinBot Can you suggest other scenarios to unit test from the hook? I believe there should be more options we will cover

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useCreateNavigationSuggestions.ts
nabi-ebrahimi: Addressed.

mountiny (approved):
Thanks for following up on that

PR #97087, opened 2026-07-27
Lazy-load contact options: shell first, createOption after getValidOptions

jmgraa:
### Explanation of Change

On large accounts, building a full `createOption` for every personal detail is expensive (policy/tag/archive lookups, icons, etc.), even though most contacts are discarded by filtering/ranking.

This PR changes contact option construction so `createFilteredOptionList` always emits lightweight `PersonalDetailShell`s instead of fully built options. Shells only carry the fields needed to filter, rank, and de-dupe (`text`, `login`, `accountID`, `participantsList`, etc.), plus an `isHydrated: false` discriminant and a memoized `hydrate()` that runs `buildFullOption` / `createOption` on demand. Display fields are not readable on a shell without narrowing on `isHydrated`.

How it works:

1. Build one shell per personal detail in `createFilteredOptionList` / `buildPersonalDetailsOptions`.
2. In `getValidOptions`, filter and rank on those shells, then hydrate only the top-N survivors with `hydrateContactOption`. Contacts that never make the cut never pay the full `createOption` cost.

Callers of `getValidOptions` / `getSearchOptions` still receive fully built personal-detail options; the lazy work is internal to the option-list pipeline. `hydrateContactOption` remains available for any shell that still needs hydration outside that path (e.g. tests or draft restore).

Unit tests check that shells match fully built options for filtering and sorting, and that after hydration the final results match the previous eager path (including search, exclusions, and top-N limits).

#### ParticipantSearchResults / createFilteredOptionList

| Before        | After         |
| 898ms / 812ms | 717ms / 624ms |
| 927ms / 838ms | 669ms / 588ms |
| 946ms / 855ms | 713ms / 618ms |

### Fixed Issues
$ https://github.com/Expensify/App/issues/97788
PROPOSAL:

### Tests
1. Log in with an account that has many contacts / personal details (ideally a high-traffic account).
2. Start a money request (or open any flow that uses the participant / people search selector).
3. Open the participant picker and wait for the initial contact/report list to load.
4. Verify the list shows expected contacts and recent reports with correct display names, logins, and avatars (no missing icons or blank rows).
5. Search for a known contact by name and by email/login.
6. Verify matching contacts appear in the correct order and match pre-change behaviour.
7. Select one or more participants and confirm selection state, chips, and submit still work.
8. Clear the search and scroll / load more if available; verify lists remain correct.

### Offline tests
1. Open the participant picker while online so options load.
2. Go offline (disable network).
3. Search for a contact that is already available locally.
4. Verify filtering/ranking still works and selected hydrated options still show full details (name, login, avatar).

### QA Steps
Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/OptionsListUtils/types.ts
```diff
@@ -97,8 +100,37 @@ type SearchOptionData = Pick<
...
 type SearchOption<T> = SearchOptionData & {
     item: T;
+
+    /**
+     * Present only on lightweight personal detail options built by createFilteredOptionList. Such options carry
+     * just the fields needed for filtering and sorting, and hydrateLazyPersonalDetailOption uses this data to
+     * build the full display option once the visible page of options has been selected.
+     */
+    lazyHydrationData?: PersonalDetailLazyHydrationData;
 };
```
dariusz-biela, line 129: Making `lazyHydrationData` optional means TypeScript cannot distinguish a lightweight shell from a fully built `SearchOption`. Consequently, consumers can access display-only fields on entries in `options.personalDetails`; the code compiles, but those fields may be `undefined` at runtime.                                                                                                                                                                          
                                                                                                                                                                                                                           
Now that lazy construction is always enabled, this affects every current and future consumer. The existing issue was found through manual code inspection rather than by the compiler. Could we model lightweight and hydrated options as separate, discriminated types so consumers must hydrate or narrow them first?

src/libs/OptionsListUtils/index.ts
dariusz-biela, line 1: `src/components/Search/FilterComponents/InSelector.tsx:54`
```ts
    const {options, isLoading} = useFilteredOptions({
        enabled: ready,
        isSearching: !!debouncedSearchTerm.trim(),
    });
```

`useFilteredOptions` defaults to `includeP2P: true`, but this component only renders `chatOptions.recentReports`.

Because the `getSearchOptions` call below passes neither `maxResults` nor `searchQuery`, `getValidOptions` hydrates every eligible personal-detail shell before the actual search term is applied by `filterAndOrderOptions`. Those hydrated `personalDetails` are then never rendered.

Compared with `main`, this adds shell creation on top of the existing full option construction for nearly every contact, causing a performance regression. Could we pass `includeP2P: false` here?

config/eslint/eslint.seatbelt.tsv
```diff
@@ -731,7 +731,7 @@
 "../../src/libs/OptionsListUtils/index.ts"	"@typescript-eslint/no-deprecated/deprecatedCachedOneTransactionThreadReportIDs[reportID]"	2
 "../../src/libs/OptionsListUtils/index.ts"	"@typescript-eslint/no-deprecated/deprecatedGetReportName"	3
 "../../src/libs/OptionsListUtils/index.ts"	"@typescript-eslint/no-deprecated/deprecatedLastReportActions[reportID]"	3
-"../../src/libs/OptionsListUtils/index.ts"	"@typescript-eslint/no-deprecated/translateLocal"	9
+"../../src/libs/OptionsListUtils/index.ts"	"@typescript-eslint/no-deprecated/translateLocal"	10
```
dariusz-biela: Is this necessary?

src/libs/OptionsListUtils/index.ts
```diff
@@ -1090,6 +1090,29 @@ type CreateOptionParams = {
...
+ * ranking of shells match the hydrated options. Mirrors createOption's historical asymmetry: the
+ * personal details data is only passed through when there is no report.
+ */
+function getPersonalDetailOptionText(
+    accountID: number | undefined,
+    hasReport: boolean,
+    personalDetails: OnyxEntry<PersonalDetailsList>,
+    login: string | undefined,
+    translate: LocalizedTranslate,
+): string {
```
dariusz-biela:
```suggestion
type GetPersonalDetailOptionTextProps = {
      accountID: number | undefined;
      hasReport: boolean;
      personalDetails: OnyxEntry<PersonalDetailsList>;
      login: string | undefined;
      translate: LocalizedTranslate;
  };

function getPersonalDetailOptionText({
      accountID,
      hasReport,
      personalDetails,
      login,
      translate,
  }: GetPersonalDetailOptionTextProps): string {

```

src/libs/OptionsListUtils/index.ts
dariusz-biela: I think we should change the default to `includeP2P: false` and add a comment noting that, in some situations, setting it to true may degrade performance.

dariusz-biela: Oh, and we have a potential performance regression in `src/pages/NewChatPage/index.tsx:117`

config/eslint/eslint.seatbelt.tsv
jmgraa: Yes, necessary. The shell has to resolve a display name and createFilteredOptionList has no translate in scope, so it calls translateLocal - one new reference.

Getting back to 9 means passing translate into createFilteredOptionList.

src/libs/OptionsListUtils/index.ts
jmgraa: **flip the default to includeP2P: false** - I'd rather not flip it. `NewChatPage`, `ShareTab`, `SearchAutocompleteList` and `useSearchSelector` all render contacts and none of them passes the flag, so a false default would quietly empty their contact sections. Opting out per screen is safer, so I documented the cost on the hook config and on `createFilteredOptionList` and set `includeP2P: false` on the one screen that renders reports only.

If you still prefer false as the default, I'll do it along with explicit `includeP2P: true` at those four call sites.

jmgraa: **potential performance regression in src/pages/NewChatPage/index.tsx:117** - Confirmed, it was an actual regression. That screen calls `getValidOptions` on every render with no `maxElements`, and search skips contact pagination, so every keystroke rebuilt every surviving contact. The eager version built once per option list and was reused from the cache, so the lazy path was slower here than main.

Fixed by caching the built option against its shell (WeakMap). Safe because a shell's inputs are frozen at build time, so the same shell always produces the same option; new option list means new shells, and entries go away with them.

Each call still returns a fresh shallow copy rather than the cached object - consumers mark options in place (`isSelected, isBold`), so sharing one instance could leave a memoized row showing a stale checkmark. Only the expensive createOption work is shared. New test covers both: work reused, object fresh, mutations don't leak.

The alternative was deferring hydration past filterAndOrderOptions, but then the in-place marking happens before the build and has to be re-applied afterwards.

src/libs/OptionsListUtils/types.ts
jmgraa: Agreed and it was slightly worse than undefined fields, since some display fields held placeholder values (`alternateText` was the raw login), so you'd read wrong data rather than nothing.

Shells now have their own type: `OptionList.personalDetails` is `LazyPersonalDetailOption[]`, exposing only the fields a shell really has, and `hydrateLazyPersonalDetailOption` widens it back to a full `SearchOption`.

Went with a narrower field list instead of a discriminated union so consumers that only need login/text don't have to narrow. Device contacts stay assignable, so the `useSearchSelector` concat path is unchanged.

One limit: all `SearchOptionData` fields are optional, so a shell still fits a `SearchOptionData` parameter. This blocks reading display data off the option list, not passing a shell somewhere that wants a full option.

src/components/Search/FilterComponents/InSelector.tsx
```diff
@@ -54,6 +54,9 @@ function InSelector({value = [], selectionListTextInputStyle, selectionListStyle
     const {options, isLoading} = useFilteredOptions({
         enabled: ready,
         isSearching: !!debouncedSearchTerm.trim(),
+        // This filter only ever renders reports (see the sections built below), and the getSearchOptions call
+        // here passes neither maxResults nor searchQuery, so every contact would be built and then dropped.
+        includeP2P: false,
```
dariusz-biela, line 59:
```suggestion
          // The sections below read recentReports and never personalDetails, so contacts would never reach the list.
        includeP2P: false,
```

src/libs/OptionsListUtils/index.ts
jmgraa: Correction to my earlier count after re-checking call sites: three screens rely on the default, not four - `ShareTab`, `SearchAutocompleteList`, `useSearchSelector` pass no flag. `NewChatPage` already passes `includeP2P: true` explicitly, so flipping the default wouldn't affect it. `SearchAutocompleteList` builds contacts only while searching (`deferContactsUntilSearch: true`) - exactly the state a `false` default would break.

Conclusion unchanged: unwanted `true` degrades performance, unwanted `false` silently drops rendered contacts, so per-screen opt-out is safer. Cost is documented on the hook config and `createFilteredOptionList`, and `InSelector` (the one report-only consumer) now passes `includeP2P: false`. The NewChatPage regression above is fixed via per-shell WeakMap memoization. Resolving - reopen if you still prefer flipping the default and I'll do it with explicit `true` at the three call sites.

tests/unit/OptionsListUtilsTest.tsx
dariusz-biela, line 1: The current parity tests use the lazy hydration path to build both the actual result and the expected baseline. Because both sides depend on the same captured context, a missing hydration input can produce identical but incorrect results and the tests will still pass.

  Could we add an independent test that compares a hydrated lazy option with an option built directly from the original inputs? It would be useful to include at least one meaningful non-default value and assert both full parity and a specific output affected by that value. This would ensure regressions in the hydration input plumbing are detected.

src/libs/OptionsListUtils/types.ts
```diff
@@ -97,13 +100,79 @@ type SearchOptionData = Pick<
...
+
 type SearchOption<T> = SearchOptionData & {
     item: T;
+
+    /**
+     * Present only on lightweight personal detail options built by createFilteredOptionList. Such options carry
+     * just the fields needed for filtering and sorting, and hydrateLazyPersonalDetailOption uses this data to
+     * build the full display option once the visible page of options has been selected.
+     */
+    lazyHydrationData?: PersonalDetailLazyHydrationData;
```
dariusz-biela: Could we model the lazy personal detail shell as a discriminated union instead of exposing optional `lazyHydrationData` on the generic `SearchOption<T>`?

Currently, `lazyHydrationData` is optional on both `SearchOption<T>` and `LazyPersonalDetailOption`, so the types do not reliably distinguish a shell from a fully built option. It also means `SearchOption<Report>` exposes a field that report options never contain.

I suggest introducing:

```ts
type LazyPersonalDetailOptionShell = Pick<...> & {
    item: PersonalDetails | null;
    lazyHydrationData: PersonalDetailLazyHydrationData;
};

type LazyPersonalDetailOption =
    | LazyPersonalDetailOptionShell
    | SearchOption<PersonalDetails | null>;
```

and removing `lazyHydrationData` from `SearchOption<T>`.

The hydrator could then narrow safely using:

```ts
if (!('lazyHydrationData' in option)) {
    return option;
}
```

The hydration `WeakMap` should also be keyed by `LazyPersonalDetailOptionShell`.

This would make the hydration payload mandatory for actual shells, preserve support for fully built device contacts, and remove the invalid lazy-hydration contract from report options. The presence of `lazyHydrationData` already serves as the discriminator, so a separate `kind` field should not be necessary.

src/hooks/useFilteredOptions.ts
dariusz-biela: Since includeP2P can significantly increase the amount of work, could we remove the default and make this property required?                                                                                                                        
                                                                                                                                                                                                                                                
Defaulting to true risks silent performance regressions, while defaulting to false risks silently removing contacts. Requiring an explicit value forces every consumer to declare its intent:                                                       
```ts                                                                                                                                                                                                                                                   
includeP2P: boolean;                                                                                                                                                                                                                                
 ```                                                                                                                                                                                                                                                 
Contact pickers should pass true, while report-only screens should pass false.

Comment (dariusz-biela) [inline src/hooks/useFilteredOptions.ts:1]: useSearchSelector still calls useFilteredOptions without forwarding getValidOptionsConfig.includeP2P.                                                                                                                                               
                                                                                                                                                                                                                                                
Consequently, when includeP2P is false, we still build a shell for every contact and only discard them later in getValidOptions. Could we propagate this decision to useFilteredOptions so contacts are skipped at the source?                      
                                                                                                                                                                                                                                               
```ts
const shouldBuildContactOptions =                                                                                                                                                                                                                   
   searchContext === CONST.SEARCH_SELECTOR.SEARCH_CONTEXT_GENERAL ||                                                                                                                                                                               
   searchContext === CONST.SEARCH_SELECTOR.SEARCH_CONTEXT_ATTENDEES                                                                                                                                                                                
        ? (getValidOptionsConfig.includeP2P ?? true)                                                                                                                                                                                                
       : true;      
```

src/libs/OptionsListUtils/index.ts
```diff
@@ -1697,6 +1720,95 @@ function clearFilteredOptionListCache() {
...
+// no maxElements) would otherwise rebuild every surviving contact each time, which is slower than the eager build
+// they replaced: that one ran once per option list and was reused from the createFilteredOptionList cache.
+// A new option list produces new shells, so entries are never reused across builds and drop out with the shells.
+//
+// The cached option is never handed out directly: consumers mark options in place (getValidOptions sets
+// isBold/isSelected/brickRoadIndicator) and list rows re-render on reference change, so every call returns a fresh
+// shallow copy, exactly like the per-call objects the eager build produced. Only the createOption work is shared.
+// The copies share nested objects (icons, participantsList, item) with the cached entry, matching the
+// cloneOptionList invariant: a consumer that mutates those must clone them first.
+const hydratedPersonalDetailOptions = new WeakMap<LazyPersonalDetailOption, SearchOption<PersonalDetails | null>>();
```
dariusz-biela: `createFilteredOptionList` never hands out its own arrays, it always returns `cloneOptionList(...)` - both after a fresh build and on a cache hit. The clone allocates new shell objects (`personalDetails.map((option) => ({...option}))`), so a `WeakMap` keyed on the shell starts empty for every clone.

Inside one mounted screen this is fine, because `useFilteredOptions` memoizes the clone and the same shell objects are reused across renders. It stops working whenever a *fresh* `createFilteredOptionList` call hits the entry cache: reopening the picker, or a second screen running the same config. There the clone is new, every key misses, and the whole visible page is rebuilt with `createOption`, while `main` returns already-built options from that same cache entry and does zero `createOption` work. So on that warm path this ends up doing work `main` does not - cold open is of course still much faster.

Suggestion: move the memo onto the shell itself as a thunk instead of keying a module-level map by object identity:

```ts
// on the shell
hydrate: () => SearchOption<PersonalDetails | null>;
```

`cloneOptionList` copies function properties by reference, so all clones of one cached option list share the memoized result, and a new build produces new shells whose results are collected with them. It would also remove the `lazyHydrationData` field, its `deepFreeze` exemption and the `SearchOption<T>` widening raised in the other comments, so it should be a net deletion.

src/libs/OptionsListUtils/index.ts
```diff
@@ -2894,7 +3002,9 @@ function getValidOptions(
             ? Math.max(maxElements - recentReportOptions.length - workspaceChats.length - (!selfDMChat ? 1 : 0), MIN_PERSONAL_DETAILS_SLOTS)
             : undefined;
         const groupedPersonalDetails = optionsOrderBy(options.personalDetails, personalDetailsComparator, maxPersonalDetailsElements, filteringFunction, true);
-        personalDetailsOptions = groupedPersonalDetails.options;
+        // Lightweight options from createFilteredOptionList get their full display fields only now, after the
+        // heap selection reduced them to a single page, so the expensive work is done for a handful of options.
+        personalDetailsOptions = groupedPersonalDetails.options.map(hydrateLazyPersonalDetailOption);
```
dariusz-biela: Hydration here is bounded by `maxPersonalDetailsElements`, which is `undefined` when the caller passes no `maxElements`. `NewChatPage` passes neither `maxElements` nor `searchString` - it does its filtering afterwards in `filterAndOrderOptions` - so this line hydrates every contact on the account. On top of that its `usePaginatedData` runs with `skipPagination: isSearching`, so the full address book reaches `getValidOptions` the moment the user starts typing.

On a large account that is one synchronous pass of `createOption` over every contact when search opens: the same total work `main` does, moved from `createFilteredOptionList` into `getValidOptions`. The optimization does not reach the largest contact picker in the app.

Two ways out, both small:

- pass `maxElements` from `NewChatPage` so the heap caps contacts before hydration (would need a check against its own `hasMore` / pagination), or
- let that one caller hydrate after its own pagination, behind an explicit opt-out such as a `deferContactHydration?: boolean` in the `getValidOptions` config, so the lazy shells stay invisible to every other consumer.

src/libs/OptionsListUtils/index.ts
```diff
@@ -1697,6 +1720,95 @@ function clearFilteredOptionListCache() {
...
+            policy,
+            privateIsArchived,
+            conciergeReportID,
+            config: {showPersonalDetails: true},
+            reportAttributesDerived,
+            policyTags: reportPolicyTags,
+            visibleReportActionsData,
+        }),
+    };
+    hydratedPersonalDetailOptions.set(option, hydrated);
```
dariusz-biela: `deepFreeze(result)` runs on the built option list, so before this PR a contact option's nested objects (`icons`, `subtitle`, ...) were frozen in dev and a consumer writing through them threw immediately. The hydrated option is built after that point and never frozen, yet it is shared by every caller that hydrates the same shell - `{...hydrated}` only copies the top level. The check that used to catch nested mutation of a contact option no longer exists.

```ts
if (__DEV__) {
    deepFreeze(hydrated);
}
```

right before the `set` restores it. `item` and the members of `participantsList` are already exempt inside `deepFreeze`, and the `{...hydrated}` return keeps the top level mutable, so the `isBold` / `isSelected` marking still works.

src/libs/OptionsListUtils/types.ts
dariusz-biela: This puts an optional `lazyHydrationData` on every `SearchOption<T>`, including `SearchOption<Report>`, where it can never be set and means nothing. Only contact shells carry it, and `LazyPersonalDetailOption` already declares it a few lines below.

If the memo moves into a thunk on the shell (see the comment on `hydratedPersonalDetailOptions`) this field disappears entirely. If it stays, it should live only on `LazyPersonalDetailOption`.

src/libs/OptionsListUtils/index.ts
```diff
@@ -1665,17 +1687,18 @@ function cloneOptionList(optionList: OptionList): OptionList {
...
+// every option build), so freezing them would crash unrelated flows in dev. Mutating them corrupts
+// app-wide Onyx state, which is beyond this cache's invariant.
 function deepFreeze(value: unknown) {
     if (typeof value !== 'object' || value === null || Object.isFrozen(value)) {
         return;
     }
     Object.freeze(value);
     for (const [key, child] of Object.entries(value)) {
-        if (key === 'item') {
+        if (key === 'item' || key === 'lazyHydrationData') {
```
dariusz-biela: The exemption is needed because the hydration inputs hang off the shell. The side effect is that anyone holding a contact option can reach `option.lazyHydrationData.context` and mutate `personalDetails`, `policiesCollection`, `policyTags` or `reportAttributesDerived` - one object shared by every shell of the run, and by the app.

A thunk would keep those inputs in a closure, unreachable from the option, and `deepFreeze` would need no exemption here at all because it only recurses into objects.

src/libs/OptionsListUtils/index.ts
```diff
@@ -1697,6 +1720,95 @@ function clearFilteredOptionListCache() {
...
+            // eslint-disable-next-line rulesdir/no-default-id-values
+            reportID: report?.reportID ?? '',
+            keyForList: report ? String(report.reportID) : String(accountID),
+            text,
+            login: detail?.login,
+            accountID: Number(detail?.accountID),
+            participantsList: detail ? [detail] : [],
+            isSelected: false,
+            selected: false,
+            brickRoadIndicator: null,
```
dariusz-biela: `getValidOptions` now hydrates before the marking loop, and that loop writes `brickRoadIndicator` on the hydrated option, so this `null` is never read. The `Pick` in `types.ts` groups `'brickRoadIndicator'` under "Written in place by getValidOptions", which is no longer true for this one field - both this line and the Pick entry can go.

`displayName` and `isOptimisticPersonalDetail` should stay: they are undefined on both paths today, but `getPersonalDetailSearchTerms` reads them, so they are genuinely filter-relevant.

src/libs/OptionsListUtils/index.ts
```diff
@@ -1697,6 +1720,95 @@ function clearFilteredOptionListCache() {
...
+ */
+function buildPersonalDetailsOptions(reportMapForAccountIDs: Record<number, Report>, context: LazyHydrationContext): LazyPersonalDetailOption[] {
+    const {personalDetails} = context;
+    return Object.values(personalDetails ?? {}).map((personalDetail) => {
+        const accountID = personalDetail?.accountID ?? CONST.DEFAULT_NUMBER_ID;
+        const report = reportMapForAccountIDs[accountID];
+        // Same lookup createOption performs (also normalizes accountID in place).
+        const detail = getPersonalDetailsForAccountIDs([accountID], personalDetails)[accountID];
+        // Same text createOption computes (translateLocal — the default createOption uses when translate is omitted).
+        const text = getPersonalDetailOptionText({accountID, hasReport: !!report, personalDetails, login: detail?.login, translate: translateLocal});
```
dariusz-biela: The shell computes `getPersonalDetailOptionText` for every contact and `createOption` computes the same value again for the survivors. It is only a page's worth of duplicated work, and passing the precomputed text into hydration would cost you `getPersonalDetailOptionText` as the single source of truth for both paths, which is the most valuable part of this refactor. Fine to leave as is - noting it so it reads as a decision rather than an oversight.

tests/unit/OptionsListUtilsTest.tsx
jmgraa: Added `should produce exactly what createOption produces from the same inputs`

src/libs/OptionsListUtils/types.ts
jmgraa: Addressed - same idea as a discriminated union, different shape.

Instead of required lazyHydrationData on the shell, we use isHydrated (false + hydrate() for shells, true for fully built options like device contacts). That still narrows safely and keeps hydration off SearchOption. The hydrate memo itself is covered in the later WeakMap comment⬇️.

staszekscp (approved):
This looks really good! 🎉

mjasikowski (approved):

mjasikowski:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

Issue #72725, opened 2025-10-15
[Due for payment 2026-08-28] [Due 2026-08-26] Migrate ProposalPolice to use responses+conversations rather than assistants


roryabraham:
### Problem
The assistants API we use for ProposalPolice is deprecated. It's also opaque because the prompt is configured manually in the OpenAI dashboard rather than checked into code and version controlled. This is problematic because I am maybe the only one who can change the prompt, and also because assistants are deprecated we can't use GPT-5, which might produce better results for cheaper than GPT-4o.

### Solution
Migrate ProposalPolice to use the "normal" Chat Completions API rather than an assistant. Start using GPT-5.

roryabraham:
Created https://github.com/Expensify/App/pull/75624 to implement this.

roryabraham:
No update. Assistants are still working for now

roryabraham:
This has a real final deadline of August 26, 2026

roryabraham:
No update

PR #97199, opened 2026-07-28
[No QA] Migrate ProposalPolice to Responses + Conversations API

roryabraham:
### Explanation of Change
ProposalPolice currently runs on the OpenAI Assistants API: the system prompt is configured manually in the OpenAI dashboard (not checked into code), and duplicate-proposal detection makes one Assistants thread/run per prior proposal on the issue (each with up to 90s of polling).

Assistants are being deprecated in favor of the Responses API. **They will be shut off on 2026-08-26**, so we're up against a real deadline soon here.

This migrates ProposalPolice to:
- The **Responses API** (`OpenAIUtils.promptResponses`) on **gpt-5.6-luna** (replacing the Assistant's GPT-4o) for the comment-intent and edit-check calls — no more thread/run polling.
- A persistent, per-issue **OpenAI Conversation** for duplicate-check, so the loop of one Assistants call per prior proposal becomes a single Responses call against the issue's Conversation history. The Conversation ID is tracked via a hidden marker in a bot comment on the issue.
- A checked-in, composed system prompt (`prompts/proposalPolice/*`) split into small, focused instruction fragments per call type (comment-intent / edit-check / duplicate-check), replacing the single dashboard-configured prompt blob — each call only gets the instructions it actually needs.
- JSON-schema structured outputs for each call's response shape, replacing manual "respond with JSON" prompt instructions.

#### Behaviour changes beyond the migration

The system prompt configured in the OpenAI dashboard is **stale** against `contributingGuides/PROPOSAL_TEMPLATE.md`: it still requires two sections that were removed in `ed87b140dc6` and `0ced4a8d38c`. On `main` that means the bot nags contributors whose proposals correctly follow the current template. The checked-in `templateDefinition.ts` matches the real template, which fixes that.

**The created-comment path is reworked.** The case-sensitive `Proposal` keyword gate is gone, because it meant an entire class of comment never reached any check at all — a bid for the job that never uses the word, [like this one](https://github.com/Expensify/App/issues/98504#issuecomment-5289204908), returned at that gate before anything ran. Instead:

1. Expensify employees are filtered out in the workflow. `author_association` isn't sufficient (it only reports `MEMBER` for publicly visible org members), so this reads `Expensify/expensify-expensify` team membership, which needs a `read:org` token. A check that errors skips the run rather than treating the author as an outside contributor.
2. Template conformance is checked **deterministically in code** (`isProposal`). It's a definition, not a judgment, and an LLM checking it can drift from the template — which is exactly the `main` bug above.
3. Only comments that fail it get an LLM call, classifying intent as `SPAM` (a claim on the job with no technical content of its own), `GENUINE_ATTEMPT` (real technical content, wrong format), or `NOT_AN_ATTEMPT` (feedback, retests, questions, discussion).
4. `SPAM` is collapsed via GraphQL `minimizeComment`, which leaves the text intact and is reversible — deliberately not the duplicate path's body overwrite. `GENUINE_ATTEMPT` stays visible. Both are pointed at the template.

The model reports intent and the code decides the action, so the enforcement policy stays in one reviewable, testable place.

**Call volume.** Measured over the 7 days to 2026-08-15 across the 61 open `Help Wanted` issues: 303 non-bot comments, of which 98 already follow the template and so cost nothing. That leaves **~29 intent calls/day (~880/month)**, an upper bound since Expensify employees are filtered out before the action runs and aren't excluded from that count. Duplicate-check volume is unchanged at ~14/day. The instruction prefix is cached via `promptCacheKey`, so the marginal cost of each call is the comment body plus a one-word response.

**Two edit-check gates are new rather than ported**, both fixing live `main` bugs: edits that didn't change the body are skipped (on `main` these reached the model as `Previous comment content: undefined`, which reads as a full rewrite and could produce a spurious banner), and edits are skipped unless either side of the edit is a proposal (so a heavily reworded discussion comment that merely mentions "Proposal" isn't bannered).

**Duplicate detection never matches a proposal against its own author's.** The old prompt asked for this but neither implementation honoured it, and the Conversation makes it more likely to bite: `main` compared pairwise and stopped at the first hit, whereas the model now sees every prior proposal at once and reports a single highest score, making a contributor's own earlier draft a real candidate.

No tests previously existed for ProposalPolice or `OpenAIUtils`; this PR adds coverage for all of the new logic (`OpenAIUtilsTest`, `ProposalPoliceConversationTest`, `ProposalPoliceInputTest`, `ProposalPoliceSchemaTest`, `ProposalUtilsTest`, `proposalPoliceCommentTest`).

### Fixed Issues
$ https://github.com/Expensify/App/issues/72725

### Tests
Outside of the automated tests we added, this will be live-tested.

### Offline tests
None — this is a GitHub Actions bot with no client-side/offline behavior.

### QA Steps
None — see `[No QA]` in the PR title. This is a backend-only GitHub Actions change with no user-facing app behavior.

### PR Author Checklist
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
n/a


prompts/proposalPolice/botActions.ts
```diff
@@ -0,0 +1,25 @@
...
+`);
+
+/**
+ * How to classify an edit to a proposal comment. Used only by the edit-check call.
+ */
+const EDITED_COMMENT_ACTIONS = Str.dedent(`
+    EDITED COMMENTS: Compare the original proposal with its latest edit.
+
+    - "${CONST.ACTION_EDIT}" if the changes are SUBSTANTIAL.
+    - "${CONST.NO_ACTION}" if the changes are MINOR.
```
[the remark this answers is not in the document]
roryabraham: fixed in 95cdde58e6b194e15fd94c798fc603eec018f68a

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -114,126 +114,147 @@ async function run() {
...
+
+        // edit comment if substantial changes were detected
+    } else if (action === CONST.ACTION_EDIT) {
         console.log('ProposalPolice™ editing issue comment...', commentID);
         await GithubUtils.octokit.issues.updateComment({
             ...context.repo,
             /* eslint-disable @typescript-eslint/naming-convention */
             comment_id: commentID,
-            body: `${formattedResponse}\n\n${payload.comment?.body}`,
+            body: `${buildSubstantiveEditMessage(formattedDate)}\n\n${payload.comment?.body}`,
```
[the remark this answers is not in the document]
roryabraham: fixed in 577b1395674e5221e5ac315b1b5c5c5370d0d40b

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -114,126 +130,167 @@ async function run() {
...
-                comment_id: commentID,
-                body: duplicateCheckWithdrawMessage,
-            });
-            // Post a comment to notify the user about the withdrawn duplicated proposal
-            console.log('ProposalPolice™ notifying contributor of withdrawn proposal...');
-            await GithubUtils.createComment(CONST.APP_REPO, issueNumber, duplicateCheckNoticeMessage);
-            console.log('DUPLICATE PROPOSAL DETECTION Check Completed, returning early.');
-            return;
+            const similarityPercentage = parsedDuplicateCheckResponse?.similarity ?? 0;
+            if (similarityPercentage >= DUPLICATE_SIMILARITY_THRESHOLD) {
```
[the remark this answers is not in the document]
roryabraham: Fixed in 2ac29f16809e0dd3dcb4cb50595b815057924a7e

.github/workflows/proposalPolice.yml
```diff
@@ -3,6 +3,12 @@ on:
   issue_comment:
     types: [created, edited]
 
+concurrency:
+  # Serializes runs per-issue so two comments posted close together can't each create their own
+  # OpenAI Conversation for duplicate-check, splitting an issue's duplicate-detection history in two.
+  group: proposal-police-${{ github.event.issue.number }}
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
roryabraham: TIL about the [new `queue: max` feature](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency). Been wanting that for years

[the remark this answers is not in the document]
roryabraham: Fixed in a2fa73c29ee504eb15d032e3a0356790341b1de3

.github/libs/isBotUser.ts
```diff
@@ -0,0 +1,16 @@
+// Note: if you are updating this set, please update it in PHP as well: https://github.com/Expensify/Web-Expensify/blob/d5d74f1ba73deed0379ff5b9f57376213a8b02bf/lib/Github/Utils.php#L51
+const KNOWN_BOT_USERS = new Set(['CLABotify', 'MelvinBot', 'OSBotify', 'botify', 'exfy-zapier']);
```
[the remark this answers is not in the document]
roryabraham: not a real problem. `chatgpt-codex-connector` is a bot

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -86,11 +87,27 @@ async function run() {
     }
 
     // If event is `edited` and comment was already edited by the bot, return early
-    if (isCommentEditedEvent(payload) && payload.comment?.body.trim().includes('Edited by **proposal-police**')) {
+    if (isCommentEditedEvent(payload) && payload.comment?.body.trim().includes(SUBSTANTIVE_EDIT_MESSAGE_PREFIX)) {
```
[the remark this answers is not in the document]
roryabraham: Fixed in d4232b5d405a650a5d38edb41df646afe00b603f

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -33,30 +62,29 @@ function isCommentEditedEvent(payload: IssueCommentEvent): payload is IssueComme
...
-/**
- * Determines if a comment author is a known bot or a bot-type account.
- */
-function getIsBotAuthor(user: {login?: string; type?: string} | null | undefined): boolean {
-    if (!user) {
-        return false;
+    const staleItemID = findConversationItemIDForComment(await openAI.listConversationItems(trackedConversationID), commentID);
+    if (!staleItemID) {
+        console.log('Could not find this proposal in the Conversation, leaving it untouched.', commentID);
+        return;
```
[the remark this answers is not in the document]
roryabraham: Good catch — fixed in 1714d89354d.

You're right that this was unreachable for a comment that only became a proposal via an edit. The created path returns at `isProposal(newProposalBody)` without recording anything, so there was never a stale item to match, and `refreshStoredProposal` bailed. Since duplicate checks read the Conversation rather than live comments, that proposal was invisible to them permanently.

Rather than only adding when no stale item exists, I made the function's rule complete: drop whatever is stored for the comment, then re-add **only while the comment is still a proposal**. That covers all four combinations —

| stored copy | proposal now | result |
| --- | --- | --- |
| yes | yes | replaced with the current text |
| no | yes | added (the case you found) |
| yes | no | removed, nothing put back |
| no | no | nothing |

The third row turned out to be a second bug in its own right: a proposal edited into a retraction kept its stored copy and went on matching future proposals from the Conversation.

This reverses the replace-only rule I added in 577b1395674. That rule was guarding against two copies of one `comment_id` if the lookup missed an item that was really there — but that requires a matching bug, whereas a proposal permanently invisible to duplicate detection is the ordinary path. Wrong trade, so I took yours.

Covered by two new tests, `records a comment that only became a proposal on this edit` and `drops the stored copy when a proposal is edited into something that is no longer one`. I verified both fail against the previous logic.

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -85,9 +117,19 @@ async function run() {
...
+    }
+
+    // The edit check only grades how much a proposal changed, so a heavily reworded discussion comment
+    // that merely mentions the word "Proposal" would otherwise come back as a substantial edit and get
+    // the bot's banner prepended. Either side being a proposal is enough: an edit can add the template
+    // to a comment that lacked it, or strip it from one that had it.
+    if (isCommentEditedEvent(payload) && !isProposal(payload.changes.body?.from) && !isProposal(payload.comment?.body)) {
+        console.log('Edited comment is not a proposal, skipping the edit check.');
         return;
```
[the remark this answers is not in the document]
roryabraham: I don't think this is realistically worth fixing

prompts/proposalPolice/duplicateDetection.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Duplicate-proposal detection rules and worked examples. Used only by the duplicate-check call.
+ */
+export default Str.dedent(`
+    DUPLICATE PROPOSAL DETECTION:
+
+    Compare the new proposal against every prior proposal already in this conversation (each was posted as its own message tagged with a comment_id attribute). Ignore every section except ROOT CAUSE and SOLUTION.
+
+    SCORING: Weight the SOLUTION section at least 80% and the ROOT CAUSE section at most 20%. Two proposals are similar only to the extent they propose the same technical change:
+    - Same mechanism or approach, even if worded differently → very high similarity.
```
ikevin127: ### 🟠 `#2` The duplicate prompt lost the "same solution, different root cause" anchor, and the remaining text now contradicts itself for that case

Old instructions prompt, section IX:

> - If both the root cause and solution are nearly identical, the similarity should be very high (close to 100).
> - **If the solution is the same but the root cause is different, the similarity should still be high (over 90).**
> - If the solution is different, even if the root cause is the same, the similarity should be much lower (well below 90).

New `duplicateDetection.ts` keeps the 80/20 weighting and both worked examples, but drops all three anchors. That leaves two clauses pulling in opposite directions for the most common duplicate shape on Expensify issues, two people proposing the same fix while describing the root cause differently:

- **The weighting says ~80.** "Weight the SOLUTION section at least 80% and the ROOT CAUSE section at most 20%" read literally on identical solution + unrelated root cause gives `0.8 x 100 + 0.2 x 0 = 80`.
- **Example 2 says >= 90.** "A and B both say clear the error in onSelectRow (even if worded differently) → same mechanism, similarity >= 90."

**Why this matters:** 80 is below `DUPLICATE_SIMILARITY_THRESHOLD` (85) and 90 is above it. Whichever clause the model weights more determines whether a real duplicate is withdrawn or waved through, and there is nothing in the prompt to break the tie.

On `main` the anchor made this deterministic and the answer was always "withdraw". This is the one delta I would call an unintentional behaviour change rather than a deliberate simplification.

**Suggested fix:** restore the anchors, which costs three lines and resolves the contradiction:

```suggestion
    - Same mechanism or approach, even if worded differently → very high similarity, even when the root causes given are different.
    - Both the root cause and the solution nearly identical → similarity close to 100.
```

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -1,28 +1,57 @@
...
 
-type DuplicateProposalResponse = AssistantResponse & {
-    similarity?: number;
-};
+/**
+ * How similar (0-100) a new proposal must be to an existing one before it's withdrawn as a duplicate.
+ * The model only reports a similarity score; this threshold is applied here so tuning it doesn't
+ * depend on the model reproducing a cutoff stated in the prompt.
+ */
+const DUPLICATE_SIMILARITY_THRESHOLD = 85;
```
ikevin127: ### 🟠 `#3` The threshold drop from 90 to 85 now contradicts both the old code and the old prompt

I flagged this before on the code alone. With the dashboard prompt in hand it is a deviation from two places, not one: the old prompt told the model "Only if the similarity is 90 or above, respond with ACTION_HIDE_DUPLICATE", and the old code independently gated on `similarityPercentage >= 90`. `main` therefore required 90 twice. The PR moves to a single gate at 85.

Stacked on top of [#2](https://github.com/Expensify/App/pull/97199/changes#r3771781896), the calibration story becomes hard to reason about: the model no longer knows any cutoff (deliberate, and I agree with the reasoning in the code comment), the anchors that made scores comparable are gone, and the cutoff itself moved down 5 points. Three simultaneous changes to a decision whose false-positive action overwrites a contributor's comment body.

For a PR whose stated goal is parity, **keep 90** and tune it later with live data:

```suggestion
const DUPLICATE_SIMILARITY_THRESHOLD = 90;
```

If 85 is deliberate, the PR description should say why, because nothing in the diff explains it.

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
```diff
@@ -114,126 +155,176 @@ async function run() {
...
+
+            const similarityPercentage = parsedDuplicateCheckResponse?.similarity ?? 0;
+
+            // The reported match must still be a live proposal on this issue. The model can name a comment
+            // that has since been withdrawn or edited into a different solution, and it can invent an ID
+            // outright. Withdrawing on one of those would destroy a contributor's legitimate work, so an
+            // unresolvable match is treated as "not a duplicate" rather than trusted.
+            const originalProposal = commentsResponse.find(
+                (comment) => comment.id === parsedDuplicateCheckResponse?.duplicateCommentID && comment.id !== commentID && isProposal(comment.body),
+            );
```
ikevin127, line 267: ### 🟡 `#4` "Posted by different users" was in the old prompt and is in neither the old nor the new implementation

Old prompt, section IX opening:

> When a new proposal is posted, compare it to existing proposals in the same issue **that were posted by different users**.

Neither implementation can honour this. On `main` the model received only two proposal bodies with no author information, and the loop filtered on bots and timestamps only. In this PR `buildSeedItems` filters on `isProposal`, bot author, and creation time, but not on same-author, and the seed items carry no author.

So this is parity, not a regression. But it is a documented intent that has never worked, and this PR is the first version where the fix is nearly free, since `commentsResponse` with full user objects is already in hand at both the seed site and the confirmation site:

```ts
// A contributor revising their own thinking in a follow-up comment is not spam, so never withdraw
// a proposal as a duplicate of one by the same author.
const originalProposal = commentsResponse.find(
    (comment) =>
        comment.id === parsedDuplicateCheckResponse?.duplicateCommentID &&
        comment.id !== commentID &&
        comment.user?.login !== newProposalAuthor &&
        isProposal(comment.body),
);
```

Worth doing here because the new architecture makes self-matches *more* likely to be found: `main` compared pairwise and broke on the first hit, whereas the Conversation holds every prior proposal at once and the model is asked for the single highest score, so a contributor's own earlier draft is now a first-class candidate.

prompts/proposalPolice/duplicateDetection.ts
roryabraham: Restored the anchors, and reworded the weighting rather than leaving it. "At least 80% / at most 20%" is arithmetic the model can perform, so keeping it next to an "over 90" anchor reproduces the same contradiction `main` had. The new block states the priority qualitatively and lets the anchors set the numbers.

"Same mechanism → over 90 even when root causes differ" is an eval case now.

`59e4b1d570c`

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
roryabraham: Reverted to 90 in `21cb765946f`

.github/actions/javascript/proposalPoliceComment/proposalPoliceComment.ts
roryabraham: Fix in `cbe6a110f88` as follows:

1. Tag each proposal passed to the duplicate-check LLM call with an author attribute
2. The LLM is instructed to never report a new proposal as a duplicate of a previous proposal from the same author.
3. In case the LLM ignores that instruction, we keep a `comment.user?.login !== newProposalAuthor` check to ensure deterministically that we never hide someone's proposal as a duplicate of their own prior proposal.

ikevin127:
## PR #97199 follow-up review — verification of round-2 changes

Re-verified at `5b7cc1570a2`. Rory pushed 6 commits since my review, and the scope grew well beyond what I flagged: the created-comment path was reworked into a deterministic template gate plus an LLM intent classifier with spam minimization, and an offline eval harness was added.

---

### My findings: what landed

| # | Finding | Status |
|---|---|---|
| 🟠 `#1` | Template shrank to 2 mandatory lines, near-unreachable template check | ✅ Resolved, better than any option I offered |
| 🟠 `#2` | Duplicate scoring anchors dropped, prompt self-contradictory | ✅ Fixed in `59e4b1d570c` |
| 🟠 `#3` | Threshold 90 → 85 | ✅ Reverted in `21cb765946f` |
| 🟡 `#4` | "Posted by different users" never implemented | ✅ Fixed in `cbe6a110f88`, prompt + deterministic guard |
| 🟢 #5a | Which duplicate gets linked | ✅ Answered, tie-break to earliest added |
| 🟢 #5b | Undocumented deviations | ✅ Description now documents all of them |
| 🟢 #5c | Legacy automated-tests sections | ✅ Answered, agreed |
| 🟢 #5d | Stricter similarity parsing | ✅ Confirmed intentional |
| 🟢 #5e | No baseline for the model swap | ✅ Eval harness added |

##

### ✅ **Two of these deserve more than a checkmark:**

**`#2` was fixed better than I suggested.** I proposed keeping "at least 80% / at most 20%" and adding the anchors back. Rory's point in `59e4b1d570c` is correct and I was wrong: leaving explicit arithmetic next to an "over 90" anchor reproduces the exact contradiction, because the model can actually perform the arithmetic. Restating the priority qualitatively ("The SOLUTION section decides the score; the ROOT CAUSE only nudges it") and letting the anchors set the numbers removes the conflict instead of papering over it.

**`#4` got belt-and-braces treatment.** The `author` attribute on both `buildDuplicateCheckInput` and `buildDuplicateCheckSeedItem`, the prompt instruction, *and* the `comment.user?.login !== newProposalAuthor` guard on the confirmation site. The deterministic guard is what actually matters, and `ProposalPoliceConversationTest` now pins per-item author tagging.

**`#1`'s resolution is the interesting one.** Rather than any of my three options, `13dddf66114` inverts the design: template conformance is decided in code by `isProposal` (a definition, not a judgment), and only the leftovers get an LLM call to classify intent. That is a better answer than moving the check around a gate, and it closes the bid-spam hole I hadn't spotted. The `ProposalUtilsTest` added alongside it is the right place to pin that definition.

##

♻️ Reviewing the latest changes, see comments below 👇

evals/proposalPolice/commentIntent.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+                const {value, agreed} = majority(samples);
+                console.log(`${fixture.id}: ${agreed}/${RUNS_PER_FIXTURE} → ${value} (expected ${fixture.expectedIntent}) — ${fixture.note}`);
+
+                // Minimizing a real contributor's comment is the one outcome here that can't be walked back
+                // by the next run, so a comment that is not even an attempt must never be called spam —
+                // not on a majority, not on a single sample.
+                if (fixture.expectedIntent === CONST.INTENT.NOT_AN_ATTEMPT) {
+                    expect(samples).not.toContain(CONST.INTENT.SPAM);
+                }
```
ikevin127, line 37: #### 🟡 N2 Spam minimization now applies to every non-employee comment, and C+ reviewers are non-employees

Dropping the `Proposal` keyword gate is justified (the linked bid spam is real), but it means the intent classifier sees every human comment on every open `Help Wanted` issue, and `SPAM` collapses it publicly.

Credit where due: `evals/proposalPolice/fixtures/commentIntent.json` anticipates this. Five of the twelve `NOT_AN_ATTEMPT` fixtures are precisely the C+ shapes I was going to raise, including `commentIntent-feedback-disputing-root-cause` ("this has technical content") and `commentIntent-contributor-plus-proposal-selection`.

And `commentIntent.test.ts` enforces the right invariant, on every sample rather than the majority:

```ts
if (fixture.expectedIntent === CONST.INTENT.NOT_AN_ATTEMPT) {
    expect(samples).not.toContain(CONST.INTENT.SPAM);
}
```

That is the single best thing in this round. What remains is that the eval does not run in CI, so the prompt or `PROPOSAL_POLICE_MODEL` can change without it, and 20 fixtures is a thin net for "every comment on every Help Wanted issue."

**Suggested mitigation**, cheap and removable: emit the classification as an annotation on every non-`NOT_AN_ATTEMPT` run so the first weeks are auditable without digging through job logs.

```ts
core.notice(`ProposalPolice classified comment ${commentID} by ${newProposalAuthor} as ${intent}`);
```

Optionally hold the `minimizeCommentAsSpam` call behind a log-only period first. I would not block on this given `minimizeComment` preserves the text and is reversible from the UI, which is a deliberately better choice than the duplicate path's body overwrite.

ikevin127:
#### 🟡 N3 Per-comment LLM cost is now a multiple of what it was, and the PR does not quantify it

**Before**: only comments containing the literal `Proposal` reached a model. Now: one `gpt-5.6-luna` call per non-bot, non-employee created comment on an open `Help Wanted` issue. The label gate bounds it, but that is still a lot of "any update on this?" traffic.

`promptCacheKey: 'proposal-police-comment-intent'` keeps the instruction prefix cached, so the marginal cost is the comment body plus output, but the call count is what changed by an order of magnitude.

A rough calls-per-day and monthly figure in the description would stop this being a surprise on the OpenAI bill, and gives a number to compare against if you later want a cheap deterministic pre-filter back.

prompts/proposalPolice/messages.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Matches the banner a previous run prepended, along with the blank line separating it from the
+ * proposal, so the proposal can be recovered from a comment that already carries one.
+ * ProposalPoliceMessagesTest pins this against buildSubstantiveEditMessage so the two cannot drift.
+ */
+const SUBSTANTIVE_EDIT_MESSAGE_REGEX = /^🚨 Edited by \*\*proposal-police\*\*:[^\n]*\n+/;
+
+function buildTemplateReminderMessage(proposalAuthor: string | undefined): string {
+    return `⚠️ @${proposalAuthor} Thanks for your proposal. Please update it to follow the [proposal template](https://github.com/Expensify/App/blob/main/contributingGuides/PROPOSAL_TEMPLATE.md?plain=1), as proposals are only reviewed if they follow that format (note the mandatory sections).`;
+}
```
ikevin127, line 23: #### 🟢 N4 The spam branch posts copy written for proposals

Someone whose "please assign me, see my Upwork application" comment was just collapsed as spam is told "Thanks for your proposal. Please update it." They did not write a proposal and there is nothing to update.

This is the only copy contributors see from this feature, on a public repo, and it reads as a bot that did not understand what it was replying to.

Either neutral wording that works for both branches, or a second message for `SPAM`:

```ts
function buildJobClaimReminderMessage(author: string | undefined): string {
    return `⚠️ @${author} Thanks for your interest. To be considered for this job, please post a proposal following the [proposal template](https://github.com/Expensify/App/blob/main/contributingGuides/PROPOSAL_TEMPLATE.md?plain=1) (note the mandatory sections). Comments claiming the job without one are not reviewed.`;
}
```

.github/libs/ProposalUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Checks if a comment body matches the criteria for a Proposal.
+ */
+function isProposal(body: string | null | undefined): boolean {
+    if (!body) {
+        return false;
+    }
+    const lowerCaseBody = body.toLowerCase();
+    return body.includes(CONST.PROPOSAL_KEYWORD) && lowerCaseBody.includes(CONST.PROPOSAL_HEADER_A) && lowerCaseBody.includes(CONST.PROPOSAL_HEADER_B);
```
ikevin127: #### 🟢 N5 Case sensitivity of `PROPOSAL_KEYWORD` is now load-bearing for whether a contributor gets nagged publicly

`ProposalUtilsTest` pins this as intended:

```ts
it('requires the keyword to be capitalized, so ordinary prose about a proposal does not qualify', () => {
    expect(isProposal(buildProposal('## proposal', ROOT_CAUSE_HEADER, 'cause', SOLUTION_HEADER, 'fix'))).toBe(false);
});
```

A contributor who writes `## proposal` with both mandatory sections filled in correctly now fails the deterministic gate, gets an intent call, comes back `GENUINE_ATTEMPT`, and is publicly told to follow the template they did follow. On `main` this silently did nothing, because the keyword gate returned early.

The two section headers are already matched case-insensitively in the same function, so the inconsistency is arbitrary and now has a user-visible consequence. Lowercasing the keyword check alongside them costs nothing:

```suggestion
    return lowerCaseBody.includes(CONST.PROPOSAL_KEYWORD.toLowerCase()) && lowerCaseBody.includes(CONST.PROPOSAL_HEADER_A) && lowerCaseBody.includes(CONST.PROPOSAL_HEADER_B);
```

If the capitalization requirement is load-bearing for some case I am not seeing, the test name should say why rather than just asserting the behaviour.

ikevin127:
### Still open from round 1 review, unaddressed

None of these were replied to, so I assume they were lost rather than declined. Ordered by what I would still push for:

- 🟡 **`<new_proposal>` versus `<proposal>` tag split.** Still both, now both carrying `author`. The duplicate-check input is persisted verbatim into the Conversation, so prior proposals sit in history tagged `<new_proposal>` while seeds and refreshes use `<proposal>`. On an issue with three post-migration proposals the model sees three `<new_proposal>` messages and has to guess which is current.
- 🟡 **Assistant verdicts accumulate in the Conversation forever.** `findConversationItemIDForComment` only matches items containing `comment_id="`, so every duplicate-check output item stays. The withdrawn-proposal cleanup removes the proposal but leaves a verdict naming its ID.
- 🟡 **`deleteConversationItem` and `listConversationItems` still have no direct test.** `OpenAIUtilsTest` mocks only `conversations.create` and `conversations.items.create`. `deleteConversationItem` is the method that removes data and it is exercised only through the action's mocks. Same for `findConversationItemIDForComment`, still no direct test.
- 🟡 **Banner guard `includes` versus `startsWith`, and the untrimmed strip.** Unchanged: `payload.comment.body.trim().includes(SUBSTANTIVE_EDIT_MESSAGE_PREFIX)` guards, `payload.comment.body.replace(SUBSTANTIVE_EDIT_MESSAGE_REGEX, ...)` strips against a `^` anchor.
- 🟢 `comment_id="N"` and now `author="..."` collisions, since `escapeForXMLWrapper` leaves quotes alone.
- 🟢 Null-user handling in `buildSeedItems` still uses the double negative and disagrees with `findTrackedConversationID`.
- 🟢 `as Partial<...>` casts in the type guards, and the dead `?? -1` in `isDuplicateCheckResponse`.
- 🟢 `ProposalUtils.ts` should be `isProposal.ts`. Now cemented by a matching `ProposalUtilsTest.ts`, so renaming gets more annoying every round.
- 🟢 `VALID_PROPOSAL_BODY` and `makeComment` still duplicated across two test files, and `ProposalComment & {html_url: string; node_id: string}` grew another bolted-on field this round. Both are fields the action genuinely reads, so they belong on `ProposalComment`.
- 🟢 No test pins instruction isolation. The evals partly cover this by running each call's real instructions, but nothing asserts that the edit check never receives duplicate-detection rules.

ikevin127:
### Approval status

**Not approvable yet, for mechanical reasons rather than review ones.**

| Gate | Status |
|---|---|
| CI | 17 checks passing. Only `Check independent approval` fails, which is waiting on me |
| `Expensify/GitHub-Actions#101` | ✅ Merged 2026-08-12, no longer blocking |
| Mergeability | ❌ `CONFLICTING` / `DIRTY`, 951 commits behind `main` |
| Review decision | `REVIEW_REQUIRED` |

The conflicts are in `config/eslint/eslint.config.mjs`, `package.json`, and `tsconfig.json`, all three from the evals setup landing in files that move constantly on `main`. Purely mechanical, but they need resolving and the bundles rebuilding afterwards.

**My position:** every blocking finding from round 1 is resolved, and `#1` and `#2` were resolved better than I proposed. Of the new findings, 🟠 N1 is the only one I would hold approval for, because it is a silent misconfiguration that ends with employee comments being collapsed as spam and it costs one API call to close. N2 through N5 I am happy to see handled here or in a follow-up, author's call.

Once N1 is in and `main` is merged, I'll approve. Given the Assistants shutoff on 2026-08-26 is 12 days out, I would not let N2 through N5 hold this.

**One process note**: this is now materially more than an API migration. Auto-collapsing contributor comments as spam is a policy change about how the repo treats external contributors, and the author checklist has a line about new patterns being agreed by multiple Expensify engineers. Worth a second internal sign-off on the spam behaviour specifically, separate from my review of the code.

**Short version**: all four blocking items from my review are fixed, two of them better than I proposed. The rework introduced one thing I'd hold on (the team-slug 404 ambiguity in the workflow, which silently makes every employee spam-eligible if the slug or token scope ever changes), plus four minor items. The actual blockers to merging are the conflicts with main and my approval, not open review threads.

cc @roryabraham

.github/libs/ProposalUtils.ts
roryabraham: Fixed in `d504199376a`. Lowercased the keyword check so it matches the two section headers beside it, and `isProposal.test.ts` now pins the opposite of what it pinned before: `## proposal` with both sections filled in is a proposal.

Also renamed `ProposalUtils.ts` to `isProposal.ts` while touching it, per your other note.

prompts/proposalPolice/messages.ts
roryabraham: Fixed in `d504199376a`. Added `buildJobClaimReminderMessage` for the spam branch, close to your wording. The template reminder is unchanged for `GENUINE_ATTEMPT`, so it keeps its character-for-character parity.

Covered in `ProposalPoliceMessages.test.ts`, which asserts the spam copy never contains "your proposal".

evals/proposalPolice/commentIntent.test.ts
roryabraham: Added the `core.notice` in `d504199376a`, so every non-`NOT_AN_ATTEMPT` classification shows up on the run itself rather than only in job logs.

Not doing the log-only period. The eval is the mitigation, and it now runs in CI's runtime — the suites moved to `tests/tooling` under `bun:test` in `4461eac011f` — though the evals themselves still don't run in CI, since each run spends tokens.

roryabraham:
Measured it and added the figures to the description.

Over the 7 days to 2026-08-15, across the 61 open `Help Wanted` issues: 303 non-bot comments, 98 of which already follow the template and cost nothing. So **~29 intent calls/day, ~880/month**. That's an upper bound — Expensify employees are filtered out before the action runs and aren't excluded from the count. Duplicate-check volume is unchanged at ~14/day.

roryabraham:
All ten are done in `d504199376a`, plus the test move in `4461eac011f`.

One correction on the framing: these weren't lost. None of them were posted in round 1 — I checked every comment on the PR, and each of these appears for the first time in the comment above. Not a complaint, just so you're not left thinking a reply went missing.

- **`<new_proposal>` vs `<proposal>`.** Both are `<proposal>` now, and the prompt names the last message as the one under review. `buildDuplicateCheckSeedItem` delegates to `buildDuplicateCheckInput` so the two can't drift.
- **Verdicts accumulating.** Pruned after each duplicate check via `findVerdictItemIDs`. `findConversationItemIDForComment` also filters to user messages now, so a verdict naming a comment ID can't be mistaken for the proposal itself — there's a test for exactly that.
- **No direct tests for the delete path.** `OpenAIUtils.test.ts` covers `listConversationItems` (including pagination) and `deleteConversationItem`, asserting the argument order, since swapping it would delete nothing and fail silently.
- **Banner guard `includes` vs `startsWith`.** Now `startsWith`, and the strip runs against the same trimmed body the guard checked.
- **Attribute collisions.** `escapeForXMLWrapper` escapes quotes as well as angle brackets.
- **Null-user double negative.** `buildSeedItems` and `findTrackedConversationID` read the field the same way.
- **`as Partial<...>` casts and the dead `?? -1`.** Guards narrow with `in` instead; the coalesce is gone.
- **`ProposalUtils.ts`.** Renamed to `isProposal.ts`, test alongside.
- **Duplicated test helpers.** `VALID_PROPOSAL_BODY` and `makeComment` live in `tests/tooling/proposalPoliceFixtures.ts`. `html_url` and `node_id` moved onto `ProposalComment`, which also let the cast in `mockComments` go.
- **Instruction isolation.** `ProposalPoliceInstructions.test.ts` asserts each call gets the role and template and none of the other two calls' rules.

roryabraham:
### On N1

There's no N1 on the PR. The findings go N2, N3, N4, N5 — it's referenced only in your approval comment, so I think it's sitting in an unsubmitted pending review that only you can see.

Going from your one-line description ("the team-slug 404 ambiguity in the workflow, which silently makes every employee spam-eligible if the slug or token scope ever changes"): the ambiguity is real, and I checked rather than assumed —

```
$ gh api /orgs/Expensify/teams/this-team-does-not-exist/memberships/roryabraham
gh: Not Found (HTTP 404)          # broken slug
$ gh api /orgs/Expensify/teams/expensify-expensify/memberships/octocat
gh: Not Found (HTTP 404)          # genuine non-member
```

So the three-way check separates errors from 404s, but can't separate those two 404s.

**I'm not fixing it.** It takes a deliberate config change to trigger — a team rename or the token losing `read:org` — and the blast radius is small: employees on `Help Wanted` issues are giving direction, so their comments classify `NOT_AN_ATTEMPT` and nothing happens to them. Closing it means resolving the team on every comment, which is a permanent API call per comment to guard against a config change that would break plenty of other things too.

Worth naming: "fail loudly" isn't available here as an alternative. A broken slug doesn't produce an error to be loud about, it produces a successful 404 meaning "not a member". Detecting it *is* the extra call.

If you think that's the wrong call, submit N1 and I'll take another look — you may have a failure mode in mind that I'm not seeing from the summary line.

tests/tooling/proposalPoliceComment.test.ts
```diff
@@ -0,0 +1,572 @@
...
+        mockFn.mockReset();
+    }
+
+    mockGetAllCommentDetails.mockResolvedValue([]);
+    mockCreateConversation.mockResolvedValue({id: 'conv_new', created_at: 0, metadata: null, object: 'conversation'});
+    mockListConversationItems.mockResolvedValue([]);
+    // Bypass the real JSON-schema validators here; OpenAIUtils.test.ts already covers parseJSONResponse itself.
+    mockParseJSONResponse.mockImplementation((text) => JSON.parse(text));
+}
+
```
ikevin127, line 139: #### 🟢 The verdict-pruning loop has no integration test

`ProposalPoliceConversation.test.ts` covers `findVerdictItemIDs` in isolation, and `OpenAIUtils.test.ts` covers `deleteConversationItem`.

But every action test's mock returns user messages only:

```ts
function conversationMessage(itemID: string, commentID: number): ConversationItem {
    return {id: itemID, type: 'message', role: 'user', status: 'completed', content: [...]};
}
```

So `findVerdictItemIDs` returns `[]` in all 27 tests and the loop body in `run()` never executes. The pieces are tested, the wiring is not. Cheapest fix is inside the test that already exercises the delete path:

```ts
mockListConversationItems.mockResolvedValue([
    conversationMessage('item_42', 42),
    conversationMessage('item_99', 99),
    {id: 'item_verdict', type: 'message', role: 'assistant', status: 'completed', content: [{type: 'output_text', text: '{"similarity":95,"duplicateCommentID":42}'}]},
]);
```

then assert both `item_verdict` and `item_99` are deleted. The existing `toHaveBeenCalledTimes(1)` becomes `2`, which is the assertion that would have caught a prune loop that silently does nothing.

ikevin127:
#### 🟢 Worth one check during live testing

`findVerdictItemIDs` matches `type === 'message' && role === 'assistant'`. If `gpt-5.6-luna` persists reasoning items into the Conversation, they are not `type: 'message'` and would accumulate past the prune. After the first real duplicate check on a live issue, list that Conversation's items once and confirm only `<proposal>` user messages remain.

If reasoning items do show up, widening the filter is a one-line change.

ikevin127 (approved):
**🟢 LGTM**

### Two new notes above ☝️ , neither blocking

All 14 items from rounds 1 and 2 are resolved, N1 was my error (never posted, your counter-argument on the merits is sound). The only real blocker left is an ESLint failure inherited from the main merge, in `src/ files` this PR never touches.

chuckdries (dismissed):

chuckdries (approved):

PR #97208, opened 2026-07-28
[No QA] Seed the AdHoc build cache from main so PR test builds hit it warm

JakubKorytko:
### Explanation of Change

Rock keys its remote cache by `traits + configuration + fingerprint`, and on both platforms pushes to `main` build a different variant than PR test builds do - `simulator`/`Debug` vs `device`/`AdHoc` on iOS, `developmentDebug`/`Debug` vs `Adhoc` on Android. Different keys, different S3 namespaces, so nothing warms the one PR test builds read from and the first PR on any new native fingerprint always compiles for ~8 minutes.

This adds a `Seed AdHoc build cache` workflow that runs the same AdHoc builds PR test builds run, on pushes to `main`. Per platform: ask Rock whether the cache already holds an artifact for the current fingerprint, build only if it doesn't, and have that build confirm before it finishes that it actually uploaded something.

**Why it calls the shared build workflows rather than adding a matrix entry.** Trait parity by construction - destination, configuration, scheme, signing material and the pinned Rock action version all come from the workflow PR test builds already use, so the cache key cannot drift. A matrix entry wouldn't have worked anyway: the Rock action SHA pinned in the remote-build workflows predates the `provisioning-profiles` array our AdHoc signing needs.

**Why it's a separate workflow.** The remote-build workflows use a serialized concurrency group on `main` (`cancel-in-progress: false`), so an 8-minute compile inside one widens the window in which pushes queue and older pending runs get dropped - the simulator build's own signal would get slower in order to warm a cache it doesn't read. And `failureNotifier.yml` watches both by name, so a seed failure (an expired AdHoc provisioning profile, say) would file an issue assigned to whoever merged the unrelated PR that happened to trigger that run. On its own trigger it costs those workflows nothing: both are byte-identical to `main` in this PR.

**Failures here are deliberately not routed to anyone.** Nothing depends on this workflow. When it fails, PR test builds just compile their own native code again - slower, never wrong.

**Honest framing.** This is a latency shift, not a compute saving. A fingerprint that one PR would have compiled gets compiled on `main` instead. It wins when several PRs want the same new fingerprint - the common case, since most PRs are JS-only - and loses when a seeded fingerprint is never requested. What it buys is that nobody waits 13 minutes for a test build unless their own PR changed native code. On a sample of 14 recent iOS test builds, ~3 (~20%) convert from ~13 minutes to ~5.

| File | Change |
|---|---|
| `seedAdHocCache.yml` | new. check -> seed, per platform, on pushes to `main` and on dispatch. The seed verifies itself before it finishes |
| `checkRockCache.yml` | new. Reusable lookup: asserts the S3 provider, computes the fingerprint, and reports whether an artifact exists |
| `buildIOS.yml`, `buildAndroid.yml` | `seed-cache-only` makes a cache hit stop right after the lookup, and makes the build ask the bucket at the end whether the artifact it was meant to write is actually there. The traits it verifies come from the same `variant` expression that configures the build, so they cannot name a different namespace than the one just built. `expected-fingerprint` covers the other half of the cache key. Both inert unless passed |

<details>
<summary>Why existing builds and deploys can't be affected</summary>

**All four pre-existing callers resolve to the literals they had.** `deploy.yml` and `verifyHybridApp.yml` pass `variant: Release`, `buildAdHoc.yml` passes `Adhoc`, none passes `seed-cache-only`, so `re-sign: ${{ !inputs.seed-cache-only }}` is `true` and `ad-hoc` is unchanged. They also fail safe: `seed-cache-only` is a typed boolean defaulting to `false`, so re-signing can't be switched off for an existing caller by accident. The step that verifies a seed landed is gated on the same input, so it never runs for them either. Release builds key on `device,Release` while seeds write `device,AdHoc`, so a seed can't contaminate what a deploy downloads.

**The cache can't be poisoned.** Upload is gated on `!ARTIFACT_URL` plus a re-check immediately before uploading, so a seed can only fill an empty slot, never overwrite. The artifact it writes is the same kind PR builds already write, because on a miss the re-sign path never executes.

**PR test builds keep their own JS.** The cache-hit path runs `rock sign:ios --build-jsbundle` / `rock sign:android --build-jsbundle`, which rebuilds the bundle from the PR checkout, so a cached artifact contributes native code only. `inject-ci-data.sh` only rewrites `src/CONST/CI.ts`, which is part of that rebuilt bundle.

**Secrets.** The lookup workflow takes three named secrets rather than `inherit`, since it only needs the submodule and the bucket. A dispatch passes the same gate the remote-build workflows apply (`validateBuildRequest.yml`: Expensify employee with write access, plus a confirmation they reviewed the code at that ref), because like them it builds whatever ref it names with the build secrets in scope.

</details>

### Fixed Issues

$ https://github.com/Expensify/App/issues/96950
PROPOSAL: N/A

### Tests

**⚠️ These need an Expensify internal - I have `pull` access only, so I can neither dispatch a workflow nor push a branch to `Expensify/App`.** Nothing in this PR has executed yet, and tests 1 and 2 can be run *before* merge, which I'd strongly prefer: test 1 needs push access for one throwaway branch plus two dispatches, test 2 needs a test build on any JS-only PR.

**1. Dry run, before merge**

**Nothing in this PR is exercised by CI.** `verifyHybridApp.yml` is the only workflow that would run `buildIOS.yml` / `buildAndroid.yml` before merge, and it is excluded twice over: its `paths` filter doesn't list `.github/**`, and its jobs are gated on `!github.event.pull_request.head.repo.fork`. So this dry run is the only pre-merge signal available.

`workflow_dispatch` resolves its `ref` only against branches in the repo hosting the workflow, so neither this PR's head nor anything on my fork is dispatchable, and `Seed AdHoc build cache` isn't dispatchable either until it's on `main`. What *is* reachable: the remote-build workflows are already on `main`, a dispatch runs the workflow file from whichever ref it names, and everything worth validating here lives in the reusable workflows those can call. So branch `korytko/seed-dry-run` on my fork carries this PR's logic with the check and seed jobs borrowed back into the remote-build workflows and the `main`-only guard dropped. The verification needs no borrowing - it lives inside the build workflows those jobs already call. It needs to be placed on `Expensify/App` by someone with push access - I don't have it:

```
git fetch https://github.com/software-mansion-labs/expensify-app-fork.git korytko/seed-dry-run
git push origin FETCH_HEAD:refs/heads/korytko/seed-dry-run
```

Pointing this at a branch is safe and isn't wasted work: the cache is keyed by fingerprint rather than branch, that branch's only diff from `main` is under `.github/**` (not a fingerprint input, so its fingerprint *is* `main`'s), and an upload can only fill an empty slot. A successful dry run seeds the real cache. It also runs each workflow's own simulator matrix, which is the cost of borrowing them. Delete the branch afterwards - `git push origin --delete korytko/seed-dry-run` - it is not meant to be merged.

1. Dispatch both, selecting ref `korytko/seed-dry-run` and ticking `reviewed_code`. (The push itself may already have triggered them, since all the branch's commits touch only `.github/**` and so may or may not be filtered out by `paths-ignore` - if it did, just read those runs instead.)
   - `gh workflow run 'Remote Build iOS' --ref korytko/seed-dry-run -f reviewed_code=true`
   - `gh workflow run 'Remote Build Android' --ref korytko/seed-dry-run -f reviewed_code=true`
2. Open `Check Rock remote cache` and verify it asserts the provider is `S3`, logs a fingerprint, and reports `NEEDS_BUILD=true` on a cold namespace
3. Open the seed job's `Compare fingerprint against the one the caller looked up` step and verify it reports a match. The iOS lookup runs on Linux while the build runs on macOS, and if the two disagree the iOS gate is guarding nothing - harmless, but then it should be dropped for iOS. I checked the inputs locally and expect a match: of the fingerprint's 2270 inputs, 2266 are content hashes and the four JSON ones (`autolinkingSources`, `env`, `reactNativeVersion`, `scripts`) contain no absolute host paths, no `darwin`/`macos`/`arm64` strings and no host-specific packages among the 52 autolinked dependencies. This step is the confirmation rather than the open question
4. Verify the `Rock Remote Build` step is a real ~8 minute compile ending in `Upload ZIP Artifact for caching`, and that no `Download and Unpack IPA`, `Re-sign IPA` or `Upload for Ad-hoc distribution` steps ran
5. Verify the artifact name is `rock-ios-device-AdHoc-<fingerprint>` / `rock-android-Adhoc-<fingerprint>`, with no PR number or commit SHA in it, since that is the namespace PR test builds read
6. Verify the seed job's last step, `Verify the seed landed in the remote cache`, logs `Seed landed for traits '...'`, and that those traits are the ones the lookup in step 2 was given - the same namespace step 5 just confirmed the artifact was named for
7. Dispatch the same thing a second time and verify the lookup now logs `Already cached: ...`, reports `NEEDS_BUILD=false`, and that the seed job is *skipped* rather than failed, taking its verification step with it
8. Verify the two pre-existing `build (...)` matrix jobs ran exactly as before in both runs

**2. PR test builds get the warm cache and keep their own JS**

1. With a fingerprint seeded by test 1, open a PR that changes only JS
2. Trigger `Build and deploy apps for testing` for that PR
3. Open the `Build iOS` job and time the `Rock Remote Build - iOS` step
4. Verify it hits the cache at roughly 100s instead of cold compiling for roughly 510s
5. Install the ad-hoc build from the link posted on the PR
6. Verify the app runs and reports this PR's number, confirming the JS bundle was rebuilt during re-sign and the cached native binary did not leak `main`'s CI data
7. Verify the PR comment and build summary still contain working iOS and Android download links, so ad-hoc distribution is unchanged for PR builds

**3. After merge: the push trigger**

This is the only part a dry run can't cover, since `.github/**` is in `paths-ignore` and this branch's own pushes never fire it.

1. Dispatch `Seed AdHoc build cache` from `main` once, ticking `reviewed_code`
2. On an already-warm fingerprint, verify the whole run is two cheap lookups and two skipped builds - which is itself confirmation the trigger and the gate are wired up
3. Watch the next push to `main` that changes native inputs (a `Mobile-Expensify` submodule bump, or a file added under `patches/`) and verify the seed runs and lands

Reverting is deleting `seedAdHocCache.yml`. The inputs added to the shared build workflows are inert unless passed, and `checkRockCache.yml` has no other callers.

**Known limitations, follow-ups rather than blockers**

- **Native code under `modules/` isn't in the fingerprint.** `calculateFingerprint` reads only `sourceDir` plus `extraSources`, and the autolinking input records dependency roots rather than file contents, so editing `modules/hybrid-app/{android,ios}/**` doesn't move the hash and Rock serves a stale binary. Pre-existing, but it only bites when the fingerprint is already cached - and making fingerprints cached is what this PR does. ~18 commits touched native code under `modules/` in the last 6 months. Fix is adding those directories to `extraSources`, which invalidates every cache namespace once and so deserves its own PR; until then, PRs touching `modules/` should use `FORCE_NATIVE_BUILD`.
- **`scripts/**` is in `paths-ignore`,** so a change to `scripts/compute-patches-hash.sh` alone moves the fingerprint without triggering any build, seed or otherwise. Pre-existing; also affects the simulator seeding that already happens today.

### Offline tests

N/A. This PR only changes GitHub Actions workflows.

### QA Steps

N/A. This PR only changes GitHub Actions workflows.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/.github/workflows/buildAndroid.yml
+++ b/.github/workflows/buildAndroid.yml
@@ -27,6 +27,14 @@ on:
         description: Force a full native build, bypassing Rock remote cache
         type: string
         default: 'false'
+      seed-cache-only:
+        description: Build only to warm the Rock remote cache. Skips re-signing and ad-hoc distribution, so a cached fingerprint costs nothing but the lookup.
+        type: string
+        default: 'false'
+      expected-fingerprint:
+        description: Fingerprint a caller already computed elsewhere. Seed builds warn when theirs differs, which means the caller's cache lookup was answering for a different build.
+        type: string
+        default: ''
 
     outputs:
       VERSION_CODE:
@@ -160,6 +168,25 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
 
+      # A caller that skips this build on a cache hit is only as good as its fingerprint matching the one used
+      # here. Warn rather than fail: a mismatch costs the savings, never correctness, because the build below
+      # repeats the lookup itself and stops early when it does hit.
+      - name: Compare fingerprint against the one the caller looked up
+        if: ${{ inputs.seed-cache-only == 'true' && inputs.expected-fingerprint != '' }}
+        env:
+          IS_HYBRID_APP: 'true'
+          EXPECTED_FINGERPRINT: ${{ inputs.expected-fingerprint }}
+        run: |
+          ACTUAL_FINGERPRINT=$(npx rock fingerprint -p android --raw 2>/dev/null) || ACTUAL_FINGERPRINT=''
+          if [ "$ACTUAL_FINGERPRINT" != "$EXPECTED_FINGERPRINT" ]; then
+            echo "::warning::The caller looked up $EXPECTED_FINGERPRINT but this build computed ${ACTUAL_FINGERPRINT:-nothing}, so its cache check is guarding a different fingerprint and can never skip this build."
+            echo "### Fingerprint mismatch" >> "$GITHUB_STEP_SUMMARY"
+            echo "- Caller looked up \`$EXPECTED_FINGERPRINT\`" >> "$GITHUB_STEP_SUMMARY"
+            echo "- This build computed \`${ACTUAL_FINGERPRINT:-nothing}\`" >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "Fingerprint matches the caller's lookup: $ACTUAL_FINGERPRINT"
+          fi
+
       - name: Rock Remote Build - Android
         uses: callstackincubator/android@4cedf4d9b5c167452c96fe67233577e0fde9a025
         env:
@@ -171,8 +198,10 @@ jobs:
           variant: ${{ inputs.variant }}
           aab: ${{ inputs.variant == 'Release' }}
           sign: true
-          re-sign: true
-          ad-hoc: ${{ inputs.variant == 'Adhoc' }}
+          # A seed build only needs to upload the base artifact it just compiled, which happens on a cache miss
+          # regardless of these two. Leaving them off means a cache hit stops right after the lookup.
+          re-sign: ${{ inputs.seed-cache-only != 'true' }}
+          ad-hoc: ${{ inputs.variant == 'Adhoc' && inputs.seed-cache-only != 'true' }}
           # We don't need Rock to setup java, because we already ran the setup-java action above
           setup-java: false
           keystore-file: './upload-key.keystore'
--- a/.github/workflows/buildAndroid.yml
+++ b/.github/workflows/buildAndroid.yml
@@ -28,9 +28,12 @@ on:
         type: string
         default: 'false'
       seed-cache-only:
+        # Typed boolean rather than a string like `force-native-build` above, which has to stay a string because it is
+        # threaded through from a dispatch input. This one is only ever set to a literal, and actionlint rejects a
+        # quoted 'true' passed to a string input, so a real boolean is both truthful and lintable.
         description: Build only to warm the Rock remote cache. Skips re-signing and ad-hoc distribution, so a cached fingerprint costs nothing but the lookup.
-        type: string
-        default: 'false'
+        type: boolean
+        default: false
       expected-fingerprint:
         description: Fingerprint a caller already computed elsewhere. Seed builds warn when theirs differs, which means the caller's cache lookup was answering for a different build.
         type: string
@@ -172,7 +175,7 @@ jobs:
       # here. Warn rather than fail: a mismatch costs the savings, never correctness, because the build below
       # repeats the lookup itself and stops early when it does hit.
       - name: Compare fingerprint against the one the caller looked up
-        if: ${{ inputs.seed-cache-only == 'true' && inputs.expected-fingerprint != '' }}
+        if: ${{ inputs.seed-cache-only && inputs.expected-fingerprint != '' }}
         env:
           IS_HYBRID_APP: 'true'
           EXPECTED_FINGERPRINT: ${{ inputs.expected-fingerprint }}
@@ -180,9 +183,11 @@ jobs:
           ACTUAL_FINGERPRINT=$(npx rock fingerprint -p android --raw 2>/dev/null) || ACTUAL_FINGERPRINT=''
           if [ "$ACTUAL_FINGERPRINT" != "$EXPECTED_FINGERPRINT" ]; then
             echo "::warning::The caller looked up $EXPECTED_FINGERPRINT but this build computed ${ACTUAL_FINGERPRINT:-nothing}, so its cache check is guarding a different fingerprint and can never skip this build."
-            echo "### Fingerprint mismatch" >> "$GITHUB_STEP_SUMMARY"
-            echo "- Caller looked up \`$EXPECTED_FINGERPRINT\`" >> "$GITHUB_STEP_SUMMARY"
-            echo "- This build computed \`${ACTUAL_FINGERPRINT:-nothing}\`" >> "$GITHUB_STEP_SUMMARY"
+            {
+              echo "### Fingerprint mismatch"
+              echo "- Caller looked up \`$EXPECTED_FINGERPRINT\`"
+              echo "- This build computed \`${ACTUAL_FINGERPRINT:-nothing}\`"
+            } >> "$GITHUB_STEP_SUMMARY"
           else
             echo "Fingerprint matches the caller's lookup: $ACTUAL_FINGERPRINT"
           fi
@@ -200,8 +205,8 @@ jobs:
           sign: true
           # A seed build only needs to upload the base artifact it just compiled, which happens on a cache miss
           # regardless of these two. Leaving them off means a cache hit stops right after the lookup.
-          re-sign: ${{ inputs.seed-cache-only != 'true' }}
-          ad-hoc: ${{ inputs.variant == 'Adhoc' && inputs.seed-cache-only != 'true' }}
+          re-sign: ${{ !inputs.seed-cache-only }}
+          ad-hoc: ${{ inputs.variant == 'Adhoc' && !inputs.seed-cache-only }}
           # We don't need Rock to setup java, because we already ran the setup-java action above
           setup-java: false
           keystore-file: './upload-key.keystore'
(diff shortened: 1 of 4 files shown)


.github/workflows/checkRockCache.yml
```diff
@@ -0,0 +1,131 @@
...
+          echo "FINGERPRINT=$FINGERPRINT" >> "$GITHUB_OUTPUT"
+          echo "### Rock remote cache: \`$PLATFORM\` \`$TRAITS\`" >> "$GITHUB_STEP_SUMMARY"
+          echo "- Fingerprint: \`${FINGERPRINT:-unknown}\`" >> "$GITHUB_STEP_SUMMARY"
+
+          # A lookup that errors out may just be a bad minute for S3, so warn and let the build decide for itself.
+          if ! OUTPUT=$(npx rock remote-cache list -p "$PLATFORM" --traits "$TRAITS" --json); then
+            echo "::warning::Could not query the remote cache, so assuming a build is needed"
+            echo "- Lookup failed, assuming a build is needed" >> "$GITHUB_STEP_SUMMARY"
+            echo "NEEDS_BUILD=true" >> "$GITHUB_OUTPUT"
+            exit 0
```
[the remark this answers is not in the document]
JakubKorytko: good catch, fixed

mateuuszzzzz (approved):

.github/workflows/buildAndroid.yml
```diff
@@ -160,6 +178,27 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
 
+      # A caller that skips this build on a cache hit is only as good as its fingerprint matching the one used
```
mateuuszzzzz, line 174: NAB: We could create composite action and re-use it in those 2 iOS and Android workflows to prevent drift between those 2 implementations

But... I also have one concern after rethinking this PR. In the past we did not have any extra guards that assure that build landed in AWS bucket.

I know this suppose to give us extra context about the error. I.e. without this we don't know if workflow failed due to build error or something else.

Nevertheless, I assume that if the upload fails, we can still investigate the logs to determine what happened. This would only save a bit of time while also adding complexity to an already quite complex chain of workflow calls. So finally I think we should not add such verification checks anywhere cc: @war-in @JakubKorytko for thoughts.

A single test run should be enough to validate that we haven't missed any issues in the current implementation.

JakubKorytko: you're right that it was overbuilt, cut it from 45 lines to 14 (-58 across the 4 files). dropped the `cache-traits` input, the `TRAITS` output on `checkRockCache.yml` and the pass-throughs - traits are just derived from `inputs.variant` in the same file now. also dropped the retry loop, provider-s3 builds the S3Client with SDK defaults so it already retries 5xx/throttling 3x with backoff, my loop was doing it again 5s apart for no reason.

kept the check itself though. every other caller of these workflows has someone waiting on the artifact, so a broken upload is loud. the seed is the first one where nothing consumes the output, so if it stops uploading the only symptom is "test builds feel slow again" and nothing points at this workflow. the two failure messages are distinct so a red run tells you which kind it is.

composite action - at 14 lines where the only diff is `-p ios` vs `-p android` I think it costs more than it saves, but happy to add it if you disagree.

the fingerprint compare step is warn-only and temporary, it's there to check the linux lookup and the macos build agree. once the dry run confirms it I'll drop it.

agreed one real run validates the rest. if you and @war-in still want the verification gone entirely, say so and I'll delete it, won't argue further.

.github/workflows/buildIOS.yml
```diff
@@ -304,3 +341,29 @@ jobs:
         with:
           name: ${{ inputs.artifact-prefix }}ios-sourcemap-artifact
           path: Mobile-Expensify/main.jsbundle.map
+
+      # A seed build that finishes without caching anything looks exactly like a working one, and nothing downstream
+      # notices: PR builds just compile their own native code again, which reads as slow CI rather than as a failure.
+      # So ask the bucket the same question a PR build's lookup asks, rather than trusting the build action's own
+      # report - an action bump could move the gate on its upload step and break the upload and its self-report
+      # together. Last in the job so a failure here costs none of the uploads above.
+      - name: Verify the seed landed in the remote cache
```
elirangoshen, line 350: This verify step and the fingerprint-comparison step above are duplicated almost verbatim in `buildAndroid.yml`. Not blocking, but a composite action under `.github/actions/composite/` (taking `platform` and `traits`) would keep the two copies from drifting apart when one of them gets tweaked later.

PR #97211, opened 2026-07-28
fix: keep Search Router results above offline indicator

nabi-ebrahimi:
### Explanation of Change

On native devices, Search Router can open during the initial keyboard animation before its available height is settled. If results appear during this timing window, the list may use the full-screen height, place the offline indicator behind the keyboard, and incorrectly disable scrolling.

This change limits the native Search Router to the visible area above the keyboard. The limit updates during keyboard transitions and orientation changes and is removed when the keyboard closes.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97182
PROPOSAL: 

### Tests

**Cold-start offline scrolling**

1. Copy `type:` to the clipboard.
2. Force-stop and reopen the app.
3. Enable forced offline mode.
4. Open Search Router.
5. Immediately paste `type:` into the search input.
6. Try to scroll through the results without closing or reopening the keyboard.

**Expected:** The results scroll immediately on the first attempt.

### Offline tests
Same as Tests.

### QA Steps
Same as Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/ca8c8cdf-0a0b-4362-b6b4-c2ce3a96bdad

</details>

<details>

</details>

<details>

>

https://github.com/user-attachments/assets/0624b333-a11d-4638-b168-4717e161453c

</details>

<details>

</details>

<details>

>

https://github.com/user-attachments/assets/552c96b2-88b7-4906-81b5-0eb6d90f681f

</details>


src/components/Search/SearchRouter/SearchRouter.tsx
```diff
@@ -482,7 +482,7 @@ function SearchRouter({onRouterClose, shouldHideInputCaret, isSearchRouterDispla
 
     return (
         <View
-            style={[styles.flex1, modalWidth, styles.h100, !shouldUseNarrowLayout && styles.mh85vh]}
+            style={[styles.flex1, modalWidth, !shouldUseNarrowLayout && [styles.h100, styles.mh85vh]]}
```
marufsharifi: The PR description says this "adds offline-indicator bottom clearance using the existing `useBottomSafeSafeAreaPaddingStyle` hook" and that it "applies only on narrow layouts while offline." Neither is true in the code:

- There's no `useBottomSafeSafeAreaPaddingStyle` usage anywhere in this diff.
- The condition here is purely `!shouldUseNarrowLayout` — there is **no offline gate**. So `h100` is dropped for **all** narrow layouts, online *and* offline.

This means online narrow-layout behavior also changes, contradicting "Online … behavior remains unchanged." Please either:
1. update the description to match what the code actually does, and confirm online narrow layout (empty query / few results / many results) has no regression, or
2. if the intent really is offline-only, gate it on `isOffline`.

src/components/Search/SearchList/ListItem/TextWithIconCell.tsx
```diff
@@ -13,13 +13,14 @@ import {View} from 'react-native';
 
 type TextWithIconCellProps = {
     icon: IconAsset;
+    iconSize?: number;
```
marufsharifi, line 16: Clean, backward-compatible prop extraction — the `= 12` default preserves every existing caller, the type is correct, and no casting. 👍

Caveat: `TextWithIconCell` is a shared Search cell. Per the shared-component checklist item, please confirm no other consumer visually depended on the hard-coded `12` in a way the default now hides. Low risk, but worth a quick look at the other usages.

marufsharifi (changes requested):

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,23 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
 
 function SearchRouterPage() {
     const {closeSearchRouter} = useSearchRouterActions();
     const {isSearchRouterDisplayed} = useSearchRouterState();
+    const {isKeyboardActive, keyboardHeight, keyboardActiveHeight} = useKeyboardState();
+    const {paddingTop} = useSafeAreaPaddings();
+    const {windowHeight} = useWindowDimensions();
+    const styles = useThemeStyles();
+    const effectiveKeyboardHeight = isKeyboardActive ? keyboardHeight || keyboardActiveHeight : 0;
+    const availableHeight = effectiveKeyboardHeight ? Math.max(windowHeight - effectiveKeyboardHeight - paddingTop, 0) : undefined;
```
marufsharifi: **Implementation doesn't match the PR description.** The description says this "adds offline-indicator bottom clearance … using the existing `useBottomSafeSafeAreaPaddingStyle` hook … only on narrow layouts while offline." But this code doesn't use that hook, has no `isOffline` check, and instead constrains height from the **keyboard**.

`useBottomSafeSafeAreaPaddingStyle` already does exactly what the description claims — it reads `isOffline` via `useNetwork` and adds `CONST.OFFLINE_INDICATOR_HEIGHT` only when offline. Please either reconcile the description with the code, or (preferred) replace this hand-rolled math with the existing hook applied to the list's content container:

```ts
const contentContainerStyle = useBottomSafeSafeAreaPaddingStyle({addOfflineIndicatorBottomSafeAreaPadding: true});
```

As written, a reviewer/QA tests the wrong thing and a future maintainer is misled about intent.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,23 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
+import {View} from 'react-native';
 
 function SearchRouterPage() {
     const {closeSearchRouter} = useSearchRouterActions();
     const {isSearchRouterDisplayed} = useSearchRouterState();
+    const {isKeyboardActive, keyboardHeight, keyboardActiveHeight} = useKeyboardState();
+    const {paddingTop} = useSafeAreaPaddings();
+    const {windowHeight} = useWindowDimensions();
+    const styles = useThemeStyles();
+    const effectiveKeyboardHeight = isKeyboardActive ? keyboardHeight || keyboardActiveHeight : 0;
```
marufsharifi: **`keyboardHeight` and `keyboardActiveHeight` are on different scales**, so the `||` fallback produces a layout hop.

In `withKeyboardState.tsx`, `keyboardHeight` is set on `keyboardDidShow` via `getKeyboardHeight(e.height, bottom)` (**inset-adjusted**), while `keyboardActiveHeight` is set on `keyboardWillShow` to the raw `e.height` (**not** inset-adjusted). During the show animation you use the raw value, then after it settles you switch to the adjusted one.

**Effect:** on a device with a home-indicator inset, `maxHeight` shifts by ~`bottom` px between the "will show" and "did show" frames — a visible jump on every focus. Pick one consistent source, or document why mixing is intentional.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
marufsharifi: **Not scoped to offline — this changes the online path too.** `availableHeight` is applied whenever `isKeyboardActive` is true, regardless of network state, so online native users now also get a `maxHeight` clamp on the Search Router. The linked issue #97182 is offline-only, and the description explicitly promises "Online … behavior remains unchanged."

Gate the clamp on `isOffline` (via `useNetwork`), or use `useBottomSafeSafeAreaPaddingStyle`, which is already offline-scoped. Otherwise we're taking unverified risk on the far more common online flow.

marufsharifi: **This may not actually clear the offline indicator (the stated root cause).** The formula subtracts only the **keyboard** height and top inset — it never reserves `CONST.OFFLINE_INDICATOR_HEIGHT` or the bottom safe-area inset (`includeSafeAreaPaddingBottom` is set on the `ScreenWrapper`). The indicator renders at the bottom of the `ScreenWrapper`, above the keyboard, so the bottom of the clamped list can still sit *behind* it.

**Failure scenario:** offline, keyboard open, a result set ending exactly at the clamped bottom → the last row is still covered by the sticky offline indicator, i.e. the original bug reproduces. Reserving `OFFLINE_INDICATOR_HEIGHT` (what `addOfflineIndicatorBottomSafeAreaPadding` does) is what actually fixes this.

marufsharifi: **Complex computation needs a "why" comment.** This subtracts `paddingTop` but not `paddingBottom`, and clamps with `Math.max(…, 0)` — non-obvious. Per the PR checklist ("comments were added to code that is not self explanatory"), add a short comment describing what region `availableHeight` represents and why only the top inset is subtracted (the `ScreenWrapper` already owns the bottom safe area via `includeSafeAreaPaddingBottom`).

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -16,11 +28,13 @@ function SearchRouterPage() {
             includePaddingTop
             includeSafeAreaPaddingBottom
         >
-            <SearchRouter
-                onRouterClose={closeSearchRouter}
-                shouldHideInputCaret={false}
-                isSearchRouterDisplayed={isSearchRouterDisplayed}
-            />
+            <View style={[styles.flex1, availableHeight !== undefined && {maxHeight: availableHeight}]}>
```
marufsharifi: **Use the existing `StyleUtils` helper instead of an inline style object.** The repo already has `getMaximumHeight(maxHeight: number): ViewStyle` (`src/styles/utils/index.ts:848`), and the PR checklist asks whether the style "can't be created with an existing StyleUtils function."

```suggestion
            <View style={[styles.flex1, availableHeight !== undefined && StyleUtils.getMaximumHeight(availableHeight)]}>
```

(add `const StyleUtils = useStyleUtils();` alongside the other hooks). This keeps styling consistent with the codebase and gives the branch a proper `ViewStyle` type.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,23 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
 
+import useKeyboardState from '@hooks/useKeyboardState';
+import useSafeAreaPaddings from '@hooks/useSafeAreaPaddings';
+import useThemeStyles from '@hooks/useThemeStyles';
+import useWindowDimensions from '@hooks/useWindowDimensions';
+
 import React from 'react';
+import {View} from 'react-native';
 
 function SearchRouterPage() {
```
marufsharifi, line 33: **No test coverage for the diff (0%).** The checklist marks "I added unit tests … to prevent regressions," but no test file is included. The layout math is hard to test as-is because it depends on `KeyboardStateContext` and native insets — which is itself an argument for reusing the already-tested `useBottomSafeSafeAreaPaddingStyle`. If the hand-rolled approach stays, extract the `availableHeight` calculation into a pure helper (inputs: `windowHeight`, `keyboardHeight`, `paddingTop`) and unit-test the clamp behavior, including the `Math.max(…, 0)` edge case.

marufsharifi (changes requested):

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

nabi-ebrahimi: addressed.

nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Thanks. I’ll keep this inline because a unit test would only cover simple arithmetic, not the native keyboard/layout race, which is covered by the focused manual tests.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -16,11 +37,13 @@ function SearchRouterPage() {
             includePaddingTop
             includeSafeAreaPaddingBottom
         >
-            <SearchRouter
-                onRouterClose={closeSearchRouter}
-                shouldHideInputCaret={false}
-                isSearchRouterDisplayed={isSearchRouterDisplayed}
-            />
+            <View style={[styles.flex1, availableHeight !== undefined && StyleUtils.getMaximumHeight(availableHeight)]}>
```
marufsharifi, line 64: 🟡 **Perf (low): fresh style allocation each render.**

```tsx
<View style={[styles.flex1, availableHeight !== undefined && StyleUtils.getMaximumHeight(availableHeight)]}>
```

This builds a new array literal and `getMaximumHeight` returns a new `{maxHeight}` object on every render. This is a single page component (not a list row) and React Compiler runs on native, so the cost is negligible — NAB. If you want it explicit:

```tsx
const maxHeightStyle = useMemo(
    () => [styles.flex1, availableHeight !== undefined && StyleUtils.getMaximumHeight(availableHeight)],
    [styles.flex1, StyleUtils, availableHeight],
);
// ...
<View style={maxHeightStyle}>
```

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,32 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
     const {closeSearchRouter} = useSearchRouterActions();
     const {isSearchRouterDisplayed} = useSearchRouterState();
+    const {isKeyboardActive, keyboardHeight, keyboardActiveHeight} = useKeyboardState();
+    const {isOffline} = useNetwork();
+    const {paddingTop} = useSafeAreaPaddings();
+    const {bottom} = useSafeAreaInsets();
+    const {windowHeight} = useWindowDimensions();
+    const StyleUtils = useStyleUtils();
+    const styles = useThemeStyles();
+    const effectiveKeyboardHeight = isOffline && isKeyboardActive ? keyboardHeight || getKeyboardHeight(keyboardActiveHeight, bottom) : 0;
```
marufsharifi: **NAB Readability:** This line combines a ternary condition, a `&&`, and a `||` fallback in one expression, so the reader has to hold three branches at once and confirm that `||` binds to the truthy branch rather than the whole condition:

```ts
const effectiveKeyboardHeight = isOffline && isKeyboardActive ? keyboardHeight || getKeyboardHeight(keyboardActiveHeight, bottom) : 0;
```

Suggest splitting into named steps so each line does one thing:

```ts
const isKeyboardOverlapping = isOffline && isKeyboardActive;
const measuredKeyboardHeight = keyboardHeight || getKeyboardHeight(keyboardActiveHeight, bottom);
const effectiveKeyboardHeight = isKeyboardOverlapping ? measuredKeyboardHeight : 0;
```

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,32 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
+    // Keep the router between the top safe area and keyboard. ScreenWrapper and the list handle bottom safe-area and offline-indicator spacing; clamp transient dimensions to zero.
```
marufsharifi: **[low] Readability:** Good "why" comment, but it's a run-on fusing three ideas (the goal, what other components own, and the clamp behavior) into one line, which loosens the tie between each clause and the code below it. Suggest splitting so the intent and the clamp rationale sit next to what they explain:

```ts
// Keep the router between the top safe area and the keyboard; ScreenWrapper and the
// list already handle bottom safe-area and offline-indicator spacing.
// Clamp to 0 so transient mid-animation dimensions never produce a negative height.
const availableHeight = effectiveKeyboardHeight ? Math.max(windowHeight - effectiveKeyboardHeight - paddingTop, 0) : undefined;
```

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,32 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
+    const availableHeight = effectiveKeyboardHeight ? Math.max(windowHeight - effectiveKeyboardHeight - paddingTop, 0) : undefined;
```
marufsharifi: ### 🔴 Missing test coverage — height computation (all 4 branches untested)

This ternary + clamp is the core of the fix and no test exercises it (`SearchRouterPage` isn't referenced anywhere under `tests/`). All four paths need coverage:

- **online / keyboard inactive** → `availableHeight === undefined`, no `getMaximumHeight` applied
- **offline + keyboard active + truthy `keyboardHeight`** → uses `keyboardHeight`
- **offline + keyboard active + falsy `keyboardHeight`** → falls back to `getKeyboardHeight(keyboardActiveHeight, bottom)`
- **clamp** → `Math.max(windowHeight - effectiveKeyboardHeight - paddingTop, 0)` returns `0` when the subtraction goes negative

Please add `tests/unit/SearchRouterPageTest.tsx`. Mock `useNetwork`, `useKeyboardState`, `useWindowDimensions`, `useSafeAreaInsets`/`useSafeAreaPaddings`, and spy on `StyleUtils.getMaximumHeight`. Assert:

- online → `getMaximumHeight` **not** called
- offline + active with `keyboardHeight = X` → called with `windowHeight - X - paddingTop`
- fallback path → uses `getKeyboardHeight(...)`
- small-`windowHeight` case → called with `0`

Cover both the happy (online, unchanged) and failure/offline paths.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
marufsharifi, line 64: ### 🟠 Missing test coverage — max-height application on the wrapper

The `availableHeight !== undefined && StyleUtils.getMaximumHeight(availableHeight)` guard has two untested outcomes: the wrapper gets a max-height style when offline + keyboard-active, and gets no max-height otherwise.

A render test should assert the `View`'s resolved `style` includes the max-height **only** in the offline + active case. This can reuse the `SearchRouterPageTest.tsx` from the comment above, asserting on the rendered wrapper style rather than (or in addition to) the spy.

src/components/Search/SearchAutocompleteList.tsx
```diff
@@ -185,6 +186,10 @@ function SearchAutocompleteList({
     const styles = useThemeStyles();
     const {translate, localeCompare} = useLocalize();
     const {shouldUseNarrowLayout} = useResponsiveLayout();
+    const contentContainerStyle = useBottomSafeSafeAreaPaddingStyle({
```
marufsharifi, line 190: ### 🟠 Missing test coverage — offline-aware `contentContainerStyle`

This is the PR's stated purpose ("keep results above the offline indicator") but neither branch of the hook is asserted. The existing tests (`tests/unit/SearchAutocompleteListTest.tsx`, `tests/ui/components/SearchAutocompleteListTest.tsx`) never touch offline/safe-area padding.

Please extend `tests/unit/SearchAutocompleteListTest.tsx` (it already renders the list and drives Onyx). Add two cases and assert on the rendered list's `contentContainerStyle`:

- **online** → `paddingBottom` stays `styles.pb2`
- **offline** (`ONYXKEYS.NETWORK = {isOffline: true}`) → `paddingBottom` increases by `CONST.OFFLINE_INDICATOR_HEIGHT`

> ⚠️ The offline branch only triggers when `ScreenWrapperOfflineIndicatorContext` provides `addSafeAreaPadding: true` — wire that provider into the test render tree or the offline case silently no-ops.

marufsharifi, line 190: **🟡 medium — Shared component: this `contentContainerStyle` change lands on every consumer, not just the native router**

This list is also rendered by [`SearchRouter.tsx`](https://github.com/Expensify/App/blob/main/src/components/Search/SearchRouter/SearchRouter.tsx) and `DeferredSearchAutocompleteList` (`index.native.tsx` + `index.tsx`), so `addOfflineIndicatorBottomSafeAreaPadding` applies on **all platforms/surfaces**, not only the native `SearchRouterPage` this PR targets. It's guarded by `isOfflineIndicatorSafeAreaPaddingEnabled` from `ScreenWrapperOfflineIndicatorContext`, so risk is low — but per the reviewer checklist we should confirm no other consumer gets doubled bottom padding or a layout shift when offline.

Could you verify offline in Search Router on **web/mWeb** (and any other `SearchAutocompleteList` usage) that the last row clears the offline indicator with no extra gap? No code change expected if it looks right.

marufsharifi (changes requested):

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Thanks for checking this. Since this is a single page wrapper rather than a repeated list item, the allocation is negligible and React Compiler can optimize it where needed. I’ll keep the simpler inline style.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: The height calculation and its online, offline, fallback, and clamp branches are now covered by focused unit tests. The resulting value is passed directly to the existing `StyleUtils.getMaximumHeight()` helper, while the actual native wrapper behavior is covered by the reported manual regression test. I’d avoid adding a heavily mocked render test for this direct mapping.

src/components/Search/SearchAutocompleteList.tsx
nabi-ebrahimi: Addressed.

nabi-ebrahimi: Verified offline Search Router on web and mWeb. The last row clears the offline indicator without doubled padding or an extra layout gap, and online spacing remains unchanged.

tests/unit/SearchAutocompleteListTest.tsx
```diff
@@ -31,6 +35,25 @@ import * as TestHelper from '../utils/TestHelper';
 import waitForBatchedUpdates from '../utils/waitForBatchedUpdates';
 import wrapOnyxWithWaitForBatchedUpdates from '../utils/wrapOnyxWithWaitForBatchedUpdates';
 
+const mockFlashListContentContainerStyles: Array<React.ComponentProps<typeof ReactNative.FlatList>['contentContainerStyle']> = [];
```
marufsharifi, line 39: **NAB** Deriving prop types via `ComponentProps<typeof FlatList>`

Our style guide prefers importing an exported prop type over `ComponentProps<typeof X>`. That said, this targets `react-native`'s `FlatList`/`FlashList` (third-party, no App-exported prop type to import) and it's a test mock, so this is effectively not actionable. Noting only for completeness — safe to leave as-is.

tests/unit/SearchAutocompleteListTest.tsx
```diff
@@ -154,14 +179,18 @@ const fakeRecentReports = [
     {reportID: '103', keyForList: '103', text: 'Charlie Report', alternateText: 'charlie alt', lastMessageText: 'hey'},
 ];
 
-function SearchRouterWrapper({isSearchRouterDisplayed}: {isSearchRouterDisplayed?: boolean}) {
+function SearchRouterWrapper({isSearchRouterDisplayed, addOfflineIndicatorSafeAreaPadding}: {isSearchRouterDisplayed?: boolean; addOfflineIndicatorSafeAreaPadding?: boolean}) {
```
marufsharifi, line 190: **Boolean prop `addOfflineIndicatorSafeAreaPadding` isn't `is`/`should`-prefixed**

Per STYLE.md, boolean props/vars should be prefixed with `is` or `should` (e.g. `shouldAddOfflineIndicatorSafeAreaPadding`). However, this deliberately mirrors existing codebase naming (`addSafeAreaPadding` context value, `addOfflineIndicatorBottomSafeAreaPadding` hook param) and is a test-only wrapper prop, so consistency with the surrounding names arguably wins. Fine to leave.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,56 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
     const {closeSearchRouter} = useSearchRouterActions();
     const {isSearchRouterDisplayed} = useSearchRouterState();
+    const {isKeyboardActive, keyboardHeight, keyboardActiveHeight} = useKeyboardState();
+    const {isOffline} = useNetwork();
+    const {paddingTop} = useSafeAreaPaddings();
+    const {bottom} = useSafeAreaInsets();
+    const {windowHeight} = useWindowDimensions();
+    const StyleUtils = useStyleUtils();
+    const styles = useThemeStyles();
+    const isKeyboardOverlapping = isOffline && isKeyboardActive;
```
marufsharifi: **Medium** `isKeyboardOverlapping` is assigned `isOffline && isKeyboardActive`, but the name says nothing about being offline. A reader tracing why the height clamps sees "overlapping" and reasonably assumes it's a pure keyboard-geometry check, then is surprised the whole branch is gated on network state. The offline coupling is the non-obvious part and it's invisible in the name.

Suggest renaming to reflect the real condition, e.g. `shouldConstrainForOfflineKeyboard` — or keeping the name and adding a note:

```ts
// Only offline, where the offline indicator occupies the bottom.
const isKeyboardOverlapping = isOffline && isKeyboardActive;
```

tests/unit/SearchAutocompleteListTest.tsx
```diff
@@ -246,6 +276,37 @@ describe('SearchAutocompleteList', () => {
...
+        await flushAllUpdates();
+
+        const getContentPaddingBottom = () => {
+            return mockFlashListContentContainerStyles
+                .map((contentContainerStyle) => StyleSheet.flatten(contentContainerStyle)?.paddingBottom)
+                .findLast((paddingBottom) => typeof paddingBottom === 'number');
+        };
+        const onlinePaddingBottom = getContentPaddingBottom();
+
+        expect(typeof onlinePaddingBottom).toBe('number');
```
marufsharifi: **NAB:** the type is asserted twice — `expect(typeof onlinePaddingBottom).toBe('number')` immediately followed by `if (typeof onlinePaddingBottom !== 'number') { throw ... }`. The intent (assert + narrow for TS) is fine, but the doubled check reads as redundant. Consider dropping the `expect` and letting the `throw` both fail the test and narrow the type, or add a short comment:

```ts
// throw (not just expect) so TS narrows the type below
```

tests/unit/SearchAutocompleteListTest.tsx
```diff
@@ -246,6 +276,37 @@ describe('SearchAutocompleteList', () => {
...
+                action: jest.fn(),
+            },
+        ]);
+        mockFlashListContentContainerStyles.length = 0;
+        render(<SearchRouterWrapper addOfflineIndicatorSafeAreaPadding />);
+        await flushAllUpdates();
+
+        const getContentPaddingBottom = () => {
+            return mockFlashListContentContainerStyles
+                .map((contentContainerStyle) => StyleSheet.flatten(contentContainerStyle)?.paddingBottom)
```
marufsharifi, line 303: **NAB:** `getContentPaddingBottom` chains `.map(...).findLast(...)` to pull "the last numeric paddingBottom", but the *why* — that FlashList re-renders push multiple style snapshots and only the latest matters — isn't stated, so `findLast` looks arbitrary. One line would fix it:

```ts
// FlashList records a style per render; take the most recent numeric paddingBottom
```

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
```diff
@@ -2,11 +2,56 @@ import ScreenWrapper from '@components/ScreenWrapper';
...
+type GetAvailableHeightParams = {
+    isKeyboardOverlapping: boolean;
+    keyboardHeight: number;
+    keyboardActiveHeight: number;
+    bottomInset: number;
+    windowHeight: number;
+    paddingTop: number;
+};
+
+function getAvailableHeight({isKeyboardOverlapping, keyboardHeight, keyboardActiveHeight, bottomInset, windowHeight, paddingTop}: GetAvailableHeightParams): number | undefined {
```
marufsharifi, line 27: **NAB:** This `windowHeight - keyboardHeight - paddingTop` computation is the same shape as the one in `BaseModal.tsx` (`windowHeight - modalHeight - keyboardActiveHeight - paddingTop`). Not identical, and there's no existing shared util, so nothing to change here — just flagging in case a shared `getAvailableHeightAboveKeyboard` helper is worth extracting in a follow-up.

marufsharifi (changes requested):

tests/unit/SearchAutocompleteListTest.tsx
nabi-ebrahimi: Thanks, I’ll keep this as-is since it is limited to the third-party test mock.

src/components/Search/SearchRouter/SearchRouterPage/index.native.tsx
nabi-ebrahimi: Addressed.

tests/unit/SearchAutocompleteListTest.tsx
nabi-ebrahimi: Addressed.

tests/unit/SearchAutocompleteListTest.tsx
nabi-ebrahimi: Addressed.

marufsharifi (approved):
LGTM

carlosmiceli (approved):

Issue #97104, opened 2026-07-27
[Due for payment 2026-08-19] [Sentry: APP-7MV] Web WebGL context creation failure on /search and /home (resurfaced on 9.4.43+)


mountiny:
## Sentry
https://expensify.sentry.io/issues/APP-7MV

## Impact (snapshot at filing)
- Users (total since first seen): 560
- Events: 8950
- Users (last 7d): 10
- First seen: 2026-03-15
- Last seen: ongoing (2026-07-26)
- Platform: web
- App version(s): 9.4.43-1, 9.4.41-4 (resurfaced on latest production)
- Affected route(s): /search, /home (previously /home only in prior tracker)
- Mechanism: auto.browser.browserapierrors.setTimeout

## Stack trace (top frames)
```
failed to create webgl context: err 0
Error: failed to create webgl context: err 0
@sentry/browser helpers → captureException
Search_Root view
```

## Suspected cause
WebGL context creation fails on low-memory or GPU-constrained browsers while Search or Home renders WebGL-dependent UI. Crash resurfaced on 9.4.43+ after prior fix was paid/closed.

## Reproduction
Unknown — see Sentry events. Prior tracker focused on /home; new events also on /search.

## Related
- Sibling Sentry IDs: none
- Prior GH issues: #93834 (paid/closed 2026-07-22 — same Sentry ID, /home-only scope)

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: <a href="https://www.upwork.com/jobs/~022081747685653182354">https://www.upwork.com/jobs/~022081747685653182354</a></li>
        <li>Upwork Job ID: 2081747685653182354</li>
        <li>Last Price Increase: 2026-07-27</li>
    </ul>
</details>

abzokhattab:
Reviewed Melvin's proposal .... root cause is correct, but the fix isn't needed. **This is already fixed in production.**

[...]

PR #97219, opened 2026-07-28
Render Skia web charts on a software surface when WebGL is unavailable

wildan-m:
### Explanation of Change

`failed to create webgl context: err 0` (Sentry APP-7MV) is still reported on production after the prior fix, now on `/search` as well as `/home`. That fix hardened the capability check that gates the web charts, and it does prevent the crash when the device is already incapable at the moment the chart mounts. But the check runs when the chart mounts, while the WebGL context is created much later — after the CanvasKit WASM module finishes loading. Anything that changes GPU availability in between (memory pressure, a GPU process restart, or the per-page context limit being exhausted) makes that verdict stale, and the check cannot see it: it validates a context it already holds, rather than the browser's ability to hand out a new one.

So the crash cannot be fully prevented by predicting it up front. This makes the failure non-fatal where it actually happens instead: CanvasKit throws rather than degrading when it can't create a WebGL surface, and that throw runs in a resize callback, so no `try/catch` or error boundary can reach it. The patch catches it and falls back to CanvasKit's software surface, so the chart still renders — mirroring the library's other web renderer path, which already treats a failed WebGL surface as recoverable. The software fallback is skipped for a canvas that already holds a WebGL context (it can never provide the 2D context the software surface flushes through, which would turn into another uncatchable error).

When neither surface can be created, the renderer announces it with a bubbling DOM event — a no-op without a listener — and the chart wrapper listens for it and swaps in the existing "Unable to display chart" empty state. The capability check cannot cover that window, so the renderer's own report is the only reliable signal. Charts on capable clients are unchanged and still render on the GPU.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97104

### Tests

1. Go to Search → **Spend** → Insights and open **Spend over time**, **Top categories**, and **Top merchants**. Verify each chart renders normally (GPU path, unchanged).
2. Reload, and before opening a chart, simulate a client that can create a context for the capability check but not for the actual chart — paste this in the console:
   ```js
   // 1st webgl2 request (the capability check) succeeds; the ones CanvasKit makes later fail.
   window.__gl = 0;
   const orig = HTMLCanvasElement.prototype.getContext;
   HTMLCanvasElement.prototype.getContext = function (type, ...args) {
       if (type === 'webgl2' && ++window.__gl > 1) return null;
       return orig.call(this, type, ...args);
   };
   // The chart may render through the offscreen renderer instead, so deny that context too.
   if (typeof OffscreenCanvas !== 'undefined') {
       const offscreen = OffscreenCanvas.prototype.getContext;
       OffscreenCanvas.prototype.getContext = function (type, ...args) {
           if (type === 'webgl2') return null;
           return offscreen.call(this, type, ...args);
       };
   }
   ```
3. **Do not reload** — navigate to an insight chart from the left sidebar so the chart mounts with the override active.
4. Verify the page does **not** crash — no `failed to create webgl context: err 0` (or any other uncaught error) in the console — and the chart area shows one of the two degraded outcomes instead of staying silently blank:
   - the chart **renders through the software surface** (the first paint takes a few seconds longer than the GPU path), or
   - when no surface can be created at all, the **"Unable to display chart"** empty state appears.

   Before this change the error was thrown and the chart area stayed blank. (The error is thrown as a plain string rather than an `Error`, so it has no stack trace — which is why it reaches Sentry as a bare title with no first-party frames.)
5. Reload to clear the override and reopen a chart; verify it renders normally again on the GPU path.

**Platform scope:** step 2 injects through the browser console, so it applies to **MacOS: Chrome/Safari**. On other platforms just confirm no regression — the three insight charts still render normally. iOS/Android **native** are unaffected: the patched file is the web renderer, which native never loads.

### Offline tests

No offline impact — the surface creation and fallback behave identically online and offline.

### QA Steps

The crash condition is environmental and can't be reliably forced on staging, so just confirm there's no regression around the charts:

1. Go to Search → **Spend** → Insights and open **Spend over time**, **Top categories**, and **Top merchants**.
2. Verify each chart renders normally, with no errors in the console.

The definitive result is only observable in production: once this deploys, the [`APP-7MV`](https://expensify.sentry.io/issues/APP-7MV) Sentry issue should stop receiving new `failed to create webgl context` events.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

https://github.com/user-attachments/assets/e37f4a11-df7b-43e8-83be-86d40add93cb


truph01 (approved):

patches/@shopify/react-native-skia/details.md
```diff
@@ -23,3 +23,48 @@
...
+
+    When no surface can be created at all, the renderer dispatches a bubbling
+    "skia-surface-unavailable" CustomEvent on its canvas (a no-op without a listener).
+    A capability check cannot cover this case - it runs at chart mount while the
+    context is created only after the CanvasKit WASM module loads - so the event is
+    the only reliable signal, and SkiaWebChart uses it to swap in its "unable to
+    display chart" empty state instead of leaving a blank canvas.
+    ```
+
+- Upstream PR/issue: 🛑 TODO — the same defensive handling should be sent upstream so `onResize` matches the already-guarded `renderPictureToSurface` path.
```
mountiny: Lets' do this now

mountiny (changes requested):

patches/@shopify/react-native-skia/details.md
wildan-m: Done — upstream PR: https://github.com/Shopify/react-native-skia/pull/3996. It applies the same defensive handling (plus the `skia-surface-unavailable` event) to upstream `main`, where the throws moved into the renderer constructor and `onResize` after the WebGL-context-reuse refactor. `details.md` now links it; once it lands in a release we consume, the patch can be dropped.

mountiny (approved):
Thanks for adding that

src/components/Charts/SkiaWebChart/index.tsx
```diff
@@ -71,12 +79,14 @@ function SkiaWebChart<TProps extends object>({getComponent, componentProps}: Ski
     );
 
     return (
-        <WithSkiaWeb
-            opts={{locateFile: (file: string) => `/${file}`}}
-            getComponent={getComponent}
-            componentProps={componentProps}
-            fallback={fallback}
-        />
+        <View ref={viewRef(containerRef)}>
```
situchan, line 82: This wrapper caused regression - [Concierge chat is cut off on medium/small screens](https://github.com/Expensify/App/issues/98456#top)

Root cause: `maxWidth` isn't set

PR #97228, opened 2026-07-28
Add SequentialQueue pause watchdog to self-heal orphaned pauses

TMisiukiewicz:
### Explanation of Change

When the client detects an Onyx-update gap, the SequentialQueue is paused (via `shouldPauseQueue` in SaveResponseInOnyx middleware, or `SequentialQueue.pause()` in applyOnyxUpdatesReliably/DeferredOnyxUpdates) and is only unpaused when the OnyxUpdateManager finishes resolving the gap. That pause/unpause pair is not balanced by construction: several early-return and failure paths in the async chain (e.g. non-leader tab early return in `handleMissingOnyxUpdates`, subscription not re-firing, a dead in-progress fetch promise) can leave the queue paused forever. When that happens, queued write commands like `OpenApp` never execute, `IS_LOADING_APP` never clears, and the user is stuck on a full-screen skeleton on any page until a hard page refresh. Reads also park forever on `waitForIdle()`. This was reproduced deterministically: switching priority mode from #focus to normal left `OpenApp` and `PusherPing` stuck in the queue with `[SequentialQueue] Unable to flush. Queue is paused.` repeating indefinitely and no unpause ever logged. It also matches user reports of intermittent "infinite skeleton" sessions lasting 10+ minutes (fixed by switching browsers/clearing cookies, i.e. resetting client state).

This PR adds a pause watchdog inside `SequentialQueue` (it owns `isQueuePaused`):

- `pause()` arms a timeout (`CONST.NETWORK.MAX_PAUSE_WATCHDOG_TIME_MS`, 60s); `unpause()` and `resetQueue()` clear it.
- A `connectWithoutView` subscription to `ONYX_UPDATES_LAST_UPDATE_ID_APPLIED_TO_CLIENT` re-arms the watchdog on every applied-update advance while paused — so a genuinely progressing (even slow) catch-up never trips it; only a pause with zero progress for the full window does.
- When it fires, it logs a `Log.alert` (visible in telemetry to find residual causes) and force-calls `unpause()`, which flushes the queue and re-runs the stuck `OpenApp`/`ReconnectApp`, re-syncing the client and clearing the skeleton.

This is a liveness safety net, not a root-cause fix for each individual orphaning path — it converts every "stuck forever until refresh" variant into a bounded ≤60s stall that self-heals.

Changed files:

- `src/CONST/index.ts` — new `NETWORK.MAX_PAUSE_WATCHDOG_TIME_MS` constant
- `src/libs/Network/SequentialQueue.ts` — watchdog implementation

### Fixed Issues

$ https://github.com/Expensify/App/issues/97331
PROPOSAL:

### Tests

1. Open account with a lot of data in Onyx in focus mode
2. If you were in most recent mode after logging in, go to Preferences -> change to focus, and then go to Troubleshoot -> Clear Onyx data
3. Go to Preferences -> switch to Most recent mode
4. In the Network tab, verify the `OpenApp` request was called maximum 60s after switching the mode
5. Once `OpenApp` resolved, verify the app works and there are no infinite skeletons

### Offline tests

N/A

### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


contributingGuides/SEQUENTIAL_QUEUE.md
```diff
@@ -389,9 +390,12 @@ Three concerns cut across every block.
...
...UPDATES_LAST_UPDATE_ID_APPLIED_TO_CLIENT` re-arms it, so a slow-but-progressing catch-up never trips it; the timer elapses only when the paused queue has applied **no** newer update for the full window. When it fires it `Log.alert`s (so residual orphaning causes stay visible in telemetry) and recovers in two steps: **close the gap first, then unpause**. The gap-closing escalation — registered by `OnyxUpdateManager` via `registerPauseWatchdogEscalation` (`SequentialQueue` cannot import it, dependency cycle) — fires an incremental `ReconnectApp` **outside the queue** (`reconnectAppWithSideEffects`), whose response applies immediately and moves the client to the server head, so the drained writes don't run against stale data. The escalation shares the stalled-fetch back-off (a client already thrashing on `GetMissingOnyxMessages` is not handed another reconnect; the watchdog then just unpauses) and is capped at `MAX_PAUSE_WATCHDOG_ESCALATION_TIME_MS`, with failures swallowed — liveness wins over ordering when the recovery itself misbehaves. After the escalation settles (or the cap elapses), `unpause()` runs and `flush()` drains the queue. `unpause()` and `resetQueue()` clear the timer.
```
adhorodyski: briefly speaking, I find this to be pure jargon. can you frame it so that a human being understands it (more) easily?

TMisiukiewicz: agree, done

src/libs/Network/SequentialQueue.ts
```diff
@@ -74,8 +74,65 @@ function setIsReadyPromisePending() {
...
+            // The normal chain may have unpaused meanwhile.
+            if (!isQueuePaused) {
+                return;
+            }
+            unpause();
+        });
+    }, CONST.NETWORK.MAX_PAUSE_WATCHDOG_TIME_MS);
+}
+
+// Progress while paused re-arms the watchdog, so only a stalled pause trips it.
```
mountiny: Can you add a clear comment for why this needs to use connectWithoutView so its not repeated elsewhere

src/libs/actions/OnyxUpdateManager/index.ts
```diff
@@ -116,6 +116,24 @@ function isFetchAlreadyStalled(lastUpdateIDFromClient: number): boolean {
...
+    if (stalledFetch && Date.now() - stalledFetch.time < CONST.NETWORK.STALLED_UPDATES_FETCH_BACKOFF_TIME_MS) {
+        Log.info('[OnyxUpdateManager] Pause watchdog escalation skipped — within the stalled-fetch back-off window', false, {lastUpdateIDFromClient});
+        return Promise.resolve();
+    }
+
+    stalledFetch = {clientUpdateID: lastUpdateIDFromClient, time: Date.now()};
+    Log.info('[OnyxUpdateManager] Pause watchdog escalation — firing an incremental ReconnectApp outside the queue to close the update gap before unpausing', false, {
+        lastUpdateIDFromClient,
+    });
+    return reconnectAppWithSideEffects(lastUpdateIDFromClient);
```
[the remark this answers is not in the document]
mountiny: @TMisiukiewicz can you address?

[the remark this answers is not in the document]
TMisiukiewicz: addressed ✅

contributingGuides/SEQUENTIAL_QUEUE.md
```diff
@@ -335,7 +336,7 @@ In the drain-end `process().finally`, `resolveIsReadyPromise` runs only when `is
 
 `unpause()` calls `flush(false)` precisely to **preserve** the existing `isReadyPromise`. Swapping in a fresh one would orphan READs already awaiting the old promise. (`setIsReadyPromisePending()` is idempotent for the same reason: it no-ops when a pending promise is already armed.)
 
-**Sharp edges.** `isReadyPromise` is a single shared one-shot resolver. If a flush round arms it but never reaches any of the four resolve sites (e.g. a persistent, unresolvable `shouldPauseQueue` data gap), subsequent READs blocked on `waitForIdle()` have no timeout.
+**Sharp edges.** `isReadyPromise` is a single shared one-shot resolver with no timeout of its own. If a flush round arms it but never reaches any of the four resolve sites (e.g. a persistent, unresolvable `shouldPauseQueue` data gap), subsequent READs blocked on `waitForIdle()` hang — bounded today only by the pause watchdog, which force-unpauses after `MAX_PAUSE_WATCHDOG_TIME_MS` without applied-update progress so the drain can reach a resolve site (see [Pausing & Data-Gap Sync](#pausing--data-gap-sync)).
```
adhorodyski: > bounded today only by the pause watchdog

i wouldn't use any of that time framing, we're simply describing code as is

contributingGuides/SEQUENTIAL_QUEUE.md
```diff
@@ -389,9 +390,14 @@ Three concerns cut across every block.
...
+
+So `pause()` starts a 60 s timer (`MAX_PAUSE_WATCHDOG_TIME_MS`) that resets whenever the client applies a newer update — a slow-but-healthy catch-up keeps pushing the deadline out, so it only fires when the queue has made **no** progress for a full minute. When it fires, it logs an alert (to keep the root cause visible in telemetry), then **closes the gap before unpausing**: it asks `OnyxUpdateManager` (via a registered callback, since `SequentialQueue` cannot import it directly) to fetch the missing updates with an incremental `ReconnectApp` that runs outside the queue, so the drained writes are not sent against stale data. That recovery is best-effort — it is skipped if the client is already retrying fetches, capped at `MAX_PAUSE_WATCHDOG_ESCALATION_TIME_MS`, and its errors are ignored — because unblocking the queue always wins over closing the gap. `unpause()` and `resetQueue()` cancel the timer.
```
adhorodyski: > (via a registered callback, since `SequentialQueue` cannot import it directly)

this is not essential I think

abzokhattab (approved):
LGTM

mountiny (commented):
Reviewed the diff plus the surrounding machinery it leans on (`SaveResponseInOnyx`, `OnyxUpdates.apply`, `reconnectApp`, `ActiveClientManager`, `Log`, `makeRequestWithSideEffects`).

The design is sound. The watchdog is correctly owned by the module that owns `isQueuePaused`, the registration callback is the right way around the `SequentialQueue → App` cycle (with a safe `?? Promise.resolve()` fallback if nothing registers), and the escalation being both capped and failure-swallowed means the recovery can't itself re-deadlock. The generation token plus the `isQueuePaused` re-check before the late `unpause()` is the right guard. It also fits the existing machinery nicely — `SIDE_EFFECT_REQUEST_COMMANDS.RECONNECT_APP` is already in `requestsToIgnoreLastUpdateID`, so the escalation's response applies straight through rather than re-pausing.

Below are the gaps I'd want closed. Numbered by risk.

---

## 1. The escalation can fire a *full* out-of-queue ReconnectApp — medium

[`OnyxUpdateManager/index.ts#L123`](https://github.com/Expensify/App/blob/ebf5441699a25b108bec1a0b28c3eed0fcc95a9c/src/libs/actions/OnyxUpdateManager/index.ts#L123) falls back to `CONST.DEFAULT_NUMBER_ID` (0) when the client has never received a `lastUpdateID`. `reconnectAppWithSideEffects(0)` then skips `params.updateIDFrom`, so the server returns the **full** app payload — applied outside the queue while WRITEs are still pending in the paused queue.

`handleMissingOnyxUpdates` has a dedicated branch for exactly this case (`!lastUpdateIDFromClient` → `finalReconnectAppAfterActivatingReliableUpdates`, guarded by "if a reconnect query is in progress, don't start another one"). The watchdog escalation has no equivalent guard.

Suggest bailing out of the escalation when `lastUpdateIDFromClient === 0` and just unpausing.

**Related nit** in [`App.ts#L565`](https://github.com/Expensify/App/blob/ebf5441699a25b108bec1a0b28c3eed0fcc95a9c/src/libs/actions/App.ts#L565): `getOnyxDataForOpenOrReconnect(false, !updateIDFrom)` — the 2nd positional is `isFullReconnect`, which only changes a log label. The JSDoc immediately above discusses `IS_LOADING_APP`, which is controlled by the 5th param (`shouldClearAppLoading`) and isn't passed at all. The behaviour is correct, but the pairing reads as though `!updateIDFrom` is what keeps `IS_LOADING_APP` from being cleared.

## 2. The watchdog fires while offline — medium

`isQueuePaused` is orthogonal to offline. Pause on a gap → user goes offline → nothing advances → 60s later the watchdog trips:

- a `Log.alert` fires that is a false positive (the pause isn't orphaned, the client is just offline) — noise in the exact telemetry signal this PR relies on to find residual causes;
- a ReconnectApp that cannot succeed, followed by the full 15s cap wait;
- a force-unpause, so once connectivity returns the queue drains against a client that is *known* to be behind — precisely the state the pause exists to prevent.

Suggest not arming (or deferring) while `isOfflineNetwork()`, and re-arming on the online transition. The same reasoning covers backgrounded native apps, where suspended JS timers make the watchdog fire immediately on foreground rather than after a real 60s of no progress.

## 3. The re-arm signal isn't causally tied to *this* pause's resolution — design gap

An advance of `ONYX_UPDATES_LAST_UPDATE_ID_APPLIED_TO_CLIENT` means "some update was applied", not "the gap chain that paused us is still alive". Commands in `requestsToIgnoreLastUpdateID` (OpenApp, ReconnectApp, GetMissingOnyxMessages, DeleteMoneyRequest, CloseAccount) route straight to `OnyxUpdates.apply`, which advances that key ([`OnyxUpdates.ts#L194`](https://github.com/Expensify/App/blob/ebf5441699a25b108bec1a0b28c3eed0fcc95a9c/src/libs/actions/OnyxUpdates.ts#L194)) **even with the gap open**.

So a leader tab whose unpause is orphaned, but which still receives responses that advance the key, keeps pushing its deadline out indefinitely — the same "stuck forever" class this PR is defending against. The leader gate closed the cross-tab version of this (good catch in the earlier review); the intra-tab version remains.

Cheap belt-and-braces: an **absolute ceiling** measured from `pause()` that re-arming cannot extend (say 5 min), so total pause time is genuinely bounded. As written, the only bound is "no signal for 60s", which isn't a bound at all.

## 4. `reconnectAppWithSideEffects` doesn't await `hasLoadedAppPromise` — low/medium

[`App.ts#L550`](https://github.com/Expensify/App/blob/ebf5441699a25b108bec1a0b28c3eed0fcc95a9c/src/libs/actions/App.ts#L550) reads `hasLoadedApp` synchronously. `reconnectApp` deliberately wraps its body in `hasLoadedAppPromise.then(...)` because `hasLoadedApp` is `undefined` until Onyx hydrates — an un-hydrated read falls through to a full `openApp()`. The comment says "Mirror `reconnectApp`'s guards", but this is the one guard that changes behaviour. Low probability (60s after a pause, hydration is done), cheap to fix.

The `openApp()` fallback itself is fine — `writeWithNoDuplicatesConflictAction` drops a duplicate OpenApp, and OpenApp is in `requestsToIgnoreLastUpdateID`.

## 5. `resetQueue()` resets `pauseGeneration = 0`, recycling the token — low (test flakiness)

`resetQueue` is test-only, but it runs in `afterEach` while an escalation may still be pending: escalation captures generation 1 → `resetQueue()` → next test's `pause()` → generation 1 again → the `pauseGeneration !== generation` guard passes and the stale escalation unpauses the *new* test's pause. That's exactly the scenario the new "a stale escalation must not unpause a pause that started after it fired" test pins *within* a test, reopened *across* tests.

Make `pauseGeneration` monotonic (don't reset it) — a token that can be recycled carries no information. Same applies to `lastSeenUpdateID = 0`.

## 6. `Onyx.clear()` reads as progress — low

The callback coerces `undefined → 0` into `lastSeenUpdateID`, so Troubleshoot → "Clear Onyx data" (literally step 2 of this PR's own test steps) resets it, and the next real value then counts as an advance and re-arms for another full 60s. Still bounded, but it silently doubles the worst-case stall in the very flow being tested. Prefer ignoring the clear/`undefined` case rather than coercing it to 0.

## 7. `escalationCap` timer is never cleared — nit

[`SequentialQueue.ts#L116`](https://github.com/Expensify/App/blob/ebf5441699a25b108bec1a0b28c3eed0fcc95a9c/src/libs/Network/SequentialQueue.ts#L116) — when the escalation wins the race, a 15s `setTimeout` stays pending. Harmless in production; in Jest it's a dangling timer that can leak across tests.

---

## Test gaps

- **Nothing asserts the queue actually drains after a force-unpause.** Every watchdog test asserts only `isPaused() === false`. The entire point of the PR is that the stuck `OpenApp`/`ReconnectApp` *fires* and `IS_LOADING_APP` clears. Worth adding: push a request → `pause()` → let the watchdog fire → assert the request was sent and `IS_LOADING_APP` is `false`. Without that, a future change could keep all these tests green while the skeleton bug returns.
- No offline test (finding 2).
- No test for `lastUpdateIDFromClient === 0` (finding 1).
- `Log.alert` is described in `SEQUENTIAL_QUEUE.md` as the primary telemetry signal for residual causes, but isn't asserted anywhere.
- The PR body has every checklist box checked, including "I included screenshots or videos for tests on all platforms", while all five `<details>` blocks are empty.

## Checked and cleared

- The new file-wide `jest.mock('@libs/ActiveClientManager')` in `SequentialQueueTest.ts` covers all three real exports (`init`, `isClientTheLeader`, `isReady`) and returns `true`, matching `index.native.ts`'s hardcoded value — no behaviour change for the pre-existing tests in that file.
- `Log.alert` goes through `post` from `@libs/Network` rather than the sequential queue, so the alert isn't itself swallowed by the pause it's reporting.

mountiny (approved):
Thanks for the updates

PR #97242, opened 2026-07-28
Share extension: pre-mount destination report and create expense before reveal transition

jmgraa:
### Explanation of Change
Before this change, submitting a shared receipt could briefly show the Inbox behind the share modal before the destination chat appeared on top - a visible flash during dismiss.

`SubmitDetailsPage` now handles two navigation strategies based on layout and report state:

**Narrow layout:** Pre-mounts the destination report behind the modal (when it already exists in Onyx) so the correct chat is ready when the modal closes.

**Wide layout (tablets/desktop):** Reveals the destination report on top of the modal first, then dismisses the modal after the animation completes.

Both strategies keep the confirm button in loading state until navigation is safe. When the destination report doesn't exist yet (new recipient), the page waits for the optimistic report to land in Onyx before navigating - with fallback timers (500ms if ID changes, 5s absolute timeout) to prevent indefinite hangs.

The expense is created immediately on submit (not after the reveal animation), and cleanup skips a second navigation (shouldNavigate: false) since reveal/pre-mount already handles routing. Stale tracking spans are cancelled before creation to avoid lifecycle conflicts, improving telemetry accuracy.

### Fixed Issues
$ https://github.com/Expensify/App/issues/97667
PROPOSAL:

### Tests
1. **Share extension with existing recipient**
   - Share a receipt/image to Expensify via share extension
   - Select an existing contact as recipient
   - Verify confirm button shows loading state briefly, then navigates to the report

2. **Share extension with new recipient**
   - Share a receipt/image via share extension
   - Select a new recipient (first-time conversation)
   - Verify destination report doesn't exist yet (pre-mount disabled)
   - Verify confirm button stays in loading state until optimistic report lands from backend
   - Verify navigation completes to newly created report within 5 seconds

3. **Share extension with destination already topmost**
   - Open a report, then trigger share extension for the same recipient
   - Verify standard navigation flow works correctly

Repeat tests 1-3 both on narrow and wide layout

### Offline tests
Same as tests

### QA Steps
Same as tests

### PR Author Checklist
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/a52f11b8-a329-48ad-8e20-fc1e43ac3a0c

</details>


src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
...
+    const hasNavigationDestination = !!transaction && !!postSubmitNavigationReportID;
+    const destinationReportInCollection = hasNavigationDestination ? getReportOrDraftReport(postSubmitNavigationReportID, undefined, undefined, {}) : undefined;
+    // Destination report isn't in COLLECTION.REPORT yet (e.g. a recipient with no existing chat) — it will only
+    // exist after the expense create writes it optimistically, so pre-mounting is impossible.
+    const isDestinationReportMissing = hasNavigationDestination && !destinationReportInCollection?.reportID;
+    // Keep pre-mount off while pending navigation owns the reveal — otherwise once the optimistic
+    // report lands, isDestinationReportMissing flips false and usePreMountDestination would schedule
+    // a narrow pre-insert alongside the pending revealRouteBeforeDismissingModal (dual nav).
+    const preMountDestinationRoute =
+        hasNavigationDestination && !isDestinationReportMissing && !pendingNavigationReportID && Navigation.getTopmostReportId() !== postSubmitNavigationReportID
```
JakubKorytko: missing the "already pre-inserted" guard from `getSubmitExpensePreMountDestinationRoute` (`hasPreInsertedFullscreen`), and I think that breaks the feature.

Once we pre-insert, `getTopmostReportId()` starts returning `postSubmitNavigationReportID` (it looks past the RHP), so this flips to `undefined` and the `[route]` cleanup in `usePreMountDestination` removes the pre-insert we just made. Submit then falls back to `runExpenseCreateAndCleanup(true)`, which is the flash we're fixing. It can also re-arm afterwards and mount ReportScreen twice.

`getSubmitExpensePreMountDestinationRoute.ts:48-51` documents this exact trap. Same fix here:

```ts
const isAlreadyPreInserted = Navigation.getIsFullscreenPreInsertedUnderRHP();
...
&& (isAlreadyPreInserted || Navigation.getTopmostReportId() !== postSubmitNavigationReportID)
```

Whether it hits on a given render depends on which dep the compiler keyed the memo on, which is probably why the tests miss it.

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
...
+            ? ROUTES.REPORT_WITH_ID.getRoute(postSubmitNavigationReportID)
+            : undefined;
```
JakubKorytko: this only gates on `getTopmostReportId`. `getSubmitExpensePreMountDestinationRoute` also checks `isSearchTopmostFullScreenRoute`, `isReportTopmostSplitNavigator` and `isReportOpenInRHP`, and bails on wide layout.

Without those we change behaviour in a few cases:

- Search tab: main just dismisses the modal and leaves you on Search. Now we reveal a report, so sharing pulls you off Search.
- Wide: `preInsertFullscreenUnderRHP` bails, but we still build the route, so reveal goes through `revealRouteBeforeDismissingModal` instead of dismissing.
- Report open in an RHP: main goes through `dismissModalWithReport`, which can dismiss to the super wide RHP. We skip that.

Intended? If not, could we reuse the prior art gates?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
...
     const isIouReport = isMoneyRequestReport(reportToSubmit);
     const policyTagsForRequestMoney = useMoneyRequestPolicyTags({
         moneyRequestReportID: isIouReport ? reportToSubmit?.reportID : undefined,
         parentChatReportPolicyID: reportToSubmit?.policyID,
         participantReportID: participant?.reportID,
     });
 
+    // Empty draft skips REPORT_DRAFT fallback — report must be in COLLECTION.REPORT to render behind the share modal.
+    const hasNavigationDestination = !!transaction && !!postSubmitNavigationReportID;
+    const destinationReportInCollection = hasNavigationDestination ? getReportOrDraftReport(postSubmitNavigationReportID, undefined, undefined, {}) : undefined;
```
JakubKorytko, line 286: nothing here subscribes to `report_<postSubmitNavigationReportID>`, so eligibility comes from `getReportOrDraftReport`'s deprecated module cache rather than a reactive value. `getSubmitExpensePreMountDestinationRoute.ts:79` passes its `useOnyx` report as the 5th arg for that reason.

Could we pass `reportToSubmit ?? report` here so it's reactive?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
...
+    // Keep pre-mount off while pending navigation owns the reveal — otherwise once the optimistic
+    // report lands, isDestinationReportMissing flips false and usePreMountDestination would schedule
+    // a narrow pre-insert alongside the pending revealRouteBeforeDismissingModal (dual nav).
+    const preMountDestinationRoute =
+        hasNavigationDestination && !isDestinationReportMissing && !pendingNavigationReportID && Navigation.getTopmostReportId() !== postSubmitNavigationReportID
+            ? ROUTES.REPORT_WITH_ID.getRoute(postSubmitNavigationReportID)
+            : undefined;
+
+    const {reveal: revealPreMountDestination, cleanupPreMount} = usePreMountDestination(preMountDestinationRoute, {
+        shouldPreservePreInsertedRouteOnUnmount: () => formHasBeenSubmitted.current,
```
JakubKorytko: `formHasBeenSubmitted` flips at the top of `performUpload`, before the file read and long before `reveal()`. So if we bail after that (the `!transaction || !participant` return, a file read that never calls back, the extension being torn down) we keep the pre-inserted route with nobody left to reveal or remove it, and the user lands on Home silently switched to the destination.

It also leaks the global pre-inserted flag, which the orchestrator reads to pick `ATTRIBUTE_FAST_PATH_HANDLER`, so the next manual submit can be labelled `report_pre_insert` for nothing.

Could we gate on reveal actually having started instead?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
...
+        if (!pendingNavigationReportID || !pendingNavigationReport?.reportID || hasStartedPendingNavigation.current) {
+            return;
+        }
+        hasStartedPendingNavigation.current = true;
+        Navigation.revealRouteBeforeDismissingModal(ROUTES.REPORT_WITH_ID.getRoute(pendingNavigationReportID), {
+            afterTransition: () => {
+                setIsConfirming(false);
+            },
+        });
+    }, [pendingNavigationReportID, pendingNavigationReport?.reportID]);
```
JakubKorytko: this only fires if the optimistic report lands under exactly this ID, and there are two cases where it won't:

- `requestMoney` isn't passed `optimisticChatReportID`, so a rebuilt chat report gets a random ID (`MoneyRequestBuilder.ts:1355`).
- `resolveReportForMoneyRequest` can return undefined (`IOUUtils.ts:520-531`), which drops us to `reportOrAccountID`, an accountID on this route.

Either way the expense is created and the confirm button spins forever, with only the back button out. Main at least dismissed. Pass `optimisticChatReportID` plus a timeout fallback?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
     const listOfParticipants = participants.filter((participant) => participant.selected);
     const participant = listOfParticipants.at(0) ?? selectedParticipants.at(0);
     const reportToSubmit = resolveReportForMoneyRequest({transaction, transactionReport, routeReport: report, policy});
+    const postSubmitNavigationReportID = (isSelfDM(report) ? report : reportToSubmit)?.reportID ?? reportOrAccountID;
```
JakubKorytko, line 276: this duplicates `finalActiveReportID` inside `cleanupAndNavigateAfterExpenseCreate` (`backToReport ?? report?.reportID ?? optimisticChatReportID`). They only agree because the call site mirrors this line term for term.

If they ever drift we reveal one report and register the pending transaction against another, silently. Compute it once and pass it down?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +270,46 @@ function SubmitDetailsPage({
...
     const isIouReport = isMoneyRequestReport(reportToSubmit);
     const policyTagsForRequestMoney = useMoneyRequestPolicyTags({
         moneyRequestReportID: isIouReport ? reportToSubmit?.reportID : undefined,
         parentChatReportPolicyID: reportToSubmit?.policyID,
         participantReportID: participant?.reportID,
     });
 
+    // Empty draft skips REPORT_DRAFT fallback — report must be in COLLECTION.REPORT to render behind the share modal.
```
JakubKorytko: this comment is about the `{}` arg on the next line, not `hasNavigationDestination`. Move it down one line?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +327,132 @@ function SubmitDetailsPage({
...
-                optimisticTransactionID,
-                isTrackIntentUser,
-                delegateAccountID,
+        };
+
+        if (preMountDestinationRoute) {
+            // Create the expense in this turn — do not defer it to afterTransition, or it can be
+            // lost if the reveal animation never completes (e.g. app backgrounded). Only the
+            // confirming-state reset waits for the transition.
+            revealPreMountDestination(() => {
```
JakubKorytko, line 492: why reveal before creating? `reveal()` dismisses straight away, so the optimistic build runs on the frame the animation starts, in our most constrained context. `handleReportPreInsert` defers `createTransaction` to `afterTransition` for that reason, and the hook doc says to do submit work before calling reveal.

Swapping these keeps the crash safety you documented:

```ts
runExpenseCreateAndCleanup(false);
revealPreMountDestination(() => setIsConfirming(false));
```

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +327,132 @@ function SubmitDetailsPage({
...
+                setIsConfirming(false);
             });
+            runExpenseCreateAndCleanup(false);
```
JakubKorytko: minor, not a blocker. `shouldNavigate: false` means we never set or end a follow-up action, which is fine while share doesn't track.

But a leftover pending `DISMISS_MODAL_AND_OPEN_REPORT` for this same report won't get cancelled while the share modal is up (`cancelIfStaleForNavState` bails on any non-fullscreen top route), and the ReportScreen we reveal will end it with the whole share flow in its duration. A `cancelTracking()` at submit avoids that, and calling `startTracking` (see below) would too.

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +327,132 @@ function SubmitDetailsPage({
...
-            optimisticChatReportID: reportOrAccountID,
-            linkedTrackedExpenseReportAction: transaction.linkedTrackedExpenseReportAction,
-        });
+
+        // Pre-mount wasn't possible because the destination report doesn't exist yet. Create the expense without
+        // navigating — the optimistic write creates the report — and let the pending-navigation effect reveal it
+        // once it lands. isConfirming stays true until the reveal transition ends, so the confirm button keeps
+        // its loading state exactly like the pre-mounted path.
+        if (isDestinationReportMissing) {
+            runExpenseCreateAndCleanup(false);
```
JakubKorytko, line 504: cleanup gets `shouldWaitForUpcomingTransition: true` hardcoded (`cleanupAndNavigateAfterExpenseCreate.ts:47`), but on this branch nothing navigates yet.

So if the optimistic report takes longer than 1s to land, the draft transactions get wiped while the modal is still up: the confirmation list loses `transaction` and the receipt disappears behind the spinner. `cleanupAfterExpenseCreate` warns about exactly this.

Could this branch run the cleanup from the reveal's `afterTransition` instead of on a timer that has nothing to do with when we navigate?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +327,132 @@ function SubmitDetailsPage({
...
+            setPendingNavigationReportID(postSubmitNavigationReportID);
```
JakubKorytko, line 505: nothing measures this branch (create, wait for the optimistic report, then reveal), so we can't tell whether the new behaviour is actually faster, and it has no upper bound.

Could share call startTracking and reuse SPAN_SUBMIT_TO_DESTINATION_VISIBLE? SendMoney and submitWithDismissFirst already do. Since the dashboards group by follow-up action it'd want its own value rather than DISMISS_MODAL_AND_OPEN_REPORT, otherwise share's baseline (share sheet, cold-ish start, HEIC conversion) drags that bucket.

Three places would need the new value: the allowlist at ReportScreen.tsx:86, the switch in cancelIfStaleForNavState (currently default: break, so abandoned share submits never get stale-cancelled), and the === DISMISS_MODAL_AND_OPEN_REPORT check in useDeferNonEssentials, which we'd want share to hit. Plus setFastPath, or we trip the missing-setFastPath dev warn on end.

src/libs/Navigation/helpers/cleanupAndNavigateAfterExpenseCreate.ts
```diff
@@ -19,6 +19,8 @@ type CleanupAndNavigateAfterExpenseCreateParams = {
     isInvoice?: boolean;
     linkedTrackedExpenseReportAction?: OnyxEntry<ReportAction>;
     action: DeepValueOf<typeof CONST.IOU.ACTION>;
+    /** When false, runs cleanup only — use when dismiss/reveal already handled navigation. */
+    shouldNavigate?: boolean;
```
JakubKorytko, line 30: two things worth spelling out on this param.

Line 47 passes `shouldWaitForUpcomingTransition: true` regardless, which that param documents as "navigation will be dispatched right after this call". Not true when `shouldNavigate` is false, so the wait turns into a 1s timeout before the draft wipe. Thread `shouldWaitForUpcomingTransition: shouldNavigate`?

And with false we skip `setPendingSubmitFollowUpAction`, so the caller has to own the span itself the way the orchestrator does (`useExpenseSubmission.ts:882-883`). Miss it and the span hangs until the 60s timeout, then `maxDurationFilter` drops the sample. A `__DEV__` warn on `isTracking() && !shouldNavigate` would catch that.

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -65,23 +81,65 @@ jest.mock('@hooks/useReportIsArchived', () => jest.fn(() => false));
 // eslint-disable-next-line @typescript-eslint/naming-convention -- numeric account IDs are the Onyx participants key shape
 jest.mock('@hooks/useReportOrReportDraft', () => jest.fn((reportID?: string) => (reportID ? {reportID, participants: {1: {}, 2: {}}, type: 'chat'} : undefined)));
 
+jest.mock('@libs/getIsNarrowLayout', () => jest.fn(() => false));
```
JakubKorytko: defaulting this to false means most tests here run in the one layout the share extension never sees, and it's why test 5 quietly covers the wide fallback instead of the pre-mounted path.

Flip the default to true and opt out where you actually want wide?

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -32,13 +39,22 @@ jest.mock('@libs/actions/IOU/TrackExpense', () => {
     };
 });
 
+jest.mock('@libs/ReportUtils', () => {
+    const actual = jest.requireActual<ReportUtilsActual>('@libs/ReportUtils');
+    return {
+        ...actual,
+        getReportOrDraftReport: jest.fn(actual.getReportOrDraftReport),
```
JakubKorytko, line 48: mocking this at module level means `mockReturnValue(undefined)` in 7 and 7b applies to every caller in the render tree, and `mockReturnValue(landedReport)` makes any ID resolve to that report. So those tests can pass for a reason other than the one they describe.

Narrow it to the ID, or lean on the Onyx seeding you're already doing in the same tests?

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -147,18 +205,37 @@ async function seedShareState() {
...
-    await waitForBatchedUpdatesWithAct();
-    fireEvent.press(screen.getByTestId('mock-confirm-button'));
-    await waitForBatchedUpdatesWithAct();
+}
+
+function resetNavigationMocksForSubmitDetailsPageTests() {
+    jest.mocked(Navigation.getTopmostReportId).mockReturnValue('report-share-1');
+    jest.mocked(getIsNarrowLayout).mockReturnValue(false);
+    jest.mocked(Navigation.getIsFullscreenPreInsertedUnderRHP).mockReturnValue(false);
+    jest.mocked(Navigation.preInsertFullscreenUnderRHP).mockImplementation(() => {
```
JakubKorytko, line 241: this flips `getIsFullscreenPreInsertedUnderRHP` but leaves `getTopmostReportId` alone, so it doesn't model the real dispatch, which puts the report where `getTopmostReportParams` finds it. That's why the teardown bug above is invisible here.

Adding `jest.mocked(Navigation.getTopmostReportId).mockReturnValue(SHARED_REPORT_ID)` inside this implementation should make it fail.

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -243,4 +330,95 @@ describe('SubmitDetailsPage', () => {
         // Confirm must bail out (no raw HEIC uploaded) until the converted file is ready.
         expect(jest.mocked(readFileAsync)).not.toHaveBeenCalled();
     });
+
+    // Error #5 — when the destination chat is not already visible, submit must reveal the pre-mounted report
+    // and defer navigation to cleanup (shouldNavigate: false) so we do not double-navigate after dismiss.
+    it('reveals the pre-mounted destination and passes shouldNavigate false to cleanup when another report is topmost', async () => {
```
JakubKorytko: `getIsNarrowLayout` is false here, so nothing is ever pre-mounted and this passes via the non-owned `revealRouteBeforeDismissingModal` branch, i.e. the wide fallback rather than the path in the name.

The actual headline path (narrow, real pre-insert, submit resolving to `dismissModal`) has no test. Add a narrow variant asserting `dismissModal` and rename this one?

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -65,23 +81,65 @@ jest.mock('@hooks/useReportIsArchived', () => jest.fn(() => false));
 // eslint-disable-next-line @typescript-eslint/naming-convention -- numeric account IDs are the Onyx participants key shape
 jest.mock('@hooks/useReportOrReportDraft', () => jest.fn((reportID?: string) => (reportID ? {reportID, participants: {1: {}, 2: {}}, type: 'chat'} : undefined)));
 
+jest.mock('@libs/getIsNarrowLayout', () => jest.fn(() => false));
+
+jest.mock('@libs/Navigation/TransitionTracker', () => ({
```
JakubKorytko, line 88: these run their callbacks synchronously, and the `runAfterTransitions` mock also drops `waitForUpcomingTransition` and `maxWaitForUpcomingTransitionMs`. So the pre-insert always beats submit, and the 1s draft-wipe case can't surface here at all.

A variant where `scheduleWhenIdle` never fires would cover confirming before the pre-insert lands.

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -243,4 +330,95 @@ describe('SubmitDetailsPage', () => {
...
+        expect(Navigation.revealRouteBeforeDismissingModal).toHaveBeenCalledTimes(1);
+        expect(Navigation.revealRouteBeforeDismissingModal).toHaveBeenCalledWith(
+            ROUTES.REPORT_WITH_ID.getRoute(SHARED_REPORT_ID),
+            expect.objectContaining({afterTransition: expect.any(Function)}),
+        );
+    });
+
+    // Error #8 — backing out before submit must tear down any pre-inserted destination route before goBack,
+    // or the stale route can flash behind the next modal dismiss.
+    it('cleans up a pre-inserted destination route before goBack when the user backs out without submitting', async () => {
```
JakubKorytko, line 430: two paths still uncovered: unmounting after `formHasBeenSubmitted` flips but before `reveal()`, and a pending submit where the optimistic report never lands under `postSubmitNavigationReportID`.

tests/unit/cleanupAndNavigateAfterExpenseCreateTest.ts
```diff
@@ -194,6 +194,7 @@ describe('cleanupAndNavigateAfterExpenseCreate', () => {
             isInvoice: true,
             hasMultipleTransactions: false,
             shouldAddPendingNewTransactionIDs: false,
+            shouldNavigate: true,
```
JakubKorytko, line 240: this pins the default only. Nothing asserts that `shouldNavigate: false` actually reaches `navigateAfterExpenseCreate`, which is the case the share flow now depends on.

src/pages/Share/SubmitDetailsPage.tsx
JakubKorytko: it can also be a follow-up - not a blocker

JakubKorytko (commented):
the pre-mount idea is right, and the refactor is clean 🎉 

Main thing: this re-does what the manual flow already has (`getSubmitExpensePreMountDestinationRoute` + `handleReportPreInsert` + `usePreMountDestination`) and drops the stability guard, which I think breaks the pre-mount. Details on the `preMountDestinationRoute` line.

On telemetry: `ManualSubmitToDestinationVisible` can't regress here - thank you for that haha. Share never calls `startTracking`, so the span is never started and every touchpoint is behind `isTracking()`. But that also means the whole flow is unmeasured, so the win this PR claims isn't verifiable, added a more precise comment in code.

Three behaviour changes also look inherited rather than decided: Search tab, wide layout, and report-already-in-RHP.

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: Added optional `navigationReportID` parameter to `cleanupAndNavigateAfterExpenseCreate` to eliminate duplication. Other call sites work unchanged since the parameter is optional and defaults to the existing fallback chain (`backToReport` → `report?.reportID` → `optimisticChatReportID`).

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: We're covered. When the destination report doesn't exist, we detect it with `isDestinationReportMissing` and then subscribe reactively to it via `pendingNavigationReportID` here:

```typescript
const [pendingNavigationReport] = useOnyx(`${ONYXKEYS.COLLECTION.REPORT}${getNonEmptyStringOnyxID(pendingNavigationReportID)}`);
```

The effect waits for `pendingNavigationReport?.reportID` to exist before navigating.

tests/ui/SubmitDetailsPageTest.tsx
jmgraa: I added the line to the mock as you mentioned. However, I'm not fully reproducing the teardown bug you're referring to. I traced through the back-button handler in SubmitDetailsPage and it unconditionally calls `cleanupPreMount()` before `goBack()`. All tests pass even after reverting to before the code review fixes and adding just this mock line.

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: Passed `optimisticChatReportID: reportOrAccountID` to `requestMoney()`. Now when the expense is created, the report lands under the ID we're expecting, so the navigation effect triggers correctly and the confirm button isn't stuck.

I didn't add the timeout fallback. In the share flow, we always have a valid destination when we confirm - you can't submit without selecting one. So the edge cases the timeout would catch (report deleted, creation fails completely) shouldn't happen in practice. If you'd still prefer to add the timeout fallback as a safeguard, just let me know and I can put it in.

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: Same pattern exists in `IOURequestStepConfirmation.tsx` - can apply fix there separately.

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +338,139 @@ function SubmitDetailsPage({
...
-        });
+        };
+
+        const runExpenseCreateAndCleanup = (shouldNavigate: boolean) => {
+            performExpenseCreate();
+            cleanupAndNavigateAfterExpenseCreate({...cleanupParams, shouldNavigate});
+        };
+
+        if (preMountDestinationRoute) {
+            cancelTracking();
```
JakubKorytko: this cancels on the pre-mount branch only, but `isDestinationReportMissing` below also passes `shouldNavigate: false`, so it lands in exactly the case the new warn at `cleanupAndNavigateAfterExpenseCreate.ts:48` exists to catch: a live span with nobody left to end it, hanging until the 60s safety timeout.

Hoist it above both branches, or move it into `runExpenseCreateAndCleanup` when `shouldNavigate` is false?

Also worth a word on why share cancels at all. Share never calls `startTracking`, so this is purely defensive against a stale span from a previous flow, which isn't obvious from the call.

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +338,139 @@ function SubmitDetailsPage({
...
+            return;
+        }
+
+        // Pre-mount wasn't possible because the destination report doesn't exist yet. Create the expense without
+        // navigating — the optimistic write creates the report — and let the pending-navigation effect reveal it
+        // once it lands. isConfirming stays true until the reveal transition ends, so the confirm button keeps
+        // its loading state exactly like the pre-mounted path.
+        if (isDestinationReportMissing) {
+            runExpenseCreateAndCleanup(false);
+            setPendingNavigationReportID(postSubmitNavigationReportID);
```
JakubKorytko, line 505: on the timeout you asked about: agreed on the happy path. `parentChatReport` is the draft report and its participants match, so `MoneyRequestBuilder.ts:1338` keeps it and the write lands under `postSubmitNavigationReportID`. The pass-`optimisticChatReportID` half was the real fix.

What the timeout would still catch isn't a deleted report, it's the write not happening at all: anything that throws inside `requestMoney` before the optimistic data is applied, or the participants-mismatch branch at `MoneyRequestBuilder.ts:1356` nulling `chatReport` so `getChatByParticipants` resolves a different ID. Both unlikely here, and neither is worse than a spinner the user can back out of.

So I'd add it, but only because it's about four lines and turns an unbounded state into the old flash, not because I can name a repro. Your call, happy either way.

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +275,52 @@ function SubmitDetailsPage({
...
+        hasNavigationDestination &&
+        !isDestinationReportMissing &&
+        !pendingNavigationReportID &&
+        canPreInsert &&
+        (hasPreInsertedFullscreen || Navigation.getTopmostReportId() !== postSubmitNavigationReportID)
+            ? ROUTES.REPORT_WITH_ID.getRoute(postSubmitNavigationReportID)
+            : undefined;
+
+    const {reveal: revealPreMountDestination, cleanupPreMount} = usePreMountDestination(preMountDestinationRoute, {
+        shouldPreservePreInsertedRouteOnUnmount: () => hasCalledReveal.current,
```
JakubKorytko, line 305: the hook already skips its own teardown when reveal has run (`usePreMountDestination/index.ts:115` checks `hasRevealBeenCalledRef`), and it sets that ref in `revealDestination` before anything else. So this callback only changes the outcome if `route` changes after reveal, which resets the hook's ref but not ours.

Is that the case you're guarding, or can the option go?

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -262,13 +275,52 @@ function SubmitDetailsPage({
...
+    const destinationReportInCollection = hasNavigationDestination ? getReportOrDraftReport(postSubmitNavigationReportID, undefined, undefined, {}) : undefined;
+    // Destination report isn't in COLLECTION.REPORT yet (e.g. a recipient with no existing chat) — it will only
+    // exist after the expense create writes it optimistically, so pre-mounting is impossible.
+    const isDestinationReportMissing = hasNavigationDestination && !destinationReportInCollection?.reportID;
+    // Keep pre-mount off while pending navigation owns the reveal — otherwise once the optimistic
+    // report lands, isDestinationReportMissing flips false and usePreMountDestination would schedule
+    // a narrow pre-insert alongside the pending revealRouteBeforeDismissingModal (dual nav).
+    const hasPreInsertedFullscreen = Navigation.getIsFullscreenPreInsertedUnderRHP();
+    const canPreInsert = !isSearchTopmostFullScreenRoute() && !isReportTopmostSplitNavigator() && !isReportOpenInRHP(navigationRef.getRootState());
+    const preMountDestinationRoute =
```
JakubKorytko: gates look right now, but this is still a hand-rolled copy of the report branch of `getSubmitExpensePreMountDestinationRoute`: same `hasPreInsertedFullscreen` guard, same three nav reads, same `getTopmostReportId` duplicate check, same empty-draft `getReportOrDraftReport` trick. Share is `isFromGlobalCreate: false` and `canPreInsertSearch: false`, so the builder would return the same route, minus the deliberate wide reveal you're keeping.

Could the shared part live in one place? Otherwise the next guard added to the builder silently won't apply here.

src/libs/Navigation/helpers/cleanupAndNavigateAfterExpenseCreate.ts
```diff
@@ -31,14 +42,21 @@ function cleanupAndNavigateAfterExpenseCreate({
     isInvoice,
     linkedTrackedExpenseReportAction,
     action,
+    shouldNavigate = true,
+    navigationReportID,
 }: CleanupAndNavigateAfterExpenseCreateParams) {
+    if (__DEV__ && isTracking() && !shouldNavigate) {
+        // eslint-disable-next-line no-console
+        console.warn('[cleanupAndNavigateAfterExpenseCreate] shouldNavigate=false but span is active. Caller must own span lifecycle — miss this and span hangs 60s until dropped.');
```
JakubKorytko: `Log.warn` is what the rest of the nav helpers use, and it drops the disable while still showing up in dev:

```ts
Log.warn('[cleanupAndNavigateAfterExpenseCreate] shouldNavigate=false but a submit span is still active');
```

It also survives release builds, so a caller that forgets span ownership shows up in logs rather than only on someone's laptop. Worth it, or is dev-only deliberate?

src/libs/Navigation/helpers/cleanupAndNavigateAfterExpenseCreate.ts
```diff
@@ -19,6 +20,16 @@ type CleanupAndNavigateAfterExpenseCreateParams = {
     isInvoice?: boolean;
     linkedTrackedExpenseReportAction?: OnyxEntry<ReportAction>;
     action: DeepValueOf<typeof CONST.IOU.ACTION>;
+    /** When false, runs cleanup only — use when dismiss/reveal already handled navigation.
+     * IMPORTANT: Caller must own telemetry span (see useExpenseSubmission:882-883).
```
JakubKorytko: `useExpenseSubmission:882-883` goes stale on the next edit to that file. Name the thing instead of the lines?

tests/unit/cleanupAndNavigateAfterExpenseCreateTest.ts
```diff
@@ -47,6 +51,48 @@ describe('cleanupAndNavigateAfterExpenseCreate', () => {
...
+            expect.objectContaining({
+                shouldWaitForUpcomingTransition: false,
+            }),
+        );
+        expect(navigateAfterExpenseCreate).toHaveBeenCalledWith(expect.objectContaining({shouldNavigate: false}));
+    });
+
+    it('should warn in __DEV__ when shouldNavigate is false but a submit span is still active', () => {
+        const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
+        jest.mocked(isTracking).mockReturnValue(true);
```
JakubKorytko, line 76: `jest.clearAllMocks()` clears calls but keeps implementations, so this leaks into every test after it and they all run with a span active. Not visible today because nothing else asserts on the warn, but the next `shouldNavigate: false` test inherits it.

Reset it next to the other two in `beforeEach`:

```ts
jest.mocked(isTracking).mockReturnValue(false);
```

tests/unit/cleanupAndNavigateAfterExpenseCreateTest.ts
```diff
@@ -47,6 +51,48 @@ describe('cleanupAndNavigateAfterExpenseCreate', () => {
...
+        cleanupAndNavigateAfterExpenseCreate({
+            action: CONST.IOU.ACTION.CREATE,
+            report: chatReport,
+            draftTransactionIDs: ['txn-1'],
+            transactionID: 'txn-1',
+            isFromGlobalCreate: false,
+            shouldNavigate: false,
+        });
+
+        if (__DEV__) {
```
JakubKorytko: `__DEV__` is always true under Jest, so the `else` never runs and the test reads as if it covers both builds. Assert the warn directly?

tests/ui/SubmitDetailsPageTest.tsx
```diff
@@ -147,18 +217,37 @@ async function seedShareState() {
...
         />,
     );
-    await waitForBatchedUpdatesWithAct();
-    fireEvent.press(screen.getByTestId('mock-confirm-button'));
-    await waitForBatchedUpdatesWithAct();
+}
+
+function resetNavigationMocksForSubmitDetailsPageTests() {
+    jest.mocked(Navigation.getTopmostReportId).mockReturnValue('report-share-1');
+    jest.mocked(getIsNarrowLayout).mockReturnValue(false);
```
JakubKorytko: still false, so most of the file runs in the layout the share extension never sees and every narrow test has to opt in. Flip the default and opt out in the wide test instead?

This is also why the revert experiment on the `preInsertFullscreenUnderRHP` mock thread passed: with narrow false nothing is ever pre-inserted, so `getTopmostReportId` returning `SHARED_REPORT_ID` after a pre-insert never happens and the guard under test is never exercised. Flipping this default is what would make that mock line bite.

JakubKorytko (approved):
the pre-insert gates and `optimisticChatReportID` close most of the last round, so what's left is mainly the pending-navigation branch (still the one that can park the modal in a loading state with nothing to resolve it) plus a few leftovers from round one. Otherwise, LGTM!

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: Added it. Timeout just caps the spinner at 5s.

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: Yes, guarding that case. The callback ensures: once reveal starts, keep the pre-mount alive even if route deps change.

src/pages/Share/SubmitDetailsPage.tsx
jmgraa: After looking deeper into `getSubmitExpensePreMountDestinationRoute`, there is no need to extract code to create a shared helper. `getSubmitExpensePreMountDestinationRoute` can be used directly in `SubmitDetailsPage`.

src/pages/Share/SubmitDetailsPage.tsx
```diff
@@ -286,106 +371,156 @@ function SubmitDetailsPage({
...
+                    existingTransactionDraft,
+                    existingTransaction: storedTransaction ?? transaction,
+                    draftTransactionIDs,
+                    isSelfTourViewed,
+                    betas,
+                    personalDetails,
+                    optimisticTransactionID,
+                    isTrackIntentUser,
+                    delegateAccountID,
+                    optimisticChatReportID: reportOrAccountID,
```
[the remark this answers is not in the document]
jmgraa: `optimisticChatReportID` is only consumed by `getMoneyRequestInformation` as the ID for a freshly built optimistic chat report, so a route ID pointing at an expense report would indeed make the chat and IOU report share an ID. In practice the share flow only ever puts a chat / draft-chat / policy-expense-chat ID in this route param (`ShareTabParticipantsSelector`), so this isn't reachable today, but the override was the one place that skipped the normalization the rest of the page already does.

ChavdaSachin (approved):
LGTM! ✅

mjasikowski (approved):

jmgraa:
As discussed with @IuliiaHerets, the bug mentioned above was just a one-off on a single device and couldn't be reproduced. There are no other cases or steps to trigger it

PR #97468, opened 2026-07-30
Send SearchForTodos concurrently with writes instead of waiting on the queue

TMisiukiewicz:
### Explanation of Change

`SearchForTodos` is fired by `loadPostDataForOpenOrReconnect()` right after `OpenApp`/`ReconnectApp` are enqueued, but `API.read` gates every read behind `waitForWrites()`, so the request didn't actually hit the network until the sequential write queue drained. On large accounts, where `OpenApp` can take 5-10s, `SearchForTodos` sat idle that whole time before even being sent.

This PR makes two changes:

1. Adds a `READS_SENT_DURING_WRITES` set (currently just `SEARCH_FOR_TODOS`) in `src/libs/API/types.ts`. `API.read` skips `waitForWrites` for commands in this set, so they go out concurrently with in-flight writes, mirroring the existing carve-out for the sign-in short-lived/support auth token commands.
2. Because `SearchForTodos`' response touches Onyx keys that `OpenApp`/`ReconnectApp` also write, its Onyx updates must not apply while those writes are still in flight. In `applyHTTPSOnyxUpdates` (`src/libs/actions/OnyxUpdates.ts`), commands in that set now apply via `waitForSequentialQueueIdle().then(() => Onyx.update(updates))` instead of the existing `queueOnyxUpdates` buffer (that buffer only drains when the queue processor finishes with an empty queue, so a response landing after the queue already went idle would be stranded). Gating on queue-idle applies immediately when nothing is in flight, since `isReadyPromise` starts resolved.

Net effect: the request stops occupying the critical path, and its data still lands no earlier than the writes it raced against. The sign-in commands are left as separate checks since they must apply immediately (deferring an auth token behind writes would break login).

Added `tests/unit/SearchForTodosDeferralTest.ts` covering (a) the request firing alongside an in-flight write while its Onyx updates apply only after that write settles, and (b) applying immediately when no write is in flight (regression guard against stranding).

### Fixed Issues

$
PROPOSAL:

### Tests

### Offline tests

N/A

### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


mountiny (commented):
Took a careful pass through this along with the surrounding queue/Onyx machinery. Ran the affected suites locally — `NetworkTest`, `APITest`, `SequentialQueueTest` and the new `SearchForTodosDeferralTest` all pass (60 tests), and lint is clean apart from pre-existing seatbelt warnings.

The goal is right and the diff is small and well-commented. The reasoning for rejecting `queueOnyxUpdates` (stranding a response that lands after the buffer already drained) is correct, and keeping the sign-in commands as separate checks with the "must apply immediately" note is the right call. One issue I think blocks, one robustness concern, and a few smaller things.

---

### 1. Staleness inversion: the payload is now older, but still applied last

The current gate does more than order the Onyx apply — it means the request is **sent** after the writes settle, so its payload reflects post-write server state.

With this change `SearchForTodos` is sent at t0 and applied at t2, on top of data that reflects t2. Onyx `merge` is last-write-wins per field, so the older payload wins on every overlapping key.

Concrete case — offline approve, then reconnect:

1. User is offline and approves report X. `ApproveMoneyRequest` is persisted.
2. Reconnect. `reconnectApp()` enqueues `ReconnectApp`, the queue holds both writes, and `SearchForTodos` now goes out immediately. The server has not processed the approval yet, so X comes back in the approve bucket with `stateNum: SUBMITTED`.
3. The queue drains, the approval succeeds, X becomes `APPROVED` in Onyx.
4. The held to-do payload then applies and merges X back to `stateNum: SUBMITTED`.

X shows as awaiting approval again until a Pusher update or a reload corrects it. The same shape applies to any Pusher update landing between t0 and t2, and `reportActions_` merges can resurrect actions deleted in that window. On a large account the window is the full `OpenApp` duration — the 5-10s cited in the description.

The description says "its data still lands no earlier than the writes it raced against", which is true, but the hazard isn't *when* it lands — it's that it is **older** and lands **last**.

Suggestion: take the fast path only when the queue holds nothing but `OPEN_APP` / `RECONNECT_APP`. Those apply no optimistic data to these keys and their own payload is the fresher one, so the inversion disappears. Fall back to the existing `waitForWrites` behaviour whenever any other write is pending.

### 2. `waitForIdle()` doesn't establish that the write's Onyx data has been applied

`resolveIsReadyPromise()` fires at `src/libs/Network/SequentialQueue.ts:420`, and `flushOnyxUpdatesQueue()` — the thing that actually hands the buffered WRITE batch to Onyx — is called at line 427, in the same synchronous block. The deferred `.then(() => Onyx.update(updates))` is a microtask queued at 420, so it runs after `Onyx.update(writeBatch)` has been *called* but well before it has resolved.

In practice the write gets a one-microtask head start (`Onyx.update` runs `afterInit` synchronously and defers its operations through `clearPromise.then(...)`, so the write's operations are enqueued first). But that is a property of `Onyx.update` internals, not something this code establishes.

There's also a window with no head start at all: if the `SearchForTodos` response lands *after* `resolveIsReadyPromise()` but *while* `flushQueue()`'s `Onyx.update` is still in flight, `waitForIdle()` is already resolved and the todos payload applies immediately, concurrent with the flush. That matters because `OpenApp` / full `ReconnectApp` send `report_`, `reportNameValuePairs_`, `transactions_` and `transactionViolations_` as `SET_COLLECTION`, and `SearchForTodos` merges into exactly those collections.

Worth noting the codebase already treats these as distinct events: `getCurrentFlushPromise()` was added to `QueuedOnyxUpdates` precisely because queued WRITE updates "only land in Onyx when `flushQueue()` runs". It can't be used directly here though — `flushQueue()` nulls `flushPromise` before its `Onyx.update` resolves, so by the time the deferred microtask runs it returns an already-resolved promise.

A gate that does hold: wait for `IS_LOADING_REPORT_DATA` to go `true -> false`. It sits in `finallyData`, which lands in the *same* `Onyx.update(copyUpdates)` batch as the report and transaction `setCollection`s, so observing `false` proves the write's data is applied. No new plumbing needed.

I have not reproduced a failure from this — flagging it as fragility rather than a demonstrated bug.

### 3. Each `updateHandler` call re-reads the gate

`applyHTTPSOnyxUpdates` invokes `updateHandler` up to three times (onyxData -> success/failureData -> finallyData), chained. Each `applyWhenQueueIsIdle` call reads `waitForSequentialQueueIdle()` at its own moment, so an unrelated write arriving between them re-arms `isReadyPromise` and splits what should be one batch across two very different times.

`SearchForTodos` carries no success/finally data today so this is latent, but it's a trap for the next command added to `READS_SENT_DURING_WRITES`. Capturing the gate promise once and reusing it for all three calls would close it.

### 4. The deferral is a no-op on non-leader tabs

`flush()` resolves `isReadyPromise` immediately when `!isClientTheLeader()`, so on a follower tab the to-do payload applies with no gating at all. Not a regression — the same is true today — but it does mean the safety argument only covers the leader tab.

### 5. The tests can't catch an ordering problem

`TODOS_ONYX_DATA` deliberately uses an unrelated key (`NVP_PRIORITY_MODE`), so the tests prove the deferral fires but not that it orders correctly against the write. A case where the write's `onyxData` does `setCollection` on a collection and the deferred read does `mergeCollection` into the same one — asserting the read's members survive — would cover the interaction that actually matters here.

### 6. `NetworkTest`: assertion weakened and the comment is now stale

Step 4 is still labelled "First API Call Verification" but now only asserts the command appears somewhere in the list. Filtering `SearchForTodos` out of `calledCommands()` and keeping the ordering assertion on the remainder would preserve what the test was checking.

### 7. Nit: Fixed Issues

`### Fixed Issues` is a bare `$` with the `PROPOSAL:` placeholder still in place — needs the full issue URL before this is out of draft.

### 8. Minor: this also widens the post-sign-out window

Reads don't set `canCancel` (`src/libs/API/makeRequest.ts` sets it for writes only), so `HttpUtils.cancelPendingRequests()` on sign-out doesn't abort `SearchForTodos`. Its response can now park on `isReadyPromise` for the length of the queue drain and apply afterwards. Reads have never gone through `flushQueue()`'s `preservedKeys` filter so this isn't a new gap, just a wider window.

PR #97502, opened 2026-07-30
Fix: keep invoice room visible after deleting an invoice instead of showing Not here

MelvinBot:
### Explanation of Change

Deleting an invoice from its report showed the “Not here” page and, offline, left the invoice/preview visible. The root cause is that `useGetIOUReportFromReportAction()` excluded invoice reports: they carry the money request action just like IOU/expense reports, but the filter only accepted `isIOUReport`/`isExpenseReport`, so `iouReport` came back undefined for invoices at every call site. That broke both delete navigation (no invoice-room back URL could be built) and optimistic cleanup.

The fix includes invoice reports wherever that same filter is applied:

- [`useGetIOUReportFromReportAction`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/hooks/useGetIOUReportFromReportAction.ts#L20-L24) and the equivalent inline check in [`useDeleteTransactions`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/hooks/useDeleteTransactions.ts#L360) now also accept `isInvoiceReport`, so `iouReport` resolves for invoices. Its `chatReportID` points to the invoice room, so deletion navigates back correctly and the optimistic delete marks the invoice/preview as deleted immediately (important offline).
- [`DeleteMoneyRequest.prepareToCleanUpMoneyRequest`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/libs/actions/IOU/DeleteMoneyRequest.ts#L180) treats invoice reports as expense-style for the total sign math, since invoices store their totals expense-style. Without this the optimistic invoice total/preview would use IOU-style signs.
- [`withReportOrNotFound`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/pages/inbox/report/withReportOrNotFound.tsx#L126) suppresses the NotFound (“Not here”) page while a delete-transaction navigation is in flight, mirroring the existing `ReportNotFoundGuard` behavior.

Regression tests cover deletion through the report-details flow (including invoice-room back navigation and suppression of the “Not here” page) and the offline invoice delete via `useDeleteTransactions`.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97399
PROPOSAL: https://github.com/Expensify/App/issues/97399#issuecomment-5131652081

### Tests

1. Enable Invoices.
2. Use the FAB to send two invoices to any user.
3. Open the invoice room, then open the first invoice report.
4. Select **More → Delete → Delete**.
5. Verify the invoice report closes and the invoice room remains visible without showing the “Not here” page.

### Offline tests

1. Create an invoice while online.
2. Enable **Force offline**.
3. Open the invoice room, then open the invoice report.
4. Select **More → Delete → Delete**.
5. Verify the invoice report closes and the invoice room remains visible without showing the “Not here” page.
6. Disable **Force offline**.

### QA Steps

Same as Tests and Offline tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/ccae6bd6-2b5c-47b4-a5b2-f9977a45d0a5

</details>

<details>

https://github.com/user-attachments/assets/c0cc810d-6e91-436c-8d78-d0b3b6a20c5b

</details>

<details>

https://github.com/user-attachments/assets/a212f8fa-e78a-4221-aa0f-1d0b9a91f379

</details>

<details>

https://github.com/user-attachments/assets/0baa413a-2553-4fe6-bb4b-4516553ac981

</details>

<details>

https://github.com/user-attachments/assets/a1930c2b-35d1-4615-b246-0cc1b2fb30f1

</details>


src/hooks/useExpenseActions.ts
```diff
@@ -533,7 +534,10 @@ function useExpenseActions({reportID, isReportInSearch = false, backTo, onDuplic
                             transaction.transactionID,
                             requestParentReportAction,
                             transactionThreadReport,
-                            iouReport,
+                            // Invoice reports are excluded by useGetIOUReportFromReportAction, so iouReport is undefined for them.
+                            // Fall back to the invoice report itself (its chatReportID points to the invoice room) so deleting an
+                            // invoice navigates back to the invoice room instead of leaving a "Not here" page. See issue #97399.
+                            iouReport ?? (isInvoiceReport(moneyRequestReport) ? moneyRequestReport : undefined),
```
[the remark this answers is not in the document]
MelvinBot: Good catch — fixed in 6dedc8f.

The delete path filtered the money request action's report down to only IOU/expense reports before calling `deleteMoneyRequest`, so for an invoice the actual optimistic delete ran with `iouReport` undefined. `deleteTransactions` now also resolves invoice reports, so the invoice reportAction and the invoice-room report preview are marked pending `DELETE` optimistically instead of the updates landing on `REPORT_ACTIONSundefined`.

- [`src/hooks/useDeleteTransactions.ts:349`](https://github.com/Expensify/App/blob/6dedc8f43491449b7227fa8b2dcfd73eedbdad57/src/hooks/useDeleteTransactions.ts#L349) — added `|| isInvoiceReport(candidateIOUReport)` to the non-split path filter.

Added offline coverage in [`tests/unit/hooks/useDeleteTransactionsInvoiceOfflineTest.tsx`](https://github.com/Expensify/App/blob/6dedc8f43491449b7227fa8b2dcfd73eedbdad57/tests/unit/hooks/useDeleteTransactionsInvoiceOfflineTest.tsx): it deletes an invoice's only expense while offline through the real `deleteTransactions` → `deleteMoneyRequest` path and asserts both the invoice money request action and the invoice-room report preview are optimistically marked deleted (this would fail before the fix, since the writes went to `REPORT_ACTIONSundefined`).

Note: I couldn't execute the Jest suite in this environment, so CI is the validation gate for the new test.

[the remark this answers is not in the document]
nyomanjyotisa: Resolved ✅

nyomanjyotisa (approved):
LGTM!

src/hooks/useExpenseActions.ts
NicolasBonet: Have you considered fixing this inside `useGetIOUReportFromReportAction` instead (adding `isInvoiceReport(candidateIOUReport)` to the filter on line 21)? The filter's comment says its purpose is to exclude *chat* reports (self-DM tracks and split bills), and an invoice report isn't a chat report — so the invoice exclusion looks accidental rather than intentional.

Fixing it at the source would:
- fix all five call sites uniformly, including `MoneyRequestReceiptView` and `useHoldRejectActions`, which may have the same latent bug;
- populate `chatIOUReport` and `isChatIOUReportArchived` for invoices too — with this fallback, `iouReport` is patched but `chatIOUReport` stays `undefined` and the archived flag stays `false`;
- remove the duplicated fallback expression and near-identical comment block here and in `DynamicReportDetailsPage`.

The trade-off is a wider blast radius (`MoneyRequestHeaderSecondaryActions` uses `iouReport` for other actions), so if the surgical approach is deliberate, that's fine — but worth a comment explaining why the hook itself wasn't changed.

src/hooks/useDeleteTransactions.ts
```diff
@@ -343,7 +343,10 @@ function useDeleteTransactions({report, reportActions, policy}: UseDeleteTransac
                 const iouReportID = isMoneyRequestAction(action) ? action?.reportID : undefined;
                 const candidateIOUReport = allReports?.[`${ONYXKEYS.COLLECTION.REPORT}${iouReportID}`];
                 // For self-DM tracks and split bills, action.reportID resolves to a chat report, not an IOU/expense report.
-                const iouReport = isIOUReport(candidateIOUReport) || isExpenseReport(candidateIOUReport) ? candidateIOUReport : undefined;
+                // Invoice reports also carry the money request action; without them the optimistic delete would run with an
+                // undefined iouReport, so the invoice/preview would not be marked deleted (visible until a server response,
+                // or indefinitely offline). Its chatReportID points to the invoice room. See issue #97399.
+                const iouReport = isIOUReport(candidateIOUReport) || isExpenseReport(candidateIOUReport) || isInvoiceReport(candidateIOUReport) ? candidateIOUReport : undefined;
```
NicolasBonet, line 360: Two things here:

1. This filter and the one in `useGetIOUReportFromReportAction.ts:21` serve the same purpose and even share the same "self-DM tracks and split bills" comment, but they now behave differently (this one includes invoice reports, the hook doesn't). That divergence is a trap for future readers — if the hook isn't updated to match, it'd be good to note why here.

2. With invoice reports now flowing through as `iouReport`, they reach `prepareToCleanUpMoneyRequest`, where `isExpenseReportType = isExpenseReport(iouReport)` (`DeleteMoneyRequest.ts:177`) is `false` for invoices — so the totals update falls into `updateIOUOwnerAndTotal`, which applies IOU-style sign math. Invoice totals are stored expense-style, and the rest of the IOU code treats them that way (`MoneyRequestBuilder.ts:1735` and `UpdateMoneyRequest.ts:1614` both use `isExpenseReport(iouReport) || isInvoiceReport(iouReport)`). The result is a wrong optimistic total and preview amount whenever the invoice report survives the delete. This is latent today (invoices are single-transaction, so `shouldDeleteIOUReport` is true and the report is nulled anyway), but this change makes the path reachable — suggest extending `DeleteMoneyRequest.ts:177` with `|| isInvoiceReport(iouReport)` for consistency.

NicolasBonet (changes requested):
Nice targeted fix with solid regression coverage — the tests exercise the real `isInvoiceReport` / `getNavigationUrlOnMoneyRequestDelete` logic and the real optimistic Onyx writes, and the offline case is covered. Requesting changes for two things (inline): invoice reports now reach `prepareToCleanUpMoneyRequest` where the totals math still treats them as IOU-style reports, and the invoice exclusion looks like it should be fixed at its source in `useGetIOUReportFromReportAction` rather than patched at individual call sites.

src/hooks/useExpenseActions.ts
MelvinBot: Good call — moved the fix to the source. [`useGetIOUReportFromReportAction.ts:24`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/hooks/useGetIOUReportFromReportAction.ts#L20-L24) now includes `isInvoiceReport(candidateIOUReport)` in its filter, since an invoice report isn't a chat report (the filter's real purpose is excluding self-DM tracks / split bills). That resolves `iouReport` uniformly at all call sites and also populates `chatReport`/`isChatIOUReportArchived` for invoices, so I removed this duplicated fallback (and the now-unused import) here.

Before doing this I checked the blast radius on all consumers of the hook. Every `iouReport` consumer is either an invoice delete/cleanup path that already carried this exact fallback, or is behind an `isTrackExpenseAction` guard that invoices never hit (e.g. `MoneyRequestHeaderSecondaryActions`' hold/unhold/pay don't read `iouReport` at all). The only real behavior change is that `chatReport`/`isChatIOUReportArchived` now populate for invoice rooms, which is only observable for *archived* invoice rooms and is the more-correct behavior. All existing invoice-delete tests still pass.

Separately, this made a latent invoice-total bug reachable, which I fixed too — see the reply on `useDeleteTransactions.ts`.

src/hooks/useDeleteTransactions.ts
MelvinBot: Both addressed.

1. **Filter divergence:** fixed at the source — [`useGetIOUReportFromReportAction.ts:24`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/hooks/useGetIOUReportFromReportAction.ts#L20-L24) now includes `isInvoiceReport`. This hook builds its `iouReport` inline and doesn't consume that one, but the two filters are now intentionally aligned (both `isIOUReport || isExpenseReport || isInvoiceReport`) and the existing comment [here at line 357](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/hooks/useDeleteTransactions.ts#L355-L360) explains why invoices must be included.

2. **Invoice-total sign math:** fixed [`DeleteMoneyRequest.ts:180`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/src/libs/actions/IOU/DeleteMoneyRequest.ts#L178-L180) to `isExpenseReport(iouReport) || isInvoiceReport(iouReport)`, matching `MoneyRequestBuilder`/`UpdateMoneyRequest`, so the optimistic invoice total/preview uses expense-style signs.

Added regression coverage in [`useDeleteTransactionsInvoiceOfflineTest.tsx`](https://github.com/Expensify/App/blob/f1b1a9f342d769e40543fece28349dab99f8d483/tests/unit/hooks/useDeleteTransactionsInvoiceOfflineTest.tsx): deleting one transaction from a multi-transaction invoice (report survives, `shouldDeleteIOUReport` false) asserts the surviving invoice total is `-100` (expense-style), with owner/manager unchanged. I confirmed the test genuinely guards the bug — reverting only the `DeleteMoneyRequest.ts` change makes it fail with `+300` (the IOU-style result).

NicolasBonet (approved):

PR #97555, opened 2026-07-31
fix: make agent-device replays fail safely

BartekObudzinski:
### Explanation of Change

**Problem:** Headless agent-device replays could silently report success from the wrong screen, spend all measured runs, and then report "No captured runs" with no indication of what went wrong. The replay CLI timeout on 0.20.x left the daemon request running, blocking the next run. Measurement flows were mixed with interactive macros inside the agent-device skill, with no single place to look up which flow owns a given Sentry span.

**Solution:** Moved all Sentry span flows into `measure-telemetry-span`, collocated with the runner and deadline wrapper. Each flow declares `# @span` with exactly one owner per span; the runner refuses to pick between two candidates. Preconditions and postconditions from flow headers are now enforced independently of the flow body. The runner fails on the first replay that emits no span and prints the current screen instead of exhausting all runs first.

### Fixed Issues

$ https://github.com/Expensify/App/issues/100391
PROPOSAL:

### Tests

1. Run `node --test .claude/skills/measure-telemetry-span/scripts/replay-with-deadline.test.mjs` and verify all 3 tests pass.
2. Run `bash -n .claude/skills/measure-telemetry-span/measure.sh` and `npx shellcheck -e SC1091 .claude/skills/measure-telemetry-span/measure.sh`.
3. On Android, run `.claude/skills/measure-telemetry-span/measure.sh ManualOpenCreateExpense 3 android`.
4. Verify the warmup and all three measured replays pass and each reset returns to Inbox.
5. Replay a flow with an intentionally wrong final assertion and verify it exits with code 1.
6. Replay a five-second selector wait through the deadline wrapper with `--timeout 1000` and verify it exits with code 124 near one second, cleans the isolated daemon, and releases the emulator.

### Offline tests

Not applicable because these changes only affect local automation tooling and do not change application networking or offline behavior.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] If the PR modifies a generic component, I tested and verified that those changes do not break usages of that component in the rest of the App (i.e. if a shared library or component like `Avatar`, I verified that `Avatar` is working as expected in all cases)

### Screenshots/Videos
<details>

Not applicable. There are no application UI changes.

</details>

<details>

Not applicable. There are no application UI changes.

</details>

<details>

Not applicable. There are no application UI changes.

</details>

<details>

Not applicable. There are no application UI changes.

</details>

<details>

Not applicable. There are no application UI changes.

</details>


.claude/skills/agent-device/flows/README.md
```diff
@@ -23,14 +23,31 @@ Before manually navigating, use this human-in-the-loop loop:
...
+    --session <name>
+AGENT_DEVICE_STATE_DIR="$HOME/.agent-device-expensify-headless" \
+node .claude/skills/agent-device/scripts/replay-with-deadline.mjs \
+    .claude/skills/agent-device/flows/scenarios/<name>.ad \
+    --session <name> \
+    --timeout 120000 \
+    -e KEY=VALUE
 ```
 
+Tested Agent Device versions 0.20.1 through 0.20.3 accept `replay --timeout` but can leave the daemon request running until a selector timeout. The repository wrapper enforces the wall-clock deadline outside the daemon, exits with code 124, and cleans the dedicated headless daemon. Always use a separate `AGENT_DEVICE_STATE_DIR` for this workflow so timeout cleanup cannot stop an interactive session. Selector waits remain the per-step bounds and produce the useful divergence report.
```
kacper-mikolajczak: This looks like an upstream fix (if still exists on latest version).

.claude/skills/agent-device/flows/macros/return-to-inbox.ad
kacper-mikolajczak, line 1: Looks like quite an expensive macro due to the wait times vs # of actions perfromed. How performant/useful it is based on your tests?

.claude/skills/agent-device/flows/scenarios/create-expense-manual.ad
```diff
@@ -0,0 +1,37 @@
+# @desc   Open Create Expense from the FAB, enter an amount on the Manual tab, fill the Merchant field, and submit. Supports configurable currency-aware submit button matching. Exercises the full submit-chain Sentry spans end-to-end. Assumes a workspace where Category is not required.
+# @pre    id="BaseSidebarScreen"
+# @pre    id="floating-action-button" || role="button" label="Open actions menu"
+# @post   id="SearchPageNarrow" || id="Search"
+# @post   role="tab" label="Expenses" || role="button" label="Expenses" || label="Expenses" hittable=true
+# @reset  .claude/skills/agent-device/flows/macros/switch-to-inbox.ad
```
kacper-mikolajczak, line 6: If I could use this as an example to understand concept behind reset tag. It is used to link macros between each other so they create kind of a pair that _resets_/mitigates itself, so the app lands on the same place?

Concept itself seems really nice, however I wonder how useful it would be. Let me know what's your experience/use-cases for that, thanks!

.claude/skills/agent-device/flows/scenarios/send-message.ad
```diff
@@ -0,0 +1,24 @@
...
+# @param  MESSAGE Unique message text to send and verify in the opened chat.
+# @param  TARGET_CHAT_LABEL Exact accessible label of the report/chat row to open from the Inbox LHN.
+# @param  EXPECTED_CHAT_TITLE Visible report/chat title expected after navigation.
+
+is exists "id=\"BaseSidebarScreen\""
+is exists "id=\"searchButton\" || role=\"button\" label=\"Search\""
+press "role=\"tab\" label=\"Inbox. Your review is required\" || role=\"tab\" label=\"Inbox\" || label=\"Inbox. Your review is required\" hittable=true || label=\"Inbox\" hittable=true"
+press "role=\"button\" label=\"${TARGET_CHAT_LABEL}\" || label=\"${TARGET_CHAT_LABEL}\" hittable=true"
+wait text "${EXPECTED_CHAT_TITLE}" 10000
+wait "id=\"composer\"" 10000
```
kacper-mikolajczak, line 20: Those waits are added in order to prevent failure due to app being too slow to actually respond to the .ad script input, right? 

How we could mitigate it on more robust level instead of manually inserting waits?

kacper-mikolajczak (commented):
Thanks for the initiative @BartekObudzinski! Left couple of comments in the first round of review to get more insight - thanks for the clarification in advance :)

.claude/skills/agent-device/flows/README.md
BartekObudzinski: Agreed on the upstream part. We tested 0.20.1 through 0.20.3 and 0.20.5 is out, so I will retest and open an upstream issue if it still hangs.

The cleanup half stays either way. On timeout the wrapper kills the process group and runs `daemon stop --clean` against the isolated state dir, so the emulator gets released. If upstream fixes the hang, the wrapper shrinks to just that.

.claude/skills/agent-device/flows/macros/return-to-inbox.ad
BartekObudzinski: `10000` is the selector wait timeout, not a sleep. `wait "<selector>" 10000` polls and continues as soon as the selector resolves, so the real cost is the transition itself. Fixed duration waits are banned in the authoring rules.

The macro runs between measured runs in `measure.sh`, never inside the span. I have no recorded timing, so I will capture an Android run and post the actual reset duration.

.claude/skills/agent-device/flows/scenarios/create-expense-manual.ad
BartekObudzinski: One way, not a pair. A scenario points at a macro that restores that scenario's own `@pre`. The macro knows nothing about the scenario, so one macro serves several scenarios.

`measure.sh` is the consumer. It reads `@reset` and runs that macro between measured repeats. With no `@reset` it relaunches the app, which costs seconds and turns the next run into a cold start. That is the use case: N comparable samples instead of N samples with a relaunch mixed in.

.claude/skills/agent-device/flows/scenarios/send-message.ad
BartekObudzinski: These are selector waits. `wait "id=\"composer\"" 10000` polls until the composer exists and continues then. `10000` is only where we give up and print the divergence report.

They are hand written because agent-device does not wait implicitly before a press. That is the upstream change worth asking for. For now `validate-flows.mjs` enforces the guard where it matters, for example a floating action button press must be followed by a selector wait.

.claude/skills/agent-device/flows/README.md
kacper-mikolajczak: Thanks! Let me know how testing ended up going.

.claude/skills/agent-device/flows/macros/return-to-inbox.ad
kacper-mikolajczak: Ok! Let me know about testing, too :thx:

.claude/skills/agent-device/flows/scenarios/create-expense-manual.ad
kacper-mikolajczak: Got it, but seems like the reset macros added are quite simple like one lie i.e. `Hit the back button`. I wonder how much value it brings vs the alternative of inlining such or let the agent handle the reset based on `pre` conditions.

.claude/skills/agent-device/flows/scenarios/send-message.ad
kacper-mikolajczak: Got it.

.claude/skills/agent-device/flows/README.md
BartekObudzinski: Retested on 0.20.6 and it still hangs, so the wrapper stays.

Probe flow is a single step that waits 5000ms on a selector that can never match, run with `--timeout 1000`. Bare `agent-device replay` took 5505ms on iOS and 5262ms on Android, both exit 1. Through the wrapper it was 1690ms on iOS and 1333ms on Android, exit 124, and it printed `Daemon stopped (graceful). Retained runner cleanup completed.`

So the daemon request really does run to the selector timeout and the flag has no effect on wall clock. I will open the upstream issue and link it here. Updated the README line to say 0.20.1 through 0.20.6 with those numbers.

.claude/skills/agent-device/flows/macros/return-to-inbox.ad
BartekObudzinski: Measured it on the iOS simulator. `return-to-inbox.ad` runs in 1895ms, 1892ms, 1914ms across three samples, 4 steps in 1.7s of replay with the rest being node startup. `switch-to-inbox.ad` is 2101ms.

The alternative is worse than I expected. When a scenario has no `@reset`, measure.sh falls back to `agent-device open --relaunch`. I timed that path twice and it took 67298ms and 67011ms, and both times it never reached the scenario `@pre`. iOS restores the deep route on relaunch, so the snapshot afterwards still showed `IOURequestStartPage` plus a React Native warning overlay. So for these scenarios the relaunch fallback does not just cost a cold start, it fails outright.

That is the case for keeping the macro. Two seconds between measured runs against a fallback that does not get back to the starting screen at all.

.claude/skills/agent-device/flows/scenarios/create-expense-manual.ad
BartekObudzinski: They are not one-liners. `return-to-inbox` is four statements: press Back, wait for the Inbox tab, press it, wait for `BaseSidebarScreen`. `switch-to-inbox` is three with an `is exists` guard first. Most of the bulk is the Inbox tab selector union, which has to cover the "Your review is required" label variant. Six scenarios share two macros, so inlining means copying that union six times and fixing it six times when a label changes.

On letting the agent reset from `@pre`: that does not work here. `@pre` is a selector saying where you must be, not a path saying how to get there. From an open expense screen there is nothing in `id="BaseSidebarScreen"` that tells you to press Back and then the Inbox tab.

Your question made me actually run this on iOS and it turned up a real bug, so thanks for pushing on it. Every one of these flows was Android-only. React Native maps `role="tab"` to no iOS accessibility trait, so the tabs come back as `Other` with `hittable: false`, and both halves of the `role="tab" ... || label=... hittable=true` union miss. `switch-to-inbox.ad` failed at step 1 on iOS and so did the five scenarios that inline the same union.

The dangerous part is what a naive fix does. I first tried appending a plain `label="..."` alternative, then checked what that actually presses: `press label="Spend"` tapped (201, 98), the page header, while the real tab sits at y=779. No error, replay continues against the wrong screen, which is exactly the false success this PR is supposed to prevent.

So the fix only adds a bare-label alternative where the label is provably unique, and the validator now requires the author to declare it with `@unique-label` instead of allowing bare labels generally. "Inbox. Your review is required" is unique in the tree. Home and Spend are not, so `switch-home-to-inbox` and `switch-home-to-reports` stay Android-only until the app puts a `testID` on each tab, which I noted in the README. Pushed with a test covering both the exempt and the still-rejected case.

BartekObudzinski: Correction to my last comment: I reverted the iOS fix out of this PR. It was scope creep on a PR you are already reviewing, and it added a new metadata field on top of everything else. The branch is back to what you reviewed.

The finding itself stands, I just moved it. Verified on Android after the revert, so nothing here regressed:

| check | result |
| --- | --- |
| `switch-to-inbox.ad` | exit 0, 2346ms |
| `open-create-expense.ad` | exit 0, 4864ms |
| `return-to-inbox.ad` | exit 0, 2907ms |
| reset macro, 3 samples | 2639 / 2595 / 3540ms |
| `measure.sh ManualOpenCreateExpense 3 android` | exit 0, avg 389ms over 3 runs |

That last one is the PR test step 4, which now has a real result: samples 400ms, 392ms, 375ms, and every reset returned to Inbox.

Two things I want to flag rather than quietly fix.

`measure.sh` defaults to `PLATFORM="${3:-ios}"`, so a bare `measure.sh <span>` targets the platform where every scenario currently fails at the first tab press. Worth either changing the default to android or documenting it until the iOS work lands separately.

The first `measure.sh` run aborted with exit 1 in `dismiss_react_native_overlays` after five attempts. The overlay was genuine, an "Open debugger to view warnings" bar sitting directly on the nav tabs, so the guard is right to exist. But the same `agent-device react-native dismiss-overlay` returned "verified gone" on the first try when I ran it by hand a moment later, and then measure.sh completed. So five attempts with a one second sleep is a bit tight and a transient overlay kills the whole measurement. I have not touched it, happy to raise the attempts or add backoff if you think it is worth it.

.claude/skills/agent-device/flows/macros/return-to-inbox.ad
BartekObudzinski: Reset numbers on Android too, now that I had an emulator up: 2639 / 2595 / 3540ms for `return-to-inbox.ad`. Slower than iOS but same order.

kacper-mikolajczak: Got it. I found those acceptable as a work-around, but in the long term how would you see us strengthening agent-device setup so it works seamlessly without static helpers like this? What is the root cause coming from your research (e.g. lack of enough locators, agent-device peformance, other setup issues)? 

Thanks for digging into this @BartekObudzinski

.claude/skills/agent-device/flows/README.md
kacper-mikolajczak: Ok, thanks! Let's see how quickly we can get upstream fix running.

.claude/skills/agent-device/flows/macros/return-to-inbox.ad
BartekObudzinski: The root cause is two things, and only one is agent-device.

First, agent-device does not wait implicitly before an interaction. A press dispatches against whatever tree exists right now, so every transition needs a hand written selector wait after it. That is the upstream ask I would make: let press retry until the target resolves and is hittable, so the wait becomes the default rather than something the author remembers.

Second, a flow can only say where it must be, not how to get there. `@pre` is a selector. There is no way to name an app state and ask the tool to reach it, which is what forces a macro to exist at all. The fallback when a scenario has no `@reset` is `open --relaunch`, and I timed that at 67298ms and 67011ms on iOS, and neither run even landed on the `@pre` because iOS restores the deep route.

The third piece is on our side, not theirs. The Inbox tab needs a four way selector union because the label flips to "Inbox. Your review is required" depending on review state. Stable ids on the bottom tabs would collapse that union to one selector and shrink the macros to almost nothing.

.claude/skills/agent-device/flows/README.md
```diff
@@ -1,12 +1,13 @@
 # Flows
 
+This file is the reference for every `.ad` file in the repository: the metadata spec, selector rules, recording, and maintenance. Measurement flows live in [`measure-telemetry-span/flows/`](../../measure-telemetry-span/SKILL.md) and follow the same rules.
```
kacper-mikolajczak: > Measurement flows live in [`measure-telemetry-span/flows/`](../../measure-telemetry-span/SKILL.md) and follow the same rules.

This could be omitted here - such information should be located in a skill, so whenever someone wants to use it, it's accessible from there but does not clutter outside.

.claude/skills/agent-device/flows/README.md
```diff
@@ -34,13 +35,11 @@ Before manually navigating, use this human-in-the-loop loop:
...
-## QA workflow (separate)
+## Replay lifecycle
 
-`flows/tests/` is reserved for dedicated QA automation and should not be offered to users as part of the interactive helper loop above. Run these flows with the dedicated test runner:
+No flow in this repository owns application lifecycle, so none of them is a self-contained `agent-device test` input. `agent-device test` creates an isolated session per attempt; use it only for scripts that own `context`, `open`, and cleanup. Repository flows expect the caller to open the app and prepare a named session first.
 
-```bash
-agent-device test .claude/skills/agent-device/flows/tests/<name>.ad -e KEY=VALUE ...
-```
+Tested Agent Device versions 0.20.1 through 0.20.6 accept `replay --timeout` but can leave the daemon request running until a selector timeout. For headless and measured runs, drive replay through [`measure-telemetry-span/scripts/replay-with-deadline.mjs`](../../measure-telemetry-span/SKILL.md), which enforces the wall-clock deadline outside the daemon and exits with code 124. Selector waits remain the per-step bounds and produce the useful divergence report.
```
kacper-mikolajczak: I am concerned about leaving such information as it could easily get outdated. E.g. current version of agent-device is `0.20.9`

kacper-mikolajczak: Also, please avoid mentioning the skill's implementation details outside of the skill itself please. Thanks!

kacper-mikolajczak (changes requested):
Thanks for the updates since last review. Looking good! Left couple of comments

.claude/skills/agent-device/flows/README.md
BartekObudzinski: Done, removed that sentence.

.claude/skills/agent-device/flows/README.md
BartekObudzinski: Removed the whole paragraph. The version range was already stale and the deadline wrapper detail belongs in `measure-telemetry-span`, not here.

BartekObudzinski:
Hey @rlinoz @Julesssss   opened a PR that reorganises the agent-device measurement setup: Sentry span flows moved into `measure-telemetry-span`, replays now fail loudly on wrong screen or missing span instead of silently spending all runs, and each span has exactly one owning flow so the runner never picks by filesystem order. wdyt?

.claude/skills/agent-device/SKILL.md
```diff
@@ -121,4 +121,6 @@ Read these files directly for device automation guidance (bootstrap, exploration
 
 ## Flows
 
-Repeatable steps (sign-in, onboarding, etc.) are captured as composable `.ad` snippets under [`flows/`](flows/README.md). For interactive usage, propose and run only `flows/macros/` helpers. `flows/tests/` belongs to a separate QA workflow and must not be proposed by this skill; QA/perf runs execute them via `agent-device test <path>`.
+This skill owns interactive automation only: reusable setup and navigation macros under [`flows/macros/`](flows/README.md), with platform overrides under `flows/macros/<platform>/`. Propose and run macros through the [Agent decision loop](flows/README.md). `flows/README.md` is also the reference for the `.ad` metadata spec, selector rules, and recording workflow, so read it before authoring any `.ad` file in this repository.
+
+Sentry span measurement lives in the [`measure-telemetry-span`](../measure-telemetry-span/SKILL.md) skill, which owns its own `flows/` and the replay deadline wrapper. Do not add measurement flows here.
```
rlinoz: Not really necessary I think

```suggestion
```

.claude/skills/agent-device/flows/macros/android/send-message.ad
```diff
@@ -1,8 +1,7 @@
 context platform=android
-# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the QA scenario that exercises the ManualSendMessage Sentry span, see flows/tests/send-message.ad.
+# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the flow that measures the ManualSendMessage Sentry span, see measure-telemetry-span/flows/send-message.ad.
```
rlinoz: Same here, no reason to mention sentry I think

```suggestion
# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible.
```

.claude/skills/measure-telemetry-span/flows/create-expense-manual.ad
```diff
@@ -2,16 +2,9 @@
...
-# @tag    navigation
-# @tag    perf
-# @tag    sentry-ManualCreateExpenseSubmit
-# @tag    sentry-ManualSubmitToDestinationVisible
-# @tag    sentry-ManualCreateExpenseServerResponse
-# @tag    sentry-ManualConfirmationMount
-# @tag    sentry-ManualConfirmationListReady
-# @tag    sentry-ManualConfirmationReceiptLoad
-# @tag    sentry-ManualGeolocationWait
+# @span   ManualConfirmationMount
```
rlinoz, line 7: Curious, is span a new thing?

rlinoz (commented):
Is there any comparisong, before/after this changes?

.claude/skills/measure-telemetry-span/flows/create-expense-manual.ad
BartekObudzinski: Sentry spans are not new. The PR just moves measurement flows into `measure-telemetry-span`, enforces one owning flow per span so the runner does not pick by filesystem order, and makes the runner fail on a missing span instead of spending all runs silently.

.claude/skills/measure-telemetry-span/SKILL.md
```diff
@@ -42,28 +55,54 @@ If you see **no Android device** (`adb devices` empty): append **`--boot`** to t
...
+To debug one flow, or to run it as a QA check, open a session in the isolated state directory and drive the same wrapper `measure.sh` uses:
+
+```bash
+export AGENT_DEVICE_STATE_DIR="$HOME/.agent-device-expensify-headless"
+agent-device open <app-id> --platform <ios-or-android> --session <name>
+node .claude/skills/measure-telemetry-span/scripts/replay-with-deadline.mjs \
+    .claude/skills/measure-telemetry-span/flows/<name>.ad \
+    --session <name> \
+    --timeout 120000 \
+    -e KEY=VALUE
```
Julesssss, line 97: Do we need to add permissions for this script example?

BartekObudzinski: No extra permissions are needed. The wrapper reuses the existing `agent-device` CLI and a separate `AGENT_DEVICE_STATE_DIR`. It does not add file system access beyond what `agent-device` already requires, and it does not touch CI or native app permissions @Julesssss

BartekObudzinski:
I'm fixing the state of the PR, because is something that should not happend, please ignore that 👀

BartekObudzinski:
The PR is updated @rlinoz @Julesssss @kacper-mikolajczak

BartekObudzinski:
Kindly bumping @Julesssss @rlinoz @kacper-mikolajczak

.claude/skills/agent-device/flows/macros/android/send-message.ad
```diff
@@ -1,8 +1,7 @@
 context platform=android
-# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the QA scenario that exercises the ManualSendMessageVisible Sentry span, see flows/tests/send-message.ad.
+# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the flow that measures the ManualSendMessageVisible Sentry span, see measure-telemetry-span/flows/send-message.ad.
```
kacper-mikolajczak: Looks like unnecessary mention of out-of-scope tool. Let's treat flows as abstract entities.

.claude/skills/agent-device/flows/macros/web/send-message.ad
```diff
@@ -1,8 +1,7 @@
 context platform=web
-# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the QA scenario that exercises the ManualSendMessageVisible Sentry span, see flows/tests/send-message.ad.
+# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the flow that measures the ManualSendMessageVisible Sentry span, see measure-telemetry-span/flows/send-message.ad.
```
kacper-mikolajczak: Same as for send-message

.claude/skills/agent-device/flows/macros/send-message.ad
```diff
@@ -1,7 +1,6 @@
-# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the QA scenario that exercises the ManualSendMessageVisible Sentry span, see flows/tests/send-message.ad.
+# @desc   Send a chat message from inside an already-open chat. Reusable helper for setting up state in other flows. Does not navigate or open a chat - assumes the composer is visible. For the flow that measures the ManualSendMessageVisible Sentry span, see measure-telemetry-span/flows/send-message.ad.
```
kacper-mikolajczak: As above

.claude/skills/agent-device/flows/README.md
```diff
@@ -78,11 +76,18 @@ agent-device replay <flow>.ad -e EMAIL=other@example.com
...
+- **A `@pre` must distinguish the start screen from the destination.** A bottom-tab label such as `text="Home"` is rendered on every tab, so asserting it proves nothing. Anchor on a screen-scoped `id=` instead - `id="HomePage"` for Home, `id="BaseSidebarScreen"` for Inbox - otherwise a flow that starts on the wrong screen reports success while the action it measures never happens.
 - **Confirm every selector on the platform it is written for.** `snapshot -i` prints display tags, which are not selector values - a node printed as `[text-field]` may only match `role="textbox"`. Check the exit code of `agent-device is visible "<selector>"` before committing a selector, and never carry one across platforms unchecked.
-- **Every flow declares `@desc` and `@pre`.** Add `@post` for outcome-bearing flows; utility flows (for example `go-back`) may omit it. Add `@tag` when applicable.
-- **Choose directory intentionally.** Put reusable setup/navigation steps in `flows/macros/`; put outcome verification scenarios in `flows/tests/`.
+- **Every flow declares `@desc` and `@pre`.** Measurement flows also declare at least one `@post`. Utility macros (for example `go-back`) may omit `@post`.
```
kacper-mikolajczak: Referencing out of scoep detail about measuremnet flows

kacper-mikolajczak (changes requested):
LGTM in general but please make sure that we are avoiding unnecessary coupling between task-specific skill and general instructions. Thanks!

.claude/skills/agent-device/flows/macros/android/send-message.ad
BartekObudzinski: Removed. The `@desc` now stops at "assumes the composer is visible" and says nothing about spans or the other skill. Same edit on the web and unscoped copies.

.claude/skills/agent-device/flows/macros/web/send-message.ad
BartekObudzinski: Removed here too.

.claude/skills/agent-device/flows/macros/send-message.ad
BartekObudzinski: Removed.

.claude/skills/agent-device/flows/README.md
BartekObudzinski: You flagged line 84 but the coupling was in five places, so I took all of them out.

Dropped the `@span` and `@reset` rows from the metadata table, the "One owner per span" rule, and the "Choose the skill intentionally" rule. Line 84 now says "Outcome-bearing flows" instead of "Measurement flows". The recording step no longer offers `measure-telemetry-span/flows/` as a destination or mentions `@span`.

Nothing is lost by moving them. `measure-telemetry-span/SKILL.md` already documents `@span`, `@reset`, and the one owner per span rule in its Contract section, so those headers stay described in the skill that consumes them. The dependency now runs one way, the measurement skill points up at this README for the shared spec and this README knows nothing about it.

One leftover I want to raise rather than fix quietly. `agent-device/SKILL.md` still has a line saying measurement flows live in `measure-telemetry-span`. Rafael asked me to delete it earlier and that thread got resolved without the edit landing. It is the same coupling you are pointing at, so I am happy to remove it, but it reads as a useful signpost for someone about to add a measurement flow in the wrong place. Your call.

BartekObudzinski: You flagged line 84 but the coupling was in five places, so I took all of them out.

Dropped the @span and @reset rows from the metadata table, the "One owner per span" rule, and the "Choose the skill intentionally" rule. Line 84 now says "Outcome-bearing flows" instead of "Measurement flows". The recording step no longer offers measure-telemetry-span/flows/ as a destination or mentions @span.

Nothing is lost by moving them. measure-telemetry-span/SKILL.md already documents @span, @reset, and the one owner per span rule in its Contract section, so those headers stay described in the skill that consumes them. The dependency now runs one way, the measurement skill points up at this README for the shared spec and this README knows nothing about it.

rlinoz:
Can we mark this one as ready for review?

BartekObudzinski:
@rlinoz Sure, but do we need ticket?

rlinoz:
Ah sorry, I thought we had one, here you go https://github.com/Expensify/App/issues/100391

BartekObudzinski:
All three fixed, plus the claims registry one, which was a single line since `measure.sh` can just default `AGENT_DEVICE_CLAIMS_DIR` inside its own state dir. Root cause was a merge from main on 20 August that resolved the flow files in favour of upstream and threw away this branch's own work, so the nine flows had lost their id anchored assertions, their waits and `return-to-inbox.ad`, all now restored with `find ... "click"` replaced by real ids. The duplicated nav is by design via `TabBarBottomContent`, and `press` has no way to pick among matches, so the three flows that must press a tab stay blocked on per tab `testID`s, and I have not run any of this on a device since I am on 0.20.6 where your stricter `press` error does not exist.

rlinoz:
@BartekObudzinski does it really need all those `\` ?

BartekObudzinski:
@rlinoz Mostly yes. Quoted arguments go through `JSON.parse`, so any selector containing a space needs the wrapper, and then its inner quotes need escaping. Without the backslashes those do not error, they parse into a different selector. I left them uniform because healing a selector means adding a `||` alternative, which on a bare line means adding the wrapper and escaping by hand

rlinoz (approved):

rlinoz:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

izarutskaya:
@BartekObudzinski Looks steps are for internal testing in Tests section, could you please share steps for QA team? Thanks

rlinoz:
This is a NO QA @izarutskaya sorry

PR #97559, opened 2026-07-31
Refactor Workspace menu data into a shared source

nabi-ebrahimi:
### Explanation of Change

Extracts the existing Workspace menu definitions from `WorkspaceInitialPage` into the shared `getWorkspaceMenuItems` helper.

The shared source owns the existing menu labels, icons, routes, ordering, permissions, feature gates, badges, RBRs, and highlight state. `WorkspaceInitialPage` continues to own hooks, policy loading, connection prefetching, rendering, and guarded navigation actions.

This is a behavior-preserving prerequisite for adding Workspace navigation suggestions in a follow-up PR. It does not add new Search Router suggestions or change the Workspace UI.

### Fixed Issues

### Tests

**Workspace menu remains unchanged**

1. Sign in and open any Workspace.
2. Review the Workspace menu and its available subpages.
3. Verify the same menu rows appear in the same order as before.

Expected: No Workspace menu item is added, removed, renamed, or reordered.

**Workspace navigation remains unchanged**

1. Open several available Workspace subpages, such as Overview, Members, Rooms, Reports, and Workflows.
2. Use the Workspace menu to move between them.

Expected: Every row opens its correct existing page, and the selected row remains highlighted.

**Permissions and feature visibility remain unchanged**

1. Open a Workspace with optional features enabled.
2. Verify its enabled feature pages appear.
3. Repeat with a limited member or a Workspace where some features are disabled.

Expected: Users only see the same pages allowed by their permissions and enabled Workspace features.

**Existing menu indicators remain unchanged**

1. Open a Workspace containing any badge, error indicator, or newly enabled feature highlight.
2. Review the related menu rows.

Expected: Existing icons, badges, error indicators, and highlights appear as before.

**Responsive layouts**

1. Open Workspace settings on narrow and wide layouts.
2. Navigate through several Workspace subpages.

Expected: The menu, header, scrolling, focus, and navigation behave as before, with no visible UI changes.

### Offline tests
Same as Tests.

### QA Steps
Same as Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/5ac53865-148b-4974-ad6f-a1d354803507

</details>

<details>

>

https://github.com/user-attachments/assets/546d85ba-6e6e-4778-b348-b8c4f1d048d9

</details>

<details>

>

https://github.com/user-attachments/assets/6e8d4893-0bff-49dd-a165-4f83f0a13e5b

</details>

<details>

>

https://github.com/user-attachments/assets/fba570a3-e723-40b0-b682-7866be27a235

</details>

<details>

>

https://github.com/user-attachments/assets/5b89a5c2-a65a-4ddf-972d-9c711d5dc147

https://github.com/user-attachments/assets/12046974-6b05-4520-abba-7b8df347535e

https://github.com/user-attachments/assets/f2a41577-7ac4-4d84-9e12-5074fd17d53d

https://github.com/user-attachments/assets/e4f20d02-6f22-4eee-8325-46235262b2ee

</details>


JmillsExpensify (commented):
No product review required.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,395 @@
+import {shouldShowQBOReimbursableExportDestinationAccountError} from '@libs/actions/connections/QuickbooksOnline';
```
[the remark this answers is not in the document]
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,398 @@
...
+    connectionInProgress = false,
+    policyCategories,
+    previousPendingFields,
+    shouldShowEnterCredentialsError = false,
+    shouldShowRBR = false,
+    isRulesRevampBetaEnabled = false,
+    isVendorMatchingBetaEnabled = false,
+    convertToDisplayString,
+}: GetWorkspaceMenuItemsParams): WorkspaceMenuItem[] {
+    const canReadPolicyFeature = (policyFeature: PolicyFeature) => canMemberRead(policy, currentUserLogin ?? '', policyFeature);
```
marufsharifi, line 109: **Routes are now built during render, so `Log.warn` fires on every render when `policyID` is missing**

Before this PR, `getRoute` ran inside the press handler. It now runs eagerly for every row, and each builder logs when the ID is missing — [`ROUTES.ts:2622`](https://github.com/Expensify/App/blob/main/src/ROUTES.ts#L2622):

```ts
if (!policyID) {
    Log.warn('Invalid policyID is used to build the WORKSPACE_OVERVIEW route');
}
```

`policyID` is derived from the policy, not the route params (`WorkspaceInitialPage.tsx:85-86`: `const policy = policyDraft?.id ? policyDraft : policyProp; const policyID = policy?.id;`), and the first three rows are pushed regardless of `policy`. `getWorkspaceMenuItems()` is also called unconditionally, before `FullPageNotFoundView` decides whether to render children.

**Bug this causes:** every workspace open logs 3 warnings on the first render pass (Onyx returns `undefined` before the policy read resolves), and a cold deeplink to `/workspaces/<id>/overview` or a not-found workspace keeps logging 3 warnings *per render* while mounted. These go to `console.debug` and into the buffer that uploads to the server at 500 lines ([`Log.ts:112`](https://github.com/Expensify/App/blob/main/src/libs/Log.ts#L112)) — console noise plus extra network traffic, and it contradicts the "no errors in the JS console" checkbox for the loading/deeplink case.

```suggestion
}: GetWorkspaceMenuItemsParams): WorkspaceMenuItem[] {
    // Every row's route is keyed on the policy ID, and ROUTES.*.getRoute() logs a warning when it is missing.
    // Return nothing until the policy has loaded instead of building invalid routes on every render.
    if (!policyID) {
        return [];
    }

    const canReadPolicyFeature = (policyFeature: PolicyFeature) => canMemberRead(policy, currentUserLogin ?? '', policyFeature);
```

If you'd rather keep the pre-PR render output byte-for-byte, the alternative is a lazy `getRoute: () => Route` on the item — but that's worse for the follow-up consumer, which wants concrete routes for suggestions.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,398 @@
...
+                route: ROUTES.POLICY_ACCOUNTING.getRoute(policyID),
+                brickRoadIndicator: hasSyncError || shouldShowQBOReimbursableExportDestinationAccountError(policy) ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : undefined,
+                screenName: SCREENS.WORKSPACE.ACCOUNTING.ROOT,
+                sentryLabel: CONST.SENTRY_LABEL.WORKSPACE.INITIAL.ACCOUNTING,
+                highlighted: highlightedPolicyFeature === CONST.POLICY.MORE_FEATURES.ARE_CONNECTIONS_ENABLED,
+            });
+        }
+
+        if (policyFeatureStates[CONST.POLICY.MORE_FEATURES.IS_HR_ENABLED] && canReadMoreFeatures) {
+            let hrBrickRoadIndicator;
```
marufsharifi: **NAB: `let hrBrickRoadIndicator;` is an implicit (evolving) `any`**

An un-annotated `let` accepts *any* future assignment. `hrBrickRoadIndicator = 'errror'` (typo) or `'warning'` would compile, and `HighlightableMenuItem` then renders **no indicator at all** — a silently missing red dot on the HR row, which is precisely the kind of thing manual QA doesn't catch. The `getHRBrickRoadIndicator()` closure this replaced had its type inferred from its `return`s, so it was safe.

`ValueOf` is already imported in this file, so this is a one-line fix:

```suggestion
            let hrBrickRoadIndicator: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS> | undefined;
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,398 @@
...
+    currentUserLogin?: string;
+    icons: WorkspaceMenuIconMap;
+    connectionInProgress?: boolean;
+    policyCategories?: OnyxTypes.PolicyCategories;
+    previousPendingFields?: OnyxTypes.Policy['pendingFields'];
+    shouldShowEnterCredentialsError?: boolean;
+    shouldShowRBR?: boolean;
+    isRulesRevampBetaEnabled?: boolean;
+    isVendorMatchingBetaEnabled?: boolean;
+    convertToDisplayString: (amount?: number, currency?: string) => string;
```
marufsharifi: **NAB: Hand-rolled signature duplicates an exported type**

The real contract already exists at [`CurrencyListContextProvider/types.ts:16`](https://github.com/Expensify/App/blob/main/src/components/CurrencyListContextProvider/types.ts#L16):

```ts
convertToDisplayString: (amount: number | undefined, currencyCode: string | undefined) => string;
```

**Bug this causes:** the underlying `CurrencyUtils.convertToDisplayString` already takes a third `shouldUseLocalCurrencySymbol` param. If the context signature is extended, this lookalike keeps compiling while the helper passes the wrong arity — an invoice-badge formatting regression with no type error to catch it. Per the type-consistency guideline, reuse the exported type:

```suggestion
    convertToDisplayString: CurrencyListActionsContextType['convertToDisplayString'];
```

Requires `import type {CurrencyListActionsContextType} from '@hooks/useCurrencyList';` (already exported there).

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,398 @@
...
+    | 'LuggageWithLines'
+    | 'ExpensifyCard'
+    | 'CreditCard'
+    | 'CalendarSolid'
+    | 'Clock'
+    | 'InvoiceGeneric'
+    | 'Gear'
+    | 'Bolt',
+    IconAsset
+>;
```
marufsharifi, line 66: **NAB: The icon-name list now exists in three places, and the keys aren't constrained to real icon names**

Injecting the icons is correct — they're lazily loaded per component via `useLazyAsset` — but the *name list* is duplicated: this union, the page's `useMemoizedLazyExpensifyIcons([...])` call (`WorkspaceInitialPage.tsx:92-115`), and a third time in the test (`getWorkspaceMenuItemsTest.ts:13-35`, 21 lines of `mockIcon`).

Also, because the union is raw string literals rather than `ExpensifyIconName`, a typo like `'Buliding'` type-checks here and only explodes at the call site.

**Bug this causes:** adding a row in the follow-up requires editing three separate lists; miss the page's list and the row renders with a `PlaceholderIcon` (blank icon), with nothing failing at compile time.

```suggestion
const WORKSPACE_MENU_ICON_NAMES = [
    'Building',
    'Users',
    'Hashtag',
    'Document',
    'Sync',
    'Receipt',
    'Briefcase',
    'Folder',
    'Tag',
    'Coins',
    'Workflows',
    'Feed',
    'Car',
    'LuggageWithLines',
    'ExpensifyCard',
    'CreditCard',
    'CalendarSolid',
    'Clock',
    'InvoiceGeneric',
    'Gear',
    'Bolt',
] as const satisfies readonly ExpensifyIconName[];

type WorkspaceMenuIconMap = Record<(typeof WORKSPACE_MENU_ICON_NAMES)[number], IconAsset>;
```

Then the page becomes `useMemoizedLazyExpensifyIcons([...WORKSPACE_MENU_ICON_NAMES, 'ExpensifyAppIcon'])` and the test imports the same const.

Stronger option if you want this genuinely reusable: return `iconName: ExpensifyIconName` and let each consumer resolve icons with its own lazy hook — then the second consumer needs zero icon plumbing and the test needs no icon mock at all.

src/pages/workspace/WorkspaceInitialPage.tsx
```diff
@@ -264,242 +166,26 @@ function WorkspaceInitialPage({policyDraft, policy: policyProp, route}: Workspac
     // isMatchingVendorListLoaded), which is empty on a non-active workspace until a page
     // requiring connections is opened. Prefetch it here, gated on read-access + beta so this
     // doesn't fire an accounting-page read on every workspace visit.
-    const canReadVendors = canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.VENDORS);
+    const canReadVendors = canMemberRead(policy, currentUserLogin ?? '', CONST.POLICY.POLICY_FEATURE.VENDORS);
```
marufsharifi, line 168: **NAB: The Vendors gate now has two owners in two files**

Prefetch gate here:
```ts
canMemberRead(policy, currentUserLogin ?? '', CONST.POLICY.POLICY_FEATURE.VENDORS) && isBetaEnabled(CONST.BETAS.VENDOR_MATCHING)
```
Row gate in `getWorkspaceMenuItems.ts:247`:
```ts
canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.VENDORS) && hasVendorFeature(policy, isVendorMatchingBetaEnabled) && isMatchingVendorListLoaded(policy)
```

Before this PR both derived from the same local `canReadVendors`, so they couldn't drift. The prefetch exists *specifically* to make the row's `isMatchingVendorListLoaded(policy)` check pass — as the comment right above states.

**Bug this causes:** change the read permission or the beta on one side only and you reintroduce exactly the bug that comment warns about — the Vendors row never appears on a non-active workspace because connections were never fetched, with nothing failing at compile time. Suggest exporting the predicate from the helper and using it in both places:

```ts
// getWorkspaceMenuItems.ts
function canPrefetchVendorConnections(policy: OnyxTypes.Policy | undefined, currentUserLogin: string | undefined, isVendorMatchingBetaEnabled: boolean) {
    return canMemberRead(policy, currentUserLogin ?? '', CONST.POLICY.POLICY_FEATURE.VENDORS) && isVendorMatchingBetaEnabled;
}
export {canPrefetchVendorConnections};
```

tests/unit/getWorkspaceMenuItemsTest.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        expect(items.map((item) => item.translationKey)).toEqual(['workspace.common.profile', 'workspace.common.members', 'workspace.common.rooms']);
+        expect(items.map((item) => item.route)).toEqual([
+            ROUTES.WORKSPACE_OVERVIEW.getRoute(policy.id),
+            ROUTES.WORKSPACE_MEMBERS.getRoute(policy.id),
+            ROUTES.WORKSPACE_ROOMS.getRoute(policy.id),
+        ]);
+    });
+
+    it('preserves the complete enabled Workspace menu order and presentation data', () => {
```
marufsharifi, line 179: **This test name overstates what it asserts — Vendors is never produced**

The Vendors row needs `hasVendorFeature(policy, isVendorMatchingBetaEnabled) && isMatchingVendorListLoaded(policy)`, and neither is satisfied by this fixture (`isVendorMatchingBetaEnabled` isn't passed, so it defaults to `false`). So the expected list on line 100 is not the "complete enabled Workspace menu" — it's the complete menu *minus Vendors*.

**Why it matters:** this is the one test guarding row order for a refactor whose entire claim is "order preserved". If someone later inserts a row between Categories and Tags, this test catches it; if they break the Vendors row's position, nothing does. Either rename it to reflect the fixture (`...with vendor matching disabled`) or pass `isVendorMatchingBetaEnabled: true` with a policy whose vendor list is loaded, and add `'workspace.common.vendors'` to the expected array in its real position (after Categories, before Tags).

tests/unit/getWorkspaceMenuItemsTest.ts
```diff
@@ -0,0 +1,126 @@
...
+            'common.perDiem',
+            'iou.time',
+            'workspace.common.invoices',
+            'workspace.common.moreFeatures',
+        ]);
+        expect(items.find((item) => item.translationKey === 'workspace.common.rules')?.icon).toBe(icons.Bolt);
+        expect(items.find((item) => item.translationKey === 'workspace.common.invoices')?.badgeText).toBe('$1.23');
+        expect(convertToDisplayString).toHaveBeenCalledWith(123, policy.outputCurrency);
+    });
+});
```
marufsharifi, line 290: **Coverage gaps: ~80–85 % statements but roughly 40–50 % branches, and 0 % of the page wiring**

Both `push` blocks execute, so statement coverage looks healthy — but nearly every condition inside them is untested:

| Untested | Consequence if it regresses |
|---|---|
| All 7 `brickRoadIndicator` error sources | A dropped RBR is invisible until a customer reports a missed error |
| HR `INFO` indicator (`isMergeHRCompleteSetupNeeded`) | Only the `undefined` branch runs today |
| `highlighted` / `previousPendingFields` logic | The most subtle extracted logic — zero coverage |
| `isGroupPolicy === false` (personal policy) | Guards the entire protected block |
| `screenName` values | Feeds `canAccessRoute`; a wrong value silently closes the RHP |
| `isRulesRevampBetaEnabled: false` → `Feed` icon | Only the `Bolt` branch is asserted |

For a behaviour-preserving refactor the highest-value additions are a highlight test, one RBR test, and a full-shape snapshot so future drift shows up as a diff rather than as a missing row:

```ts
    it('marks the most recently enabled feature as highlighted', () => {
        const policy = createMock<Policy>({
            ...buildPolicy(CONST.POLICY.ROLE.ADMIN),
            areCategoriesEnabled: true,
            pendingFields: {areCategoriesEnabled: CONST.RED_BRICK_ROAD_PENDING_ACTION.UPDATE},
        });

        const items = getWorkspaceMenuItems({policy, policyID: policy.id, currentUserLogin, icons, previousPendingFields: {}, convertToDisplayString: () => ''});

        expect(items.find((item) => item.translationKey === 'workspace.common.categories')?.highlighted).toBe(true);
        expect(items.filter((item) => item.highlighted)).toHaveLength(1);
    });

    it('flags a red brick road on the members row when the employee list has errors', () => {
        const policy = createMock<Policy>({
            ...buildPolicy(CONST.POLICY.ROLE.ADMIN),
            employeeList: {[currentUserLogin]: {role: CONST.POLICY.ROLE.ADMIN, errors: {error: 'Whoops'}}},
        });

        const items = getWorkspaceMenuItems({policy, policyID: policy.id, currentUserLogin, icons, convertToDisplayString: () => ''});

        expect(items.find((item) => item.translationKey === 'workspace.common.members')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
    });

    it('returns only the base rows for a personal policy', () => {
        const policy = createMock<Policy>({...buildPolicy(CONST.POLICY.ROLE.ADMIN), type: CONST.POLICY.TYPE.PERSONAL});

        const items = getWorkspaceMenuItems({policy, policyID: policy.id, currentUserLogin, icons, convertToDisplayString: () => ''});

        expect(items).toHaveLength(3);
    });

    it('returns stable row metadata', () => {
        // guards translationKey + route + screenName + sentryLabel together
        expect(items.map(({translationKey, route, screenName, sentryLabel}) => ({translationKey, route, screenName, sentryLabel}))).toMatchSnapshot();
    });
});
```

Nothing here or elsewhere covers the page side (`.map()` → `Navigation.navigate(item.route)`, and `canAccessRoute` matching on `screenName`) — which is where the eager-`getRoute` regression I flagged on `getWorkspaceMenuItems.ts:106` lives.

marufsharifi (changes requested):

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: On `main`, the page owns the icon-name list and passes the resolved IconAsset values directly into menu rows. Our refactor preserves that runtime behavior: `WorkspaceInitialPage` still owns `useMemoizedLazyExpensifyIcons()`, and the extracted code receives those same resolved icons.

src/pages/workspace/WorkspaceInitialPage.tsx
nabi-ebrahimi: Addressed.

tests/unit/getWorkspaceMenuItemsTest.ts
nabi-ebrahimi: Addressed.

tests/unit/getWorkspaceMenuItemsTest.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,401 @@
...
+import SCREENS from '@src/SCREENS';
+import type * as OnyxTypes from '@src/types/onyx';
+import type {PolicyFeatureName} from '@src/types/onyx/Policy';
+import {isEmptyObject} from '@src/types/utils/EmptyObject';
+import type IconAsset from '@src/types/utils/IconAsset';
+import type WithSentryLabel from '@src/types/utils/SentryLabel';
+
+import type {ValueOf} from 'type-fest';
+
+type WorkspaceMenuIconMap = Record<
```
marufsharifi, line 43: The 21-key icon list is now hardcoded in two places — here and in the `useMemoizedLazyExpensifyIcons([...])` call in `WorkspaceInitialPage.tsx`. Injecting icons is the right call (it preserves lazy icon loading and keeps this helper unit-testable), but the intent isn't obvious, and a future contributor could "simplify" it by importing `Expensicons` directly and silently regress startup TTI.

TypeScript guards a *missing* icon structurally, but please add a comment explaining why injection is used:

```suggestion
// Icons are injected (not imported from Expensicons) so the page keeps lazy-loading
// them via useMemoizedLazyExpensifyIcons, and so this helper stays unit-testable.
type WorkspaceMenuIconMap = Record
```

tests/unit/getWorkspaceMenuItemsTest.ts
```diff
@@ -0,0 +1,290 @@
...
+            convertToDisplayString: () => '',
+        });
+
+        expect(items.find((item) => item.translationKey === 'workspace.common.hr')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.INFO);
+    });
+
+    it('uses the existing Rules icon when the Rules revamp beta is disabled', () => {
+        const policy = createMock<Policy>({...buildPolicy(CONST.POLICY.ROLE.ADMIN), areRulesEnabled: true});
+
+        const items = getWorkspaceMenuItems({
```
marufsharifi, line 168: `canReadVendors` is recomputed here for the prefetch gate, while the helper independently computes the same value via `canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.VENDORS)` for the Vendors row. Same result today, but the vendor read-access rule now lives in two places and can drift.

Not blocking — but consider computing it once and passing it into `getWorkspaceMenuItems` (e.g. a `canReadVendors` param), so both the prefetch and the menu row read a single source of truth.

src/pages/workspace/WorkspaceInitialPage.tsx
```diff
@@ -263,242 +165,26 @@ function WorkspaceInitialPage({policyDraft, policy: policyProp, route}: Workspac
...
+        policyCategories,
+        previousPendingFields: prevPendingFields,
+        shouldShowEnterCredentialsError,
+        shouldShowRBR,
+        isRulesRevampBetaEnabled: isBetaEnabled(CONST.BETAS.RULES_REVAMP),
+        isVendorMatchingBetaEnabled: isBetaEnabled(CONST.BETAS.VENDOR_MATCHING),
+        convertToDisplayString,
+    }).map((item) => ({
+        ...item,
+        action: singleExecution(waitForNavigate(() => Navigation.navigate(item.getRoute()))),
```
marufsharifi, line 186: Just flagging that this lazy wiring is load-bearing and should stay as-is. The helper returns `getRoute: () => ROUTES.X.getRoute(policyID)` and it's only invoked here on click. Building the route eagerly (e.g. `getRoute(undefined)` during creation/loading) triggers a `Log.warn`, which the `does not build routes until a menu item is selected` test guards against. Please keep `item.getRoute()` inside the `action` callback rather than hoisting it into the map body.

marufsharifi (changes requested):

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

tests/unit/getWorkspaceMenuItemsTest.ts
nabi-ebrahimi: Both places reuse the same centralized `canMemberRead()` rule for different purposes: prefetching and menu visibility. Passing it in would unnecessarily couple the shared menu helper to page-specific logic.

src/pages/workspace/WorkspaceInitialPage.tsx
nabi-ebrahimi: Confirmed, `item.getRoute()` remains inside the action callback, and the focused test protects this lazy behavior.

nabi-ebrahimi:
@marufsharifi resolved conflicts, thanks.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,402 @@
...
+
+type WorkspaceMenuItem = WithSentryLabel & {
+    translationKey: TranslationPaths;
+    icon: IconAsset;
+    getRoute: () => Route;
+    brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>;
+    screenName: WorkspaceTopLevelScreens;
+    badgeText?: string;
+    highlighted?: boolean;
+};
```
marufsharifi, line 84: @nabi-ebrahimi, could you please add docs for these. thanks.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,402 @@
...
+    currentUserLogin?: string;
+    icons: WorkspaceMenuIconMap;
+    connectionInProgress?: boolean;
+    policyCategories?: OnyxTypes.PolicyCategories;
+    previousPendingFields?: OnyxTypes.Policy['pendingFields'];
+    shouldShowEnterCredentialsError?: boolean;
+    shouldShowRBR?: boolean;
+    isRulesRevampBetaEnabled?: boolean;
+    isVendorMatchingBetaEnabled?: boolean;
+    convertToDisplayString: CurrencyListActionsContextType['convertToDisplayString'];
```
marufsharifi, line 111: @nabi-ebrahimi also here.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,402 @@
...
+    | 'LuggageWithLines'
+    | 'ExpensifyCard'
+    | 'CreditCard'
+    | 'CalendarSolid'
+    | 'Clock'
+    | 'InvoiceGeneric'
+    | 'Gear'
+    | 'Bolt',
+    IconAsset
+>;
```
marufsharifi: **NAB:** `WorkspaceMenuIconMap` hard-codes a 21-name icon union. Every future row that needs a new icon forces edits in **two** places — the `icon: icons.X` usage *and* this union — or you get a compile error pointing at the type instead of the real edit site. The union just re-encodes what the `icons.X` accesses already prove.

The injection itself is correct (the page supplies icons via `useMemoizedLazyExpensifyIcons(...)`, so this keeps lazy-loading + testability). Only the shape is heavier than it needs to be. Consider relaxing it to a subset of Expensicons so there's a single source of truth:

```ts
import type * as Expensicons from '@components/Icon/Expensicons';

// Icons are injected so the page keeps lazy-loading them and this helper stays unit-testable.
type WorkspaceMenuIconMap = Partial<Record<keyof typeof Expensicons, IconAsset>>;
```

The `icon: icons.X` accesses still enforce that each required icon is present, so nothing loosens in practice.

tests/unit/getWorkspaceMenuItemsTest.ts
```diff
@@ -0,0 +1,290 @@
...
+            convertToDisplayString: () => '',
+        });
+
+        expect(items).toHaveLength(3);
+        expect(warnSpy).not.toHaveBeenCalled();
+
+        items.at(0)?.getRoute();
+        expect(warnSpy).toHaveBeenCalledTimes(1);
+        warnSpy.mockRestore();
+    });
```
marufsharifi, line 67: This test proves laziness only as a *side effect* of `ROUTES.WORKSPACE_OVERVIEW.getRoute(undefined)` happening to call `Log.warn`. If `ROUTES` ever stops warning on an undefined `policyID`, this test fails even though `getWorkspaceMenuItems` is still correct — a false failure that sends the next dev on a wrong hunt. Assert the actual property under test (route builder not invoked until `getRoute()` is called):

```suggestion
    it('does not build routes until a menu item is selected', () => {
        const getRouteSpy = jest.spyOn(ROUTES.WORKSPACE_OVERVIEW, 'getRoute');

        const items = getWorkspaceMenuItems({
            policy: undefined,
            policyID: undefined,
            icons,
            convertToDisplayString: () => '',
        });

        expect(items).toHaveLength(3);
        expect(getRouteSpy).not.toHaveBeenCalled();

        items.at(0)?.getRoute();
        expect(getRouteSpy).toHaveBeenCalledTimes(1);
        getRouteSpy.mockRestore();
    });
```

*(This drops the last use of `Log`, so also remove `import Log from '@libs/Log';` on line 1.)*

tests/unit/getWorkspaceMenuItemsTest.ts
```diff
@@ -0,0 +1,290 @@
...
+        role,
+        employeeList: {
+            [currentUserLogin]: {role},
+        },
+        pendingAction: undefined,
+        errorFields: {},
+    });
+}
+
+describe('getWorkspaceMenuItems', () => {
```
marufsharifi, line 50: The suite locks down ordering/visibility well, but most `brickRoadIndicator` branches are untested. These branches all live in this diff and are exactly the "indicators unchanged" promise from the PR's test plan — a future edit could silently drop one and every test still passes:

- Accounting — `hasSyncError || shouldShowQBOReimbursableExportDestinationAccountError` (`getWorkspaceMenuItems.ts:208`)
- Categories — `hasPolicyCategoriesError` (`:244`)
- Taxes — `shouldShowTaxRateError` (`:277`)
- Workflows — `errorFields.reimburser` (`:289`)
- Rules — `hasPolicyRulesError` (`:301`)
- Receipt partners — `shouldShowEnterCredentialsError` **(new pass-through param)** (`:232`)
- Company cards — `shouldShowRBR` **(new pass-through param)** (`:346`)

At minimum please cover the two new pass-through params, since they're the indirection this PR introduces. e.g.:

```ts
it('flags company cards when shouldShowRBR is set', () => {
    const policy = createMock<Policy>({...buildPolicy(CONST.POLICY.ROLE.ADMIN), areCompanyCardsEnabled: true});
    const items = getWorkspaceMenuItems({policy, policyID: policy.id, currentUserLogin, icons, shouldShowRBR: true, convertToDisplayString: () => ''});
    expect(items.find((i) => i.translationKey === 'workspace.common.companyCards')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
});
```

marufsharifi:
@nabi-ebrahimi, could you please check the above comments. thanks.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Address.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: I’ll keep the exact `Record` because it guarantees every required icon is provided. Using `Partial` would make icon values optional and weaken this compile-time safety.

tests/unit/getWorkspaceMenuItemsTest.ts
nabi-ebrahimi: Addressed.

tests/unit/getWorkspaceMenuItemsTest.ts
nabi-ebrahimi: Addressed.

nabi-ebrahimi:
> @nabi-ebrahimi, could you please check the above comments. thanks.

@marufsharifi thanks for the review. All actionable comments have been addressed. Could you please review again?

trjExpensify:
Same here, @marufsharifi. Please complete your re-review today!

marufsharifi:
I'am testing this, i will finish in a few hours. thanks.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+import {isEmptyObject} from '@src/types/utils/EmptyObject';
+import type IconAsset from '@src/types/utils/IconAsset';
+import type WithSentryLabel from '@src/types/utils/SentryLabel';
+
+import type {ValueOf} from 'type-fest';
+
+/**
+ * Icons used by Workspace menu items. They are injected instead of imported from Expensicons so the page keeps lazy-loading them and this helper remains unit-testable.
+ */
+type WorkspaceMenuIconMap = Record<
```
marufsharifi, line 47: **Maintainability — hand-maintained icon union.** `WorkspaceMenuIconMap` is a manual union of 21 icon names. Every future menu-icon add/rename must touch this union *and* the item that uses it, or the caller silently passes the wrong shape. The type errors on a *missing* key (nice safety net), so this is an acceptable tradeoff for keeping the page's lazy-loading + this helper's testability — the comment above already explains the injection rationale. Optional: derive it to avoid drift, at the cost of re-coupling to Expensicons:

```ts
type WorkspaceMenuIconMap = Pick<typeof Expensicons, 'Building' | 'Users' | /* … */ 'Bolt'>;
```

If we keep the explicit union, a `// keep in sync with the icons referenced below` note would help the next maintainer.

src/pages/workspace/WorkspaceInitialPage.tsx
```diff
@@ -265,242 +167,26 @@ function WorkspaceInitialPage({policyDraft, policy: policyProp, route}: Workspac
     // narrowed to vendor-capable workspaces because that answer lives in the very data being
     // fetched. The hook already skips the fetch when the app is offline, when the workspace has
     // no accounting connection, and when the data has already been fetched.
-    const canReadVendors = canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.VENDORS);
+    const canReadVendors = canMemberRead(policy, currentUserLogin ?? '', CONST.POLICY.POLICY_FEATURE.VENDORS);
```
marufsharifi, line 170: **NAB:** `canReadVendors` is computed here for `usePolicyConnectionsPrefetch`, and `getWorkspaceMenuItems` independently recomputes the same `canMemberRead(policy, currentUserLogin ?? '', VENDORS)` internally for the Vendors row. Same result → not a bug, just duplicated across the boundary. Passing it into the helper would leak an implementation detail, so leaving it is defensible — flagging for awareness.

tests/unit/getWorkspaceMenuItemsTest.ts
```diff
@@ -0,0 +1,318 @@
...
+            currentUserLogin,
+            icons,
+            shouldShowEnterCredentialsError: true,
+            convertToDisplayString: () => '',
+        });
+
+        expect(items.find((item) => item.translationKey === 'workspace.common.receiptPartners')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
+    });
+
+    it('shows an error indicator when company cards require attention', () => {
```
marufsharifi, line 200: **Test coverage gap — untested brick-road branches.** Ordering/visibility/highlight and 4 indicator branches (members, receipt partners, company cards, Merge-HR info) are covered well. The remaining RBR branches are untested and are the most regression-prone part of the helper:
- `hasGeneralSettingsError` → Profile RBR (errorFields `name`/`avatarURL`/`outputCurrency`/`address`)
- `hasSyncError` **or** `shouldShowQBOReimbursableExportDestinationAccountError` → Accounting RBR
- `shouldShowTaxRateError` → Taxes RBR
- `hasPolicyRulesError` → Rules RBR
- workflows `errorFields.reimburser` → Workflows RBR
- `hasPolicyCategoriesError(policyCategories)` → Categories RBR (also exercises the `policyCategories` param)

Each is a one-liner mirroring this `company cards` test, e.g.:

```ts
it('shows an error indicator when workflows have a reimburser error', () => {
    const policy = createMock<Policy>({...buildPolicy(CONST.POLICY.ROLE.ADMIN), areWorkflowsEnabled: true, errorFields: {reimburser: {error: 'Whoops'}}});
    const items = getWorkspaceMenuItems({policy, policyID: policy.id, currentUserLogin, icons, convertToDisplayString: () => ''});
    expect(items.find((i) => i.translationKey === 'workspace.common.workflows')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
});
```

Not a blocker — cheap to lock down.

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+        if (hasHRError) {
+            return CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR;
+        }
+        if (isMergeHRCompleteSetupNeeded(policy)) {
+            return CONST.BRICK_ROAD_INDICATOR_STATUS.INFO;
+        }
+    };
+    const hasMembersError = shouldShowEmployeeListError(policy);
+    const hasPolicyCategoryError = hasPolicyCategoriesError(policyCategories);
+    const hasGeneralSettingsError =
```
marufsharifi, line 158: **Missing failure-path coverage: `hasGeneralSettingsError` (profile row).**

No test exercises this 4-way OR — none of `errorFields.{name,avatarURL,outputCurrency,address}` is ever set, so the profile `brickRoadIndicator` branch is untested.

Please add to `tests/unit/getWorkspaceMenuItemsTest.ts`:
```ts
it('shows an error indicator on the profile row when general settings have errors', () => {
    const policy = createMock<Policy>({...buildPolicy(CONST.POLICY.ROLE.ADMIN), errorFields: {name: {error: 'Whoops'}}});
    const items = getWorkspaceMenuItems({policy, policyID: policy.id, currentUserLogin, icons, convertToDisplayString: () => ''});
    expect(items.find((item) => item.translationKey === 'workspace.common.profile')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
});
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+                sentryLabel: CONST.SENTRY_LABEL.WORKSPACE.INITIAL.REPORTS,
+            });
+        }
+
+        if (policyFeatureStates[CONST.POLICY.MORE_FEATURES.ARE_CONNECTIONS_ENABLED] && canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.ACCOUNTING)) {
+            items.push({
+                translationKey: 'workspace.common.accounting',
+                icon: icons.Sync,
+                getRoute: () => ROUTES.POLICY_ACCOUNTING.getRoute(policyID),
+                brickRoadIndicator: hasSyncError || shouldShowQBOReimbursableExportDestinationAccountError(policy) ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : undefined,
```
marufsharifi, line 226: **Missing failure-path coverage: accounting sync-error indicator.**

The accounting row appears in the "complete menu" test but its `brickRoadIndicator` is never asserted, and the `connectionInProgress` param has **no** effect exercised anywhere in the suite.

Please add a test that gives the policy a syncing accounting connection in an error state (or passes `connectionInProgress: true` with an errored connection) and asserts:
```ts
expect(items.find((item) => item.translationKey === 'workspace.common.accounting')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+        CONST.POLICY.POLICY_FEATURE.EXPENSIFY_CARD,
+        CONST.POLICY.POLICY_FEATURE.COMPANY_CARDS,
+        CONST.POLICY.POLICY_FEATURE.PER_DIEM,
+        CONST.POLICY.POLICY_FEATURE.MORE_FEATURES,
+    ].some(canReadPolicyFeature);
+
+    const accountingConnectionNames = CONST.POLICY.CONNECTIONS.ACCOUNTING_CONNECTION_NAMES;
+    const hasSyncError = shouldShowSyncError(policy, connectionInProgress, accountingConnectionNames);
+    const hasHRError = shouldShowHRConnectionError(policy, connectionInProgress, isPolicyAdmin(policy));
+    const getHRBrickRoadIndicator = () => {
```
marufsharifi, line 148: **Partial coverage: `getHRBrickRoadIndicator` — only the INFO branch is tested.**

"shows an information indicator when Merge HR setup is incomplete" covers INFO, but the **ERROR** branch (`shouldShowHRConnectionError` → true) and the `undefined` fallthrough are untested.

Please extend the HR tests with an ERROR case — admin + a connected HR provider whose `lastSync.isAuthenticationError` is `true`:
```ts
expect(items.find((item) => item.translationKey === 'workspace.common.hr')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+                highlighted: highlightedPolicyFeature === CONST.POLICY.MORE_FEATURES.ARE_RECEIPT_PARTNERS_ENABLED,
+            });
+        }
+
+        if (policyFeatureStates[CONST.POLICY.MORE_FEATURES.ARE_CATEGORIES_ENABLED] && canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.CATEGORIES)) {
+            items.push({
+                translationKey: 'workspace.common.categories',
+                icon: icons.Folder,
+                getRoute: () => ROUTES.WORKSPACE_CATEGORIES.getRoute(policyID),
+                brickRoadIndicator: hasPolicyCategoryError ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : undefined,
```
marufsharifi, line 262: **Missing failure-path coverage: categories `hasPolicyCategoriesError(policyCategories)`.**

The `policyCategories` param is never passed with an errored category, so this branch is dead in tests.

Please add a test passing errored categories and asserting the indicator:
```ts
const items = getWorkspaceMenuItems({
    policy, policyID: policy.id, currentUserLogin, icons,
    policyCategories: {Food: {name: 'Food', enabled: true, errors: {err: 'Whoops'}}},
    convertToDisplayString: () => '',
});
expect(items.find((item) => item.translationKey === 'workspace.common.categories')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+                highlighted: highlightedPolicyFeature === CONST.POLICY.MORE_FEATURES.ARE_TAGS_ENABLED,
+            });
+        }
+
+        if (policyFeatureStates[CONST.POLICY.MORE_FEATURES.ARE_TAXES_ENABLED] && canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.TAXES)) {
+            items.push({
+                translationKey: 'workspace.common.taxes',
+                icon: icons.Coins,
+                getRoute: () => ROUTES.WORKSPACE_TAXES.getRoute(policyID),
+                brickRoadIndicator: shouldShowTaxRateError(policy) ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : undefined,
```
marufsharifi, line 295: **Missing failure-path coverage: taxes `shouldShowTaxRateError`.**

Please add a test with an admin policy that has taxes enabled (`tax.trackingEnabled: true`) and an errored tax rate (`taxRates.taxes.{...}.errors`), asserting:
```ts
expect(items.find((item) => item.translationKey === 'workspace.common.taxes')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+                highlighted: highlightedPolicyFeature === CONST.POLICY.MORE_FEATURES.ARE_TAXES_ENABLED,
+            });
+        }
+
+        if (policyFeatureStates[CONST.POLICY.MORE_FEATURES.ARE_WORKFLOWS_ENABLED] && canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.WORKFLOWS)) {
+            items.push({
+                translationKey: 'workspace.common.workflows',
+                icon: icons.Workflows,
+                getRoute: () => ROUTES.WORKSPACE_WORKFLOWS.getRoute(policyID),
+                brickRoadIndicator: !isEmptyObject(policy?.errorFields?.reimburser ?? {}) ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : undefined,
```
marufsharifi, line 307: **Missing failure-path coverage: workflows reimburser-error indicator.**

Please add a test with workflows enabled and `errorFields: {reimburser: {error: 'Whoops'}}`, asserting:
```ts
expect(items.find((item) => item.translationKey === 'workspace.common.workflows')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+                highlighted: highlightedPolicyFeature === CONST.POLICY.MORE_FEATURES.ARE_WORKFLOWS_ENABLED,
+            });
+        }
+
+        if (policyFeatureStates[CONST.POLICY.MORE_FEATURES.ARE_RULES_ENABLED] && canReadPolicyFeature(CONST.POLICY.POLICY_FEATURE.RULES)) {
+            items.push({
+                translationKey: 'workspace.common.rules',
+                icon: isRulesRevampBetaEnabled ? icons.Bolt : icons.Feed,
+                getRoute: () => ROUTES.WORKSPACE_RULES.getRoute(policyID),
+                brickRoadIndicator: hasPolicyRulesError(policy) ? CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR : undefined,
```
marufsharifi, line 319: **Missing failure-path coverage: rules `hasPolicyRulesError`.**

"uses the existing Rules icon…" asserts the icon but not the error indicator. Please add a test that makes `hasPolicyRulesError(policy)` true with rules enabled, asserting:
```ts
expect(items.find((item) => item.translationKey === 'workspace.common.rules')?.brickRoadIndicator).toBe(CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR);
```

src/pages/workspace/getWorkspaceMenuItems.ts
```diff
@@ -0,0 +1,419 @@
...
+    /** Workspace policy used to determine menu visibility and indicators. */
+    policy: OnyxTypes.Policy | undefined;
+    /** Workspace policy ID used to build destination routes. */
+    policyID: string | undefined;
+    /** Login used to determine the member's feature access. */
+    currentUserLogin?: string;
+    /** Lazily loaded icons used by the menu items. */
+    icons: WorkspaceMenuIconMap;
+    /** Whether an accounting connection is currently syncing. */
+    connectionInProgress?: boolean;
```
marufsharifi: **Naming — boolean must be prefixed with `is`/`should`**

Per [`STYLE.md` → Boolean variables and props](contributingGuides/STYLE.md), boolean props/variables must be prefixed with `should` or `is`. `connectionInProgress` is a boolean introduced in this PR with no prefix — let's rename it to `isConnectionInProgress`.

```suggestion
    isConnectionInProgress?: boolean;
```

The rename touches 5 spots, so the suggestion above only covers this line — please also update:
- `getWorkspaceMenuItems.ts:118` — destructured default (`connectionInProgress = false,`)
- `getWorkspaceMenuItems.ts:145` — `shouldShowSyncError(policy, connectionInProgress, …)`
- `getWorkspaceMenuItems.ts:146` — `shouldShowHRConnectionError(policy, connectionInProgress, …)`
- `WorkspaceInitialPage.tsx:178` — call site (`connectionInProgress: isConnectionInProgress(connectionSyncProgress, policy),`)

marufsharifi (changes requested):
Tested works perfect, but some minor suggestions.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/WorkspaceInitialPage.tsx
nabi-ebrahimi: NAB and intentionally unchanged because prefetching and menu visibility are separate concerns.

tests/unit/getWorkspaceMenuItemsTest.ts
nabi-ebrahimi: The new pass-through parameters introduced by this refactor are already covered.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/workspace/getWorkspaceMenuItems.ts
nabi-ebrahimi: Addressed.

nabi-ebrahimi:
@marufsharifi thanks for the detailed review. All actionable comments have been addressed. Could you please review again?

marufsharifi (approved):
LGTM

carlosmiceli (approved):

Issue #95343, opened 2026-07-05
[Due for payment 2026-08-19] [$250] [Sentry: APP-GNR] removeChild crash in NativeSafeAreaProvider cleanup on Search


mountiny:
## Sentry
https://expensify.sentry.io/issues/APP-GNR

## Impact (snapshot at filing)
- Users (total since first seen): 2
- Events: 2
- Users (last 14d): 2
- First seen: 2026-06-30
- Last seen: 2026-07-05 (ongoing)
- Platform: web
- App version(s): 9.4.26-9
- Affected route(s) (parameterized): /search
- Mechanism: auto.browser.global_handlers.onerror

## Stack trace (top frames, first-party only)
```
NativeSafeAreaProvider.web.js:62 (destroy_) → document.body.removeChild(element)
react-dom-client.production.js → commitHookEffectListUnmount → passive unmount chain
```

## Suspected cause
`react-native-safe-area-context` web provider cleanup calls `document.body.removeChild(element)` after the element was already removed during Search navigation unmount.

## Reproduction
Unknown — see Sentry events linked from the Sentry issue. Occurred on `/search` with `distance-odometer` view active.

## Related
- Sibling Sentry IDs (if any): none
- Prior GH issues: none

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: <a href="https://www.upwork.com/jobs/~022077756078585815489">https://www.upwork.com/jobs/~022077756078585815489</a></li>
        <li>Upwork Job ID: 2077756078585815489</li>
        <li>Last Price Increase: 2026-07-23</li>
    </ul>
</details>

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

mallenexpensify:
### Payment Summary
Contributor: @wildan-m  paid $250 via Upwork
Contributor+: @gijoe0295  due $250 via NewDot

JmillsExpensify:
$250 approved for @gijoe0295

PR #97595, opened 2026-07-31
Prevent uncaught removeChild error in safe area provider cleanup

wildan-m:
### Explanation of Change

Navigating on Search intermittently throws an uncaught `NotFoundError: Failed to execute 'removeChild' on 'Node'` on web, reported to Sentry as APP-GNR. The safe area library's web provider appends a hidden measurement element to the page body on mount, and its cleanup removes it by asking the body to remove it. That call throws when the element is no longer a child of the body, which is what happens when the provider's subtree is torn down with the element already detached. The throw lands in React's passive-unmount chain where nothing catches it, so it surfaces as an uncaught error.

The patch makes the cleanup tolerate that state: it detaches the transition listener first, then removes the element through the element itself, which does nothing when the node has no parent. The attached case is unchanged, so only the previously-crashing path behaves differently. The unguarded removal is still on the library's `main`, so this stays a patch rather than a version bump.

### Fixed Issues

$ https://github.com/Expensify/App/issues/95343
PROPOSAL: https://github.com/Expensify/App/issues/95343#issuecomment-4992812017

### Tests

**A UI reproduction is not possible** — the provider mounts once at the app root and never unmounts, so its cleanup never runs on its own. Steps 1–2 invoke that cleanup directly; step 3 is a regression check. Everything here creates the failing state artificially, so **the definitive confirmation is Sentry after deploy** (see below).

1. **On `main`** — `npm install && rm -rf node_modules/.cache/rspack && npm run web`. Sign in, open Search (`/search`), open the console (⌘⌥J) and paste the script below. It runs the real cleanup twice — attached, which must succeed, then detached, which is the race — and prints:
   ```
   removal call : document.body.removeChild(element)   [unpatched]
   control  (element attached) : no error
   crash    (element detached) : NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
   VERDICT: BUG REPRODUCED
   ```
2. **On this branch** — same command, then reload, open Search and paste the same script:
   ```
   removal call : element.remove()                     [patched]
   control  (element attached) : no error
   crash    (element detached) : no error
   VERDICT: FIXED
   ```
3. **Regression check** — on this branch, navigate between a few Search views, open and close an expense in the RHP, and go back to Inbox. The layout keeps its usual spacing and no errors appear in the console.

> Clearing `node_modules/.cache/rspack` on each branch is required: rspack treats `node_modules` as immutable and its persistent cache is only invalidated by the rsbuild config, so without it both branches report the same verdict. The `removal call` line is the tell.

Reading React internals makes this a dev-build-only check, so run it against a local dev build rather than staging or production.

<details>
<summary>Console script</summary>

```js
/* #95343 — invokes the provider's REAL effect cleanup via React's fiber tree.
 * Runs it twice: attached (must succeed), then detached (the teardown race).
 * Reload the page afterwards — this tears the provider's effect down. */
(() => {
    const RULE = '──────────────────────────────────────────────────────────────────';
    const log = (s) => console.log(s === undefined ? '' : s);

    const root = document.getElementById('root');
    const key = root && Object.keys(root).find((k) => k.startsWith('__reactContainer$'));
    if (!key) {
        log('Could not find the React container on #root — is the app loaded?');
        return;
    }

    // find the provider's fiber
    let target = null;
    const seen = new Set();
    (function walk(f) {
        while (f && !target) {
            if (seen.has(f)) { return; }
            seen.add(f);
            if (typeof f.type === 'function' && f.type.name === 'NativeSafeAreaProvider') { target = f; return; }
            if (f.child) { walk(f.child); }
            f = f.sibling;
        }
    })(root[key]);

    if (!target) {
        log('NativeSafeAreaProvider fiber not found (a minified build renames it — use a dev build).');
        return;
    }

    // pull the effect cleanup out of the hook list.
    // React 19 stores it on effect.inst.destroy; React 18 on effect.destroy.
    let holder = null;
    for (let hook = target.memoizedState; hook; hook = hook.next) {
        const eff = hook.memoizedState;
        if (!eff || typeof eff !== 'object') { continue; }
        if (eff.inst && typeof eff.inst.destroy === 'function' && /removeChild|\.remove\(\)/.test(String(eff.inst.destroy))) { holder = eff.inst; break; }
        if (typeof eff.destroy === 'function' && /removeChild|\.remove\(\)/.test(String(eff.destroy))) { holder = eff; break; }
    }
    if (!holder) {
        log('Found the provider but not its cleanup effect — React internals may have changed shape.');
        return;
    }

    const src = String(holder.destroy);
    const unpatched = /document\.body\.removeChild/.test(src);
    const isMeasure = (n) => n && n.style && n.style.position === 'fixed' && n.style.transitionProperty === 'padding' && n.style.visibility === 'hidden';
    const findEl = () => Array.prototype.filter.call(document.body.children, isMeasure)[0];

    log(RULE);
    log(' safe-area-context removeChild crash · Expensify/App #95343');
    log(RULE);
    log(' cleanup this build ships (the real closure, read off the fiber):');
    log(src.replace(/^/gm, '     '));
    log('');
    log(' removal call : ' + (unpatched ? 'document.body.removeChild(element)   [unpatched]' : 'element.remove()                     [patched]'));
    log('');

    let el = findEl();
    if (!el) { log(' No measurement element on <body> — is the app finished loading?'); return; }

    // 1. CONTROL — element attached, cleanup should complete normally
    let control;
    try { holder.destroy(); control = 'no error'; } catch (e) { control = e.name + ': ' + e.message; }
    log(' control  (element attached) : ' + control);

    // put it back so the failing case runs against the same closure
    if (!document.body.contains(el)) { document.body.appendChild(el); }

    // 2. CRASH — the teardown race: element already detached when cleanup runs
    el.remove();
    let crash;
    try { holder.destroy(); crash = 'no error'; } catch (e) { crash = e.name + ': ' + e.message; }
    log(' crash    (element detached) : ' + crash);

    if (!document.body.contains(el)) { document.body.appendChild(el); }

    const reproduced = /NotFoundError/.test(crash);
    log('');
    log(' VERDICT: ' + (reproduced
        ? 'BUG REPRODUCED — ' + crash
        : 'FIXED — the same cleanup completed with no error'));
    log(' (reload the page — this script tore the provider\'s effect down)');
    log(RULE);
})();
```

</details>

#### Confirming the actual fix in production

The checks above prove the cleanup no longer throws on an already-detached element; they cannot prove the crash stopped for real users, because they manufacture a state normal use never produces. Sentry **APP-GNR** — last seen ongoing at 2 users / 2 events on web `/search` — should stop receiving events once deployed. If events continue, the patch is not the whole story and the issue should be reopened rather than closed.

### Offline tests

The change only affects a DOM cleanup that runs on unmount. It reads and writes no data and does not depend on network state, so behaviour is identical offline.

### QA Steps

No user-visible behaviour changes, so QA is a regression check only: sign in, open Search (`/search`), navigate between a few Search views and back to Inbox, and verify the layout renders with the usual spacing and no errors appear in the JS console. Steps 1–2 above are developer-only.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

https://github.com/user-attachments/assets/5c3edcdc-d549-46ab-a5ba-13d10b1539d5


patches/react-native-safe-area-context/details.md
```diff
@@ -0,0 +1,18 @@
...
+    ```
+    The web build of NativeSafeAreaProvider appends a hidden measurement element to document.body and
+    its effect cleanup removes it with document.body.removeChild(element), which throws NotFoundError
+    when the element is no longer a child of body. During navigation the provider subtree can be torn
+    down with the element already detached, and the throw escapes as an uncaught error. This patch
+    detaches the transition listener first and then uses element.remove(), which is a no-op when the
+    node has no parent.
+    ```
+
+- Upstream PR/issue: 🛑 — the unguarded `document.body.removeChild` is still present on the library's `main` branch, so upgrading does not resolve it.
```
gijoe0295: We still need to raise an upstream PR to fix this. We can do it later after merging this one

```suggestion
- Upstream PR/issue: 🛑 TODO
```

tests/unit/SafeAreaProviderWebCleanupTest.tsx
gijoe0295, line 1: No need

patches/react-native-safe-area-context/details.md
wildan-m: Applied — will raise the upstream PR once this merges.

tests/unit/SafeAreaProviderWebCleanupTest.tsx
wildan-m: Removed.

gijoe0295 (approved):
LGTM. Please create the upstream PR soon @wildan-m

patches/react-native-safe-area-context/details.md
```diff
@@ -0,0 +1,18 @@
...
+    ```
+    The web build of NativeSafeAreaProvider appends a hidden measurement element to document.body and
+    its effect cleanup removes it with document.body.removeChild(element), which throws NotFoundError
+    when the element is no longer a child of body. During navigation the provider subtree can be torn
+    down with the element already detached, and the throw escapes as an uncaught error. This patch
+    detaches the transition listener first and then uses element.remove(), which is a no-op when the
+    node has no parent.
+    ```
+
+- Upstream PR/issue: 🛑 TODO
```
mountiny: Can you please do this now? If we dont do it now, we wont ever do it

mountiny (changes requested):

patches/react-native-safe-area-context/details.md
wildan-m: Done — upstream PR is https://github.com/appandflow/react-native-safe-area-context/pull/746, and `details.md` now links it instead of the TODO.

mountiny (approved):
Thanks for adding that!

PR #97629, opened 2026-08-01
[Payment due @aimane-chnaif] Support OFX and QFX bank statements in the personal card importer

allgandalf:
### Explanation of Change

NewDot's importer only takes xls/xlsx/csv/txt, so a bank OFX/QFX file gets rejected outright. Classic hands those to `UploadOFX` and parses them on the backend, so this does the same rather than adding a client side parser. The card is built optimistically and its `cardID`/`cardName` go up with the file, which https://github.com/Expensify/Auth/pull/23696 added.

Web-Expensify had to forward `cardID`, `cardName` and `reimbursable` on to Auth, that is https://github.com/Expensify/Web-Expensify/pull/55873.

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/666815
PROPOSAL:

### Tests

1. Go to Settings > Wallet > Import transactions
2. Give the card a name, hit Next
3. Upload a `.ofx` or `.qfx` file from your bank
4. Verify it goes straight to the success modal with no column mapping step
5. Verify the card shows up under Assigned Cards and the statement's transactions land on it
6. Repeat with a `.csv` and verify the column mapping step still works
7. Open a card you imported from a statement, hit Import transactions, and verify a `.ofx` file can be picked and lands on that same card
8. Open a card you imported from a spreadsheet, hit Import transactions, and verify `.ofx` and `.qfx` are not offered there

### Offline tests

The card is added optimistically and removed again if the upload fails.

Offline the Choose file spinner does not stop, because the success modal only resolves once the server responds. The spreadsheet import behaves the same way, both await `waitForImportFinalModal`, so this is not specific to statements.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/6d2d1f44-edd3-469a-a966-0ff0175ec985

</details>


src/libs/OFXUtils.ts
```diff
@@ -0,0 +1,75 @@
+import CONST from '@src/CONST';
+
+// The headers stay untranslated because they are persisted in the saved column layout and matched by name on a later import.
+const OFX_COLUMNS = [
+    {header: 'Date', role: CONST.CSV_IMPORT_COLUMNS.DATE},
+    {header: 'Merchant', role: CONST.CSV_IMPORT_COLUMNS.MERCHANT},
+    {header: 'Amount', role: CONST.CSV_IMPORT_COLUMNS.AMOUNT},
+];
+
+const STATEMENT_TRANSACTION_REGEX = /<STMTTRN>([\s\S]*?)<\/STMTTRN>/gi;
```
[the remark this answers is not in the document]
allgandalf: i checked this against real uploads before touching it. the backend that Classic posts OFX files to requires the same closing `</STMTTRN>`, and every OFX file we have actually received parses fine through it, none of them come back with zero transactions. so this isn't something the banks are doing in practice.

matching that behaviour matters too, if i loosen it here then NewDot starts accepting files Classic rejects and the two disagree. happy to revisit if we ever see one fail.

src/components/ImportSpreadsheet.tsx
```diff
@@ -109,6 +123,29 @@ function ImportSpreadsheet({backTo, goTo, shouldForceReplaceNavigation = false,
...
+        const ofxExtensions: readonly string[] = CONST.OFX_STATEMENT_EXTENSIONS;
+        if (ofxExtensions.includes(fileExtension.toLowerCase())) {
+            fetch(fileURI)
+                .then((data) => data.text())
+                .then((text) => {
+                    const rows = parseOFXToSpreadsheetRows(text);
+                    if (!rows) {
+                        throw new Error('No transactions found in the statement');
+                    }
+                    return setSpreadsheetData(rows, fileURI, file.type, file.name, isImportingMultiLevelTags ?? false, importedSpreadsheet?.importTransactionSettings, getOFXColumnRoles());
```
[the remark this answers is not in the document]
allgandalf: good catch, that was a real double flip. fixed in [bea1954](https://github.com/Expensify/App/pull/97629/commits/bea19549c7aa8e81d081b27ab800954cafbce5c6), OFX defines its own sign so the manual toggle is forced off for those files now.

src/CONST/index.ts
```diff
@@ -367,6 +367,12 @@ const CONST = {
     // Allowed extensions for text files that are used as spreadsheets
     TEXT_SPREADSHEET_EXTENSIONS: ['txt', 'csv'],
 
+    // Bank statement extensions, parsed with the OFX parser instead of the spreadsheet library
+    OFX_STATEMENT_EXTENSIONS: ['ofx', 'qfx'],
+
+    // Allowed extensions for the card transaction importer, which also accepts bank statement exports
+    ALLOWED_TRANSACTION_IMPORT_EXTENSIONS: ['xls', 'xlsx', 'csv', 'txt', 'ofx', 'qfx'],
```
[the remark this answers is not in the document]
allgandalf: fixed in [768e1c7](https://github.com/Expensify/App/pull/97629/commits/768e1c773e08656dfe54cb738a9114c7301f0584), dropped the hardcoded list and build it from `ALLOWED_SPREADSHEET_EXTENSIONS` + `OFX_STATEMENT_EXTENSIONS` instead.

src/libs/OFXUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Reads a single OFX element. OFX 1.x is SGML and leaves leaf elements unclosed, so a value runs until the next tag.
+ */
+function getElementValue(statementTransaction: string, tag: string): string {
+    const match = new RegExp(`<${tag}>([^<]*)`, 'i').exec(statementTransaction);
+    return match?.at(1)?.trim() ?? '';
```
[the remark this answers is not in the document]
allgandalf: good catch, fixed in [c63c4cf](https://github.com/Expensify/App/pull/97629/commits/c63c4cf767f) using `Str.htmlDecode`, so `AT&amp;T` imports as `AT&T` now.

worth flagging that the backend Classic uses has the same gap, i checked and it hands back the escaped string too. so this is a fix on our side rather than a difference i introduced, and it might be worth a separate issue for the classic path.

JmillsExpensify (dismissed):
Testing steps LGTM. Didn't text with an OFX file.

src/libs/OFXUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Converts an OFX DTPOSTED (YYYYMMDD, optionally followed by a time and time zone) to YYYY-MM-DD.
+ */
+function getPostedDate(postedDateTime: string): string {
+    const date = postedDateTime.slice(0, 8);
+    if (!/^\d{8}$/.test(date)) {
+        return '';
+    }
+    return `${date.slice(0, 4)}-${date.slice(4, 6)}-${date.slice(6, 8)}`;
+}
```
aimane-chnaif, line 33: Bug: Dates land one day early in US timezone

`getPostedDate` emits `yyyy-MM-dd`, which flows into [parseCSVDate](https://github.com/Expensify/App/blob/c70be59a7e5de24857bbfec0ab26a68396f89bed/src/libs/CSVDateUtils.ts#L26). Its first branch is `new Date(trimmedInput)`, and V8 parses a bare ISO date as **UTC midnight**, while `format()` renders in **local** time:

```
TZ=America/New_York → format(new Date('2026-03-02'), 'yyyy-MM-dd') === '2026-03-01'
```

Every OFX transaction shifts back a day for any user west of UTC.

Also, include a unit test that runs under a non-UTC `TZ`.


Image
Image
Image
Image

[edge-cases.ofx.zip](https://github.com/user-attachments/files/30710582/edge-cases.ofx.zip)

src/libs/OFXUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * OFX signs a charge negative, while the importer treats a positive amount as a charge, so the sign is flipped here.
+ */
+function getAmount(transactionAmount: string): string {
+    const amount = Number(transactionAmount);
+    if (!transactionAmount || !Number.isFinite(amount)) {
+        return '';
+    }
+    return String(-amount);
```
aimane-chnaif, line 61: Bug: Comma decimal separators silently drop rows

The OFX spec permits `,` as the decimal separator, and European exports use it.
`Number('-24,99')` is `NaN`, so `getAmount` returns `''` and the row is skipped at [OFXUtils.ts:54](https://github.com/Expensify/App/pull/97629/changes#diff-022a1fe1abf4ce1c5eedb4f3e517d90a3c6c79e1cc89bb3106abe3a64f85f372R54) — no error, just missing transactions.
If every row uses commas, the user gets the generic "invalid file" modal with no clue why.

src/components/ImportSpreadsheet.tsx
```diff
@@ -109,6 +123,31 @@ function ImportSpreadsheet({backTo, goTo, shouldForceReplaceNavigation = false,
...
+        if (ofxExtensions.includes(fileExtension.toLowerCase())) {
+            fetch(fileURI)
+                .then((data) => data.text())
+                .then((text) => {
+                    const rows = parseOFXToSpreadsheetRows(text);
+                    if (!rows) {
+                        throw new Error('No transactions found in the statement');
+                    }
+                    // OFX fixes the sign of an amount, so the manual flip the user picked for spreadsheets must not apply.
+                    const importTransactionSettings = {...importedSpreadsheet?.importTransactionSettings, flipAmountSign: false};
```
aimane-chnaif: Forcing `flipAmountSign: false` here might skip the user's persisted choice

This writes `flipAmountSign: false` back into `importTransactionSettings` via `Onyx.set`.
Overriding it *for this import* is correct, but it's persisted: after an OFX upload the user can navigate back to `ImportTransactionsPage`, where the toggle re-initializes from Onyx ([ImportTransactionsPage.tsx:32](https://github.com/Expensify/App/blob/c70be59a7e5de24857bbfec0ab26a68396f89bed/src/pages/settings/Wallet/ImportTransactionsPage.tsx#L32)) and now reads off.
They then upload a CSV with their flip preference silently discarded.
The `false` also gets baked into the card's saved column layout via `buildColumnLayout`.

Prefer passing the override through to the parse path rather than mutating shared settings.

cspell.json
```diff
@@ -211,6 +213,7 @@
         "CODEOWNERS",
         "luacmartins",
         "mountiny",
+        "STMTTRN",
```
aimane-chnaif: This is inside GitHub usernames (`luacmartins`, `mountiny`, `tgolen`) rather than the alphabetized word list where `BANKTRANLIST`/`DTPOSTED`/`TRNAMT` went.

cspell.json
```diff
@@ -35,6 +35,7 @@
         "Amina",
         "Areport",
         "Authy",
+        "BANKTRANLIST",
```
aimane-chnaif: This only appears in the test file, which is already wrapped in `// cspell:disable` — that entry looks unnecessary.

src/languages/es.ts
```diff
@@ -1110,6 +1110,10 @@ const translations: TranslationDeepObject<typeof en> = {
         dragAndDropMultiLevelTag: `<muted-link>Arrastra y suelta un archivo de hoja de cálculo aquí, o elige un archivo a continuación. <a href="${CONST.IMPORT_SPREADSHEET.MULTI_LEVEL_TAGS_ARTICLE_LINK}">Aprende más</a> sobre los formatos de archivo soportados.</muted-link>`,
         chooseSpreadsheet: '<muted-link>Elige un archivo de hoja de cálculo para importar. Los formatos soportados son .csv, .txt, .xls y .xlsx.</muted-link>',
         chooseSpreadsheetMultiLevelTag: `<muted-link>Elige un archivo de hoja de cálculo para importar. <a href="${CONST.IMPORT_SPREADSHEET.MULTI_LEVEL_TAGS_ARTICLE_LINK}">Aprende más</a> sobre los formatos de archivo soportados.</muted-link>`,
+        dragAndDropTransactions:
+            '<muted-link>Arrastra y suelta un archivo de hoja de cálculo o extracto bancario aquí. Los formatos soportados son .csv, .txt, .xls, .xlsx, .ofx y .qfx.</muted-link>',
```
aimane-chnaif: Bug: `choose a file below` phrase is missing in Spanish

And please pull main. We now support Greek locale (el.ts) which is missing in this PR

src/libs/OFXUtils.ts
allgandalf: real one, thanks. the shift comes from `parseCSVDate` trying `new Date()` before the explicit formats, so a bare `yyyy-MM-dd` reads as UTC midnight and formats back a day early. moved the format loop first in [b83c43d](https://github.com/Expensify/App/pull/97629/commits/b83c43dc76b6cb674f42edd461853ab40e892d42). anything with a time or zone still falls through to native parsing, so nothing else changes.

on the non-UTC test, i wrote one and then threw it away because it was fake. setting `process.env.TZ` inside a jest test does nothing here, i probed it and the value stayed UTC, so the test went green with the bug still in place. what does catch it is running the suite with `TZ=America/New_York`, the existing CSVDateUtils canary fails on main today. if you want that guarded in CI it needs a TZ in the jest config, which felt like a separate call, lmk if you'd rather i do it here.

src/libs/OFXUtils.ts
allgandalf: fixed in [de7844c](https://github.com/Expensify/App/pull/97629/commits/de7844ce303c85ea172381152740f62b88564c2c). the right-most separator counts as the decimal point when one or two digits follow it, so `-24,99`, `-1.234,56` and `-1,234.56` all land correctly and `-1,234` stays thousands.

for context the backend Classic posts to doesn't skip these, it strips the comma, so `-24,99` imports as -$2,499 over there. so this was worth doing properly.

src/components/ImportSpreadsheet.tsx
allgandalf: you're right, that persisted and would have eaten their next CSV. dropped the settings mutation in [f8a4051](https://github.com/Expensify/App/pull/97629/commits/f8a405177f9bf2e609b514e99e5dfb9c3f9fc6d6) and the flip is now skipped at import time for a statement file in [699ee77](https://github.com/Expensify/App/pull/97629/commits/699ee776a21a5d04af43e1248ecf6cd47722f0d8), so the saved toggle is untouched.

cspell.json
allgandalf: fixed in [2f29ffd](https://github.com/Expensify/App/pull/97629/commits/2f29ffd38fc7eab5301dfb713b14868009817c09), it sits next to `SSAE` now. `TRNAMT` had the same problem sitting in the mixed-case block so i moved that one too.

cspell.json
allgandalf: half right so i took the other half. it was actually still needed, two of the inline fixtures were not inside a `cspell:disable`. wrapped those in [eb11a8b](https://github.com/Expensify/App/pull/97629/commits/eb11a8bfc7fece4f4644faef755fb101ba86979f) and dropped the entry in [2f29ffd](https://github.com/Expensify/App/pull/97629/commits/2f29ffd38fc7eab5301dfb713b14868009817c09), so now it genuinely isn't needed.

src/languages/es.ts
allgandalf: both done. spanish has the missing phrase in [5e3740f](https://github.com/Expensify/App/pull/97629/commits/5e3740f308086dd14e6e65968c44e8dcbdf92bbf), and main is merged so el.ts exists now, greek copy added in [00edde8](https://github.com/Expensify/App/pull/97629/commits/00edde8103438df7706f71a6d3fd9f0aba14f4f8).

src/libs/actions/ImportTransactions.ts
```diff
@@ -253,7 +254,9 @@ async function importTransactionsFromCSV(
     existingCardID?: number,
     previouslySavedLayout?: SavedCSVColumnLayoutData,
 ): Promise<ImportFinalModal> {
-    const settings = spreadsheet.importTransactionSettings ?? {};
+    // OFX already signs a charge, so the manual flip is dropped for this import without touching the saved preference.
+    const storedSettings = spreadsheet.importTransactionSettings ?? {};
+    const settings = isOFXStatement(spreadsheet.fileName ?? '') ? {...storedSettings, flipAmountSign: false} : storedSettings;
```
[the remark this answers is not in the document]
allgandalf: fair, the forced settings were reaching `buildColumnLayout` too. fixed in [931934f](https://github.com/Expensify/App/pull/97629/commits/931934f2dadaefa30a29836f4ecb1c35adce8f98): an OFX import into an existing card now keeps that card's saved layout instead of overwriting it. OFX always maps its own columns so it never needs the saved one, and a brand new card still saves the statement layout since there is nothing to preserve.

one correction on the reasoning, the sign preference for the next import comes from `importTransactionSettings`, not the saved layout, and that has been untouched since the earlier fix. App also never reads `flipAmountSign` back out of a saved layout, only the column names, so the real impact was the layout Classic shows for that card.

src/libs/OFXUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+function normalizeSeparators(transactionAmount: string): string {
+    const decimalIndex = Math.max(transactionAmount.lastIndexOf(','), transactionAmount.lastIndexOf('.'));
+    if (decimalIndex < 0) {
+        return transactionAmount;
+    }
+
+    const tail = transactionAmount.slice(decimalIndex + 1);
+    const head = transactionAmount.slice(0, decimalIndex).replaceAll(/[,.]/g, '');
+    return /^\d{1,2}$/.test(tail) ? `${head}.${tail}` : `${head}${tail}`;
```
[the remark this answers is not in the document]
allgandalf: good catch, this one was mine and it was bad. `-24.9900` came out as $249,900 because the grouping heuristic only accepted one or two digits after the separator. fixed in [a5879a0](https://github.com/Expensify/App/pull/97629/commits/a5879a0ad8ec48aed50071bb212556c6133d0189): a period always marks the decimal point however many digits follow, since that is the separator the spec defines, and the one-or-two-digit rule now only applies to a comma. `-24.9900`, `-24,99`, `-1.234,56`, `-1,234.56` and `-1,234` all land correctly.

src/components/ImportSpreadsheet.tsx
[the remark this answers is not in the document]
allgandalf: you're right, i checked `TransactionsImportedPage` and it does run that for every existing-card import. skipped it for statements in [d2246dc](https://github.com/Expensify/App/pull/97629/commits/d2246dcb677a23a44ae5749c5b57b1cdffff591f).

the collision needs a saved layout that maps one of `Date`/`Merchant`/`Amount` to a different role, so a plain matching layout was already a no-op, but it would have knocked out the amount role and failed validation when it did happen.

src/libs/OFXUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    const lastPeriod = transactionAmount.lastIndexOf('.');
+    const rightMost = Math.max(transactionAmount.lastIndexOf(','), lastPeriod);
+    if (rightMost < 0) {
+        return transactionAmount;
+    }
+
+    // A period is the separator the spec defines, so it always marks the decimal point, however many digits
+    // follow. A comma only marks it when it is right-most with one or two digits after it, otherwise it groups.
+    if (rightMost !== lastPeriod && !/,\d{1,2}$/.test(transactionAmount)) {
+        return transactionAmount.replaceAll(/[,.]/g, '');
```
aimane-chnaif: @MelvinBot see https://github.com/Expensify/App/pull/97629/changes#r3737684220.
Is there any existing pattern throughout the app to reuse?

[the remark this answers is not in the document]
MelvinBot: Short answer: there's no drop-in helper that fully covers this. The App's designated "any locale" parser is [`NumberUtils.parseFloatAnyLocale`](https://github.com/Expensify/App/blob/1d499e59ac424149c6a3c84944da264df45fdb90/src/libs/NumberUtils.ts#L52), but it's just `parseFloat(value.replace(',', '.'))`, which mishandles thousands grouping — the exact cases your `normalizeSeparators` + tests already cover:

- `parseFloatAnyLocale('-1.234,56')` → `-1.234` ❌ (should be `-1234.56`)
- `parseFloatAnyLocale('-1,234.56')` → `-1.234` ❌ (should be `-1234.56`)
- `parseFloatAnyLocale('-1,234')` → `-1.234` ❌ (should be `-1234`)

So reusing it as-is would regress grouped amounts.

The other candidates don't fit either:
- [`MoneyRequestUtils.stripCommaFromAmount` / `replaceCommasWithPeriod`](https://github.com/Expensify/App/blob/1d499e59ac424149c6a3c84944da264df45fdb90/src/libs/MoneyRequestUtils.ts#L17) are for fixed-format keypad input in a known layout — same grouping blind spot.
- [`LocaleDigitUtils.fromLocaleDigit`](https://github.com/Expensify/App/blob/1d499e59ac424149c6a3c84944da264df45fdb90/src/libs/LocaleDigitUtils.ts#L64) / `NumberFormatUtils` are keyed to the **viewer's** app locale (read from Onyx), but a bank file's separator convention is per-file/per-bank and independent of who imports it — so they'd misread a file whose locale differs from the importer's.

So a dedicated parser here is justified. If the goal is to consolidate rather than keep two amount-parsers that can disagree, the cleaner move is to strengthen `parseFloatAnyLocale` to handle grouping and reuse it, since it's already the app's canonical "any locale" number helper.

One thing worth flagging while you're here: the case in the comment above (`-24,9900`) is genuinely mis-parsed by the current `normalizeSeparators` today. Tracing it: the lone comma is right-most with 4 trailing digits, so it fails the `,\d{1,2}$` test and falls into the grouping branch → strips to `-249900` → after the sign flip and the importer's cents conversion that's off by orders of magnitude. Note the inherent ambiguity for commas specifically: `1,234` (grouping) and `24,9900` (fixed-precision decimal) both have >2 digits after the comma, so digit count alone can't disambiguate — trimming trailing-zero precision first (as suggested above) is the pragmatic tie-breaker.

src/libs/CSVDateUtils.ts
```diff
@@ -30,20 +30,21 @@ function parseCSVDate(input: string): string | null {
...
         if (isValid(parsedDate)) {
             return format(parsedDate, CONST.DATE.FNS_FORMAT_STRING);
         }
     }
 
+    // Falls back to native parsing for anything with a time or zone, which the formats above reject
+    let date = new Date(trimmedInput);
+    if (isValid(date) && !Number.isNaN(date.getTime())) {
+        return format(date, CONST.DATE.FNS_FORMAT_STRING);
+    }
```
aimane-chnaif, line 57: Two-digit years now parse to year 00xx

Moving the `CSV_DATE_FORMATS` loop ahead of `new Date()` fixes one day shift bug, but `new Date()` was also what handled two-digit years — and date-fns' `yyyy` token accepts a 2-digit string as a literal year:

```
input        old (main)     new (this PR)
"01/15/25"   2025-01-15  →  0025-01-15
"9/8/25"     2025-09-08  →  0025-09-08
"12/31/99"   1999-12-31  →  0099-12-31
"2 Nov 25"   2025-11-02  →  0025-11-02
"08-09-25"   2025-08-09  →  0008-09-25   ← field order also flips
```

This is not OFX-specific — `parseCSVDate` is the shared path for **every** CSV/xlsx transaction import, and `MM/DD/YY` is one of the most common bank export formats. A transaction imports with `created: '0025-01-15'` and goes to `ImportCSVTransactions` that way.

`tests/unit/CSVDateUtils.test.ts` exists but has no two-digit-year case, so it still passes — and the PR doesn't touch it despite changing the function.

Suggested fix — add explicit 2-digit formats to `CSV_DATE_FORMATS` (date-fns `yy` maps to the 2000s window correctly), placed after their 4-digit counterparts:

```ts
'MM/dd/yy', 'M/d/yy', 'dd/MM/yy', 'MM-dd-yy', 'dd-MM-yy', 'd MMM yy', 'MMM d, yy',
```

Keeping `new Date()` as the final fallback (as the PR now does) then covers the rest.

Please add test cases to `CSVDateUtils.test.ts` for both the two-digit years and a non-UTC `TZ` — the reorder is the kind of change that needs its own regression net.

@MelvinBot confirm

src/components/ImportSpreadsheet.tsx
```diff
@@ -109,6 +123,29 @@ function ImportSpreadsheet({backTo, goTo, shouldForceReplaceNavigation = false,
 
         setIsReadingFile(true);
 
+        const ofxExtensions: readonly string[] = CONST.OFX_STATEMENT_EXTENSIONS;
+        if (ofxExtensions.includes(fileExtension.toLowerCase())) {
```
aimane-chnaif: Can we reuse `isOFXStatement`?

src/libs/OFXUtils.ts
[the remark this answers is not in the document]
allgandalf: same bug on the comma side, fixed in [97d1e92](https://github.com/Expensify/App/pull/97629/commits/97d1e92c08b9a8b479d9222058e70a812b100cb2). a trailing comma group now only counts as grouping at exactly three digits, which is the one width a decimal amount is never written in, so `-24,9900` reads as 24.99 and `-1,234` still reads as 1234.

@MelvinBot's point about `parseFloatAnyLocale` matches what i found, it does a bare `replace(',', '.')` so it would turn `-1.234,56` into -1.234. reusing it would regress the grouped cases this already covers, so keeping the local parser.

src/libs/CSVDateUtils.ts
allgandalf: your table was right, all five. fixed in [9d12ee6](https://github.com/Expensify/App/pull/97629/commits/9d12ee6879b73d0c82b563602dfb96a8216de87a) with tests in [6721e1a](https://github.com/Expensify/App/pull/97629/commits/6721e1a41d942376392d12e55187309bdc53fa91).

one correction on the placement though: after the four-digit counterparts does not work, `yyyy` happily takes "25" as year 0025 so it matches first and the `yy` entry is never reached. they have to go before, which is safe because `yy` rejects a four-digit year outright. `08-09-25` also needed them ahead of `yyyy-MM-dd`, which was reading "08" as the year. all 14 cases i checked now match main.

on the non-UTC test, i tried again and it still cannot be done in-process. setting `process.env.TZ` inside a jest test has no effect here, i probed it by asserting the formatted value and it stayed UTC, so the test passes with the bug in place. running the suite with `TZ=America/New_York` does catch it and i ran everything both ways. guarding it in CI needs a TZ in the jest config, happy to do that if you want it, just did not want to change the whole suite's timezone inside this PR.

src/components/ImportSpreadsheet.tsx
allgandalf: done in [98fcc2f](https://github.com/Expensify/App/pull/97629/commits/98fcc2f1e1a02ad48ad6f693bfc93a8e4770b579).

src/libs/OFXUtils.ts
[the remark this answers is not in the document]
MelvinBot: Confirmed — the fix is correct and agrees with what I found. I traced [`normalizeSeparators`](https://github.com/Expensify/App/blob/0f59ab8fd4a572e715cd58638765322bcf9d8802/src/libs/OFXUtils.ts#L38-L51) against all the cases and they land right:

- `-24,99` → comma + 2 digits, `,\d{3}$` misses → decimal → `24.99` ✓
- `-24,9900` → comma + 4 digits, `,\d{3}$` misses → decimal → `24.99` ✓
- `-1,234` → comma + exactly 3 digits, [`,\d{3}$`](https://github.com/Expensify/App/blob/0f59ab8fd4a572e715cd58638765322bcf9d8802/src/libs/OFXUtils.ts#L47) hits → grouping → `1234` ✓
- `-1.234,56` / `-1,234.56` → both → `1234.56` ✓

The `,\d{3}$` tie-break is sound: thousands groups are always exactly three digits, and in a two-decimal-cents world a fractional part is written with 2 (`,99`) or padded precision (`,9900`) but never exactly 3 — so the one width that's unambiguously grouping is exactly the one you match. The [tests cover all six](https://github.com/Expensify/App/blob/0f59ab8fd4a572e715cd58638765322bcf9d8802/tests/unit/OFXUtils.test.ts#L100-L121) too. 👍

And agreed on keeping the local parser — `parseFloatAnyLocale` does a bare `replace(',', '.')` and would turn `-1.234,56` into `-1.234`, regressing exactly the grouped cases this handles.

One edge case, purely for the record (not a blocker): three-decimal currencies with a comma — e.g. `-24,990` (24.990 KWD) — would satisfy `,\d{3}$` and read as grouping → `24990`. But those currencies already break the importer's two-decimal cents assumption regardless of this parser, so it's pre-existing and out of scope here.

aimane-chnaif (approved):

aimane-chnaif:
Nice! It's better to parse OFX/QFX files in backend.
Let me know if you want me to retest after backend deployed.

src/components/ImportSpreadsheet.tsx
```diff
@@ -109,6 +133,23 @@ function ImportSpreadsheet({backTo, goTo, shouldForceReplaceNavigation = false,
 
         setIsReadingFile(true);
 
+        // A statement carries its own columns, so it is parsed by the backend and skips the column mapping step.
+        if (isOFXStatement(file.name ?? '')) {
+            uploadOFXStatement(file, importedSpreadsheet?.importTransactionSettings ?? {}, accountID, existingCardID)
```
aimane-chnaif: Not tested yet but existing card uploads might rename the card.

It's consistent with CSV flow which seeds settings from card first:
https://github.com/Expensify/App/blob/1235a0908350f75ae2093b5ff38eeac8c82ee92a/src/pages/settings/Wallet/TransactionsImportedPage.tsx#L134

But OFX path skips it, so re-uploading a statement to an existing card sends `cardName: 'Imported Card'` whenever `importTransactionSettings` is empty.

src/libs/API/parameters/UploadOFXParams.ts
```diff
@@ -0,0 +1,17 @@
...
+    file: FileObject;
+
+    /** Optimistic cardID the statement imports into */
+    cardID: number;
+
+    /** Display name for the card */
+    cardName: string;
+
+    /** Whether the transactions are reimbursable */
+    reimbursable: boolean;
```
aimane-chnaif, line 14: `currency` is not supported?
The user still picks a currency on `ImportTransactionsPage` before reaching the upload screen, and for OFX that choice is silently dropped

src/components/ImportSpreadsheet.tsx
```diff
@@ -109,6 +133,23 @@ function ImportSpreadsheet({backTo, goTo, shouldForceReplaceNavigation = false,
...
+                .then((importFinalModal) => showImportSpreadsheetConfirmModal(importFinalModal, {shouldHandleNavigationBack: false}))
+                .then((didShowImportFinalModal) => {
+                    if (!didShowImportFinalModal) {
+                        return;
+                    }
+                    setIsClosing(true);
+                    Navigation.dismissModal();
+                })
+                .finally(() => {
```
aimane-chnaif: Let's add `.catch`. There might be exceptions

src/libs/CSVDateUtils.ts
```diff
@@ -30,20 +41,21 @@ function parseCSVDate(input: string): string | null {
...
```
aimane-chnaif: Do we still need this change?
`parseCSVDate` is not used anymore in OFX/QFX flow.

src/components/ImportSpreadsheet.tsx
```diff
@@ -1,15 +1,19 @@
...
 import useConfirmModal from '@hooks/useConfirmModal';
+import useImportSpreadsheetConfirmModal from '@hooks/useImportSpreadsheetConfirmModal';
 import {useMemoizedLazyExpensifyIcons} from '@hooks/useLazyAsset';
 import useLocalize from '@hooks/useLocalize';
 import useOnyx from '@hooks/useOnyx';
 import useResponsiveLayout from '@hooks/useResponsiveLayout';
 import useThemeStyles from '@hooks/useThemeStyles';
 
 import {setSpreadsheetData} from '@libs/actions/ImportSpreadsheet';
+import {uploadOFXStatement} from '@libs/actions/ImportTransactions';
```
aimane-chnaif: This component is the entry screen for 8 flows (categories, tags, members, per diem, workflows, merchant rules, company cards, and transactions)

and previously had no dependency on `ImportTransactions` at all.

It now pulls that module's whole graph (CardUtils, CSVDateUtils + date-fns, NumberUtils, OFXUtils, the API layer) into seven flows that will never call it. 

We can avoid this by keeping the OFX branch in `ImportTransactionsSpreadsheetPage`, passing it down as a callback.

aimane-chnaif (changes requested):

allgandalf:
Auth PR still in review

src/components/ImportSpreadsheet.tsx
allgandalf: moved it to ImportTransactionsSpreadsheetPage and pass it down as `onStatementPicked`, so the other flows do not pull ImportTransactions in

src/components/ImportSpreadsheet.tsx
allgandalf: added

src/components/ImportSpreadsheet.tsx
allgandalf: UploadOFX rejects a statement for a card the spreadsheet importer made, it holds the name you typed where the account number goes. so statements are only offered when creating a card now

src/libs/CSVDateUtils.ts
allgandalf: reverted. separate thing worth raising though, on main an ISO date in a CSV imports a day early west of UTC

src/libs/API/parameters/UploadOFXParams.ts
allgandalf: UploadOFX takes no currency, ScrapeCard reads CURDEF out of the statement. the picker comes before the file so we cannot hide it, open to ideas

allgandalf:
@aimane-chnaif lets test this now, the Web-E PR is on staging for now, so make sure to keep that toggle on

aimane-chnaif:
### Currency selection is silently discarded

Repro:
1. Settings → Wallet → tap **Import transactions**
2. Set **Card display name**
3. Tap **Currency**, choose **EUR** (a currency different from the one in your statement's `<CURDEF>`)
4. Tap **Next**
5. On **Import spreadsheet**, tap **Choose file** and pick a `.ofx` file whose `<CURDEF>` is `USD`
6. Confirm `Import successful`, wait for the transactions to appear, then open the new card

Expected: transactions in EUR, the currency you chose.
Actual: transactions in USD. No warning that your selection was ignored.


### A card created from a statement cannot accept another statement

Repro:
1. Complete an OFX import as above so a new card exists
2. Settings → Wallet → tap that card → **Import spreadsheet**
3. Tap **Choose file** and try to pick `.ofx` file

Expected: the statement imports into the existing card.
Actual: non-csv files (.ofx, .qfx) are not selectable in native picker modal

### Offline upload

Repro:
1. Wallet → **Import transactions** → name it → **Next**
2. Turn off the network
3. Choose file, pick a `.ofx`

Expected: not sure. need product confirmation
Actual: infinite loading shows

### Import failed with specific ofx

502 Invalid response from backend

[edge-cases.ofx.zip](https://github.com/user-attachments/files/31962675/edge-cases.ofx.zip)

allgandalf:
> Currency selection is silently discarded

`UploadOFX` has no currency param, `ScrapeCard` reads `CURDEF` out of the statement and falls back to USD. Same for flip amount sign, OFX signs its own amounts. Neither can be honoured, so the question is whether we hide those two rows once a statement is picked. That is a product call, the file is chosen after the settings screen.

> A card created from a statement cannot accept another statement

Fixed. Statements are offered on any card now except one imported from a spreadsheet, those store the name you typed where the account number goes and `UploadOFX` throws `403 Card is for a different account number`.

> Offline upload

Not new, the CSV import does the same. Both await `waitForImportFinalModal`, which only resolves when `successData` applies, so offline it never settles. Worth its own issue.

> Import failed with specific ofx

That one is the scraper. `PostProcess/Expense/Merge.php` does `-str_replace(',', '', $trnamt)` and your file has a `STMTTRN` with no `TRNAMT`, so PHP 8 throws `Unsupported operand types: string * int` and the whole parse dies. Auth then sees no acctid and returns 502. Logs: request `a37ed0b5cc2ef3bd-SJC`, `2026-09-08T14:59:07Z`. Classic hits it the same way. Same line also reads `-24,99` as 2499.

@MelvinBot Web-E [merged](https://github.com/Expensify/Web-Expensify/pull/55873) on the 4th. Tests added, `OFXUtils` gone, prop split, object URL moved, `Log.warn` on failure, magic string into CONST, export order fixed. Left the duplicated modal setup alone, pulling it out means refactoring the CSV path.

aimane-chnaif:
Are all comments addressed?

src/pages/settings/Wallet/ImportTransactionsSpreadsheetPage.tsx
```diff
@@ -1,21 +1,52 @@
...
+    const [accountID = CONST.DEFAULT_NUMBER_ID] = useOnyx(ONYXKEYS.SESSION, {selector: accountIDSelector});
+    const {setIsClosing} = useCloseImportPage();
+    const showImportSpreadsheetConfirmModal = useImportSpreadsheetConfirmModal();
+
+    // Importing a spreadsheet stores the name the person gave the card where its account number goes,
+    // and UploadOFX rejects a statement for a card holding a name.
+    const hasImportedSpreadsheet = !!cardID && !!savedColumnLayouts?.[String(cardID)];
+
+    const uploadStatement = async (file: FileObject) => {
+        const importFinalModal = await uploadOFXStatement(file, importedSpreadsheet?.importTransactionSettings ?? {}, accountID, cardID ? Number(cardID) : undefined);
```
aimane-chnaif: Bug: Existing-card upload sends the wrong card name and reimbursable flag

Repro:
1. Wallet → Import transactions → name the card `Chase Personal` → Next → Choose file → pick a `.ofx` → wait for the card to appear
2. Settings → Wallet → tap that card → toggle `Mark transactions as reimbursable off`
3. On the same screen, tap `Import spreadsheet`
4. Choose file → pick next month's `.ofx`

Expected: the statement imports under `Chase Personal`, still non-reimbursable.
Actual: the request carries `cardName: 'Imported Card'` and `reimbursable: true`, discarding both.

src/components/ImportSpreadsheet.tsx
```diff
@@ -96,14 +112,25 @@ function ImportSpreadsheet({backTo, goTo, shouldForceReplaceNavigation = false,
...
+        const {fileExtension} = splitExtensionFromFileName(file?.name ?? '');
+        const statementExtensions: readonly string[] = CONST.OFX_STATEMENT_EXTENSIONS;
+
+        // A statement carries its own columns, so the backend parses it and the column mapping step is skipped.
+        if (onStatementPicked && statementExtensions.includes(fileExtension.toLowerCase())) {
+            setIsReadingFile(true);
+            onStatementPicked(file).finally(() => {
+                setIsReadingFile(false);
+            });
+            return;
```
aimane-chnaif, line 129: Missing .catch

allgandalf:
> the gate still blocks statements on a statement-created card once any CSV has been imported into it

That is the right call. Importing a CSV sends the card name as the account number, and `ScrapeCard` rewrites `cardNumber` to it on mismatch, so after step 2 of your repro the card no longer holds the real account number and `UploadOFX` would 403. Keying off `accountDetails.accountID` would not help, that is the card display name in both cases.

> the existing-card path still sends 'Imported Card' / reimbursable: true

Fixed, it calls `getExistingCardImportSettings` now like the CSV path.

> statements are allowed on a spreadsheet card while Onyx is still loading

Fixed with `isLoadingOnyxValue`. Also guarded the branch on `shouldAllowBankStatements && onStatementPicked`, and the drop target on `isReadingFile`.

@aimane-chnaif `.catch` added, it logs instead of showing a second modal since `uploadOFXStatement` already returns the failure one.

src/pages/settings/Wallet/ImportTransactionsSpreadsheetPage.tsx
allgandalf: fixed, it reads the card settings through `getExistingCardImportSettings` now like the CSV path does

https://github.com/Expensify/App/pull/97629/commits/4f51fc97b75663cbfa9a1a181a8b0dd88645fa89

src/components/ImportSpreadsheet.tsx
allgandalf: added, it logs rather than showing a second modal since `uploadOFXStatement` already returns the failed one

https://github.com/Expensify/App/pull/97629/commits/c07a7611d55a2a696758c09fdbfdceddc9f1e557

aimane-chnaif (approved):

allgandalf:
@robertjchen this PR is ready for your review

robertjchen (approved):
Thanks!! This looks good to me! 🙇

mitarachim:
@allgandalf This PR is failing because of a regression issue https://github.com/Expensify/App/issues/101281

The issue is reproducible in: iOS


https://github.com/user-attachments/assets/75ea1d82-029c-4d61-b324-1b260052cde8

Issue #96716, opened 2026-07-22
[Due for payment 2026-08-27] [$250] [DEV] Infinite loading when switching to Inbox


izarutskaya:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding. To join our Slack channel, fill out [this form](https://forms.gle/Q7hnhUJPnQCK7Fe56).
___

**Version Number:** 9.4.41-4 (DEV only)
**Reproducible in staging?:** Needs Reproduction
**Reproducible in production?:** Needs Reproduction
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @bernhardoj 
**Slack conversation** (hyperlinked to channel name): [#expensify_bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1784694615651199)

## Action Performed:
1. Go to New Expensify on Dev
2. Open Inbox

## Expected Result:
A report is shown

## Actual Result:
Infinite loading. The report URL doesn't have reportID

## Workaround:
Unknown

## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [ ] MacOS: Chrome / Safari

## Screenshots/Videos

https://github.com/user-attachments/assets/1aa37487-b7ec-46dd-8e2f-7405cf078ed1

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: <a href="https://www.upwork.com/jobs/~022080712484395675250">https://www.upwork.com/jobs/~022080712484395675250</a></li>
        <li>Upwork Job ID: 2080712484395675250</li>
        <li>Last Price Increase: 2026-07-24</li>
<li>Automatic offers: </li>
<ul>
<li>wildan-m | Contributor | 112132364</li>
</ul></li>
    </ul>
</details>

[...]

PR #97636, opened 2026-08-01
Recover the missing report ID once the reports collection loads

wildan-m:
### Explanation of Change

Opening the Inbox on a wide layout can land on a report URL that carries no report ID, leaving the central pane on a loading skeleton that never resolves. The screen filling the central pane picks its report by scanning the report list once, at mount. That list is read from a plain module-level copy that is filled in asynchronously, so a screen that mounts before the reports finish loading scans an empty list and the route is created without a report ID. With no report ID the screen's own fetch bails out on its empty-ID guard, so nothing is ever requested and the skeleton stays.

The app already has a safety net that backfills a missing report ID, but it reads that same module-level copy — and a plain variable being filled in later re-renders nothing, so the effect never runs again and the empty route never heals. Subscribing that handler to the reports collection makes the data's arrival the trigger: the moment the reports load, the handler re-runs and fills the param through the navigation call it already makes. This heals every route created without a report ID, not just the Inbox tab switch. When the reports are already loaded at mount nothing changes — the same list is read, through a subscription rather than a one-shot snapshot.

### Fixed Issues

$ https://github.com/Expensify/App/issues/96716
PROPOSAL: https://github.com/Expensify/App/issues/96716#issuecomment-5072845324

### Tests

The bug is a race between the report route mounting and the reports collection loading, so it cannot reproduce on a warm cache. Steps 1–3 make the reports absent at mount, which reproduces it every time. `window.Onyx` is exposed on dev builds — paste the snippets into the DevTools console. Note the fix backfills the route param, so the signal to watch is the **address bar**, not the loading state alone.

**Setup:** dev build, window wide enough for the two-column layout (~1000px+), app open in **one tab only** (Onyx syncs across tabs, so a second tab pushes the reports back).

1. Sign in, then open the **Home** tab — clearing from the Inbox would just update the open Inbox instead of leaving the route to mount without reports.
2. Stash your reports and clear the collection (the stash is restored in step 5, so nothing is lost):
   ```js
   sessionStorage.setItem('savedReports', JSON.stringify(await window.Onyx.get('report_')));
   await window.Onyx.setCollection('report_', {});
   ```
3. Go to `https://dev.new.expensify.com:8082/r`.
4. Verify the loading skeleton and the address bar stuck on `/r` with no report ID — the state from the issue.
5. Put the reports back — this is the "reports finished loading" moment:
   ```js
   await window.Onyx.mergeCollection('report_', JSON.parse(sessionStorage.getItem('savedReports')));
   ```
6. Verify the address bar becomes `/r/<reportID>` and the report renders with its messages. **On `main`** the chats all appear in the left list but the central pane stays on the skeleton and the address bar stays `/r`.
7. Open the Inbox and verify a report opens with `/r/<reportID>`.
8. Open a report from the left list and verify it still opens normally.

If a run is interrupted between steps 2 and 5, the reports stay cleared — the server only sends deltas, so a reload will not bring them back. Restore them with **Settings → Troubleshoot → Clear cache and restart**, which refetches everything and keeps you signed in.

### Offline tests

This change adds no network request and alters none. Offline, the reports collection is served from local storage, and the handler resolves the report ID from that collection exactly as it does online — including the recovery step, which fires when the locally stored reports finish loading.

### QA Steps

The issue was reported against a dev build, where the slower bundle lets the report route mount before the reports finish loading, and it was never reproduced on staging or production. On staging the reports are normally loaded first, so these steps are a no-regression pass rather than a reproduction — they are expected to pass both before and after this change. The deterministic reproduction is in the `Tests` section above and needs the DevTools console.

1. Sign in to New Expensify, with the browser window wide enough for the two-column layout (roughly 1000px or more).
2. Open the Inbox tab.
3. Verify a report is shown in the central pane and the address bar shows `/r/<reportID>`, not a bare `/r`.
4. Navigate directly to the bare report path `/r`.
5. Verify the address bar gains a report ID and the last accessed report renders, rather than sitting on a loading skeleton.
6. Open a report from the left-hand list and verify it opens as before.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

https://github.com/user-attachments/assets/9e119253-a312-401e-b623-f4a2b1bf5748


src/pages/inbox/ReportRouteParamHandler.tsx
```diff
@@ -45,6 +49,7 @@ function ReportRouteParamHandler() {
             'openOnAdminRoom' in route.params && !!route.params.openOnAdminRoom,
             undefined,
             reportNameValuePairs,
+            reports,
```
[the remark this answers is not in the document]
wildan-m: Good catch — fixed in 6d6136cb0cb by returning early while the report name-value pairs are still loading, so an archived report can no longer be pinned into the route.

JmillsExpensify (commented):
No product review required.

src/pages/inbox/ReportRouteParamHandler.tsx
```diff
@@ -27,7 +28,11 @@ function ReportRouteParamHandler() {
     const route = useRoute<ReportScreenRoute>();
     const navigation = useNavigation();
     const {isBetaEnabled} = usePermissions();
-    const [reportNameValuePairs] = useOnyx(ONYXKEYS.COLLECTION.REPORT_NAME_VALUE_PAIRS);
+    const [reportNameValuePairs, reportNameValuePairsMetadata] = useOnyx(ONYXKEYS.COLLECTION.REPORT_NAME_VALUE_PAIRS);
+    // Subscribing to the reports collection (instead of relying on the module-scoped copy inside ReportUtils)
+    // makes this handler re-run once the reports finish loading, so a route that was created without a reportID
+    // recovers instead of staying stuck on the loading skeleton.
+    const [reports] = useOnyx(ONYXKEYS.COLLECTION.REPORT);
```
mananjadhav: **question:** After reportID is set (early return at 39–46), this still re-renders on every report write. Can we improvise this?

src/pages/inbox/ReportRouteParamHandler.tsx
```diff
@@ -40,11 +45,19 @@ function ReportRouteParamHandler() {
...
         }
 
+        // Archived status is stored in the name-value pairs, and a missing entry reads as "not archived".
+        // Resolving before they arrive could pin an archived report into the route, which is never revisited
+        // because the effect returns early once a reportID is set.
+        if (isLoadingOnyxValue(reportNameValuePairsMetadata)) {
+            return;
+        }
+
         const lastAccessedReportID = findLastAccessedReport(
```
mananjadhav: I am wondering if the behavior is changed somehow over here. Today the flow is no reportID -> skeleton. Reports start arriving in Onyx (may be not all at once).

As soon as `findLastAccessedReport` can pick something from the reports it currently sees, the handler calls `setParams({ reportID })`.

So if the real “last accessed” report hasn’t loaded yet, it could lock in some other report from the first partial batch. wdyt?

mananjadhav (commented):
@wildan-m I've reviewed and left 2 comments. Can you take a look at these?

src/pages/inbox/ReportRouteParamHandler.tsx
wildan-m: Good point — resolution now happens inside a `useOnyx` selector that short-circuits once the route has a reportID, so report writes no longer re-render this handler or run the scan.

src/pages/inbox/ReportRouteParamHandler.tsx
wildan-m: You are right — the guard now also waits for the reports collection to finish loading, so it can no longer pin a candidate picked from a partial batch.

mananjadhav (approved):

mananjadhav:
Quick bump @danieldoglas

danieldoglas (approved):
LGTM

PR #97645, opened 2026-08-02
Expense -Deleted split moved from 1:1 chat to Self DM with negative amount after clear cache

lorretheboy:
### Fixed Issues

$ https://github.com/Expensify/App/issues/92521
PROPOSAL: https://github.com/Expensify/App/issues/92521#issuecomment-5009545786

### Tests

Prerequisite: Account has Self DM.
Prerequisite 2: Have a private room created with at least one member invited.

1. Open the Expensify app.
2. Navigate to a private room conversation.
3. Tap on "+" button > "Split Expense"
4. Enter any amount and create the split expense.
5. Navigate to "Account" > "Troubleshoot" > "Clear cache and restart"
6. Return to "Inbox" and open the individual chat where split IOU was sent.
7. Long tap on IOU preview and delete it.
8. Navigate to Self DM.
9. Verify that Split deleted from individual chat should not be moved to Self DM.

### QA Steps

Prerequisite: Account has Self DM.
Prerequisite 2: Have a private room created with at least one member invited.

1. Open the Expensify app.
2. Navigate to a private room conversation.
3. Tap on "+" button > "Split Expense"
4. Enter any amount and create the split expense.
5. Navigate to "Account" > "Troubleshoot" > "Clear cache and restart"
6. Return to "Inbox" and open the individual chat where split IOU was sent.
7. Long tap on IOU preview and delete it.
8. Navigate to Self DM.
9. Verify that Split deleted from individual chat should not be moved to Self DM.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/7efb7a8e-0d0b-4f9c-89f8-e1f83de87219

</details>

<details>

https://github.com/user-attachments/assets/2850193d-66c3-4900-998a-252e31029bdc

</details>

<details>

https://github.com/user-attachments/assets/b753ad8f-a722-477f-9290-2407044a809f

</details>

<details>

https://github.com/user-attachments/assets/2d934b6c-30f2-432e-b508-ad6d998392e9

</details>

<details>

https://github.com/user-attachments/assets/feb62d64-04c2-49c4-ba51-62f8d4c80eb2

</details>


joekaufmanexpensify (dismissed):
Good for product 👍

src/pages/inbox/report/ContextMenu/BaseReportActionContextMenu.tsx
```diff
@@ -223,10 +223,12 @@ function BaseReportActionContextMenu({
...
     const {reportActions: paginatedReportActions} = usePaginatedReportActions(childReport?.reportID);
     const currentUserPersonalDetails = useCurrentUserPersonalDetails();
-    const transactionThreadReportID = useMemo(
-        () => getOneTransactionThreadReportID(childReport, childChatReport, paginatedReportActions ?? [], isOffline),
+    const transactionThreadReportAction = useMemo(
+        () => getOneTransactionThreadReportAction(childReport, childChatReport, paginatedReportActions ?? [], isOffline),
         [paginatedReportActions, isOffline, childReport, childChatReport],
     );
+    // Since we don't always create the transaction thread optimistically, we fall back to CONST.FAKE_REPORT_ID
+    const transactionThreadReportID = transactionThreadReportAction ? (transactionThreadReportAction.childReportID ?? CONST.FAKE_REPORT_ID) : undefined;
```
marufsharifi: This is a verbatim re-implementation of `getOneTransactionThreadReportID` in `src/libs/ReportActionsUtils.ts`, comment included:

```ts
function getOneTransactionThreadReportID(...args: Parameters<typeof getOneTransactionThreadReportAction>): string | undefined {
    const reportAction = getOneTransactionThreadReportAction(...args);
    if (reportAction) {
        // Since we don't always create transaction thread optimistically, we return CONST.FAKE_REPORT_ID
        return reportAction.childReportID ?? CONST.FAKE_REPORT_ID;
    }
}
```

Deriving the ID from the action instead of calling `getOneTransactionThreadReportID` a second time is the **right** call performance-wise — that helper re-runs the O(n) scan over `paginatedReportActions`. The only problem is that the derivation got copy-pasted rather than shared. Extract the last step:

```ts
// src/libs/ReportActionsUtils.ts
/** Since we don't always create the transaction thread optimistically, we fall back to CONST.FAKE_REPORT_ID. */
function getTransactionThreadReportIDFromAction(reportAction: ReportAction<'IOU'> | undefined): string | undefined {
    return reportAction ? (reportAction.childReportID ?? CONST.FAKE_REPORT_ID) : undefined;
}

function getOneTransactionThreadReportID(...args: Parameters<typeof getOneTransactionThreadReportAction>): string | undefined {
    return getTransactionThreadReportIDFromAction(getOneTransactionThreadReportAction(...args));
}
```

then here (add `getTransactionThreadReportIDFromAction` to the `@libs/ReportActionsUtils` import at the top of this file):

```suggestion
    const transactionThreadReportID = getTransactionThreadReportIDFromAction(transactionThreadReportAction);
```

**Why this matters:** the `FAKE_REPORT_ID` sentinel is branched on in at least four other places — `src/libs/SearchUIUtils.ts:2767`, `src/hooks/useLoadReportActions.ts:51`, `src/pages/inbox/ReportFetchHandler.tsx:299`, `src/pages/Search/SearchMoneyRequestReportPage.tsx:202`. If that convention is ever changed (e.g. to `undefined` plus a separate flag), the copy in this component keeps silently producing the old value and this exact bug regresses — with no compiler error to catch it.

src/pages/inbox/report/ContextMenu/BaseReportActionContextMenu.tsx
```diff
@@ -235,16 +237,26 @@ function BaseReportActionContextMenu({
 
     const requestParentReportAction = useMemo(() => {
         if (isMoneyRequestReport || isInvoiceReport) {
-            if (transactionThreadReportID === CONST.FAKE_REPORT_ID) {
-                return Object.values(childReportActions ?? {}).find((action) => action.actionName === CONST.REPORT.ACTIONS.TYPE.IOU && !isDeletedAction(action));
+            // The transaction thread report can be missing from Onyx (e.g. right after clearing the cache), so we can't look up its parent action.
+            // In that case we use the action the transaction thread was derived from, which applies the same request-action selection logic.
```
marufsharifi: The second line isn't accurate: `transactionThreadReportAction` doesn't "apply the same request-action selection logic" — it **is** the transaction thread's parent action. That distinction matters a lot to the next reader, because it's the difference between "this fallback is an approximation" (risky, someone will want to remove it) and "this fallback is an identity" (safe).

```suggestion
            // The transaction thread report can be missing from Onyx (e.g. right after clearing the cache), so its
            // parentReportActionID is unavailable. Fall back to the IOU action the thread was derived from — that action
            // is by definition the thread's parent action, so both paths resolve to the same report action.
```

src/pages/inbox/report/ContextMenu/BaseReportActionContextMenu.tsx
```diff
@@ -235,16 +237,26 @@ function BaseReportActionContextMenu({
...
+            if (transactionThreadReportID === CONST.FAKE_REPORT_ID || !transactionThreadReport?.parentReportActionID) {
+                return transactionThreadReportAction;
             }
-            if (!paginatedReportActions || !transactionThreadReport?.parentReportActionID) {
+            if (!paginatedReportActions) {
                 return undefined;
             }
             return paginatedReportActions.find((action) => action.reportActionID === transactionThreadReport.parentReportActionID);
```
marufsharifi: **NAB:** The `paginatedReportActions.find(...)` branch below is unreachable in any meaningful sense — it can only ever return `transactionThreadReportAction` or `undefined`:

1. `transactionThreadReportID` (line 231) is derived **from** `transactionThreadReportAction.childReportID`.
2. `transactionThreadReport` (line 233) is the Onyx report at that exact ID.
3. A transaction thread's `parentReportActionID` is by definition the `reportActionID` of the IOU action whose `childReportID` points at it — i.e. `transactionThreadReportAction.reportActionID`.
4. `getOneTransactionThreadReportAction` sources its return value from the **same** `paginatedReportActions` array you then `.find()` over — including the `getSendMoneyFlowAction` early return, which also iterates the passed-in `actions`.

So the `.find()` returns the identical object reference.

**And the one case where they diverge is a case this branch gets wrong.** During the optimistic→server reconciliation of a newly created expense, `transactionThreadReport.parentReportActionID` can still hold the optimistic action ID after that action has been replaced in `paginatedReportActions`. Your new guard doesn't catch this — the field is *present*, just stale — so `.find()` returns `undefined` and `moneyRequestAction` collapses to `undefined` again. That's the same class of bug this PR is fixing, just triggered by a create race instead of a cache clear.

**Concrete repro if left as-is:** create an expense in a 1:1 chat and long-press the preview within the ~1s optimistic window. `moneyRequestAction` is `undefined`, so **Hold/Unhold** (`ContextMenuActions.tsx:648`) and **Edit** (`:611`) silently vanish from the menu, and **Delete** (`:1583`, `showDeleteModal(reportID, moneyRequestAction ?? reportAction)`) once again targets the `REPORT_PREVIEW` action instead of the `IOU` action — the exact mechanism behind the orphaned negative-amount expense in Self DM.

```suggestion
            return transactionThreadReportAction;
```

src/pages/inbox/report/ContextMenu/BaseReportActionContextMenu.tsx
```diff
@@ -235,16 +237,26 @@ function BaseReportActionContextMenu({
...
-    }, [parentReportAction, isMoneyRequestReport, isInvoiceReport, paginatedReportActions, transactionThreadReport?.parentReportActionID, transactionThreadReportID, childReportActions]);
+    }, [
+        parentReportAction,
+        isMoneyRequestReport,
+        isInvoiceReport,
+        paginatedReportActions,
+        transactionThreadReport?.parentReportActionID,
+        transactionThreadReportID,
+        transactionThreadReportAction,
+    ]);
```
marufsharifi: `transactionThreadReportID` is now derived from `transactionThreadReportAction` (line 231), so listing both is redundant — the action already covers it. Harmless as written, just noise.

If you take the simplification from lines 242–248, this collapses to:

```suggestion
    }, [parentReportAction, isMoneyRequestReport, isInvoiceReport, transactionThreadReportAction]);
```

If you keep the current branching, at minimum drop `transactionThreadReportID` and keep the rest.

marufsharifi (commented):
### `### Offline tests` section is empty

but the section in the description is blank. Deleting an expense is a queued, optimistically-applied operation, so this is exactly the change where offline steps are load-bearing. Suggested minimum:

1. Reproduce through step 6 (private room split → clear cache → open the 1:1 chat), go offline, long-press the IOU preview → **Delete**. Verify the expense greys out with the pending style, Self DM stays clean, then reconnect and confirm the deletion persists with no phantom entry reappearing.
2. Go offline, create a split in a private room, immediately long-press the new preview. Verify the menu shows the correct options for the pending expense and that Delete removes the optimistic entry cleanly on reconnect. (This exercises the offline pending-delete branch in `getOneTransactionThreadReportAction`, `ReportActionsUtils.ts:1977`.)
3. Delete offline, force-quit, reopen while still offline — Onyx rehydrates from disk without the thread report, the same missing-report state as a cache clear. Verify the menu and pending state are still correct, then reconnect and confirm no negative-amount expense lands in Self DM.

tests/ui/components/BaseReportActionContextMenuTest.tsx
```diff
@@ -411,3 +418,171 @@ describe('BaseReportActionContextMenu hold/unhold action', () => {
...
+        amount: 1000,
+        currency: CONST.CURRENCY.USD,
+    },
+};
+
+/**
+ * Seeds a 1:1 chat containing a report preview action for a one-transaction IOU report, which is the shape the context
+ * menu sees when long-pressing an expense preview in a chat.
+ */
+async function seedMoneyRequestPreviewData({
```
marufsharifi, line 450: **NAB: Missing explicit return type.** 

Per [`STYLE.md`](https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md), we type function return values (except trivially simple functions). This helper does substantial setup, so an explicit `Promise<void>` documents intent:

```ts
async function seedMoneyRequestPreviewData({...}): Promise<void> {
```

Borderline since the return is `Promise<void>` — treat as optional, not a blocker.

tests/ui/components/BaseReportActionContextMenuTest.tsx
```diff
@@ -411,3 +418,171 @@ describe('BaseReportActionContextMenu hold/unhold action', () => {
...
+        reportID: previewIOUReportID,
+        amount: 1000,
+        currency: CONST.CURRENCY.USD,
+    });
+
+    mockPaginatedReportActions.reportActions = [iouAction];
+    await waitForBatchedUpdates();
+}
+
+async function renderMenuAndPressDelete() {
```
marufsharifi, line 515: **NAB: Missing explicit return type.** 

```ts
async function renderMenuAndPressDelete(): Promise<void> {
```

Same borderline note as the other helpers — `Promise<void>` arguably falls under the "trivially simple" carve-out in [`STYLE.md`](https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md), so this is optional.

tests/ui/components/BaseReportActionContextMenuTest.tsx
```diff
@@ -411,3 +418,171 @@ describe('BaseReportActionContextMenu hold/unhold action', () => {
...
+        />,
+    );
+
+    const onPress = await getContextMenuItemOnPress(CONST.SENTRY_LABEL.CONTEXT_MENU.DELETE);
+    await act(async () => {
+        onPress({});
+    });
+}
+
+function expectDeleteTargetedTheIOUAction() {
```
marufsharifi, line 531: **NAB: Missing explicit return type.** 

```ts
function expectDeleteTargetedTheIOUAction(): void {
```

Optional — a `void` side-effect assertion helper is close to the "trivially simple" exemption in [`STYLE.md`](https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md).

tests/ui/components/BaseReportActionContextMenuTest.tsx
```diff
@@ -411,3 +418,171 @@ describe('BaseReportActionContextMenu hold/unhold action', () => {
...
+    iouAction = iouCreateAction,
+}: {
+    /** When omitted, no transaction thread report exists in Onyx (the state right after clearing the cache). */
+    transactionThreadReport?: Record<string, unknown>;
+    iouAction?: ReportAction;
+}) {
+    await Onyx.clear();
+
+    await Onyx.merge(ONYXKEYS.SESSION, {accountID: currentUserAccountID, email: currentUserLogin});
+    await Onyx.merge(ONYXKEYS.PERSONAL_DETAILS_LIST, testPersonalDetails);
```
marufsharifi, line 461: **NAB: Duplicated Onyx-seeding boilerplate — optional**

This preamble repeats what `seedOnyxData` already does above (around line 159):

```ts
await Onyx.clear();
await Onyx.merge(ONYXKEYS.SESSION, {accountID: currentUserAccountID, email: currentUserLogin});
await Onyx.merge(ONYXKEYS.PERSONAL_DETAILS_LIST, testPersonalDetails);
// ...
await waitForBatchedUpdates();
```

The two seeders build genuinely different report topologies (hold/expense-report vs. report-preview-in-DM), so I'm **not** suggesting merging the full functions. Only the current-user init + trailing `waitForBatchedUpdates()` are identical.

**Suggested fix (optional):** extract just the shared bit into a small local helper both seeders call, e.g.:

```ts
async function seedCurrentUser() {
    await Onyx.clear();
    await Onyx.merge(ONYXKEYS.SESSION, {accountID: currentUserAccountID, email: currentUserLogin});
    await Onyx.merge(ONYXKEYS.PERSONAL_DETAILS_LIST, testPersonalDetails);
}
```

Note: `TestHelper.signInWithTestUser` is **not** a drop-in here — it's a network-mock sign-in, not the direct `Onyx.merge` seed these tests rely on. Fine to leave as-is if you prefer each scenario's seeder fully self-contained.

tests/ui/components/BaseReportActionContextMenuTest.tsx
```diff
@@ -411,3 +418,171 @@ describe('BaseReportActionContextMenu hold/unhold action', () => {
...
+const expenseTransactionID = '900';
+
+const iouCreateAction: ReportAction<typeof CONST.REPORT.ACTIONS.TYPE.IOU> = {
+    reportActionID: iouActionID,
+    actionName: CONST.REPORT.ACTIONS.TYPE.IOU,
+    actorAccountID: currentUserAccountID,
+    reportID: previewIOUReportID,
+    childReportID: transactionThreadReportID,
+    created: '2025-03-05 16:34:27',
+    // An action with an empty message array is treated as a legacy deleted comment by isDeletedAction()
```
marufsharifi: This comment references code this PR removes — `isDeletedAction` is dropped from `BaseReportActionContextMenu` in this same diff, so a reader following it into the component won't find it, leaving it unclear what still depends on the message being non-empty. Suggest pointing it at the code path that actually relies on it:

```suggestion
    // getOneTransactionThreadReportAction skips IOU actions with an empty message array (treated as deleted), so give this one a real message.
```

src/libs/ReportActionsUtils.ts
```diff
@@ -2003,16 +2003,20 @@ function getOneTransactionThreadReportAction(
     return iouRequestAction;
 }
 
+/**
+ * Resolves the transaction thread reportID from the IOU action the thread is derived from.
+ * Since we don't always create the transaction thread optimistically, we fall back to CONST.FAKE_REPORT_ID.
+ */
+function getTransactionThreadReportIDFromAction(reportAction: ReportAction<'IOU'> | undefined): string | undefined {
+    return reportAction ? (reportAction.childReportID ?? CONST.FAKE_REPORT_ID) : undefined;
```
marufsharifi, line 2012: **NAB:** the ternary + `??` on one line is a little dense for the `undefined → undefined` edge case. An early return separates the two concerns:

```suggestion
    if (!reportAction) {
        return undefined;
    }
    return reportAction.childReportID ?? CONST.FAKE_REPORT_ID;
```

Minor — current form is fine if you prefer the one-liner.

marufsharifi (changes requested):

tests/ui/components/BaseReportActionContextMenuTest.tsx
lorretheboy: I think we don't need this, I follow pattern in other tests. Besides, this is test so I don't think we should update it

src/libs/ReportActionsUtils.ts
lorretheboy: I think it is fine to keep the current logic

tests/ui/components/BaseReportActionContextMenuTest.tsx
lorretheboy: It is test file, besides, I don't think we need explicitly define void type (we do like my approach in many places)

tests/ui/components/BaseReportActionContextMenuTest.tsx
lorretheboy: It is test file, besides, I don't think we need explicitly define void type (we do like my approach in many places)

tests/ui/components/BaseReportActionContextMenuTest.tsx
lorretheboy: It is test file, besides, I don't think we need explicitly define void type (we do like my approach in many places)

marufsharifi (approved):
LGTM

src/pages/inbox/report/ContextMenu/BaseReportActionContextMenu.tsx
```diff
@@ -223,28 +224,24 @@ function BaseReportActionContextMenu({
...
+            // The transaction thread report can be missing from Onyx (e.g. right after clearing the cache) or hold a stale
+            // parentReportActionID (e.g. during the optimistic->server reconciliation of a newly created expense), so we don't look it up.
+            // The IOU action the thread was derived from is by definition the thread's parent action, so it always resolves to the same report action.
+            return transactionThreadReportAction;
         }
         return parentReportAction;
-    }, [parentReportAction, isMoneyRequestReport, isInvoiceReport, paginatedReportActions, transactionThreadReport?.parentReportActionID, transactionThreadReportID, childReportActions]);
+    }, [parentReportAction, isMoneyRequestReport, isInvoiceReport, transactionThreadReportAction]);
 
     const moneyRequestAction = transactionThreadReportID ? requestParentReportAction : parentReportAction;
```
Valforte: This is looking like dead code, `requestParentReportAction` now returns `transactionThreadReportAction` whenever the guard passes, and `transactionThreadReportID` is truthy iff `transactionThreadReportAction` is defined (FAKE_REPORT_ID = 'FAKE_REPORT_ID', truthy)

```suggestion
    // The transaction thread report can be missing from Onyx (e.g. right after clearing the cache) or hold a stale
    // parentReportActionID (e.g. during the optimistic->server reconciliation of a newly created expense), so we don't look it up.
    // The IOU action the thread was derived from is by definition the thread's parent action, so it always resolves to the same report action.
    const moneyRequestAction = transactionThreadReportAction ?? parentReportAction;
```

Valforte (changes requested):
Small simplifaction suggestion

Valforte (approved):

Issue #96967, opened 2026-07-24
[Due for payment 2026-09-07] [$375] Report incorrectly shows a policy violation (red-dot/RBR) after a split is reverted, even though there are no actual violations remaining.


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding. To join our Slack channel, fill out [this form](https://forms.gle/Q7hnhUJPnQCK7Fe56).
___

**Version Number:**
**Reproducible in staging?:** Needs Reproduction (Unable to reproduce)
**Reproducible in production?:** Needs Reproduction
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @lschurr 
**Slack conversation** (hyperlinked to channel name): [#Expensify Bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1784843356985579)

## Action Performed:

1. Open a report and move some expenses onto it.
2. Split an expense on the report.
3. Revert the split.
4. Open the report.

## Expected Result:
The report’s RBR (red-dot) indicator matches what the opened report can actually render. Once a transaction is deleted/removed (including a reverted split child), its violations should no longer keep the indicator lit.
## Actual Result:
The report shows a policy-violation red-dot indicator (RBR), but when the report is opened there are no per-transaction violations to fix. A lingering violation on the reverted/removed transaction keeps the dot lit while the opened report looks clean.

## Workaround:
Unknown

## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: <a href="https://www.upwork.com/jobs/~022080654765149918620">https://www.upwork.com/jobs/~022080654765149918620</a></li>
[...]

PR #97706, opened 2026-08-03
Filter DELETE-pending transactions out of LHN RBR aggregation

dilshodmackbook-sketch:
### Explanation of Change

Reverting a split on a report removes the split child transaction(s) optimistically — they are marked with `pendingAction: DELETE` and stay in Onyx (with their `reportID` and their `transactionViolations` entry) until the server confirms and nulls them.

The LHN red-dot (RBR) is computed by `getViolatingReportIDForRBRInLHN` in `src/libs/ReportUtils.ts`. It built the report's transaction set with a plain `getReportTransactions(...)` and did **not** exclude transactions that are pending deletion. That set is passed down into `hasVisibleViolationsForUser` / `hasViolations` / `hasWarningTypeViolations` / `hasNoticeTypeViolationsForRBRInLHN`, each of which reads `transactionViolations_<transactionID>` for every transaction — so a `DELETE`-pending transaction whose violation entry still existed kept the red dot lit.

Meanwhile the report contents (and the sibling "is this report empty" check a few lines above in the same file) already filter out `DELETE`-pending transactions, which is why the opened report looked clean. The LHN aggregation and the report view disagreed about whether the reverted transaction still existed.

This PR makes the RBR aggregation ignore transactions that are already pending deletion, mirroring the empty-report check in the same function:

```ts
const transactions = getReportTransactions(potentialReport.reportID).filter((transaction) => transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE);
```

Because this filtered array is the one handed to every downstream violation check, the single change fixes the whole RBR path in one place and keeps the LHN consistent with what the opened report actually renders — for reverted splits and any other flow that leaves a transaction queued for deletion.

A unit test was added to `tests/unit/ReportUtilsTest.ts` covering the case where the only violating transaction is `DELETE`-pending, asserting `getViolatingReportIDForRBRInLHN` returns `null`.

### Fixed Issues

$ https://github.com/Expensify/App/issues/96967
PROPOSAL: https://github.com/Expensify/App/issues/96967#issuecomment-5070690516

### Tests

1. Open a report and move some expenses onto it.
2. Split an expense on the report.
3. Revert the split.
4. Open the report and observe there are no per-transaction violations to fix.
5. Verify the report's RBR (red-dot) indicator in the LHN is **not** shown while the reverted split's child transaction is still queued for deletion.
6. Verify reports that genuinely have violations still show the RBR indicator.

### Offline tests

1. Go offline.
2. Split an expense on a report, then revert the split.
3. Verify the LHN does not show a policy-violation red-dot for that report while the reverted child transaction sits pending deletion.
4. Go back online and verify the state stays consistent after the server confirms the removal.

### QA Steps

1. Open a report and move some expenses onto it.
2. Split an expense on the report.
3. Revert the split.
4. Open the report and confirm there are no per-transaction violations to fix.
5. Verify the report no longer shows a policy-violation red-dot (RBR) indicator in the LHN.
6. Verify reports that do have real violations still display the RBR indicator.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

https://github.com/user-attachments/assets/5b46ab0c-863e-4c2d-820b-e20511021e22

</details>

<details>

https://github.com/user-attachments/assets/e37da14d-bead-4bc8-8429-aed7ab232789

</details>

<details>

https://github.com/user-attachments/assets/a7098e05-a127-4f2a-8d47-f11a7eeee6c6

</details>

<details>

https://github.com/user-attachments/assets/a1f260e4-e8e8-4f53-8113-fcbe1a6a5f0d

</details>

<details>

https://github.com/user-attachments/assets/21c0c63c-dd05-44a9-9331-004efa9a8c76

</details>


src/libs/ReportUtils.ts
```diff
@@ -9460,7 +9451,9 @@ function getViolatingReportIDForRBRInLHN(report: OnyxEntry<Report>, transactionV
                 return false;
             }
             const policy = allPolicies?.[`${ONYXKEYS.COLLECTION.POLICY}${potentialReport.policyID}`];
-            const transactions = getReportTransactions(potentialReport.reportID);
+            // Ignore transactions that are already pending deletion (e.g. a reverted split child) so the LHN RBR stays
+            // consistent with what the opened report renders, which also filters out DELETE-pending transactions.
+            const transactions = getReportTransactions(potentialReport.reportID).filter((transaction) => transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE);
```
marufsharifi: **Reuse the existing `isTransactionPendingDelete` predicate instead of the raw comparison.**

`TransactionUtils` already exports a named predicate for exactly this check:

https://github.com/Expensify/App/blob/main/src/libs/TransactionUtils/index.ts#L3001-L3003

For a transaction with a set `pendingAction` it's behaviorally identical to this line (the `pendingFields` branch in `getTransactionPendingAction` only ever yields `UPDATE`, never `DELETE`), so this is a safe swap:

```ts
const transactions = getReportTransactions(potentialReport.reportID).filter((transaction) => !isTransactionPendingDelete(transaction));
```

`isTransactionPendingDelete` isn't imported into `ReportUtils.ts` yet, but this file already imports heavily from `TransactionUtils`, so just add it to that import. This keeps the intent ("skip transactions pending deletion") readable rather than hardcoding the `CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE` comparison.

marufsharifi: **This "drop DELETE-pending transactions" filter is now the third copy of the same inline pattern in this file.**

The same raw `pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE` filter over a transaction list already lives at:

- https://github.com/Expensify/App/blob/main/src/libs/ReportUtils.ts#L9300 (treat report as empty once removal is queued)
- https://github.com/Expensify/App/blob/main/src/libs/ReportUtils.ts#L13040-L13042 (`doesReportContainRequestsFromMultipleUsers`, gated by `shouldExcludeDeletedTransactions`)

The comment on this new line even notes the opened-report path does the same filtering — that's the smell. Switching all three to `!isTransactionPendingDelete(...)` (see the other comment) collapses them onto one shared predicate.

Extracting a small `getActiveReportTransactions` / `filterOutDeletedTransactions` helper in `TransactionUtils` would make the reuse fully explicit, but that's a larger refactor and out of scope here — the predicate swap is the minimal DRY fix for this PR.

marufsharifi: Filter runs before the cheap `isOpenOrProcessingReport` early-return

`getReportTransactions(...).filter(...)` executes before the `isOpenOrProcessingReport(potentialReport)` guard just below (line 9459). For any report that isn't open/processing, both the transaction lookup and the new allocation-and-iterate are wasted. This is a pre-existing ordering issue — this PR only widens the wasted work by adding the `.filter()` to it. Per-report transaction lists are small and bounded, so the absolute cost is negligible even across many LHN rows; flagging only for a cheap cleanup.

Move the transactions computation below the early-return so it runs only for reports that survive the guard:

```ts
// Allow both open and processing reports to show RBR for violations
if (!isOpenOrProcessingReport(potentialReport)) {
    return false;
}

const policy = allPolicies?.[`${ONYXKEYS.COLLECTION.POLICY}${potentialReport.policyID}`];
const transactions = getReportTransactions(potentialReport.reportID).filter((transaction) => transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE);
```

marufsharifi: **NAB:** This line chains `getReportTransactions(...).filter(...)` with a long inline predicate, pushing it well past the width of the surrounding lines. The meaningful `pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE` check ends up buried at the tail of the line.

Consider extracting a named predicate so the intent reads at a glance:

```

const isNotPendingDeletion = (transaction: Transaction) => transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE;
const transactions = getReportTransactions(potentialReport.reportID).filter(isNotPendingDeletion);

File: tests/unit/ReportUtilsTest.ts [hunk 2]
```diff
@@ -15399,6 +15383,96 @@ describe('ReportUtils', () => {
...
+                Onyx.merge(`${ONYXKEYS.COLLECTION.TRANSACTION}${transaction.transactionID}`, transaction),
+                Onyx.merge(transactionViolationsKey, transactionViolationsCollection[transactionViolationsKey]),
+            ]);
+            await waitForBatchedUpdates();
+
+            const result = getViolatingReportIDForRBRInLHN(chatReport, transactionViolationsCollection);
+            expect(result).toBeNull();
+
+            await Onyx.clear();
+        });
```
marufsharifi, line 15666: **Missing test: mixed DELETE-pending + live transaction on the same report.**

Both this new test and the existing positive test use a **single** transaction, so nothing proves the filter is per-transaction (drops only the reverted child) rather than "if any transaction is DELETE-pending, drop the whole report's RBR." That's the exact split-report shape this fix targets.

Please add, right after this test:

```ts
it('should still surface RBR when a report has a live violating transaction alongside a DELETE-pending one', async () => {
    // ...same policy / chatReport / expenseReport setup as the pending-delete test...

    // Two violating transactions on the SAME expense report:
    // one reverted split child (DELETE-pending) and one still live.
    const deletedChild: Transaction = {
        ...createRandomTransaction(809),
        transactionID: 'transaction-rbr-mixed-deleted',
        reportID: expenseReportID,
        amount: 5000,
        currency: CONST.CURRENCY.USD,
        status: CONST.TRANSACTION.STATUS.POSTED,
        reimbursable: true,
        pendingAction: CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE,
    };
    const liveChild: Transaction = {
        ...createRandomTransaction(810),
        transactionID: 'transaction-rbr-mixed-live',
        reportID: expenseReportID,
        amount: 5000,
        currency: CONST.CURRENCY.USD,
        status: CONST.TRANSACTION.STATUS.POSTED,
        reimbursable: true,
        // no pendingAction
    };

    // Give BOTH transactions a MISSING_CATEGORY violation, then merge policy/reports/
    // transactions/violations into Onyx as in the tests above.

    const result = getViolatingReportIDForRBRInLHN(chatReport, transactionViolationsCollection);
    expect(result).toBe(expenseReportID); // RBR still shown because the live child violates

    await Onyx.clear();
});
```

**Verdict:** Covered 2/3 changed behaviors — the mixed-transaction case is the one gap, and it's the realistic split-report scenario the fix is about.

tests/unit/ReportUtilsTest.ts
```diff
@@ -15399,6 +15383,96 @@ describe('ReportUtils', () => {
...
+                Onyx.merge(`${ONYXKEYS.COLLECTION.POLICY}${policyID}`, policyData),
+                Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${chatReport.reportID}`, chatReport),
+                Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${expenseReport.reportID}`, expenseReport),
+                Onyx.merge(`${ONYXKEYS.COLLECTION.TRANSACTION}${transaction.transactionID}`, transaction),
+                Onyx.merge(transactionViolationsKey, transactionViolationsCollection[transactionViolationsKey]),
+            ]);
+            await waitForBatchedUpdates();
+
+            const result = getViolatingReportIDForRBRInLHN(chatReport, transactionViolationsCollection);
+            expect(result).toBeNull();
```
marufsharifi, line 15546: **Add a positive control so this test can't pass vacuously.**

Right now the test only asserts the `DELETE`-pending case returns `null`. That doesn't prove the *filter* is what produces the `null` — if any unrelated gate in `getViolatingReportIDForRBRInLHN` (`isCurrentUserSubmitter`, `isOpenOrProcessingReport`, or `shouldShowViolation` rejecting `MISSING_CATEGORY`) were the real reason, the test would still pass while testing nothing, and it would keep passing even if the fix were reverted in a way that broke a different gate.

Please assert that the *same* setup lights the RBR when the transaction is **not** pending deletion. Cheapest version — flip `pendingAction` and re-assert right after line 15471:

```ts
// Positive control: the same violating transaction lights the RBR when it isn't pending deletion.
await Onyx.merge(`${ONYXKEYS.COLLECTION.TRANSACTION}${transactionID}`, {pendingAction: null});
await waitForBatchedUpdates();
expect(getViolatingReportIDForRBRInLHN(chatReport, transactionViolationsCollection)).toBe(expenseReportID);
```

This guarantees the test actually exercises the new filter (and would fail on `main`), instead of relying on QA step 6 to cover the positive case.

marufsharifi (changes requested):

garrettmknight (dismissed):
Looks good from a UX perspective

src/libs/ReportUtils.ts
```diff
@@ -9326,7 +9327,7 @@ function getPolicyIDsWithEmptyReportsForAccount(
         }
 
         // Ignore transactions that are already pending deletion so we treat the report as empty once the removal is queued.
-        const transactions = (reportsTransactionsParam[report.reportID] ?? []).filter((transaction) => transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE);
+        const transactions = (reportsTransactionsParam[report.reportID] ?? []).filter((transaction) => !isTransactionPendingDelete(transaction));
```
marufsharifi: This swap to `isTransactionPendingDelete(transaction)` is a drive-by refactor.

This function (`getPolicyIDsWithEmptyReportsForAccount`, the "empty report" check) isn't part of the LHN RBR path this PR fixes. I confirmed the new helper is functionally equivalent to the previous inline check — `getTransactionPendingAction` only diverges from the raw `pendingAction` for the `UPDATE` case, never `DELETE` — so this is a cosmetic consistency change, not a behavior fix.

Per Expensify's guidance, a refactor should land as its own separate change rather than bundled into an unrelated fix. Suggest reverting to the original inline check here, or splitting the helper cleanup into its own PR.

```ts
// before (in scope to leave untouched)
const transactions = (reportsTransactionsParam[report.reportID] ?? []).filter((transaction) => transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE);
```

src/libs/ReportUtils.ts
```diff
@@ -13064,9 +13067,7 @@ function hasExportError(reportActions: OnyxEntry<ReportActions> | ReportAction[]
 }
 
 function doesReportContainRequestsFromMultipleUsers(iouReport: OnyxEntry<Report>, shouldExcludeDeletedTransactions = false): boolean {
-    const transactions = getReportTransactions(iouReport?.reportID).filter(
-        (transaction) => !shouldExcludeDeletedTransactions || transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE,
-    );
+    const transactions = getReportTransactions(iouReport?.reportID).filter((transaction) => !shouldExcludeDeletedTransactions || !isTransactionPendingDelete(transaction));
```
marufsharifi: Same drive-by refactor as in `getPolicyIDsWithEmptyReportsForAccount`.

`doesReportContainRequestsFromMultipleUsers` is unrelated to the LHN RBR aggregation being fixed here. Replacing the inline `transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE` with `!isTransactionPendingDelete(transaction)` is functionally equivalent (verified), so it adds diff noise without changing behavior.

Suggest keeping the original inline check in this function and, if the team wants the helper applied consistently, doing that in a dedicated refactor PR.

```ts
// before (in scope to leave untouched)
const transactions = getReportTransactions(iouReport?.reportID).filter(
    (transaction) => !shouldExcludeDeletedTransactions || transaction.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE,
);
```

tests/unit/ReportUtilsTest.ts
```diff
@@ -15458,6 +15458,213 @@ describe('ReportUtils', () => {
...
 
+        it('should return null when the only violating transaction is pending deletion (e.g. a reverted split child)', async () => {
+            await Onyx.clear();
+
+            const policyID = 'policy-rbr-deleting';
+            const chatReportID = 'chat-rbr-deleting';
+            const expenseReportID = 'expense-rbr-deleting';
+            const transactionID = 'transaction-rbr-deleting';
+
+            const policyData: Policy = {
```
marufsharifi, line 15469: **Duplicated test scaffold — reuse a shared factory instead of copy-pasting the fixtures**

This `policyData` / `chatReport` / `expenseReport` block is a near-verbatim copy of the existing test `'should return the violating report ID for an open expense report with violations'` (`tests/unit/ReportUtilsTest.ts:15383`–`15421`): same `TEAM` policy with the admin `employeeList`, same `createPolicyExpenseChat` chat report shape, and the same open `createExpenseReport` (`managerID: 42`, `total`, `stateNum/statusNum: OPEN`). Only the string IDs and `name` differ.

The whole `describe('getViolatingReportIDForRBRInLHN')` block now repeats these fixtures 4+ times, and this PR adds two more copies.

**Suggested fix:** extract a factory inside the `describe` block, e.g.

```ts
async function setupRbrScenario({suffix, transactions}: {suffix: string; transactions: Transaction[]}) {
    const policyID = `policy-rbr-${suffix}`;
    // build policy/chat/expense report from the shared shape, merge everything into Onyx,
    // return {chatReport, expenseReportID, transactionViolationsCollection}
}
```

and have this test (and its siblings) call it. Since this is pre-existing duplication the PR is only following — not introducing a new pattern — the cleanest sequencing is a separate refactor of the block first, per Expensify's "refactor-to-reuse first" rule. Flagging as **low** because these two new tests add two more copies rather than starting the cleanup.

tests/unit/ReportUtilsTest.ts
```diff
@@ -15458,6 +15458,213 @@ describe('ReportUtils', () => {
...
+        it('should still surface RBR when a report has a live violating transaction alongside a DELETE-pending one', async () => {
+            await Onyx.clear();
+
+            const policyID = 'policy-rbr-mixed';
+            const chatReportID = 'chat-rbr-mixed';
+            const expenseReportID = 'expense-rbr-mixed';
+            const deletedTransactionID = 'transaction-rbr-mixed-deleted';
+            const liveTransactionID = 'transaction-rbr-mixed-live';
+
+            const policyData: Policy = {
```
marufsharifi, line 15566: **Same duplicated scaffold as the sibling test above**

This `policyData` / `chatReport` / `expenseReport` setup is identical (apart from IDs, `name`, and `total: 10000`) to the block in the `'…pending deletion…'` test at `tests/unit/ReportUtilsTest.ts:15469`–`15507` and to the existing positive test at `15383`–`15421`.

**Suggested fix:** consume the shared `setupRbrScenario(...)` factory proposed in the comment on line 15469, passing the two transactions (`deletedTransaction`, `liveTransaction`) in. That collapses this ~40-line preamble to a single call and keeps the RBR test suite DRY.

tests/unit/ReportUtilsTest.ts
```diff
@@ -15458,6 +15458,213 @@ describe('ReportUtils', () => {
             await Onyx.clear();
         });
 
+        it('should return null when the only violating transaction is pending deletion (e.g. a reverted split child)', async () => {
```
marufsharifi, line 15461: **Extract the shared scenario setup to reduce duplication**

These two new tests are ~100 lines each and share ~80% identical setup — `policyData`, `chatReport`, and `expenseReport` are near-verbatim copies (same shape, same `managerID: 42`, same currency/state fields), differing only in IDs and totals. A reader has to diff the two blocks line-by-line to find what actually distinguishes the scenarios (one DELETE-pending transaction vs. one DELETE-pending + one live).

Consider a shared builder for the common scaffold:

```ts
function buildRbrScenario({policyID, chatReportID, expenseReportID, total}: {...}) {
    return {policyData, chatReport, expenseReport};
}
```

so each test body reduces to its distinguishing transactions + violations + assertions. The intent of each test then reads at a glance.

tests/unit/ReportUtilsTest.ts
```diff
@@ -15458,6 +15458,213 @@ describe('ReportUtils', () => {
...
+                    [currentUserEmail]: {
+                        role: CONST.POLICY.ROLE.ADMIN,
+                    },
+                },
+                owner: currentUserEmail,
+                isPolicyExpenseChatEnabled: true,
+            };
+
+            const chatReport: Report = {
+                ...createPolicyExpenseChat(807),
```
marufsharifi, line 15487: **Duplicated seed numbers across the two new tests**

Both new tests seed with `createPolicyExpenseChat(807)` / `createExpenseReport(808)` — and the first test also uses `createRandomTransaction(807)`, colliding with the chat's `807`. It's harmless because each test does `await Onyx.clear()` and overrides `reportID` / `transactionID` explicitly, but reusing the same seeds across sibling tests is mildly confusing and diverges from the existing positive test's `803`/`804`. Consider unique seeds per test (e.g. `809`/`810`, `811`/`812`) for readability. Non-blocking.

marufsharifi (changes requested):

src/libs/ReportUtils.ts
dilshodmackbook-sketch: Good catch. Reverted back to the inline `pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE` check here — the helper swap was out of scope. Only `getViolatingReportIDForRBRInLHN` keeps `isTransactionPendingDelete`, since that's the actual LHN RBR path this PR fixes.

src/libs/ReportUtils.ts
dilshodmackbook-sketch: Reverted this one too, back to the original inline check. Keeping the change scoped to the RBR fix.

tests/unit/ReportUtilsTest.ts
dilshodmackbook-sketch: Agreed. Since this is pre-existing duplication these new tests only follow, I'll extract the shared scaffold into a factory as a separate refactor per the refactor-to-reuse-first rule — same reasoning as the helper reverts above. Keeping this PR scoped to the RBR fix for now.

tests/unit/ReportUtilsTest.ts
dilshodmackbook-sketch: Same as above — I'll fold this into the factory refactor follow-up rather than starting the cleanup mid-fix.

tests/unit/ReportUtilsTest.ts
dilshodmackbook-sketch: Same here, covered by that same factory refactor follow-up.

tests/unit/ReportUtilsTest.ts
dilshodmackbook-sketch: Fixed. Gave each test unique seeds — 809 for the pending-delete test and 810-813 for the mixed one — so the chat/report/transaction seeds no longer collide.

garrettmknight (approved):

marufsharifi (approved):
LGTM

chuckdries (approved):
Clean!

PR #97708, opened 2026-08-03
[Payment due @huult] [NO QA] Add Amount debited and Amount reimbursed filters, columns and sorting

ishpaul777:
### Explanation of Change

A cross-border reimbursement moves two amounts: the company is debited in the currency it settles in, and the employee is credited in the currency of their deposit account. [#97167](https://github.com/Expensify/App/pull/97167) shows those on withdrawal groups in the reconciliation dashboard. This adds them as filters and as columns on expense reports.

- Two new numeric filters, **Amount debited** and **Amount reimbursed**, sitting next to Total and behaving like it: equal to, greater than, less than, and a range. Query syntax is `amountDebited` / `amountReimbursed`, with `amount-debited` / `amount-reimbursed` as the readable form used in saved searches and URLs.
- Two new optional columns of the same names in the expense report view, each rendered in the currency stored alongside its amount rather than the report's own currency, and each sortable by clicking its header or picking it from **Sort by**.

Decisions worth a look:

- A column is hidden unless a loaded row carries both the amount and its currency, so a workspace that never pays across borders does not gain two permanently empty columns. Amounts are never displayed without their currency, since the two sides are denominated differently and neither matches the row's own currency.
- Sorting ranks the amounts as recorded, without converting between currencies, which is what the backend does and what Total already does. So a report settled in one currency can rank above a report worth more in another, and reports with no conversion sort first ascending and last descending.
- A column that is being sorted by stays visible even when nothing on the page fills it. Ascending puts the reports without that amount first, so a first page of domestic reports would otherwise drop the column and take both its header and its **Sort by** entry with it, leaving the sort applied with no way to change it.
- The new keys are placed ahead of `amount` in the parser grammar, otherwise `amount` matches first and swallows `amountDebited` as `amount` followed by stray text.
- Neither the filters nor the columns are offered on an expense search. A payment pays a whole report, so an expense has no amount of its own on either side, and a per-expense column would repeat its report's conversion on every row next to that expense's own amount, reading as though the expense itself had been converted. This matches `total`, which is also absent from the expense filter menu.

Needs [Expensify/Auth#23431](https://github.com/Expensify/Auth/pull/23431), which accepts the two filter keys and the two sort keys. Stacked on [#97167](https://github.com/Expensify/App/pull/97167), so only eleven commits belong to this PR, and one of those - "Restore the currency formatter the cross-border payment messages use" - is the same fix as [#97723](https://github.com/Expensify/App/pull/97723), carried here only so this branch can typecheck until that lands on main.

### Fixed Issues

$ https://github.com/Expensify/Expensify/issues/655933
PROPOSAL: N/A, internal

### Tests

Requires a workspace whose reports were paid with a currency conversion, since Auth records these amounts on the payment. Locally that means a workspace on a Corpay withdrawal account paying a member whose deposit account is in another currency, with at least one report left unconverted for comparison.

1. Open Search and switch the type to **Expense reports**.
2. Open the filters, confirm **Amount debited** and **Amount reimbursed** appear next to **Total**, and set Amount debited to _equal to_ the debited amount of one of the converted reports.
3. Verify only that report is returned, and that the readable query in the search bar reads `amount-debited:<value>`.
4. Change the filter to _greater than_ a value between the two converted reports. Verify only the larger one is returned, and that the report which was never paid across borders is returned by neither.
5. Repeat steps 2 to 4 with **Amount reimbursed**. Verify it ranks the reports independently of Amount debited, since the two sides are in different currencies.
6. Open the column picker and enable **Amount debited** and **Amount reimbursed**. Verify each converted report shows its amount in its own currency, and that the report that was never converted shows blank cells rather than a zero.
7. Click the **Amount debited** header. Verify the reports reorder by that amount, that clicking again reverses them, and that the report which was never converted sits at the top ascending and the bottom descending. Repeat on **Amount reimbursed** and verify it produces a different order, since the two sides are in different currencies.
8. With **Amount debited** as the active sort, narrow the search so the page holds only reports that were never converted. Verify the column and its header stay put rather than disappearing, and that **Sort by** still lists it so the sort can be changed.
9. Switch the type to **Expenses**. Verify neither filter is offered in the filter menu and neither column is offered in the column picker, and that switching type with either filter applied drops it from the query rather than sending it to the server.
10. Open a workspace that has never paid across borders. Verify neither column appears in the picker, so no empty columns are added.
11. Type `amountDebited:100` directly into the search bar and verify it parses as a filter chip rather than as free text, then reload the page and verify the filter survives in the URL.

### Offline tests

Search results come from the server, so with no connection the list keeps showing the last loaded results. Applying either filter offline shows the offline indicator and no results change until the connection returns, the same as every other search filter. Column selection is local, so enabling or disabling the two columns works offline against already-loaded rows.

### QA Steps

Same as Tests, on a workspace configured for global reimbursement whose reports have been paid across borders.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/SearchUIUtils.ts
```diff
@@ -6176,6 +6189,12 @@ function getColumnsToShow({
                 return true;
             }
 
+            // The column the results are sorted by has to stay, even with nothing to show in it. Both the header and the
+            // Sort by option come from this list, so dropping it would leave the sort applied with no way to change it.
+            if (column === sortBy) {
+                return true;
```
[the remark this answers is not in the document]
ishpaul777: Good catch, fixed in 7d69efc370c. `getCSVExportParameters` now passes `sortBy: queryJSON?.sortBy`, so Export current view keeps the sorted conversion column instead of dropping a column the table still shows. The other `getColumnsToShow` callers don't need it: `GroupHeader` returns the parent's `columns` for expense-report type, and the remaining ones pass transaction-type data or a transaction array, neither of which reaches the report column filtering.

src/CONST/index.ts
```diff
@@ -7129,6 +7129,8 @@ const CONST = {
                     SUBMITTER_USER_ID: this.TABLE_COLUMNS.SUBMITTER_USER_ID,
                     SUBMITTER_PAYROLL_ID: this.TABLE_COLUMNS.SUBMITTER_PAYROLL_ID,
                     ORDER_DEAL_NUMBERS: this.TABLE_COLUMNS.ORDER_DEAL_NUMBERS,
+                    AMOUNT_DEBITED: this.TABLE_COLUMNS.AMOUNT_DEBITED,
+                    AMOUNT_REIMBURSED: this.TABLE_COLUMNS.AMOUNT_REIMBURSED,
```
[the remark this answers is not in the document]
ishpaul777: Not changing this one, since it's the established behaviour for this whole column family rather than something these two entries introduce. `SUBMITTER_USER_ID`, `SUBMITTER_PAYROLL_ID` and `ORDER_DEAL_NUMBERS` are already listed unconditionally in the same `TYPE_CUSTOM_COLUMNS.EXPENSE_REPORT` block and are already value-filtered by the same `reportCustomColumns` branch in `getColumnsToShow`, so selecting one with no matching data has always left it checked in the picker with no table column. Making the picker data-aware for only the two new columns would make them behave differently from the three siblings, and doing it for all five would change shipped behaviour for columns outside the scope of this PR. Worth a separate issue if we want the picker to be data-aware across the board.

JmillsExpensify (dismissed):
Part of an existing product initiative.

src/components/Search/SearchList/ListItem/ExpenseReportListItemRow/ExpenseReportListItemRowWide.tsx
```diff
@@ -232,6 +233,26 @@ function ExpenseReportListItemRowWide({
                 <TextCell text={orderDealNumbers} />
             </View>
         ),
+        [CONST.SEARCH.TABLE_COLUMNS.AMOUNT_DEBITED]: (
+            <View style={[StyleUtils.getReportTableColumnStyles(CONST.SEARCH.TABLE_COLUMNS.AMOUNT_DEBITED)]}>
+                {!!debitedAmount && !!debitedCurrency && (
```
huult, line 238: Can I ask: if debitedAmount and creditedAmount are both 0, should this still be displayed?

ishpaul777: No. Nothing shows for those.

src/libs/SearchUIUtils.ts
```diff
@@ -6078,6 +6113,12 @@ function getColumnsToShow({
                 return true;
             }
 
+            // The column the results are sorted by has to stay, even with nothing to show in it. Both the header and the
+            // Sort by option come from this list, so dropping it would leave the sort applied with no way to change it.
```
huult: @ishpaul777 Can we shorten this comment? I don’t think we need such a long comment here.

src/libs/SearchUIUtils.ts
```diff
@@ -341,6 +341,10 @@ const expenseReportColumnNamesToSortingProperty: ExpenseReportSorting = {
     [CONST.SEARCH.TABLE_COLUMNS.FROM]: 'formattedFrom' as const,
     [CONST.SEARCH.TABLE_COLUMNS.TO]: 'formattedTo' as const,
     [CONST.SEARCH.TABLE_COLUMNS.TOTAL]: 'total' as const,
+    // Both the backend page selection and this local sort rank the amounts as stored, without converting between
+    // currencies, so a report can outrank one that is worth more in another currency.
```
huult: Can we shorten this comment? I don’t think we need such a long comment here.

ishpaul777: done!

src/libs/SearchUIUtils.ts
ishpaul777: done

huult (approved):
LGTM.

madmax330 (approved):

PR #97823, opened 2026-08-04
Fix: return focus and soft keyboard to description input on iOS after canceling discard modal

MelvinBot:
### Explanation of Change

On the expense **Description** step, canceling the "Discard changes?" modal is supposed to return focus (and the soft keyboard) to the description input. On iOS App it didn't — the keyboard never came back.

**Root cause:** opening the "Discard changes?" modal resigns the native soft keyboard on iOS — the modal takes over as the active responder and the IME is dismissed — while the description `TextInput` keeps reporting `isFocused() === true` at the JS layer (the field's focus/editable state is toggled natively, but React Native's focus state is preserved). So the actual driver of the missing keyboard is the **native keyboard being resigned**, not the `isFocused()` value itself.

The `onCancel` handler in [`IOURequestStepDescription.tsx`](https://github.com/Expensify/App/blob/e570d231d4c0a5f085c53f2aed21d2fc98a39632/src/pages/iou/request/step/IOURequestStepDescription.tsx#L179) called:

```ts
focusComposerWithDelay(inputRef.current)(true);
```

Because `isFocused()` still returns `true` after the modal closes, this hits the "already focused" guard in [`focusComposerWithDelay/index.ts`](https://github.com/Expensify/App/blob/e570d231d4c0a5f085c53f2aed21d2fc98a39632/src/libs/focusComposerWithDelay/index.ts#L37-L43): with `forceKeyboardIfAlreadyFocused` left at its default `false`, the helper early-returns and never re-requests the IME (`input.focus()` is skipped too). So the preserved JS focus state is exactly what stops the keyboard from being restored — the call becomes a no-op and the soft keyboard stays down.

The fix passes the third argument so the native keyboard is re-requested even when iOS still considers the field focused:

```ts
focusComposerWithDelay(inputRef.current)(true, undefined, true);
```

This routes through the existing native re-focus path ([`requestKeyboardForFocusedComposer/index.native.ts`](https://github.com/Expensify/App/blob/e570d231d4c0a5f085c53f2aed21d2fc98a39632/src/libs/focusComposerWithDelay/requestKeyboardForFocusedComposer/index.native.ts#L14-L21) → `KeyboardController.setFocusTo('current')`), which re-shows the IME without a blur/refocus, and leaves web untouched — the web implementation of [`requestKeyboardForFocusedComposer`](https://github.com/Expensify/App/blob/e570d231d4c0a5f085c53f2aed21d2fc98a39632/src/libs/focusComposerWithDelay/requestKeyboardForFocusedComposer/index.ts#L3-L5) is a `NOOP`. So it targets the iOS regression without side effects on other platforms.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97604
PROPOSAL: https://github.com/Expensify/App/issues/97604#issuecomment-5148828487

### Tests

**Verifies the fix (iOS App):**
1. Start creating an expense: choose **Manual**, enter an amount, and continue to the **Description** step.
2. Type any text in the Description input.
3. Press **Back** to trigger the "Discard changes?" modal.
4. Tap **Cancel**.
5. Verify that when the modal closes, focus returns to the Description input **and the soft keyboard is shown again** (previously the keyboard did not return on iOS).

**Regression check (all other platforms — web / Android):**
1. Repeat steps 1–4 above.
2. Verify that after Cancel the Description input still contains the typed text, focus is restored, and there are no console errors or crashes (behavior unchanged from before this PR).

> Note: this is a soft-keyboard timing behavior and is environment-sensitive. It may not reproduce on a local dev / simulator HybridApp build (dev-mode JS timing, simulator keyboard differences, local native binary). It should be ad-hoc tested on a **release-mode build on a real iOS device** (matching how QA caught it) to confirm the keyboard/focus is restored before merge.

### Offline tests

Same as Tests — this change only affects focus/keyboard behavior and is independent of network state.

### QA Steps

Same as Tests. Verify primarily on **iOS: Native** (release build on a real device) that the keyboard/focus returns to the Description input after tapping Cancel on the discard modal, and confirm no regression on the other platforms.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

Automated regression check on dev NewDot web (the fix is a NOOP on web, so this confirms the Description discard-modal cancel flow is not regressed — text is preserved and the input regains focus):

<h3>Description step with input</h3>
Description step with input

<h3>Discard changes modal after pressing Back</h3>
Discard changes modal after pressing Back

<h3>Description step after Cancel, text preserved and input focused</h3>
Description step after Cancel, text preserved and input focused

</details>


src/pages/iou/request/step/IOURequestStepDescription.tsx
ikevin127, line 217: 🟡 **The stated root cause skips the actual mechanism, and the mechanism is fragile**

The description attributes the bug to "the description `TextInput` still reports `isFocused() === true` after the modal closes". That is the guard the call trips on, not the reason the keyboard disappeared.

The driver is on `IOURequestStepDescription.tsx:219`:

```tsx
editable={!isDiscardModalVisible}
```

While the modal is open the input flips to `editable={false}`, so iOS resigns the IME natively while the JS-side `isFocused()` bookkeeping stays `true`. That divergence is exactly why `KeyboardController.setFocusTo('current')` (`reloadInputViews` + `focus` on iOS) is the correct tool and why a plain `focus()` cannot work.

This matters for what to look at during device testing. Check the ordering in `useDiscardChangesConfirmation/index.native.ts:51-54`:

```ts
onVisibilityChange?.(false);
if (result.action !== ModalActions.CONFIRM) {
    blockedNavigationAction.current = undefined;
    onCancel?.();
    return;
}
```

Both run in the same tick, so `editable` has not re-committed to `true` when `onCancel` fires. The fix only lands because `shouldDelay = true` awaits `ComposerFocusManager.isReadyToFocus()`, then `setTimeout(..., COMPOSER_FOCUS_DELAY)`, and then `requestKeyboardForFocusedComposer` defers once more inside `requestIdleCallback`. Three deferrals stacked up are what get us past the React commit.

That works, but it means the failure mode on a slow device is "keyboard sometimes does not return" rather than "never returns". @MelvinBot What do you think about this ?

src/pages/iou/request/step/IOURequestStepDescription.tsx
```diff
@@ -176,7 +176,7 @@ function IOURequestStepDescription({
 
     useDiscardChangesConfirmation({
         onCancel: () => {
-            focusComposerWithDelay(inputRef.current)(true);
+            focusComposerWithDelay(inputRef.current)(true, undefined, true);
```
ikevin127, line 177: 🟡 **`(true, undefined, true)` is unreadable at the call site and has no comment**

```ts
focusComposerWithDelay(inputRef.current)(true, undefined, true);
```

Nothing tells a reader what either boolean means, and the `undefined` is a positional hole for `forcedSelectionRange`. The realistic risk is a future cleanup "simplifying" this back to `(true)` and silently reintroducing the exact bug this PR fixes, because the diff would look like dead-argument removal.

`IOURequestStepDistanceOdometer.tsx:547-550` already sets the precedent for documenting a focus call whose shape is load-bearing. Match it:

```ts
useDiscardChangesConfirmation({
    onCancel: () => {
        // The input is `editable={!isDiscardModalVisible}`, so iOS resigns the IME while the modal is open
        // even though `isFocused()` still reports true. Pass `forceKeyboardIfAlreadyFocused` so the keyboard
        // is re-requested via KeyboardController instead of early-returning on the already-focused guard.
        focusComposerWithDelay(inputRef.current)(true, undefined, true);
    },
    ...
});
```

This also covers the checklist item the author left unchecked: "I verified that comments were added to code that is not self explanatory". @MelvinBot Let me know your take on this, and apply the change if it makes sense.

ikevin127 (approved):
> @ikevin127 QA is done

I'm assuming the issue passed QA / is fixed on adhoc build, so we're good to merge.

mountiny (approved):
thanks for the fix

IuliiaHerets:
QA team failed this PR on Android with an [original issue](https://github.com/Expensify/App/issues/97604) 

https://github.com/user-attachments/assets/54b5319e-24e4-4fee-a45e-17839c3c55c8

cc @mountiny @heyjennahay @ikevin127

ikevin127:
Will verify and open a follow-up PR once confirmed ✅

mountiny:
Thanks

PR #97855, opened 2026-08-05
Refactor Domain menu data into shared source

nabi-ebrahimi:
### Explanation of Change

This PR extracts the existing Domain menu definitions into a shared `getDomainMenuItems` builder and updates `DomainInitialPage` to consume it.

The menu order, labels, icons, routes, access rules, and error indicators remain unchanged. This is a behavior-preserving prerequisite for adding Domain navigation suggestions in a separate PR.

### Fixed Issues

#### Domain menu remains unchanged

Precondition: The user administers a domain.

1. Open **Workspaces**.
2. Open **Domains**.
3. Select a domain.
4. Verify the menu displays **Members**, **Admins**, **Groups**, and **SAML** in that order.
5. Verify every row retains its expected label and icon.
6. Open each row.

Expected: Every row opens its corresponding Domain page, and the menu looks and behaves as it did before this refactor.

### Offline tests
Same as Tests.

### QA Steps
Same as Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/6335780d-ddc0-41d3-bd57-73a8114ff025

</details>

<details>

>

https://github.com/user-attachments/assets/435e37e4-5950-4383-9d56-834410bfcf84

</details>

<details>

>

https://github.com/user-attachments/assets/cbd16033-ec81-4592-980d-c31fbd5e4ae3

</details>

<details>

>

https://github.com/user-attachments/assets/a2e46b49-3634-4a04-8819-1bd902303804

</details>

<details>

>

https://github.com/user-attachments/assets/e6846489-7418-4c46-9ace-431c59765c68

</details>


src/pages/domain/DomainInitialPage.tsx
```diff
@@ -71,35 +54,14 @@ function DomainInitialPage({route}: DomainInitialPageProps) {
...
-            translationKey: 'domain.saml',
-            icon: icons.UserLock,
-            action: singleExecution(waitForNavigate(() => Navigation.navigate(ROUTES.DOMAIN_SAML.getRoute(domainAccountID)))),
-            screenName: SCREENS.DOMAIN.SAML,
-        },
-    ];
+    const domainMenuItems = getDomainMenuItems({
+        domainAccountID,
+        domainErrors,
+        icons: {User: icons.User, UserShield: icons.UserShield, Users: icons.Users, UserLock: icons.UserLock},
```
marufsharifi, line 60: **Redundant icon remapping.** `icons` here comes from `useMemoizedLazyExpensifyIcons(['UserLock', 'UserShield', 'User', 'Users'])`, which already returns an object keyed by exactly those four names — the same key set as `DomainMenuIconMap`. Reconstructing it by hand adds four spots to drift out of sync for zero type-safety gain. Pass it straight through:

```suggestion
        icons,
```

tests/unit/getDomainMenuItemsTest.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    const groupErrors: DomainErrors = {errors: {}};
+    const securityGroupErrorKey = `${CONST.DOMAIN.DOMAIN_SECURITY_GROUP_PREFIX}1` as const;
+    groupErrors[securityGroupErrorKey] = {errors: {groupError: 'Group error'}};
+
+    it.each([
+        ['Members', {errors: {}, memberErrors: {member: {errors: {memberError: 'Member error'}}}}, 'domain.domainMembers'],
+        ['Admins', {errors: {}, adminErrors: {[adminAccountID]: {errors: {adminError: 'Admin error'}}}}, 'domain.domainAdmins'],
+        ['Groups', groupErrors, 'domain.groups.title'],
```
marufsharifi, line 42: **Coverage gap: no case asserts SAML stays clean.** Each `it.each` row checks one error type in isolation, so a regression that wrongly flagged the SAML row with a brick road wouldn't be caught. Since SAML is the only item with no `brickRoadIndicator`, it's worth an explicit guard. Suggested addition to the block:

```ts
it('never shows a brick road on the SAML row, even when other errors exist', () => {
    const domainErrors: DomainErrors = {
        errors: {},
        memberErrors: {member: {errors: {memberError: 'Member error'}}},
    };
    const items = getDomainMenuItems({domainAccountID, domainErrors, icons});
    const saml = items.find((item) => item.translationKey === 'domain.saml');
    expect(saml?.brickRoadIndicator).toBeUndefined();
});
```

marufsharifi (changes requested):

src/pages/domain/DomainInitialPage.tsx
nabi-ebrahimi: Addressed.

tests/unit/getDomainMenuItemsTest.ts
nabi-ebrahimi: Addressed.

nabi-ebrahimi:
@marufsharifi kindly bump, thanks.

marufsharifi:
@nabi-ebrahimi, thanks, i am reviewing.

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
...
+import type {Route} from '@src/ROUTES';
+import SCREENS from '@src/SCREENS';
+import type DomainErrors from '@src/types/onyx/DomainErrors';
+import type IconAsset from '@src/types/utils/IconAsset';
+
+import type {ValueOf} from 'type-fest';
+
+type DomainTopLevelScreens = keyof typeof DOMAIN_TO_RHP;
+
+type DomainMenuItem = {
```
marufsharifi, line 19: no docs for the the defined type

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
...
+    translationKey: TranslationPaths;
+    icon: IconAsset;
+    route: Route;
+    brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>;
+    screenName: DomainTopLevelScreens;
+};
+
+type DomainMenuIconMap = Record<'User' | 'UserShield' | 'Users' | 'UserLock', IconAsset>;
+
+type GetDomainMenuItemsParams = {
```
marufsharifi, line 38: also here

trjExpensify:
How are you getting on, do you think you'll complete the review today?

marufsharifi:
@trjExpensify, yes, but Github is down.

trjExpensify:
Yeah, bad GH. Seems like we're back on track now!

marufsharifi:
I will finish this in an hour. thanks.

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Builds the ordered Domain menu items with their routes and indicator states.
+ */
```
marufsharifi: **[low] JSDoc documents "what", not "why"**

Per [`STYLE.md`](https://github.com/Expensify/App/blob/main/contributingGuides/STYLE.md), JSDoc should *add information only* and document **why**, not **what** — a comment that mostly restates a self-explanatory name adds no value (`getDomainMenuItems` → "Builds the … Domain menu items").

This one is borderline since `"ordered … with their routes and indicator states"` adds a little detail, so treat it as optional.

Suggested: either drop the comment, or make it carry the non-obvious contract — e.g. that order is significant, or that the caller is expected to attach `action`/navigation afterward.

```ts
/**
 * Menu order is significant — it drives the on-screen layout.
 * Callers attach `action`/navigation onto each returned item.
 */
```

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
...
+
+type DomainMenuIconMap = Record<'User' | 'UserShield' | 'Users' | 'UserLock', IconAsset>;
+
+type GetDomainMenuItemsParams = {
+    domainAccountID: number;
+    domainErrors?: DomainErrors;
+    icons: DomainMenuIconMap;
+};
+
+function getDomainMenuItems({domainAccountID, domainErrors, icons}: GetDomainMenuItemsParams): DomainMenuItem[] {
```
marufsharifi, line 52: **This reinvents an existing pattern with a different shape.**

We already extract an InitialPage's menu-item list into its own module: `src/pages/settings/useInitialSettingsPageMenuData.ts` (L89). That precedent does it as a **hook** that owns its own Onyx + icons (`useMemoizedLazyExpensifyIcons`) and builds each item's `action` internally, exporting `MenuData`/`MenuSection`.

This function does the same job but with a divergent convention — a pure function that takes an `icons` map + `domainErrors` as params and returns `route` instead of a built `action`, leaving `DomainInitialPage` to wrap it. Two conventions for the same concept is what our "a new code pattern must be agreed by multiple engineers" rule is meant to catch.

Could we either:
- follow the settings precedent (`useDomainMenuData` hook that owns its Onyx/icons and returns built items), **or**
- keep the pure-function form (it genuinely reads cleaner in the new unit test) but call it out in the PR description and get a second engineer to sign off?

No change strictly required — just want the divergence to be a deliberate, agreed choice rather than a silent third pattern.

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
...
+
+type DomainTopLevelScreens = keyof typeof DOMAIN_TO_RHP;
+
+type DomainMenuItem = {
+    translationKey: TranslationPaths;
+    icon: IconAsset;
+    route: Route;
+    brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>;
+    screenName: DomainTopLevelScreens;
+};
```
marufsharifi, line 34: **`DomainMenuItem` is now the third copy of this type shape.**

The same core shape already exists in two places:
- `src/pages/workspace/WorkspaceInitialPage.tsx` (L85) — `WorkspaceMenuItem`
- `src/pages/settings/useInitialSettingsPageMenuData.ts` (L61) — `MenuData`

All three model "a top-level nav menu item keyed to a `*_TO_RHP` relation" with the identical core: `translationKey: TranslationPaths`, `icon: IconAsset`, `brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>`, and `screenName` typed as `keyof typeof <X>_TO_RHP`.

Consider extracting a shared generic base so these can't drift, e.g.:

```ts
type TopLevelMenuItem<TScreens> = {
    translationKey: TranslationPaths;
    icon: IconAsset;
    brickRoadIndicator?: ValueOf<typeof CONST.BRICK_ROAD_INDICATOR_STATUS>;
    screenName: TScreens;
};
```

Since that touches two files outside this PR's scope, the DRY-correct move is to land the base type as its own prep change first, then have all three specialize it. Fine to defer — flagging so we don't grow a fourth copy later.

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
...
+
+type DomainMenuIconMap = Record<'User' | 'UserShield' | 'Users' | 'UserLock', IconAsset>;
```
marufsharifi: **Bespoke icon-map param duplicates the icon-name list.**

These four icon names are already declared in `src/pages/domain/DomainInitialPage.tsx` (L59):

```ts
const icons = useMemoizedLazyExpensifyIcons(['UserLock', 'UserShield', 'User', 'Users']);
```

`DomainMenuIconMap` restates the same set here, so adding/removing a menu item now means editing the list in two spots. The established analog (`useInitialSettingsPageMenuData`) avoids this by resolving icons inside the module rather than passing them in.

Mostly resolves itself if we adopt the hook form from the other comment (icons resolved internally). If we keep the pure function, consider deriving the map keys from a single source so the two lists can't diverge.

src/pages/domain/DomainInitialPage.tsx
```diff
@@ -71,35 +54,14 @@ function DomainInitialPage({route}: DomainInitialPageProps) {
...
-        },
-    ];
+    const domainMenuItems = getDomainMenuItems({
+        domainAccountID,
+        domainErrors,
+        icons,
+    }).map((item) => ({
+        ...item,
+        action: singleExecution(waitForNavigate(() => Navigation.navigate(item.route))),
+    }));
```
marufsharifi, line 64: **Per-render double allocation via `.map` spread**

`getDomainMenuItems(...)` already builds 4 fresh objects, then this `.map((item) => ({...item, action}))` spread-copies each into 4 *more* objects on every render (~8 allocations/render vs. ~4 before, plus 4 new closures). It's negligible here — the array is a fixed 4 items on a screen that renders infrequently — so no change is required.

If you want to fold the extra copy away, pass `singleExecution`/`waitForNavigate` into `getDomainMenuItems` and build `action` there, so a single object is produced per item:

```ts
const domainMenuItems = getDomainMenuItems({domainAccountID, domainErrors, icons, singleExecution, waitForNavigate});
```

src/pages/domain/getDomainMenuItems.ts
```diff
@@ -0,0 +1,68 @@
...
+    domainErrors?: DomainErrors;
+    icons: DomainMenuIconMap;
+};
+
+function getDomainMenuItems({domainAccountID, domainErrors, icons}: GetDomainMenuItemsParams): DomainMenuItem[] {
+    return [
+        {
+            translationKey: 'domain.domainMembers',
+            icon: icons.User,
+            route: ROUTES.DOMAIN_MEMBERS.getRoute(domainAccountID),
```
marufsharifi, line 57: **NAB: ** Route strings now computed eagerly every render instead of lazily on press

Previously each `ROUTES.*.getRoute(domainAccountID)` ran inside the `action` closure, i.e. only when the item was tapped. Now all four `getRoute` calls run eagerly whenever `getDomainMenuItems` is called (every render of `DomainInitialPage`). Each `getRoute` is a single template-string interpolation ([ROUTES.ts:4759](../blob/main/src/ROUTES.ts#L4759)), so the cost is ~4 trivial string ops/render — effectively free. Flagging only for completeness; no action needed.

tests/unit/getDomainMenuItemsTest.ts
```diff
@@ -0,0 +1,55 @@
...
+            ROUTES.DOMAIN_SAML.getRoute(domainAccountID),
+        ]);
+        expect(items.map((item) => item.icon)).toEqual([icons.User, icons.UserShield, icons.Users, icons.UserLock]);
+        expect(items.map((item) => item.screenName)).toEqual([SCREENS.DOMAIN.MEMBERS, SCREENS.DOMAIN.ADMINS, SCREENS.DOMAIN.GROUPS, SCREENS.DOMAIN.SAML]);
+        expect(items.every((item) => item.brickRoadIndicator === undefined)).toBe(true);
+    });
+
+    const groupErrors: DomainErrors = {errors: {}};
+    const securityGroupErrorKey = `${CONST.DOMAIN.DOMAIN_SECURITY_GROUP_PREFIX}1` as const;
+    groupErrors[securityGroupErrorKey] = {errors: {groupError: 'Group error'}};
```
marufsharifi: **NAB:** `groupErrors` is declared as a `const` literal and then mutated with a computed key on the next line. The two-step build reads as leftover procedural setup and mutates a `const`. Hoisting the key and using a computed property makes it a single, clear fixture:

```suggestion
    const securityGroupErrorKey = `${CONST.DOMAIN.DOMAIN_SECURITY_GROUP_PREFIX}1` as const;
    const groupErrors: DomainErrors = {
        errors: {},
        [securityGroupErrorKey]: {errors: {groupError: 'Group error'}},
    };
```

src/pages/domain/DomainInitialPage.tsx
marufsharifi, line 64: This `.map` — turning each item's `route` into `action: singleExecution(waitForNavigate(() => Navigation.navigate(item.route)))` — is the only changed logic with no test.

`getDomainMenuItemsTest.ts` proves the `route` **values** are correct, but nothing asserts the component actually wires `item.route` → `Navigation.navigate`.

It's non-branching glue and there was no `DomainInitialPage` component test before this PR, so this is **not** a coverage regression. To close it to a literal 100%, add a component test (new `tests/unit/DomainInitialPageTest.tsx`): mock `@libs/Navigation/Navigation`, render the page, press each `HighlightableMenuItem`, and assert `Navigation.navigate` was called with `ROUTES.DOMAIN_MEMBERS.getRoute(domainAccountID)` (and the other three routes). This requires mocking `useOnyx`, `useNavigationState`, `useCurrentUserPersonalDetails`, and `useMemoizedLazyExpensifyIcons` — heavier than the payoff for pure glue, so treat as optional.

marufsharifi (commented):
Tested, it works perfect, but some minor suggestions.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: The pure builder is intentional because Search Router may build items for multiple domains, where hooks cannot be called in a loop. Returning routes also lets each consumer attach its own navigation lifecycle, so I propose keeping this shape and would appreciate an engineer confirming the pattern.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: Thanks, this is worth considering separately. Since the three types have different requirements and sharing them would touch unrelated Settings and Workspace code, I’ll defer it to a focused follow-up and keep this PR scoped to the Domain extraction.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: Addressed.

src/pages/domain/DomainInitialPage.tsx
nabi-ebrahimi: Thanks for flagging this. Since the allocation is negligible and moving page-specific navigation handling into the shared builder would reduce its reusability, I’ll keep the current implementation.

src/pages/domain/getDomainMenuItems.ts
nabi-ebrahimi: Thanks for noting this. The eager route creation is negligible and supports the shared builder’s reusable route-based shape, so no change is needed.

tests/unit/getDomainMenuItemsTest.ts
nabi-ebrahimi: Addressed.

src/pages/domain/DomainInitialPage.tsx
nabi-ebrahimi: Thanks for noting this. Since the mapping is non-branching glue, all route values are already unit-tested, and a component test would require disproportionate mocking, I’ll keep the existing focused tests and manual navigation coverage.

nabi-ebrahimi:
@marufsharifi thanks for the detailed review. All actionable comments have been addressed, and I replied to the optional items with the reasoning for keeping the current approach. Could you please review again?

trjExpensify:
... and another! @marufsharifi, please re-review it today.

marufsharifi:
Thanks, i will finish this in afternoon.

tests/unit/getDomainMenuItemsTest.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    it.each([
+        ['Members', {errors: {}, memberErrors: {member: {errors: {memberError: 'Member error'}}}}, 'domain.domainMembers'],
+        ['Admins', {errors: {}, adminErrors: {[adminAccountID]: {errors: {adminError: 'Admin error'}}}}, 'domain.domainAdmins'],
+        ['Groups', groupErrors, 'domain.groups.title'],
+    ] satisfies Array<[string, DomainErrors, string]>)('preserves the %s error indicator', (_name, domainErrors, expectedTranslationKey) => {
+        const items = getDomainMenuItems({domainAccountID, domainErrors, icons});
+
+        expect(items.filter((item) => item.brickRoadIndicator === CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR).map((item) => item.translationKey)).toEqual([expectedTranslationKey]);
+    });
```
marufsharifi, line 49: Each case here isolates a single erroring section. Consider adding one case where multiple sections error at once, to lock in that the indicators are independent (a real backend payload can carry member + admin + group errors simultaneously):

```ts
it('shows indicators on every erroring section at once', () => {
    const domainErrors: DomainErrors = {
        errors: {},
        memberErrors: {member: {errors: {e: 'x'}}},
        adminErrors: {[adminAccountID]: {errors: {e: 'x'}}},
    };
    const flagged = getDomainMenuItems({domainAccountID, domainErrors, icons})
        .filter((item) => item.brickRoadIndicator === CONST.BRICK_ROAD_INDICATOR_STATUS.ERROR)
        .map((item) => item.translationKey);

    expect(flagged).toEqual(['domain.domainMembers', 'domain.domainAdmins']);
});
```

Why it matters: a future change to the error-selection logic could accidentally make indicators mutually exclusive, and the current one-at-a-time cases wouldn't catch it.

marufsharifi (commented):
Looks good, but a minor suggestion.

tests/unit/getDomainMenuItemsTest.ts
nabi-ebrahimi: Addressed.

nabi-ebrahimi:
@marufsharifi thanks for the review. All actionable comments have been addressed. Please take another look.

marufsharifi (approved):
LGTM

nabi-ebrahimi:
@carlosmiceli can you please also merge this one too if it looks good to you, cause domain navigation is also blocked by this one. Thanks.

carlosmiceli (approved):

nabi-ebrahimi:
@trjExpensify I think we are waiting for your review as Product here, Thanks!

carlosmiceli:
No need, I just had to leave yesterday before CI checks completed.

PR #92498, opened 2026-06-03
Introduce the spinner on the `FormAlertWithSubmitButton` buttons

OlGierd03:
### Explanation of Change

A spinner has been added to the buttons in `FormAlertWithSubmitButton` to improve responsiveness for users and thus reduce the INP metric for interactions labelled `[data-sentry-label="Form-SubmitButton"]`.

A loading spinner now appears on vast majority of the `FormAlertWithSubmitButton` buttons the instant the button is pressed, before `onSubmit` or any loading state driven by Onyx. These changes add a shared `usePressLoading` hook that holds the logic for showing a spinner immediately on press. The instant spinner is on by default through a new `shouldShowLoadingImmediatelyOnPress` prop that defaults to `true`. Gatekeepers that validate before submitting, such as `FormProvider` with `FormWrapper` and `BaseLoginForm`, opt out and drive the spinner themselves through the same hook. Some consumers who make early returns via `onSubmit` with state-only change have also been excluded.

Under the hood a press flips a local `isPressed` flag to true and the actual work is wrapped in `setTimeout(…,0)`. Deferring by one macrotask lets React commit `isPressed=true` and paint the spinner before consumer code that may block the JS thread runs. `isPressed` then resets on the next tick and hands the loading state over to the consumer's real `isLoading`, which is usually driven by Onyx. If validation or violations short circuit the submit and `isLoading` never turns true the spinner cancels itself naturally.

The navigate forward case needs special handling. When `onSubmit` pushes a new screen on top of the current one the current screen stays mounted so a local loading flag would never reset and the spinner would stay stuck on return. Those screens (such as `DynamicWorkspaceInvitePage`) clear the flag when the screen regains focus through `useEffectFocus`.

### Fixed Issues

$ https://github.com/Expensify/App/issues/85596

### Tests

<details>
<summary>FormProvider</summary>
<br />

**Account > Profile > Display name**
  1. Go to Account (bottom tab) > Profile > Display name.
  2. Change the first name and press Save.
  3. Verify the spinner appears on Save the instant it is pressed (before the value persists) and the form saves correctly.
  4. Reopen the form, clear a required field and press Save.
  5. Verify the validation error shows and the spinner does NOT stay stuck (it cancels on violations).
  6. Rapidly double-press Save on a valid form.
  7. Verify the form is submitted only once (no duplicate request).

**Account > Profile > Contact method (or Workspace > Profile > Name)**
  1. Open the form, enter a valid value and press the submit button.
[...]

PR #97956, opened 2026-08-06
Introduce immediate spinner to ButtonCompose

OlGierd03:
### Explanation of Change

Adds an opt-in immediate loading state to `ButtonComposed`, so a press is acknowledged within a frame instead of after the handler's blocking work. The new `shouldShowLoadingImmediatelyOnPress` prop wraps the whole `onPress`: Button commits a local pressed state, lets React paint the spinner, then runs the handler one macrotask later and hands over to the Onyx-driven `isLoading` (where one exists) with no flicker. Turning it on is a one-line change at the call site.

The engine is the existing `usePressLoading`, generalized for this: `isLoading` is now optional (with none passed the pressed state clears once the work settles, instead of waiting for a hand-over that never comes), `startWithLoading` awaits async handlers and the focus reset subscribes through `NavigationContext` directly so a Button rendered outside a navigator is fine.

For handlers that validate or branch before doing the work, the prop is the wrong tool (it would spin on a press that does nothing), so those consumers keep calling `usePressLoading` themselves and pass its `isLoading` in.

`ButtonContext` is split into a state context and an actions context (rulesdir/context-provider-split-values), since `onPress` published to the primitives is now the wrapped handler.

The same mechanism already shipped for form submits in #92498, and the chart below shows why it's worth spreading: daily INP spans for [data-sentry-label="Form-SubmitButton"] fall from 1K to 1.7K to near zero at the 16 Jul deploy, with the badness score flatlining on the same date (the single badness spike on 12 Aug is an unrelated anomaly):

improvement

### Fixed Issues

https://github.com/Expensify/App/issues/96058 (see comments: [#1](https://github.com/Expensify/App/issues/96058#issuecomment-4970022507), [#2](https://github.com/Expensify/App/issues/96058#issuecomment-4970119567))
https://github.com/Expensify/App/issues/96125

### Tests

1. Open an expense report.
2. Enter the selection mode and select several expenses (the more, the better because the JS thread block on save is easier to observe).
3. Press `Edit multiple` in the selection menu.
4. In the `Edit multiple expenses` RHP, open any field (e.g. `Description`), set a new value and press `Save` on the field page.
5. Back on the `Edit multiple expenses` page, press `Save`.
6. Verify the spinner appears on `Save` the instant it is pressed, before the RHP freezes/dismisses, and that it stays visible until the RHP is dismissed (no frame without the spinner).
7. Verify the edited value is applied to all selected expenses.

### PR Author Checklist

    - [ ] Android: Native
    - [ ] Android: mWeb Chrome
    - [ ] iOS: Native
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari

### Screenshots/Videos

Before:

https://github.com/user-attachments/assets/3a2c6048-507d-46bd-b1c4-a81ca738b31f

After:

https://github.com/user-attachments/assets/7945740e-31bf-4d0c-9e10-83759e6590aa


src/components/ButtonComposed/Button.tsx
```diff
@@ -12,21 +12,23 @@ import type {SkeletonSpanReasonAttributes} from '@libs/telemetry/useSkeletonSpan
...
-import {ButtonContext} from './context';
+import {ButtonActionsContext, ButtonStateContext} from './context';
 
 function Button({
     children,
     contentContainerStyle = [],
     size = CONST.BUTTON_SIZE.MEDIUM,
-    isLoading = false,
+    isLoading: isOnyxLoading = false,
+    shouldShowLoadingImmediatelyOnPress = false,
```
mikolajpochec, line 30: `showLoadingImmediatelyOnPress` matches `ButtonComposed`'s migration style better. I'd shorten it even more to `showInstantLoadingOnPress`.

src/components/ButtonComposed/types.ts
```diff
@@ -38,6 +54,16 @@ type ButtonBehaviorProps = {
...
 
+    /**
+     * Opt-in: show the loading spinner the moment the button is pressed, ahead of `onPress`. Defaults to false.
+     *
+     * The button owns the pressed state and paints the spinner before running `onPress`, so a JS-blocking handler still
+     * gives instant feedback. When an external `isLoading` turns true the button hands the spinner over to it; otherwise
+     * the pressed state resets when the screen regains focus. Leave it off for handlers that only navigate, toggle local
+     * state, or bail out on validation — those should use the `loading` controller passed to `onPress` instead.
+     */
+    shouldShowLoadingImmediatelyOnPress?: boolean;
```
mikolajpochec, line 50: Same as in `Button.tsx`. Migration generally did follow pattern of not using `should*` props.

src/components/ButtonComposed/Button.tsx
```diff
@@ -62,17 +64,77 @@ function Button({
...
+        await new Promise((resolve) => {
+            setTimeout(resolve, 0);
+        });
+        try {
+            await runAfterPaint();
+        } catch (error) {
+            setIsPressed(false);
+            throw error;
+        }
+    };
```
mikolajpochec: I would rather replace this logic with an existing `usePressLoading` hook.

src/components/ButtonComposed/Button.tsx
```diff
@@ -62,17 +64,77 @@ function Button({
...
+    // press, minus the mouse-only blur/haptic handling — so the immediate spinner works on Enter too. Living in the actions
+    // context (functions only) keeps it clear of rulesdir/context-provider-split-values.
+    const handleEnterPress = () => {
+        if (isDisabled || isLoading) {
+            return;
+        }
+        if (shouldShowLoadingImmediatelyOnPress) {
+            return startWithLoading(() => onPress(undefined, loadingController));
+        }
+        return onPress(undefined, loadingController);
```
mikolajpochec: Repeated logic from the same file, lines 116-123

src/components/ButtonComposed/types.ts
```diff
@@ -8,9 +8,25 @@ import type React from 'react';
...
 type ButtonEventsProps = {
-    /** A function that is called when the button is clicked on */
-    onPress?: (event?: GestureResponderEvent | KeyboardEvent) => void | Promise<void>;
+    /**
+     * A function that is called when the button is clicked on.
+     *
+     * Receives a `loading` controller as its second argument. Call `loading.run(work)` to show the spinner immediately and
+     * defer the heavy work until after paint — useful when the handler validates first and only some branches should load.
+     */
+    onPress?: (event?: GestureResponderEvent | KeyboardEvent, loading?: PressLoadingController) => void | Promise<void>;
```
mikolajpochec: AI:

> I'd drop this second argument. `onPress` is not a new API — handlers all over the app already match it, so adding a second positional arg means Button now passes something into a slot some handlers already use for their own value. That already happened in `MoneyRequestAttendeeSelector.tsx:132`, where `option?: OptionData` is only checked for truthiness: the controller lands there, the check flips, the early return never runs, and `onFinish` fires with nothing selected.
> 
> TypeScript only caught it because `{run}` and `OptionData` have nothing in common. Any handler with a second param typed `unknown`, `any`, `object`, or anything wider than `{run}` takes the controller with no error at all. So we cannot rely on the compiler here.
> 
> The fix used, `onPress={() => confirmSelection()}`, works by throwing away both args. Fine here since the event was unused, but it is the pattern people will copy, and in other places it silently loses the event. It also looks identical to a normal arrow wrapper, so a reviewer cannot tell it is load-bearing.
> 
> "Validate first, then show the spinner" also already has an answer on the consumer side — `DateFilterBase`, `TermsStep` and `ReportAddApproverPage` call `usePressLoading` themselves.
> 
> Suggestion: keep only `shouldShowLoadingImmediatelyOnPress`, and let handlers that need per-branch control use the hook. If we keep the controller, move it out of the positional slot — its own prop or a single options object — so no existing handler can pick it up by accident.

I generally agree with that. After introducing already existing `usePressLoading` hook this argument becomes redundant.

mikolajpochec (commented):
Great work, but I'd like to address a few nitpicks and a possible code cleanup.

src/components/ButtonComposed/Button.tsx
OlGierd03: That's true, but since this is a POC, for now I'll keep `showLoadingImmediatelyOnPress`, which aligns with the existing naming convention in the app that was introduced along with `usePressLoading`. When migrating, we should consider renaming this variable to something more intuitive, but for now I wouldn't want to complicate matters with additional changes

src/components/ButtonComposed/types.ts
OlGierd03: We'll make sure to consider this when the migration takes place

src/components/ButtonComposed/Button.tsx
OlGierd03: Thanks for the catch and the proposed improvement 🫶

src/components/ButtonComposed/types.ts
OlGierd03: Thanks for the catch 👍

src/components/ButtonComposed/Button.tsx
OlGierd03: Yeah, it makes sense. Initially I wanted to encapsulate all the logic related to the spinner inside the files associated with the button, but that's probably not the best solution for keeping the code clean

PR #97972, opened 2026-08-06
Add Workspace navigation suggestions to Search Router

nabi-ebrahimi:
### Explanation of Change

Adds accessible Workspace pages to Search Router using the shared Workspace menu source.

Workspace suggestions:
- Respect existing workspace access, feature, and visibility rules.
- Display the workspace avatar and name.
- Match a workspace name to its Overview page only.
- Support the exact `HR` query.
- Keep results alphabetically ordered.
- Synchronize the Workspace LHN, title, and content when navigating between workspaces.

### Fixed Issues

PROPOSAL: https://github.com/Expensify/App/issues/92752#issuecomment-4967287007

### Tests

1. **Search for a Workspace page**
   1. Open Search Router.
   2. Search for `members`.
   3. Select a Workspace Members result.
   4. Verify the correct workspace Members page opens and the workspace avatar/name is shown in the result.

2. **Search by workspace name**
   1. Open Search Router.
   2. Search for a workspace name.
   3. Verify only that workspace's Overview result matches its name.
   4. Verify unrelated workspace subpages are not shown.

3. **Search for HR**
   1. Open Search Router.
   2. Search for `HR`.
   3. Verify accessible Workspace HR results appear.

4. **Switch between workspaces**
   1. On a wide layout, open a subpage in Workspace A.
   2. Open Search Router and select a page in Workspace B.
   3. Verify the content, workspace title, and Workspace LHN all update to Workspace B.
   4. Repeat using the same subpage in both workspaces.

5. **Verify visibility rules**
   1. Disable a Workspace feature such as Workflows.
   2. Search for that page.
   3. Verify the disabled page is not suggested.
   4. Verify inaccessible or deleted workspaces are not suggested.

### Offline tests
Same as the Tests.

### QA Steps
Same as the Tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

>

https://github.com/user-attachments/assets/777e13ba-9e87-4c74-a07e-d4601861d05e

</details>

<details>

>

https://github.com/user-attachments/assets/d132f053-9451-4c64-9fcd-cf85e2bccf65

</details>

<details>

>

https://github.com/user-attachments/assets/edf0c2cb-73e5-44ea-bb29-6619eb45b0cd

</details>

<details>

>

https://github.com/user-attachments/assets/6833e5bc-264e-47bf-9948-c5eabf557ca2

</details>

<details>

>

1. **Search for a Workspace page**
>

https://github.com/user-attachments/assets/4fab5a37-d217-44da-acdd-ddd0267694c1

2. **Search by workspace name**
>

https://github.com/user-attachments/assets/b374a97d-a081-4621-9b9f-f0760f65040c

3. **Search for HR**
>

https://github.com/user-attachments/assets/75819b45-30ae-4774-9e8b-da8a58941de7

4. **Switch between workspaces**
>

https://github.com/user-attachments/assets/a5553dee-d48e-4c7a-8d41-d34665ea8478

5. **Verify visibility rules**
>

https://github.com/user-attachments/assets/4fa696f8-e5bd-41f2-87e8-956a369431a4

</details>


src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -64,6 +101,46 @@ type BuildSpendNavigationItemsParams = {
...
+    isRulesRevampBetaEnabled: boolean;
+    isVendorMatchingBetaEnabled: boolean;
+    shouldUseNarrowLayout: boolean;
+    convertToDisplayString: Parameters<typeof getWorkspaceMenuItems>[0]['convertToDisplayString'];
+    getItemText: (item: ReturnType<typeof getWorkspaceMenuItems>[number]) => string;
+    getDestinationText: (destination: string) => string;
+};
+
+type WorkspaceContextProps = {
+    policy: OnyxTypes.Policy;
```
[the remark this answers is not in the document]
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -123,11 +200,59 @@ function buildSpendNavigationItems({sections, icons, rightElement, getItemText,
...
+    isOffline,
+    isRulesRevampBetaEnabled,
+    isVendorMatchingBetaEnabled,
+    shouldUseNarrowLayout,
+    convertToDisplayString,
+    getItemText,
+    getDestinationText,
+}: BuildWorkspaceNavigationItemsParams): NavigationSuggestionSourceItem[] {
+    return Object.values(policies ?? {})
+        .filter((policy): policy is OnyxTypes.Policy => !!policy?.id && shouldShowPolicy(policy, isOffline, currentUserLogin))
```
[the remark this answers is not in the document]
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function navigateToWorkspaceSettingsRoute(targetRoute: Route, policyID: string, shouldUseNarrowLayout: boolean) {
+    if (shouldUseNarrowLayout) {
+        Navigation.navigate(targetRoute);
+        return;
+    }
+
+    const activeRoute = Navigation.getActiveRouteWithoutParams();
+    const targetSuffix = getWorkspaceRouteSuffix(targetRoute);
+    const isWorkspaceSettingsTransition = targetSuffix !== undefined && activeRoute !== targetRoute && WORKSPACE_ROUTE_PATTERN.test(activeRoute);
```
marufsharifi: **`getWorkspaceRouteSuffix` and the `targetSuffix !== undefined` guard are dead code.**

`getWorkspaceRouteSuffix` (lines 25–29) ends with `return match?.[1] ?? '';`, so it **always returns a string — never `undefined`**. That makes `targetSuffix !== undefined` on line 78 **always `true`**, and `targetSuffix` is used nowhere else. So the whole helper + local var have no effect, and the transition decision actually rests only on `activeRoute !== targetRoute && WORKSPACE_ROUTE_PATTERN.test(activeRoute)`.

Why it matters: the `string | undefined` return type and the guard read as if they validate "the target is a workspace route," but they don't — and they only ever check the **active** route, never the **target**. If this exported helper is later reused with a non-workspace `targetRoute`, nothing stops it from `forceReplace`-ing + `setParams`-ing a policyID onto a mismatched target. Today it's masked only because callers always pass a workspace route.

Suggested fix — make the intent real, or delete the helper entirely:
```ts
// gate on the TARGET actually being a workspace settings route
const isWorkspaceSettingsTransition =
    WORKSPACE_ROUTE_PATTERN.test(targetRoute) && activeRoute !== targetRoute && WORKSPACE_ROUTE_PATTERN.test(activeRoute);
// then remove getWorkspaceRouteSuffix (25–29) and targetSuffix (77)
```

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,93 @@
...
+    name?: string;
+    key?: string;
+    params?: {
+        policyID?: string;
+    };
+    state?: {
+        key?: string;
+        routes?: NavigationRouteWithState[];
+    };
+};
```
marufsharifi: **Hand-rolled `NavigationRouteWithState` duplicates React Navigation's state typing.**

This type re-describes the navigator state shape (`name`/`key`/`params`/`state.routes`) that React Navigation already models. It works structurally, but reusing the existing navigation-state types (or a `PartialState`-derived type) keeps `findWorkspaceSidebarRoute`/`getActiveWorkspaceSidebarRoute` aligned with the real navigator types, so a future change to the navigation state shape is caught by the compiler here instead of silently traversing wrong. Non-blocking, but preferable for consistency with the rest of the navigation layer.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,93 @@
...
+    return undefined;
+}
+
+function navigateToWorkspaceSettingsRoute(targetRoute: Route, policyID: string, shouldUseNarrowLayout: boolean) {
+    if (shouldUseNarrowLayout) {
+        Navigation.navigate(targetRoute);
+        return;
+    }
+
+    const activeRoute = Navigation.getActiveRouteWithoutParams();
```
marufsharifi: **`activeRoute` is param-stripped but compared against the raw `targetRoute`.**

`Navigation.getActiveRouteWithoutParams()` (line 76) strips the query string, but the `activeRoute !== targetRoute` check on line 78 uses `targetRoute` as-is. For today's workspace routes (no query string) this is fine, but some workspace routes do append a query — e.g. `ROUTES.WORKSPACE_WORKFLOWS.getRoute(policyID, tab)` puts the tab in a query param. If such a route is ever passed here, the equality would compare a stripped string against one with `?tab=...` and misfire (treating same-page as a different-page transition). Worth a one-line comment documenting the "targets have no query param" assumption, or normalize `targetRoute` the same way before comparing.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -124,11 +202,59 @@ function buildSpendNavigationItems({sections, icons, rightElement, getItemText,
...
+                return {
+                    text: getDestinationText(itemText),
+                    singleIcon: item.icon,
+                    action: () => navigateToWorkspaceSettingsRoute(item.getRoute(), policy.id, shouldUseNarrowLayout),
+                    keyForList: `workspace_${policy.id}_${item.screenName}`,
+                    rightElement: <WorkspaceContext policy={policy} />,
+                    matchTerms: item.screenName === SCREENS.WORKSPACE.PROFILE ? [itemText, policy.name] : [itemText],
+                };
+            });
+        });
```
marufsharifi, line 310: **Build cost scales with workspace count — confirm this is memoized.**

`buildWorkspaceNavigationItems` maps **every accessible policy** through the full `getWorkspaceMenuItems` (per-policy sync-error / HR / vendor checks) and creates a `<WorkspaceContext policy={policy} />` React element for **every** row (line 239) before any query filtering. `buildNavigationSuggestions` then sorts the *entire unfiltered* source on each run.

Mitigation already in place: the caller feeds a debounced value — `useNavigationSuggestions(textInputValue, …)` where `textInputValue = useDebouncedState(initialQuery, 500)` — so this runs ~every 500ms, not per keystroke. 👍

Remaining ask: `buildWorkspaceNavigationItems`' inputs don't include the query, so the result should be memoized so it doesn't re-materialize hundreds of React elements on each debounced query change (only when policies/betas/layout actually change). Please confirm React Compiler memoizes this file:
```bash
npm run react-compiler-compliance-check check-changed
```
If it doesn't, a High-Traffic account (many workspaces) rebuilds the whole workspace menu + elements repeatedly while typing — the exact input-lag the High-Traffic checklist item guards against.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -65,6 +102,47 @@ type BuildSpendNavigationItemsParams = {
...
+    getItemText: (item: ReturnType<typeof getWorkspaceMenuItems>[number]) => string;
+    getDestinationText: (destination: string) => string;
+};
+
+type WorkspaceContextProps = {
+    /** The policy used to identify the workspace suggestion */
+    policy: OnyxTypes.Policy;
+};
+
+function WorkspaceContext({policy}: WorkspaceContextProps) {
```
marufsharifi: **`WorkspaceContext` is a misleading name for a presentational cell.**

This renders a right-hand avatar + name cell, but the name reads like a React Context. In a codebase full of real `*Context` / `*Provider` symbols, this will confuse anyone grepping for context providers. Suggest renaming to something presentational, e.g. `WorkspaceSuggestionCell` or `WorkspaceIdentityCell` (update the `WorkspaceContextProps` type and the usage on line 239 too).

marufsharifi:
@MelvinBot Review

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Thanks for the suggestion. Since this is non-blocking and the current type is intentionally limited to the fields this helper reads, I’ll keep it unchanged to avoid introducing additional narrowing or casts in this scoped PR. The navigation behavior is covered by focused tests.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Confirmed that `npm run react-compiler-compliance-check check-changed` passes for all changed React files, so the Workspace suggestions are compiler-managed and no manual memoization is needed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Addressed.

nabi-ebrahimi:
@marufsharifi, all review comments have been addressed or clarified, and the focused checks pass. Could you please take another look? Thanks!

trjExpensify:
Nice. Same here, @marufsharifi. Let's get these re-reviews done tomorrow!

marufsharifi:
This is almost done, i will finish this before afternoon.

marufsharifi:
## Reviewer Checklist

    - [ ] Android: HybridApp
    - [ ] Android: mWeb Chrome
    - [ ] iOS: HybridApp
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

marufsharifi:
@nabi-ebrahimi, could you please merge the main. thanks.

nabi-ebrahimi:
@marufsharifi Done, thanks

nabi-ebrahimi:
@marufsharifi can you please do another review here, Thanks!

marufsharifi:
@nabi-ebrahimi, could you please check this. 

[RF-1.webm](https://github.com/user-attachments/assets/766f46bf-f717-4d45-8b15-86b6da810dc2)

nabi-ebrahimi:
@marufsharifi thanks for the review, fixed this:

https://github.com/user-attachments/assets/7cb18549-2fed-456d-a511-566df80b1622

Can you please do another review.

marufsharifi:
@nabi-ebrahimi, could you please check this. thanks. 



https://github.com/user-attachments/assets/f9062597-00dc-43c1-b52c-8f3776766522

nabi-ebrahimi:
> @nabi-ebrahimi, could you please check this. thanks.

Resolved. Thanks for catching this, can you please review again?

trjExpensify:
@marufsharifi can you give this a re-review? Thanks!

marufsharifi:
@nabi-ebrahimi, could you please merge the main. thanks.

nabi-ebrahimi:
> @nabi-ebrahimi, could you please merge the main. thanks.

Done, thanks.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+    name?: string;
+    key?: string;
+    params?: {
+        policyID?: string;
+    };
+    state?: {
+        key?: string;
+        routes?: NavigationRouteWithState[];
+    };
+};
```
marufsharifi: No docs added to this type.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -70,6 +105,47 @@ type BuildSpendNavigationItemsParams = {
...
+    currentUserLogin: string | undefined;
+    icons: Parameters<typeof getWorkspaceMenuItems>[0]['icons'];
+    isOffline: boolean;
+    isRulesRevampBetaEnabled: boolean;
+    isVendorMatchingBetaEnabled: boolean;
+    shouldUseNarrowLayout: boolean;
+    convertToDisplayString: Parameters<typeof getWorkspaceMenuItems>[0]['convertToDisplayString'];
+    getItemText: (item: ReturnType<typeof getWorkspaceMenuItems>[number]) => string;
+    getDestinationText: (destination: string) => string;
+};
```
marufsharifi, line 150: please add docs.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function navigateToWorkspaceSettingsRoute(targetRoute: Route, policyID: string, shouldUseNarrowLayout: boolean) {
+    if (shouldUseNarrowLayout) {
+        Navigation.navigate(targetRoute);
+        return;
+    }
+
+    const activeRoute = Navigation.getActiveRouteWithoutParams();
+    const targetRouteWithoutParams = targetRoute.replaceAll(/\?.*/g, '');
```
marufsharifi: **Route comparison mixes leading-slash and slash-less formats — this guard never short-circuits in production.**

`Navigation.getActiveRouteWithoutParams()` returns a path **with** a leading slash (it's `getPathFromState()`, which always prefixes `/` — see the convention at `VideoPlayerPreview/index.tsx` comparing against `` `/${ROUTES.REPORT_ATTACHMENTS.route}` ``). But `targetRoute` comes from `ROUTES.WORKSPACE_*.getRoute()`, whose templates have **no** leading slash (`workspaces/:policyID/...`).

So `activeRoute` is `"/workspaces/A/members"` while `targetRouteWithoutParams` is `"workspaces/A/members"` → `activeRoute !== targetRouteWithoutParams` is **always true** for real workspace routes, and the intended "already on the exact target, skip the sidebar dance" short-circuit can never fire.

**Failure scenario:** no user-visible bug *today* only because the downstream `params?.policyID !== policyID` check independently prevents a redundant `setParams`. But the guard is dead in production, and any future side effect added to the transition path would now run in exactly the case this line was written to exclude. It also means the unit test for this line passes only under an unrealistic mock (see the test-file comment).

Normalize both sides before comparing (and `.replace(/\?.*/, '')` is clearer than `replaceAll` — note `replaceAll` throws unless the regex has the `/g` flag, so the `/g` there is load-bearing):

```suggestion
    const activeRoute = Navigation.getActiveRouteWithoutParams().replace(/^\//, '');
    const targetRouteWithoutParams = targetRoute.replace(/\?.*/, '');
```

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
marufsharifi: **Reinvents navigation-state types instead of reusing the established ones.**

This bespoke, all-optional `NavigationRouteWithState` plus the hand-rolled recursion below duplicates what the sibling helper `src/libs/Navigation/helpers/navigateToWorkspacesPage.ts` already does the standard way — it imports `NavigationState`/`PartialState` from `@react-navigation/native` and uses `findFocusedRoute`.

**Why it matters:** the loose type hard-codes assumptions about the sidebar route's params shape (`params?: {policyID?: string}`) and the nesting shape by hand. If the navigation state shape shifts, those assumptions drift silently rather than surfacing as a type error, and you lose real type checking on the traversal. Please reuse `NavigationState`/`NavigationRoute` + `findFocusedRoute` to stay consistent with the rest of `libs/Navigation`.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -204,6 +332,20 @@ function useNavigationSuggestions(query: string, shouldWatchForApprovals = true)
         onSelect: (searchQuery) => navigateToCannedSpendSearch(searchQuery, clearSelectedTransactions),
     });
 
+    const workspaceItems = buildWorkspaceNavigationItems({
```
marufsharifi, line 415: **This hook-level wiring is not integration-tested.** `buildWorkspaceNavigationItems` and `navigateToWorkspaceSettingsRoute` are well covered in isolation, but the wiring here — correct betas passed, `isOffline` plumbed, `getItemText` using the real `translate`, and `workspaceItems` placed between `spendItems` and `accountItems` in the sources array — has no coverage. In the unit tests `getItemText`/`getDestinationText` are hand-rolled label maps, so a wrong `translationKey` on a menu item wouldn't be caught. Consider a `renderHook`-level test asserting workspace rows appear in the right order relative to the other sources.

tests/unit/navigateToWorkspaceSettingsRouteTest.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        navigateToWorkspaceSettingsRoute(targetRoute, 'workspace-b', false);
+
+        expect(Navigation.navigate).toHaveBeenCalledWith(targetRoute);
+        expect(Navigation.setParams).not.toHaveBeenCalled();
+    });
+
+    it('compares Workspace targets without query parameters while preserving them for navigation', () => {
+        const targetRoute = ROUTES.WORKSPACE_WORKFLOWS.getRoute('workspace-a', CONST.TAB.WORKFLOWS.APPROVALS);
+        jest.mocked(Navigation.getActiveRouteWithoutParams).mockReturnValue(ROUTES.WORKSPACE_WORKFLOWS.getRoute('workspace-a'));
```
marufsharifi: **This mock returns the wrong format, so the test validates a branch that can't happen in production.**

`getActiveRouteWithoutParams()` returns a path **with** a leading slash, but this mocks it with `ROUTES.WORKSPACE_WORKFLOWS.getRoute('workspace-a')` (**no** leading slash). That makes `activeRoute === targetRouteWithoutParams`, exercising the `!isWorkspaceSettingsTransition` early-exit — but in the running app the strings never match (slash mismatch), so this code path is never hit. The test gives false confidence for the exact guard flagged in `navigateToWorkspaceSettingsRoute.ts:70-72`. Return the realistic value:

```suggestion
        jest.mocked(Navigation.getActiveRouteWithoutParams).mockReturnValue(`/${ROUTES.WORKSPACE_WORKFLOWS.getRoute('workspace-a')}`);
```

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+    params?: {
+        policyID?: string;
+    };
+    state?: {
+        key?: string;
+        routes?: NavigationRouteWithState[];
+    };
+};
+
+function findWorkspaceSidebarRoute(route: NavigationRouteWithState | undefined): {sidebarRoute: NavigationRouteWithState; splitStateKey?: string} | undefined {
```
marufsharifi: **This nav-state traversal already exists — please reuse it instead of re-walking the tree.**

`findWorkspaceSidebarRoute` / `getActiveWorkspaceSidebarRoute` walk `… → WORKSPACE_SPLIT_NAVIGATOR → SCREENS.WORKSPACE.INITIAL` and pull `policyID` out of the sidebar params. That's almost line-for-line what [`useRestoreWorkspacesTabOnNavigate.ts`](https://github.com/Expensify/App/blob/main/src/hooks/useRestoreWorkspacesTabOnNavigate.ts#L39-L59) already does — using the **real `NavigationState` types** plus the shared helpers [`getTabState`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/helpers/tabNavigatorUtils.ts#L16) and [`isWorkspaceNavigatorRouteName`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/helpers/isNavigatorName.ts#L58). The same drill-down also lives in [`Navigation.popToSidebar`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/Navigation.ts#L608).

Here we instead invent a bespoke, weaker-typed `NavigationRouteWithState` and hand-roll the recursion.

**Fix:** drop the custom walker and reuse `getTabState` + `isWorkspaceNavigatorRouteName` with the real `State<RootNavigatorParamList>` types. If a helper is warranted, it belongs in `src/libs/Navigation/helpers/`, not under `SearchRouter/`.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    if (!isWorkspaceSettingsTransition) {
+        Navigation.navigate(targetRoute);
+        return;
+    }
+
+    const workspaceSidebarRoute = getActiveWorkspaceSidebarRoute();
+    if (workspaceSidebarRoute?.sidebarRoute.key && workspaceSidebarRoute.sidebarRoute.params?.policyID !== policyID) {
+        Navigation.setParams({policyID}, workspaceSidebarRoute.sidebarRoute.key, workspaceSidebarRoute.splitStateKey);
+    }
```
marufsharifi: **Manual sidebar-policy sync duplicates the split navigator's built-in behavior.**

This `Navigation.setParams({policyID}, sidebarKey, splitStateKey)` block largely re-implements [`SplitRouter.adaptStateIfNecessary`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/AppNavigator/createSplitNavigator/SplitRouter.ts#L41-L72), which already copies the central screen's params onto the sidebar screen on wide layouts (the comment there literally says *"e.g. policyID for workspace initial"*).

**Fix:** verify whether `adaptStateIfNecessary` already keeps the sidebar `policyID` in sync on navigation — if so, this manual `getActiveWorkspaceSidebarRoute()` + `setParams` block is unnecessary and can be removed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Navigates between Workspace settings while keeping the wide-layout sidebar policy in sync.
+ */
+import Navigation from '@libs/Navigation/Navigation';
+import navigationRef from '@libs/Navigation/navigationRef';
+
+import NAVIGATORS from '@src/NAVIGATORS';
+import type {Route} from '@src/ROUTES';
+import SCREENS from '@src/SCREENS';
+
+const WORKSPACE_ROUTE_PATTERN = /^\/?workspaces\/[^/]+(\/.*)?$/;
```
marufsharifi: **This reintroduces a pattern the team deliberately removed.**

Both a `switchPolicyID`-style manual policy sync and a `workspaces/` **path regex** were intentionally deleted in commit [`7af7fba9b6d`](https://github.com/Expensify/App/commit/7af7fba9b6d) *("Remove switchPolicyID and extracting policyID from path")*, which removed `src/libs/Navigation/helpers/switchPolicyID.ts` and the `POLICY_ID_FROM_PATH` / `PATH_WITHOUT_POLICY_ID` regexes from `CONST.ts`.

`WORKSPACE_ROUTE_PATTERN = /^\/?workspaces\/[^/]+(\/.*)?$/` revives exactly that. The codebase's convention is to match routes by **screen name** ([`isWorkspaceNavigatorRouteName`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/helpers/isNavigatorName.ts#L58) + the linking config), not by path regex.

**Fix:** replace the path regex with screen-name matching, and please confirm with the author before reintroducing retired navigation patterns.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -70,6 +105,47 @@ type BuildSpendNavigationItemsParams = {
...
+            />
+            <TextWithTooltip
+                text={policy.name}
+                shouldShowTooltip
+                numberOfLines={1}
+                style={[styles.textLabelSupporting, styles.flexShrink1]}
+            />
+        </View>
+    );
+}
```
marufsharifi, line 176: **`WorkspaceIdentityCell` duplicates an existing cell and drops the avatar fallback.**

This is the same `<View flexRow … alignItemsCenter>` + workspace avatar (`XXX_SMALL`) + `<TextWithTooltip shouldShowTooltip>` layout already implemented by [`WorkspaceCell.tsx`](https://github.com/Expensify/App/blob/main/src/components/Search/SearchList/ListItem/WorkspaceCell.tsx#L20-L50) (and inlined again in [`WorkspaceTableRow.tsx`](https://github.com/Expensify/App/blob/main/src/components/Tables/WorkspaceListTable/WorkspaceTableRow.tsx#L129)).

It also uses a **raw** `<Avatar type={CONST.ICON_TYPE_WORKSPACE}>` instead of the wrapper [`WorkspaceAvatar`](https://github.com/Expensify/App/blob/main/src/components/Avatar/WorkspaceAvatar.tsx#L28), so it loses the default-workspace-avatar fallback (seeded icon + background color) when `policy.avatarURL` is missing.

**Fix:** at minimum swap the raw `<Avatar>` for `WorkspaceAvatar` (it sets `ICON_TYPE_WORKSPACE` and the fallback internally). Better: reuse `WorkspaceCell` — but it's currently `Report`-coupled and gated to EXPENSE/INVOICE types, so decouple it to accept a `policy` first, then reuse.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -143,6 +219,51 @@ function buildSpendNavigationItems({sections, icons, rightElement, getItemText,
...
+    isVendorMatchingBetaEnabled,
+    shouldUseNarrowLayout,
+    convertToDisplayString,
+    getItemText,
+    getDestinationText,
+}: BuildWorkspaceNavigationItemsParams): NavigationSuggestionSourceItem[] {
+    return Object.values(policies ?? {})
+        .filter(
+            (policy): policy is OnyxTypes.Policy =>
+                !!policy?.id && !policy.isJoinRequestPending && policy.pendingAction !== CONST.RED_BRICK_ROAD_PENDING_ACTION.DELETE && shouldShowPolicy(policy, isOffline, currentUserLogin),
```
marufsharifi, line 284: **This "accessible workspaces" filter repeats an existing predicate.**

`!policy.isJoinRequestPending && … && shouldShowPolicy(policy, …, currentUserLogin)` is already the shape used in [`useWorkspaceList.ts`](https://github.com/Expensify/App/blob/main/src/hooks/useWorkspaceList.ts#L50-L56), [`getActivePolicies` (PolicyUtils.ts)](https://github.com/Expensify/App/blob/main/src/libs/PolicyUtils.ts#L110), and [`PolicyUtils.ts#L2809`](https://github.com/Expensify/App/blob/main/src/libs/PolicyUtils.ts#L2809).

It already composes `shouldShowPolicy`, so this is minor — but note the extra `policy.pendingAction !== RED_BRICK_ROAD_PENDING_ACTION.DELETE` clause is redundant with `shouldShowPolicy` (which excludes pending-deletes when `shouldShowPendingDeletePolicy` is `false`), and this call passes `isOffline` into that `shouldShowPendingDeletePolicy` slot — a separate correctness smell worth a second look.

**Fix:** reuse `getActivePolicies(policies, currentUserLogin)` (or the `useWorkspaceList` predicate) and drop the redundant `pendingAction` check.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
```diff
@@ -143,6 +219,51 @@ function buildSpendNavigationItems({sections, icons, rightElement, getItemText,
...
+                const itemText = getItemText(item);
+                return {
+                    text: getDestinationText(itemText),
+                    singleIcon: item.icon,
+                    action: () => navigateToWorkspaceSettingsRoute(item.getRoute(), policy.id, shouldUseNarrowLayout),
+                    keyForList: `workspace_${policy.id}_${item.screenName}`,
+                    rightElement: <WorkspaceIdentityCell policy={policy} />,
+                    matchTerms: item.screenName === SCREENS.WORKSPACE.PROFILE ? [itemText, policy.name] : [itemText],
+                    sortText: policy.name,
+                };
```
marufsharifi, line 308: **All workspaces × all sub-pages are materialized (data + a React element each) regardless of the query**

For every accessible policy this calls `getWorkspaceMenuItems` (~20 feature predicates: `canMemberRead`, `shouldShowSyncError`, `arePolicyRulesEnabled`, …) producing up to ~19 items, and for **each** item allocates an `action` closure (line 257) and a `<WorkspaceIdentityCell policy={policy} />` element (line 259). For a user with dozens of workspaces that's hundreds of predicate evaluations and element allocations to surface ≤8 suggestions — built eagerly even when the query is empty, and fully rebuilt on any change inside the large, frequently-updated `POLICY` / `POLICY_CATEGORIES` collections while the router is open.

**Fix:** keep the per-policy element/closure allocation out of the eager path. Return lightweight descriptors here (text + `getRoute` + policy id + `sortText`), and construct `rightElement` / `action` only for the ≤8 rows that survive `buildNavigationSuggestions`. That turns hundreds of `<WorkspaceIdentityCell>` allocations into at most 8, and drops the wasted work when the query matches few/no workspaces.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
marufsharifi: **Extract the repeated return type into a named `type`.**

The shape `{sidebarRoute: NavigationRouteWithState; splitStateKey?: string} | undefined` is spelled out inline on both `findWorkspaceSidebarRoute` (this line) and `getActiveWorkspaceSidebarRoute` (line 48). A repeated complex object type is what our "extract complexity into a named `type`" rule targets, and duplicating it means two places to keep in sync.

```ts
type WorkspaceSidebarMatch = {
    sidebarRoute: NavigationRouteWithState;
    splitStateKey?: string;
};
```

…then return `WorkspaceSidebarMatch | undefined` from both functions.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    if (route.name === NAVIGATORS.WORKSPACE_SPLIT_NAVIGATOR) {
+        const sidebarRoute = route.state?.routes?.find((nestedRoute) => nestedRoute.name === SCREENS.WORKSPACE.INITIAL);
+        if (sidebarRoute?.key) {
+            return {sidebarRoute, splitStateKey: route.state?.key};
+        }
+    }
+
+    const routes = route.state?.routes ?? [];
+    for (let index = routes.length - 1; index >= 0; index--) {
```
marufsharifi: **Coverage gap: recursive descent is never exercised.**

`findWorkspaceSidebarRoute` recurses into `route.state.routes` to locate a `WORKSPACE_SPLIT_NAVIGATOR` nested below the root — but `mockWorkspaceNavigationState` in `navigateToWorkspaceSettingsRouteTest.ts` always places the split navigator as the sole top-level route, so this loop never runs. A regression that broke the recursion would pass every current test, even though recursion is the reason this helper exists.

**Add** a test *"updates the sidebar when the Workspace split navigator is nested below the root"*: build a root state whose route is a wrapper stack that **contains** the `WORKSPACE_SPLIT_NAVIGATOR`, trigger a workspace→workspace transition, and assert `Navigation.setParams` is still called with the sidebar `key`.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+    const targetRouteWithoutParams = targetRoute.replaceAll(/\?.*/g, '');
+    const isWorkspaceSettingsTransition = WORKSPACE_ROUTE_PATTERN.test(targetRouteWithoutParams) && activeRoute !== targetRouteWithoutParams && WORKSPACE_ROUTE_PATTERN.test(activeRoute);
+
+    if (!isWorkspaceSettingsTransition) {
+        Navigation.navigate(targetRoute);
+        return;
+    }
+
+    const workspaceSidebarRoute = getActiveWorkspaceSidebarRoute();
+    if (workspaceSidebarRoute?.sidebarRoute.key && workspaceSidebarRoute.sidebarRoute.params?.policyID !== policyID) {
```
marufsharifi: **Coverage gap: workspace transition with no resolvable sidebar route.**

When it's a genuine workspace→workspace transition but `getActiveWorkspaceSidebarRoute()` returns `undefined` (root state has no split navigator, or the sidebar route has no `key`), the expected behavior is: skip `setParams`, still call `Navigation.navigate`. No test asserts this fallback — every transition test supplies a valid sidebar via `mockWorkspaceNavigationState`.

**Add** a test that sets up a workspace transition against a root state with **no** `WORKSPACE_SPLIT_NAVIGATOR`, asserting `Navigation.setParams` is **not** called and `Navigation.navigate` **is** called with the target route.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+
+function navigateToWorkspaceSettingsRoute(targetRoute: Route, policyID: string, shouldUseNarrowLayout: boolean) {
+    if (shouldUseNarrowLayout) {
+        Navigation.navigate(targetRoute);
+        return;
+    }
+
+    const activeRoute = Navigation.getActiveRouteWithoutParams();
+    const targetRouteWithoutParams = targetRoute.replaceAll(/\?.*/g, '');
+    const isWorkspaceSettingsTransition = WORKSPACE_ROUTE_PATTERN.test(targetRouteWithoutParams) && activeRoute !== targetRouteWithoutParams && WORKSPACE_ROUTE_PATTERN.test(activeRoute);
```
marufsharifi: `activeRoute` and `targetRouteWithoutParams` are in **different formats**, so `activeRoute !== targetRouteWithoutParams` is always `true` and the "same page" short-circuit never fires.

- `Navigation.getActiveRouteWithoutParams()` returns a path **with** a leading slash — see the canonical comparison in [`VideoPlayerPreview/index.tsx`](https://github.com/Expensify/App/blob/main/src/components/VideoPlayerPreview/index.tsx#L60): `=== `/${ROUTES.REPORT_ATTACHMENTS.route}``.
- `getRoute()` returns **no** leading slash (e.g. `workspaces/${policyID}/workflows`).

So `'/workspaces/a/members' !== 'workspaces/a/members'` is always true. Today it's masked by the downstream `sidebarRoute.params?.policyID !== policyID` guard (which skips `setParams` when the workspace is unchanged), so there's no user-visible bug — but the guard is effectively dead code that doesn't do what it reads as, and the unit test for it passes for the wrong reason (see my comment on `navigateToWorkspaceSettingsRouteTest.ts`).

Normalize both sides to the same format:

```suggestion
    const targetRouteWithoutParams = `/${targetRoute.replaceAll(/\?.*/g, '')}` as const;
    const isWorkspaceSettingsTransition = WORKSPACE_ROUTE_PATTERN.test(targetRouteWithoutParams) && activeRoute !== targetRouteWithoutParams && WORKSPACE_ROUTE_PATTERN.test(activeRoute);
```

`WORKSPACE_ROUTE_PATTERN` starts with `^\/?`, so it still matches with the leading slash.

src/libs/Navigation/helpers/navigateToWorkspaceSettingsRoute.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
marufsharifi, line 1: This file traverses `navigationRef.getRootState()` and dispatches `Navigation.setParams` — it's a navigation-state helper, not Search-Router UI. The established home for this is `src/libs/Navigation/helpers/`, right next to [`navigateToWorkspacesPage.ts`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/helpers/navigateToWorkspacesPage.ts), which does very similar sidebar/tab restoration.

Suggest moving it to `src/libs/Navigation/helpers/navigateToWorkspaceSettingsRoute.ts` so it's discoverable for reuse and the layering stays clean. The import updates in `useNavigationSuggestions.tsx` and the two test files are mechanical.

marufsharifi (changes requested):

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Addressed.

tests/unit/navigateToWorkspaceSettingsRouteTest.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed this one too after investigating more, thanks!

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Thanks for checking this. `adaptStateIfNecessary()` only copies `policyID` when creating a missing sidebar route; it does not update an existing sidebar when switching workspaces. Removing this synchronization would restore the stale Workspace LHN issue, so I’m keeping it.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Thanks for checking this. `getActivePolicies()` has different filtering semantics, and the explicit pending-delete check is intentional because deleted workspaces must never be navigable, including offline or errored deletions. I’m keeping the current filter to avoid changing Workspace visibility behavior.

src/components/Search/SearchRouter/useNavigationSuggestions.tsx
nabi-ebrahimi: Thanks for raising this. `getWorkspaceMenuItems()` must still run per workspace because its permission and feature checks determine valid destinations. Deferring row elements would only reduce allocations while requiring a broader shared-pipeline change, so I’m keeping the tested approach unless profiling shows a measurable issue.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: This is covered by the normal-navigation fallback test, which verifies that navigation still occurs without a resolvable Workspace sidebar and `setParams` is not called. The earlier transition structure has also changed, so no additional test is needed.

src/components/Search/SearchRouter/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

nabi-ebrahimi:
@marufsharifi Thanks for the review. All applicable comments have been addressed and verified. Could you please review again?

src/libs/Navigation/helpers/navigateToWorkspaceSettingsRoute.ts
nabi-ebrahimi: Addressed.

marufsharifi (approved):
LGTM

carlosmiceli (dismissed):

nabi-ebrahimi:
@trjExpensify I think we are waiting for your review, Thanks!

nabi-ebrahimi:
@trjExpensify and @carlosmiceli Can you please help us here and merge this, Thanks!

trjExpensify (dismissed):
Core PR for the navigational search rollout. 👍

trjExpensify:
I've been OOO until today. I'll leave the merge for @carlosmiceli.

carlosmiceli:
@nabi-ebrahimi there seems to be confclits 🙏

nabi-ebrahimi:
> @nabi-ebrahimi there seems to be confclits 🙏

@carlosmiceli sorry for the delay, resolved the conflicts.

This failure is also unrelated to this PR.

carlosmiceli:
@nabi-ebrahimi tests are failing now 🫂

trjExpensify:
Ran a build on this to confirm the navigation from inside the workspace editor doesn't have the same bug as domains found [here](https://github.com/Expensify/App/issues/100112#issuecomment-5507564284). Confirmed it doesn't. 👍

marufsharifi:
@nabi-ebrahimi, could you please merge the main?  thanks.

marufsharifi:
@carlosmiceli, the test failure was already present on main, so I believe we can proceed with merging this. Thanks!

carlosmiceli:
@nabi-ebrahimi let's do one more main merge to see if it fixes the test, we'll merge if not anyway,.

nabi-ebrahimi:
@carlosmiceli, I merged the `main`, and everything passed successfully. All yours now. Thanks!

carlosmiceli (approved):

mitarachim:
Deploy Blocker https://github.com/Expensify/App/issues/100369 was identified to be related to this PR.

PR #98006, opened 2026-08-06
[No QA] Fix Pusher onResubscribe firing twice on web and on the first handshake on native

adhorodyski:
### Explanation of Change

The private user channel subscribes twice, once for `PONG` and once for `MULTIPLE_EVENTS`. Each call passed its own `onResubscribe`, so one socket drop sent two `ReconnectApp` requests. The duplicate came from the number of registrations at the call site, not from the handler binding.

`Pusher.onChannelResubscribe(channelName, callback)` replaces the fourth parameter of `subscribe()`. Each registration owns its handler and its own first-handshake flag, so one drop runs it one time. `subscribeToUserEvents` now registers one time instead of two, and `usePusherDraftPacing` one time instead of six. Draft clearing also survives when typing events create the report channel first.

`subscribe()` calls `socket.subscribe()` every time. pusher-js ignores it when the channel is subscribed, reinstates one that a screen cancelled mid-handshake, and authorizes again after a failure. A channel that exists but does not work is no longer treated as usable. Every caller that arrives before the handshake binds its own error handler, so all of them learn when authorization fails.

Native fired `onResubscribe` on the first handshake, before any drop. It now reads the channel status before it overwrites it. Native also bound the event callback again on every handshake, and `eventsBoundToChannels` survives reconnects, so each handshake left another copy behind.

One behaviour change: `subscribe()` no longer resolves early for a caller that arrives mid-handshake. It waits for the handshake.

Measured over `2026-08-05T10:30:00Z` to `2026-08-05T14:30:00Z`.

| Measure | Today | Target |
| --- | --- | --- |
| Web moments with two resubscribe handlers | 82,150 | under 1,000 |
| Web moments with three or more handlers | 244 | 0 |
| iOS resubscribes with no socket drop | 1.11 per user, 6,743 total | under 0.2 per user |
| Pushed `ReconnectApp` requests | 184,212 | 2,431 fewer, 1.3% |

### Fixed Issues

$ https://github.com/Expensify/App/issues/98038
PROPOSAL:

### Tests

1. Run `npx jest tests/unit/PusherSubscribeWebTest.ts tests/unit/PusherSubscribeTest.ts`. All 15 tests pass. Three tests are new. Two run on web: two callers in one tick fire `onResubscribe` one time, and a caller inside a reconnect window binds no extra handler. One runs on native: nothing fires on the first handshake, and one call follows every later handshake.
2. Revert the two changes in `src/libs/Pusher/` and run the same command. Three tests fail. Web reports `Received number of calls: 2`. Native reports `Received number of calls: 1` on the first handshake.
3. Web, manual: sign in and search the console for `Pusher re-subscribed to private user channel`. Kill the network, then restore the network, so Pusher re-subscribes. Verify that the line appears one time for each resubscribe.
4. iOS, manual: sign in and watch the log for `Pusher re-subscribed to private user channel`. Verify that no line appears during a clean boot, where the socket never drops. Then background the app until the socket drops. Foreground the app and verify that one line appears.

### Offline tests

Nothing changes offline. `Pusher.subscribe()` waits on `initPromise`, and `initPromise` settles only when the socket connects. The code never reaches either branch of the gate while the app is offline. The callback that now runs less often is `reconnect()`, and `reconnect()` already does nothing until the client is online.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

No UI change. Verified with unit tests in `tests/unit/PusherSubscribeTest.ts`.

</details>

<details>

No UI change. Verified with unit tests in `tests/unit/PusherSubscribeWebTest.ts`.

</details>

<details>

No UI change. Verified with unit tests in `tests/unit/PusherSubscribeTest.ts`.

</details>

<details>

No UI change. Verified with unit tests in `tests/unit/PusherSubscribeWebTest.ts`.

</details>

<details>

No UI change. Verified with unit tests in `tests/unit/PusherSubscribeWebTest.ts`.

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/libs/Pusher/index.ts
```diff
@@ -262,7 +262,7 @@ function subscribe<EventName extends PusherEventName>(
                         Log.info('[Pusher] Attempting to subscribe to channel', false, {channelName, eventName});
                         let channel = getChannel(channelName);
 
-                        if (!channel?.subscribed) {
+                        if (!channel) {
                             channel = socket.subscribe(channelName);
                             let isBound = false;
                             channel.bind('pusher:subscription_succeeded', () => {
```
mkhutornyi, line 268: - a channel cancelled mid-reconnect is never reinstated
- no retry after `pusher:subscription_error`
- secondary callers no longer learn about subscription failures
- later subscribers' `onResubscribe` discarded

Suggested change which fixes all of them:

Track per-channel state in the module instead of inferring it from pusher-js's registry:
```
const channelWaiters = new Map<string, Array<{resolve: () => void; reject: (e: unknown) => void}>>();

channel = socket.subscribe(channelName);   // ALWAYS — reinstates cancelled, retries errored, no-ops when subscribed
if (!channelWaiters.has(channelName)) {
    channelWaiters.set(channelName, []);
    channel.bind('pusher:subscription_succeeded', ...);  // one handler per channel; drains waiters, then fires resubscribe callbacks
    channel.bind('pusher:subscription_error', ...);      // drains waiters with reject
}

```

BartekObudzinski: Agreed on all four, went with your design.

`socket.subscribe()` now runs on every call, which covers the three states the old guard collapsed. Channel state lives in a module level map instead of being read off pusher-js's registry, so waiters resolve or reject individually and every subscriber's `onResubscribe` is kept.

Two things on top of the sketch.

The bind once guard cannot key on the map entry alone. `socket.unsubscribe()` on a subscribed channel hits `channels.remove`, so the next `subscribe()` gets a fresh channel object with no bindings. The guard would skip re-binding and leave the channel deaf. State is now cleared alongside pusher-js at every teardown site.

Keeping every handler also reintroduces the bug this PR fixes. `subscribeToPrivateUserChannelEvent` runs twice for the same user channel, once for `pong` and once for `multipleEvents`, and `onPusherResubscribeToPrivateUserChannel` was declared inside it. Two closures means two `reconnect()` calls per drop, and `reconnect()` has no dedupe of its own. So handlers are deduped by function reference and that one is now module level. `usePusherDraftPacing` already shares a single handler, so it collapses for free.

Worth flagging: reference identity is the weak part. It gives us one call per distinct function, when what we want is one `reconnect()` per channel resubscribe. Those agree only because that caller passes the same reference twice. A channel scoped registration would be the real fix. Happy to do it here if you want it, otherwise I'll open a follow up.

Five new tests cover this and all fail if you revert the source.

src/libs/Pusher/index.ts
```diff
@@ -262,7 +262,7 @@ function subscribe<EventName extends PusherEventName>(
                         Log.info('[Pusher] Attempting to subscribe to channel', false, {channelName, eventName});
                         let channel = getChannel(channelName);
 
-                        if (!channel?.subscribed) {
+                        if (!channel) {
```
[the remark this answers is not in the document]
BartekObudzinski: Correct, fixed. `socket.subscribe()` is now unconditional, so a channel still in the map after `pusher:subscription_error` gets a fresh attempt instead of resolving through the `else` path. There's a test that fails the subscription, asserts the rejection, subscribes again and checks pusher-js was asked to retry.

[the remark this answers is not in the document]
BartekObudzinski: Right, and the `subscribeToReportTypingEvents` ordering is the case that mattered. Every subscriber's `onResubscribe` is tracked per channel now, so the draft pacing handler still runs after a reconnect whichever caller created the channel. Handlers are deduped by function reference, which keeps the six `usePusherDraftPacing` subscriptions at one call. Test covers a late subscriber joining an already created channel and still getting its call.

src/libs/Pusher/index.native.ts
```diff
@@ -215,17 +217,22 @@ function bindEventToChannel<EventName extends PusherEventName>(
     return boundCb;
 }
 
+function onChannelResubscribe(channelName: string, callback: () => void) {
+    const callbacks = resubscribeCallbacks.get(channelName) ?? [];
+    resubscribeCallbacks.set(channelName, callbacks);
+    callbacks.push(callback);
+
+    return () => {
+        callbacks.splice(callbacks.indexOf(callback), 1);
```
mkhutornyi: If the new registration happens _before_ the prior disposer runs and the prior callback is no longer in the array, `indexOf` returns `-1` and `splice(-1, 1)` deletes the **newly registered** callback, so `reconnect()` then never fires at all.

mkhutornyi (commented):
Please check codex comments

adhorodyski:
@codex review

mountiny:
@MelvinBot review

src/libs/Pusher/index.native.ts
```diff
@@ -276,9 +276,12 @@ function subscribe<EventName extends PusherEventName>(
                                     }
                                 },
                                 onSubscriptionSucceeded: () => {
+                                    // Fires again on every later handshake, so read the status before overwriting it.
+                                    const wasSubscribed = channels[channelName] === CONST.PUSHER.CHANNEL_STATUS.SUBSCRIBED;
```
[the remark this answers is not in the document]
adhorodyski: This is real but narrow. It needs auth failure + a second drop. It only costs us one lost `reconnect()`, then the same handshake sets `SUBSCRIBED`, so next drop counts right again. 

I do not want to fix this with a second copy of channel state, that duplication is what this PR deletes on web. Leaving as is.

mountiny:
@mkhutornyi over to you

src/libs/Pusher/index.ts
```diff
@@ -260,55 +290,56 @@ function subscribe<EventName extends PusherEventName>(
...
+                            return;
+                        }
+
+                        const onSubscriptionSucceeded = () => {
+                            unbindHandshakeHandlers();
+                            bindAndResolve();
+                        };
+
+                        const onSubscriptionError = (data: PusherSubscriptionErrorData = {}) => {
+                            unbindHandshakeHandlers();
```
mkhutornyi: This unbinds both `pusher:subscription_succeeded` and `pusher:subscription_error`.
When pusher-js later re-authorizes the channel on its own, nothing binds the subscriber's event callback so the channel is subscribed at the socket level and the app receives nothing.

Should we unbind only the error handler here, leaving the success handler bound?

@MelvinBot can you confirm?

MelvinBot: Confirmed — this is a real regression, and your suggested fix is correct.

pusher-js (8.3.0) keeps an errored channel in its registry — `subscription_error` only sets `subscriptionPending = false`, it never removes the channel — and on the next reconnect its `connected` handler runs `subscribeAll()`, which re-authorizes every channel still in the registry and re-emits `pusher:subscription_succeeded`. Because [`onSubscriptionError`](https://github.com/Expensify/App/blob/df7ac0e2f1b988743c8598c0b7a54487ba5391fc/src/libs/Pusher/index.ts#L316-L326) calls `unbindHandshakeHandlers()` (which removes **both** handlers), that later success has no listener from this `subscribe()` call, so `bindEventToChannel` never runs. The socket is subscribed but the app's event callback is gone — silent data loss. For the private user channel this is nasty: `subscribeToUserEvents` only calls `Pusher.subscribe` once and just logs on reject, so a single transient `AuthenticatePusher` failure could leave real-time updates permanently dead until the next full teardown.

This is a behavior change from the pre-PR code, where the success handler was bound once and never unbound on error, so an eventual re-subscribe still bound the callback (via the `if (!isBound)` path).

Leaving the success handler bound restores that recovery:
- On eventual re-subscribe success, `onSubscriptionSucceeded` fires → binds the event callback → the stray `resolve()` is a no-op since the promise already rejected. ✅
- The `onChannelResubscribe` handler still won't misfire on that recovery success — its `hasSubscribed` flag is still `false` from the failed first attempt, so it treats the first real success as the initial subscription, not a resubscribe. ✅

Two things worth doing alongside the change:
1. Add a unit test in `PusherSubscribeWebTest.ts` for exactly this path — subscribe → `failAuthorization()` (reject) → `startSubscription()` + `completeHandshake()` on the same channel → assert the event callback now receives an event. The current suite covers re-auth on a *new* `subscribe()` call but not pusher-js's own auto-resubscribe after an error.
2. Consider the multi-caller case: with the handler left bound, every rejected waiter's event callback also rebinds on recovery, which is the intended (pre-PR) behavior — just confirm that's what you expect for callers that already handled the rejection.

Note there's a separate, related concern in the same area worth a look regardless of this fix: `onChannelResubscribe`'s `indexOf`/`splice(-1, 1)` disposer flagged elsewhere can delete a newly-registered callback when re-registration races the old disposer.

adhorodyski: On the disposer note: line 227 returns early when `indexOf` gives `-1`, so a stale disposer removes nothing. `onPrivateUserChannelResubscribe` also passes a new closure on every call, so two registrations never share one reference.

adhorodyski:
@mkhutornyi updated

mkhutornyi (approved):

mkhutornyi:
> iOS, manual: sign in and watch the log for `Pusher re-subscribed to private user channel`. Verify that no line appears during a clean boot, where the socket never drops. Then background the app until the socket drops. Foreground the app and verify that one line appears.

Instead of backgrounding, I disabled network and then enabled to see the log.

tests/unit/PusherSubscribeTest.ts
mountiny, line 1: Can you please update the tests and your skills to include and follow the documenting tests guidelines wiht the Given, When, Then comments https://github.com/Expensify/App/blob/main/tests/README.md#documenting-tests

mountiny (changes requested):

mountiny (approved):
Thanks for updating this

Issue #97891, opened 2026-08-05
[Due for payment 2026-09-18] Staging server toggle resets on app force-close in GH release build


Julesssss:
Slack Thread Context: https://expensify.slack.com/archives/C9YU7BX5M/p1774961694877359?thread_ts=1774961694.877359&cid=C9YU7BX5M

**^ Original report:**

> Following up on this request: the Staging toggle is disabled by default on a fresh installation. It would be great if this setting could be enabled by default for staging builds, as it would save testers from having to enable it manually every time.
This behavior is observed on both Android and iOS builds.
 
https://github.com/user-attachments/assets/36c67fce-b056-45ee-967d-e4983d2163b4

**What the problem is:**
When the staging Android app is force-closed and relaunched, the "Use staging server" toggle automatically turns off, reverting to production. This occurs only in the GH release build and is not reproducible in the Play Store build. The issue is inconsistent but observed during force-close with account switch, cache clear, rapid PR verification actions, and NewDot/OldDot switching.

**What the expected output is:**
The "Use staging server" toggle setting should persist across app sessions (force-close, relaunch, logout). On cold app start, the app should correctly detect and restore the staging server preference instead of defaulting to production.

**Debugging info:**
- Issue reproducible only in GH release build, not in Play Store build
- Observed scenarios: force-close with account switch, app cache clear, rapid actions during PR verification, NewDot/OldDot switching
- Setting is stored locally in Onyx (not server-persisted)
- Related code: [App.ts#L138](https://github.com/Expensify/App/blob/c600f7bf33b6811e21f4ca14a45fbccfda53b09f/src/libs/actions/App.ts#L138) (listed as key to preserve on logout)
- Suspected bug: [ApiUtils.ts#L14-L34](https://github.com/Expensify/App/blob/c600f7bf33b6811e21f4ca14a45fbccfda53b09f/src/libs/ApiUtils.ts#L14-L34) (may incorrectly assume production on cold start instead of staging)

**Potential solutions or implementation notes:**
- Verify that cold app start logic correctly reads and restores the staging server preference from Onyx
- Investigate why the GH release build exhibits this behavior while Play Store build does not

Created with the :create-gh-issue: automation

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>
<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
[...]

PR #98149, opened 2026-08-10
Fix staging server toggle resetting on cold start in release builds

mikolajpochec:
### Explanation of Change

Fixes the NewDot staging server toggle resetting itself on cold start. Two independent causes, which is why it looked intermittent: the first fires on every HybridApp cold start, the second only on Android after a production deploy or a rate-limit burst.

**Why it reset.** `setupNewDotAfterTransitionFromOldDot` merged every key of OldDot's transition payload into Onyx unfiltered, and `shouldUseStagingServer` is one OldDot sends on every transition. NewDot's value survived sign-out (it's in every `KEYS_TO_PRESERVE` list) but not that handoff, which happens on every mount. Separately on Android, `betaChecker` decides the environment by comparing `pkg.version` against the latest GitHub release, so once that version was promoted to production the installed build stopped being "ahead of production" and `ApiUtils` force-disabled the toggle — Onyx value intact, `TestToolMenu` still showing ON while requests went to production. A rate-limited response (no `tag_name`) reached the same outcome and wrote `IS_BETA = false`, poisoning the offline fallback for later launches.

**What changed.** OldDot now bumps a `stagingToggleGeneration` counter when the user deliberately flips its toggle (see linked PR), so NewDot can tell that apart from the ambient flag. NewDot stores the last generation it accepted in the new `ONYXKEYS.LAST_ACCEPTED_OLD_DOT_STAGING_TOGGLE_GENERATION` key (preserved alongside `ONYXKEYS.SHOULD_USE_STAGING_SERVER`) and applies OldDot's flag only on the first transition (seeding the key) or when the generation is higher — a lower one means OldDot wiped its state, so we re-baseline instead of ignoring every later flip. `HybridAppHandler` waits for both keys to load before comparing.

On Android, `ONYXKEYS.IS_BETA` is replaced by `ONYXKEYS.BETA_BUILD_VERSION`, pinning the beta verdict to the version it was made for and short-circuiting the `isBetaBuild()` fetch; rate-limited or non-semver responses now leave the stored value untouched. `getEnvironment` caches the promise rather than the resolved value, so concurrent startup callers share one beta check instead of each firing its own GitHub request.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97891
PROPOSAL:

### Tests

### QA Steps

### PR Author Checklist

    - [ ] Android: Native
    - [ ] Android: mWeb Chrome
    - [ ] iOS: Native
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


Guccio163 (approved):
Generally looks good to me, when I was chatting with Claude it found a nitpick cache problem; It has a low importance, but I'm pasting it below:

```
Mechanism: environmentPromise ??= resolveEnvironment() caches the Promise object, not its resolved value. ??= only re-assigns if left side is null/undefined. A rejected Promise is still a non-null object reference — so once cached, ??= never fires again, ever, for rest of app session.

Old code (src/libs/Environment/getEnvironment/index.native.ts, pre-PR) cached the plain value environment: Environment | null. Failure path = environment stays null = falsy = next call retries whole resolution from scratch. Built-in retry, free.

New code trades that away. If resolveEnvironment()'s returned promise ever rejects, environmentPromise holds that dead promise forever. Every future getEnvironment() call returns same rejected promise → rejects again → forever. No retry path exists anymore.

Concrete trigger check — where could resolveEnvironment() actually reject:

- betaChecker.isBetaBuild() android variant (src/libs/Environment/betaChecker/index.android.ts): every branch returns/catches, ends in .catch(() => false). Never rejects. Safe.
- betaChecker.isBetaBuild() iOS variant (index.ios.ts, untouched by this PR):
function isBetaBuild(): IsBetaBuild {
        NativeModules.EnvironmentChecker.isBeta().then((isBeta) => {
            resolve(isBeta);
        });
    });
}
No .catch() here. If native EnvironmentChecker.isBeta() rejects (native exception, module not linked, whatever), rejection propagates up untouched, into resolveEnvironment()'s returned promise, into cached environmentPromise. That's the one real path to permanent poisoning.
- Config?.ENVIRONMENT access itself throwing synchronously — not actually a risk: sync throw during ??= RHS evaluation means assignment never completes, environmentPromise stays null, next call retries fine. Ruled out.

So severity low because: this PR only touches the Android beta-check path (which is reject-safe), and the one real gap (iOS, no .catch) predates this PR entirely — untouched, same risk existed in spirit before too (old code called betaChecker.isBetaBuild().then(...) the same way, no catch on iOS there either). This PR just changes what gets cached (promise vs value), which flips a pre-existing "iOS native call could theoretically reject" from "self-healing on next call" to "permanently stuck." Real bug precondition (native module error) is rare/never-seen-in-practice, hence nitpick not blocker.
```

src/libs/actions/Session/index.ts
```diff
@@ -783,15 +790,41 @@ function setupNewDotAfterTransitionFromOldDot(hybridAppSettings: HybridAppSettin
...
                 {
                     onyxMethod: Onyx.METHOD.MERGE,
                     key: ONYXKEYS.HYBRID_APP,
                     value: hybridAppUpdate,
                 },
             ];
 
+            // OldDot bumps the generation only on deliberate toggle flips, so an unaccepted generation marks a real
+            // change rather than the ambient value sent on every transition.
+            const oldDotStagingToggleGeneration = hybridApp?.stagingToggleGeneration ?? 0;
```
staszekscp, line 803: Why is this called "stagingToggleGeneration"? I'm confused because of the word "generation" - to me it doesn't explain what it does 😄

staszekscp (commented):
In general this solution seems to me a bit like an overkill with all the counters - what was the reasoning behind exactly this solution? It feels to me like a cleaner way to solve this is to have eg. a native module that would sync both values between OldDot and NewDot 🤔

PR #98189, opened 2026-08-10
Duplicate GPS distance expenses as manual distance

x-dev90:
### Explanation of Change

GPS distance expenses are now duplicated as manual distance expenses because the saved transaction no longer contains the original GPS coordinate trace.

The duplicate flow now maps `DISTANCE_GPS` to `DISTANCE_MANUAL`, skips copied GPS endpoint waypoints, keeps the saved measured distance, and uses the duplicated request type consistently for both workspace and self-DM duplicates.

Added tests for GPS distance duplication in both workspace and self-DM flows, plus coverage to confirm normal map distance duplicates still preserve waypoints.
### Fixed Issues

$https://github.com/Expensify/App/issues/91814
PROPOSAL:https://github.com/Expensify/App/issues/91814#issuecomment-5005880544

### Tests

1. Open the app and sign in.
2. Open self-DM.
3. Create a GPS distance expense.
4. Open the created GPS distance expense.
5. Tap More > Duplicate.
6. Open the duplicated expense.
7. Verify the duplicated expense opens without an unexpected error.

### Offline tests

N/A
### QA Steps

Same as tests

### PR Author Checklist

    - [X] Android: Native
    - [X] Android: mWeb Chrome
    - [X] iOS: Native
    - [X] iOS: mWeb Safari
    - [X] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/2c8e327b-00d8-4b9b-9636-dca45e9b4dbb

</details>

<details>

</details>

<details>

>

https://github.com/user-attachments/assets/a99d72f4-ad07-451e-90ec-83c379b13ccd

</details>

<details>

>

https://github.com/user-attachments/assets/7fdf456e-632d-48f9-81e6-42b8a33be3eb

</details>

<details>

>

https://github.com/user-attachments/assets/c03faa38-21b1-4b8f-b5c3-af14b1d5bb7c

</details>


x-dev90:
> @x-dev90, could you please check the failed actions. thanks.

Hi @marufsharifi, all actions have now passed successfully. Could you please review them again when you have a chance? Thank you.

garrettmknight (dismissed):
Looks good from a product perspective.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -648,10 +649,18 @@ function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, tra
...
  */
 function getDuplicateRequestType(transaction: OnyxTypes.Transaction) {
     const sourceRequestType = getRequestType(transaction);
-    return sourceRequestType === CONST.IOU.REQUEST_TYPE.SCAN ? CONST.IOU.REQUEST_TYPE.MANUAL : sourceRequestType;
+    if (sourceRequestType === CONST.IOU.REQUEST_TYPE.SCAN) {
+        return CONST.IOU.REQUEST_TYPE.MANUAL;
+    }
+    if (sourceRequestType === CONST.IOU.REQUEST_TYPE.DISTANCE_GPS) {
+        return CONST.IOU.REQUEST_TYPE.DISTANCE_MANUAL;
+    }
```
marufsharifi: The "a GPS distance duplicate becomes a manual distance" rule is now encoded in **two** independent places that must stay in lockstep:

- here (type remap): `if (sourceRequestType === CONST.IOU.REQUEST_TYPE.DISTANCE_GPS) return CONST.IOU.REQUEST_TYPE.DISTANCE_MANUAL;`
- and in `buildDuplicateTransactionParams` (line ~609): `const shouldDuplicateAsManualDistance = isGPSDistanceRequest(transaction);` which drives both the waypoint-strip and the `distance` copy.

If a future type ever needs the same treatment, a contributor has to remember to edit both. Getting only one right produces exactly the class of silent bug this PR fixes (type says `distance-manual` but the endpoint waypoints are still copied → duplicate reopens with a stale route).

Suggest a single predicate both sites consume:

```ts
// GPS distance duplicates must become manual distances: the saved transaction keeps the measured
// quantity but not the raw GPS trace, so the endpoint waypoints can't be reused as a real route.
function shouldDuplicateAsManualDistance(transaction: OnyxTypes.Transaction) {
    return isGPSDistanceRequest(transaction);
}
```

Then this branch reads `if (shouldDuplicateAsManualDistance(transaction)) return CONST.IOU.REQUEST_TYPE.DISTANCE_MANUAL;`. Non-blocking — just future-proofing the coupling.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -606,7 +606,8 @@ function resolveDuplicates({
  */
 function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, transactionDetails: ReturnType<typeof getTransactionDetails>) {
     const {linkedTrackedExpenseReportAction, ...transactionWithoutLinkedAction} = transaction;
-    const waypoints = !isExpenseSplit(transaction) ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
+    const shouldDuplicateAsManualDistance = isGPSDistanceRequest(transaction);
+    const waypoints = !isExpenseSplit(transaction) && !shouldDuplicateAsManualDistance ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
```
marufsharifi: Flagging that this line carries an `as WaypointCollection | undefined` cast (the repo forbids type casting). It's **pre-existing** — carried over from the original ternary, not introduced by this PR — so it's out of scope to fix here. But since you're already editing the line: if `getTransactionDetails`' return type typed `waypoints` precisely, this cast could be dropped. Leave it unless you're comfortable touching that return type.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -638,7 +639,7 @@ function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, tra
         unit: transaction.comment?.units?.unit,
     };
 
-    if (isDistanceRequest(transaction) && (isExpenseSplit(transaction) || isOdometerDistanceRequest(transaction))) {
+    if (isDistanceRequest(transaction) && (isExpenseSplit(transaction) || isOdometerDistanceRequest(transaction) || shouldDuplicateAsManualDistance)) {
```
marufsharifi:
```ts
if (isDistanceRequest(transaction) && (isExpenseSplit(transaction) || isOdometerDistanceRequest(transaction) || shouldDuplicateAsManualDistance)) {
    transactionParams.distance = transaction.comment?.customUnit?.quantity ?? undefined;
}
```

This copies `customUnit.quantity` straight into `distance`, reusing the exact odometer/split path — good, so if that path is correct this is too. The only thing worth confirming: `quantity` must be in the **display unit** (mi/km), since that's what the manual-distance creation expects. The added tests only cover `DISTANCE_UNIT_MILES`. Could you add one GPS case with a **kilometers** workspace unit to lock in that a km GPS duplicate keeps the right distance/amount? That's the most likely spot for a subtle unit bug to hide.

tests/actions/IOUTest/DuplicateTest.ts
```diff
@@ -2976,6 +3132,50 @@ describe('actions/Duplicate', () => {
...
+
+            duplicateReport(getDefaultParams([gpsDistanceTx]));
+            await waitForBatchedUpdates();
+
+            const distanceCall = writeSpy.mock.calls.find(isWriteMockCallForCommand(WRITE_COMMANDS.CREATE_DISTANCE_REQUEST));
+            expect(distanceCall).toBeDefined();
+            expect(distanceCall?.[1]).toEqual(
+                expect.objectContaining({
+                    distance: 42,
+                    distanceRequestType: CONST.IOU.REQUEST_TYPE.DISTANCE_MANUAL,
```
marufsharifi: Nice — this asserts `distanceRequestType: DISTANCE_MANUAL` for GPS. One gap to close while you're here: the refactor changed the `distanceRequestType` **source** for *every* distance type (`getDistanceRequestType(transaction)` → `getDuplicateRequestType(transaction)` in `createExpenseByType`), not just GPS. The existing regression test `should preserve waypoints for non-split distance expenses` only asserts `waypoints).not.toBe('null')` and never checks the request type, so the non-GPS path is unguarded.

Add one line there to lock in the non-regression:

```ts
expect(distanceCall?.[1]).toEqual(
    expect.objectContaining({distanceRequestType: CONST.IOU.REQUEST_TYPE.DISTANCE}),
);
```

The values are equal today for plain/`map` distance, so this is low risk — but the assertion prevents a future change to `getDuplicateRequestType` from silently breaking normal distance duplicates.

x-dev90: Thanks for the feedback. I added a regression assertion confirming non-GPS distance duplicates retain the DISTANCE request type while preserving waypoints. Please review.

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the feedback. I added kilometer workspace coverage and verified the distance is preserved when duplicating GPS expenses as manual distance. Please review.

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the feedback. I centralized the GPS-to-manual-distance logic in `shouldDuplicateAsManualDistance` and reused it in both code paths.Please review.

x-dev90:
@marufsharifi , gentle bump. Thanks!

x-dev90:
@marufsharifi, quick bump—thanks!

marufsharifi:
@x-dev90, could you please, merge the main. thanks.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -907,6 +922,7 @@ function duplicateExpenseTransaction({
             transactionParams: {
                 ...(params.transactionParams ?? {}),
                 validWaypoints: waypoints,
+                ...(isDistanceRequest(transaction) && {distanceRequestType: duplicateRequestType}),
```
marufsharifi: **This behavioral change isn't covered by a test.** The guard here is `isDistanceRequest(...)`, not "is GPS" — so *every* distance type duplicated into a self-DM now sends `distanceRequestType` on the `TRACK_EXPENSE` payload, where previously non-GPS distances sent nothing here.

I traced `TrackExpense.ts` (`distanceRequestType` is only forwarded as an API param, and optimistic waypoint handling keys off `validWaypoints`, which is still preserved for non-GPS), so I don't believe this regresses — for a normal map distance it now sends `'distance'`/`'distance-map'` instead of `undefined`, which is arguably more correct.

But the new self-DM test only exercises GPS, so this non-GPS path has no assertion. Could we add a self-DM duplicate test for a normal **map** distance asserting waypoints are still preserved **and** `distanceRequestType` is the expected value? That locks in the intended behavior and guards the payload change.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -603,13 +603,18 @@ function resolveDuplicates({
...
+
 /**
  * Builds the transactionParams object and computes waypoints used when duplicating a transaction.
  * Shared between duplicateExpenseTransaction and duplicateReport.
  */
 function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, transactionDetails: ReturnType<typeof getTransactionDetails>) {
     const {linkedTrackedExpenseReportAction, ...transactionWithoutLinkedAction} = transaction;
-    const waypoints = !isExpenseSplit(transaction) ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
+    const duplicateAsManualDistance = shouldDuplicateAsManualDistance(transaction);
+    const waypoints = !isExpenseSplit(transaction) && !duplicateAsManualDistance ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
```
marufsharifi: This ternary packs a double negation, a compound condition, and a cast branch into one line — the reader has to hold "not split and not manual-distance" in their head while parsing the cast. Naming the positive intent first reads much cleaner:

```suggestion
    const shouldKeepWaypoints = !isExpenseSplit(transaction) && !duplicateAsManualDistance;
    const waypoints = shouldKeepWaypoints ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
```

marufsharifi (commented):
Looks good and works perfect, but just tow more refinement suggestions

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for flagging this. Since the cast was pre-existing and fixing it properly would require touching the shared `getTransactionDetails` return type, I’ll leave it out of scope for this PR to avoid unrelated changes.

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the suggestion. I updated this to name the waypoint-retention condition first, which makes the GPS-to-manual distance case easier to read. Please review.

x-dev90:
@marufsharifi, during testing, I found a bug in the Workspace section. I’m working on fixing it and will inform you once it’s resolved. Thanks!

x-dev90:
Hi @marufsharifi , while testing an additional deleted-workspace edge case, I found a limitation that requires a backend change.

Steps:

1. Create a workspace using miles.
2. Create a GPS distance expense in the self-DM.
3. Delete the workspace.
4. Duplicate the GPS expense.

The PR correctly converts the GPS expense to manual distance and converts `5.84 mi` to `9.40 km`. However, the deleted workspace rate is unavailable, so the backend applies the current P2P rate. For example, `5.84 mi @ TJS0.76/mi = TJS4.44` becomes `9.40 km @ TJS11.2854/km = TJS106.08`.

The frontend sends the original amount, but the backend recalculates it from the current rate. The existing distance APIs do not accept an explicit historical rate, so preserving the physical distance, equivalent converted rate, and original amount requires backend support. Could you please advise whether this edge case should be handled separately as a backend issue?

marufsharifi:
> Hi @marufsharifi , while testing an additional deleted-workspace edge case, I found a limitation that requires a backend change.
> 
> Steps:
> 
> 1. Create a workspace using miles.
> 2. Create a GPS distance expense in the self-DM.
> 3. Delete the workspace.
> 4. Duplicate the GPS expense.
> 
> The PR correctly converts the GPS expense to manual distance and converts `5.84 mi` to `9.40 km`. However, the deleted workspace rate is unavailable, so the backend applies the current P2P rate. For example, `5.84 mi @ TJS0.76/mi = TJS4.44` becomes `9.40 km @ TJS11.2854/km = TJS106.08`.
> 
> The frontend sends the original amount, but the backend recalculates it from the current rate. The existing distance APIs do not accept an explicit historical rate, so preserving the physical distance, equivalent converted rate, and original amount requires backend support. Could you please advise whether this edge case should be handled separately as a backend issue?

@x-dev90, could you please share a demo. thanks.

x-dev90:
> @x-dev90, could you please share a demo. thanks.

Hi @marufsharifi, I confirmed that the unit conversion issue also exists on `main` in the above flow, so I reverted that commit as out of scope. The original duplication bug remains fixed. Please review.

**Duplicate Manual GPS Distance Demo on the `main` Branch:**

IMG_20260829_121958




**Duplicate GPS distancen Demo on the `main` Branch :**

IMG_20260829_121849

marufsharifi:
@x-dev90, could you please merge the main. thanks.

x-dev90:
> @x-dev90, could you please merge the main. thanks.

I merged main. Thanks!

marufsharifi:
@x-dev90, i am still testing, i will let you know. thanks.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -834,7 +849,11 @@ function duplicateExpenseTransaction({
     const participants = getMoneyRequestParticipantsFromReport(targetReport, currentUserAccountID);
 
     const transactionDetails = getTransactionDetails(transaction);
-    const {transactionParams, waypoints} = buildDuplicateTransactionParams(transaction, transactionDetails);
+    // A duplicate mirrors the source, so an unreported source stays unreported even when a workspace is available
+    const isSourceUnreported = !transaction.reportID || transaction.reportID === CONST.REPORT.UNREPORTED_REPORT_ID;
+    const shouldDuplicateSelfDMExpense = !targetPolicy;
```
marufsharifi: **The GPS→manual gate is narrower than the branch that actually routes to the self-DM path, so the fix won't fire in the real repro.**

`shouldDuplicateSelfDMExpense` is gated on `!targetPolicy`, but the track (self-DM) creation path below is entered on the **broader** `!targetPolicy || isSourceUnreported` (line 902). The moved comment on line 853 even documents the mismatched state: *"an unreported source stays unreported even when a workspace is available"* — i.e. `isSourceUnreported === true` while `targetPolicy` is truthy is real and handled.

Every production caller passes the default expense policy as `targetPolicy`, never `undefined`:
- [`useExpenseActions.ts:275`](https://github.com/Expensify/App/blob/main/src/hooks/useExpenseActions.ts#L275) — the "More ▸ Duplicate" action from the repro: `targetPolicy: defaultExpensePolicy ?? undefined`
- [`useBulkDuplicateAction.ts:88`](https://github.com/Expensify/App/blob/main/src/hooks/useBulkDuplicateAction.ts#L88) — `targetPolicy: (defaultExpensePolicy ?? undefined)`

So for any user who has a default expense policy (anyone in a workspace), duplicating an unreported self-DM GPS expense evaluates to `targetPolicy` truthy → `shouldDuplicateSelfDMExpense = false`, while `isSourceUnreported = true` still routes into the track branch. Result: `duplicateRequestType` stays `DISTANCE_GPS`, `buildDuplicateTransactionParams(..., false)` keeps the waypoints and skips `distance` → **the exact pre-fix broken state issue #91814 reports.** The unit test only passes because it hard-codes `targetPolicy: undefined`, the *only* config where the gate flips to `true`.

Align the gate with the routing branch (both vars are already in scope, in order):

```suggestion
    const shouldDuplicateSelfDMExpense = !targetPolicy || isSourceUnreported;
```

Please also add a test that passes a **real** `targetPolicy` **and** an unreported GPS source (`reportID: CONST.REPORT.UNREPORTED_REPORT_ID`) and still asserts `distanceRequestType: DISTANCE_MANUAL` / `waypoints: undefined` — that single case catches this.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -698,13 +712,14 @@ function createExpenseByType({
 }) {
     switch (transactionType) {
         case CONST.SEARCH.TRANSACTION_TYPE.DISTANCE: {
+            const duplicateRequestType = getDuplicateRequestType(transaction);
```
marufsharifi, line 716: **`getDuplicateRequestType` is called here without `shouldDuplicateSelfDMExpense`, so it defaults to `false` and a GPS source is never mapped on the workspace path.**

The PR body states the request type is used *"consistently for both workspace and self-DM duplicates,"* but this branch — the workspace half — leaves GPS as `DISTANCE_GPS` for both `iouRequestType` (line 715) and `distanceRequestType` (line 742). Combined with the gating issue in `duplicateExpenseTransaction` (line 854), the net effect is that **there is no configuration with a default policy in which a GPS duplicate becomes manual**: if `isSourceUnreported` is false you fall through to this `createExpenseByType` path, which also doesn't convert.

Either thread the flag through so this path converts too, or, if leaving workspace GPS duplicates as `DISTANCE_GPS` is intentional and safe, add a comment stating why — relying on a defaulted argument to silently disable the conversion here is easy to misread.

tests/actions/IOUTest/DuplicateTest.ts
```diff
@@ -1938,6 +1938,85 @@ describe('actions/Duplicate', () => {
...
+                transaction: mockGPSDistanceTransaction,
+                optimisticChatReportID: mockOptimisticChatReportID,
+                optimisticIOUReportID: mockOptimisticIOUReportID,
+                isASAPSubmitBetaEnabled: mockIsASAPSubmitBetaEnabled,
+                introSelected: undefined,
+                quickAction: undefined,
+                policyRecentlyUsedCurrencies: [],
+                isSelfTourViewed: false,
+                customUnitPolicyID: '',
+                targetPolicy: undefined,
```
marufsharifi: **This test passes only because `targetPolicy: undefined` — the one configuration in which the fix activates.**

`mockTransaction.reportID` is `'1'`, so `isSourceUnreported` is `false`; with `targetPolicy: undefined` the gate `shouldDuplicateSelfDMExpense = !targetPolicy` becomes `true`. The realistic production configuration — `targetPolicy = defaultExpensePolicy` with an unreported GPS source — is never exercised, so a green result here does **not** prove the user-facing fix works (see the blocker on `Duplicate.ts:854`).

Please add a sibling case:
- `targetPolicy` set to a real policy **and** `reportID: CONST.REPORT.UNREPORTED_REPORT_ID`
- assert the `TRACK_EXPENSE` call still gets `distanceRequestType: DISTANCE_MANUAL`, `waypoints: undefined`, `distance: <quantity>`

That case fails on the current gate and passes once it's widened to `!targetPolicy || isSourceUnreported`.

marufsharifi (changes requested):

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for catching this. I aligned the GPS-to-manual condition with the self-DM tracking path and added coverage for an unreported GPS source with a target policy. Please review.

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the feedback. Keeping workspace GPS duplicates unchanged is intentional to limit the fix to self-DM expenses and avoid regressions. I’ll clarify this in the PR description. Please review.

tests/actions/IOUTest/DuplicateTest.ts
x-dev90: Thanks for catching this. I aligned the condition with the self-DM tracking path and added coverage for an unreported GPS expense with a target policy. Please review.

x-dev90:
@marufsharifi ,I’ve addressed all your feedback. Could you please review the latest changes? thanks!

marufsharifi:
@x-dev90, could you please merge the main. thanks.

x-dev90:
> @x-dev90, could you please merge the main. thanks.

@marufsharifi, Done.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -908,6 +925,7 @@ function duplicateExpenseTransaction({
             transactionParams: {
                 ...(params.transactionParams ?? {}),
                 validWaypoints: waypoints,
+                ...(isDistanceRequest(transaction) && {distanceRequestType: duplicateRequestType}),
```
marufsharifi: **This guard is `isDistanceRequest`, not `isGPSDistanceRequest` — it widens the change beyond the stated fix.**

Before this PR the self-DM (`trackExpense`) path sent **no** `distanceRequestType` at all — it was only sent on the workspace path via `createDistanceRequest`. With `isDistanceRequest`, **every** distance type (map / manual / odometer) duplicated into a self-DM now sends a new `distanceRequestType` to `TRACK_EXPENSE`.

For non-GPS distances `duplicateRequestType === getRequestType(transaction)`, so the optimistic transaction is unchanged (its `iouRequestType` comes from `existingTransaction`), and `trackExpense` only forwards `distanceRequestType` as an API param (`TrackExpense.ts:2884`). So local behavior is safe — but the **server payload changes** on flows with **zero test coverage**. If the backend branches on the now-present param for a waypoint-carrying map duplicate, "duplicate my map mileage into self-DM" could regress silently.

Two acceptable fixes:
- Keep the broad guard (consistent with the workspace path) **and** add the map-distance self-DM test I flag in `DuplicateTest.ts` below, or
- Narrow it to the only type whose type actually differs from the source:

```suggestion
                ...(isGPSDistanceRequest(transaction) && {distanceRequestType: duplicateRequestType}),
```

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -641,7 +647,7 @@ function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, tra
         unit: transaction.comment?.units?.unit,
     };
 
-    if (isDistanceRequest(transaction) && (isExpenseSplit(transaction) || isOdometerDistanceRequest(transaction))) {
+    if (isDistanceRequest(transaction) && (isExpenseSplit(transaction) || isOdometerDistanceRequest(transaction) || duplicateAsManualDistance)) {
```
marufsharifi: **GPS→manual with a missing `customUnit.quantity` yields a distance-less manual expense.**

When `duplicateAsManualDistance` is true and `quantity` is absent, `transactionParams.distance` resolves to `undefined`, producing a manual distance expense with **no distance**. In practice a GPS transaction always carries a measured `quantity`, so risk is low — but there's no guard and no test for the empty case. Worth a defensive check or at least a code comment noting that quantity is assumed present.

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -603,13 +603,19 @@ function resolveDuplicates({
...
  * Builds the transactionParams object and computes waypoints used when duplicating a transaction.
  * Shared between duplicateExpenseTransaction and duplicateReport.
  */
-function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, transactionDetails: ReturnType<typeof getTransactionDetails>) {
+function buildDuplicateTransactionParams(transaction: OnyxTypes.Transaction, transactionDetails: ReturnType<typeof getTransactionDetails>, shouldDuplicateSelfDMExpense = false) {
     const {linkedTrackedExpenseReportAction, ...transactionWithoutLinkedAction} = transaction;
-    const waypoints = !isExpenseSplit(transaction) ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
+    const duplicateAsManualDistance = shouldDuplicateAsManualDistance(transaction, shouldDuplicateSelfDMExpense);
+    const shouldKeepWaypoints = !isExpenseSplit(transaction) && !duplicateAsManualDistance;
+    const waypoints = shouldKeepWaypoints ? (transactionDetails?.waypoints as WaypointCollection | undefined) : undefined;
```
marufsharifi: **NAB: Boolean variable should be prefixed with `is`/`should`** 

Per `STYLE.md`, boolean vars must be prefixed `is` or `should`. `duplicateAsManualDistance` holds the boolean result of `shouldDuplicateAsManualDistance(...)` but drops the prefix — note `shouldKeepWaypoints` right below does it correctly.

```suggestion
    const isManualDistanceDuplicate = shouldDuplicateAsManualDistance(transaction, shouldDuplicateSelfDMExpense);
    const shouldKeepWaypoints = !isExpenseSplit(transaction) && !isManualDistanceDuplicate;
```

src/libs/actions/IOU/Duplicate.ts
```diff
@@ -603,13 +603,19 @@ function resolveDuplicates({
     API.write(WRITE_COMMANDS.RESOLVE_DUPLICATES, parameters, {optimisticData, failureData});
 }
 
+function shouldDuplicateAsManualDistance(transaction: OnyxTypes.Transaction, shouldDuplicateSelfDMExpense: boolean) {
```
marufsharifi, line 606: **Uncovered negative branch: reported GPS distance must *not* convert to manual**

This is the `shouldDuplicateAsManualDistance` false branch when `shouldDuplicateSelfDMExpense=false` — the scope-guard for the whole feature. Every test that feeds a `DISTANCE_GPS` transaction sets `shouldDuplicateSelfDMExpense=true` (no policy, or unreported), so nothing exercises a GPS distance that is **reported *and* has a `targetPolicy`**. That path must go through `createExpenseByType` (`createDistanceRequest`), keep its waypoints, and stay a GPS/distance request rather than becoming `DISTANCE_MANUAL`.

Without this, a regression that made the guard fire for workspace expenses (e.g. dropping the `&&` or weakening the self-DM condition) would pass silently.

**Suggested test** — add a new `it` next to the GPS `it.each` at `tests/actions/IOUTest/DuplicateTest.ts:1941`:
- `DISTANCE_GPS` transaction with `reportID` set (reported), `targetPolicy: mockPolicy`, `targetReport: policyExpenseChat`
- Assert the `CREATE_DISTANCE_REQUEST` call is used (not `TRACK_EXPENSE`)
- Assert its `waypoints` are defined/non-null
- Assert the duplicated transaction's `iouRequestType`/`distanceRequestType` is **not** `DISTANCE_MANUAL`

marufsharifi (changes requested):

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the feedback. I narrowed the self-DM distanceRequestType payload change to GPS duplicates only, so map/manual/odometer duplicates remain unchanged. Please review.

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the feedback. I added a comment clarifying that completed GPS expenses use the saved customUnit.quantity when duplicated as manual because the raw GPS track is unavailable. Please review.

src/libs/actions/IOU/Duplicate.ts
x-dev90: addressed. thanks!

src/libs/actions/IOU/Duplicate.ts
x-dev90: Thanks for the feedback. I added coverage to ensure reported workspace GPS duplicates continue using `CREATE_DISTANCE_REQUEST`, keep their waypoints, and remain `DISTANCE_GPS`. Please review.

x-dev90:
Hi @marufsharifi , I’ve addressed all feedback on the PR. Could you please review the latest changes? thanks

marufsharifi (approved):
LGTM

blimpich (approved):
Nice job!

Issue #87030, opened 2026-04-02
Melvin ability to run web and mobile device testing


Julesssss:
### Problem

We are using playwright within Melvin for simple bug verification and test fixes (limited to web).
Our plans to extend Melvin include improved testing and expanding testing functionality to mobile.

`agent-device` [looks](https://x.com/thymikee/status/2039771602217242840) like a great potential tool for this. The demo shows fast UI testing of the Expensify ND app. 

### Solution

Lets create a POC using the tool. Initially we would like to enable Melvin to:

1. Verify test cases similar to playwright
2. Post screenshots to App PRs (building into Neil's work)

@AndrewGable @justinpersaud
_____

- [Here's](https://github.com/Expensify/App/blob/main/.claude/skills/playwright-app-testing/SKILL.md) our existing Playwright skill
- Here's [documentation ](https://github.com/Expensify/Expensify/tree/main/.claude)on Melvin
- [Here's](https://github.com/Expensify/Expensify/issues/601721) how Melvin uploads images and attatches them to Github comments
- [Here's](https://github.com/Expensify/melvin/pull/234#event-27386275486) a collection of test runs

**Initial Questions:**
- Limitations? 
- How to integrate with Melvin, which is designed for one-shot requests

kacper-mikolajczak:
Hey folks 👋 Here's a design doc on the PoC after our internal discussion with @Julesssss:

# PoC Design: Agentic Mobile Testing for Melvin

## Goal

Improve PR QA/review process by enabling agentic mobile testing via [agent-device](https://github.com/callstackincubator/agent-device), integrated with the existing MelvinBot App workflow. Additionally, enable agent-device for local developer use.

## Background

MelvinBot currently handles ~200 requests per 36 hours on Expensify/App - automated issue analysis, CI fixes, code review, implementation, and more. The workflow (`claude-melvinbot-app.yml`) includes Playwright browser testing infrastructure, but analysis of ~650 production runs shows it's invoked only ~0.3% of the time. Mobile testing is entirely absent.

[agent-device](https://github.com/callstackincubator/agent-device) is a CLI tool by Callstack that gives AI agents control over mobile app UIs using accessibility tree snapshots (not screenshots), making interactions token-efficient and deterministic. It supports Android emulators headlessly on Linux - a natural fit for the existing CI setup.

## Architecture

```
                        MelvinBot App Workflow
                     (claude-melvinbot-app.yml)
                               |
            +------------------+------------------+
            |                                     |
[...]

PR #98272, opened 2026-08-11
[NO QA] Scope agent-device text-input macros per platform

kacper-mikolajczak:
### Explanation of Change

The agent-device sign-in flow does not resolve on web. `flows/macros/sign-in.ad` declares `@pre role="textfield" label="Phone or email"`, and every arm of its `fill` union is the same native shape, so on web the flow cannot even satisfy its own precondition. `send-message.ad` and `complete-onboarding.ad` target text inputs the same way and fail for the same reason.

Resolved against a live web session, by the exit code of `agent-device is visible "<selector>"`:

| Selector | Web |
| --- | --- |
| `role="textbox"` | resolves for text inputs |
| `role="textfield"` | fails |
| `role="text-field"` | fails |
| `editable=true` | fails on inputs that are genuinely editable - the same node matches `editable=false` |
| `id="username"` | not exposed on the login input |
| `label="Phone or email"` (bare) | resolves |
| `role="button" label="Continue"` | resolves |
| `role="textbox" label="Write something..."` | resolves |

Worth calling out because it is how the wrong selectors got written in the first place: `snapshot -i` prints the login input as `[text-field]` and `[editable]`. Those are display tags, not selector values. A selector has to be confirmed by exit code, never read off a snapshot.

This PR adds `flows/macros/web/` and `flows/macros/android/`. A caller driving platform `P` uses `flows/macros/<P>/<name>.ad` when it exists and `flows/macros/<name>.ad` otherwise. Only the three flows whose selectors actually diverge are split, and all three of them fill text inputs. `go-back.ad` is platform-neutral and stays shared.

- The `web/` variants use the confirmed web selectors, in the `@pre` and `@post` headers as well as in the executable steps.
- The `android/` variants are byte-identical to what `flows/macros/` has today. No android arm has been probed, so nothing about android behaviour changes.
- The unscoped copies stay in `flows/macros/` as the fallback for platforms with no folder yet, which today means iOS. That is deliberately today's unprobed behaviour rather than a change to it - if a reviewer would rather see iOS probed and the unscoped copies deleted, say so and I will do that instead of leaving the duplicate.

Not touched here: `flows/lib/sign-in-drive.sh` carries the same selector strings in its own `SEL_*` constants.

### The general problem this exposes

A `||` union conflates two different meanings: "fallback if this shape is absent" and "the shape this platform uses". They are not the same thing, and merging them makes selector rot invisible.

An arm that is dead on one platform stays in the file, because a different arm carries the union on another platform. Nothing reports which arm matched, so a union that is really one arm wide on each platform looks healthy on both. When the last live arm on a platform finally breaks, the failure reads as "the flow does not work", with no signal that the other arms had been dead for a long time.

`sign-in.ad` is the end state of that. Its three-arm `fill` union looks like portability insurance, and all three arms are the same native shape - so the union is exactly what hid the fact that the flow was never portable.

Platform folders make the choice explicit per platform. A `||` union should be a real fallback within one platform, not a cross-platform compatibility layer.

Every remaining `||` arm in the flows this PR touches is enumerated in a comment below, with a keep-or-drop proposal for each. Android arms are marked unprobed rather than guessed at, and nothing untested has been deleted - each one needs a reviewer's call.

### Fixed Issues

$ https://github.com/Expensify/App/issues/87030

### Tests

N/A - no product code changes. The web selectors were each resolved against a live web session and checked by exit code.

### QA Steps

N/A - no product code changes.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/.claude/skills/agent-device/flows/README.md
+++ b/.claude/skills/agent-device/flows/README.md
@@ -3,17 +3,29 @@
 ## Directory layout
 
 - `macros/` - reusable helpers for common setup/navigation actions that stop in a navigable state for further interactive work.
+- `macros/<platform>/` - platform-specific overrides of a `macros/` flow, for flows whose selectors differ per platform. See [Platform scoping](#platform-scoping).
 - `tests/` - critical-scenario scripts for QA/perf verification that assert explicit outcomes (for example Sentry spans) and then stop.
 - `lib/` - bash drive libraries for flows that need conditional steering the linear `.ad` format cannot express (snapshot classification, state-dependent branching). Each file documents its own contract; the caller always owns the session lifecycle (`open`/`close`/`record`). Source them from an orchestrator or run them standalone against an already-open session (for example `lib/sign-in-drive.sh --platform web --session <name> --email <email>`).
 
 Composable `.ad` snippets - bounded units of work. A flow may span one or multiple screens as long as it represents a coherent, reusable action with clear start (`@pre`) and completion (`@post`) checkpoints. Each flow advertises machine-matchable metadata (`@pre`, `@post`, `@tag`, `@param`) via `# @`-prefixed comment headers, while flow type is derived from location (`flows/macros/` or `flows/tests/`).
+
+## Platform scoping
+
+Most flows are platform-neutral and live directly in `macros/`. A flow whose selectors genuinely differ per platform gets a copy per platform under `macros/<platform>/`, where `<platform>` is the value passed to `--platform`.
+
+A caller driving platform `P` resolves a macro by name:
+
+1. `macros/<P>/<name>.ad` when that file exists.
+2. `macros/<name>.ad` otherwise.
+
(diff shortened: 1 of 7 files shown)


Julesssss (approved):
Nice to see the macros coming together 🎉 

Is there a reason we need per-platform macros?

rlinoz (approved):
Love this

Julesssss:
> Hi @kacper-mikolajczak. Any QA steps here?
> 
> cc @rlinoz @Julesssss @mjasikowski

No testing required here @IuliiaHerets, thank you

Issue #98048, opened 2026-08-07
[Due for payment 2026-08-25] Search API re-fires with an identical query hash every 2-8 seconds — 26% of all Search traffic is duplicate


mountiny:
## Summary

Clients are re-sending the **exact same `Search` API request** — identical query hash, identical `jsonQuery`, identical `offset` — every 2 to 8 seconds, continuously, for minutes at a time, with no user interaction driving it.

This is not rare. In one sampled hour, **26% of all `Search` API calls** were a user re-sending a query they had already sent more than 10 times in that same hour. It affects web, iOS and Android, and it costs roughly **28 minutes of Auth compute per hour**.

I have confirmed the behaviour in the logs and measured its scale. I have **not** identified which client-side code path re-fires the request, so this issue is deliberately evidence-only. Details below, with reproducible queries.

## Evidence: a single looping client

One client, 2026-08-07 `12:40`–`12:50` UTC. **67 calls in 10 minutes**, continuous, with no gap longer than 17 seconds anywhere in the window. Every request was byte-identical.

| Field | Value | Varies between calls? |
| --- | --- | --- |
| `command` | `Search` | no |
| `hash` | `959171759` | no |
| `inputQuery` | `type:expense sortBy:date sortOrder:desc` | no |
| `searchKey` | `expenses` | no |
| `offset` | `50` | no |
| `shouldCalculateTotals` | `false` | no |
| `filters` | `null` | no |
| platform / appversion | `web` / `9.4.50-3` | no |
| source IP, browser fingerprint | identical throughout | no |
| `clientUpdateID` | `15181637929`, `15181653964`, `15181656766`, `15181657747`, `15181659336`, … | **yes, increases every call** |

Intervals between the first 15 calls, in seconds:

```
6.3, 1.5, 3.8, 3.9, 3.2, 2.3, 1.5, 6.6, 5.6, 8.4, 5.9, 1.4, 3.9, 2.2
```

Two observations worth recording, without claiming they explain the loop:

- The query is at `offset: 50`, so this is a paginated page, not an initial page load.
- `clientUpdateID` advances on every single call, meaning the client is receiving and applying Onyx updates between iterations.

Sample request IDs, for anyone who wants to pull the full trace:

```
a27658f9c98aa497-LAX   12:40:02
a2765920d870a497-DTW   12:40:09
a2765929f81ea497-LAX   12:40:10
a276594169c0a497-LAX   12:40:14
a276595a099da497-DTW   12:40:18
[...]

PR #98305, opened 2026-08-11
fix: stop Search re-firing an identical query on unrelated Onyx updates

BartekObudzinski:
### Explanation of Change

Search sends a byte-identical request every 2 to 8 seconds with no user interaction. Issue #98048 measured this at roughly 26 percent of all Search calls in a sampled hour, across web, iOS, and Android, costing about 28 minutes of Auth compute per hour. It is not a recent regression. The repeat share has stayed flat at 24 to 28 percent across the whole retained log window.

Two defects in the same effect in `src/hooks/useSearchHighlightAndScroll.ts` cause this.

First, the skip guard could never fire on a non-chat search. `newIDs` was built from `Object.keys(transactions)`, which are Onyx collection keys shaped `transactions_<id>`, and compared against the bare `transactionID` values that `extractTransactionIDsFromSearchResults` returns. Those two sets can never intersect, so `hasAGenuinelyNewID` was always true, the early return meant to skip the refetch was unreachable, and the `hasDeletedID` branch below it was dead code.

Second, report actions triggered refetches on searches that have nothing to do with report actions. The condition was `(!isChat && hasTransactionsIDsChange) || hasReportActionsIDsChange`, so any new report action anywhere in the account, an incoming chat message for example, re-fired an expense search. The comment right above the effect already describes the intended behavior: trigger a search when a new report action is added while on chat, or when a new transaction is added for the other search types. The code did not match its own comment.

The second defect pumps the loop, and the first defect is the safety valve that should have caught it but could not. Together they produce exactly the reported signature: an identical query hash, no user interaction, and a cadence that tracks request latency, because the internal `searchTriggeredRef` latch is released once per response by the reset effect right below.

The fix is two lines. Gate the report action trigger behind `isChat` so it matches the documented intent, and strip the collection key prefix before comparing against the search result IDs.

This bug survived because the existing unit tests build the `transactions` fixture with bare keys, for example `{'1': {transactionID: '1'}}`, while production passes `useOnyx(ONYXKEYS.COLLECTION.TRANSACTION)`, whose keys carry the `transactions_` prefix. Every existing test agreed with itself and never exercised the mismatch.

Three new tests were written before the fix, and each one was watched failing, all showing a single spurious call with the identical query hash. They cover a report action arriving on a non-chat search with no transaction change, an added transaction that is already present in the results, and a report action arriving while Onyx holds a transaction the query filters out. That third test still failed after gating the report action trigger alone, which is what proved the ID prefix fix was needed too.

**Note for reviewers:** a non-chat search no longer refetches on a report action that changes no transaction, for example approving a report from an RHP while on the approve todo view. The flows that need that refresh already call `refreshSearchAfterReportAction` explicitly right after the action, at `src/hooks/useLifecycleActions.tsx:252` and `src/hooks/useSelectionModePayment.ts:235`, which is what `src/libs/SearchRefreshUtils.ts` exists for. This change also makes the `hasDeletedID` branch reachable for the first time on non-chat searches, so deletion from search results is worth explicit QA attention.

### Fixed Issues

$ https://github.com/Expensify/App/issues/98048
PROPOSAL:

### Tests

**Preconditions**

1. Sign in to two accounts, Account 1 and Account 2, on two devices or browsers.
2. On Device 1, open DevTools > Network and filter the requests to `Search`.

**The fix**

1. Account 1: open Spend > Expenses.
2. Account 1: clear the network log in DevTools.
3. Account 2: open the DM with Account 1 and send a message.
4. Account 1: verify that no `Search` request appears in DevTools.

**Regression checks**

5. Account 1: create an expense from the right hand panel of the Expenses page, and verify that it appears in the list with its row highlighted.
6. Account 1: delete an expense from the list, and verify that it disappears from the list.
7. Account 1: use the search router to run a keyword search and open the Chats results. Have Account 2 send a message to that chat, and verify that the chat results refresh.
8. Account 1: open Needs approval, approve a report, and verify that the list refreshes.
9. Account 1: enter selection mode on Ready to pay, pay a report, and verify that the list refreshes.

### Offline tests

N/A

### QA Steps

Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/ca55865a-f048-4e81-b39c-8e010d38e80c

</details>

### DIFF
--- a/src/hooks/useSearchHighlightAndScroll.ts
+++ b/src/hooks/useSearchHighlightAndScroll.ts
@@ -114,25 +114,36 @@ function useSearchHighlightAndScroll({
             }
             hasPendingSearchRef.current = false;
 
-            // Read the IDs off the transactions themselves: `transactionsIDs` are Onyx collection keys
-            // (`transactions_<id>`) while the search results yield bare IDs, so the two never compare equal.
-            const newIDs = isChat
-                ? reportActionsIDs
-                : Object.values(transactions ?? {})
-                      .map((transaction) => transaction?.transactionID)
-                      .filter((id): id is string => !!id);
+            // Read the IDs off the transactions themselves: their Onyx keys are `transactions_<id>` while the
+            // search results yield bare IDs, so the two never compare equal.
+            // `addedIDs` drives the new-item check and `currentIDs` the deletion check. They cannot be the same
+            // list: the collection holds every transaction cached on the account, so on a filtered or paginated
+            // query it always contains an ID the results omit, which would make every pass look like a new item.
+            const addedIDs: string[] = [];
+            const currentIDs: string[] = [];
+            for (const [key, transaction] of Object.entries(transactions ?? {})) {
+                const transactionID = transaction?.transactionID;
+                if (!transactionID) {
+                    continue;
+                }
+                currentIDs.push(transactionID);
+                if (!previousTransactionsIDsSet.has(key)) {
+                    addedIDs.push(transactionID);
+                }
+            }
+
             let currentSearchResultIDs: string[] = [];
             if (searchResultsData) {
                 currentSearchResultIDs = isChat ? extractReportActionIDsFromSearchResults(searchResultsData) : extractTransactionIDsFromSearchResults(searchResultsData);
             }
             const existingSearchResultIDsSet = new Set(currentSearchResultIDs);
-            const hasAGenuinelyNewID = newIDs.some((id) => !existingSearchResultIDsSet.has(id));
+            const hasAGenuinelyNewID = (isChat ? reportActionsIDs : addedIDs).some((id) => !existingSearchResultIDsSet.has(id));
 
             // Only skip search if there are no new items AND search results aren't empty
             // This ensures deletions that result in empty data still trigger search
             if (!hasAGenuinelyNewID && currentSearchResultIDs.length > 0) {
-                const newIDsSet = new Set(newIDs);
-                const hasDeletedID = currentSearchResultIDs.some((id) => !newIDsSet.has(id));
+                const currentIDsSet = new Set(isChat ? reportActionsIDs : currentIDs);
+                const hasDeletedID = currentSearchResultIDs.some((id) => !currentIDsSet.has(id));
                 if (!hasDeletedID) {
                     return;
                 }
--- a/src/hooks/useSearchHighlightAndScroll.ts
+++ b/src/hooks/useSearchHighlightAndScroll.ts
@@ -114,11 +114,7 @@ function useSearchHighlightAndScroll({
             }
             hasPendingSearchRef.current = false;
 
-            // Read the IDs off the transactions themselves: their Onyx keys are `transactions_<id>` while the
-            // search results yield bare IDs, so the two never compare equal.
-            // `addedIDs` drives the new-item check and `currentIDs` the deletion check. They cannot be the same
-            // list: the collection holds every transaction cached on the account, so on a filtered or paginated
-            // query it always contains an ID the results omit, which would make every pass look like a new item.
+            // Transaction Onyx keys are `transactions_<id>` but search results yield bare IDs, so read the ID off the value.
             const addedIDs: string[] = [];
             const currentIDs: string[] = [];
             for (const [key, transaction] of Object.entries(transactions ?? {})) {
--- a/src/hooks/useSearchHighlightAndScroll.ts
+++ b/src/hooks/useSearchHighlightAndScroll.ts
@@ -115,16 +115,16 @@ function useSearchHighlightAndScroll({
             hasPendingSearchRef.current = false;
 
             // Transaction Onyx keys are `transactions_<id>` but search results yield bare IDs, so read the ID off the value.
-            const addedIDs: string[] = [];
-            const currentIDs: string[] = [];
+            const addedTransactionIDs: string[] = [];
+            const currentTransactionIDs: string[] = [];
             for (const [key, transaction] of Object.entries(transactions ?? {})) {
                 const transactionID = transaction?.transactionID;
                 if (!transactionID) {
                     continue;
                 }
-                currentIDs.push(transactionID);
+                currentTransactionIDs.push(transactionID);
                 if (!previousTransactionsIDsSet.has(key)) {
-                    addedIDs.push(transactionID);
+                    addedTransactionIDs.push(transactionID);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


tests/unit/useSearchHighlightAndScrollTest.ts
```diff
@@ -186,7 +186,134 @@ describe('useSearchHighlightAndScroll', () => {
...
             ...baseProps,
             transactions: {
-                '1': {transactionID: '1'},
+                transactions_1: {transactionID: '1'},
+            },
+        };
+
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-expect-error
+        rerender(updatedProps);
```
QichenZhu: NAB: should we make it a file-level suppression and add a comment, since it's used throughout the file?

src/hooks/useSearchHighlightAndScroll.ts
```diff
@@ -116,7 +114,13 @@ function useSearchHighlightAndScroll({
...
             hasPendingSearchRef.current = false;
 
-            const newIDs = isChat ? reportActionsIDs : transactionsIDs;
+            // Read the IDs off the transactions themselves: `transactionsIDs` are Onyx collection keys
+            // (`transactions_<id>`) while the search results yield bare IDs, so the two never compare equal.
+            const newIDs = isChat
+                ? reportActionsIDs
+                : Object.values(transactions ?? {})
+                      .map((transaction) => transaction?.transactionID)
+                      .filter((id): id is string => !!id);
```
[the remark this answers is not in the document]
BartekObudzinski: Fixed. The collection holds every transaction cached on the account, so on a filtered or paginated query it always contained an ID the results omit, and the check never went false.

One caveat on the suggestion. `newIDs` was also feeding the deletion check right below it, so building it only from added IDs made nearly every result ID look deleted and the search fired on every pass anyway. I split it into two lists instead, added IDs for the new item check and the full current set for the deletion check. Test added for it.

QichenZhu (commented):
The test steps are confusing:

> 1. Sign in on web and open Search with the query `type:expense sortBy:date sortOrder:desc`.

The Search page input only supports keywords and doesn't support these query parameters.

> 3. Leave the page idle for 90 seconds without interacting with it.

This step doesn't trigger the bug.

> 5. In a second tab of the same session, open any chat and send three messages.

Native apps don't have tabs.

**Suggested changes:**

Preconditions:

1. Sign in to two accounts on two devices or browsers
2. Open DevTools > Network on Device 1

Steps:

1. Account 1: Open Spend > Expenses
2. Account 1: Clear the network log in DevTools
3. Account 2: Open the DM with Account 1 and send a message
4. Account 1: Verify that no `Search` requests appear in DevTools

Please update them and the recording videos accordingly. Thanks!

src/hooks/useSearchHighlightAndScroll.ts
```diff
@@ -116,19 +114,36 @@ function useSearchHighlightAndScroll({
             }
             hasPendingSearchRef.current = false;
 
-            const newIDs = isChat ? reportActionsIDs : transactionsIDs;
+            // Read the IDs off the transactions themselves: their Onyx keys are `transactions_<id>` while the
+            // search results yield bare IDs, so the two never compare equal.
+            // `addedIDs` drives the new-item check and `currentIDs` the deletion check. They cannot be the same
+            // list: the collection holds every transaction cached on the account, so on a filtered or paginated
+            // query it always contains an ID the results omit, which would make every pass look like a new item.
```
QichenZhu: Is this comment describing the problem or the solution?

QichenZhu: Oh, I see. It describes the problem with the previous code, but the reader of this comment will only see the new code and might be confused. Let's remove it.

BartekObudzinski: Right, that one was about the old code. Removed.

tests/unit/useSearchHighlightAndScrollTest.ts
BartekObudzinski: Good call. Moved it to a file-level disable with a comment explaining why the props need `@ts-expect-error`, and dropped the 22 inline ones.

QichenZhu (approved):

src/hooks/useSearchHighlightAndScroll.ts
```diff
@@ -116,19 +114,32 @@ function useSearchHighlightAndScroll({
             }
             hasPendingSearchRef.current = false;
 
-            const newIDs = isChat ? reportActionsIDs : transactionsIDs;
+            // Transaction Onyx keys are `transactions_<id>` but search results yield bare IDs, so read the ID off the value.
+            const addedIDs: string[] = [];
+            const currentIDs: string[] = [];
```
mountiny:
```suggestion
            const addedTransactionIDs: string[] = [];
            const currentTransactionIDs: string[] = [];
```

src/hooks/useSearchHighlightAndScroll.ts
```diff
@@ -116,19 +114,32 @@ function useSearchHighlightAndScroll({
...
+                }
+            }
+
             let currentSearchResultIDs: string[] = [];
             if (searchResultsData) {
                 currentSearchResultIDs = isChat ? extractReportActionIDsFromSearchResults(searchResultsData) : extractTransactionIDsFromSearchResults(searchResultsData);
             }
             const existingSearchResultIDsSet = new Set(currentSearchResultIDs);
-            const hasAGenuinelyNewID = newIDs.some((id) => !existingSearchResultIDsSet.has(id));
+            const hasAGenuinelyNewID = (isChat ? reportActionsIDs : addedIDs).some((id) => !existingSearchResultIDsSet.has(id));
```
mountiny:
```suggestion
            const hasAGenuinelyNewID = (isChat ? reportActionsIDs : addedTransactionIDs).some((id) => !existingSearchResultIDsSet.has(id));
```

src/hooks/useSearchHighlightAndScroll.ts
```diff
@@ -116,19 +114,32 @@ function useSearchHighlightAndScroll({
...
 
             // Only skip search if there are no new items AND search results aren't empty
             // This ensures deletions that result in empty data still trigger search
             if (!hasAGenuinelyNewID && currentSearchResultIDs.length > 0) {
-                const newIDsSet = new Set(newIDs);
-                const hasDeletedID = currentSearchResultIDs.some((id) => !newIDsSet.has(id));
+                const currentIDsSet = new Set(isChat ? reportActionsIDs : currentIDs);
```
mountiny:
```suggestion
                const currentIDsSet = new Set(isChat ? reportActionsIDs : currentTransactionIDs);
```

src/hooks/useSearchHighlightAndScroll.ts
```diff
@@ -116,19 +114,32 @@ function useSearchHighlightAndScroll({
...
+            const addedIDs: string[] = [];
+            const currentIDs: string[] = [];
+            for (const [key, transaction] of Object.entries(transactions ?? {})) {
+                const transactionID = transaction?.transactionID;
+                if (!transactionID) {
+                    continue;
+                }
+                currentIDs.push(transactionID);
+                if (!previousTransactionsIDsSet.has(key)) {
+                    addedIDs.push(transactionID);
```
mountiny:
```suggestion
                currentTransactionIDs.push(transactionID);
                if (!previousTransactionsIDsSet.has(key)) {
                    addedTransactionIDs.push(transactionID);
```

mountiny (changes requested):

src/hooks/useSearchHighlightAndScroll.ts
BartekObudzinski: Renamed, thanks.

src/hooks/useSearchHighlightAndScroll.ts
BartekObudzinski: Renamed, thanks.

src/hooks/useSearchHighlightAndScroll.ts
BartekObudzinski: Renamed, thanks.

src/hooks/useSearchHighlightAndScroll.ts
BartekObudzinski: Renamed, thanks.

mountiny (approved):

PR #98405, opened 2026-08-12
Fix new manual expense flow regressions (Part 3)

thelullabyy:
### Fixed Issues

$ https://github.com/Expensify/App/issues/98230
$ https://github.com/Expensify/App/issues/98317 
$ https://github.com/Expensify/App/issues/98240
PROPOSAL:

### Tests

Verify that all of these issues are not reproducible
$ https://github.com/Expensify/App/issues/98230
$ https://github.com/Expensify/App/issues/98317 
$ https://github.com/Expensify/App/issues/98240

### QA Steps

Verify that all of these issues are not reproducible
$ https://github.com/Expensify/App/issues/98230
$ https://github.com/Expensify/App/issues/98317 
$ https://github.com/Expensify/App/issues/98240

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/74ced781-5ecd-4d3f-a41e-df91ca94692e

https://github.com/user-attachments/assets/7d4e5391-d368-42a9-8e70-a15de4bd4d7d

https://github.com/user-attachments/assets/76d4d63e-2e0b-40e4-9d55-cb740fe74649

</details>


src/libs/Navigation/OnyxTabNavigator.tsx
```diff
@@ -225,6 +230,15 @@ function OnyxTabNavigator<TTabName extends string = SelectedTabRequest>({
     const persistSelectedTab = Tab.setSelectedTab as (tabID: string, tabName: string) => void;
     const notifyTabSelected = onTabSelected as (newTabName: string | undefined) => void;
 
+    useEffect(() => {
```
thelullabyy: For this issue https://github.com/Expensify/App/issues/98240,

On the first Manual press:

1. tabPress fires and is not prevented (no discard guard involved).
2. The tab navigator's own state moves to index 0 → manual.
3. The parent never takes that change — the URL stays /create/track/start/1/0/scan.
4. The navigator is then re-hydrated from the parent's pre-press snapshot and snaps back to scan, with no dispatch, no remount, and no Navigation.* call.

The second press works because by then the parent's snapshot has caught up

Solution: We record what each press asked for (from → to), and if the navigator's state later drifts back to the tab we came from

joekaufmanexpensify (dismissed):
Fixing deploy blockers.

src/libs/Navigation/OnyxTabNavigator.tsx
```diff
@@ -121,6 +121,11 @@ function OnyxTabNavigator<TTabName extends string = SelectedTabRequest>({
     const [focusTrapContainerElementMapping, setFocusTrapContainerElementMapping] = useState<Record<string, HTMLElement>>({});
     const [selectedTab, selectedTabResult] = useOnyx(`${ONYXKEYS.COLLECTION.SELECTED_TAB}${id}`);
 
+    const pressedTabRef = useRef<{from: string | undefined; to: string} | undefined>(undefined);
```
ikevin127: ### 🟠 **`src/libs/Navigation/OnyxTabNavigator.tsx:233-240` and `:303`: `pressedTabRef` is never cleared once a press succeeds, so it stays armed for the whole life of the navigator and can hijack a later legitimate tab jump**

```tsx
pressedTabRef.current = pressedTabName && pressedTabName !== currentTabName ? {from: currentTabName, to: pressedTabName} : undefined;
```

The ref is only ever overwritten by the *next* `tabPress`. In the happy path (press lands and sticks) the entry `{from: 'manual', to: 'scan'}` survives with `hasReappliedPressRef.current === false`. Any later state event where the active tab equals `from` fires the re-apply, and it does not have to come from the tab bar: `Navigation.navigate` on a route that lives inside a tab navigator dispatches `TabActions.jumpTo` directly (`src/libs/Navigation/Navigation.ts:503`), which produces a `state` event with no `tabPress`.

**Bug this causes**: user is on **Manual**, taps **Scan** (armed), shoots a receipt, drills into a sub-step, and the flow later navigates back to the manual create **URL**. `setActiveTabName('manual')` matches `pressedTab.from`, the effect force jumps to Scan, and the user lands on the camera instead of the manual form. That is the same class of bug as #98240, just inverted.

Suggested fix, disarm as soon as the press is no longer in flight:

```tsx
useEffect(() => {
    const pressedTab = pressedTabRef.current;
    if (!pressedTab || !activeTabName) {
        return;
    }
    if (hasReappliedPressRef.current) {
        // The re-apply already ran (or was consumed); stop watching so a later intentional jump isn't undone.
        pressedTabRef.current = undefined;
        return;
    }
    if (activeTabName !== pressedTab.from) {
        return;
    }
    hasReappliedPressRef.current = true;
    tabNavigationRef.current?.dispatch(TabActions.jumpTo(pressedTab.to));
}, [activeTabName]);
```

That still covers the `to` then bounce back to `from` sequence the fix targets, but the arm dies on the next settled state event instead of living forever.

If that ordering turns out to be too tight in practice, the alternative is a time bound (drop the arm after `CONST.ANIMATED_TRANSITION`), but please do not leave it unbounded.

src/libs/Navigation/OnyxTabNavigator.tsx
```diff
@@ -276,6 +292,15 @@ function OnyxTabNavigator<TTabName extends string = SelectedTabRequest>({
                                     return;
                                 }
                                 handleTabPress(navigation, e);
+                                if (e.defaultPrevented) {
+                                    return;
+                                }
```
ikevin127, line 332: ### 🟡 **`src/libs/Navigation/OnyxTabNavigator.tsx:294-303`: the discard-changes confirm path is excluded from the fix, so #98240 is still reproducible there**

When a tab registers an unsaved-changes guard, `handleTabPress` calls `event.preventDefault()` and jumps to the target itself after the user confirms (`:205`).

The new code returns before arming `pressedTabRef`, so the confirmed jump gets none of the re-apply protection, and it also leaves whatever stale arm was there from an earlier press.

`IOURequestStepAmount` registers a guard, so this is the Manual tab in the exact flow this PR is fixing: type an amount on Manual, tap Scan, confirm "discard changes". If the same bounce-back race happens, the tab does not switch and the user has to tap twice again.

**Suggested fix**: arm from inside `handleTabPress` right before the post-confirm dispatch, so both paths share one code path:

```tsx
.then(() => {
    hasReappliedPressRef.current = false;
    pressedTabRef.current = {from: currentRouteName, to: targetRoute.name};
    navigation.dispatch(TabActions.jumpTo(targetRoute.name));
});
```

src/libs/Navigation/OnyxTabNavigator.tsx
ikevin127: ### 🟡 **`src/libs/Navigation/OnyxTabNavigator.tsx:124-127`: this is a behavior change for every screen that uses the navigator, not just the IOU start page**

`OnyxTabNavigator` backs `NewChatSelectorPage`, `ShareRootPage`, `SplitExpensePage`, `DistanceRequestStartPage`, `IOURequestStepDistance` and `DynamicEditInviteReceiptPartnerPolicyPage`.

All of them now get a self-healing "if the tab bounces back, jump again" heuristic plus an extra render per tab state event, to fix a race that (as far as the linked issue shows) only happens in the new manual expense flow.

**Two asks:**
1. What actually re-jumps the tab after the RHP back navigation? If it is `useResetIOUType` / `initialRouteName` recomputing from the persisted `selectedTab`, fixing it there is a real fix instead of a re-apply on top of a wrong jump, and it will not leak into the other six screens.
2. If the workaround has to stay, gate it behind an opt-in prop (`shouldReapplyInterruptedTabPress`, default `false`) and set it only from `IOURequestStartPage`. Blast radius drops to the screen that needs it.

**Bug this causes if left global**: any future screen whose tab is legitimately driven programmatically inherits finding 1's bounce, and it will be very hard to trace back to a shared navigator.

src/libs/Navigation/OnyxTabNavigator.tsx
```diff
@@ -121,6 +121,11 @@ function OnyxTabNavigator<TTabName extends string = SelectedTabRequest>({
     const [focusTrapContainerElementMapping, setFocusTrapContainerElementMapping] = useState<Record<string, HTMLElement>>({});
     const [selectedTab, selectedTabResult] = useOnyx(`${ONYXKEYS.COLLECTION.SELECTED_TAB}${id}`);
 
+    const pressedTabRef = useRef<{from: string | undefined; to: string} | undefined>(undefined);
+    const hasReappliedPressRef = useRef(false);
+    const tabNavigationRef = useRef<NavigationProp<ParamListBase> | undefined>(undefined);
+    const [activeTabName, setActiveTabName] = useState<string | undefined>(undefined);
```
ikevin127: ### 🟡 **`src/libs/Navigation/OnyxTabNavigator.tsx:127` and `:279`: the new `activeTabName` state is not needed, and it re-renders the navigator subtree on every tab state event**

The `state` listener already has `newSelectedTab` in hand and the ref holds the navigation object, so the whole thing can technically be done without adding state:

```tsx
const newSelectedTab = routeNames.at(index);
const pressedTab = pressedTabRef.current;
if (pressedTab && !hasReappliedPressRef.current && newSelectedTab === pressedTab.from) {
    hasReappliedPressRef.current = true;
    navigation.dispatch(TabActions.jumpTo(pressedTab.to));
}
```

That drops one `useState`, one `useRef` and one `useEffect`. If dispatching from inside the state listener feels re-entrant, defer with a `requestAnimationFrame` rather than round-tripping through React state (note `InteractionManager` is off limits per `contributingGuides/INTERACTION_MANAGER.md`).

Worth doing because `setActiveTabName` currently fires on first mount and on every tab change for all six consumer screens, including the pager-heavy `SplitExpensePage`.

src/hooks/useDefaultParticipants.ts
```diff
@@ -81,8 +82,11 @@ function useDefaultParticipants({sourceReport, transaction, iouType, isNewManual
             return [];
         }
 
-        const globalCreateIOUType = iouType === CONST.IOU.TYPE.TRACK ? CONST.IOU.TYPE.CREATE : iouType;
-        const canUseDefaultPolicy = shouldUseDefaultExpensePolicy(globalCreateIOUType, defaultExpensePolicy, amountOwed, userBillingGracePeriodEnds, ownerBillingGracePeriodEnd, accountID);
+        if (iouType === CONST.IOU.TYPE.TRACK) {
```
ikevin127, line 86: ### 🟡 **`src/hooks/useDefaultParticipants.ts:85-87`: the new `TRACK` branch always goes through `useSelfDMReport()`, which fabricates a report with a random ID when the real self-DM is not loaded**

`useSelfDMReport` ends with `return selfDMReport ?? buildOptimisticSelfDMReport(DateUtils.getDBTime())`, and `buildOptimisticSelfDMReport` calls `generateReportID()`.

When the real self-DM is not in Onyx yet, the hook returns a brand new object with a brand new random `reportID` on every render. So `participants` recomputes with a different `reportID` each render, and `IOURequestStepConfirmation`'s auto-assign effect (`:506-524`) keeps re-running (`setMoneyRequestParticipants` plus `navigation.setParams({iouType: TRACK})`) until the Onyx write lands.

This sharp edge existed before for the `autoReporting === false` case, but the PR routes *every* track expense through it, and `isLoading` does not wait on the report collection, so a track expense started right after boot can be seeded with a participant pointing at a report that does not exist.

**Suggested fix**: hold off until the real report resolves, rather than seeding a fabricated one:

```tsx
if (iouType === CONST.IOU.TYPE.TRACK) {
    // `useSelfDMReport` falls back to an optimistic report with a random ID; don't seed participants off that.
    return selfDMReport?.reportID && !isOptimisticSelfDM ? getMoneyRequestParticipantsFromReport(selfDMReport, accountID).filter((p) => p.selected) : [];
}
```

or read `ONYXKEYS.SELF_DM_REPORT_ID` here and fold its loading state into `isLoading`. The unit test mocks `useSelfDMReport` with a stable object, so this is invisible to the suite as written.

src/libs/IOUUtils.ts
```diff
@@ -547,6 +547,14 @@ function pickReportForPolicy(...reports: Array<OnyxEntry<Report>>): OnyxEntry<Re
...
 }
 
+function getSelectedWorkspacePolicyID(transaction: OnyxEntry<Transaction>, action: IOUAction): string | undefined {
+    if (action === CONST.IOU.ACTION.EDIT) {
+        return undefined;
+    }
+
+    return transaction?.participants?.find((participant) => participant?.isSender)?.policyID ?? transaction?.participants?.find((participant) => participant?.isPolicyExpenseChat)?.policyID;
+}
+
```
ikevin127, line 558: 🟡 **`src/libs/IOUUtils.ts:550-556`: the new exported helper has no JSDoc, and it swallowed four inline comments that carried real context.**

Every neighbour in this file (`reportHasRealPolicy`, `pickReportForPolicy`, `resolveReportForMoneyRequest`) has a doc block. The four call sites each lost the `Skip the placeholder '_FAKE_' self-DM policy so it doesn't shadow the selected workspace chat's real policy. See #96576.` comment, and none of that reasoning moved into the helper. Two non-obvious rules are now completely undocumented in code:

- why `EDIT` returns `undefined` (the reason only exists in a test name: an existing expense is authoritative about its report)
- why `isSender` wins over `isPolicyExpenseChat` (it mirrors `getIOURequestPolicyID`, which already prefers the sender, so the two stay consistent when a transaction carries both a receiver workspace chat and an invoice sender)

**Suggested fix:**

```ts
/**
 * The policyID of the workspace the user explicitly picked for this expense, taken from the transaction participants.
 *
 * Call sites use this ahead of the report-derived policyID so a placeholder/stale route report (e.g. the self-DM a
 * submissions-disabled workspace flow is seeded onto, which carries the '_FAKE_' policyID) can't shadow the selected
 * workspace chat's real policy. See #96576.
 *
 * The invoice sender wins over the workspace chat participant to match `getIOURequestPolicyID`'s precedence, which
 * matters when a transaction carries both. Returns undefined while editing: an existing expense's own report is
 * authoritative, so the report-derived policy must keep driving the page.
 */
```

Without this the next person reading `getSelectedWorkspacePolicyID(transaction, action) ?? getIOURequestPolicyID(...)` has no way to know the `EDIT` early return is load-bearing for split-expense editing, and will "simplify" it away.

tests/unit/hooks/useDefaultParticipants.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    chatType: CONST.REPORT.CHAT_TYPE.SELF_DM,
+} as Report;
+
+const workspaceChat = {
+    reportID: 'workspaceChat1',
+    type: CONST.REPORT.TYPE.CHAT,
+    chatType: CONST.REPORT.CHAT_TYPE.POLICY_EXPENSE_CHAT,
+    policyID: POLICY_ID,
+    ownerAccountID: ACCOUNT_ID,
+} as Report;
```
ikevin127: ### 🟢 **`tests/unit/hooks/useDefaultParticipants.test.ts:18-30`: `as Report` casts where a typed factory already exists**

The file already imports `createRandomTransaction` from `tests/utils/collections`, and `createRandomReport(index, chatType)` in `tests/utils/collections/reports.ts` returns a real `Report`. Use it and drop both casts:

```ts
const mockSelfDMReport: Report = {...createRandomReport(1, CONST.REPORT.CHAT_TYPE.SELF_DM), reportID: 'selfDM1'};
const workspaceChat: Report = {...createRandomReport(2, CONST.REPORT.CHAT_TYPE.POLICY_EXPENSE_CHAT), reportID: 'workspaceChat1', policyID: POLICY_ID, ownerAccountID: ACCOUNT_ID};
```

**Casts here are not cosmetic**: `as Report` silences the compiler, so if `getMoneyRequestParticipantsFromReport` starts reading a field these fixtures do not have, the test keeps passing against an object that could never come out of Onyx.

Same file, three smaller ones:
- `mockDefaultExpensePolicy` (`:32-38`) has no type annotation, so it is never checked against `Policy` and will silently drift from `useDefaultExpensePolicy`'s return type. Annotate it or build it from `createRandomPolicy`.
- `useCurrentUserPersonalDetails` is mocked with a hardcoded `accountID: 1` while the file defines `const ACCOUNT_ID = 1`. Rename the constant to `mockAccountID` so the jest factory can reference it (jest allows out-of-scope variables prefixed with `mock`) and the two cannot drift.
- Mocking `@hooks/useOnyx` wholesale (`:49-52`) means `isLoading` is never exercised by any test in this file, and any future `useOnyx` read added to the hook silently returns `undefined`. Since the comment says the only reason is settle timing, seeding the NVP keys in `beforeAll` and awaiting `waitForBatchedUpdates` would keep the real hook in play.

ikevin127 (approved):
**🟢 LGTM** - Thanks for addressing all comments. I just re-verified the 3 issues fixed by this PR after the latest commit, and all are passing!

puneetlath (approved):

PR #98423, opened 2026-08-12
Fix workspace navigation after distance settings deep link

fedirjh:
### Explanation of Change

Opening the distance settings RHP from another tab could construct an incomplete workspace background on narrow screens. The navigation action retained the `WorkspaceSplitNavigator` name but dropped its nested state, where the policy-scoped Workspace Initial and Distance Rates routes live. Closing the RHP could therefore reveal `/workspaces/undefined`.

This change:
- Preserves the nested split state when constructing the matching workspace background beneath an RHP.
- Prevents `WorkspaceInitialPage` from fetching policy data, showing Not Found, or closing the RHP before a route policy ID exists.
- Clears consumed deep-link hints when revealing a newly created workspace so an earlier Distance Rates route cannot be replayed.
- Adds regression tests for missing policy parameters, nested workspace background state, and repeated workspace creation after a distance-settings deep link.

### Fixed Issues

$ https://github.com/Expensify/App/issues/98365
PROPOSAL:

### Tests

1. Use a narrow layout and sign in.
2. Create a workspace and enable a fixed commuter exclusion.
3. Create a distance expense in that workspace.
4. Open the expense and select the workspace distance settings link.
5. Verify Distance Rates Settings opens for the expense workspace.
6. Press Back and verify Distance Rates opens for the same workspace.
7. Press Back and verify the workspace page opens for the same workspace; confirm the URL never contains `/workspaces/undefined`.
8. Press Back and verify the Workspaces list opens.
9. Select **New workspace** and confirm creation.
10. Verify the new workspace opens without a crash and the previous workspace's Distance Rates page is not restored.
11. Repeat steps 4–10 and verify the same result on every attempt.

### Offline tests

1. Complete steps 1–5 above while online so the workspace data is available locally.
2. Disable the network connection.
3. Complete steps 6–8.
4. Verify every Back action opens the expected cached workspace screen, the URL never contains `/workspaces/undefined`, and the app does not crash.
5. Re-enable the network connection and complete steps 9–11.

### QA Steps

// TODO: These must be filled out, or the issue title must include "[No QA]".

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

https://github.com/user-attachments/assets/b1586135-25be-48da-ae88-ca41e14ecd66

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


src/libs/Navigation/AppNavigator/createRootStackNavigator/GetStateForActionHandlers.ts
```diff
@@ -449,7 +449,8 @@ function handleReplaceFullscreenUnderRHP(
         }
         const staleTabState = existingTabState ? markFocusedTabRouteForRemount(updatedTabState, existingTabState) : updatedTabState;
 
-        const updatedTabRoute = {...existingTabRoute, state: staleTabState} as StackNavigationState<ParamListBase>['routes'][number];
+        // Drop consumed deep-link hints before remounting, or React Navigation can replay the old target over the new state.
+        const updatedTabRoute = {...withSanitizedDeepLinkParams(existingTabRoute, undefined), state: staleTabState} as StackNavigationState<ParamListBase>['routes'][number];
```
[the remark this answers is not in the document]
fedirjh: Good catch. The state-only hint comes from the separate nested-push flow, while this PR addresses stale `params.screen` hints from the reported distance-settings flow. Expanding the shared sanitizer would increase this regression fix’s blast radius, so I propose handling the state-only form in a focused follow-up with dedicated coverage.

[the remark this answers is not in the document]
fedirjh: cc @sumo-slonik

src/libs/Navigation/helpers/linkTo/index.ts
```diff
@@ -134,7 +134,19 @@ function shouldChangeToMatchingFullScreen(
     return newFocusedRoute?.name === SCREENS.SETTINGS.SUBSCRIPTION.ADD_PAYMENT_CARD && lastActiveScreen !== SCREENS.SETTINGS.SUBSCRIPTION.ROOT;
 }
 
-export {isSwitchingTabsWithinTabNavigator, getActiveScreenInRoute, shouldChangeToMatchingFullScreen, isNavigatingToReportActionWithinSameReport};
+function getMatchingFullScreenRouteParams(matchingFullScreenRoute: NavigationPartialRoute) {
```
marufsharifi: **Annotate the union return type.**

This helper returns two structurally different shapes — a bare params object (`return matchingFullScreenRoute.params`) or a `{screen, params}` hint object — but the return type is left to inference, so a caller can't tell from the signature that a `screen`-shaped hint may come back.

```ts
function getMatchingFullScreenRouteParams(
    matchingFullScreenRoute: NavigationPartialRoute,
): NavigationPartialRoute['params'] | {screen: string; params: NavigationPartialRoute['params'] | undefined} {
```

The ambiguity between "raw params" and "nested `{screen, params, state}` hint" is exactly what caused the bug being fixed here. An explicit type makes the two contracts visible at every call site and guards against a future edit dropping `state` again.

tests/navigation/linkToHelpersTests.ts
```diff
@@ -114,6 +120,30 @@ describe('getActiveScreenInRoute', () => {
     });
 });
 
+describe('getMatchingFullScreenRouteParams', () => {
```
marufsharifi, line 123: **Cover the two untested branches of `getMatchingFullScreenRouteParams`.**

The new `describe` only exercises the `lastRoute` + nested `state` path. The other two branches — which decide what gets passed to the navigator — are unverified:

1. `!lastRoute` → returns `matchingFullScreenRoute.params`
2. `lastRoute` present but **no** `state` → returns bare `lastRoute.params` (not `{...params, state}`)

```ts
it('returns the route params when there is no nested state', () => {
    const route: NavigationPartialRoute = {name: NAVIGATORS.WORKSPACE_NAVIGATOR, params: {policyID: '1'}};
    expect(getMatchingFullScreenRouteParams(route)).toEqual({policyID: '1'});
});

it('omits state when the last route has none', () => {
    const route: NavigationPartialRoute = {
        name: NAVIGATORS.WORKSPACE_NAVIGATOR,
        state: {index: 0, routes: [{name: SCREENS.WORKSPACE.INITIAL, params: {policyID: '1'}}]},
    };
    expect(getMatchingFullScreenRouteParams(route)).toEqual({screen: SCREENS.WORKSPACE.INITIAL, params: {policyID: '1'}});
});
```

Branch (2) is where the old regression lived — a test pins the "don't wrap params in a `state` key when there's no nested state" contract so it can't silently break.

src/libs/Navigation/helpers/swapBackgroundTabForRHPTarget.ts
```diff
@@ -95,12 +95,11 @@ function swapBackgroundTabForRHPTarget(currentState: NavigationState | undefined
         navigationRef.dispatch(additionalAction);
     } else {
         // Plain tab switch within the existing TAB_NAVIGATOR.
-        const lastRouteInMatchingFullScreen = matchingFullScreenRoute.state?.routes?.at(-1);
         const additionalAction: StackNavigationAction = {
             type: CONST.NAVIGATION.ACTION_TYPE.NAVIGATE,
             payload: {
                 name: matchingFullScreenRoute.name,
-                params: lastRouteInMatchingFullScreen ? {screen: lastRouteInMatchingFullScreen.name, params: lastRouteInMatchingFullScreen.params} : matchingFullScreenRoute.params,
+                params: getMatchingFullScreenRouteParams(matchingFullScreenRoute),
```
marufsharifi, line 102: **This path now carries the fix but has no dedicated test.**

`swapBackgroundTabForRHPTarget` was switched to the shared `getMatchingFullScreenRouteParams`, so it now also preserves nested split `state` — the same behavior this PR is fixing. But unlike the `linkTo` and `handleReplaceFullscreenUnderRHP` paths, there's no unit test asserting this navigator produces a payload with the nested state intact.

Both cross-tab entry points feed the RHP background reconstruction; if only one is tested, a future refactor could regress `/workspaces/undefined` through this branch while all existing tests stay green. A small test mirroring the `getMatchingFullScreenRouteParams` case, asserting the dispatched `payload.params` retains `state`, would close the gap.

src/pages/workspace/WorkspaceInitialPage.tsx
```diff
@@ -232,17 +234,18 @@ function WorkspaceInitialPage({policyDraft, policy: policyProp, route}: Workspac
...
     // so checkIfShouldShowPolicy returns false. Suppress NotFound during this loading window.
-    const computedShouldShowNotFoundPage = isWorkspacesTabFocused && !shouldShowPolicy && !policy?.isLoading && (!isPendingDelete || prevIsPendingDelete);
+    const computedShouldShowNotFoundPage = !!routePolicyID && isWorkspacesTabFocused && !shouldShowPolicy && !policy?.isLoading && (!isPendingDelete || prevIsPendingDelete);
     // Latch to true: once the not-found state is detected, keep showing it so the normal
     // workspace content doesn't flash during the exit animation when navigation state
     // changes (e.g., isWorkspacesTabFocused becomes false after StackActions.pop()).
     const prevShouldShowNotFoundPage = usePrevious(computedShouldShowNotFoundPage);
-    const shouldShowNotFoundPage = computedShouldShowNotFoundPage || !!prevShouldShowNotFoundPage;
+    const prevRoutePolicyID = usePrevious(routePolicyID);
+    const shouldShowNotFoundPage = computedShouldShowNotFoundPage || (prevRoutePolicyID === routePolicyID && !!prevShouldShowNotFoundPage);
```
marufsharifi: **Latch reset across `policyID` is the right fix — please add a test for it.**

```ts
const shouldShowNotFoundPage = computedShouldShowNotFoundPage || (prevRoutePolicyID === routePolicyID && !!prevShouldShowNotFoundPage);
```

Gating the latch on `prevRoutePolicyID === routePolicyID` is exactly what prevents a previous policy's Not-Found state from leaking onto a newly created workspace (the repeated-creation scenario). But this behavior — "the latch does **not** persist when `routePolicyID` changes" — isn't asserted anywhere, and it's the subtle part most likely to regress. See my comment on `WorkspaceInitialPageTest.tsx`.

src/libs/Navigation/helpers/linkTo/index.ts
```diff
@@ -134,7 +134,21 @@ function shouldChangeToMatchingFullScreen(
...
+    matchingFullScreenRoute: NavigationPartialRoute,
+): NavigationPartialRoute['params'] | {screen: string; params: NavigationPartialRoute['params'] | undefined} {
+    const lastRoute = matchingFullScreenRoute.state?.routes?.at(-1);
+    if (!lastRoute) {
+        return matchingFullScreenRoute.params;
+    }
+
+    return {
+        screen: lastRoute.name,
+        params: lastRoute.state ? {...lastRoute.params, state: lastRoute.state} : lastRoute.params,
```
marufsharifi, line 151: **This now preserves nested `state` for *all* matching-fullscreen navigators, not just Workspace.**

`getMatchingFullScreenRouteParams` runs for every matching-fullscreen tab switch, so Reports / Settings / Search split navigators are also affected. Before this PR their split was **reset** to its initial route on tab switch; now the full nested stack is **restored**. That's very likely the more-correct behavior and aligns this branch with the `state`-preserving branch just above (~L249), but the added tests only cover `WORKSPACE_SPLIT_NAVIGATOR`.

- Could you confirm a non-workspace matching-fullscreen RHP open (e.g. Reports/Settings) still lands on the expected background screen and doesn't jump to an unexpected deep sub-screen?
- If it checks out, consider a one-line test for a non-workspace navigator to lock the generalization in.

garrettmknight (dismissed):

src/libs/Navigation/helpers/linkTo/index.ts
fedirjh: Confirmed — added a regression test verifying the Subscription page remains behind the payment-card RHP when opened from Reports.

src/libs/Navigation/AppNavigator/createRootStackNavigator/GetStateForActionHandlers.ts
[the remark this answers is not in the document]
sumo-slonik: I'm starting the review now. I'll let you know this afternoon what I think.

src/libs/Navigation/helpers/linkTo/index.ts
```diff
@@ -134,7 +134,21 @@ function shouldChangeToMatchingFullScreen(
     return newFocusedRoute?.name === SCREENS.SETTINGS.SUBSCRIPTION.ADD_PAYMENT_CARD && lastActiveScreen !== SCREENS.SETTINGS.SUBSCRIPTION.ROOT;
 }
 
-export {isSwitchingTabsWithinTabNavigator, getActiveScreenInRoute, shouldChangeToMatchingFullScreen, isNavigatingToReportActionWithinSameReport};
+function getMatchingFullScreenRouteParams(
```
marufsharifi, line 141: The `state`-nested-into-`params` branch on L147 is the actual fix here — it's the one behavioral difference from the old inline code, which dropped `lastRoute.state` and left `WorkspaceSplitNavigator` stateless (→ `/workspaces/undefined`, #98365). Because that's non-obvious and easy to "simplify" back into the buggy form, it deserves a `why` comment (also required by the PR checklist):

```suggestion
/**
 * Builds the `params` for a tab-switch NAVIGATE that reveals a matching full-screen background.
 * When the matched split carries nested state (e.g. WorkspaceSplitNavigator: [Initial, DistanceRates]),
 * forward it under `params.state` — RN's deep-link hint shape — otherwise the split is rebuilt
 * stateless and closing the RHP lands on `/workspaces/undefined` (#98365).
 */
function getMatchingFullScreenRouteParams(
```

marufsharifi (commented):
Looks good, but just one minor suggestion

src/libs/Navigation/AppNavigator/createRootStackNavigator/GetStateForActionHandlers.ts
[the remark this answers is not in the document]
sumo-slonik: I'm wondering if it would be better to fix this properly right away in this spot. As it stands, we're introducing a case here that this code is specifically trying to prevent.

That said, if the fix PR is going to be a direct follow-up, I'm okay with keeping it separate, as long as that PR contains the complete fix for the underlying mechanism and appropriate test coverage.

What do you think, @WojtekBoman?

src/libs/Navigation/helpers/linkTo/index.ts
sumo-slonik, line 151: Could you please check whether this change affects the behavior of swipe-back gestures and the back button? Overwriting the navigation state like this could potentially impact the navigation history.

sumo-slonik: I'm particularly curious about the case where we arrive here from a flow that doesn't have the Workspace Initial Page in its navigation history.

For example:

1. Go to Inbox
2. Open any chat
3. Open Chat Details (RHP)
4. Use "Go to Workspace"

From there, could you check the behavior of both the back button and the swipe-back gesture? We should end up back in the Inbox, following the path we took to get there, rather than being taken to the Workspace Initial Page.

src/libs/Navigation/AppNavigator/createRootStackNavigator/GetStateForActionHandlers.ts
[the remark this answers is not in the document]
WojtekBoman: Hey 👋 I went through the PR and noticed 2 things :)

- I agree with @sumo-slonik, I think it's worth folding this one in here. `getRehydratedTabNavigatorStateAfterPush()` writes `params.state` onto the very same `TAB_NAVIGATOR` route this call sanitizes, and `getStateFromParams` returns `params.state` before it even looks at `params.screen`, so both encodings replay through the same code path. That means a follow-up PR would end up touching this exact line again, and until then the fix only covers one of the two ways the same stale state reaches us, which makes it easy to read the sanitizer as complete when it isn't.

  It's a one-line predicate rather than a new mechanism. It just has to match RN's state shape instead of the presence of a `state` key, because `SETTINGS.PROFILE.DYNAMIC_ADDRESS_STATE` has a legitimate `state?: string` param:

  ```ts
  const isLongFormDeepLinkHint = (params: unknown): boolean => isRecord(params) && isRecord(params.state) && Array.isArray(params.state.routes);

  const hintParams = isRecord(route.params) && (typeof route.params.screen === 'string' || isLongFormDeepLinkHint(route.params)) ? route.params : undefined;
  ```

  Everything downstream already handles it, since `STALE_DEEP_LINK_PARAM_KEYS` contains `state` already. The first-visit branches in `linkTo` and `swapBackgroundTabForRHPTarget` should keep the long form though, a never-visited tab needs the full nested state.

- `WorkspaceInitialPage` could be a smaller change. A `WORKSPACE.INITIAL` route with no `policyID` means the navigation state is corrupt, so Not Found is the right answer. Only the `!routePolicyID` guard in `fetchPolicyData` is load-bearing for the crash, and with it main's existing expression already renders Not Found on its own.

marufsharifi:
@fedirjh, Gentle bump. thanks.

marufsharifi:
@fedirjh, could you please merge the main. thanks.

src/libs/Navigation/AppNavigator/createRootStackNavigator/GetStateForActionHandlers.ts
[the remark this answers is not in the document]
fedirjh: Updated the shared sanitizer to recognize the long-form hint only when `params.state` contains a `routes` array. This clears the stale nested state while preserving legitimate scalar `state` params. I added helper and handler regression coverage and reduced `WorkspaceInitialPage` to the missing-route fetch guard.

src/libs/Navigation/helpers/linkTo/index.ts
fedirjh: Done.

src/libs/Navigation/helpers/linkTo/index.ts
fedirjh: Added integration coverage for navigating from Reports with an open RHP to a workspace and back. It verifies that the Reports split is restored with Inbox as its base and the original report route still on top, rather than replacing that history with Workspace Initial.

Gonals:
There are a few lint errors

marufsharifi:
@fedirjh, could you please fix the lint failure and merge the main. thanks.

fedirjh:
@marufsharifi Fixed and ready.

marufsharifi (approved):
LGTM

marufsharifi:
@Gonals, could you please take a look here when you get a chance. thanks.

Gonals (approved):

Issue #98416, opened 2026-08-12
[Due for payment 2026-09-09] [Domain Control - Request access] [FE] Show a "Domain already set up" page when you add a domain that already exists


rayane-d:
## Action Performed:

1. Sign in as a user who is not an admin of a domain another account already set up.
2. Go to Settings > Domains > Add domain.
3. Type that domain and click Continue.

## Actual Result:

`AddDomainPage` shows the error under the input: "This domain already exists. If the domain is listed on
your screen, please click it to become an administrator...".

The backend sends a small `domain_<domainAccountID>` entry with the error. It lands in
`ONYXKEYS.COLLECTION.DOMAIN`, so a new row shows up in `DomainsListPage` and the user can click it to
reach the Access restricted page.

## Expected Result:

Go to a new page instead of showing an inline error, and ask the user there if they want to request
admin access:

<img width="750" height="1624" alt="Image" src="https://github.com/user-attachments/assets/891c73a9-8f97-4f5e-9895-728c59392905" />

## What the API returns

The backend already sends what this page needs. When `CreateDomain` fails because the domain exists,
the response has `jsonCode` 300 and these Onyx updates:

- Key: `createDomainForm`, Value: `{errors: {<timestamp>: <message>}}`
- Key: `createDomainForm`, Value:  `{domainAccountID: <number>}`
- Key:  `domain_<domainAccountID>`, Value: `{accountID: <number>, email: "+@<domain>"}`

When `CreateDomain` fails because the domain exists, the value in `createDomainForm.domainAccountID`  tells which domain to request access to.

---

BE issue:
- https://github.com/Expensify/App/issues/97980

PRs:
- https://github.com/Expensify/Auth/pull/23587
- https://github.com/Expensify/Web-Expensify/pull/55217

<details><summary>Issue Owner</summary>Current Issue Owner: @ZhenjaHorbach </details>

<details><summary>Issue Owner</summary>Current Issue Owner: @mallenexpensify</details>

jakubstec:
> OD checks whether the user has the SNVP shared with them to disable the button. So, we'll need the data that will be added in the other sub-issues:

Thank you! So I'll note that and handle it in the separate PR then 😄

jakubstec:
in progress

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100185 was identified to be related to the PR https://github.com/Expensify/App/pull/98489 that attempted to fix this issue.

[...]

Issue #97982, opened 2026-08-06
[Domain Control - Request access] [BE] Onyx updates when a user requests domain adminship


rayane-d:
Part of https://github.com/Expensify/App/issues/97053

Add Onyx updates to the existing `RequestDomainAdminship` command: the new request entry in the domain data and the requester's personal details to the domain admins, and the requester's own entry entry in the domain data back to them.

Add a NewDot command that takes a `domainAccountID` and admin auth token, instead of `domainName` and domain auth token.

<details><summary>Issue Owner</summary>Current Issue Owner: @rayane-d</details>

rayane-d:
Auth PR is ready for review:
- https://github.com/Expensify/Auth/pull/23737

rayane-d:
BE PRs are merged, awaiting deployment

- [x] https://github.com/Expensify/Auth/pull/23737
- [x] https://github.com/Expensify/Web-Expensify/pull/55385

rayane-d:
done

jakubstec:
hey, could we add something to OpenApp that tells the requester they already have a pending adminship request? right now the only feedback about pending requests is in `OpenDomainPage`, which a requester can't reach and the state only survives in Onyx from the `RequestDomainAdminship` response. once we clear Onyx, user can retry the request access flow, despite already requesting it.

rayane-d:
Hey @jakubstec, we already return the request itself to the requester in the domain Onyx data in `OpenApp`:


<img width="1511" height="803" alt="Image" src="https://github.com/user-attachments/assets/cdca1b4c-d62a-4aff-b8b2-6d766c533411" />

Let me know if I'm missing something.

jakubstec:
> Hey @jakubstec, we already return the request itself to the requester in the domain Onyx data in OpenApp:

@rayane-d 
Hmm, did you send me OpenApp of the requester or an domain admin? because in my case I don't see that - can you take a look at this recording? I might've missed something too 😅

on recording
- there is pending request (confirmed on OD)
- on requester account OpenApp after clearing onyx data returns no `domain_adminRequesters`
- When I check it on domain admin account i can see it

https://github.com/user-attachments/assets/2c3533c6-c3ae-41cd-84bf-9391b9231fd1

rayane-d:
> did you send me OpenApp of the requester or an domain admin?

@jakubstec - In my screenshot, it's the OpenApp of the requester.

Admins get every pending adminship request, while requesters only get their own.

[...]

Issue #97984, opened 2026-08-06
[Domain Control - Request access] [BE] Clear the pending request in Onyx for all admins on domain adminship request accept and decline


rayane-d:
Part of https://github.com/Expensify/App/issues/97053

Declining on NewDot needs a new command, `DeclineDomainAdminshipRequest` that use the admin authToken and domainAccountID. it should do necessary checks and queues the Onyx updates to the admins.

Accepting already works through `AddDomainAdmin`, but no Onyx updates are sent. We need to queue Onyx updates to the requester and domain admins.

<details><summary>Issue Owner</summary>Current Issue Owner: @rayane-d</details>

rayane-d:
in progress

rayane-d:
done

Issue #97981, opened 2026-08-06
[Domain Control - Request access] [BE] Send the pending adminship requests and requester details in the domain Onyx data


rayane-d:
Part of https://github.com/Expensify/App/issues/97053

Add a `domain_adminRequesters` field to the domain Onyx object, keyed by requester accountID with the request `created` time. Domain admins get all requesters, other users only their own entry. 

Also include the requesters in the personal details update that `OpenDomainPage` sends, so the Domain Admins page can show their name and avatar. Details are needed there because requesters can be non-members or on unvalidated domains.

<details><summary>Issue Owner</summary>Current Issue Owner: @rayane-d</details>

rayane-d:
done

PR #98489, opened 2026-08-12
feat: add "Domain already set up" page when you add a domain that already exists

jakubstec:
### Explanation of Change
Previously, when a user tried to add a domain that someone else had already set up in Expensify, `AddDomainPage` just showed an inline error under the input telling them to find the domain in their list and click it - even though that domain wasn't actually in their list, since they had no access to it.

This PR adds a dedicated **"Domain already set up"** page instead:

- New screen/route `WORKSPACES_DOMAIN_ALREADY_EXISTS` (`workspaces/domain-already-exists/:domainAccountID`) rendering the new `DomainAlreadyExistsPage`. It's a `ConfirmationPage` with "Never mind" (back to the domains list) and "Ask for admin access", which sends the adminship request (that needed a new `isButtonDisabled` prop on `ConfirmationPage`).
- **Ask for admin access** (#98908) calls the new `RequestDomainAdminship` command via `requestDomainAdminship`. The BE answers by merging `{domain_adminRequesters: {<accountID>: 'read'}}` into `domain_<accountID>`, and we write the same thing optimistically, then go back to the domains list. Coming back into the flow for that domain shows the button as a disabled "Request sent"; a failed request rolls the optimistic write back, so the button is actionable again (the full request-sent UX is #97982).
- Because the entry the failure came with was already dropped, that optimistic merge would otherwise *resurrect* the `domain_<accountID>` key. An Onyx merge rollback removes fields but never the key, so a failed request would leave `{domain_adminRequesters: {}}` behind, and the next add-domain attempt would see that key in its snapshot and wrongly show "you already have access". `requestDomainAdminship` therefore takes an `isTransientDomainEntry` flag (derived from `!domain?.accountID`) and uses `Onyx.set(key, null)` in that case.
- When `CreateDomain` fails because the domain is taken, the BE returns a generic error plus the `domainAccountID` of the existing domain and a minimal `domain_<accountID>` Onyx entry. The response is identical whether or not the user already has access to that domain, so `AddDomainPage` snapshots the domain keys it had right before submitting and compares the returned accountID against it:
    - **accountID was in the snapshot** (a domain the user can already see) → show the inline error `domain.addDomain.alreadyHaveAccessError` via the new `setCreateDomainAlreadyHaveAccessError` action and stay on the page.
    - **accountID wasn't in the snapshot** → `clearDomainFromFailedCreation` deletes the transient `domain_<accountID>` entry so a domain the user has no access to never leaks into the domains list, then we navigate to `DomainAlreadyExistsPage` (`forceReplace`).
    - **no snapshot** (the failure landed without a submit in this mount, e.g. after a remount) → `clearCreateDomainAccountID` and just let the regular BE error show.
- To make sure we never compare against an incomplete domains list, the submit button is held in its loading state while `useIsAppLoadPending()` is `true`, so we can't submit before `OpenApp` has delivered the domains.
- New `shouldHideServerError` prop on `FormProvider`: while `domainAccountID` is set we suppress the generic server error, so it doesn't flash under the input right before we navigate away.
- `createDomain`'s optimistic data now clears `errors` and `domainAccountID`, and the create-domain form is reset both on mount and on unmount, so a stale result from a previous attempt can't drive the flow.
- Extracted `DomainNameOrNotFoundWrapper` out of `DomainAccessRestrictedPage` (loading indicator → not-found page → children with the resolved domain name), with unit tests.

### Fixed Issues
$ https://github.com/Expensify/App/issues/98416
$ https://github.com/Expensify/App/issues/98908
PROPOSAL:

### Tests
1. Sign in as a user who is **not** an admin/member of a domain another account has already set up.
2. Go to Settings > Domains > Add domain.
3. Enter that domain and click Continue.
4. Verify you land on the new "Domain exists" page ("Domain already set up. Request access?") and that **no error flashes under the input** before the navigation.
5. Click "Ask for admin access" and verify you're taken back to the Domains list, with no stray row for the domain you don't have access to.
6. Repeat steps 2-3 and verify you land on the "Domain already set up" page with the button already showing the disabled "Request sent" state, since the request is still pending.
7. Click "Never mind" and verify you're taken back to the Domains list, still with no stray row.
8. Repeat steps 2-3 once more and go back using the header back button - verify you land on the Domains list again
9. Sign in as a user who **is** already an admin/security group member of an existing domain.
10. Go to Settings > Domains > Add domain, enter that same domain, and click Continue.
11. Verify an inline error appears under the input ("This domain already exists in your account.") instead of navigating to the new page, and that the generic backend error is not shown.
12. Verify that domain still appears exactly once, correctly, in the Domains list.
13. Reload the app and open Settings > Domains > Add domain as fast as possible (before the domains finish loading) - verify Continue stays in its loading state until the domains have arrived, and then the flow above still behaves correctly.

### Offline tests
N/A

### QA Steps
Same as Tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/8e36f90e-a837-41ef-96cf-b2d3310df3dc

</details>


jakubstec:
@codex review

src/pages/domain/DomainAlreadyExistsPage.tsx
```diff
@@ -0,0 +1,72 @@
...
+                description={translate('domain.domainAlreadyExists.description')}
+                descriptionStyle={styles.textSupporting}
+                shouldShowSecondaryButton
+                secondaryButtonText={translate('domain.domainAlreadyExists.neverMind')}
+                onSecondaryButtonPress={goToDomainsList}
+                shouldShowButton
+                buttonText={translate('domain.domainAlreadyExists.requestAccess')}
+                onButtonPress={() => {
+                    // TODO: call the BE "request domain admin access" endpoint for domainAccountID once it exists.
+                    goToDomainsList();
```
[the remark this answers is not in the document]
jakubstec: will be handled in separate sub-issue. right now we can put this PR on hold

src/languages/en.ts
```diff
@@ -10715,6 +10715,14 @@ const translations = {
             subtitle: 'Enter the name of the private domain you want to access (e.g. expensify.com).',
             domainName: 'Domain name',
             newDomain: 'New domain',
+            alreadyHaveAccessError: 'This domain already exists in your account.',
+        },
+        domainAlreadyExists: {
+            headerTitle: 'Domain exists',
+            title: 'Domain already set up. Request access?',
+            description: 'Someone already set this domain up in Expensify. Want to request admin access?',
+            neverMind: 'Nevermind',
```
war-in: We already have `neverMind` in translations. How about extracting it to common?

src/libs/Navigation/AppNavigator/ModalStackNavigators/index.tsx
```diff
@@ -119,6 +119,10 @@ const OPTIONS_PER_SCREEN: Partial<Record<Screen, PlatformStackNavigationOptions>
     [SCREENS.TWO_FACTOR_AUTH.SUCCESS]: {
         animationTypeForReplace: 'push',
     },
+    // Opened via forceReplace from the add domain page, so it needs a forward (push) transition instead of the default reversed (pop) one.
```
war-in: Other screens don't have the description here. I think we should follow the same pattern here, or standardise comments for every screen with `push` animation

src/selectors/Domain.ts
```diff
@@ -213,6 +213,20 @@ function isAdminSelector(accountID: number) {
     };
 }
 
+/**
+ * Creates a selector that checks if a given account ID has any access to the domain,
+ * either as an admin or as a member of one of its security groups.
+ */
+function hasDomainAccessSelector(accountID: number) {
```
war-in: We usually use names with "selector" inside when defining `useOnyx`-related selectors. Do we have any non-Onyx selectors there? Maybe we could find a better place for it, or at least rename the function so it's not misleading?

src/pages/domain/DomainsListPage.tsx
war-in: I don't think we need the entire personal details here. Wouldn't 

```tsx
const [currentUserAccountID] = useOnyx(ONYXKEYS.SESSION, {selector: accountIDSelector});
```

be enough?

src/pages/domain/DomainsListPage.tsx
```diff
@@ -57,7 +57,14 @@ function DomainsListPage() {
                 continue;
             }
 
-            const isDomainAdmin = isAdminSelector(currentUserPersonalDetails?.accountID)(domain);
+            const currentUserAccountID = currentUserPersonalDetails?.accountID;
+            const isDomainAdmin = isAdminSelector(currentUserAccountID)(domain);
```
war-in, line 80: We won't need this boolean if we skip the entire row in the next condition. Let's move it after the condition

src/pages/domain/DomainsListPage.tsx
```diff
@@ -57,7 +57,14 @@ function DomainsListPage() {
...
+
+            // we don't want to show domains that the user has no access to, so we skip them in the list
```
war-in: Is that comment really necessary? I'd say the code is self-explanatory here. Maybe it'd be better to link to some discussion that confirms this is expected behaviour? You can also leave the link in this thread and skip the code comment entirely

src/pages/domain/AddDomainPage.tsx
```diff
@@ -70,6 +73,26 @@ function AddDomainPage() {
         }
     }, [form?.hasCreationSucceeded, allDomains]);
 
+    useEffect(() => {
+        const domainAccountID = form?.domainAccountID;
+        if (!domainAccountID) {
```
war-in, line 83: Don't we also need to check if the creation succeeded? `form?.hasCreationSucceeded`
Or maybe it's always true that this field (domainAccountID) is set when there is already a domain?

src/pages/domain/DomainsListPage.tsx
```diff
@@ -57,7 +57,14 @@ function DomainsListPage() {
...
+            if (!currentUserAccountID || !hasDomainAccessSelector(currentUserAccountID)(domain)) {
```
war-in: Currently, when I create a new account I can see this on the domains list page

Image

I think that after this condition, we'll lose that entry. is it expected?

src/pages/domain/AddDomainPage.tsx
war-in, line 1: Will the `createDomain` command stop returning

```
This domain already exists. If the domain is listed on your screen, please click it to become an administrator. Otherwise please ask an existing domain administrator to add you as an administrator.
```

error for failed requests when the domain already exists? If not, we could still see the flashing error

Image

src/pages/domain/AddDomainPage.tsx
```diff
@@ -70,6 +73,26 @@ function AddDomainPage() {
...
+            return;
+        }
+
+        const existingDomain = allDomains?.[`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`];
+        if (!existingDomain) {
+            return;
+        }
+
+        if (!!currentUserAccountID && hasDomainAccessSelector(currentUserAccountID)(existingDomain)) {
+            setCreateDomainAlreadyHaveAccessError();
```
war-in, line 97: Couldn't the backend send this error directly? I'm worried that we could miss the domain in Onyx. Or maybe this is not possible? 🤔

src/pages/domain/AddDomainPage.tsx
war-in: Isn't it already handled by the next useEffect? 🤔

src/pages/domain/DomainAlreadyExistsPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type DomainAlreadyExistsPageProps = PlatformStackScreenProps<WorkspacesDomainModalNavigatorParamList, typeof SCREENS.WORKSPACES_DOMAIN_ALREADY_EXISTS>;
+
+function DomainAlreadyExistsPage({route}: DomainAlreadyExistsPageProps) {
+    const {asset: GlobeLock} = useMemoizedLazyAsset(() => loadIllustration('GlobeLock'));
+    const {translate} = useLocalize();
+    const styles = useThemeStyles();
+    const domainAccountID = route.params.domainAccountID;
+    // TODO: the BE only sends this domain_<domainAccountID> entry because CreateDomain failed with
+    // "already exists" - the current user has no access to it, so it's not real data. Once "Request access"
```
war-in: What do you mean by "real data"? Also, is this TODO comment really useful?

src/pages/domain/DomainAlreadyExistsPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const goToDomainsList = () => Navigation.goBack(ROUTES.DOMAINS_LIST.getRoute());
+
+    const isDomainNameLoading = isLoadingOnyxValue(domainNameResults);
+    if (isDomainNameLoading) {
+        return <FullScreenLoadingIndicator />;
+    }
+
+    if (!domainName) {
+        return <NotFoundPage onLinkPress={goToDomainsList} />;
+    }
```
war-in: `DomainAccessRestrictedPage` has this logic too. Would it be useful to extract a wrapper component?

src/pages/domain/DomainAlreadyExistsPage.tsx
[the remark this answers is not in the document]
war-in: What do you mean by in a separate sub-issue? Is this the first phase of the feature and the `Ask for admin access` button action will be handled elsewhere? If yes, we should disable it because the current behaviour is confusing

src/pages/domain/DomainAlreadyExistsPage.tsx
jakubstec: Comment might sound weird without context 😅 it was a placeholder note for myself; we're intentionally not cleaning up Onyx `domain_<domainAccountID`> data to handle request access flow, and we need the clean-up afterwards. I suspected, that in review someone would ask about not cleaning it up, but it's required for the flow (implemented here https://github.com/Expensify/App/issues/97980 😄)

jakubstec: but i'll delete this comment, as we are extracting a wrapper component 🙏

src/pages/domain/AddDomainPage.tsx
jakubstec: this clean-up is needed, so the second `Add domain` flow won't have any stale Onyx data, that causes redirect to `Domain exists` page, instead of `Add domain` for the fresh flow. I've actually refactored it a little, and moved L93 `resetCreateDomainForm()` to the next useEffect

```
useEffect(() => {
        return () => {
            clearDraftValues(ONYXKEYS.FORMS.CREATE_DOMAIN_FORM);
            resetCreateDomainForm();
        };
    }, []);
```

src/pages/domain/DomainsListPage.tsx
jakubstec: yes, because when you login and logout this domain with disappear from the list page. same behavior is observable on OldDot - it's expected fix for that 😄

src/pages/domain/DomainAlreadyExistsPage.tsx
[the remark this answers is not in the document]
jakubstec: I was thinking that we are waiting for #97981, #97982 and #97984 so we can handle the adminship request on FE side, and that `Ask for admin access` functionality is out of scope of this PR. Then:
- do we want to put this PR on HOLD until all BE get merged
- or do we want to disable this button, create separate sub-issue for that and go forward with this one?

cc @mountiny @rayane-d 

Screenshot 2026-08-18 at 11 37 01

src/pages/domain/DomainsListPage.tsx
jakubstec: I think so too, didn't want to touch that as it's outside of the scope of this PR - but if you also noticed that I'll change it 😄

src/pages/domain/DomainsListPage.tsx
jakubstec: > I think that after this condition, we'll lose that entry. is it expected?

No it's not expected

this guard was supposed to fix pre-existing behavior, when trying to add already existing domain, where user had no access it added a new entry which disappears after logout (same behavior as OldDot).

jakubstec: Regarding the screenshot you send above, let's imagine a flow

- user is an domain member, not admin
- wants to open domain (e.g swmansion.com)
- we don't want to show this version`Access Restricted` page, but new one right?

Screenshot 2026-08-18 at 13 28 20

also this one:
- user wants to create a new domain
- and types already existing one he belongs to (as a member)
- do we want to redirect user to updated `Access Restricted` page here too, instead of inline error?

Screenshot 2026-08-18 at 13 31 14

So do we need a sub-issue for that? 

cc @mountiny @rayane-d

src/pages/domain/AddDomainPage.tsx
jakubstec: we don't have to check that, because only success returns `hasCreationSucceeded` - when BE sends `domainAccountID` it means that specific error happened (which is a "domain already exists" case)

src/pages/domain/AddDomainPage.tsx
jakubstec: Answering the question above: no, `createDomain` won't stop returning this error, we actually need that failure signal to know we're in the "already exists" case at all.

I think the underlying issue is that we currently don't distinguish 2 flows that share the same error:
a) user is already a member of the domain they're trying to create
b) user is not a member, and gets redirected to the new "Domain already exists" page

For a), we might want to keep this inline error as is, or redirect to an updated "Access restricted" page where the user can request admin access instead (still waiting on feedback here: https://github.com/Expensify/App/pull/98489#discussion_r3803635746)

For b), this generic error is what's causing the flash while we redirect.

For now, I think it's correct to suppress the inline error on the FE only for flow b) while keeping flow a) as is until we get feedback on the sub-issue above.

src/pages/domain/AddDomainPage.tsx
jakubstec: I think it's neccessary because of #97980, `domainAccountID` is used further for request access flow. also, i don't think that we could miss it in Onyx though

jakubstec:
@codex review

mountiny:
@MelvinBot review

src/pages/domain/DomainsListPage.tsx
mountiny: > do we want to redirect user to updated Access Restricted page here too, instead of inline error?

Yes I believe that is what we want.

mountiny: @MelvinBot can you create sub issue for it please

jakubstec:
there are still some things to be done and clarify 😄

jakubstec:
@codex review

src/pages/domain/AddDomainPage.tsx
```diff
@@ -70,9 +76,34 @@ function AddDomainPage() {
...
+            return;
+        }
+
+        if (domainKeysBeforeCreation.current.has(`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`)) {
+            setCreateDomainAlreadyHaveAccessError();
+            return;
+        }
+
+        clearDomainFromFailedCreation(domainAccountID);
+        Navigation.setNavigationActionToMicrotaskQueue(() => Navigation.navigate(ROUTES.WORKSPACES_DOMAIN_ALREADY_EXISTS.getRoute(domainAccountID), {forceReplace: true}));
```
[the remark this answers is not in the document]
jakubstec: @MelvinBot Can we rely on every legitimate domain_<accountID> entry returned by OpenApp having a boolean validated field, including domains shown to non-admin users?
The minimal domain entry returned by a failed CreateDomain request contains only accountID and email, without validated.
If validated is guaranteed for all regular domain entries, we could safely filter domain.validated === undefined from DomainsListPage. This would exclude only the response stub while preserving both validated (true) and legitimate unvalidated (false) domains. I'm trying to find a fix for that codex comment without using Onyx.ConnectWithoutView. Take under consideration that when you create a new acount (e.g abc@swmansion.com , you can see swmansion.com as a domain in the domains list - but, I don't see the account in Domain Members list.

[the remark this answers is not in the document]
MelvinBot: **Short answer: no — I wouldn't rely on `validated === undefined` as the discriminator, and I can't confirm the guarantee you're asking about from code.** The code that assembles `domain_<accountID>` entries for OpenApp (and the "already exists" `CreateDomain` failure stub) lives in the backend (Web-Expensify/Auth), which I don't have access to, so I can't verify whether `validated` is *unconditionally* set on every legitimate entry. The [`validated: boolean`](https://github.com/Expensify/App/blob/2c1165973dc13081cd7fe1d879875bda23639272/src/types/onyx/Domain.ts#L9-L11) in the Onyx type is an App-side assumption, not a runtime guarantee.

Two concrete reasons the filter is risky even if it happens to work today:

1. **It sniffs an undocumented serialization detail.** Filtering on the *absence* of a field couples the list to whatever fields the BE happens to omit from an error response right now. If the stub ever starts carrying `validated` (or `validated:false`), the stub silently reappears; if a real entry ever omits it, a legitimate domain silently vanishes. Neither failure is caught by a type check.
2. **Your swmansion.com observation is the counterexample that matters.** OpenApp already sends a `domain_<accountID>` entry for a domain the user is *not* a member of (just because the email domain matches) — that's a **legitimate OpenApp entry, not the `CreateDomain` stub**, and we don't know if it carries `validated`. So "no `validated`" isn't reliably equivalent to "this is the error stub." You'd be asking one field-presence heuristic to separate three distinct backend paths (real accessible domain / error stub / non-member own-domain entry), none of which document their field sets.

**What I'd do instead (no `Onyx.connectWithoutView`, no field sniffing):**

Filter [`DomainsListPage`](https://github.com/Expensify/App/blob/2c1165973dc13081cd7fe1d879875bda23639272/src/pages/domain/DomainsListPage.tsx#L54-L74) on a *semantic* signal — whether the current user actually has access to the domain (admin via [`isAdminSelector`](https://github.com/Expensify/App/blob/2c1165973dc13081cd7fe1d879875bda23639272/src/selectors/Domain.ts#L203-L214), OR security‑group membership) — rather than on the presence of `validated`. This is more robust than the current source-cleanup-only approach and directly closes the Codex race:

- The Codex P2 is that [`clearDomainFromFailedCreation`](https://github.com/Expensify/App/blob/2c1165973dc13081cd7fe1d879875bda23639272/src/libs/actions/Domain.ts#L414-L416) runs in a component effect, so a back-button unmount before the response lands leaves the stub in Onyx forever. An **access-based filter hides the stub whenever it's present, mounted or not** — you don't depend on any cleanup timing.
- It also hides the swmansion.com non-member entry, which you already said [is the expected behavior](https://github.com/Expensify/App/pull/98489#discussion_r3803582318). Note: the filter must check *real* access (admin entries + security‑group membership), **not** a bare "email domain matches" check — otherwise the @swmansion.com user would still match their own non-member domain and it wouldn't be hidden.
- Add a loading guard on the security‑group data so a member's domain doesn't flicker out of the list while that data is still loading (the concern I raised on the earlier `hasDomainAccess` version).

You can keep the stub in Onyx for the request-access flow (#97980) — the access filter simply prevents it from rendering in the list while it's there, which is what you want.

If you'd rather keep the `validated` approach, please confirm the backend contract first: have someone with Web-Expensify/Auth access check the OpenApp domain-assembly path (is `validated` set for admin **and** non-admin member **and** own-email-domain entries?) and the `CreateDomain` "already exists" failure builder (does the stub omit `validated`?). Absent that confirmation, the access-based filter is the safer bet.

jakubstec:
@codex review

jakubstec:
@MelvinBot review

mountiny:
@MelvinBot review

src/pages/domain/DomainAlreadyExistsPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+                heading={translate('domain.domainAlreadyExists.title')}
+                innerContainerStyle={styles.p10}
+                description={translate('domain.domainAlreadyExists.description')}
+                descriptionStyle={styles.textSupporting}
+                shouldShowSecondaryButton
+                secondaryButtonText={translate('domain.common.neverMind')}
+                onSecondaryButtonPress={goToDomainsList}
+                shouldShowButton
+                buttonText={translate('domain.domainAlreadyExists.requestAccess')}
+                onButtonPress={() => {}}
```
ZhenjaHorbach: Do we need this?
I suppose onButtonPress is an optional param which uses `() => {}` by default

jakubstec: probably not 😅

src/pages/domain/AddDomainPage.tsx
```diff
@@ -70,9 +78,37 @@ function AddDomainPage() {
...
+        if (!domainKeysBeforeCreation.current) {
+            const domainKey = `${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`;
+            if (form.domainKeysBeforeCreation && !form.domainKeysBeforeCreation.includes(domainKey)) {
+                clearDomainFromFailedCreation(domainAccountID);
+            }
+            clearCreateDomainAccountID();
+            return;
+        }
+
+        if (domainKeysBeforeCreation.current.has(`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`)) {
```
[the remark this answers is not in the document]
jakubstec: practically there is a very minor chance for that in my opinion

assets/images/product-illustrations/globe-lock.svg
shawnborton, line 1: Where did you get this illustration? It looks quite borked, the lines are all out of whack: 

Image

ZhenjaHorbach: Nice that you noticed
It's a bit difficult to notice in the light theme 😁

shawnborton: Since they are so similar, perhaps we just reuse the same empty state we have on the Domains page? 
CleanShot 2026-08-21 at 13 16 39@2x

shawnborton: cc @dubielzyk-expensify for thoughts there

dubielzyk-expensify: Yeah that sounds good to me  👍 Let's reuse that one. That globe doesn't have great lines haha

jakubstec: > Where did you get this illustration? It looks quite borked, the lines are all out of whack:

Hey, I got it here https://github.com/Expensify/App/issues/98416#issuecomment-5276337854

> Since they are so similar, perhaps we just reuse the same empty state we have on the Domains page?

On it! Changed 😄

Screenshot 2026-08-24 at 15 45 44

shawnborton: Nice, that feels better. Why is the button disabled there though?

shawnborton: Actually can we use the same dimensions that we are using on the Domains page?

jakubstec: > Nice, that feels better. Why is the button disabled there though?

It's functionality will be implemented as a part of another sub-issue (https://github.com/Expensify/App/issues/98908)

> Actually can we use the same dimensions that we are using on the Domains page?

Of course! It will look like this:

Screenshot 2026-08-24 at 16 03 13

mountiny: I am a slightly confused around this one, I thought that issue for separate flow, but this PR should already implement the request to `Ask for admin access`. Otherwise this is super weird expereince when the user is taken to this page which says "Want to request admin access?" Ah sorry but disabled with no explanation 😄

mountiny (changes requested):
Code looks good to me but I think it does not make sense to ship this without being able to actually request the admin access https://github.com/Expensify/App/pull/98489#discussion_r3846894978

assets/images/product-illustrations/globe-lock.svg
ZhenjaHorbach: Probably yes
Let's remove this button then
Since we already have a separate issue for that

jakubstec: Sorry! I assumed that if there is no mention about sending request in this issue, that disabled button is okay here 😅

> PR should already implement the request to Ask for admin access

I'm working on it, @ZhenjaHorbach could you review it once I'm done today? 😄

ZhenjaHorbach: Sure!
Will recheck when we add the last requirements!

rayane-d: Yeah, I think we should have the request access functionality implemented before shipping this page. @jakubstec there will also be some shared logic with the "Request access" button on the "Access restricted" screen that we should consider when implementing this, as outlined in issue https://github.com/Expensify/App/issues/99382.

jakubstec:
@codex review

src/libs/actions/Domain.ts
```diff
@@ -475,6 +515,44 @@ function clearSetPrimaryContactError(domainAccountID: number) {
...
+            value: {
+                // eslint-disable-next-line @typescript-eslint/naming-convention
+                domain_adminRequesters: {[currentUserAccountID]: 'read'},
+            },
+        },
+    ];
+
+    const failureData: Array<OnyxUpdate<typeof ONYXKEYS.COLLECTION.DOMAIN>> = [
+        isTransientDomainEntry
+            ? {onyxMethod: Onyx.METHOD.SET, key: domainKey, value: null}
```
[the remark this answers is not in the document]
jakubstec: fixed here https://github.com/Expensify/App/pull/98489/commits/b15b55f5e00974009ff92bd7d60b09c9c13f0511

jakubstec:
hey, I've put this on HOLD, as we want to ensure parity with OD, and for that I think we need to wait for this issue - https://github.com/Expensify/App/issues/97982, so there is one change required (get adminship pending requests Onyx data for domain non-member users in `OpenApp`) 

edit: also, i want to note, that I'll be OoO from Wednesday (25.08) to Sunday (30.08), so I'll be back on Monday to work on it ASAP. @war-in will oversee this PR when I'm away

mountiny:
@jakubstec @war-in @rayane-d I think we dont have to hold and we just go with the case you are member but not admin yet and you can ask for admin access

war-in:
@mountiny as discussed in Slack, if it's safe to request admin access twice, we can remove `HOLD` and merge the PR. Unfortunately, I can't do that because this is Jakub's PR. @mountiny could you rename the PR title? 🙏

ZhenjaHorbach:
@MelvinBot 
Remove the HOLD here

ZhenjaHorbach:
Let me recheck and approve then!

ZhenjaHorbach:
I suppose we have the wrong padding between the error text and the button on Domain exists screen

https://github.com/user-attachments/assets/65b19f7a-506c-4625-8a7e-44b942c808b3

ZhenjaHorbach:
In offline mode, we can press to Ask for admin access
As a result, nothing happens

https://github.com/user-attachments/assets/ac4727ae-ee07-440c-8bcc-b01ac240d09a

src/pages/domain/AddDomainPage.tsx
```diff
@@ -92,18 +128,26 @@ function AddDomainPage() {
...
                         const submitDomain = () => {
                             submittedDomainName.current = domainName;
-                            createDomain(domainName);
+                            domainKeysBeforeCreation.current = isLoadingOnyxValue(allDomainsResult)
+                                ? undefined
+                                : new Set(
+                                      Object.entries(allDomains ?? {})
+                                          .filter(([, domain]) => !!domain?.accountID)
+                                          .map(([domainKey]) => domainKey),
+                                  );
```
ZhenjaHorbach: Can we use something like this?
`new Set(Object.keys(allDomains ?? {}).filter((key) => allDomains?.[key]?.accountID));`

war-in:
@ZhenjaHorbach all comments addressed!

ZhenjaHorbach:
Will recheck soon!

ZhenjaHorbach:
Okay 
Now as for me everything looks good!

ZhenjaHorbach (approved):

ZhenjaHorbach:
But no

ZhenjaHorbach:
The main content shouldn't jump when we have an error
@jakubstec @war-in 
Can you fix this, please?


https://github.com/user-attachments/assets/e7faaac4-5bbd-4480-853e-6fe63de6be4e

shawnborton:
The two similar empty states here doesn't feel the best:
CleanShot 2026-08-31 at 15 30 31@2x

What if we did something like this?

CleanShot 2026-08-31 at 15 39 23@2x


File is here:  [empty-state_domain-exists.svg.zip](https://github.com/user-attachments/files/31647385/empty-state_domain-exists.svg.zip) Dimensions: 180x164

ZhenjaHorbach:
We have small visual glitches in the new image
Or is it okay? 😅
CC: @shawnborton 

Снимок экрана — 2026-08-31 в 15 52 10

mountiny:
> Or is it okay? 😅

silly

shawnborton:
How does it look when you aren't super zoomed in? I think it should render fine at a normal size.

ZhenjaHorbach:
> How does it look when you aren't super zoomed in? I think it should render fine at a normal size.

Without zoom, it's a bit difficult to notice 😁

shawnborton:
Great! This one is hard for us to test - so if you don't mind, could you give us screenshots of the two cases (assuming you are implementing both from the mockup) - thank ya!

jakubstec:
> The main content shouldn't jump when we have an error

@ZhenjaHorbach fixed, could you check whether it's correct?


https://github.com/user-attachments/assets/3f3f7ba3-4741-47dc-a00a-afd6c66db292

@shawnborton new illustration looks like this - is size okay?

Screenshot 2026-08-31 at 16 16 23

zoomed
Screenshot 2026-08-31 at 16 23 46

shawnborton:
Looks good. And are you also doing the pigeon illustration case as well?

ZhenjaHorbach:
I suppose now we have the wrong padding between the error text and the button again 😅
Before we had [the same issue](https://github.com/Expensify/App/pull/98489#issuecomment-5452657422)

> > The main content shouldn't jump when we have an error
> 
> @ZhenjaHorbach fixed, could you check whether it's correct?
> 
>  Screen.Recording.2026-08-31.at.16.30.53.mov 
> @shawnborton new illustration looks like this - is size okay?

shawnborton:
Padding seems okay from what I can tell? Specs:
CleanShot 2026-08-31 at 16 34 26@2x

jakubstec:
> I suppose now we have the wrong padding between the error text and the button again 😅

Oh, my bad! Haven't pulled the changes Marcin has made 😅 Should be all good now! @ZhenjaHorbach 

https://github.com/user-attachments/assets/539897af-9f57-4d4e-920a-ecd536c378c7

ZhenjaHorbach:
> Great! This one is hard for us to test - so if you don't mind, could you give us screenshots of the two cases (assuming you are implementing both from the mockup) - thank ya!

The second case is out of scope for this PR
I think we will update it [here](https://github.com/Expensify/App/issues/99382)

ZhenjaHorbach (approved):

mountiny:
@shawnborton all good from your side as well?

shawnborton:
Yup all good!

jakubstec:
@mountiny so I think we are good to merge it 😄

src/components/ConfirmationPage.tsx
mountiny, line 1: I guess this one would be also good to decompose/ create more specific pages @adhorodyski @staszekscp

rayane-d:
reviewing

src/pages/domain/DomainAlreadyExistsPage.tsx
```diff
@@ -0,0 +1,109 @@
...
+function DomainAlreadyExistsPage({route}: DomainAlreadyExistsPageProps) {
+    const {domainAccountID} = route.params;
+    const {asset: EmptyStateDomainExists} = useMemoizedLazyAsset(() => loadIllustration('EmptyStateDomainExists'));
+    const {translate} = useLocalize();
+    const styles = useThemeStyles();
+    const {isOffline} = useNetwork();
+
+    const [currentUserAccountID] = useOnyx(ONYXKEYS.SESSION, {selector: accountIDSelector});
+    const [hasPendingRequest] = useOnyx(`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`, {selector: hasPendingAdminshipRequestSelector(currentUserAccountID)});
+    // The add domain page drops the entry that came with the failure, so anything left here without an accountID only carries this flow.
```
mountiny, line 38: We should be having empty lines before comments
```suggestion

    // The add domain page drops the entry that came with the failure, so anything left here without an accountID only carries this flow.
```

src/pages/domain/DomainAlreadyExistsPage.tsx
```diff
@@ -0,0 +1,109 @@
...
+    const [isTransientDomainEntry] = useOnyx(`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`, {selector: (domain) => !domain?.accountID});
+    const [isRequestPending] = useOnyx(`${ONYXKEYS.COLLECTION.DOMAIN_PENDING_ACTIONS}${domainAccountID}`, {selector: (pendingActions) => !!pendingActions?.requestAdminship});
+    const [requestError] = useOnyx(`${ONYXKEYS.COLLECTION.DOMAIN_ERRORS}${domainAccountID}`, {selector: (errors) => errors?.requestAdminshipError});
+
+    // Tracks whether the user has submitted a request this visit, so navigating away on success doesn't also fire for a request that was already pending before the page mounted.
+    const [hasSubmittedRequest, setHasSubmittedRequest] = useState(false);
+    // The optimistic pending flag lands a render or two after the submit itself, so we can't treat "not pending yet" as "already settled" until we've actually seen it pending once.
```
mountiny, line 45:
```suggestion

    // The optimistic pending flag lands a render or two after the submit itself, so we can't treat "not pending yet" as "already settled" until we've actually seen it pending once.
```

mountiny (approved):
Nice, looks good to me at this point, just some NABs we dont need to hold on

src/pages/domain/AddDomainPage.tsx
```diff
@@ -70,9 +78,37 @@ function AddDomainPage() {
...
+            return;
+        }
+
+        if (domainKeysBeforeCreation.current.has(`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`)) {
+            setCreateDomainAlreadyHaveAccessError();
+            return;
+        }
+
+        clearDomainFromFailedCreation(domainAccountID, allDomains?.[`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`]?.domain_adminRequesters);
+        Navigation.setNavigationActionToMicrotaskQueue(() => Navigation.navigate(ROUTES.WORKSPACES_DOMAIN_ALREADY_EXISTS.getRoute(domainAccountID), {forceReplace: true}));
```
rayane-d, line 102: Should this branch also clear `form.domainAccountID`/`domainKeysBeforeCreation` before navigating? The other branch above (line 89) calls `clearCreateDomainAccountID()`, but this one doesn't

```suggestion
        clearDomainFromFailedCreation(domainAccountID, allDomains?.[`${ONYXKEYS.COLLECTION.DOMAIN}${domainAccountID}`]?.domain_adminRequesters);
        clearCreateDomainAccountID();
        Navigation.setNavigationActionToMicrotaskQueue(() => Navigation.navigate(ROUTES.WORKSPACES_DOMAIN_ALREADY_EXISTS.getRoute(domainAccountID), {forceReplace: true}));
```

rayane-d (commented):
This looks correct against the BE logic. One question about an effect below.

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100185 was identified to be related to this PR.

jponikarchuk:
This PR failing because of the issue https://github.com/Expensify/App/issues/100185
This issue is reproducible in: All platforms

Issue #97013, opened 2026-07-24
[$250] The generated receipt for mileage expenses is showing the wrong distance and the wrong total amount


m-natarajan:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding. To join our Slack channel, fill out [this form](https://forms.gle/Q7hnhUJPnQCK7Fe56).
___

**Version Number:** 9.4.43-1
**Reproducible in staging?:** Needs Reproduction (Unable to reproduce)
**Reproducible in production?:** Needs Reproduction
**If this was caught during regression testing, add the test name, ID and link from BrowserStack:**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Expensify/Expensify Issue URL:**
**Issue reported by:** @timothysmall
**Slack conversation** (hyperlinked to channel name): [#Expensify Bugs](https://expensify.slack.com/archives/C049HHMV9SM/p1784913552355649)

## Action Performed:

1. Log in to New Expensify using a +@team.expensify.com account.
2. Open a mileage expense 
3. View the receipt:
In Classic: The receipt preview and the enlarged image do not match the mileage and amount of the actual expense.
In ND: The receipt preview matches the expense amount total and mileage, but when clicked to enlarge it shows inaccurate mileage and amount mumbers
## Expected Result:
The receipt should match the expense for mileage and total amount
## Actual Result:
They do not match
## Workaround:
Unknown
## Platforms:
Select the officially supported platforms where the issue was reproduced:
- [ ] Android: App
- [ ] Android: mWeb Chrome
- [ ] iOS: App
- [ ] iOS: mWeb Safari
- [ ] iOS: mWeb Chrome
- [x] Windows: Chrome
- [x] MacOS: Chrome / Safari

## Screenshots/Videos

The video for the issue is in Slack [thread](https://expensify.slack.com/archives/C049HHMV9SM/p1784913552355649)

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
    <ul>
        <li>Upwork Job URL: <a href="https://www.upwork.com/jobs/~022080798705040546748">https://www.upwork.com/jobs/~022080798705040546748</a></li>
[...]

PR #98506, opened 2026-08-12
Render the generated distance receipt on every New Expensify surface

yuwenmemon:
Use PullerBear to assign a reviewer

CC anyone who you think should/could look at these changes

### Fixed Issues

$ https://github.com/Expensify/App/issues/97013

## Details

1. A distance expense carries two receipts. New Expensify builds its own e-receipt in the frontend from the expense, which is why the server stores only the route map as the thumbnail — the card draws the total, the mileage line and the waypoints around it. Beside that sits a generated PDF, added by [Web#48504](https://github.com/Expensify/Web-Expensify/pull/48504) for Expensify Classic, which cannot build a receipt in the frontend. Its design issue, [Expensify#545298](https://github.com/Expensify/Expensify/issues/545298), says it is to be displayed "EXCLUSIVELY in oldDot".
2. `AttachmentView` had started showing that PDF. Its generic PDF branch returns before the distance branch, so once distance receipts became PDFs the enlarged view stopped drawing the card. Every other New Expensify surface still drew it, and the two disagree: the card shows what the expense bills, the PDF shows the routed trip, and a manual distance edit or a commuter exclusion moves them apart. That is the mismatch in the issue.
3. Now the distance branch runs first, so New Expensify draws its own e-receipt everywhere and never falls back to the Classic PDF. Odometer expenses are untouched, since their receipt is the photos the user took.
4. The card lays out at a fixed 375px, so on a full screen it sat small in the middle of the modal, and a card with many waypoints ran off the bottom. `ScaledDistanceEReceipt` measures the space it is given and scales the card to fit, capped at 3x because the map inside it is a 1024px thumbnail.
5. Third fix, same area: the card's map goes stale during a regeneration. It only checked `pendingFields.waypoints`, but editing a distance or a rate also makes the server rebuild the receipt and invalidates the stored URL, so the card would request an image that is no longer there. It now shows the pending map for any of those edits, and for a route error.

Not fixed here, and worth stating: in Expensify Classic the receipt and the expense still disagree after a manual distance edit, because the PDF records the routed trip on purpose ([Web#49853](https://github.com/Expensify/Web-Expensify/pull/49853), for [Expensify#571771](https://github.com/Expensify/Expensify/issues/571771)). This PR stops that reaching New Expensify; whether Classic should keep behaving that way is a separate product question.

## Tests

1. Create a map distance expense on a workspace with distance rates and wait for the receipt to arrive.
2. Hover the receipt in the expense view, hover the same expense's row on the Search page, and expand the receipt. Verify all three show the e-receipt with the same total and mileage as the expense.
3. Edit the distance to a different number and save.
4. Before the new receipt arrives, verify the receipt box draws the route map rather than a stale image, and that expanding shows the e-receipt with a pending map rather than a broken one.
5. Wait for the receipt to regenerate and repeat step 2. Verify the three surfaces still agree with the expense.
6. Change the rate on the same expense and repeat step 2.
7. Open an odometer expense, expand its receipt, and verify it still shows the stitched odometer photos.
8. Open a manual distance expense with no map and verify it is unchanged.
9. Verify the enlarged e-receipt scales to fill the modal rather than sitting small in the middle, and that a route with many waypoints still fits on screen.
10. Resize the window while the receipt is open and verify it rescales without clipping.
11. Open a per diem expense and a normal PDF receipt and verify both still expand as before.

Automated: 410 tests in `tests/unit/TransactionUtilsTest.ts`, plus `tests/ui/components/MoneyRequestReceiptViewTest.tsx`, `tests/ui/IOURequestStepDistanceTest.tsx` and `tests/unit/components/ReceiptPreviewPositionTest.ts`. Typecheck, lint and the React Compiler compliance check pass.

- [x] [PR Author] I made sure this PR correctly handles default values ([SO](https://stackoverflowteams.com/c/expensify/q/20777/2))


src/components/DistanceEReceipt.tsx
```diff
@@ -65,7 +65,9 @@ function DistanceEReceipt({transaction, hoverPreview = false}: DistanceEReceiptP
                     />
 
                     <View style={[styles.moneyRequestViewImage, styles.mh0, styles.mt0, styles.mb5, styles.borderNone]}>
-                        {isFetchingWaypointsFromServer(transaction) || !thumbnailSource ? (
+                        {/* The thumbnail is a page of the stored receipt file, therefore an edit that makes the server
+                            build a new file also invalidates it. Draw the pending map until the new file arrives. */}
+                        {hasPendingDistanceReceiptRegeneration(transaction) || !thumbnailSource ? (
```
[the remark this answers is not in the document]
yuwenmemon: (Yuwen's Agent) Good catch — fixed in eef9971. The card now draws the pending map whenever the stored file cannot be trusted, which is `hasPendingDistanceReceiptRegeneration || hasDistanceRouteErrors`, so the error case no longer loads the stale receipt page while the report tile shows `ConfirmedRoute`.

garrettmknight (dismissed):

src/libs/TransactionUtils/index.ts
```diff
@@ -1282,6 +1282,41 @@ function hasPendingDistanceReceiptRegeneration(transaction: OnyxInputOrEntry<Tra
     return hasPendingRegenerationField;
 }
 
+/**
+ * Whether the route of a distance expense failed. The stored map receipt is then missing or wrong, so the
+ * surfaces draw the route locally instead.
+ */
+function hasDistanceRouteErrors(transaction: OnyxInputOrEntry<Transaction>): boolean {
+    return !isEmptyObject(transaction?.errors) || !isEmptyObject(transaction?.errorFields?.route) || !isEmptyObject(transaction?.errorFields?.waypoints);
```
Valforte: Isn't `transaction.errors` written by other errors too? That would mean this would return `true` for any error, not only route errors.

src/pages/media/AttachmentModalScreen/routes/TransactionReceiptModalContent.tsx
```diff
@@ -292,7 +293,9 @@ function TransactionReceiptModalContent({navigation, route}: AttachmentModalScre
         draftTransactionID,
     });
 
-    const allowDownload = !isEReceipt;
+    // A distance expense that falls back to the local card has no receipt file to hand over, or has one that the
+    // server is rebuilding. Downloading it would save a file that does not match what is on screen.
+    const allowDownload = !isEReceipt && !shouldRenderLocalDistanceEReceipt(transaction);
```
Valforte: 🤖 **Claude Code review** (run by @Valforte)

**Blocker:** this disables Download for distance receipts that do have a real, matching file.

`shouldRenderLocalDistanceEReceipt` returns `true` for *any* stored receipt that isn't a PDF. That includes the legacy map image and any file the user attached through Replace receipt (`shouldShowReplaceReceiptButton` is true for non-odometer distance expenses). In both cases the file exists, and `DistanceEReceipt` renders that very file as the card's map — so the download would match what's on screen, but it is no longer offered.

Suggested fix: gate download on the two cases where there is genuinely nothing to hand over:
```ts
const allowDownload = !isEReceipt && !(!hasReceiptSource(transaction) || hasPendingDistanceReceiptRegeneration(transaction));
```

src/components/ReportActionItem/MoneyRequestReceiptView.tsx
```diff
@@ -185,13 +186,12 @@ function MoneyRequestReceiptView({
...
-    // While the receipt is regenerating (e.g. after an offline waypoint edit) the stored map is stale and can't be
-    // redrawn locally, so don't surface the e-receipt overlay — the receipt box already shows the pending map.
-    const canShowDistanceEReceipt = isMapDistanceRequest && !isPendingReceiptRegeneration;
+    // The overlay draws the distance e-receipt card from the expense, therefore it must appear only where the
+    // receipt box has no generated file to show. Over a generated file it would state a total and a mileage that
+    // the file itself can contradict, which is the whole of https://github.com/Expensify/App/issues/97013.
+    const canShowDistanceEReceipt = shouldRenderLocalDistanceEReceipt(displayedTransaction);
     // The Expand button opens the full-screen receipt on the stored map. While regeneration is pending that map is
     // stale and can't be redrawn locally, so disable Expand for map distance requests until the refreshed receipt arrives.
     const shouldDisableExpandReceipt = isMapDistanceRequest && isPendingReceiptRegeneration;
```
Valforte, line 197: 🤖 **Claude Code review** (run by @Valforte)

**Test steps won't reproduce.** Step 4 says "hovering shows the card with a pending map rather than a stale image", but that state is unreachable.

`HoveredDistanceEReceipt` renders only when `isHovering` comes from `ReceiptHoverZoom`, and `canHoverZoomReceipt = canZoomReceipt && !isPendingReceiptRegeneration`. With regeneration pending, `isEnabled` is false, so `isActive` is false in `useReceiptHoverZoom` and `isHovering` never becomes `true` — no card on hover.

The same step's "expanding shows the card" also conflicts with `shouldDisableExpandReceipt` on this line, which still disables Expand for exactly that state.

Either wire hover/expand for the local-card fallback, or correct the QA steps. Also worth noting this file now has two definitions of "map distance": `isMapBasedDistanceRequest` for Expand, and `shouldRenderLocalDistanceEReceipt` for the card. Since the new `AttachmentView` branch already handles pending regeneration safely, Expand could probably stay enabled.

Valforte (changes requested):

src/pages/media/AttachmentModalScreen/routes/TransactionReceiptModalContent.tsx
yuwenmemon: (Yuwen's Agent) Agreed, fixed in eef9c94.

Gating on `shouldRenderLocalDistanceEReceipt` was too broad: for a legacy image receipt the card draws that very file, so the download would have matched the screen.

I went slightly narrower than the suggestion, because a route error should also withhold it — the stored file then describes a different trip. The shared condition is now `hasUsableStoredDistanceReceipt` (has a source, not regenerating, no route error), which is the single answer to "is the stored file showable" and is used by both the download gate and the card's map slot:

```ts
const allowDownload = !isEReceipt && (!shouldRenderLocalDistanceEReceipt(transaction) || hasUsableStoredDistanceReceipt(transaction));
```

Legacy image receipts stay downloadable; only the no-file, regenerating and route-error cases withhold it.

src/components/ReportActionItem/MoneyRequestReceiptView.tsx
yuwenmemon: (Yuwen's Agent) Correct on both counts — the QA steps were wrong, rewritten in the description.

`isEnabled={canHoverZoomReceipt}` is `canZoomReceipt && !isPendingReceiptRegeneration`, so `isActive` is false and the overlay cannot render during regeneration, and `shouldDisableExpandReceipt` still greys out Expand in that same window. Step 4 now says both are disabled then, which is the pre-existing behaviour and is unchanged by this PR.

On the substance: `canShowDistanceEReceipt` being true during regeneration is unreachable rather than wrong, so the change still holds for the cases where the overlay does render.

On enabling Expand — I agree it now looks safe, since the new `AttachmentView` branch renders the card for exactly that state instead of loading the invalidated URL. I left it out because it is a behaviour change beyond the bug this PR fixes and would want its own QA pass. Happy to fold it in if you would rather it ship together.

On the two definitions of "map distance": they genuinely differ. `isMapBasedDistanceRequest` includes `distance-manual` transactions that still carry waypoints, whereas the new predicate matches the set `AttachmentView` already gated on, so collapsing them would change which expenses get the card. If Expand moves onto the new predicate, that difference goes away.

yuwenmemon:
(Yuwen's Agent) Heads up — this PR now does the **opposite** of what it did before, and the description has been rewritten. Reviews above refer to the earlier approach.

It previously made New Expensify render the server-generated PDF everywhere, on the reasoning that Expensify Classic shows that file so New Expensify should match. That was wrong. [Web#48504](https://github.com/Expensify/Web-Expensify/pull/48504) keeps the receipt thumbnail a bare map on purpose, and its design issue [Expensify#545298](https://github.com/Expensify/Expensify/issues/545298) is explicit about why:

> Currently, we just save a map for distance receipts. This works well in newdot, as we build the e-receipt in the frontend and display all the info. […] The idea is to display this EXCLUSIVELY in oldDot.

So the PDF is a Classic-only artifact, and `DistanceEReceipt` is the intended New Expensify rendering rather than a legacy fallback. The fix is therefore to stop New Expensify showing the PDF at all, which is what the current diff does — `AttachmentView`'s distance branch now runs before its PDF branch.

Net effect on the earlier review points:

- @ahmedGaber93's two reports are resolved by the reversal. Nothing shows a bare map thumbnail any more, because the card is drawn on every surface again.
- @Valforte's three points no longer apply: the download gate, the `MoneyRequestReceiptView` overlay predicate and the `transaction.errors` widening are all reverted. The narrowed `hasDistanceRouteErrors` survives, used only by the card's pending-map guard.
- The Codex finding survives too, as that guard.

A Web-Expensify PR that would have changed the thumbnail has been closed for the same reason: https://github.com/Expensify/Web-Expensify/pull/55423

The diff is down to 4 files. Sorry for the churn on your reviews.

yuwenmemon:
@ahmedGaber93 bump for review!

Valforte (dismissed):
LGTM

ahmedGaber93:
The receipt thumbnail is not clickable and lacks the enlarge button.

@yuwenmemon Do you have the same bug?



https://github.com/user-attachments/assets/1619ca07-eea4-4bc4-bcd7-0c5a27ff00cd

yuwenmemon:
Thanks for checking. Hmm. I don't think this one is from the PR. Can you check if it's maybe from your local environment? 

The receipt box renders `ConfirmedRoute` (the live map) rather than the receipt thumbnail when `showMapAsImage` is true. In other words, the transaction has errors or a pending receipt regeneration. That branch in `ReportActionItemImage` returns early without the pressable wrapper and without `onLoad`, so `isLoading` never clears and `canShowReceiptActions` stays false. That's precisely "not clickable, no enlarge button", and it's on main too. This PR doesn't touch either file.

Local dev generates the distance receipt through PDFBox, which usually isn't reachable there, so the receipt fails to generate and leaves the transaction in that state. Your chat preview also shows the live map, which points the same way.

Could you compare `main` against this branch on the same environment and the same expense?

ahmedGaber93:
@yuwenmemon Ah, the issue was on my local, sorry about that!

I retested and it's working. One note though: the enlarged view now returns early with `DistanceEReceipt`, which doesn't really fit the enlarge mode.

Do we okay with that?



https://github.com/user-attachments/assets/c5e206b9-846f-469a-be1b-720a69f5592c

ahmedGaber93:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>


https://github.com/user-attachments/assets/d37c845e-7a45-42ed-b59d-0b0df8c1228d



</details>

<details>


https://github.com/user-attachments/assets/66acbef0-34a2-40b7-814a-ab68ecbfa1f1



</details>

<details>



</details>

<details>


https://github.com/user-attachments/assets/c20bbed3-dae6-4a3c-b757-835d6f396041



</details>

<details>

https://github.com/user-attachments/assets/a9c4bb2e-2386-4d94-b634-4de6a3abfc32




</details>

yuwenmemon:
Hmm, I would say not. Let me fix that.

ahmedGaber93:
@yuwenmemon, it looks better, but it’s still quite far from matching the size of the PDF. Are we okay with that?

https://github.com/user-attachments/assets/84f1977e-6fe2-46f2-91f8-f8185fe589f9


https://github.com/user-attachments/assets/11b24cb8-8b45-4658-bed8-0649194c5b8a

Valforte (approved):

ahmedGaber93 (approved):

mitarachim:
Deploy Blocker https://github.com/Expensify/App/issues/99376 was identified to be related to this PR.

Issue #98254, opened 2026-08-11
perf(navigation): deprioritize covered screens with React Activity instead of react-freeze


mountiny:
## Background

In our stack navigation, every screen below the top one stays fully mounted. This is what keeps its state and scroll position, and what keeps it painted behind the RHP overlay or during an iOS swipe-back gesture. It also means that a covered screen re-renders on every Onyx broadcast, exactly like the screen the user is looking at. The heaviest screens we have, the Inbox tab and the Spend tab, are also the ones that routinely sit covered, because every drill-down into a report, an expense or an editor opens as an RHP or a modal on top of them.

We tried to stop that rendering once before, with `react-freeze` ([proposal](https://github.com/Expensify/App/issues/33725#issuecomment-3768435794)). That attempt failed and had to be rolled back.

Since then, two efforts have shown a safer path. In July, together with Jakub Magiera, we shipped React `<Activity>` on the MoneyRequest/Share modal stacks (https://github.com/Expensify/App/pull/96485), and it has been stable on main since. In parallel, Lukasz Modzelewski proved the same idea inside the react-navigation 8 alpha (https://github.com/Expensify/App/issues/97081, PoC: https://github.com/Expensify/App/pull/97477). We are synced with each other; because the final shape of the RN8 API and its release date are unknown, this proposal does not wait for it.

## Problem

When a thread or a side panel is open over a large report or the search table, sending a message, saving a field, or opening another panel briefly freezes the app. Clicks give no visual feedback, the response delay can exceed half a second, and the lag scales with both the amount of data and the number of screens open in the background.

## Solution

Wrap covered screens in React `<Activity mode="hidden">`, so their render work leaves the critical path of the interaction. Core PR: https://github.com/Expensify/App/pull/97990. The PR currently wraps almost every screen in the app, 837 of them, but that is deliberate and **for demo purposes only**, so the mechanism and the numbers can be evaluated at full scale. The real rollout will activate Activity **screen by screen**, as described in the migration strategy below.

**This is not a second run of `react-freeze`.** Freeze suspended the whole subtree, so a frozen screen painted nothing and could not process a modal dismiss, which is where its blockers came from; a hidden Activity keeps paint and state, processes updates at background priority and runs real effect cleanups, so those blockers do not apply. The full comparison is below.

[...]

PR #97990, opened 2026-08-06
perf(navigation): deprioritize covered screens with React Activity (core)

dariusz-biela:
### Explanation of Change

This PR replaces react-freeze with React `<Activity>` for covered (non-top) screens. A hidden Activity keeps its state, cleans up its effects and processes updates at background priority, which removes the freeze regression classes (modal-race hard freezes, stale side effects) by construction. It is the core of https://github.com/Expensify/App/pull/97666, extracted on its own so the mechanism can be reviewed without the follow-up work around it, and it generalizes the pattern from https://github.com/Expensify/App/pull/96485 from a single selector to every covered screen in the opted-in navigators.

**In the app**

- Covered screens are deprioritized while they are covered, which makes navigation over a heavy screen (open an RHP over the Reports list, send a message in a thread over a heavy report) noticeably cheaper.
- Covered screens stay painted, so the screen behind the RHP overlay and the underlay of an iOS swipe-back gesture look exactly as they do today.
- Covered screens keep their state and their scroll position, and they are taken out of the accessibility tree, the tab order and the reach of the pointer while their content is stale.

**In the code**

- `nonTopScreenBehavior: 'none' | 'freeze' | 'activity'` replaces the `freezeNonTopScreens` navigator flag. It is a normal navigation option, so a navigator sets the behavior for all of its screens and a single screen can override it. Freeze stays available as a fallback, although nothing selects it after this PR.
- `wrapDescriptorsWithNonTopScreensBehavior` replaces `wrapDescriptorsWithFreeze` and picks the wrapper per screen from that option. Persistent screens (web sidebars) are never wrapped.
- `ScreenActivityWrapper` renders the screen inside `<Activity>` and decides its mode. Hiding is immediate, revealing waits for the navigation transition to end so a pop commits cheaply and animates undisturbed. A screen also stays visible for its first render and for the duration of a window width change, because a hidden Activity runs no effects and could otherwise neither fetch its data nor lay itself out against the new size.
- `CustomViewWrapper` neutralizes on both platforms the `display: none` a hidden Activity commits, and carries the accessibility and pointer flags a hidden subtree can no longer set from an effect. The navigator's card visibility, not Activity, decides what is on screen.
- Rolled out per navigator to the modal stacks, RightModal, SearchFullscreen, Workspace and Split. The root stack keeps its current behavior, and Split is the only navigator whose behavior changes rather than being added, because it was the one navigator freezing so far.

**Wide RHP width across a hide**

[...]

PR #98575, opened 2026-08-13
perf(navigation): add per-screen Activity behavior for covered screens

dariusz-biela:
### Explanation of Change

This PR ships the core mechanism of the accepted proposal in https://github.com/Expensify/App/issues/98254: a per-screen navigation option that decides what happens to a screen while another screen covers it, with React `<Activity>` as the new behavior and `StrictMode` as the qualification gate for every screen that will migrate to it. The mechanism is the one evaluated at full scale in the demo PR https://github.com/Expensify/App/pull/97990, but here it ships disabled, so screens can opt in one by one as the rollout progresses.

**In the app**

- No screen opts into Activity yet, and the split navigators keep freezing exactly the same screens as they do today, only now through the new option.
- On native, a frozen screen now renders one extra layout node inside its wrapper. On a phone that node sits on the split navigator sidebar while a central screen covers it, so layout, touch handling and the swipe-back underlay are the places worth checking. The web freeze path is untouched.
- One screen changes: the money request participant selector, the only place that already ships `<Activity>`. On web its content now stays painted while hidden, which is what native already did, and its wrapper node stops taking part in layout on both platforms.

**In the code**

- `nonTopScreenBehavior: 'none' | 'freeze' | 'activity'` replaces the `freezeNonTopScreens` navigator flag. It is a normal navigation option, so a navigator sets a default and a single screen can override it. The split navigators select `'freeze'`; everything else stays at `'none'`.
- `wrapDescriptorsWithNonTopScreensBehavior` replaces `wrapDescriptorsWithFreeze` and picks the wrapper per screen (freeze or Activity) from that option.
- `ScreenActivityWrapper` renders a screen inside `<Activity>` and manages its visible/hidden mode (see the internals section below).
- `AlwaysPaintedView` (renamed from `DisplayContentsView`) becomes a real `display: contents` host with an `inert` prop on both platforms.
- Screens that pick `'activity'` are additionally wrapped in `StrictMode` in dev, the qualification gate from the rollout plan.
- Two new guides document the mechanism: `contributingGuides/ACTIVITY_SCREENS.md` explains the Activity lifecycle model (hide + reveal = full effect unmount + remount with surviving state), the classes of regressions unsafe effects cause and the checklist for opting a screen in, and `contributingGuides/STRICT_MODE.md` covers the StrictMode gate.
- Unit test suites cover each of the pieces above.

**ScreenActivityWrapper internals**

The wrapper composes a few small mechanisms, each in its own module:

- `useScreenActivityState` derives the Activity mode. A screen counts as covered when another screen of its own navigator sits on top of it or when the whole navigator lost focus to a route higher in the tree. The covered flag drives `inert` with no delay, while the mode itself is allowed to lag on the reveal.
- First render: a screen that mounts while already covered still renders its first frame as visible, because React never mounts the effects of a hidden Activity, so a screen born hidden would never fetch its data. The window closes on the next animation frame, with a timeout fallback for backgrounded apps and hidden browser tabs where `requestAnimationFrame` never fires.
- `useDeferVisibleUntilFocusTransitionEnd` delays the reveal until the navigation transition ends, because revealing in the same commit as the navigation update blocked the main thread for hundreds of milliseconds on a pop. Hiding is never delayed.
- `useIsWindowSizeChanging` and `windowSizeChangeStore` keep a covered screen visible while the window width is changing, so it lays itself out against the new size before it is revealed. One shared `Dimensions` listener serves every screen of every stack, and only width changes count, so the soft keyboard does not remount the effects of every hidden screen.
- `AlwaysPaintedView` is the `display: contents` host that keeps the hidden content painted on both platforms, and its `inert` prop takes the covered content out of touch handling and the accessibility tree.
- `StrictModeMountGate` commits `StrictMode` one commit ahead of the screen content, because React only double-invokes effects for content that mounts below an already committed `StrictMode`. Production builds render a passthrough, and `USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV` opts a local profiling session out of the gate.

**Deliberately out of scope**

- No screen migrations. Each screen opts in via its own PR once it runs clean under `StrictMode`.
- The wide RHP fix for registrations dropped while a covered screen is hidden stays in the demo PR and will land with the RHP migration that needs it.

<details>
<summary>A patch to launch the Activity throughout the entire app. (Note: The app hasn't been fixed yet, so there may be quite a few bugs.)</summary>

```patch
Index: src/libs/Navigation/AppNavigator/createRightModalNavigator/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/libs/Navigation/AppNavigator/createRightModalNavigator/index.tsx b/src/libs/Navigation/AppNavigator/createRightModalNavigator/index.tsx
--- a/src/libs/Navigation/AppNavigator/createRightModalNavigator/index.tsx	(revision 9622d85f2163f0bf6ab96659596a6062022f6f08)
+++ b/src/libs/Navigation/AppNavigator/createRightModalNavigator/index.tsx	(date 1787136784869)
@@ -20,7 +20,8 @@
 
 const RightModalNavigatorComponent = createPlatformStackNavigatorComponent(NAVIGATORS.RIGHT_MODAL_NAVIGATOR, {
     createRouter: RightModalRouter,
-    defaultScreenOptions: defaultPlatformStackScreenOptions,
+    // TEMPORARY: whole-navigator Activity opt-in, ported from https://github.com/Expensify/App/pull/97990 for manual verification only.
+    defaultScreenOptions: {...defaultPlatformStackScreenOptions, nonTopScreenBehavior: 'activity'},
     Effects: RightModalNavigatorEffects,
 });
 
Index: src/libs/Navigation/AppNavigator/createSearchFullscreenNavigator/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/libs/Navigation/AppNavigator/createSearchFullscreenNavigator/index.tsx b/src/libs/Navigation/AppNavigator/createSearchFullscreenNavigator/index.tsx
--- a/src/libs/Navigation/AppNavigator/createSearchFullscreenNavigator/index.tsx	(revision 9622d85f2163f0bf6ab96659596a6062022f6f08)
+++ b/src/libs/Navigation/AppNavigator/createSearchFullscreenNavigator/index.tsx	(date 1787136784870)
@@ -25,7 +25,8 @@
 
 const SearchFullscreenNavigatorComponent = createPlatformStackNavigatorComponent('SearchFullscreenNavigator', {
     createRouter: addPushParamsRouterExtension(SearchFullscreenRouter),
-    defaultScreenOptions: defaultPlatformStackScreenOptions,
+    // TEMPORARY: whole-navigator Activity opt-in, ported from https://github.com/Expensify/App/pull/97990 for manual verification only.
+    defaultScreenOptions: {...defaultPlatformStackScreenOptions, nonTopScreenBehavior: 'activity'},
     Effects: SearchFullscreenNavigatorEffects,
     getCustomState,
     ExtraContent: SearchSidebar,
Index: src/libs/Navigation/AppNavigator/ModalStackNavigators/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/libs/Navigation/AppNavigator/ModalStackNavigators/index.tsx b/src/libs/Navigation/AppNavigator/ModalStackNavigators/index.tsx
--- a/src/libs/Navigation/AppNavigator/ModalStackNavigators/index.tsx	(revision 9622d85f2163f0bf6ab96659596a6062022f6f08)
+++ b/src/libs/Navigation/AppNavigator/ModalStackNavigators/index.tsx	(date 1787136784869)
@@ -152,7 +152,8 @@
                 aria-modal={isSmallScreenWidth || undefined}
                 role={isSmallScreenWidth ? 'dialog' : undefined}
             >
-                <ModalStackNavigator.Navigator>
+                {/* TEMPORARY: whole-navigator Activity opt-in, ported from https://github.com/Expensify/App/pull/97990 for manual verification only. */}
+                <ModalStackNavigator.Navigator screenOptions={{nonTopScreenBehavior: 'activity'}}>
                     {Object.keys(screens as Required<Screens>).map((name) => (
                         <ModalStackNavigator.Screen
                             key={name}
Index: src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx b/src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx
--- a/src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx	(revision 9622d85f2163f0bf6ab96659596a6062022f6f08)
+++ b/src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx	(date 1787136784870)
@@ -41,9 +41,8 @@
 const SplitNavigatorComponent = createPlatformStackNavigatorComponent('SplitNavigator', {
     createRouter: SplitRouter,
     Effects: SplitNavigatorEffects,
-    // Covered split screens stay frozen, as they were before this option existed. Their migration to 'activity'
-    // is a per-screen decision left to the rollout in https://github.com/Expensify/App/issues/98254.
-    defaultScreenOptions: {...defaultPlatformStackScreenOptions, nonTopScreenBehavior: 'freeze'},
+    // TEMPORARY: whole-navigator Activity opt-in, ported from https://github.com/Expensify/App/pull/97990 for manual verification only.
+    defaultScreenOptions: {...defaultPlatformStackScreenOptions, nonTopScreenBehavior: 'activity'},
     getCustomState: getCustomSplitNavigatorState,
     NavigationContentWrapper: SidebarSpacerWrapper,
 });
Index: src/libs/Navigation/AppNavigator/createWorkspaceNavigator/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/libs/Navigation/AppNavigator/createWorkspaceNavigator/index.tsx b/src/libs/Navigation/AppNavigator/createWorkspaceNavigator/index.tsx
--- a/src/libs/Navigation/AppNavigator/createWorkspaceNavigator/index.tsx	(revision 9622d85f2163f0bf6ab96659596a6062022f6f08)
+++ b/src/libs/Navigation/AppNavigator/createWorkspaceNavigator/index.tsx	(date 1787136784870)
@@ -22,7 +22,8 @@
 
 const WorkspaceNavigatorComponent = createPlatformStackNavigatorComponent(NAVIGATORS.WORKSPACE_NAVIGATOR, {
     createRouter: WorkspaceRouter,
-    defaultScreenOptions: defaultPlatformStackScreenOptions,
+    // TEMPORARY: whole-navigator Activity opt-in, ported from https://github.com/Expensify/App/pull/97990 for manual verification only.
+    defaultScreenOptions: {...defaultPlatformStackScreenOptions, nonTopScreenBehavior: 'activity'},
     Effects: WorkspaceNavigatorEffects,
 });
```
 </details>

### Fixed Issues

$ https://github.com/Expensify/App/issues/98254
PROPOSAL: https://github.com/Expensify/App/issues/98254

### Tests

Two things change at runtime. On native, a covered screen inside a split navigator (Account, Inbox, Workspaces) renders one extra layout node while it is covered; the web freeze path is untouched. On all platforms the money request participant picker swaps that same wrapper and gains `inert`. The set of screens that freeze is unchanged, so everything else should behave as it does on `main`.

1. **Covered screen stays painted (iOS).** Open Account, scroll the list down, then open Preferences. Drag slowly from the left edge and hold halfway. Verify that the Account list behind the gesture is fully drawn and still scrolled where you left it, never blank or white. Release to cancel, then repeat and complete the gesture, and verify both land on the expected screen.
2. **Layout and scroll after uncovering (iOS, Android).** With the Account list scrolled down, open Preferences, then go back. Verify that Account kept its scroll position, that the rows and the bottom tab bar sit where they did before, and that nothing is clipped or collapsed.
3. **Touch after uncovering (iOS, Android).** On that Account list, tap a row and verify that it navigates.
4. **Participant picker (web, iOS, Android).** Tap the green + button, choose Create expense, switch to the Manual tab, enter an amount and tap Next to reach "Choose recipient" (dismiss the contacts import prompt if it appears). Type a few letters and verify that the list filters, that it is interactive, and that the picker is laid out with no clipping and no collapsed height. Pick a recipient to land on "Confirm details" and verify that no part of the picker shows through. Go back and verify that the picker is laid out with no clipping and no collapsed height, that typing filters the list again, and that neither direction flashes blank. The search field is empty after going back, on this branch and on `main` alike, so it is not part of what this step verifies.
5. **Picker is unreachable while covered (web).** On "Confirm details" press Tab repeatedly and verify that focus never enters the picker behind it, then go back and verify that Tab reaches it again. The picker now stays painted while covered instead of being removed with `display: none`, so `inert` is the only thing keeping it out of the tab order.

Steps 1 to 3 use Account > Preferences on purpose, because both are screens of the same split navigator. Account > Profile > Display name does not exercise this change: Display name is an RHP screen, so Profile never becomes a covered screen of the split navigator.

### Offline tests

N/A - this PR only changes how covered screens are wrapped for rendering and makes no API, Onyx or network changes.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

**Covered screen, steps 2 and 3.** Account is scrolled down, Preferences opens over it, then back. Account
keeps its scroll position, the rows and the tab bar sit where they did, nothing is clipped or collapsed, and
tapping a row still navigates. Step 1 is not shown here: Android renders no predictive-back preview for this
transition, so there is no mid-gesture underlay to look at.

https://github.com/user-attachments/assets/0e008cc3-7512-4fbf-a4ef-163c4fd8437c

**Participant picker, step 4.** Amount, Next to "Choose recipient", search text typed, recipient picked,
"Confirm details", then back. The round trip is done twice. The picker returns with its search text and its
selection, the list stays interactive, and neither direction flashes blank.

https://github.com/user-attachments/assets/2d90911f-f49e-4090-942a-9c77802de757

</details>

<details>

**Covered screen, steps 1 to 3.** Account is scrolled to the bottom and Preferences opens over it. The edge
pan is held at mid-screen: the Account list behind the gesture is fully drawn and still scrolled where it was
left, never blank or white. The gesture is released back to the edge to cancel, then repeated and completed.
Back on Account the scroll position, the rows and the tab bar are unchanged, and tapping Security navigates.

https://github.com/user-attachments/assets/74ceb6d2-fa85-4d8c-9fa4-91fd0e6d2fe5

**Participant picker, step 4.** Same round trip as on Android, twice in both directions.

https://github.com/user-attachments/assets/b2e9895b-64f7-4c5d-87de-791f660d7b78

</details>

<details>

**Participant picker and tab order, steps 4 and 5.** The picker round trip runs first, then on "Confirm
details" ten Tab presses cycle inside the RHP and focus never lands on the picker behind it; after going back,
Tab reaches the picker again. This is the check that matters most on web: the picker now stays painted while
covered instead of being removed with `display: none`, so `inert` is the only thing keeping it out of the tab
order.

https://github.com/user-attachments/assets/f1f90ace-5fa4-4a14-8906-f7ad714eba52

</details>


src/pages/iou/request/MoneyRequestParticipantsSelector.tsx
```diff
@@ -105,7 +103,7 @@ function MoneyRequestParticipantsSelector({
 
     return (
         <Activity mode={isActivityVisible ? 'visible' : 'hidden'}>
-            <DisplayContentsView style={styles.flex1}>
+            <DisplayContentsView>
```
JakubKorytko: This DisplayContentsView is used without an inert prop, unlike ScreenActivityWrapper which always passes inert={isScreenCovered}. Before this PR, hiding here relied on React writing display:none, which also removed the content from the accessibility tree and tab order for free. Now DisplayContentsView refuses that write on web and keeps the content painted, but without inert the hidden participant list stays focusable and exposed to screen readers while Activity mode is 'hidden'. Consider passing inert={!isActivityVisible} here so this screen gets the same accessibility behavior ScreenActivityWrapper gives every other screen.

JakubKorytko: Separately, this call site dropped useThemeStyles/styles.flex1 when DisplayContentsView's style prop was replaced by inert. On native, flex1 is now only applied to the inner View that gets rendered when inert is passed, and since this call site never passes inert, the participant list loses its flex:1 sizing on native. On web it is now a plain div with no layout style at all. Worth a quick visual check on both platforms to confirm layout is unaffected, since nothing in this diff tests that.

src/components/DisplayContentsView/index.tsx
```diff
@@ -1,16 +1,62 @@
...
- * native implementation relies on internal React Native APIs.
+ * It is a plain div because react-native-web's View neither accepts `display: contents` nor declares `inert`. The
+ * attribute goes on that same div, so `inert` needs no extra node here, unlike on native.
  */
-function DisplayContentsView({children, style}: DisplayContentsViewProps) {
-    return <View style={style}>{children}</View>;
+function DisplayContentsView({inert, children}: DisplayContentsViewProps) {
+    const pinnedElementRef = useRef<HTMLDivElement | null>(null);
+
+    const attachDisplayContentsEnforcer = (element: HTMLDivElement | null) => {
```
JakubKorytko: attachDisplayContentsEnforcer is a new function on every render, so its identity as the ref callback changes every render too. React detaches and reattaches a ref whose callback identity changed, even though the pinnedElementRef check makes the reattach a no-op functionally. Wrapping it in useCallback (no deps needed, it only touches a ref) would keep the ref stable and skip that churn. Worth checking whether React Compiler already stabilizes this before changing it.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import ScreenActivityWrapper from './ScreenActivityWrapper';
+import ScreenFreezeWrapper from './ScreenFreezeWrapper';
+
+type Descriptor = {
+    /** Route object containing the screen name, used to check if the screen is persistent */
+    route: {name: string};
+
+    /** Resolved options of the screen, carrying the behavior it picked for the time it is covered by another one */
+    options: PlatformSpecificNavigationOptions & {nonTopScreenBehavior?: NonTopScreenBehavior};
```
JakubKorytko: nonTopScreenBehavior is already a field on the merged PlatformStackNavigationOptions type in NavigationOptions.ts, so intersecting it onto PlatformSpecificNavigationOptions here re-declares the same field in a second place. If the real descriptor.options type can be typed as PlatformStackNavigationOptions directly, dropping this intersection would avoid the two types drifting apart later.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/index.tsx
```diff
@@ -105,7 +103,7 @@ function PlatformNavigatorImpl<RouterOptions extends PlatformStackRouterOptions
         }),
     };
 
-    const wrappedDescriptors = freezeNonTopScreens ? wrapDescriptorsWithFreeze(descriptors, state, persistentScreens) : descriptors;
+    const wrappedDescriptors = wrapDescriptorsWithNonTopScreensBehavior(descriptors, state, persistentScreens);
```
JakubKorytko: Migration from the removed opt-in freezeNonTopScreens flag to always calling wrapDescriptorsWithNonTopScreensBehavior (gated per screen by nonTopScreenBehavior) has no test at this integration point, here or in index.native.tsx. Only the wrapping function itself is unit tested with hand built descriptors. A test here would catch a future navigator accidentally inheriting wrapping through defaultScreenOptions.

src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx
```diff
@@ -41,10 +41,11 @@ function getCustomSplitNavigatorState({state, shouldUseNarrowLayout}: CustomStat
 const SplitNavigatorComponent = createPlatformStackNavigatorComponent('SplitNavigator', {
     createRouter: SplitRouter,
     Effects: SplitNavigatorEffects,
-    defaultScreenOptions: defaultPlatformStackScreenOptions,
+    // Covered split screens stay frozen, as they were before this option existed. Their migration to 'activity'
+    // is a per-screen decision left to the rollout in https://github.com/Expensify/App/issues/98254.
+    defaultScreenOptions: {...defaultPlatformStackScreenOptions, nonTopScreenBehavior: 'freeze'},
```
JakubKorytko: No test mounts SplitNavigator (or an equivalent minimal stack) to confirm this defaultScreenOptions value actually reaches each screen's resolved options.nonTopScreenBehavior and is picked up correctly, nor is a per-screen override of this default tested end to end.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/StrictModeMountGate.tsx
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Production builds get the plain content because StrictMode does nothing there, and profiling sessions opt out of the
+ * gate through USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV to keep double renders out of the measurements.
+ */
+function StrictModeMountGatePassThrough({children}: StrictModeMountGateProps) {
+    return children;
+}
+
+export {StrictModeMountGate};
+export default __DEV__ && CONFIG.USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV ? StrictModeMountGate : StrictModeMountGatePassThrough;
```
JakubKorytko: StrictModeMountGateTest.tsx only imports the named StrictModeMountGate export directly. This default export's branch, which is what ScreenActivityWrapper actually consumes, is never exercised. There is no test asserting that CONFIG.USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV=false picks StrictModeMountGatePassThrough, or that production (__DEV__ false) does the same.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
```diff
@@ -0,0 +1,66 @@
...
+ * waits for the navigation transition to end, because revealing in the same commit as the navigation update
+ * blocked the main thread for hundreds of milliseconds on a pop.
+ */
+function useScreenActivityState(isScreenBlurred: boolean): ScreenActivityState {
+    const isFocused = useIsFocused();
+    const isWindowSizeChanging = useIsWindowSizeChanging();
+    const [hasCompletedFirstRender, setHasCompletedFirstRender] = useState(false);
+
+    useEffect(() => {
+        const rafID = requestAnimationFrame(() => setHasCompletedFirstRender(true));
```
JakubKorytko: No test unmounts the hook/wrapper before the raf or the 100ms fallback timeout fires (e.g. user backs out of a screen still on its first render). The cleanup itself (cancelAnimationFrame/clearTimeout being called) is never asserted.

tests/unit/Navigation/nonTopScreenBehavior/wrapDescriptorsWithNonTopScreensBehaviorTest.tsx
```diff
@@ -0,0 +1,91 @@
...
+        preloadedRoutes: [],
+    };
+}
+
+function renderWrapped(descriptor: {render: () => React.JSX.Element}): ReactElement<NonTopScreenWrapperProps> {
+    return descriptor.render();
+}
+
+describe('wrapDescriptorsWithNonTopScreensBehavior', () => {
+    it('leaves a screen that picked no behavior untouched', () => {
```
JakubKorytko: No test covers state.routes[state.index] being undefined (e.g. an empty routes array right after a deep link replaces the stack, or a stale state snapshot mid-navigation). topRouteKey would be undefined, so every route including the intended top one would evaluate as blurred/covered. Worth a test for the fallback, or a comment documenting that state is always guaranteed non-empty here.

src/pages/iou/request/MoneyRequestParticipantsSelector.tsx
JakubKorytko: Only `Activity` call site that does not pass `inert`.

Before, this was a plain `View`, so React hid it with `display: none` when the mode went hidden. That also removed it from the accessibility tree and tab order. `pinDisplayToContents` now refuses that write, so when hidden the list stays painted, focusable and visible to screen readers.

Dropping `styles.flex1` also matters on web: an RNW `View` is a div with `flex: 1` and a column layout, and `display: contents` has neither, so the list becomes a direct flex child of the grandparent. Native is unaffected, `flex1` was already dead style there.

Why not `inert={!isActivityVisible}` here? plus please make sure the list lookin good on web

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
```diff
@@ -0,0 +1,66 @@
...
+ * navigator lost focus to a route higher in the tree (useIsFocused).
+ *
+ * A covered screen still renders visible on its first render, because React never mounts the effects of a hidden
+ * Activity, so a screen that mounts while covered would never fetch its data. It also stays visible while the
+ * window size is changing, so it lays itself out against the new size before it is revealed. The reveal itself
+ * waits for the navigation transition to end, because revealing in the same commit as the navigation update
+ * blocked the main thread for hundreds of milliseconds on a pop.
+ */
+function useScreenActivityState(isScreenBlurred: boolean): ScreenActivityState {
+    const isFocused = useIsFocused();
```
JakubKorytko: Both test files mock `useIsFocused`, so this never runs against real react-navigation. The wrapper sits above the screen's route context, so it should read the navigator's parent route, which matches the doc comment. Worth confirming by hand: if it resolved to the screen instead, it would just duplicate `isScreenBlurred`.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
```diff
@@ -0,0 +1,66 @@
...
+    const isWindowSizeChanging = useIsWindowSizeChanging();
+    const [hasCompletedFirstRender, setHasCompletedFirstRender] = useState(false);
+
+    useEffect(() => {
+        const rafID = requestAnimationFrame(() => setHasCompletedFirstRender(true));
+        const timeoutID = setTimeout(() => setHasCompletedFirstRender(true), FIRST_RENDER_FALLBACK_DELAY_MS);
```
JakubKorytko: The raf and the timeout both run, so the timeout still fires 100ms later with the same value. React bails out, but whichever wins could clear the other.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
```diff
@@ -0,0 +1,66 @@
...
+        return () => {
```
JakubKorytko: No test unmounts before the raf or the timeout fires, so nothing asserts this cleanup runs. Easy for a refactor to drop it silently.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
JakubKorytko: `nonTopScreenBehavior` is already declared in `NavigationOptions.ts`, so this is a second copy that can drift. If the descriptor options can be typed as `PlatformStackNavigationOptions`, the intersection could go.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * Wraps each screen's render function with the wrapper its nonTopScreenBehavior option picked, which keeps covered
+ * screens off the critical path. Screens that picked no behavior and persistent screens (e.g. the sidebar on web)
+ * stay unwrapped, because a persistent screen remains interactive alongside the top screen.
+ */
+function wrapDescriptorsWithNonTopScreensBehavior<T extends Descriptor>(
+    descriptors: Record<string, T>,
+    state: PlatformStackNavigationState<ParamListBase>,
+    persistentScreens?: string[],
+): Record<string, T> {
+    const topRouteKey = state.routes[state.index]?.key;
```
JakubKorytko, line 39: If `state.routes[state.index]` is undefined, `topRouteKey` is undefined and every route counts as blurred, including the top one. Could happen with an empty routes array or a stale state snapshot. A test for that case, or a comment saying state is always non-empty here, would settle it.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
```diff
@@ -0,0 +1,57 @@
...
+            continue;
+        }
+        const NonTopScreenWrapper = WRAPPER_FOR_BEHAVIOR[behavior];
+        const isScreenBlurred = key !== topRouteKey;
+        result[key] = {
+            ...descriptor,
+            render: () => <NonTopScreenWrapper isScreenBlurred={isScreenBlurred}>{descriptor.render()}</NonTopScreenWrapper>,
+        };
+    }
+    return result;
```
JakubKorytko: This used to be skipped unless `freezeNonTopScreens` was set. Now every navigator walks its descriptors and builds a new object on every render, even when all behaviors are `'none'`, which is all of them except split navigators. Returning `descriptors` unchanged when nothing got wrapped would avoid that and keep the object identity stable.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/StrictModeMountGate.tsx
JakubKorytko: The test only imports the named export, so this default export, the one `ScreenActivityWrapper` uses, is untested. Nothing checks that `USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV: false` picks the pass-through. Since that flag gets flipped locally for profiling, a regression here would show up as unexplained double renders.

src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx
JakubKorytko, line 46: Nothing mounts a navigator to check this value actually reaches each screen's resolved `options.nonTopScreenBehavior`, or that a per-screen value overrides it. That is the whole replacement for `freezeNonTopScreens`, and the existing test uses hand-built descriptors, which skips option resolution. Worth a test.

JakubKorytko (approved):
nice one. I checked the migration keeps parity: on main only `createSplitNavigator` set `freezeNonTopScreens: true`, and that is the navigator that now gets `nonTopScreenBehavior: 'freeze'`. The option also survives `buildPlatformSpecificNavigationOptions`, so it does reach `descriptor.options`.

Main finding is `MoneyRequestParticipantsSelector`: it is the only `Activity` call site that does not pass `inert`, and it used to rely on React writing `display: none` to hide its content. That write is now refused. Rest are small nits and test gaps.

src/libs/Navigation/AppNavigator/createSplitNavigator/index.tsx
dariusz-biela: Added in 18d6c04: `tests/unit/Navigation/nonTopScreenBehavior/splitNavigatorNonTopScreenBehaviorTest.tsx` mounts a real `SplitNavigator` inside a `NavigationContainer` and asserts the wrapper each screen resolved to, so it goes through the full option resolution. It covers the navigator default reaching every covered screen as `ScreenFreezeWrapper`, a per-screen `nonTopScreenBehavior: 'activity'` overriding it, and the top screen coming out as not blurred.

Verified it fails for both ways this can break: dropping `nonTopScreenBehavior: 'freeze'` from `defaultScreenOptions`, and dropping `nonTopScreenBehavior` from the `...rest` passthrough in `getCommonNavigationOptions`. In the second case the rest of the suite still passes, which is exactly the gap you pointed at.

src/pages/iou/request/MoneyRequestParticipantsSelector.tsx
dariusz-biela: Added in 80febb3. This call site now passes `inert={!isActivityVisible}`, so while the mode is `hidden` the list is out of the tab order and out of the accessibility tree, even though the pinned `display: contents` keeps it painted.

On `styles.flex1`: you are right that it was dead on native, and it turns out it was dead on web too. The only child here is the `SelectionList` root, which already carries `flex: 1`, and the grandparent is a column flex container, so under `display: contents` the list simply becomes a direct flex item of the grandparent and ends up in the same box. I checked that with a small Playwright harness replaying the react-native-web styles for every shape this component renders in (step screen wrapper, the same with the `FormHelpMessage` sibling, the picker modal, the bare picker, the share tab scene). The list root box, the scroller height and the scrolling behavior come out identical to the old structure in all of them. The two structures only diverge when more than one child sits under a row direction parent, which no call site does.

The `flex: 1` that native carries now is a different node: the `inert` host inside the contents view. A `display: contents` node generates no box, so `inert` needs a real node of its own, and there `flex: 1` is load bearing. A Yoga 3.2 layout check confirms that without it the node collapses to height 0, because its only child uses `flex-basis: 0`, and the whole list disappears.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
dariusz-biela: Confirmed with a test instead of by hand: `screenActivityNavigationContextTest.tsx` (63398a3) mounts a real `NavigationContainer` with a nested navigator and no `useIsFocused` mock. Pushing a route above the inner navigator leaves `isScreenBlurred` at `false` while the screen turns covered, so the hook does resolve to the parent route and does not duplicate `isScreenBlurred`.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
dariusz-biela: Fixed in d84f445fa180 - the result object is now built lazily, so a navigator whose screens all picked `none` gets the original `descriptors` back by reference. Covered by an identity assertion in the test.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
dariusz-biela: Fair point about the missing coverage, but I would rather leave it untested here.

If the cleanup were dropped in a refactor, the only consequence is one pending raf or timeout firing after the unmount and calling `setHasCompletedFirstRender` on an unmounted component. React takes care of that itself: since React 18 a state update on an unmounted component is a silent no-op, with no warning and nothing kept alive beyond the handle, which resolves within 100ms at the very latest. There is no difference in behavior and no measurable cost either way, so a test asserting the cleanup would be pinning an implementation detail rather than anything observable.

The cleanup stays because it is the correct shape for an effect that schedules work, not because anything depends on it.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
dariusz-biela: Both run, but the losing call sets the value already in state, and React compares it before scheduling anything, so on an idle fiber it bails out without a render at all. What is left is one pending handle for at most 100ms per screen, and the cleanup clears both if the screen unmounts first.

Cross-cancelling means hoisting `timeoutID` into a mutable `let` and turning both arrows into blocks, since the callbacks need each other's handle. More moving parts around a first-render latch for nothing observable, so I am leaving it.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
dariusz-biela: Good catch, fixed in 6f8ff970bee6.

The blurred check is now guarded, so a missing top route key leaves every screen unblurred instead of blurring all of them, the top one included:

```ts
const isScreenBlurred = topRouteKey !== undefined && key !== topRouteKey;
```

I picked the guard over an early `return descriptors`. Bailing out early would render the screens without the wrapper, and once the state recovered on the next render the wrapper would come back as a new parent and remount the whole screen subtree. The guard keeps the tree shape stable in both cases.

The line also carries a comment stating that the state always carries a top route, and there is a unit test for it in `wrapDescriptorsWithNonTopScreensBehaviorTest.tsx`: `keeps the wrappers mounted and blurs no screen when the navigation state has no top route`.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/wrapDescriptorsWithNonTopScreensBehavior.tsx
dariusz-biela: Fixed the duplicate by deriving it: `options: PlatformSpecificNavigationOptions & Pick<PlatformStackNavigationOptions, 'nonTopScreenBehavior'>`. Single declaration in `NavigationOptions.ts` now, and the type follows it.

The intersection has to stay. I tried `options: PlatformStackNavigationOptions` and it does not typecheck: `CommonStackNavigationOptions` is `CommonProperties<StackNavigationOptions, NativeStackNavigationOptions>`, which intersects the types of shared keys, so `header` becomes `((props: StackHeaderProps) => ReactNode) & ((props: NativeStackHeaderProps) => ReactNode)` (`headerTitleStyle` collapses the same way). Neither platform options type is assignable to that. Descriptors carry the adapter output, not the `PlatformStackNavigationOptions` input.

Done in 6fe8a53cb436f.

src/CONFIG.ts
```diff
@@ -133,6 +133,8 @@ export default {
     },
     // to read more about StrictMode see: contributingGuides/STRICT_MODE.md
     USE_REACT_STRICT_MODE_IN_DEV: false,
+    // Turn off locally when profiling to keep the Activity gate's StrictMode double renders out of the measurements
+    USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV: true,
```
staszekscp, line 137: I think this is risky, because all devs working on profiling will have to remember to switch that. Wouldn't it be better to enable it eg. for testing/CI checks?

dariusz-biela: I have mixed feelings about this.

On the one hand, I agree with you that it’s an extra step to keep in mind.

But on the other hand, there may be potential errors related to React effects that won’t be caught by any automated tool—and which are also very hard to spot when working on the Activity itself without React strict mode. That’s why I’m convinced it’s better to enable React strict mode by default. In other projects, you don’t even have the option to disable React strict mode, and you have to remove the component from the App during thorough profiling for this very reason.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/StrictModeMountGate.tsx
dariusz-biela: Fixed in 20168aa2b098c. Both tests now import the default export, so the selection is what gets exercised: one mocks `USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV: true` and asserts the effect, cleanup, effect cycle, the other mocks it `false` and asserts a single effect. The named export is gone, so nothing can bypass the flag anymore.

Checked both directions by mutation: dropping the extra commit from the gate fails the strict test, and swapping the branches of the ternary fails the pass-through one.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/StrictModeMountGate.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Production builds get the plain content because StrictMode does nothing there, and profiling sessions opt out of the
+ * gate through USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV to keep double renders out of the measurements.
+ */
+function StrictModeMountGatePassThrough({children}: StrictModeMountGateProps) {
+    return children;
+}
+
+const DevStrictModeMountGate = __DEV__ && CONFIG.USE_ACTIVITY_SCREEN_STRICT_MODE_IN_DEV ? StrictModeMountGate : StrictModeMountGatePassThrough;
```
dariusz-biela, line 33: I'm wondering if it might also be a good idea to enable strict mode on the **staging** environment to make it easier for the QA team to catch bugs?

src/CONFIG.ts
roryabraham: I think long-term we should probably work to re-enable React Strict Mode project-wide, and fix all the "dev-only bugs", since the point of React Strict Mode is to surface those bugs from code that is effectively breaking the Rules of React.

That's what most projects would likely do, and in those projects, disabling strict mode would be a pretty normal part of profiling, right?

src/components/AlwaysPaintedView/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * and the browser would pay for the layout reflow. The React version in use touches 'display' only through these
+ * two entry points, so re-verify this after React upgrades.
+ */
+function pinDisplayToContents(element: HTMLDivElement) {
+    const {style} = element;
+    const setStyleProperty = style.setProperty.bind(style);
+
+    setStyleProperty('display', 'contents', 'important');
+
+    Object.defineProperty(style, 'setProperty', {
```
roryabraham, line 19: *question:* Would a `Proxy` be more robust here than `Object.defineProperty`?

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * covered screen that is still shown (for example dimmed under the RHP overlay) does not disappear, and it takes
+ * that content out of accessibility and touch handling while it is covered.
+ *
+ * StrictMode is the qualification gate for screens that opt into Activity. Its double effect mount in dev exercises
+ * the same cleanup and re-run lifecycle as a hide and reveal cycle, so an effect that would misbehave under a cover
+ * fails during development instead. StrictModeMountGate commits StrictMode one commit ahead of the screen content,
+ * which is what makes React run that cycle for a StrictMode nested below the root.
+ */
+function ScreenActivityWrapper({isScreenBlurred, children}: NonTopScreenWrapperProps) {
+    const {mode, isScreenCovered} = useScreenActivityState(isScreenBlurred);
```
roryabraham: Did you put this in a separate hook just so it's easier to test? It looks like it's only used once in the code, so ideally its logic would be inlined in this component. That would make it easier to read this component and understand what it does

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useIsWindowSizeChanging.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Reports whether the window width is changing, on both platforms. A screen hidden by
+ * <Activity> has no mounted effects, so it reads this flag to become visible for the duration of the change and
+ * lay itself out while still covered. The existing useIsResizing does not fit here, because it is web only and
+ * also counts the height changes the soft keyboard causes. The hook lives apart from windowSizeChangeStore because
+ * the store compiles differently under the two React Compiler toolchains, which the compliance check reads as a
+ * memoization divergence.
+ */
+function useIsWindowSizeChanging() {
```
roryabraham: Same deal here - this is only used once so having it in a separate file imo is just obfuscating things rather than being a useful abstraction.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useIsWindowSizeChanging.ts
```diff
@@ -0,0 +1,17 @@
+import {useSyncExternalStore} from 'react';
+
+import {getSnapshot, subscribe} from './windowSizeChangeStore';
+
+/**
+ * Reports whether the window width is changing, on both platforms. A screen hidden by
+ * <Activity> has no mounted effects, so it reads this flag to become visible for the duration of the change and
+ * lay itself out while still covered. The existing useIsResizing does not fit here, because it is web only and
+ * also counts the height changes the soft keyboard causes. The hook lives apart from windowSizeChangeStore because
```
roryabraham: It would be great to dig into the differences in the React Compiler memoization of windowSizeChangeStore and see if we can get both compiling rather than introducing this. In my experience there are often workarounds to syntactic limitations of the React Compiler.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
```diff
@@ -0,0 +1,66 @@
...
+    isScreenCovered: boolean;
+};
+
+/**
+ * Returns the Activity mode of a screen and whether it is covered. The two are separate because the accessibility
+ * state must follow the navigation state with no delay, while the mode deliberately lags behind on the reveal. A
+ * screen is covered when another screen of its own navigator is on top of it (isScreenBlurred) or when the whole
+ * navigator lost focus to a route higher in the tree (useIsFocused).
+ *
+ * A covered screen still renders visible on its first render, because React never mounts the effects of a hidden
```
roryabraham: > A covered screen still renders visible on its first render, because React never mounts the effects of a hidden Activity, so a screen that mounts while covered would never fetch its data

This feels like a code smell: we're fighting how the React Activity API is designed.

We're talking about a case where we want to mount a hidden screen that's likely to become visible soon, fetch its data, and pre-render it so it can be revealed faster when it should become visible. That's the whole point of this change.

This is precisely the use case showcased in the [Pre-rendering content that's likely to become visible](https://react.dev/reference/react/Activity#pre-rendering-content-thats-likely-to-become-visible) example from the React docs.

Those docs confirms what this code comment says:

> When an Activity boundary is hidden during its initial render, its children won’t be visible on the page — but they will still be rendered, albeit at a lower priority than the visible content, and without mounting their Effects.

The solution we have here is to force the hidden view to render with `<Activity mode='visible'>` once, so that it runs its effects and fetches its data synchronously, then resume rendering invisibly in the background with lower priority (`<Activity mode='hidden'>`).

Notably, that is _not_ what the docs recommend, and it's not ideal because it forces a high-priority synchronous render of an invisible screen.

Instead, the docs suggest:

- Start by wrapping the screen in a `Suspense` boundary. This allows the screen to render in the background with lower priority and show a lightweight loading indicator if it ever becomes visible before it's ready.
- Then, make the thing that fetches data return a Promise.
- Then, `use(fetchData())` to suspend the component while data is fetching. As long as the screen is wrapped in a `Suspense` boundary, this will fetch the data and render the component in the background, intelligently show a loading indicator if it ever becomes visible before it's ready, and swap in the real component without re-rendering the parent as soon as it is ready.
    - _Note:_ This concurrent render + fetch only begins once the component is mounted.
    - _Note:_ React has built-in logic to ensure that, if we use this pattern correctly, it should "just work" without seeing flashes of loading indicators
- Lastly, wrap the hidden screen we want to pre-render with `<Activity mode='hidden'>`. This allows its data fetching and pre-rendering to happen eagerly in the background (before the component would normally mount).
    - If it becomes focused _before_ data fetching and pre-rendering finish, we gracefully show a suspense fallback while that finishes in the background.
    - If the screen becomes focused _after_ data fetching and pre-rendering finish, then React just has to set `display=block` (or anything other than `display=none`), and nearly instantly the content becomes visible and interactive.

I understand this could mean to a pretty significant change from the current design (particularly the part where we make the data fetching return a Promise), but I think that's ultimately the direction we should go with this.

This PR in its current form is already a significant paradigm shift, so imo it's worth putting in the extra work to do it right from the start. It should end up being not only more performant, but substantially simpler as well. Ultimately, using the concurrent React APIs are they are designed to be used rather than fighting them will be important for the predictability and maintainability of the product.

Another benefit of implementing suspense-driven data loading in screens _first_ is that React Navigation 8 is going to wrap screen transitions in `startTransition` by default. Pairing that with a Suspense boundary for each screen wil mean that screens can concurrently render and fetch data with lower priority during transitions, improving interactivity during transitions (i.e: if a user clicks on two different navigation buttons in quick succession, the first action can be abandoned and the second one can take precedence), and potentially reducing our dependence on `TransitionTracker.runAfterTransitions` (i.e: we don't have to think about loading indicator logic as much. Components just suspend until they have what they need, then render). This is also more efficient that keeping `isLoading` states in parent components, because suspended children are pre-rendered and efficiently swapped out, rather than the full parent tree re-rendering

roryabraham (changes requested):

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useScreenActivityState.ts
dariusz-biela: Thanks for the deep dive. To start with, you have a point about that comment: it was not precise enough and described the mechanism as a data-fetching workaround, so I'm updating it to explain the real rationale. I agree that Suspense + `use()` + hidden `<Activity>` is the right long-term model, especially once React Navigation 8 wraps transitions in `startTransition` where possible (v8 is still prerelease and we are on v7). The reason the frame exists in its current form is two mechanisms already shipped in the app that mount screens covered from their very first commit: `usePreMountDestination` and dynamic path deep links. Neither opts into `activity` yet (they run under `freeze` today), but they define the mount semantics any covered screen has to keep after migrating.

`usePreMountDestination` is our imperative version of the "pre-rendering content that's likely to become visible" pattern from the docs, built without `use()` because our reads are Onyx store snapshots rather than suspendable resources. On narrow layout it pre-inserts the destination fullscreen under the open RHP, scheduled in idle after transitions, so the render stays off the RHP opening transition path. The visible first frame is what makes that pre-mount useful: it runs the screen's mount lifecycle at mount time, so work like `openReport` fires early and its data lands in the Onyx cache while the screen is hidden. To be precise about what it does not do: hiding an Activity cleans effects up, so subscriptions are not live while hidden, and the reveal re-runs every effect. The win is prewarming, not liveness. The reveal re-subscribes against a warm cache and renders real content in one catch-up pass, instead of mounting effects for the first time and starting the first fetch after the user is already looking at the screen.

Dynamic path deep links hit the same mechanics on cold start. The parser builds a state where the base fullscreen mounts in the same commit as the dynamic route screen on top of it, so the base screen is covered from birth. With the frame it mounts live once, matching how those underlays behave today, so closing the RHP reveals a screen whose data fetch started at mount. Without it, closing the RHP would reveal a screen that only then starts loading.

Migrating to your model needs a cached, promise-based read resource over Onyx for `use()` (writes do return promises, but there is nothing suspendable to read), an offline-first answer for suspending, and it still would not cover side-effectful mount work like `openReport` or Pusher subscriptions, which stay in effects a hidden Activity will not run. That is a data-layer project plus a lifecycle audit of the screens, so I would keep it out of this PR's scope.

My proposal is to keep the frame as the effect-prewarming bridge these mount semantics need and track suspense-driven screens as a follow-up. The wrapper is already shaped for that end state: screens pick their covered behavior through the `nonTopScreenBehavior` option (`none` / `freeze` / `activity`), and the visible first frame is just one term in the mode the wrapper computes, separate from the covered detection that drives `inert`. So the migration path is a new behavior value (for example `activity-suspense`) or a flag on the existing one that skips the first frame for screens that fetch through `use()`. Screens converge on the model you describe one by one, while everything else they still need (covered detection, the resize hold, the deferred reveal) stays shared.

src/CONFIG.ts
dariusz-biela: > That's what most projects would likely do, and in those projects, disabling strict mode would be a pretty normal part of profiling, right?

In other projects, that's what I've done myself.

> re-enable React Strict Mode project-wide

Yes, that would be ideal, but I'm afraid it would require a massive amount of work, so personally, I'd like to push it through for Activity for now, because that's where we need it.

As for refactoring the entire app to React Strict Mode, it might be worth making that a separate proposal, with a detailed migration plan.

src/components/AlwaysPaintedView/index.tsx
dariusz-biela: Good question! I prototyped it (patch below, all tests pass), but I'd keep `Object.defineProperty`:

- A Proxy can't replace `element.style` on its own (`style` is a prototype accessor), so it still needs `defineProperty` to shadow the getter - it's Proxy + defineProperty, not Proxy instead of it.
- The current patch covers exactly the two paths React uses (`hideInstance` -> `style.setProperty('display', 'none', 'important')`, `unhideInstance` -> `style.display = ...`, verified in `react-dom-client`); the Proxy traps every style access from all code instead.
- It's roughly 2x the code (binding native methods to avoid `Illegal invocation`, overriding `style`/`setAttribute`/`removeAttribute`) and even then not airtight - `setAttributeNode`, `attributeStyleMap`, and element cloning still bypass it.

I'd only switch given a concrete repro that hits one of the extra paths (`cssText`, style-attribute rewrite) - the patch is ready if you disagree.

<details>
<summary>Proxy version (tested, lint/typecheck/React Compiler clean)</summary>

```diff
diff --git a/src/components/AlwaysPaintedView/index.tsx b/src/components/AlwaysPaintedView/index.tsx
index b9ca2f1020b0..4b44c64c9d04 100644
--- a/src/components/AlwaysPaintedView/index.tsx
+++ b/src/components/AlwaysPaintedView/index.tsx
@@ -2,33 +2,82 @@ import {useRef} from 'react';
 
 import type AlwaysPaintedViewProps from './types';
 
+type StyleMethod = (...args: unknown[]) => unknown;
+
+function isStyleMethod(member: unknown): member is StyleMethod {
+    return typeof member === 'function';
+}
+
 /**
  * Refuses every write to 'display' on the element. React hides a host element with
- * 'style.setProperty(display, none, important)' and reveals it by assigning to 'style.display', so both are
- * replaced. Blocking the writes keeps the element from ever reaching 'display: none'. A MutationObserver
- * fallback would only restore the value after the fact, so the element would still pass through 'display: none'
- * and the browser would pay for the layout reflow. The React version in use touches 'display' only through these
- * two entry points, so re-verify this after React upgrades.
+ * 'style.setProperty(display, none, important)' and reveals it by assigning to 'style.display'. A Proxy over the
+ * real CSSStyleDeclaration traps both plus 'removeProperty', while writes that replace the whole declaration
+ * ('cssText', 'element.style = ...', and the style attribute) go through and the display pin is re-applied in the
+ * same task, before the browser lays out the intermediate value. Native methods are bound to the real declaration
+ * before the trap returns them, because calling one with the Proxy as the receiver throws an 'Illegal invocation'
+ * error.
  */
 function pinDisplayToContents(element: HTMLDivElement) {
     const {style} = element;
-    const setStyleProperty = style.setProperty.bind(style);
+    const pinDisplay = () => style.setProperty('display', 'contents', 'important');
+    pinDisplay();
+
+    const styleProxy = new Proxy(style, {
+        get(target, property) {
+            if (property === 'setProperty') {
+                return (name: string, value: string | null, priority?: string) => {
+                    if (name === 'display') {
+                        return;
+                    }
+                    target.setProperty(name, value, priority);
+                };
+            }
+            if (property === 'removeProperty') {
+                return (name: string) => (name === 'display' ? 'contents' : target.removeProperty(name));
+            }
+            const member: unknown = Reflect.get(target, property);
+            return isStyleMethod(member) ? member.bind(target) : member;
+        },
+        set(target, property, value) {
+            if (property === 'display') {
+                return true;
+            }
+            const didSet = Reflect.set(target, property, value);
+            if (property === 'cssText') {
+                pinDisplay();
+            }
+            return didSet;
+        },
+    });
 
-    setStyleProperty('display', 'contents', 'important');
+    Object.defineProperty(element, 'style', {
+        configurable: true,
+        get: () => styleProxy,
+        set: (cssText: string) => {
+            style.cssText = cssText;
+            pinDisplay();
+        },
+    });
 
-    Object.defineProperty(style, 'setProperty', {
+    const setAttribute = element.setAttribute.bind(element);
+    Object.defineProperty(element, 'setAttribute', {
         configurable: true,
-        value: (property: string, value: string | null, priority?: string) => {
-            if (property === 'display') {
-                return;
+        value: (name: string, value: string) => {
+            setAttribute(name, value);
+            if (name.toLowerCase() === 'style') {
+                pinDisplay();
             }
-            setStyleProperty(property, value, priority);
         },
     });
-    Object.defineProperty(style, 'display', {
+    const removeAttribute = element.removeAttribute.bind(element);
+    Object.defineProperty(element, 'removeAttribute', {
         configurable: true,
-        get: () => 'contents',
-        set: () => {},
+        value: (name: string) => {
+            removeAttribute(name);
+            if (name.toLowerCase() === 'style') {
+                pinDisplay();
+            }
+        },
     });
 }
 
```

</details>

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/index.tsx
dariusz-biela: Yep, testing only. Inlined it into the component and moved those tests into `ScreenActivityWrapperTest`, where they assert the same behavior through the rendered output. 791823037d66

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useIsWindowSizeChanging.ts
dariusz-biela: Agreed, the file is gone. The component now calls `useSyncExternalStore` on the store directly.

src/libs/Navigation/PlatformStackNavigation/createPlatformStackNavigatorComponent/ScreenActivityWrapper/useIsWindowSizeChanging.ts
dariusz-biela: Dug in, and it turned out to be a false positive in our own detector rather than a compiler limitation. `config/reactCompiler/checkWithOxc.mjs` falls back to diffing the compiled output against a plain transform when there is no `_c(` marker, and the only difference here was the comments the compiler pass drops from `handleDimensionsChange`. No memo cache is emitted anywhere and Babel reports `memoBlocks = 0`, so both compilers actually agree.

Inlining the hook (per your other comment) leaves the store with no hook at all, so the compliance check now passes with no divergence and nothing had to be worked around. I will file the `checkWithOxc.mjs` heuristic fix separately, since it can hit any hook-only `.ts` file.

src/components/AlwaysPaintedView/index.native.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * no effects.
+ */
+function AlwaysPaintedView({inert = false, children}: AlwaysPaintedViewProps) {
+    const styles = useThemeStyles();
+
+    return (
+        <NativeAlwaysPaintedView style={DISPLAY_CONTENTS}>
+            <View
+                aria-hidden={inert}
+                style={[styles.flex1, {pointerEvents: inert ? 'none' : 'box-none'}]}
```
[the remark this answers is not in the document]
dariusz-biela: `styles.pointerEventsBoxNone` resolves to `{}` on native (see `index.native.ts`), so this branch would silently drop `box-none` in a native-only component. Switched to the `pointerEvents` prop instead - no inline style object, behavior preserved.

trjExpensify:
PR doesn’t need product input as a perf refactor PR. Unassigning and unsubscribing myself.

roryabraham (approved):

truph01:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] I verified that any new or modified comments were clear, correct English, and explained """"why"""" the code was doing something instead of only explaining """"what"""" the code was doing.


### Screenshots/Videos
<details>






</details>

<details>



</details>

<details>






</details>

<details>



</details>

<details>



</details>

truph01 (approved):

roryabraham (approved):

PR #98759, opened 2026-08-17
[No QA] refactor: remove unsafe assertions from deep replacement utility

KJ21-ENG:
### Explanation of Change

This PR removes five unsafe type assertions from `deepReplaceKeysAndValues`. The utility now accepts the request-data records it actually processes and transforms nested keys and values without unsafe assertions.

Behavior remains unchanged for missing values, nested and frozen objects, replacement matches, fallback values, and `File` or `Blob` values.

### Fixed Issues

### Count change

5 -> 0 (5 removed)

### Changed files

- `src/libs/deepReplaceKeysAndValues.ts`: removes the unsafe assertions and narrows the function to its real request-data input.
- `tests/unit/deepReplaceKeysAndValuesTest.ts`: keeps the utility tests aligned with that input.

### Tests

- `tests/unit/deepReplaceKeysAndValuesTest.ts`: 20/20 passed.
- The relevant `HandleUnusedOptimisticID` tests in `tests/unit/MiddlewareTest.ts`: 11/11 passed.

### Offline tests

N/A. This utility does not perform requests or change offline behavior.

### QA Steps

[No QA]. This is a non-UI utility cleanup covered by the unit and middleware tests above.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

N/A. No UI or visual behavior changed.


ikevin127:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -49,7 +64,17 @@ function deepReplaceKeysAndValues<T extends ReplaceableValue>(target: T, oldVal:
...
-    return newObj as T;
+    return newObj;
+}
+
+/**
+ * @param target the object or value to transform
+ * @param oldVal the value to search for
+ * @param newVal the replacement value
+ */
+function deepReplaceKeysAndValues<T extends ReplaceableValue>(target: T, oldVal: string, newVal: string): DeepReplaceResult<T>;
```
ikevin127: 🟠 **`deepReplaceKeysAndValues.ts:75` — the new return type is never checked by the compiler, so this trades 5 visible assertions for 1 invisible one.**

TypeScript does not verify an overload signature's return type against the implementation. `isImplementationCompatibleWithOverload` passes if *either* direction is assignable, and everything is assignable to the implementation's `ReplaceableValue` here. So `DeepReplaceResult<T>` is an unverified claim, exactly like the `as T` casts it replaced, except ESLint can no longer see it.

I proved this compiles clean under `--strict`:

```ts
function lying<T extends ReplaceableValue>(target: T, oldVal: string, newVal: string): number;
function lying(target: ReplaceableValue, oldVal: string, newVal: string): ReplaceableValue {
    return target;
}
const lie = lying({a: 'x'}, 'a', 'b');
const proof: number = lie; // compiles. At runtime `lie` is an object.
```

Zero errors. The same lack of checking applies to the `transformReplaceableValue` pair at `:20-21`, where overload 1 exists solely so the outer implementation's `ReplaceableValue` return type is satisfied by a body that returns `unknown`.

**Why this matters for the seatbelt series**: the rule `@typescript-eslint/no-unsafe-type-assertion` exists to surface unproven type claims. Moving an unproven claim from `as T` into an overload return position makes the count go to zero while the unsoundness stays. The next person who edits a runtime branch gets no compiler feedback that the declared type is now wrong.

There is a version of this refactor where the return type is genuinely checked. Split the record case out so its return type is validated against real `return` statements, and let the recursive worker honestly return `unknown`:

```ts
/** Recursive worker. Returns `unknown` because an erased nested value genuinely is unknown. */
function replaceDeep(target: unknown, oldVal: string, newVal: string): unknown {
    if (!target) {
        return target;
    }
    if (typeof target === 'string') {
        return target.replace(oldVal, newVal);
    }
    if (typeof target !== 'object') {
        return target;
    }
    if (Array.isArray(target)) {
        return target.map((item: unknown) => replaceDeep(item, oldVal, newVal));
    }
    return replaceRecordKeysAndValues(target, oldVal, newVal);
}

/** Return type is CHECKED: every return statement is a real Record. */
function replaceRecordKeysAndValues(target: object, oldVal: string, newVal: string): Record<string, unknown> {
    const newObj: Record<string, unknown> = {};
    // ...unchanged body...
    return newObj;
}
```

I compiled that under `--strict` and it exits 0 with zero assertions, zero overloads, and zero unverified return types. The call site gets what it actually needs, also fully checked:

```ts
persistedRequestClone.data = persistedRequest.data ? replaceRecordKeysAndValues(persistedRequest.data, oldReportID, preexistingReportID) : undefined;
```

That is a real removal of the 5 violations rather than a relocation.

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -1,29 +1,44 @@
 type ReplaceableValue = Record<string, unknown> | unknown[] | string | number | boolean | undefined | null;
 
+type DeepReplaceResult<T> = unknown extends T
+    ? unknown
+    : T extends string
+      ? string
+      : T extends unknown[]
```
ikevin127: 🟡 **`deepReplaceKeysAndValues.ts:7` — `DeepReplaceResult` already has a soundness hole for `readonly` arrays.**

`T extends unknown[]` is false for `readonly` arrays, since a `readonly T[]` is not assignable to a mutable `unknown[]`. Those fall all the way through to the `Record<string, unknown>` default, but at runtime `Array.isArray` is true and they come back as arrays.

Confirmed at the type level (this assignment compiles):

```ts
type RO = DeepReplaceResult<Array<readonly string[]>>;
declare const ro: RO;
const proof: Array<Record<string, unknown>> = ro; // compiles
```

And confirmed at runtime with the actual function body:

```
readonly nested array -> [["newVal"]] | Array.isArray(out[0]) = true
```

So the type says `Record<string, unknown>` and the value is an array. Fix is one word:

```suggestion
      : T extends readonly unknown[]
```

This is narrow in practice, but it is the concrete demonstration of the problem in the finding above: a hand-maintained conditional type that the compiler never checks against the implementation drifts, and nothing tells you.

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -1,29 +1,44 @@
...
+        ? Array<DeepReplaceResult<T[number]>>
+        : T extends (...args: never[]) => unknown
+          ? T
+          : T extends abstract new (...args: never[]) => unknown
+            ? T
+            : T extends number | boolean | bigint | symbol | null | undefined | void
+              ? T
+              : Record<string, unknown>;
```
ikevin127: 🟡 **`deepReplaceKeysAndValues.ts:3-15` — the 13-line conditional type buys nothing at the only call site.**

At `HandleUnusedOptimisticID.ts:146` and `:153` the input is `Record<string, unknown> | undefined`, so `DeepReplaceResult` evaluates to `Record<string, unknown> | undefined`. I verified that resolution:

```ts
type CallSite = DeepReplaceResult<Record<string, unknown> | undefined>;
const proof: Record<string, unknown> | undefined = cs; // compiles
```

Because `Record<string, unknown>` values are `unknown`, the recursion immediately hits the `unknown extends T ? unknown` branch and the whole distributive machinery collapses. Every branch below `T extends string` is reachable only from the test file.

Carrying a recursive conditional type, two overload pairs, and a callable/constructable special case for a single caller that needs `Record<string, unknown> | undefined` is a maintenance cost with no consumer. 

The split above gets the same call-site type in a third of the lines.

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -1,29 +1,44 @@
...
 
     if (Array.isArray(target)) {
-        return target.map((item) => deepReplaceKeysAndValues(item as T, oldVal, newVal)) as T;
+        return target.map((item) => transformReplaceableValue(item, oldVal, newVal));
     }
 
     const newObj: Record<string, unknown> = {};
-    for (const [key, val] of Object.entries(target)) {
+    for (const [key, entryValue] of Object.entries(target)) {
+        const val: unknown = entryValue;
```
ikevin127, line 8: 🟢 **`deepReplaceKeysAndValues.ts:41` — `const val: unknown = entryValue;` needs a one-line comment.**

This is doing real work: `Object.entries` on an `object`-narrowed value resolves to the `[string, any][]` overload, and the annotation is an `any` barrier.

Without a comment it reads as a redundant rebind and the next cleanup pass will inline it straight back to `for (const [key, val] of ...)`, quietly restoring `any` through the whole loop body.

```suggestion
      // `Object.entries` on a bare `object` yields `[string, any][]`, so re-bind through `unknown` to stop `any` leaking into the branches below.
      const val: unknown = entryValue;
```

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -1,29 +1,44 @@
...
+            ? T
+            : T extends number | boolean | bigint | symbol | null | undefined | void
+              ? T
+              : Record<string, unknown>;
+
 /**
- * @param target the object or value to transform
- * @param oldVal the value to search for
- * @param newVal the replacement value
+ * Recursively replaces keys and values while preserving the broad type of recursive values.
```
ikevin127: 🟢 **`deepReplaceKeysAndValues.ts:17-19` — the JSDoc describes the wrong function.**

`Recursively replaces keys and values while preserving the broad type of recursive values` sits on `transformReplaceableValue`, which preserves nothing and returns `unknown`.

The type preservation happens at `:75`. Either move the sentence or reword it to say this is the untyped worker and the public wrapper declares the shape.

ikevin127:
@KJ21-ENG Dropped (5) comments that should be addressed before approval here.

src/libs/deepReplaceKeysAndValues.ts
KJ21-ENG: Addressed in ba9fff393f9f0b02d21b48e396425c5b63a0c8e2.

I removed DeepReplaceResult and both overload pairs. The exported function now has the compiler-checked record contract required by the real caller, while the private recursive worker honestly returns unknown.

src/libs/deepReplaceKeysAndValues.ts
KJ21-ENG: Addressed by removing the conditional result type entirely. There is no longer a separate readonly-array branch or an unchecked type claim that could disagree with the runtime array behavior.

src/libs/deepReplaceKeysAndValues.ts
KJ21-ENG: Addressed. The public API is now limited to Record<string, unknown> | undefined, matching the actual production caller. Primitive and array test scenarios pass through that record boundary, so a broader generic API is no longer retained only for tests.

src/libs/deepReplaceKeysAndValues.ts
KJ21-ENG: Addressed. I retained the entryValue → unknown barrier and added a comment explaining that it prevents the any produced by Object.entries from propagating into recursion and assignments.

src/libs/deepReplaceKeysAndValues.ts
KJ21-ENG: Addressed. The private worker’s JSDoc now describes processing an erased nested value, while the exported function’s JSDoc describes the optional request-data record boundary.

KJ21-ENG:
@ikevin127 Thanks for the detailed review. I’ve addressed all five findings and replied to each thread with the corresponding change. 

The updated head is ba9fff393f9f0b02d21b48e396425c5b63a0c8e2.

When you get a chance, could you please take another look?

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -49,7 +49,45 @@ function deepReplaceKeysAndValues<T extends ReplaceableValue>(target: T, oldVal:
         newObj[newKey] = val;
     }
 
-    return newObj as T;
+    return newObj;
```
ikevin127, line 34: ### 🟠 New problem: The object-traversal loop now exists twice

The fix I suggested had **one** loop, with the worker delegating into it. The implementation went the other way and copied the loop body into the public function, so `transformReplaceableValue` (lines 23-52) and `deepReplaceKeysAndValues` (lines 63-90) now contain the same five branches in the same order.

**I diffed them line by line**: they are logically identical today. That is the point. Any future change to the replacement rules has to land in both, and nothing enforces it.

This is measurable, not theoretical. Running the suite locally on `ba9fff3`:

```
deepReplaceKeysAndValues.ts | 89.13% Stmts | 90% Branch | Uncovered: 16, 30-31, 68-69
Tests: 18 passed, 18 total
```

Lines **30-31** are the worker's `File`/`Blob` branch. Lines **68-69** are the *duplicate* `File`/`Blob` branch in the public function. The duplication created a second uncovered copy of the same untested branch, which is precisely the `-2.95%` Codecov is reporting.

Line 16 is also now unreachable except through an array holding a non-string primitive, since the public loop handles primitives inline.

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -49,7 +49,45 @@ function deepReplaceKeysAndValues<T extends ReplaceableValue>(target: T, oldVal:
...
+}
+
+/**
+ * Recursively replaces keys and values in an optional request-data record.
+ */
+function deepReplaceKeysAndValues(target: Record<string, unknown> | undefined, oldVal: string, newVal: string): Record<string, unknown> | undefined {
```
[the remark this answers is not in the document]
ikevin127: @KJ21-ENG ### 🔴 Do not apply the bot's CONSISTENCY-3 suggestion

The bot correctly diagnosed the duplication, then proposed this:

```ts
return transformReplaceableValue(target, oldVal, newVal) as Record<string, unknown>;
```

That reintroduces an `@typescript-eslint/no-unsafe-type-assertion` violation into the one file whose entire purpose in this PR is removing them. It would put a new entry straight back into the seatbelt baseline and undo finding `#1`, since the `as` is unverified in exactly the same way the old overload return type was. Same failure mode as its comment on #98846: correct smell, fix that fights the issue being worked ❌ 

### 🟢 The fix that satisfies both

Delegate in the opposite direction. Keep one loop, in a record-shaped function whose return type is checked, and have the worker call *it* for the object case:

```ts
function transformReplaceableValue(target: unknown, oldVal: string, newVal: string): unknown {
    // ...unchanged guards...
    if (Array.isArray(target)) {
        return target.map((item) => transformReplaceableValue(item, oldVal, newVal));
    }
    return replaceRecordKeysAndValues(target, oldVal, newVal);
}

/** Return type is checked: every return statement is a real Record. */
function replaceRecordKeysAndValues(target: object, oldVal: string, newVal: string): Record<string, unknown> {
    // ...the single copy of the loop, unchanged...
    return newObj;
}

function deepReplaceKeysAndValues(target: Record<string, unknown> | undefined, oldVal: string, newVal: string): Record<string, unknown> | undefined {
    if (!target) {
        return target;
    }
    return replaceRecordKeysAndValues(target, oldVal, newVal);
}
```

I verified this: compiles under `--strict` with exit 0, **zero assertions, zero overloads**, and the public contract stays byte-identical. I also ran both shapes against 16 inputs (nested objects, nested arrays, `Blob` as a direct property, `Blob` inside an array, `Date`, `null`, `0`, `false`, `''`, deep nesting) and they produce identical output, with `Blob` identity preserved by reference.

It also removes the duplicate uncovered branch at 68-69, which recovers the Codecov delta ✅

tests/unit/deepReplaceKeysAndValuesTest.ts
```diff
@@ -123,7 +123,11 @@ describe('deepReplaceKeysAndValues', () => {
...
             },
         ],
-    ])('deepReplaceKeysAndValues(%s)', (input: ReplaceableValue, expected: ReplaceableValue) => {
-        expect(deepReplaceKeysAndValues(input, 'oldVal', 'newVal')).toStrictEqual(expected);
+    ])('transforms scenario value %s through the request-data record boundary', (input: ReplaceableValue, expected: ReplaceableValue) => {
+        expect(
+            input === undefined || (typeof input === 'object' && input !== null && !Array.isArray(input))
+                ? deepReplaceKeysAndValues(input, 'oldVal', 'newVal')
+                : deepReplaceKeysAndValues({value: input}, 'oldVal', 'newVal'),
+        ).toStrictEqual(expected === undefined || (typeof expected === 'object' && expected !== null && !Array.isArray(expected)) ? expected : {value: expected});
```
ikevin127: ### 🟡 The test rewrite is worse than what it replaced

**Two problems**:

1. **The test computes its expectation instead of stating it.** The same shape predicate runs on both the input and the expected value, so a bug in the predicate cancels itself out on both sides. A test that derives what it should equal cannot fail for the reason you want it to.

2. **The `null` case silently stopped testing what its name says.** The new signature does not accept `null`, so the `[null, null]` row falls into the wrapping branch and asserts `{value: null}` equals `{value: null}`. Top-level `null` is never passed to the function at all, yet the runner still prints:

   ```
   ✓ transforms scenario value null through the request-data record boundary
   ```

   That is a green test claiming coverage the code no longer has.

Since the public API is now `Record<string, unknown> | undefined`, the honest move is to split the table: keep record-shaped scenarios calling `deepReplaceKeysAndValues` directly with literal expected values, and drop or explicitly re-express the primitive/array/`null` rows as the nested cases they have actually become (`{value: [...]}` in, `{value: [...]}` out, written out rather than computed).

While you are here, a `Blob` case would cover lines 30-31 and retire the oldest untested branch in the file.

src/libs/deepReplaceKeysAndValues.ts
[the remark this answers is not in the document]
KJ21-ENG: Addressed in 6599abf54a0358546ea93a2a3b5ca71cc2f8f661. I delegated in the opposite direction as suggested: transformRecord() now owns the single object-traversal loop, while both the recursive worker and public entry point delegate to it. No cast or overload was introduced.

src/libs/deepReplaceKeysAndValues.ts
KJ21-ENG: Addressed in 6599abf54a0358546ea93a2a3b5ca71cc2f8f661. There is now exactly one Object.entries() traversal loop in transformRecord(). transformReplaceableValue() delegates object values to it, and deepReplaceKeysAndValues() delegates the public record boundary to it.

tests/unit/deepReplaceKeysAndValuesTest.ts
KJ21-ENG: Addressed in 6599abf54a0358546ea93a2a3b5ca71cc2f8f661. The tests now use explicit literal inputs and expected outputs instead of computing expectations. Top-level absence has its own undefined test, primitive/array/null behavior is honestly exercised inside the supported record boundary, and direct File and Blob property cases cover identity preservation.

KJ21-ENG:
@ikevin127 When you get a chance, could you please take another look?

KJ21-ENG:
@ikevin127 When you get a chance, could you please take another look?

ikevin127:
Catching up to this one tomorrow!

ikevin127 (approved):
**🟢 LGTM**

Both coments addressed at 6599abf54a0 👍

**Delegation direction:** `transformRecord()` now owns the single `Object.entries()` loop with a checked `Record<string, unknown>` return, and both `transformReplaceableValue()` and `deepReplaceKeysAndValues()` delegate to it. 

Zero `as` casts (the bot's suggestion correctly not applied), zero overloads, duplicate File/Blob branch removed, public contract unchanged.

src/libs/deepReplaceKeysAndValues.ts
```diff
@@ -1,29 +1,11 @@
...
-    }
-
-    if (Array.isArray(target)) {
-        return target.map((item) => deepReplaceKeysAndValues(item as T, oldVal, newVal)) as T;
-    }
-
+// `object` intentionally accepts every non-null object variant reached by recursive traversal.
+// A string-keyed `Record<string, unknown>` would reject valid object instances and change the preserved boundary.
+// eslint-disable-next-line @typescript-eslint/no-restricted-types
+function transformRecord(target: object, oldVal: string, newVal: string): Record<string, unknown> {
```
blimpich, line 4: Noting that it would've made the diff a lot nicer if the transformRecord function was defined below `deepReplaceKeysAndValues()` since then it wouldn't have appeared like we were deleting it entirely right here

blimpich (approved):

PR #97339, opened 2026-07-29
[No QA] MenuItem decomposition - simplest usages

war-in:
### Explanation of Change

Part 1 of the `MenuItem` decomposition. Moves the legacy monolith into `src/components/MenuItem/` and introduces a compound API (`Root`/`Row`/`Content`/`Icon`/`Title`/`Description`/`Trailing`/`Chevron`) sharing interaction state via context, plus a11y-label derivation from `Title`/`Description`. Adds a `MenuItemStandard` preset and migrates the simplest call sites to it. Legacy default export unchanged.

Legacy `isInteractive` prop is derived from `onPress`, so callers should pass undefined when the row is not interactive

### Fixed Issues
$ https://github.com/Expensify/App/issues/96202
PROPOSAL:

### Tests

1. Run `npm run storybook` and verify that all components in the `MenuItemComparison` section are identical

### Offline tests

N/A — pure component refactor, no API or Onyx changes.

### QA Steps

N/A

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
[...]

PR #98862, opened 2026-08-18
[No QA] MenuItem decomposition - usages with avatars

war-in:
### Explanation of Change

Part 2 of the `MenuItem` decomposition — the avatar-led rows. Adds the `Leading` and `Label` primitives, a typography `variant` on `Description`, `testID` on `Root` and the presets, and a `MenuItemEntity` preset, then migrates the avatar call sites.

Two notes for reviewers:

- `MenuItemEntity` is deliberately narrow — a `UserAvatar` plus an always-on chevron. Rows needing a workspace avatar, an ID-driven `ReportActionAvatars`, or no chevron drop to raw composition rather than growing the preset.
- `MenuItem.Description` gained a `variant`. `placeholder` reproduces the legacy font bump that happened automatically when a row had no `title`, so title-less rows keep rendering at the standalone size.

### Fixed Issues

$ https://github.com/Expensify/App/issues/99408
PROPOSAL:

### Tests

1. Run `npm run storybook` and open the `Components/MenuItemComparison` story
2. Verify that in every card of the `Phase 2 — avatars` section, the `Legacy`, `Composable` and `Preset` renders are identical
3. Verify that the `Phase 1 — icon rows` cards below are still identical

Migrated rows, checked against `main` side by side:

1. Open Settings > Security > Add copilot, pick a member and continue, and verify the confirmation row shows their avatar, name and login
2. Open Settings > Profile > Status and verify each delegator row shows an avatar, display name and login, spaced evenly
3. Open Settings > Security, set a vacation delegate, and verify the row shows the delegate's avatar, name and login; clear it and verify the empty row shows `Vacation delegate` at the larger standalone font size
4. Start a new task from the global create menu and verify the `Assignee` and `Share` rows show avatars, labels, tooltips on hover, the `Required` hint while `Share` is empty, and the chevron
5. Open a workspace > Members > a member and verify the `Profile` row still navigates and is spaced from the section below
6. Open Settings > Security > multifactor authentication revoke page and verify each device row shows its title with a working `Revoke` button on the right
7. Send an invoice from the global create menu and verify the `Send from` field shows the workspace avatar, name and `Workspace`, and only shows a chevron when the sender can be changed

### Offline tests

Mostly N/A — a component refactor with no API or Onyx changes. One row keeps offline feedback:

1. Go offline
2. Open Settings > Security and set a vacation delegate
3. Verify the delegate row greys out with the pending style, and that the error row and its dismiss action still appear below it when the request fails

### QA Steps

N/A

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


war-in:
@codex review

war-in:
@MelvinBot review

src/pages/workspace/companyCards/assignCard/ConfirmationStep.tsx
```diff
@@ -191,13 +190,12 @@ function ConfirmationStep({route}: ConfirmationStepProps) {
...
                     </View>
-                    <MenuItem
+                    <MenuItemEntity
                         title={cardholderName}
                         description={cardholderEmail}
-                        icon={cardholder?.avatar ?? getDefaultAvatarURL({accountID: cardholderAccountID ?? CONST.DEFAULT_NUMBER_ID})}
-                        iconType={CONST.ICON_TYPE_AVATAR}
-                        shouldShowRightIcon
-                        pressableTestID={CONST.ASSIGN_CARD_CARDHOLDER_ROW_TEST_ID}
+                        avatarSource={cardholder?.avatar}
```
JakubKorytko: I think this changes what a cardholder with no uploaded avatar sees. Legacy resolved the default itself:

```tsx
icon={cardholder?.avatar ?? getDefaultAvatarURL({accountID: cardholderAccountID ?? CONST.DEFAULT_NUMBER_ID})}
```

`avatarSource={cardholder?.avatar}` does not do the same thing. `UserAvatar` calls `getAvatar`, which only resolves the account default when `isDefaultAvatar(source)` is true. That function starts with `typeof avatarSource === 'string'`, so it returns `false` for `undefined`. `getAvatar` then returns `undefined`, `useFallBackAvatar` trips on `!resolvedSource`, and the row shows the grey `FallbackAvatar` instead of the cardholder's coloured default.

Worth testing on a fresh account. Fix is either `avatarSource={cardholder?.avatar ?? getDefaultAvatarURL({accountID})}` here, or resolving it inside `MenuItemEntity` (see the comment there).

src/pages/tasks/DynamicNewTaskPage.tsx
```diff
@@ -183,25 +262,37 @@ function DynamicNewTaskPage() {
...
-                                shouldShowRightIcon={!task?.parentReportID}
-                                titleWithTooltips={shareDestination?.shouldUseFullTitleToDisplay ? undefined : shareDestination?.displayNamesWithTooltips}
-                                rightLabel={translate('common.required')}
+                            <TaskFieldRow
+                                label={translate('common.share')}
+                                avatars={
+                                    !!task?.shareDestination && (
+                                        <ReportActionAvatars
+                                            singleAvatarContainerStyle={[styles.actionAvatar]}
+                                            reportID={task.shareDestination}
```
JakubKorytko: Legacy rendered this avatar itself and passed the row's interaction state into it:

```tsx
<ReportActionAvatars
    subscriptAvatarBorderColor={getSubscriptAvatarBackgroundColor(isHovered, pressed, theme.hoverComponentBG, theme.buttonHoveredBG)}
    secondaryAvatarContainerStyle={[...hover/press background styles...]}
    ...
/>
```

Now the call site builds the avatar, so neither prop is passed, and `ReportActionAvatars` has no default for `subscriptAvatarBorderColor`. A share destination that is a `policy_expense_chat` renders a subscript avatar, so on hover its ring keeps the unhovered colour and shows as a seam. `ScheduleCallConfirmationPage` has the same omission, but that row is not interactive so it does not show there.

The hover state is in `MenuItemInteractionContext`, but taking the avatars as a ready-made `ReactNode` means this row can never reach it, so fixing it needs a prop change. Either `TaskFieldRow` takes `accountIDs`/`reportID` and renders `ReportActionAvatars` itself, or a `MenuItem.Avatars` leaf reads the context and forwards both props.

src/components/MenuItem/layout/MenuItemRoot.tsx
```diff
@@ -37,7 +39,7 @@ type MenuItemRootProps = PropsWithChildren &
         accessibilityLabel?: string;
     };
 
-function MenuItemRoot({children, onPress, isDisabled = false, sentryLabel, accessibilityLabel}: MenuItemRootProps) {
+function MenuItemRoot({children, onPress, isDisabled = false, sentryLabel, testID, accessibilityLabel}: MenuItemRootProps) {
```
JakubKorytko, line 42: Legacy `MenuItem` defaults `shouldCheckActionAllowedOnPress` to `true` and wraps the handler:

```tsx
onPress={shouldCheckActionAllowedOnPress ? callFunctionIfActionIsAllowed(onPressAction, isAnonymousAction) : onPressAction}
```

The presets keep that wrap, `Root` does not. So every row migrated to raw composition drops it: `InvoiceSenderField`, `VacationDelegateMenuItem`, `TaskFieldRow`, `IOURequestEditReportCommon`, `WorkspaceCompanyCardsSettingsPage`. An anonymous user probably cannot reach most of those screens, so it likely does not bite today, but it is a silent change that every future migration inherits.

Could `Root` apply `callFunctionIfActionIsAllowed` itself, with an opt-out prop? Then the presets stop repeating it and raw composition matches legacy by default.

src/components/MenuItem/presets/MenuItemAvatarNavigation.tsx
```diff
@@ -0,0 +1,75 @@
...
+        title: string;
+
+        /** Supporting line under the title — an email, address, or other secondary identifier */
+        description: string;
+
+        /** Account ID the avatar belongs to. Picks the default avatar when `avatarSource` is absent */
+        accountID: number;
+
+        /** The entity's avatar. Falls back to the default avatar for `accountID` when omitted */
+        avatarSource?: AvatarSource;
```
JakubKorytko, line 34: This doc does not match what `UserAvatar` does. `getAvatar` only resolves the `accountID` default when `isDefaultAvatar(source)` is true, and that returns `false` for `undefined`, so leaving `avatarSource` out gives the grey `FallbackAvatar`, not the account's default avatar.

Either reword the doc, or make it true:

```tsx
<UserAvatar
    source={avatarSource ?? getDefaultAvatarURL({accountID})}
    accountID={accountID}
/>
```

That also fixes `ConfirmationStep`. The beneficial owner rows keep their grey fallback either way since they pass `accountID={CONST.DEFAULT_NUMBER_ID}`.

src/stories/MenuItemComparison.stories.tsx
```diff
@@ -55,16 +105,518 @@ function Card({title, legacy, composable, preset}: {title: string; legacy: React
...
+                        title="Alex Reed"
+                        description="alex@example.com"
+                        titleWithTooltips={STORY_TOOLTIP_DETAILS}
+                        shouldShowRightIcon
+                        onPress={noop}
+                    />
+                }
+                composable={
+                    <MenuItem.Root onPress={noop}>
+                        <MenuItem.Label>Assignee</MenuItem.Label>
```
JakubKorytko: The story puts `MenuItem.Label` straight inside `Root`, but both real call sites wrap it:

```tsx
<View style={styles.mb2}>
    <MenuItem.Label>{label}</MenuItem.Label>
</View>
```

(`TaskFieldRow` and `ScheduleCallConfirmationPage`. Legacy applied `isIDPassed ? styles.mb2 : null` internally.) So the Legacy and Composable renders in this card are not identical, which is what test step 2 asks the reviewer to check, and label spacing ends up being the one thing the story does not cover. Same at line 602.

Simplest fix is to give `MenuItemLabel` its own `mb2` and drop the wrapper `View` from all three places. The spacing is not really a call-site decision.

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
```diff
@@ -1,29 +1,64 @@
...
+            style={[styles.textLabelSupporting, variantStyles[variant], styles.breakWord]}
+            numberOfLines={numberOfLines}
         >
             {children}
         </Text>
     );
 }
 
 export default MenuItemDescription;
+export {MENU_ITEM_DESCRIPTION_VARIANT};
```
JakubKorytko: `MENU_ITEM_DESCRIPTION_VARIANT` is not re-exported from `src/components/MenuItem/index.ts`, so call sites have to reach past the barrel:

```ts
import {MENU_ITEM_DESCRIPTION_VARIANT} from '@components/MenuItem/leaves/text/MenuItemDescription';
```

That is the only deep import into `leaves/` from outside the folder, and it will spread as more variants land. Re-exporting it from the barrel (say `MenuItem.DescriptionVariant`) keeps call sites on one import and leaves the internal layout free to move.

src/pages/iou/request/step/IOURequestEditReportCommon.tsx
```diff
@@ -309,20 +309,26 @@ function IOURequestEditReportCommon({
...
         }
 
         return (
-            <MenuItem
-                onPress={handleCreateReport}
-                title={translate('report.newReport.createReport')}
-                description={policyForMovingExpenses?.name}
-                icon={icons.Document}
-            />
+            <MenuItem.Root onPress={handleCreateReport}>
```
JakubKorytko, line 319: Icon plus title plus description with no chevron is now hand-composed three times in this PR: both rows here and `WorkspaceCompanyCardsSettingsPage`. `MenuItemAction` is the same shape minus the description.

Would a description slot on `MenuItemAction`, or its own preset, be better than raw composition at each call site? Otherwise the next migration copies these 10 lines again.

src/pages/workspace/members/WorkspaceMemberDetailsPage.tsx
```diff
@@ -427,14 +428,14 @@ function WorkspaceMemberDetailsPage({personalDetails, policy, route}: WorkspaceM
...
                             )}
-                            <MenuItem
-                                style={styles.mb5}
-                                title={translate('common.profile')}
-                                icon={icons.Info}
-                                onPress={navigateToProfile}
-                                shouldShowRightIcon
-                                pressableTestID="member-profile-menu-item"
-                            />
+                            <View style={styles.mb5}>
```
JakubKorytko, line 434: The presets have no style escape hatch, so this migration swaps `style={styles.mb5}` for a wrapper `View` just to carry spacing. Same pattern shows up around the labels in `TaskFieldRow` and `ScheduleCallConfirmationPage`.

An optional `style` on `MenuItemRoot`, forwarded into the `PressableWithFeedback` style array where legacy `containerStyle` already went, would drop the extra node and keep migrations mechanical.

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
```diff
@@ -1,29 +1,64 @@
...
 /** The supporting text block of a `MenuItem.Content` */
-function MenuItemDescription({children}: MenuItemDescriptionProps) {
+function MenuItemDescription({children, numberOfLines = 2, variant = MENU_ITEM_DESCRIPTION_VARIANT.SUPPORTING}: MenuItemDescriptionProps) {
     const styles = useThemeStyles();
+    const StyleUtils = useStyleUtils();
 
     useMenuItemAccessibilityLabel('description', String(children));
 
+    /** Typography applied on top of the shared supporting-label base, keyed by variant */
+    const variantStyles: MenuItemDescriptionVariantStyles = {
```
JakubKorytko: For a two-way branch this builds a whole record plus two type aliases (`MenuItemDescriptionVariant` is used elsewhere, `MenuItemDescriptionVariantStyles` only here), and both branches get built every render even though one is thrown away. A ternary is simpler and drops the alias:

```tsx
const variantStyle =
    variant === MENU_ITEM_DESCRIPTION_VARIANT.PLACEHOLDER
        ? [StyleUtils.getFontSizeStyle(variables.fontSizeNormal), StyleUtils.getLineHeightStyle(variables.fontSizeNormalHeight)]
        : styles.textLineHeightNormal;
```

Fine to leave as is if a third variant is coming soon.

src/pages/ReimbursementAccount/USD/BeneficialOwnerInfo/subSteps/CompanyOwnersListUBO.tsx
```diff
@@ -92,18 +87,20 @@ function CompanyOwnersListUBO({isAnyoneElseUBO, isUserUBO, handleUBOsConfirmatio
...
-                        iconHeight={40}
-                        interactive={false}
-                        shouldShowRightIcon={false}
-                        displayInDefaultIconColor
-                    />
+                    <MenuItem.Root>
+                        <MenuItem.Row>
+                            <MenuItem.Leading>
+                                <UserAvatar
+                                    source={icons.FallbackAvatar}
```
JakubKorytko: This row and the `MenuItemEntity` rows above it render the same shape two different ways, and only this one passes `source={icons.FallbackAvatar}`. That prop is a no-op anyway: `UserAvatar` sets `useFallBackAvatar` when `resolvedSource === defaultAvatars.FallbackAvatar`, so it renders the fallback either way.

If the only real difference is the missing chevron, an optional `shouldShowChevron` on `MenuItemEntity` would keep the list uniform, and `icons`/`useMemoizedLazyExpensifyIcons` could go away here like it did in `BeneficialOwnersList`.

src/components/MenuItem/presets/MenuItemEntity.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * and a secondary identifier below, that navigates.
+ */
+function MenuItemEntity({title, description, accountID, avatarSource, onPress, isDisabled = false, sentryLabel, testID}: MenuItemEntityProps) {
+    return (
+        <MenuItemRoot
+            onPress={callFunctionIfActionIsAllowed(onPress)}
+            isDisabled={isDisabled}
+            sentryLabel={sentryLabel}
+            testID={testID}
+            accessibilityLabel={[title, description].filter(Boolean).join(', ')}
```
JakubKorytko: `title` and `description` are both required `string`, so `filter(Boolean)` can never drop anything and `[title, description].join(', ')` does the same. If a caller can legitimately have no second line, make it `description?: string` and then the filter is doing real work.

src/components/DelegatorList.tsx
```diff
@@ -27,25 +29,30 @@ function DelegatorList({delegators, message}: DelegatorListProps) {
...
-                        avatarID={delegatorDetails?.accountID ?? CONST.DEFAULT_NUMBER_ID}
-                        icon={delegatorDetails?.avatar ?? icons.FallbackAvatar}
-                        iconType={CONST.ICON_TYPE_AVATAR}
-                        numberOfLinesDescription={1}
-                        containerStyle={[styles.pr2, styles.mt1]}
-                        interactive={false}
-                    />
-                );
-            })}
+            <View style={[styles.mt1, styles.gap1]}>
```
JakubKorytko, line 32: `mt1` plus `gap1` reproduce the old per-row `mt1` fine, but the old `containerStyle={[styles.pr2, styles.mt1]}` is gone. `containerStyle` landed on the pressable next to `popoverMenuItem`'s `paddingHorizontal: 20`, and the more specific `paddingRight` wins in Yoga, so rows used to have 28px of right padding and now sit flush at 20px. The one-line login gets about 8px more room before it truncates. Intentional, to line up with other rows?

src/components/MoneyRequestConfirmationList/sections/InvoiceSenderField.tsx
```diff
@@ -58,29 +61,49 @@ function InvoiceSenderField({selectedParticipants, isReadOnly, didConfirm, trans
...
-            description={translate('workspace.common.workspace')}
-            label={translate('workspace.invoices.sendFrom')}
-            isLabelHoverable={false}
-            interactive={!isReadOnly && !!canUpdateSenderWorkspace}
-            onPress={() => {
-                if (!transaction?.transactionID) {
-                    return;
+        <>
+            <View style={[styles.ph5, styles.mt2]}>
+                <MenuItem.Label>{translate('workspace.invoices.sendFrom')}</MenuItem.Label>
```
JakubKorytko: `MenuItem.Label` is outside `Root` here but inside `Root` in `ScheduleCallConfirmationPage` and `TaskFieldRow`. Both are right: this one matches legacy `isLabelHoverable={false}`, the others match the default `true`. But `MenuItemLabel`'s doc only says "The label above a menu item's `Row`", so nothing records that the placement is what decides whether the label picks up the row's hover background. Worth saying so in the component doc.

src/stories/MenuItemComparison.stories.tsx
```diff
@@ -24,27 +41,60 @@ const story: Meta<typeof MenuItem> = {
 
 const CARD_WIDTH = 360;
 
+/** Account and report the ID-driven (`iconAccountID`/`iconReportID`) avatar cases resolve against */
+const STORY_ACCOUNT_ID = 90210;
+const STORY_REPORT_ID = 'menuItemComparisonStoryReport';
+const STORY_POLICY_ID = 'menuItemComparisonStoryPolicy';
+
+Onyx.merge(ONYXKEYS.PERSONAL_DETAILS_LIST, {
```
JakubKorytko: These `Onyx.merge` calls run when the module is imported and nothing awaits them, so the seeding races the first render and happens even when a different story is open. A Storybook decorator or loader would make the ordering explicit and keep the side effect inside this story.

JakubKorytko (commented):
good split, and the typography matches legacy where it counts. I checked the placeholder variant against legacy `descriptionTextStyles` and they are the same.

Two things I would fix before merge. The cardholder avatar in `ConfirmationStep` now renders the grey fallback instead of the coloured default. And the subscript avatar no longer follows the row's hover state. Everything else is API shape and consistency.

JakubKorytko:
@MelvinBot could you double-check that all of my comments in the [review](https://github.com/Expensify/App/pull/98862#pullrequestreview-4970883974) are accurate?

JakubKorytko:
@MelvinBot there is more than 3 of my comments in the review, I think it is around 14 comments in total, could you double-check that all of my comments in the review are accurate?

src/components/MenuItem/layout/MenuItemRoot.tsx
war-in: That's against the composition-over-configuration rules. I'd rather prepare a preset than add such prop if it's not required. Those usages with `shouldCheckActionAllowedOnPress` and `isAnonymousAction` are on my list

src/pages/iou/request/step/IOURequestEditReportCommon.tsx
war-in: We definitely can't put the optional `description` on `MenuItemAction`. We discussed it last time with @roryabraham [here](https://github.com/Expensify/App/pull/97339#discussion_r3755500527)

I could prepare a `MenuItemActionWithDescription` preset if you find it reasonable

src/pages/workspace/members/WorkspaceMemberDetailsPage.tsx
war-in: Yes, but we actually would like to get rid of the style prop if possible
One of the future migration PRs will be focued on those styles directly so I'll revisit this comment there

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
war-in: I can't see that pattern across the repository. Please link me to any guides if they exist; otherwise, we should start a broader discussion in some public channel 😄

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
war-in: I removed unnecessary `MenuItemDescriptionVariantStyles`, thanks!
But the pattern with `variantStyles` was established during Button decomposition and is already used in the codebase.

@dariusz-biela could you jump in and share your thoughts on style variants

```tsx
const variantStyles = {
    [MENU_ITEM_DESCRIPTION_VARIANT.SUPPORTING]: styles.textLineHeightNormal,
    [MENU_ITEM_DESCRIPTION_VARIANT.PLACEHOLDER]: [StyleUtils.getFontSizeStyle(variables.fontSizeNormal), StyleUtils.getLineHeightStyle(variables.fontSizeNormalHeight)],
};
```

vs ternary Kuba suggested? 🙏

src/components/MenuItem/presets/MenuItemEntity.tsx
war-in: nice one! thanks ❤️

src/components/DelegatorList.tsx
war-in: I think the old `pr2` is wrong and shouldn't be used here. I contacted Shawn to confirm it

Before (main)
image


After (this PR)
image

And that change will be aligned with what we have for the empty row of `Vacation delegate`

image

src/components/MoneyRequestConfirmationList/sections/InvoiceSenderField.tsx
war-in: Actually, we [decided](https://github.com/Expensify/App/issues/96202#issuecomment-5330298133) to always show the label outside the pressable, so I'll fix those usages. Thanks for noticing!

src/stories/MenuItemComparison.stories.tsx
war-in: It's not applicable anymore as we decided to always place the label outside the pressable. I'll update the gallery

src/pages/ReimbursementAccount/USD/BeneficialOwnerInfo/subSteps/CompanyOwnersListUBO.tsx
war-in: We won't add `shouldShowChevron` as this is against composition-over-configuration rules. But I removed the unnecessary `FallbackAvatar`, thanks

src/pages/workspace/companyCards/assignCard/ConfirmationStep.tsx
war-in: I resolved it in MenuItemEntity directly, thanks!

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
dariusz-biela: Personally, I think this style is definitely more readable and easier to extend. 


> ```tsx
> const variantStyles = {
>     [MENU_ITEM_DESCRIPTION_VARIANT.SUPPORTING]: styles.textLineHeightNormal,
>     [MENU_ITEM_DESCRIPTION_VARIANT.PLACEHOLDER]: [StyleUtils.getFontSizeStyle(variables.fontSizeNormal), StyleUtils.getLineHeightStyle(variables.fontSizeNormalHeight)],
> };
> ```

Tennary will encourage nesting in future versions, which is why I personally favor using config objects in most cases.

war-in:
@MelvinBot Review

war-in:
> ⚠️ Cardholder grey avatar — still present

It's already handled for free by `MenuItemEntity`

image

> MenuItemWithLabel silently drops accessibilityLabel

Nice find; I forgot about it. Fixed!

src/components/MenuItem/presets/MenuItemField.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+        /** Whether the menu item is disabled */
+        isDisabled?: boolean;
+    };
+
+/** The field MenuItem preset — a form field whose value the user picks on another screen */
+function MenuItemField({label, value, isRequired = false, onPress, isDisabled = false, sentryLabel, testID, children}: MenuItemFieldProps) {
+    const {translate} = useLocalize();
+
+    if (!value && !!onPress) {
```
JakubKorytko: Two gaps in this branch:

- A field that is empty and non-interactive (no `onPress`) skips the placeholder and falls into the filled branch below with nothing to show. Legacy fell back to the label text as the description whether or not the row was pressable. Keying off `!value` alone would cover it.
- `children` are dropped whenever the empty branch wins. A caller passes `Leading` and `Content` and they silently vanish. Nothing in the type hints at that, so it is worth saying in the `children` doc.

The Share row in `DynamicNewTaskPage` hits the first one: once `task.parentReportID` is set `onPress` is undefined, so with `shareDestination` unresolved the row renders blank and the `Required` hint disappears.

src/pages/tasks/DynamicNewTaskPage.tsx
```diff
@@ -34,6 +40,28 @@ import {isEmptyObject} from '@src/types/utils/EmptyObject';
...
+    const theme = useTheme();
+    const styles = useThemeStyles();
+    const {isHovered, isPressed} = useMenuItemInteraction();
+    const {isInteractive} = useMenuItemConfig();
+
+    const borderColor = isPressed ? theme.buttonHoveredBG : theme.hoverComponentBG;
+
+    return (
+        <ReportActionAvatars
+            singleAvatarContainerStyle={[styles.actionAvatar]}
```
JakubKorytko: The subscript badge styling did not come across. Legacy also passed:

```tsx
secondaryAvatarContainerStyle={[
    StyleUtils.getBackgroundAndBorderStyle(theme.sidebar),
    pressed && interactive ? StyleUtils.getBackgroundAndBorderStyle(theme.buttonPressedBG) : undefined,
    isHovered && !pressed && interactive ? StyleUtils.getBackgroundAndBorderStyle(theme.border) : undefined,
]}
```

Without it the small workspace badge on the Share row loses its ring and stops reacting to hover and press. `subscriptAvatarBorderColor` only covers the single-avatar case.

src/components/MenuItem/presets/MenuItemEmptyField.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+/** The empty-field MenuItem preset — a form field the user has not filled in yet */
+function MenuItemEmptyField({description, onPress, children, isDisabled = false, sentryLabel, testID}: MenuItemEmptyFieldProps) {
+    return (
+        <MenuItemRoot
+            onPress={callFunctionIfActionIsAllowed(onPress)}
+            isDisabled={isDisabled}
+            sentryLabel={sentryLabel}
+            testID={testID}
+            accessibilityLabel={description}
```
JakubKorytko: This label only carries the description, so the `Required` hint passed as `children` is never announced. `RightLabel` also does not register itself the way `Title` and `Description` do, and passing `accessibilityLabel` here switches off that registry anyway. A screen reader user gets no signal the field is required.

src/components/MenuItem/presets/MenuItemWithLabel.tsx
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * The labeled MenuItem preset — a `Root` with a label above it. The label sits outside `Root`, so
+ * it stays out of the press target and does not pick up the row's hover background.
+ */
+function MenuItemWithLabel({label, onPress, isDisabled = false, sentryLabel, testID, accessibilityLabel, children}: MenuItemWithLabelProps) {
+    const styles = useThemeStyles();
+
+    return (
+        <View>
+            <View style={[styles.ph5, styles.pv1, isDisabled && styles.opacitySemiTransparent]}>
```
JakubKorytko: `pv1` is baked in, where legacy left the label's vertical spacing to the call site through `labelStyle`. `InvoiceSenderField` passed `labelStyle={styles.mt2}`, so that label goes from 8px above to 4px above and 4px below. Worth a side by side on the `Send from` field.

src/components/MenuItem/layout/MenuItemLeading.tsx
```diff
@@ -0,0 +1,20 @@
...
+import type {PropsWithChildren} from 'react';
+
+import React from 'react';
+import {View} from 'react-native';
+
+type MenuItemLeadingProps = PropsWithChildren;
+
+/**
+ * The leading cell of a `MenuItem.Row`. Sets no width of its own — it centers and sizes to its
+ * content, so an icon, avatar or spinner brings its own footprint.
```
JakubKorytko: isn't "footprint" a weird word here?

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
```diff
@@ -1,29 +1,61 @@
...
+
+import type {ValueOf} from 'type-fest';
+
 import React from 'react';
 
+const MENU_ITEM_DESCRIPTION_VARIANT = {
+    /** The small supporting-label look, for a description that sits under a title */
+    SUPPORTING: 'supporting',
+
+    /** Normal-size text, for a description standing in for a value the row does not have yet */
```
JakubKorytko: this reads awkward to me, "standing in"  is unusual, perhaps "temporary description" or "placeholder"?

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
```diff
@@ -1,29 +1,61 @@
...
 type MenuItemDescriptionProps = {
     /** Text to render as the description */
     children: string | number;
+
+    /** Maximum number of lines to render before the text is truncated */
+    numberOfLines?: number;
+
+    /**
+     * Typography variant. `supporting` (default) is the small label look; `placeholder` bumps the font to
+     * the normal size — use it on rows with no title, where the description carries the row on its own.
```
JakubKorytko: I think we should not describe the type itself here as it may change and no-one will change the description for props later

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
```diff
@@ -8,19 +8,31 @@ import convertToLTR from '@libs/convertToLTR';
...
+          children: string | number;
+
+          accessibilityLabel?: never;
+      }
+    | {
+          /** Element to render in place of plain text, e.g. a `DisplayNames` with per-name tooltips */
+          children: ReactElement;
+
+          /** Plain-text form of the title, contributed to the row's accessibility label */
+          accessibilityLabel: string;
```
JakubKorytko: imo this comment is essentially useless, in this union the second part does not have it so I think it's not about eslint

src/components/MenuItem/presets/MenuItemField.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import React from 'react';
+
+import MenuItemEmptyField from './MenuItemEmptyField';
+import MenuItemWithLabel from './MenuItemWithLabel';
+
+type MenuItemFieldProps = PropsWithChildren &
+    WithSentryLabel &
+    WithTestID & {
+        /** Name of the field. Stands in for the value while the field is empty, and moves up into the label once it has one */
```
JakubKorytko: "Stands in" -> "placeholder"?

src/components/MenuItem/presets/MenuItemField.tsx
```diff
@@ -0,0 +1,76 @@
...
+        label: string;
+
+        /** The picked value. This is what decides if the field is empty or filled */
+        value?: string;
+
+        /** Whether to show the `Required` hint. Only reaches the screen while the field is empty, since a filled field cannot be missing */
+        isRequired?: boolean;
+
+        /** Function to fire when the row is pressed. Omit to make the row non-interactive, which also drops the chevron */
```
JakubKorytko: I think it being non-interactive is kind of obvious from the fact that onPress is not passed - the chevron part is worth keeping though

src/components/MenuItem/presets/MenuItemField.tsx
```diff
@@ -0,0 +1,76 @@
...
+import MenuItemEmptyField from './MenuItemEmptyField';
+import MenuItemWithLabel from './MenuItemWithLabel';
+
+type MenuItemFieldProps = PropsWithChildren &
+    WithSentryLabel &
+    WithTestID & {
+        /** Name of the field. Stands in for the value while the field is empty, and moves up into the label once it has one */
+        label: string;
+
+        /** The picked value. This is what decides if the field is empty or filled */
```
JakubKorytko: "This is what decides if the field is empty or filled" is either ambiguous or useless, if it means that when value is empty the field is empty - that is obvious imo, and if it means something else then I don't know what it means 😄

src/components/MoneyRequestConfirmationList/sections/InvoiceSenderField.tsx
```diff
@@ -58,29 +59,45 @@ function InvoiceSenderField({selectedParticipants, isReadOnly, didConfirm, trans
...
+            onPress={
+                isInteractive
+                    ? () => {
+                          if (!transaction?.transactionID) {
+                              return;
+                          }
+                          Navigation.navigate(createDynamicRoute(DYNAMIC_ROUTES.MONEY_REQUEST_STEP_SEND_FROM.path));
+                      }
+                    : undefined
+            }
```
JakubKorytko: can't we extract? this is kinda ugly

src/pages/MultifactorAuthentication/RevokePage.tsx
```diff
@@ -31,6 +31,42 @@ const confirmPromptKeys = {
...
+    title: string;
+
+    /** Whether this row's revoke request is in flight */
+    isLoading: boolean;
+
+    /** Opens the confirmation modal for this row's set of devices */
+    onPress: () => void;
+};
+
+/** A non-interactive row naming a set of registered devices, with a `Revoke` button in the trailing cell */
```
JakubKorytko: "a non-interactive" seems contradictory with "onPress" prop

src/components/MenuItem/presets/MenuItemField.tsx
JakubKorytko: Two gaps in this branch:

- A field that is empty and non-interactive (no `onPress`) skips the placeholder and falls into the filled branch below with nothing to show. Legacy fell back to the label text as the description whether or not the row was pressable. Keying off `!value` alone would cover it.
- `children` are dropped whenever the empty branch wins. A caller passes `Leading` and `Content` and they silently vanish. Nothing in the type hints at that, so it is worth saying in the `children` doc.

The Share row in `DynamicNewTaskPage` hits the first one: once `task.parentReportID` is set `onPress` is undefined, so with `shareDestination` unresolved the row renders blank and the `Required` hint disappears.

src/pages/tasks/DynamicNewTaskPage.tsx
JakubKorytko: The subscript badge styling did not come across. Legacy also passed:

```tsx
secondaryAvatarContainerStyle={[
    StyleUtils.getBackgroundAndBorderStyle(theme.sidebar),
    pressed && interactive ? StyleUtils.getBackgroundAndBorderStyle(theme.buttonPressedBG) : undefined,
    isHovered && !pressed && interactive ? StyleUtils.getBackgroundAndBorderStyle(theme.border) : undefined,
]}
```

Without it the small workspace badge on the Share row loses its ring and stops reacting to hover and press. `subscriptAvatarBorderColor` only covers the single-avatar case.

src/components/MenuItem/presets/MenuItemEmptyField.tsx
JakubKorytko: This label only carries the description, so the `Required` hint passed as `children` is never announced. `RightLabel` also does not register itself the way `Title` and `Description` do, and passing `accessibilityLabel` here switches off that registry anyway. A screen reader user gets no signal the field is required.

src/components/MenuItem/presets/MenuItemWithLabel.tsx
JakubKorytko: `pv1` is baked in, where legacy left the label's vertical spacing to the call site through `labelStyle`. `InvoiceSenderField` passed `labelStyle={styles.mt2}`, so that label goes from 8px above to 4px above and 4px below. Worth a side by side on the `Send from` field.

src/stories/MenuItemComparison.stories.tsx
```diff
@@ -1,50 +1,107 @@
...
 import React from 'react';
 import {View} from 'react-native';
+import Onyx from 'react-native-onyx';
 
 /**
  * Grid comparison of the legacy `MenuItem` monolith, the new composable
  * `MenuItem.Root`/`Row`/... API, and the `MenuItemAction`/`MenuItemNavigation` presets.
  * Each card shows the same visual case built with every API that can currently express it.
+ *
+ * The "Phase 2" avatar sections lead, because those rows are what is under review, and they carry no
```
JakubKorytko: This paragraph is about the review cycle, not the story. "what is under review", "Phase 1 as the already-settled baseline" and the ordering rationale all go stale the moment this merges, and phase 3 will have to rewrite them again. The first paragraph above already says what the file is for. Something like "Sections are grouped by migration phase" would survive.

src/components/MenuItem/presets/MenuItemField.tsx
JakubKorytko: `!value && !!onPress` - shouldn't this be just `!value`? an empty field with no onPress falls into the filled branch and renders nothing, legacy always fell back to the label text. the Share row in DynamicNewTaskPage hits it once parentReportID is set - blank row, no Required hint.

also `children` get silently dropped in the empty branch, worth a word in the prop doc I think

src/pages/tasks/DynamicNewTaskPage.tsx
JakubKorytko, line 61: the subscript badge styling did not come across. Legacy also passed:

```tsx
secondaryAvatarContainerStyle={[
    StyleUtils.getBackgroundAndBorderStyle(theme.sidebar),
    pressed && interactive ? StyleUtils.getBackgroundAndBorderStyle(theme.buttonPressedBG) : undefined,
    isHovered && !pressed && interactive ? StyleUtils.getBackgroundAndBorderStyle(theme.border) : undefined,
]}
```

that kept the subscript badge's background matching the row, including on hover and press. Without it the badge keeps the default colour and stops tracking the row. `subscriptAvatarBorderColor` only covers the single-avatar case.

src/components/MenuItem/presets/MenuItemEmptyField.tsx
JakubKorytko, line 39: this label only carries the description, so the `Required` hint passed as `children` is never carried and I think it's important for the screen readers, no?

src/components/MenuItem/presets/MenuItemWithLabel.tsx
JakubKorytko, line 26: `pv1` is baked in, where legacy left the label's vertical spacing to the call site through `labelStyle`. `InvoiceSenderField` passed `labelStyle={styles.mt2}`, so that label goes from 8px above to 4px above and 4px below. Worth a side by side on the `Send from` field.

src/stories/MenuItemComparison.stories.tsx
JakubKorytko: This paragraph is about the review cycle, not the story. "what is under review", "Phase 1 as the already-settled baseline" and the ordering rationale all go stale the moment this merges, and phase 3 will have to rewrite them again. The first paragraph above already says what the file is for. Something like "Sections are grouped by migration phase" would survive.

src/components/MenuItem/layout/MenuItemLeading.tsx
JakubKorytko: isn't "footprint" a weird word here?

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
JakubKorytko: this reads awkward to me, "standing in"  is unusual, perhaps "temporary description" or "placeholder"?

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
JakubKorytko: I think we should not describe the type itself here as it may change and no-one will change the description for props later

src/components/MenuItem/leaves/text/MenuItemTitle.tsx
```diff
@@ -8,19 +8,31 @@ import convertToLTR from '@libs/convertToLTR';
...
+          /** Text to render as the title */
+          children: string | number;
+
+          accessibilityLabel?: never;
+      }
+    | {
+          /** Element to render in place of plain text, e.g. a `DisplayNames` with per-name tooltips */
+          children: ReactElement;
+
+          /** Plain-text form of the title, contributed to the row's accessibility label */
```
JakubKorytko: imo this comment is essentially useless, in this union the second part does not have it so I think it's not about eslint

src/components/MenuItem/presets/MenuItemField.tsx
JakubKorytko: "Stands in" -> "placeholder"?

src/components/MenuItem/presets/MenuItemField.tsx
JakubKorytko: "This is what decides if the field is empty or filled" is either ambiguous or useless, if it means that when value is empty the field is empty - that is obvious imo, and if it means something else then I don't know what it means 😄

src/components/MenuItem/presets/MenuItemField.tsx
JakubKorytko: I think it being non-interactive is kind of obvious from the fact that onPress is not passed - the chevron part is worth keeping though

src/components/MoneyRequestConfirmationList/sections/InvoiceSenderField.tsx
```diff
@@ -58,29 +59,45 @@ function InvoiceSenderField({selectedParticipants, isReadOnly, didConfirm, trans
...
-            disabled={didConfirm}
+            onPress={
+                isInteractive
+                    ? () => {
+                          if (!transaction?.transactionID) {
+                              return;
+                          }
+                          Navigation.navigate(createDynamicRoute(DYNAMIC_ROUTES.MONEY_REQUEST_STEP_SEND_FROM.path));
+                      }
+                    : undefined
```
JakubKorytko: can't we extract? this is kinda ugly

src/pages/MultifactorAuthentication/RevokePage.tsx
JakubKorytko: "a non-interactive" seems contradictory with "onPress" prop

JakubKorytko (commented):
some comments, overall nice one

src/pages/tasks/DynamicNewTaskPage.tsx
war-in: This is intended. `ReportActionAvatars` use `secondaryAvatarContainerStyle` in only one place - `DiagonalAvatars`. To render that branch, multiple `accountIDs` are needed so this path is dead on `DynamicNewTaskPage`

src/components/MenuItem/presets/MenuItemEmptyField.tsx
war-in: the `Required` hint was never passed in legacy MenuItem, so I skipped it here too

src/components/MenuItem/presets/MenuItemWithLabel.tsx
war-in: Yes, we're adding a bit more vertical space between the label and the rest of the item, but this will stay aligned across the entire app. It'll actually look better because the label won't touch the highlight anymore

https://github.com/user-attachments/assets/3fd293af-721b-4ae7-8eae-bb5b735f3690

src/components/MenuItem/presets/MenuItemField.tsx
war-in: I decided to remove this component entirely as it didn't conform to composition-over-configuration rules

war-in:
@situchan I think this one belongs to @Pujan92 as he reviewed the previous one, sorry for the ping

src/pages/tasks/DynamicNewTaskPage.tsx
```diff
@@ -185,26 +217,81 @@ function DynamicNewTaskPage() {
...
+                                            <MenuItem.Trailing>
+                                                <MenuItem.Chevron />
+                                            </MenuItem.Trailing>
+                                        )}
+                                    </MenuItem.Row>
+                                </MenuItemWithLabel>
+                            ) : (
+                                <MenuItemEmptyField
+                                    description={translate('common.share')}
+                                    onPress={navigateToShareDestination}
```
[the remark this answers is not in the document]
war-in: fixed - https://github.com/Expensify/App/pull/98862/commits/7192da4d449a3d9a0888465e187f99f91a565c55

roryabraham:
@codex review

src/pages/tasks/DynamicNewTaskPage.tsx
```diff
@@ -34,6 +41,28 @@ import {isEmptyObject} from '@src/types/utils/EmptyObject';
...
+    const {isHovered, isPressed} = useMenuItemInteraction();
+    const {isInteractive} = useMenuItemConfig();
+
+    const borderColor = isPressed ? theme.buttonHoveredBG : theme.hoverComponentBG;
+
+    return (
+        <ReportActionAvatars
+            singleAvatarContainerStyle={[styles.actionAvatar]}
+            subscriptAvatarBorderColor={isInteractive && (isHovered || isPressed) ? borderColor : undefined}
+            reportID={reportID}
```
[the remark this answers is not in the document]
war-in: But it never does, already answered [here](https://github.com/Expensify/App/pull/98862#discussion_r3852838966)

src/pages/tasks/DynamicNewTaskPage.tsx
```diff
@@ -185,26 +217,81 @@ function DynamicNewTaskPage() {
...
+                                <MenuItemWithLabel
+                                    label={translate('common.share')}
+                                    onPress={navigateToShareDestination}
+                                >
+                                    <MenuItem.Row>
+                                        <MenuItem.Leading>
+                                            <TaskFieldAvatar reportID={task?.shareDestination} />
+                                        </MenuItem.Leading>
+                                        <MenuItem.Content>
+                                            {shareDestination.shouldUseFullTitleToDisplay ? (
```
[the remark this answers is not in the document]
war-in: fixed - https://github.com/Expensify/App/pull/98862/commits/1809170ee628ab8c4fe6c931cf89350934458925

src/components/MenuItem/presets/MenuItemEntity.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+        /** Whether the menu item is disabled */
+        isDisabled?: boolean;
+    };
+
+/**
+ * The entity MenuItem preset — a tappable row led by a person's avatar, with their name as the title
+ * and a secondary identifier below, that navigates.
+ */
+function MenuItemEntity({title, description, accountID, avatarSource, onPress, isDisabled = false, sentryLabel, testID}: MenuItemEntityProps) {
```
roryabraham: It's unclear to me without reading the code what `Entity` means in this context. Can we come up with a better name?

src/components/MoneyRequestConfirmationList/sections/InvoiceSenderField.tsx
```diff
@@ -58,29 +59,44 @@ function InvoiceSenderField({selectedParticipants, isReadOnly, didConfirm, trans
...
-            style={styles.moneyRequestMenuItem}
-            labelStyle={styles.mt2}
-            titleStyle={styles.flex1}
-            disabled={didConfirm}
+            onPress={onPress}
+            isDisabled={didConfirm}
             sentryLabel={CONST.SENTRY_LABEL.REQUEST_CONFIRMATION_LIST.SEND_FROM_FIELD}
-        />
+        >
+            <MenuItem.Row>
```
roryabraham, line 78: Sorry if this should have been a question for a previous PR, but is there any case when we would have multiple `<MenuItem.Row>` nested within a single `<MenuItem.Root>`?

src/pages/MultifactorAuthentication/RevokePage.tsx
```diff
@@ -31,6 +31,42 @@ const confirmPromptKeys = {
...
+
+    /** Whether this row's revoke request is in flight */
+    isLoading: boolean;
+
+    /** Opens the confirmation modal for this row's set of devices */
+    onPress: () => void;
+};
+
+/** A row naming a set of registered devices, with a `Revoke` button in the trailing cell */
+function RevokeRow({title, isLoading, onPress}: RevokeRowProps) {
```
roryabraham: Let's try to keep to just 1 component per file

src/components/MoneyRequestConfirmationList/sections/InvoiceSenderField.tsx
war-in: Kinda, we will have 
```tsx
<MenuItem.Root>
    <MenuItem.Row>
        {...}
    </MenuItem.Row>
    {badge}
    {error}
</MenuItem.Root>
```

So we needed a layout component, so badge and error can appear below the main MenuItem content.
I didn't want to introduce it later during the migration because it would require changing already migrated call sites, but I agree this can be misleading until we add error handling. Sorry for the confusion!

src/pages/MultifactorAuthentication/RevokePage.tsx
war-in: fixed - https://github.com/Expensify/App/pull/98862/commits/f369507e7fb07f64b864ff5a4271694b2bd92138

src/components/MenuItem/presets/MenuItemEntity.tsx
war-in: fixed - https://github.com/Expensify/App/pull/98862/commits/8c4bce3e22b4d4eafe09afe1d99704d5f166b157

war-in:
@Pujan92 all yours 😄

src/pages/iou/request/step/IOURequestEditReportCommon.tsx
Pujan92: In that case we can add a new preset `MenuItemActionWithDescription` as the pattern seems to be used at multiple places.

src/pages/settings/Profile/CustomStatus/StatusPage.tsx
```diff
@@ -252,7 +252,7 @@ function StatusPage() {
                         />
                     ) : (
                         <>
-                            <Text style={[styles.mh5, styles.mb1]}>{translate('statusPage.setVacationDelegate')}</Text>
+                            <Text style={[styles.mh5]}>{translate('statusPage.setVacationDelegate')}</Text>
```
Pujan92: Can we keep the bottom margin?

Image

Image

war-in: Oh, good catch. Thank you! Fixed [here](https://github.com/Expensify/App/pull/98862/commits/6e6d3528d3cd371cdc66bed4949f09b51d482db1)

src/pages/ReimbursementAccount/NonUSD/BeneficialOwnerInfo/BeneficialOwnersList.tsx
```diff
@@ -49,21 +47,14 @@ function BeneficialOwnersList({handleConfirmation, ownerKeys, handleOwnerEdit}:
...
             return (
-                <MenuItem
+                <MenuItemAvatarNavigation
                     key={ownerKey}
                     title={`${ownerData.firstName} ${ownerData.lastName}`}
                     description={`${ownerData.street}, ${ownerData.city}, ${ownerData.state} ${ownerData.zipCode}`}
-                    wrapperStyle={[styles.ph5]}
-                    icon={icons.FallbackAvatar}
-                    iconType={CONST.ICON_TYPE_AVATAR}
+                    accountID={CONST.DEFAULT_NUMBER_ID}
```
Pujan92, line 54: Not sure whether this can be a regression or not, but earlier we were passing fallbackAvatar and now it renders a default avatar

Image

Pujan92:
Overall looks good. Only two comments https://github.com/Expensify/App/pull/98862#discussion_r3874058160 & https://github.com/Expensify/App/pull/98862/changes#r3872159099

src/pages/ReimbursementAccount/NonUSD/BeneficialOwnerInfo/BeneficialOwnersList.tsx
shawnborton: Hmm ideally we'd be using the initials avatars there, right?

Pujan92: No, These owners details is the free text and not tied with any account ID, so it's a fallback earlier.

war-in: yeah, this was a regression. I restored the previous behaviour here - https://github.com/Expensify/App/pull/98862/commits/1c5501eb9ba70d7b53d7aa655e2984b86049338b

src/pages/iou/request/step/IOURequestEditReportCommon.tsx
war-in: @roryabraham how do you feel about it? @Pujan92 could you suggest a shape of `MenuItemActionWithDescription` I wonder if we really should add such preset. There will be a lot of usages with descriptions, but some might require a leading icon, some won't.
Also, the trailing part will differ significantly between usages

war-in:
@Pujan92 how do you feel about not blocking on [this](https://github.com/Expensify/App/pull/98862/changes#r3879321735)? We can add such preset later and proceed with the next PRs now 😄

Pujan92 (approved):
LGTM!

war-in:
@roryabraham it's all yours

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
```diff
@@ -1,29 +1,58 @@
...
+
 type MenuItemDescriptionProps = {
     /** Text to render as the description */
     children: string | number;
+
+    /** Maximum number of lines to render before the text is truncated */
+    numberOfLines?: number;
+
+    /** The visual variant of the description, which controls its font size and line height */
+    variant?: MenuItemDescriptionVariant;
```
roryabraham: **blocker:** This is a configuration prop: split it into two components, `MenuItemDescriptionSupporting` and `MenuItemDescriptionPlaceholder`

really good evidence for why we should do it in this case: the supporting variant doesn't need to subscribe to `useStyleUtils`.

src/components/MenuItem/presets/MenuItemEmptyField.tsx
```diff
@@ -0,0 +1,54 @@
...
+        >
+            <MenuItemRow>
+                <MenuItemContent>
+                    <MenuItemDescription variant={MENU_ITEM_DESCRIPTION_VARIANT.PLACEHOLDER}>{description}</MenuItemDescription>
+                </MenuItemContent>
+                <MenuItemTrailing>
+                    {children}
+                    {!!onPress && <MenuItemChevron />}
```
roryabraham: **question:** What is the use-case for an empty field without an `onPress`?

roryabraham (changes requested):
Looking really close, just noticed a couple small things.

src/components/MenuItem/leaves/text/MenuItemDescription.tsx
war-in: fixed! - https://github.com/Expensify/App/pull/98862/commits/3d16db5b5cc94006a510bfed537dce19cc7e880b

src/components/MenuItem/presets/MenuItemEmptyField.tsx
war-in: I checked the callsites once again and noticed that optional `onPress` is there only because of the one callsite. I decided to use composition directly there and make the prop required

https://github.com/Expensify/App/pull/98862/commits/00663077720f42e4a6f2fca78b2ae90fe143d66a

war-in:
@roryabraham both comments addressed

roryabraham (approved):

kavimuru:
@war-in Just to ensure , we don't have to QA this PR

roryabraham:
I think regression tests are sufficient

PR #98846, opened 2026-08-18
refactor: remove unsafe Plaid connection assertions

KJ21-ENG:
### Explanation of Change

This PR removes unsafe type assertions from the Plaid connection flows used for workspace company cards and personal cards.

The web and native Plaid SDKs return institution IDs under different field names. A shared helper now reads the correct field without forcing either SDK response into the other shape. Plaid feed IDs are handled as plain strings, and their optional domain suffix is removed in one place.

Plaid behavior remains the same: successful connections keep the selected institution and accounts, existing feeds keep their base identifier, and cancelling or continuing without an institution does not start an import.

### Fixed Issues

$ https://github.com/Expensify/App/issues/94739
PROPOSAL: https://github.com/Expensify/App/issues/94739#issuecomment-4882049995

### Count change

12 → 0 targeted unsafe type assertions (12 removed).

### Changed files

- Plaid helpers and connection flows: `src/libs/PlaidUtils.ts`, `src/libs/actions/Plaid.ts`, `src/pages/settings/Wallet/PersonalCards/FixPersonalCardConnectionPage/index.native.tsx`, `src/pages/settings/Wallet/PersonalCards/steps/PlaidConnectionStep.tsx`, `src/pages/workspace/companyCards/addNew/PlaidConnectionStep.tsx`
- Plaid Onyx types: `src/types/onyx/AssignCard.ts`, `src/types/onyx/CardFeeds.ts`
- Tests: `tests/actions/PlaidTest.ts`, `tests/unit/PlaidUtilsTest.ts`

### Tests

- `tests/actions/PlaidTest.ts`
- `tests/unit/PlaidUtilsTest.ts`

Both focused suites passed: 14 tests.

- [x] Verify that no errors appear in the JS console

### QA Steps

1. Sign in as a workspace admin and open **Workspace settings → Company cards**.
2. Start a Plaid connection, select an institution and account, and complete the connection.
3. Verify the selected institution and accounts are imported normally.
4. Repair an existing Plaid feed and verify it refreshes without creating a duplicate.
5. Cancel the flow without selecting an institution and verify no import is attempted.
6. Repeat the relevant connection and cancellation checks for a personal card.

- [x] Verify that no errors appear in the JS console

### Offline tests

1. Start a Plaid connection while online.
2. Disconnect before completing it.
3. Verify no false success or invalid import occurs.
4. Reconnect and verify the connection completes normally.

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
[...]

PR #98879, opened 2026-08-18
refactor: remove unsafe type assertions from emoji, selection, and sound utilities

KJ21-ENG:
### Explanation of Change

Removes 12 unsafe type assertions from emoji search, formatted selection copying, and native sound loading.

The change keeps the existing emoji and native sound type contracts. It also keeps transformed DOM parent and sibling links coherent before serialization. SVG `foreignObject` and MathML integration-point children now serialize with paired HTML tags. That is an intentional correction to malformed serialized output, not a claim that every previous byte of output is unchanged.

### Fixed Issues

### Tests

Repeatable development test steps are below. The attached September 12 Chrome recording is partial evidence, not a completed retest of the latest main merge.

1. Open a report and use the emoji picker to search for `smile` and `thumbs up`. Select a skin-tone variant and verify the selected emoji appears correctly in the composer.
2. Send a message containing bold text, a link, and two lines. Select the rendered message, copy it, and paste into the composer. Verify text order, link, formatting, and line breaks are preserved.
3. Repeat copying from the composer, including nested formatting and an email address. Verify content is retained without extra blank lines or zero-width breaks in the email address.
4. On native Android and iOS with in-app sounds enabled, create and complete a task. Verify the completion sound plays and the app remains stable. Mute in-app sounds, repeat, verify no sound plays, then restore the setting.

### Offline tests

1. Load the app and report while online, then disconnect the network.
2. Repeat emoji selection and formatted copy/paste from Tests. Verify these local interactions match the online results.
3. Reconnect and verify the composer remains usable.

### QA Steps

Repeat Tests steps 1–4 on staging for the applicable web and native platforms. Compare any formatting or search failure against current main before attributing it to this PR.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

September 12 Chrome recording, before the latest main merge. It shows emoji selection and formatted copy/paste online and offline. Online smile search returned no results, and formatted paste lost bold formatting and changed line breaks; these observations have not been established as PR regressions. SVG/MathML fixtures and native sound behavior are not demonstrated in this recording.

https://github.com/user-attachments/assets/acfc1ed5-d42b-4e14-b5e0-0d162c4d9d55

</details>


src/libs/SelectionScraper/index.ts
```diff
@@ -12,6 +12,63 @@ import type GetCurrentSelection from './types';
 const markdownElements = new Set(['h1', 'strong', 'em', 'del', 'blockquote', 'q', 'code', 'pre', 'a', 'br', 'li', 'ul', 'ol', 'b', 'i', 's', 'mention-user']);
 const tagAttribute = 'data-testid';
```
ikevin127, line 16: 🟠 **`SelectionScraper/index.ts` — roughly 60 lines of dom-serializer-internals machinery were added to preserve two invalid-HTML outputs, in order to delete 4 assertions.**

The core refactor is right: `replaceNodes` used to return `{...dom, ...} as Element & DataNode`, a plain object spread that was not a real domhandler node and whose `parent` / `prev` / `next` still pointed into the *original* untransformed tree. Replacing that with real `cloneNode()` instances plus `installTransformedChildren` is a genuine improvement, and it is what removes the assertions.

But that fix also silently corrected a serialization bug, so the PR adds two shims to put the bug back.

I reproduced this with the repo's own `htmlparser2` and `dom-serializer`:

```
1) coherent parse -> render     : <svg><foreignObject><div></div></foreignObject></svg>
2) foreignobject name BEFORE    : foreignobject
   foreignobject name AFTER     : foreignObject
3) old spread-style render      : <svg><foreignObject><div/></foreignObject></svg>
```

And for the editor-collapse case the PR's own test pins:

```
coherent (span parented to mtext) : <math><mtext><span></span></mtext></math>
stale parent (removed <div>)      : <math><mtext><span/></mtext></math>
```

What is happening: `dom-serializer` decides HTML vs foreign (SVG/MathML) mode by reading `elem.parent.name`, and it *mutates* `elem.name` in place during render (`foreignobject` becomes `foreignObject`). Under the old code the children's `parent` pointed at the original, never-renamed node, so the integration-point lookup missed and the children serialized in XML mode. With a coherent tree the lookup hits and they serialize as HTML, which is correct.

`installForeignObjectRenderCompatibility` recreates the miss by cloning the `foreignObject` and re-parenting its children onto an unattached lowercase copy. `installEditorCollapseRenderCompatibility` does the same for children hoisted out of a collapsed editor div.

Two problems with paying that cost:

1. **The preserved output is not valid HTML, and it reaches the clipboard.** `useCopySelectionHelper.ts:19` calls `Clipboard.setHtml(selection, clipboardText)`, so this string is what lands in the user's clipboard as `text/html`. `<div/>` is not self-closing in HTML; every HTML parser treats it as an unclosed `<div>`. Paste a selection containing an SVG `foreignObject` into Gmail, Word, or Notion and the unclosed div can swallow the content that follows it. The new test `preserves pinned SVG foreignObject serialization` asserts that exact string, so the defect is now pinned by a green test.

2. **The shims depend on dom-serializer internals that are not public API**: that it mutates `elem.name` in place, and the exact contents of its private `foreignModeIntegrationPoints` set. Nothing here would fail loudly if a patch bump changed either. The comments explain the mechanism carefully but never say the behavior being preserved is wrong.

**Recommendation:** delete both shims and let the coherent tree produce the correct output. That removes about 60 lines, removes the dom-serializer coupling, and still removes all 4 assertions. The two test expectations become `<div></div>` and `<span></span>`, and the PR description gains one line saying SVG `foreignObject` and MathML integration-point children now serialize as well-formed HTML.

If you would rather keep it strictly behavior-preserving, that is defensible for a refactor, but then the comments need to say so plainly:

```ts
/**
 * Reproduces a known-incorrect serialization: children of SVG `foreignObject` are emitted in XML
 * mode (`<div/>`) rather than HTML mode. This is invalid HTML that reaches the clipboard via
 * `Clipboard.setHtml`. Preserved deliberately to keep this PR behavior-neutral. See <issue link>.
 */
```

Right now a future reader has no way to tell this is bug-compatibility rather than a requirement.

src/types/global.d.ts
```diff
@@ -12,6 +12,11 @@ declare module '*.jpg' {
     export default value;
 }
 
+declare module '*.mp3' {
+    const value: number;
+    export default value;
+}
+
```
ikevin127, line 20: 🟡 **`src/types/global.d.ts:15` — the `*.mp3` declaration is a global claim that is only true for Metro.**

Metro resolves asset imports to a numeric registry ID, so `number` is right on native. Web bundlers resolve them to a URL **string**. The neighbouring declarations avoid this: `*.png` and `*.jpg` both use `ImageSourcePropType`, which is honest on both platforms.

Today only `src/libs/Sound/index.native.ts` imports an `.mp3`, and the web sound path builds string URLs by hand (`index.ts:53`), so nothing is broken. But this is a global declaration, and the first web-side `import x from './foo.mp3'` will be typed `number` while holding a string, with no error anywhere.

`AudioSource` is `string | number | null | {...}` (`expo-audio/build/Audio.types.d.ts:16`), so the honest declaration costs nothing and needs no other change:

```ts
declare module '*.mp3' {
    // Metro resolves assets to a numeric registry ID; web bundlers resolve them to a URL string.
    const value: number | string;
    export default value;
}
```

**Worth naming what this change actually is**: the hand-written module declaration *is* the type claim that replaced `require(...) as AudioSource`. Nothing verifies it. That is the same shape as the overload trick from #98759, though far more defensible here since it matches four existing declarations in the same file.

Making it honest for both bundlers is the difference between a convention and a cast in disguise.

tests/unit/SelectionScraperTest.ts
```diff
@@ -0,0 +1,96 @@
...
+        expect(SelectionScraper.getCurrentSelection()).toBe('<svg><foreignObject><div/></foreignObject></svg><span>selected</span>');
+    });
+
+    it('retains coherent transformed MathML relationships', () => {
+        const documentCloneSpy = jest.spyOn(Document.prototype, 'cloneNode');
+        selectFixture('<math><mtext><div>first</div><span>middle</span><p>last</p></mtext></math>');
+
+        expect(SelectionScraper.getCurrentSelection()).toBe('<math><mtext><div>first</div><span>middle</span><p>last</p></mtext></math>');
+
+        const firstCloneResult = documentCloneSpy.mock.results.at(0);
```
ikevin127: 🟡 **`tests/unit/SelectionScraperTest.ts:52-64` — the MathML test asserts on implementation internals.**

```ts
const documentCloneSpy = jest.spyOn(Document.prototype, 'cloneNode');
...
const firstCloneResult = documentCloneSpy.mock.results.at(0);
```

This reaches into the transformed document by grabbing the return value of the *first* `Document.prototype.cloneNode` call. It encodes how many times the implementation clones and in what order. Reorder `parsedDom.cloneNode()` and `domRepresentation.cloneNode(true)` in `getCurrentSelection`, and this test fails without any behavior changing.

Since the other three tests already assert on rendered output, this one is the odd one out.

The parent/prev/next wiring it checks is worth testing, but it belongs on `installTransformedChildren` directly. Export it and test it as a pure function, and the spy disappears.

Separately, line 1:

```ts
import type * as SelectionScraperModule from '@libs/SelectionScraper/index.native';
...
jest.requireActual<typeof SelectionScraperModule>('@libs/SelectionScraper/index.ts').default;
```

This types the **web** module using the **native** module's type. They happen to match today, and `index.native.ts` is a four-line stub returning `''`. If the web module ever gains an export, the test's type silently lies. `src/libs/SelectionScraper/types.ts` already exports `GetCurrentSelection`, so use that:

```ts
jest.requireActual<{default: {getCurrentSelection: GetCurrentSelection}}>('@libs/SelectionScraper/index.ts').default;
```

src/libs/SelectionScraper/index.ts
```diff
@@ -12,6 +12,63 @@ import type GetCurrentSelection from './types';
...
 
+const installTransformedChildren = (parent: ParentNode, children: ChildNode[]) => {
+    const transformedParent = parent;
+    transformedParent.children = children;
+
+    for (const [index, child] of children.entries()) {
+        child.parent = transformedParent;
+        child.prev = index > 0 ? (children.at(index - 1) ?? null) : null;
```
ikevin127: 🟢 **`installTransformedChildren` has a no-op alias and a dead fallback.**

```ts
const transformedParent = parent;
transformedParent.children = children;
...
child.prev = index > 0 ? (children.at(index - 1) ?? null) : null;
```

`transformedParent` is the same object as `parent`, so the alias only obscures that the function mutates its argument. And for `index > 0`, `children.at(index - 1)` is always defined, so `?? null` never fires.

Same phantom-fallback pattern that got fixed in #98846 by destructuring with a default.

```ts
const installTransformedChildren = (parent: ParentNode, children: ChildNode[]) => {
    parent.children = children;
    for (const [index, child] of children.entries()) {
        child.parent = parent;
        child.prev = children[index - 1] ?? null;
        child.next = children[index + 1] ?? null;
    }
};
```

trjExpensify:
PR doesn’t need product input as a refactor PR. Unassigning and unsubscribing myself.

src/libs/SelectionScraper/index.ts
KJ21-ENG: Addressed in bce709d34cefd1fdaf5dedb7fd5a5fd391d2dcd8. I removed both compatibility shims and kept the transformed domhandler tree coherent. SVG foreignObject and MathML integration-point children now serialize with paired HTML tags. The tests now assert the corrected clipboard output.

src/types/global.d.ts
KJ21-ENG: Addressed in bce709d34cefd1fdaf5dedb7fd5a5fd391d2dcd8. The *.mp3 declaration now uses number | string, which covers Metro asset IDs and web URLs. The native sound map still narrows each entry to the scalar variants accepted by AudioSource.

tests/unit/SelectionScraperTest.ts
KJ21-ENG: Addressed in bce709d34cefd1fdaf5dedb7fd5a5fd391d2dcd8. The test no longer spies on Document.prototype.cloneNode. It tests installTransformedChildren directly and derives the web module type from the actual web implementation, so export-shape changes remain type-checked.

src/libs/SelectionScraper/index.ts
KJ21-ENG: Addressed in bce709d34cefd1fdaf5dedb7fd5a5fd391d2dcd8. I removed the transformedParent alias and now update parent.children directly. The sibling assignments use indexed neighbors with ?? null, so the dead conditional fallback is gone.

tests/unit/SelectionScraperTest.ts
```diff
@@ -0,0 +1,96 @@
+import type * as SelectionScraperModule from '@libs/SelectionScraper/index.native';
+
+import CONST from '@src/CONST';
+
+import {Document, Element} from 'domhandler';
+
+// cspell:ignore mtext
+
+// Selection scraping only exists in the web implementation; the native variant always returns an empty string.
```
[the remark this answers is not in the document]
KJ21-ENG: Addressed in bce709d34cefd1fdaf5dedb7fd5a5fd391d2dcd8. The comment now uses two sentences instead of joining the clauses with a semicolon.

KJ21-ENG:
Addressed all four review findings in bce709d34cefd1fdaf5dedb7fd5a5fd391d2dcd8. Ready for another look.

ikevin127:
@KJ21-ENG LGTM, but PR is in draft mode so I can't approve.

KJ21-ENG:
@ikevin127 Could you please take a look at this when you get a chance? Thanks!

ikevin127:
Sure, taking a second-review pass at the PR and will report on findings.

src/libs/SelectionScraper/index.ts
```diff
@@ -167,6 +178,7 @@ const getCurrentSelection: GetCurrentSelection = () => {
     return newHtml || '';
 };
 
+export {installTransformedChildren};
```
ikevin127: 🟡 **`src/libs/SelectionScraper/index.ts:181` - the new export makes this platform-split module asymmetric.**

Exporting the helper was the right call for testing it as a pure function, but `index.native.ts` does not export it, and TS resolves `@libs/SelectionScraper` to `index.ts` since there is no `moduleSuffixes` configured. So this typechecks and is `undefined` at runtime on iOS and Android:

```ts
import {installTransformedChildren} from '@libs/SelectionScraper';
```

This repo takes the opposite approach elsewhere, including in a file this PR touches. `Sound/index.native.ts:56` carries `function clearSoundAssetsCache() {}`, an empty stub whose only job is keeping the native export surface identical to `Sound/index.ts:97`. `SelectionScraper` cannot do that here, since a native stub for this helper would be nonsense.

Cleanest fix is to lift the helper out of the platform pair entirely:

```ts
// src/libs/SelectionScraper/installTransformedChildren.ts
import type {ChildNode, ParentNode} from 'domhandler';

const installTransformedChildren = (parent: ParentNode, children: ChildNode[]) => { ... };

export default installTransformedChildren;
```

`index.ts` imports it, `index.native.ts` stays a four-line stub, the test imports one honest path with no `requireActual`, and the platform modules keep matching surfaces.

ikevin127:
🟡 **PR description still claims behavior is preserved, and the QA steps do not cover what changed.**

This was the second half of the first pass shim comment and it did not land. The body still reads "Existing search, copy/paste, and sound behavior is preserved", which is now the one inaccurate sentence in it.

The shim removal deliberately changed clipboard output, and the QA steps only mention bold, link, and line break. Nobody testing from that list will touch the path this PR actually changed. Suggested edits:

```
Explanation of Change: ...Children of SVG `foreignObject` and MathML integration points
now serialize as well-formed HTML (`<div></div>` rather than `<div/>`), correcting invalid
markup that previously reached the clipboard via `Clipboard.setHtml`.

QA step: Copy a selection spanning a message that contains an inline SVG, paste into Gmail
or Notion, and verify no content after the SVG is swallowed.
```

##

@KJ21-ENG Please tag me as soon as you address this comment and the one above so I can do a final pass and test. Also please merge latest main because at the moment it's behind in terms of version and that's blocking me from testing on dev since I get the "Update available" blocking screen.

src/libs/SelectionScraper/index.ts
KJ21-ENG: The web-only `installTransformedChildren` helper is now in: `src/libs/SelectionScraper/installTransformedChildren.ts`

The platform entry point no longer exports this web-only helper. Both platform modules continue to expose only the existing default `getCurrentSelection()` API, and the focused unit test imports the helper directly from its support path. This keeps the native surface free of the web-only DOM implementation while preserving the transformed-parent and sibling-link behavior.

KJ21-ENG:
The PR description was corrected so it no longer claims that every prior serialized output byte is preserved. It now identifies the SVG `foreignObject` and MathML integration-point paired-tag behavior as an intentional correction.

ikevin127:
@KJ21-ENG PR looks good, thanks for addressing the review comments.

**❌ There are two blockers at the moment**: 
1. Looks like you pushed changes to `‎.github/actions/javascript/` which should not be pushed as this PR has no business in that area, CI is 🔴.
2. The PR Author Checklist is incomplete, CI is 🔴.

Let me know once the blockers are resolved and I'll do a final pass 🙌

KJ21-ENG:
@ikevin127 Ah my bad! It looks like some generated GitHub Actions files were unintentionally included while I was testing locally (they were for my fork, I messed up heads) . I’ve removed them, confirmed the PR contains only the intended changes.

KJ21-ENG:
FYI, failing Jest tests are unrelated to this PR. They are still on the latest `main` as well. I will keep an eye on it. Once it's fixed on `main`, I will merge it here.

KJ21-ENG:
> @KJ21-ENG PR looks good, thanks for addressing the review comments.
> 
> **❌ There are two blockers at the moment**:
> 
> 1. Looks like you pushed changes to `‎.github/actions/javascript/` which should not be pushed as this PR has no business in that area, CI is 🔴.
> 2. The PR Author Checklist is incomplete, CI is 🔴.
> 
> Let me know once the blockers are resolved and I'll do a final pass 🙌

Both Done BTW

tests/unit/SelectionScraperTest.ts
```diff
@@ -0,0 +1,80 @@
+import type * as SelectionScraperModule from '@libs/SelectionScraper/index.native';
```
ikevin127: 🟡 **`tests/unit/SelectionScraperTest.ts:1` - The native-type-for-web-module problem is back**

This is the exact defect from the first pass, which he fixed in `bce709d` and has now reverted:

```diff
-import type * as SelectionScraperWebModule from '@libs/SelectionScraper/index';
+import type * as SelectionScraperModule from '@libs/SelectionScraper/index.native';
```

The `requireActual` call still loads `@libs/SelectionScraper/index.ts`, so the **web** module is once again typed with the **native** module's type.

It compiles today only because moving the helper out made the two surfaces identical again, which is precisely what makes it fragile: the next export added to `index.ts` puts the test's type silently out of sync with what it loads, and `index.native.ts` is a four-line stub returning `''`.

There was no reason to revert it, probably AI mishap. I confirmed TS resolves `@libs/SelectionScraper/index` to the web file (no `moduleSuffixes` in `tsconfig.json` or in the inherited `expo/tsconfig.base`), and ESLint was green on the previous form.\

One-line fix, back to what he already had:

```ts
import type * as SelectionScraperWebModule from '@libs/SelectionScraper/index';
...
const {default: SelectionScraper} = jest.requireActual<typeof SelectionScraperWebModule>('@libs/SelectionScraper/index.ts');
```

ikevin127:
🟡 **The `### Tests` section was dropped from the description.**

The rewrite replaced it rather than adding to it. The body now runs Explanation, Fixed Issues, Count change, QA Steps, Offline tests, with no `### Tests` anywhere, while `.github/PULL_REQUEST_TEMPLATE.md:32` requires it and line 41 of his own checklist is ticked for "I added steps for local testing in the `Tests` section". The previous version had it:

```
### Tests
- `npm test -- tests/unit/SelectionScraperTest.ts` - 4 tests passed.
- `npm run typecheck` - passed.
```

Worth restoring before merge so the section order matches the template and the ticked box is true.

tests/unit/SelectionScraperTest.ts
```diff
@@ -0,0 +1,80 @@
...
+    it('installs transformed children with coherent parent and sibling links', () => {
+        const parent = new Element('div', {});
+        const first = new Text('first');
+        const middle = new Text('middle');
+        const last = new Text('last');
+        const children = [first, middle, last];
+
+        installTransformedChildren(parent, children);
+
+        expect(parent.children).toBe(children);
```
ikevin127: 🟢 **`tests/unit/SelectionScraperTest.ts:54` - `expect(parent.children).toBe(children)` still asserts array identity.** Carried over from the last pass. It pins that the helper assigns the caller's array rather than copying it, so a defensive copy would fail the test with every DOM relationship still correct.

`toEqual` is the assertion that matches the intent. The nine `toBe` calls below it are correct, those really are node identity checks.

src/libs/Sound/index.native.ts
```diff
@@ -6,12 +11,12 @@ import {createAudioPlayer, setAudioModeAsync} from 'expo-audio';
 import {getIsMuted, SOUNDS, withMinimalExecutionTime} from './BaseSound';
 
 // Sound assets must be required at compile time
```
ikevin127: 🟢 **`src/libs/Sound/index.native.ts:13` - comment is still stale.** "Sound assets must be required at compile time" predates the switch to static imports.

ikevin127:
@KJ21-ENG Completed PR Reviewer Checklist, tests well. Dropped (4) more comments above after the final-pass review, these should be addressed before we merge - including the unrelated failing jest unit test.

Tag me once done 🙌

KJ21-ENG:
@ikevin127 Addressed the four comments in 95407e7d155:

- The test now imports the web module type used by `jest.requireActual`.
- The children-array assertion uses `toEqual`; the parent/sibling node identity assertions remain.
- Updated the sound comment to match static imports.
- Restored `### Tests` and the template section order in the PR body, preserving the existing recording and clarifying its coverage.

Also merged latest main

ikevin127 (approved):
**🟢 LGTM** - Thanks for addressing all comments 🙌

blimpich (approved):
Looks good. Nice job @KJ21-ENG and @ikevin127 👍

jponikarchuk:
The Deploy Blocker issue https://github.com/Expensify/App/issues/101346  is repro on all PRs. Could you please double check if it's not introduced by https://github.com/Expensify/App/pull/98879

KJ21-ENG:
Nah, I checked #98879’s diff and the attached logs. This points to #100800: browser Back restores the Insights route without params, and `route.params.dashboardID` throws the reported error. That code was already present before #98879 merged, and our PR doesn’t modify Insights or navigation.

cc: @blimpich @ikevin127

blimpich:
Agreed

PR #98887, opened 2026-08-18
Add a one-click deeplink that signs the user in and creates a Submit workspace

abzokhattab:
### Explanation of Change

There's no way to link directly to "start with Submit" from an email today: creating a Submit workspace requires a signed-in user to pick the intent in the onboarding UI (or click "Get started" in Old Dot pricing), neither of which is reproducible with a plain URL.

This PR adds an `intent` param to the onboarding route so a single link produces "click link → sign in → Submit workspace created". The accepted values live in the new `CONST.ONBOARDING_INTENTS`. They're short, stable aliases rather than the internal onboarding choice strings (`newDotEmployer`), so renaming those later can't break links already sitting in an inbox. The link itself is assembled outside the app, so nothing here builds the URL.

The intent survives the logged-out → logged-in transition by riding along in the deeplink itself, the same mechanism `exitTo=workspace/new` already relies on. `getOnboardingIntentFromUrl` reads it either from a direct link (`/onboarding?intent=submit`) or from the `exitTo` of a transition or magic link, encoded or not.

**Who the link acts on.** Per the Slack discussion on the issue, this is going out to existing users, so it only acts on recipients who have completed guided setup. For them `ApplySubmitOnboardingIntent` creates the workspace through `useAutoCreateSubmitWorkspace` — the same hook the onboarding purpose page and the Submit plan welcome modal already use — and skips `CompleteGuidedSetup`, so their recorded onboarding choice is left alone. Anyone who still has onboarding ahead of them is left to it untouched, since that flow already offers the Submit outcome. That fallback is why nothing here suppresses or reroutes onboarding, and idempotency comes free with the hook, which already skips creation when there's an editable group workspace, an onboarding policy ID, or restricted policy creation.

**Where they land** is not encoded in the link. The hook finishes by calling `navigateToSubmitWorkspaceAfterOnboardingWithMicrotaskQueue` — today Spend > Expenses with the side panel open on #admins — so changing that destination later is a one-line change and every link already sitting in an inbox keeps working.

On the route name: `onboarding?intent=submit` names a flow most recipients never see, but the param genuinely is an onboarding intent (it resolves to the same `CONST.ONBOARDING_CHOICES.EMPLOYER` the purpose step would set), and for recipients who haven't onboarded it's literally where they land. Happy to rename if you'd prefer something else.

### Fixed Issues

$ https://github.com/Expensify/App/issues/98644
PROPOSAL:

### Tests

On native, force-close the app before opening a link so it arrives as a cold launch; the intent is applied at most once per app process. The native commands target the standalone build, which is the one that registers the `new-expensify` scheme — on Android the trailing package keeps the intent off a HybridApp build if you have both installed.

1. **Signed-out recipient (the email-link case)**
    1. Sign out of New Expensify.
    2. Request a magic link for an account that has completed onboarding and has no Submit workspace, then open it:
        - **Web:** `https://dev.new.expensify.com:8082/v/<accountID>/<validateCode>?exitTo=onboarding%3Fintent%3Dsubmit`
        - **iOS:** run `xcrun simctl openurl booted "new-expensify://v/<accountID>/<validateCode>?exitTo=onboarding%3Fintent%3Dsubmit"` in terminal
        - **Android:** run `adb shell am force-stop com.expensify.chat.dev` then `adb shell am start -W -a android.intent.action.VIEW -d "new-expensify://v/<accountID>/<validateCode>?exitTo=onboarding%3Fintent%3Dsubmit" com.expensify.chat.dev` in terminal
    3. Verify you're signed in and land on Spend > Expenses with the side panel open on the #admins room. Verify the onboarding flow never appears.
    4. Go to Workspaces and verify exactly one workspace was created and its plan is Submit.
2. **Signed-in recipient**
    1. On another account that has finished onboarding and has no Submit workspace, open the link:
        - **Web:** `https://dev.new.expensify.com:8082/onboarding?intent=submit`
        - **iOS:** run `xcrun simctl openurl booted "new-expensify://onboarding?intent=submit"` in terminal
        - **Android:** run `adb shell am force-stop com.expensify.chat.dev` then `adb shell am start -W -a android.intent.action.VIEW -d "new-expensify://onboarding?intent=submit" com.expensify.chat.dev` in terminal
    2. Verify the same outcome as test 1, and that the Submit plan welcome modal does not appear.
3. **Recipient who never completed onboarding**
    1. On a brand new account that has not been through guided setup, open the link.
    2. Verify the normal onboarding flow appears and behaves exactly as it does on `main` — no step is skipped and no workspace is created behind it.
4. **Idempotency**
    1. With the workspace from test 1 or 2 in place, open the same link again. Verify no second workspace is created and onboarding does not start.
    2. On an account that already belongs to a Team or Corporate workspace, open the link. Verify no Submit workspace is created.

### Offline tests

Creating the workspace is an optimistic Onyx write, so offline behaviour matches the existing onboarding purpose flow: the workspace appears immediately with a pending state and the request is queued.

### QA Steps

Same as the Tests section above.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos

<details>

https://github.com/user-attachments/assets/5a47db1c-57d6-4491-a9d3-da7857f6b591

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/769fc071-5458-4806-9a1d-050ecb2bddd0

https://github.com/user-attachments/assets/78830a0a-f3ba-47ed-8533-3b94d857584b

https://github.com/user-attachments/assets/57614478-8d71-4649-a266-39708180f731

</details>

<details>

https://github.com/user-attachments/assets/e92e786e-9fd7-4246-afd8-616af93cab90

https://github.com/user-attachments/assets/37a9c0ab-fe27-4b03-9bd4-d0bd499af790

</details>

<details>

https://github.com/user-attachments/assets/817fd0a7-57c4-4680-b722-c0a5e2b1f7ab

https://github.com/user-attachments/assets/4e7b148c-4807-47c1-bb98-a88624eb0547

https://github.com/user-attachments/assets/a2c8ee24-1abf-43e8-8c62-c8b50e4b9e19

https://github.com/user-attachments/assets/01d7df7f-124e-437a-b566-63800839c147

</details>


abzokhattab:
@MelvinBot @codex review the PR, find bugs or regressions

abzokhattab:
Thanks both — three findings between you, two of which were real native bugs. Fixed in 3215c16b871.

### Fixed: custom-scheme URLs lost the route (Codex P1 #1 / Melvin #1)

Confirmed and reachable, not just latent: `new-expensify://` is a registered deeplink prefix in [`prefixes.ts`](https://github.com/Expensify/App/blob/main/src/libs/Navigation/linkingConfig/prefixes.ts#L9). In `new-expensify://onboarding?intent=submit` the route sits exactly where a host sits in a web URL, so stripping the origin discarded it and the intent came back `undefined`.

The origin is now stripped differently per scheme — host dropped for `http(s)`, scheme only for custom schemes, plus the `app://-/` placeholder host. Anchoring the host on `[^/?#]*` also fixes `https://dev.new.expensify.com:8082/...`, where the port previously survived. Four new test cases cover these.

### Fixed: warm-app links were never seen (Codex P1 #2)

Also correct. `InitialURLContextProvider` only calls `getInitialURL()` once, and the warm listener in `DeepLinkHandler` records into that context [only for secure links](https://github.com/Expensify/App/blob/main/src/DeepLinkHandler.tsx#L147), so a link opened against an already-running app left `urlAtMount` stale and the handler never fired.

Extended that same condition to onboarding-intent links rather than adding a parallel mechanism — it is the precedent already in place for exactly this problem.

### Not changing: the migration NVP write (Melvin #2)

Correct that `setSubmitMigrationModalShown()` runs before `autoCreateSubmitWorkspace` decides whether to create anything, so a recipient who already owns a Team/Corporate workspace gets the NVP without a workspace.

Leaving it. Making it conditional means duplicating `!isRestrictedPolicyCreation && !onboardingPolicyID && !hasEditableGroupPolicy` in the handler, and that copy would drift from the hook. It also reads as correct to me semantically: the NVP records that the user was offered the Submit migration, and clicking the link *is* that offer, delivered over email instead of the modal. Suppressing a later nudge for someone who already took the emailed one seems right. Happy to revisit if you disagree.

Unit tests, lint, typecheck and the React Compiler check are green. Native still needs a manual pass, which is where all three of these live.

abzokhattab:
@MelvinBot @codex review the PR, find bugs or regressions

abzokhattab:
> **P1** Consume the intent before a later account signs in

Good catch, this one is real. Fixed in f1887b48a9a.

`InitialURLContextProvider` sits at the top of `App.tsx`, above the navigator and the auth boundary, and nothing ever clears `initialURL`. Signing out and into a second already-onboarded account therefore remounts `ApplySubmitOnboardingIntent` with the original `intent=submit` still readable, and the `useRef` guard is reinitialised along with it — so that account gets a Submit workspace it never asked for.

The guard is now module scope, so it survives the remount and the intent applies at most once per app process.

I went with that over clearing `initialURL` deliberately. Consuming it is the more precise fix and would keep a genuinely new warm link working, but `initialURL` is shared with the secure-link check in `useOnboardingFlow`, `useAbsentPageSpan`, `ScreenWrapper` and `LogOutPreviousUserPage`, and I can't verify I wouldn't disturb one of those on HybridApp. The cost of the contained fix is that clicking the link a second time in the same process is a no-op, which is acceptable given the user already has the workspace by then and creation was idempotent anyway.

Worth noting the same replay characteristic already exists for the secure-link signal in `useOnboardingFlow`, which reads `initialURL` the same way. It's harmless there because it only suppresses onboarding rather than writing anything, so I've left it alone.

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/ApplySubmitOnboardingIntent.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    // Waiting on HAS_LOADED_APP keeps the eligibility checks inside useAutoCreateSubmitWorkspace from running
+    // against a half-populated store, where they would miss an existing workspace and create a duplicate.
+    const [hasLoadedApp] = useOnyx(ONYXKEYS.HAS_LOADED_APP);
+    const [isOnboardingCompleted] = useOnyx(ONYXKEYS.NVP_ONBOARDING, {selector: hasCompletedGuidedSetupFlowSelector});
+    const [isSupportalSession] = useOnyx(ONYXKEYS.SESSION, {selector: isSupportalSessionSelector});
+
+    useEffect(() => {
+        if (hasAppliedIntent || !hasLoadedApp || isOnboardingCompleted === undefined || isSupportalSession) {
+            return;
```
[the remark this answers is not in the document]
abzokhattab: Good catch, fixed in 418fd88.

`isSupportalSession` was sitting in the same bail-out as `!hasLoadedApp` and `isOnboardingCompleted === undefined`, but those two mean "not ready yet, come back" while supportal means "never do this". Taking the wait path left the intent live, and since `initialURL` is sticky for the life of the process, the next normal sign-in would act on it — the same cross-account replay the module flag was added to close. It now sits below `hasAppliedIntent = true` alongside `!isOnboardingCompleted`.

I checked the opposite risk before moving it: `getShortLivedLoginParams` passes `false` for ordinary logins, so a legitimate intent can't be swallowed by a transient true.

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/ApplySubmitOnboardingIntent.tsx
```diff
@@ -0,0 +1,55 @@
...
+        }
+        hasAppliedIntent = true;
+
+        if (!isOnboardingCompleted) {
+            return;
+        }
+
+        // The deeplink delivers the same outcome as the Submit plan welcome modal, so keep that modal from opening too.
+        setSubmitMigrationModalShown();
```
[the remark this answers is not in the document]
abzokhattab: Agreed, fixed in 418fd88.

The race is worse than just late: `scheduleSubmitPlanWelcomeModalEvaluation` runs on a microtask off the guard's `HAS_LOADED_APP` subscription, and microtasks drain before React flushes passive effects, so no React-side timing wins it — `useLayoutEffect` would lose too. The audiences also overlap heavily, since the guard wants EMPLOYER intent, completed onboarding and no group policies, which describes the same dormant users this link targets.

So the deeplink now tells the guard directly instead of racing it: `suppressWelcomeModalForSubmitDeeplink()` is called from a mount effect in `ApplySubmitOnboardingIntent`, and the guard checks that flag in both its proactive and reactive paths. Mount happens well before `OpenApp` resolves, so it lands ahead of the microtask. `setSubmitMigrationModalShown()` stays for durability across sessions. Reset is folded into `resetSessionFlag`, and there are two new tests, one per guard path.

One deliberate limitation worth flagging: the flag stays latched for the rest of the process, so if someone opens the link, signs out, and a different modal-eligible user signs in without a reload, that user won't see the modal until the next launch. Clearing it once the intent resolves reintroduces the race in a nastier form, because `autoCreateSubmitWorkspace`'s own navigation re-triggers `evaluate` in the window before the NVP write lands. A suppressed modal for one session seemed the cheaper failure, but happy to revisit.

iwiznia:
@hungvu193 ping me when you approve this please

hungvu193:
> @hungvu193 ping me when you approve this please

Sure

hungvu193:
@codex review

src/DeepLinkHandler.tsx
```diff
@@ -142,10 +143,10 @@ function DeepLinkHandler({onInitialUrl}: DeepLinkHandlerProps) {
...
-            // A Submit-via-PDF secure access link can arrive while the app is already running (warm), where
-            // getInitialURL() is empty. Record it so onboarding suppression has a session-sticky signal, the same
-            // way the cold path does via onInitialUrl above. Scoped to secure links so other deep links are unaffected.
-            if (hasSecureLinkKey(state.url)) {
+            // A Submit-via-PDF secure access link, or an onboarding deeplink carrying an intent, can arrive while the
+            // app is already running (warm), where getInitialURL() is empty. Record it so the handlers reading the
+            // initial URL have a session-sticky signal, the same way the cold path does via onInitialUrl above.
+            if (hasSecureLinkKey(state.url) || getOnboardingIntentFromUrl(state.url)) {
                 onInitialUrl(state.url as Route);
             }
```
hungvu193: @MelvinBot I remember we already had logic to redirect to url after login, do we need this logic here?

src/libs/getOnboardingIntentFromUrl.ts
```diff
@@ -0,0 +1,38 @@
...
+ * carries the intent across the logged-out to logged-in transition, since the deeplink outlives the sign-in itself.
+ */
+import type {OnboardingIntent} from '@src/CONST';
+import CONST from '@src/CONST';
+import ROUTES from '@src/ROUTES';
+
+import {getRouteFromLink} from './ReportUtils';
+import {getSearchParamFromPath} from './Url';
+
+const ONBOARDING_INTENT_VALUES = new Set<string>(Object.values(CONST.ONBOARDING_INTENTS));
```
hungvu193: Why do we need to convert it to a Set here? We only use this `ONBOARDING_INTENT_VALUES` inside `isOnboardingIntent`, I'd prefer to inline it.

```typescript
Object.values(CONST.ONBOARDING_INTENTS).includes(value);
```

src/hooks/useOnboardingDeeplinkIntent.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Resolves the onboarding outcome requested by the deeplink this session was opened with, e.g. `onboarding?intent=submit`.
+ *
+ * The URL is latched at mount because the app rewrites it as soon as it navigates. The initial URL covers native,
+ * where the browser URL is empty, and links opened while the app is already running.
+ */
+function useOnboardingDeeplinkIntent(): OnboardingIntent | undefined {
+    const {initialURL} = useInitialURLState();
+    const [urlAtMount] = useState(getCurrentUrl);
+
+    return getOnboardingIntentFromUrl(urlAtMount) ?? getOnboardingIntentFromUrl(initialURL);
```
hungvu193, line 57: I don't know man, this looks confusing 😄 . Mind explaining this one? OR leaving a comment here?

src/DeepLinkHandler.tsx
MelvinBot: You're right that redirect-after-login already exists — but it doesn't cover this case, and separately, **this line is a no-op as written**, so it should either be dropped or the plumbing behind it fixed.

**Why the existing logic doesn't substitute.** `exitTo` is consumed by [`handleExitToNavigation`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/actions/Session/index.ts#L1529) and by `openReportFromDeepLink`'s `waitForUserSignIn()`. Both *navigate to a route*. For an already-onboarded recipient — this link's entire audience — `intent=submit` isn't a destination; the PR does side-effect work (create/reuse a Submit workspace) instead. `openReportFromDeepLink` also [returns early for authenticated users](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/actions/Link.ts#L606-L609).

**Why this line doesn't work anyway.** `onInitialUrl` is **not** the initial-URL context setter — it's `Expensify.tsx`'s local `useState`:

- [`src/Expensify.tsx:79`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/Expensify.tsx#L79) — `const [initialUrl, setInitialUrl] = useState(...)`
- [`src/Expensify.tsx:302`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/Expensify.tsx#L302) — `<DeepLinkHandler onInitialUrl={setInitialUrl} />`
- Its only consumer is `NavigationRoot`'s `initialUrl` prop, read inside a `[]`-dep `useMemo` ([`NavigationRoot.tsx:190`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/NavigationRoot.tsx#L190) — *"The initialState value is relevant only on the first render"*). On a warm link `NavigationRoot` is already mounted, so the value is discarded.

Meanwhile [`useOnboardingDeeplinkIntent`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/hooks/useOnboardingDeeplinkIntent.ts#L17-L20) reads `useInitialURLState().initialURL`, which is written **only** by the provider's own one-shot [`Linking.getInitialURL()`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/components/InitialURLContextProvider.tsx#L40-L47). Grepping `src/` for `setInitialURL` returns nothing outside `InitialURLContextProvider.tsx` itself. The two paths never meet — so the warm-native case this targets returns `undefined` with or without the change.

**Suggestion:** drop the `|| getOnboardingIntentFromUrl(state.url)` clause, or make it real by having `InitialURLContextProvider` subscribe to `Linking.addEventListener('url', …)` itself. The latter also closes the pre-existing gap in the `hasSecureLinkKey` branch this clause was copied from.

<details>
<summary>Per-scenario check — everything else already works without this line</summary>

| Scenario | Works without the change? | Why |
|---|---|---|
| Web, signed out, magic link `?exitTo=onboarding%3Fintent%3Dsubmit` | Yes | `getCurrentUrl()` still holds the `/v/` URL at mount (`handleExitToNavigation` awaits `waitForProtectedRoutes()`, which resolves after `AuthScreens` renders), and `getOnboardingIntentFromUrl` already unwraps `exitTo`. Also, react-native-web caches `window.location.href` at bundle load, so the context fallback yields the same. |
| Web, signed in, `/onboarding?intent=submit` | Yes | Both `getCurrentUrl()` and the cached context value contain the param. |
| Native cold launch | Yes | `Linking.getInitialURL()` → `InitialURLContextProvider`. `initialURL` isn't latched in the hook, so the late resolve still re-renders the handler. |
| Native/desktop **warm** launch | **No — and still no with the change** | `getCurrentUrl()` is `''` on native, the `useState` latch is stale (`AuthScreens` doesn't remount on a warm deeplink), and the new `onInitialUrl` call lands in the local state described above. |

Two side notes:

- The `'url'` listener block is native-only — react-native-web never dispatches a `'url'` event (it only emits `onOpen` from `openURL`), which matches the [existing comment at `DeepLinkHandler.tsx:137`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/DeepLinkHandler.tsx#L137).
- The `hasSecureLinkKey` branch has the same defect on `main`, but it's partly rescued because [`useOnboardingFlow`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/hooks/useOnboardingFlow.ts#L44) also checks `Navigation.getActiveRoute()`. That fallback isn't available here: `intent` isn't a declared route param on any onboarding screen, so `getActiveRoute()` wouldn't serialize it.
- No test covers the warm `'url'` path (`tests/unit/DeepLinkHandlerTest.tsx` passes `jest.fn()` for `onInitialUrl`), so CI wouldn't flag this either way.

</details>

---
_[view run](https://github.com/Expensify/melvin/actions/runs/33316941517)_

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+function SubmitIntentDeeplinkHandler() {
+    const onboardingDeeplinkIntent = useOnboardingDeeplinkIntent();
+
+    if (onboardingDeeplinkIntent !== CONST.ONBOARDING_INTENTS.SUBMIT) {
+        return null;
+    }
+
+    return <ApplySubmitOnboardingIntent />;
+}
```
hungvu193, line 20: @MelvinBot What's the current logic to handle all deeplink nowaday in App? Do we really need to create a specific one here for submit?

MelvinBot: **Short answer:** the *component* isn't bespoke — it's the fourth sibling of an established pattern. The *URL* is the problem. `onboarding` is a route that [`OnboardingGuard` deliberately redirects to HOME](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/guards/OnboardingGuard.ts#L220-L225) for exactly this PR's audience (users who finished guided setup), because [the onboarding navigator isn't even mounted for them](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/AppNavigator/AuthScreens.tsx#L342). So none of the generic routing machinery can carry this link, and the PR has to sniff the raw URL out-of-band instead. **A dedicated route would let the existing machinery do it and delete 3 of the 4 new files** — if the email URL is still changeable.

### Current deeplink handling, in three layers

1. **React Navigation `linking` — the canonical path.** [`linkingConfig`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/linkingConfig/index.ts#L11-L17) → prefixes stripped → [`getAdaptedStateFromPath`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/helpers/getAdaptedStateFromPath.ts#L431) → nav state. Same primitives back in-app `Navigation.navigate`.
2. **Navigation guards — the interception chokepoint.** [`RootStackRouter.getStateForAction` runs `handleNavigationGuards` first](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/AppNavigator/createRootStackNavigator/RootStackRouter.ts#L192-L198), before any routing. Three registered today: `OnboardingGuard`, `MigratedUserWelcomeModalGuard`, `SubmitPlanWelcomeModalGuard`.
3. **Narrow legacy side-channels.** [`openReportFromDeepLink`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/actions/Link.ts#L606-L609) — for signed-in users it says *"Navigation for signed users is handled by react-navigation"* and returns; it only serves anonymous/just-signed-in. Plus four separate `exitTo` consumers.

### Why the PR couldn't just reuse one

I checked the four obvious candidates:

| Candidate | Verdict |
|---|---|
| `setUpPoliciesAndNavigate`'s [`exitTo=workspace/new` branch](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/actions/App.ts#L872) | **No.** It's the same *shape* (deeplink creates a workspace), but it's web-only ([`getCurrentUrl()` is `''` on native](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/currentUrl/index.native.ts#L3)), requires a `/transition` URL, and calls a plain function — whereas [`useAutoCreateSubmitWorkspace` is a hook](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/hooks/useAutoCreateSubmitWorkspace.ts#L27) and can't be called from an action module. |
| A guard | **Decision only.** [`GuardResult` is `ALLOW｜BLOCK｜REDIRECT`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/guards/types.ts#L8) — no "do work". It'd still need a screen to redirect *to*. |
| Auto-selecting EMPLOYER in the onboarding flow | **No.** The navigator isn't mounted for completed users, and `OnboardingGuard` bounces them to HOME. |
| A dedicated route | **Yes, and it's the cleanest.** See below. |

### The one thing worth pushing on

`exitTo=<real-route>` already gets the signed-out magic-link case for free via [`handleExitToNavigation`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/actions/Session/index.ts#L1529-L1536) (`waitForUserSignIn` → `waitForProtectedRoutes` → `navigate`). A real route + screen (mirroring `SubmitPlanWelcomeModalNavigator`, whose `handleConfirm` already calls `autoCreateSubmitWorkspace(firstName, lastName, false)` — the exact call needed, just behind a button) would let you drop `getOnboardingIntentFromUrl.ts`, `useOnboardingDeeplinkIntent.ts`, the `DeepLinkHandler` hook, and the `hasAppliedIntent` module global. The screen mount *is* the one-shot.

The catch: [the CONST comment says these values are "embedded in emails we can't redeploy"](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/CONST/index.ts#L151). `abzokhattab` — is that URL already live in sent emails, or is it still changeable? That's the deciding input.

<details>
<summary>Two smaller notes</summary>

**The component pattern is fine.** `SubmitIntentDeeplinkHandler` sits [between `SearchRouterWarmup` and `ThreeDSAuthHandler`](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/AppNavigator/AuthScreens.tsx#L166-L171), alongside `AuthScreensInitHandler` and `UserStatusHandler` — null-rendering handlers mounted in `AuthScreens` are the convention, not an invention. Other side-effect deeplinks use it too (`DelegatorConnectGate` for `?delegatorEmail=`, `ReportFetchHandler` for `?secureKey=`).

**Even if the route stays, the modal race can be fixed structurally.** The [`hasPendingSubmitDeeplink` module global](https://github.com/Expensify/App/blob/4273e1d057ad8e7a0b1e16c8b816b46b8c0db90f/src/libs/Navigation/guards/SubmitPlanWelcomeModalGuard.ts#L45) exists only because two decisions live in different modules on different clocks (React mount effect vs. the guard's `HAS_LOADED_APP` microtask). Folding the intent check into `navigateToSubmitPlanWelcomeModalIfReady()` — same function, same subscription, same microtask — makes it a local condition and the race impossible rather than merely sequenced around.

**Also worth knowing:** guards only run on navigation *actions*, never on cold-start initial state, which is why both welcome-modal guards need a second Onyx-driven "proactive" path.

</details>

---
_[view run](https://github.com/Expensify/melvin/actions/runs/33317523331)_

hungvu193:
@abzokhattab Can you provide android test steps as well?

I was trying to test Android but it doesn't seem to work:


https://github.com/user-attachments/assets/f2ed91ff-9b62-4243-9218-5aaac6caa958

abzokhattab:
> I was trying to test Android but it doesn't seem to work:

make sure to try the standalone script or you can just remove the MobileExpensify folder before running npm run android

i also added the steps for android in the description with a video attached

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/index.tsx
abzokhattab: The URL is still ours to change — nothing has shipped, so no email can contain a working link yet. That `CONST` comment was my rationale for picking short stable aliases over the internal onboarding-choice strings, not a statement that the value is already live. Sorry for the misdirection.

So I'm taking the dedicated-route option. Beyond deleting the URL-sniffing files, it fixes the warm-native case structurally rather than by patching shared context: React Navigation's own linking handles `'url'` events, so the path that's currently a no-op stops existing instead of needing `InitialURLContextProvider` to grow a subscription. The screen mount becomes the one-shot, so the `hasAppliedIntent` global goes too.

Two things that don't come for free with the route, which I'm handling separately:

- The paid-workspace-member bug Codex flagged is in the hook call, not the routing, so it needs the `getGroupPoliciesWhereReportCanBeCreated` check regardless.
- The welcome-modal race survives, since the guard still decides on its `HAS_LOADED_APP` microtask. But with a real route the guard can key off the incoming navigation instead of a module flag set from a React effect, which is the structural version of your suggestion.

Will push shortly.

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/ApplySubmitOnboardingIntent.tsx
```diff
@@ -0,0 +1,65 @@
...
+        if (!isOnboardingCompleted || isSupportalSession) {
+            return;
+        }
+
+        // Persists the suppression applied above, so the modal stays away in later sessions too.
+        setSubmitMigrationModalShown();
+
+        // `false` skips CompleteGuidedSetup, which is already done. The hook navigates to the user's existing Submit
+        // workspace rather than creating a second one, which is what makes repeat clicks idempotent.
+        autoCreateSubmitWorkspace(firstName ?? '', lastName ?? '', false);
```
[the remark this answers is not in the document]
abzokhattab: Correct, fixed in b8a894a. `useAutoCreateSubmitWorkspace` gates on `canEditWorkspaceSettings`, which is admin-level, so a plain member of their employer's Team workspace still got a personal Submit workspace. The handler now checks `getGroupPoliciesWhereReportCanBeCreated` — the same check the welcome guard uses — and drops the intent if the user already belongs to one.

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/ApplySubmitOnboardingIntent.tsx
```diff
@@ -0,0 +1,65 @@
...
+    useEffect(() => {
+        if (hasAppliedIntent || !hasLoadedApp || isOnboardingCompleted === undefined) {
+            return;
+        }
+        hasAppliedIntent = true;
+
+        // Marked applied above before these checks, since they are reasons to drop the intent rather than to wait for
+        // it to become actionable. Returning without consuming would leave it live for whoever signs in next.
+        if (!isOnboardingCompleted || isSupportalSession) {
+            return;
```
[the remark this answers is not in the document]
abzokhattab: Right, fixed in b8a894a. `releaseWelcomeModalForSubmitDeeplink()` now runs on every drop path, so a suppression set at mount can't outlive the intent that requested it.

src/DeepLinkHandler.tsx
abzokhattab: You were both right — the line was a no-op. Reverted the file to `main` in e83da96.

src/libs/getOnboardingIntentFromUrl.ts
abzokhattab: Agreed, inlined.

src/hooks/useOnboardingDeeplinkIntent.ts
abzokhattab: Fair — comment added. Two sources because neither covers both platforms: on web the link is in the address bar, but only until the app navigates and rewrites it, hence latching at mount; on native there's no address bar, so it arrives through the initial-URL context instead.

src/libs/Navigation/AppNavigator/SubmitIntentDeeplinkHandler/index.tsx
abzokhattab: Good question — the closest existing machinery is `setUpPoliciesAndNavigate`, which does create a workspace off a URL param (`exitTo=workspace/new`). It doesn't cover this case though: it returns early unless the URL contains `exitTo`, its creation branch requires a `/transition` OldDot handoff, it builds a generic workspace rather than a Submit plan landing on Spend + #admins, and it runs at AuthScreens mount without waiting for `HAS_LOADED_APP` — we have to wait for that, or the existing-workspace checks read a half-loaded store and create a duplicate.

The bigger blocker is that Submit creation is a hook (`useAutoCreateSubmitWorkspace` → `useOnboardingWorkspaceCreationState`, which aggregates ~15 values plus four `useOnyx` subscriptions) while `setUpPoliciesAndNavigate` is an imperative action, so reusing it there would mean duplicating that state gathering and forking the Submit creation path in two.

The split into two files is deliberate for the same reason: `index.tsx` returns `null` unless the intent is present, so that heavy hook only mounts on sessions actually launched from this link rather than on every authenticated session.

what do you think @MelvinBot

MelvinBot: All four `setUpPoliciesAndNavigate` points check out, and the split is *more* defensible than you argued. I also need to walk back part of what I said about the route alternative. But the deciding question is still unanswered.

**Your `setUpPoliciesAndNavigate` case is sound — and understated.** The strongest point isn't in your list: [`getCurrentUrl()` returns `''` on native](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/currentUrl/index.native.ts#L3), so the function [returns at `App.ts:856`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/actions/App.ts#L854-L857) before it ever reaches the `exitTo` check. It's not that it doesn't cover this case — it's web-only. And `/transition` is a hard gate, not incidental: [three required conjuncts](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/actions/App.ts#L870-L872).

**The split is mandatory, not stylistic.** Rules of Hooks forbid conditional hook calls, so gating that hook tree *requires* a second component — the only real question is one file or two. And the cost avoided is bigger than "heavy": 23 transitive `useOnyx` subscriptions and 5 full `POLICY`-collection scans, on every authenticated session. Precedent is the line directly above yours — [`SearchRouterWarmup` at `AuthScreens.tsx:167`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/AppNavigator/AuthScreens.tsx#L166-L171) is the same cheap-gate/heavy-child shape. Only nit: the more common house style is two components in *one* file — e.g. [`DelegatorConnectGate.tsx:21-23`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/AppNavigator/DelegatorConnectGate.tsx#L21-L23) in the same folder. Take it or leave it.

**Correcting myself on the dedicated route.** I oversold it. The deletion list is two files plus `hasAppliedIntent` — **not** the module globals generally. [`hasPendingSubmitDeeplink`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/guards/SubmitPlanWelcomeModalGuard.ts#L45-L55) survives either way, because the guard's proactive redirect still fires on a `HAS_LOADED_APP` microtask that beats any `exitTo` navigation gated on `waitForUserSignIn()`. And a route costs ~10-12 registration touch points (SCREENS, NAVIGATORS, ROUTES, linkingConfig, three spots in `types.ts`, navigator, screen, AuthScreens, `EXCLUDE_FROM_LAST_VISITED_PATH`). Against your current 4 new source files, that's roughly a wash. I'm not pushing it.

**What I'd still put to you.** Two of your last five commits are the argument I was reaching for: reverting warm-link support, and [fixing `isOnboardingPath` to match on a boundary](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/getOnboardingIntentFromUrl.ts#L18-L21) after `onboarding/work-email?intent=submit` matched. Both are costs of hand-parsing a URL the router already parses. That's not fatal, but it's the recurring bill.

So the deciding input is still the one from my last comment, and it's unanswered: **is the emailed URL locked yet?** If it isn't, there's a cheaper option nobody has costed — [`submit-plan-welcome` is already a registered, linkable route](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/ROUTES.ts#L192-L195), and [its confirm button already calls `autoCreateSubmitWorkspace(firstName, lastName, false)`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/components/SubmitPlanWelcomeModal.tsx#L48-L50). An `exitTo=<entryScreen>/submit-plan-welcome` link might ship this with zero new code, trading one-click for one tap. **Unverified** — I didn't confirm `handleExitToNavigation`'s `goBack` + `navigate` sequence resolves that dynamic-suffix path. Worth 20 minutes before committing to 276 lines.

<details>
<summary>Claim-by-claim verification, plus one thing to check before merge</summary>

| Your claim | Verdict | Anchor |
|---|---|---|
| Returns early unless URL contains `exitTo` | True (also requires `session`; and it's a raw `includes()` on the href, not a param lookup) | [`App.ts:854-857`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/actions/App.ts#L854-L857) |
| Creation branch requires a `/transition` handoff | True — hard requirement | [`App.ts:870-872`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/actions/App.ts#L870-L872) |
| Builds a generic workspace, not Submit + Spend/#admins | True — it passes no `type`, which is typed `TEAM｜CORPORATE`, so it structurally can't produce Submit; defaults to TEAM and lands on workspace settings | [`App.ts:592`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/actions/App.ts#L592), [`Policy.ts:2571`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/actions/Policy/Policy.ts#L2571) vs [`useAutoCreateSubmitWorkspace.ts:85`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/hooks/useAutoCreateSubmitWorkspace.ts#L85) |
| Runs at mount without waiting for `HAS_LOADED_APP` | True — [`AuthScreensInitHandler.tsx:192`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/AppNavigator/AuthScreensInitHandler.tsx#L192-L203) sits in a `[]`-dep effect and the file never reads `HAS_LOADED_APP`/`IS_LOADING_APP` | same |
| "~15 values plus four `useOnyx`" | Four is exact; the aggregate is 17 returned / 16 destructured, and 18 transitive subscriptions. Understated, not overstated | [`useAutoCreateSubmitWorkspace.ts:51-60`](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/hooks/useAutoCreateSubmitWorkspace.ts#L51-L60) |

**Worth checking before merge:** [`isNavigatingToOnboardingFlow` only matches `NAVIGATE`/`PUSH` actions](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/guards/OnboardingGuard.ts#L178-L187). A cold boot straight to `/onboarding?intent=submit` builds initial state from the linking config with no navigation action, so `OnboardingGuard` never fires — and [the onboarding navigator isn't mounted for completed users](https://github.com/Expensify/App/blob/520bc994475d4fa543c8d83f25bcc03e5890276b/src/libs/Navigation/AppNavigator/AuthScreens.tsx#L342-L346) either. Your videos suggest it's fine in practice, but it'd be good to know what's on screen during the window before `navigateToSubmitWorkspaceAfterOnboardingWithMicrotaskQueue` runs, especially on a slow connection. I tried to check this in a browser myself and couldn't — my test session lost auth before I could navigate, so I learned nothing about the PR either way.

</details>

---
_[view run](https://github.com/Expensify/melvin/actions/runs/33339880508)_

src/libs/getOnboardingIntentFromUrl.ts
```diff
@@ -0,0 +1,41 @@
...
+    // getRouteFromLink strips whichever linking-config prefix matched, so web URLs, the desktop `app://-/` origin and
+    // the native scheme all reduce to the same route. It leaves the leading slash on in-app paths.
+    const pathWithQuery = getRouteFromLink(url).replace(/^\/+/, '');
+    const onboardingPathWithQuery = isOnboardingPath(pathWithQuery) ? pathWithQuery : getSearchParamFromPath(pathWithQuery, 'exitTo');
+
+    if (!isOnboardingPath(onboardingPathWithQuery)) {
+        return undefined;
+    }
+
+    const intent = getSearchParamFromPath(onboardingPathWithQuery, 'intent');
```
hungvu193: ~@abzokhattab We have linting here~

Nevermind, it was outdated

hungvu193:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



https://github.com/user-attachments/assets/6c201c2b-0d33-4585-b7fb-f70e1f97bb62

</details>

<details>



</details>

<details>



https://github.com/user-attachments/assets/21cf28d3-5415-4bba-bf61-012f228f6928

</details>

hungvu193 (approved):

src/libs/getOnboardingIntentFromUrl.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
iwiznia: Do we need this as a separate file/lib? Can't we put this in `src/hooks/useOnboardingDeeplinkIntent.ts`?

flaviadefaria:
Assigning Garrett since this is his project.

src/libs/getOnboardingIntentFromUrl.ts
abzokhattab: Agreed ... moved it to the hook to remove duplication

iwiznia (dismissed):
Tests failed

abzokhattab:
ts is failing on main

iwiznia (approved):

iwiznia:
Yeah, started a thread about it here https://expensify.slack.com/archives/C01GTK53T8Q/p1788207872549709

iwiznia:
Not an emergency, failures are unrelated to this PR and we are discussing them in slack

kavimuru:
@abzokhattab help with the following to test in staging

Request a magic link for an account that has completed onboarding and has no Submit workspace, then open it:
Web: https://dev.new.expensify.com:8082/v/<accountID>/<validateCode>?exitTo=onboarding%3Fintent%3Dsubmit
iOS: run xcrun simctl openurl booted "new-expensify://v/<accountID>/<validateCode>?exitTo=onboarding%3Fintent%3Dsubmit" in terminal
Android: run adb shell am force-stop com.expensify.chat.dev then adb shell am start -W -a android.intent.action.VIEW -d "new-expensify://v/<accountID>/<validateCode>?exitTo=onboarding%3Fintent%3Dsubmit" com.expensify.chat.dev in terminal

francoisl:
@abzokhattab for QA testing, can we not just use a `https://staging.expensify.com/v/XXX/YYY? exitTo=onboarding%3Fintent%3Dsubmit` link and click it from the a mobile device? Might be easier than having to run an `adb`/`xcrun` command for the QA team.

Anyway, I don't think those links are used yet, going to check this off the deploy checklist so it doesn't block other stuff.

Issue #98169, opened 2026-08-10
NumberForm foundation 1: root API and NumberForm.Input


mountiny:
Sub-issue of [#97970](https://github.com/Expensify/App/issues/97970).
Migration plan: [jmgraa comment](https://github.com/Expensify/App/issues/97970#issuecomment-5240813264).

This is the foundation of the whole decomposition. It defines the root `NumberForm` contract and ships `NumberForm.Input` in the same PR, so the public types do not churn in a follow up. Nothing else in the decomposition can start until this lands.

`NumberWithSymbolForm` is untouched by this issue and keeps working.

## Scope

- **Step:** 1 of the migration plan
- **Scope:** new `src/components/NumberForm/**`, plus 2 lines moved in the lint baseline

## Root API

```tsx
<NumberForm value={amount} onInputChange={handleAmount}>
    <NumberForm.Input symbol="$" position="prefix" />
</NumberForm>
```

Root props: `value`, `onInputChange`, `negativeMode`, `children`.

The root must also satisfy the `InputWrapper` contract, because `src/pages/workspace/taxes/ValuePage.tsx:95` registers the component as a `Form` `InputComponent` and `src/components/Form/types.ts:61` lists it in the `ValidInputs` union. That means accepting `ref`, `onBlur`, `errorText` and `value` as `Form` injects them. Design this in now rather than discovering it during the migration batch.

## Files

Mirror the folder shape of `src/components/ButtonComposed/`, which is the accepted precedent in this codebase:

- `src/components/NumberForm/index.tsx` — root, attaches the primitives
- `src/components/NumberForm/types.ts`
- `src/components/NumberForm/context/` — shared state for the primitives
- `src/components/NumberForm/primitives/NumberFormInput.tsx`
- `config/eslint/eslint.seatbelt.tsv:275-276` — see below

## What moves into NumberForm.Input

Everything the numeric input owns today, lifted out of `src/components/NumberWithSymbolForm.tsx`: parsing, formatting, validation, selection, keyboard handling, locale and ref behaviour.

Reuse the existing primitives instead of reimplementing them:

- `src/components/TextInputWithSymbol/` — symbol display and platform variants
- `src/components/AmountTextInput.tsx`

Two things to reconcile while moving:

[...]

Issue #98171, opened 2026-08-10
NumberForm foundation 3: BigNumberPad primitive and layout composition


mountiny:
Sub-issue of [#97970](https://github.com/Expensify/App/issues/97970).
Migration plan: [jmgraa comment](https://github.com/Expensify/App/issues/97970#issuecomment-5240813264).

Last foundation issue. It adds the number pad primitive, composes every primitive through `NumberForm` across all four layout combinations, and rewires `NumberWithSymbolForm` to delegate to that composition.

## Scope

- **Step:** 3 of the migration plan
- **Scope:** 1 new primitive, layout composition, `NumberWithSymbolForm` becomes an adapter

## Target API

```tsx
<NumberForm value={amount} onInputChange={handleAmount}>
    <NumberForm.Input symbol="$" position="prefix" />
    <NumberForm.FlipButton />
    <NumberForm.CurrencyButton currency="USD" onPress={handleCurrency} />
    <NumberForm.BigNumberPad onNumberPressed={handleNumber} />
</NumberForm>
```

`NumberForm.BigNumberPad` wraps the existing `src/components/BigNumberPad.tsx` rather than reimplementing it.

## Layouts to cover

Four combinations, currently spread across three early return paths in `src/components/NumberWithSymbolForm.tsx`:

- portrait, touch
- portrait, non touch
- landscape, touch
- landscape, non touch

Duplication being collapsed:

- `BigNumberPad` is rendered twice, at lines 686 (landscape) and 785 (portrait), with identical props. The landscape branch also has a redundant double `shouldShowBigNumberPad ?` check at lines 680 and 685.
- The ref fanout closure is duplicated verbatim at lines 551-559 and 587-595.
- Lines 581 to 635 build a shared `textInputComponent` with roughly 35 props, used by the landscape and portrait paths but not by `displayAsTextInput`.

## Correction to step 3 of the migration plan

The plan says step 3 should "Remove the remaining duplicated layout/button paths from the old component". Do not do that literally. `NumberWithSymbolForm` stays in use until the cleanup issue [#98178](https://github.com/Expensify/App/issues/98178), and roughly 30 files depend on it transitively through `AmountForm`, `MoneyRequestAmountInput` and `MoneyRequestAmountForm`. Gutting its layout paths here breaks all of them.

[...]

Issue #98170, opened 2026-08-10
NumberForm foundation 2: FlipButton and CurrencyButton primitives


mountiny:
Sub-issue of [#97970](https://github.com/Expensify/App/issues/97970).
Migration plan: [jmgraa comment](https://github.com/Expensify/App/issues/97970#issuecomment-5240813264).

The currency button is implemented four times inside `NumberWithSymbolForm` and the flip button three times. This issue collapses each into one primitive with a minimal prop set and its own tests.

## Scope

- **Step:** 2 of the migration plan
- **Scope:** 2 new primitives under `src/components/NumberForm/primitives/`

## Target API

```tsx
<NumberForm value={amount} onInputChange={handleAmount}>
    <NumberForm.Input symbol="$" position="prefix" />
    <NumberForm.FlipButton />
    <NumberForm.CurrencyButton currency="USD" onPress={handleCurrency} />
</NumberForm>
```

## Duplication being collapsed

Line numbers are against `src/components/NumberWithSymbolForm.tsx` on `main`.

**Currency button, 4 copies:**

| Line | Path | Gate |
|---|---|---|
| 515 | `displayAsTextInput` right hand side | `shouldShowCurrencyButton && currencyOrUnitButtonText`, calls `onTrailingDropdownPress` |
| 647 | landscape | `isSymbolPressable` |
| 727 | portrait | `isSymbolPressable && !!currency && !canUseTouchScreen` |
| 753 | portrait | `isSymbolPressable && canUseTouchScreen` |

The last three are near identical — same `icons.CoinsButton`, same `styles.minWidth18`, same `onSymbolButtonPress`, same `text={currency}` — and differ only in touch gating.

**Flip button, 3 copies:**

| Line | Path | Gate | Negate mechanism |
|---|---|---|---|
| 502 | `displayAsTextInput` | `shouldShowFlipButton && allowNegativeInput && canUseTouchScreen` | internal `handleFlipPress` (line 470) |
| 659 | landscape | `allowFlippingAmount` | caller supplied `toggleNegative` |
| 765 | portrait | `allowFlippingAmount && canUseTouchScreen` | caller supplied `toggleNegative` |

## Three things to fix while collapsing

1. **Pick one negate mechanism.** Two coexist today: the internal `handleFlipPress` and the caller supplied `toggleNegative`. `NumberForm.FlipButton` should own negation through the root's `negativeMode`, and callers should not have to hand in a toggle.
[...]

Issue #16974, opened 2023-04-05
[HOLD for payment 2023-04-27] [$1000] Mweb - Send money - Can't paste number into sum field when at least 1 number is present


kbecciv:
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## Action Performed:
Precondition - User is logged in and has a chat available in account
1. Open http://staging.new.expensify.com/ on iOS Safari or Android Chrome
2. Tap on FAB "+" icon
3. Tap "Send money"
4. Type a number into the "amount" field (e.g. 50)
5. Copy the number
6. Enter one number into the "amount" field (e.g. 2)
7. Long press on the space around the number (do not press on the number) 
8. Use "Paste" keyboard shortcut (CMD V)

## Expected Result:

It is possible to paste a number into summary field when there is a number present in the field already on iOS mobile, Android mobile, Mac Chrome, Mac Desktop

## Actual Result:

Nothing happens when trying to paste a number that doesn't exceed field limits into summary field when there is at least a number present in the field.

If there are no numbers present in the field (a placeholder 0 is displayed) then the copied numbers are pasted successfully.

## Workaround:
Unknown

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [ ] Android / native
- [X] Android / Chrome
- [ ] iOS / native
- [x] iOS / Safari
- [ ] MacOS / Chrome / Safari
- [ ] MacOS / Desktop

**Version Number:** iOS v1.3.0.0 / Android v1.3.0.0

**Reproducible in staging?:** Yes

**Reproducible in production?:** Yes

**If this was caught during regression testing, add the test name, ID and link from TestRail:**

**Email or phone of affected tester (no customers):**

**Logs:** https://stackoverflow.com/c/expensify/questions/4856

**Notes/Photos/Videos:** Any additional supporting documentation

https://user-images.githubusercontent.com/93399543/230189004-7b436dc9-d28f-4344-b41a-6bdcaf8b83e0.mp4

**Expensify/Expensify Issue URL:**

**Issue reported by:** Applause - Internal Team

**Slack conversation:**

[View all open jobs on GitHub](https://github.com/Expensify/App/issues?q=is%3Aopen+is%3Aissue+label%3A%22Help+Wanted%22)

<details><summary>Upwork Automation - Do Not Edit</summary>
[...]

PR #98967, opened 2026-08-19
[No QA] Add NumberForm root with SymbolInput and TextInput primitives

jmgraa:
### Explanation of Change

This PR introduces the composable `NumberForm` and `NumberComposer` foundations and extracts shared numeric editing behavior into `NumberInput`.

`NumberInput` centralizes the logic previously duplicated between `setNewNumber` and `setFormattedNumber` in `NumberWithSymbolForm`.

The legacy component had two setters because its input paths used different update flows:

- `setNewNumber` handled values from `TextInputWithSymbol` and the number pad.
- `setFormattedNumber` handled values from the standard `TextInput`.

Both functions performed similar operations, including removing spaces, normalizing separators, adding leading zeros, validating the value, updating the selection, and notifying `onInputChange`. However, the operations were duplicated and applied through separate state-update paths.

The new `NumberInput.setNumber` uses one canonical pipeline for both input types:

1. Remove spaces added by Safari and convert locale-specific digits.
2. Normalize decimal and comma separators.
3. Add a leading zero where needed.
4. Validate decimal precision, maximum length, and negative-value rules.
5. Convert the edited display value back to the canonical value.
6. Commit the value and update the caret through one shared path.
7. Notify the parent with the canonical value.

`NumberComposer` uses display-to-canonical mapping for cases where the visual representation differs from the stored value. For example, it stores `-12` canonically while displaying `12` and rendering the minus sign separately. This allows the symbol input and future composed primitives to share the same editing logic without duplicating setters.

`NumberForm` provides form-level state and callbacks through context and exposes a `TextInput` primitive. `NumberComposer` provides a composable `SymbolInput` primitive for future controls such as currency selectors, flip buttons, and number pads.

Error rendering is also composable:

- `NumberForm.TextInput` passes `errorText` to `TextInput`, which owns the label, input border, and inline error message.
- `NumberComposer` exposes a separate `NumberComposer.Error` primitive. It renders `FormHelpMessage` outside the input so the surrounding composition can decide where the message belongs.
- This preserves the legacy symbol-input layout, where the error is rendered below the input instead of being passed to `TextInputWithSymbol`. Passing it through would change the input border, error position, and overall layout.

`AmountForm` now uses `NumberForm.TextInput` for the standard text-input path. The legacy `NumberWithSymbolForm` remains in use for currency-button variants and other flows that still depend on its additional controls and layout behavior.

Normalization is now performed by the shared `NumberInput` controller. `TextInputWithSymbol` can opt out of its own normalization so values are not normalized twice, while the legacy component keeps its existing behavior.

Dedicated tests cover shared numeric editing, formatting, validation, external value updates, imperative refs, negative values, caret handling, selection guards, context state, and error propagation.

### Fixed Issues
$ https://github.com/Expensify/App/issues/98169
PROPOSAL: https://github.com/Expensify/App/issues/97970

### Tests

### Offline tests

N/A

### QA Steps

N/A

### PR Author Checklist

    - [ ] Android: Native
    - [ ] Android: mWeb Chrome
    - [ ] iOS: Native
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari


src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+        ...rest
+    } = props;
+    const {errorText, formattedNumber, handleBlur, handleInputRef, handleKeyPress, handleSelectionChange, inputPosition, negativeMode, numberFormat, selectionForRender, setNumber} =
+        useNumberFormInputLogic({...props, maxLength});
+
+    // TODO: Unify both input paths (NumberForm.SymbolInput and NumberForm.TextInput) around a shared NumberForm.Error primitive so error rendering is no longer conditional.
+    return (
+        <>
+            <TextInputWithSymbol
+                {...rest}
```
jmusial: This leaks some props that do not belong in `TextInputWithSymbol` (fe. `isNegative`)

src/components/NumberForm/context/NumberFormContext.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import type {NumberFormActionsContextValue, NumberFormContext, NumberFormStateContextValue} from './types';
+
+const createNumberFormContext = createContextNamespace('NumberForm');
+
+const [NumberFormStateContext, useNumberFormState] = createNumberFormContext<NumberFormStateContextValue>('State');
+const [NumberFormActionsContext, useNumberFormActions] = createNumberFormContext<NumberFormActionsContextValue>('Actions');
+
+function useNumberFormContext(): NumberFormContext {
+    return {...useNumberFormState('useNumberFormContext'), ...useNumberFormActions('useNumberFormContext')};
```
jmusial: This will cause every subscriber to rerender for every action / state change anyways. 

I'd consider if split context brings any benefit here ?

src/components/TextInputWithSymbol/BaseTextInputWithSymbol.tsx
```diff
@@ -20,6 +20,7 @@ function BaseTextInputWithSymbol({
     symbolPosition = CONST.TEXT_INPUT_SYMBOL_POSITION.PREFIX,
     onSymbolButtonPress = () => {},
     onChangeAmount = () => {},
+    shouldNormalizeAmountOnChange = true,
```
jmusial: if we have one consumer maybe we can flip the default ?

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
jmusial: Also some of the props are passed again after spread

src/components/NumberForm/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+import NumberFormTextInput from './primitives/NumberFormTextInput';
+
+const NumberForm = Object.assign(NumberFormComponent, {
+    SymbolInput: NumberFormSymbolInput,
+    TextInput: NumberFormTextInput,
+});
+
+export default NumberForm;
+export {useNumberFormContext} from './context';
+export type {NumberFormProps, NumberFormRef, NumberFormSymbolInputProps, NumberFormTextInputProps} from './types';
```
jmusial: Do we need those type exports from `index` ?

src/components/NumberForm/types.ts
```diff
@@ -0,0 +1,132 @@
...
+        /** Style applied to the symbol. */
+        symbolTextStyle?: StyleProp<TextStyle>;
+
+        /** Style applied to the negative symbol. */
+        negativeSymbolStyle?: StyleProp<TextStyle>;
+    };
+
+type NumberFormTextInputProps = NumberFormInputBaseProps & Pick<BaseTextInputProps, 'inputMode' | 'label' | 'onSubmitEditing' | 'prefixStyle' | 'suffixStyle'>;
+
+type NumberFormProps = {
```
jmusial: I think we can keep `type` with component if it's not reused across files ? (This case just component and it's test)

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
```diff
@@ -0,0 +1,237 @@
...
+        handleSelectionChange,
+        inputPosition,
+        negativeMode,
+        numberFormat,
+        onSubmitEditing,
+        selectionForRender,
+        setNumber,
+        shouldAllowNegativeInput,
+        shouldFlipNegative,
+    };
```
jmusial: Those 2 have no consumers

src/components/NumberForm/types.ts
```diff
@@ -0,0 +1,132 @@
...
+        | 'onBlur'
+        | 'onFocus'
+        | 'prefixContainerStyle'
+        | 'shouldApplyPaddingToContainer'
+        | 'shouldUseDefaultLineHeightForPrefix'
+        | 'submitBehavior'
+        | 'testID'
+        | 'touchableInputWrapperStyle'
+    >,
+    'onKeyPress'
```
jmusial: does this line do anything ? Do we have `onKeyPress` `in this type ?

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
```diff
@@ -0,0 +1,237 @@
...
+    start: number;
+    end: number;
+};
+
+const getNewSelection = (oldSelection: NumberSelection, previousLength: number, newLength: number): NumberSelection => {
+    const cursorPosition = oldSelection.end + (newLength - previousLength);
+    return {start: cursorPosition, end: cursorPosition};
+};
+
+function setRef<T>(ref: ForwardedRef<T> | undefined, value: T | null) {
```
jmusial: Doesn't `mergeRefs` do the same ?

src/components/NumberForm/index.tsx
```diff
@@ -0,0 +1,12 @@
+import NumberFormComponent from './NumberForm';
+import NumberFormSymbolInput from './primitives/NumberFormSymbolInput';
+import NumberFormTextInput from './primitives/NumberFormTextInput';
+
+const NumberForm = Object.assign(NumberFormComponent, {
```
jmusial: Sample usage jsdoc would be useful

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+                accessibilityLabel={rest.accessibilityLabel}
+                autoFocus={rest.autoFocus}
+                autoGrow={autoGrow}
+                autoGrowExtraSpace={rest.autoGrowExtraSpace}
+                autoGrowMarginSide={rest.autoGrowMarginSide}
+                containerStyle={containerStyle}
+                disableKeyboard={disableKeyboard ?? true}
```
jmusial: I think this default does nothing

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+        keyboardType,
+        isSymbolPressable = false,
+        onSymbolButtonPress,
+        maxLength,
+        ...rest
+    } = props;
+    const {errorText, formattedNumber, handleBlur, handleInputRef, handleKeyPress, handleSelectionChange, inputPosition, negativeMode, numberFormat, selectionForRender, setNumber} =
+        useNumberFormInputLogic({...props, maxLength});
+
+    // TODO: Unify both input paths (NumberForm.SymbolInput and NumberForm.TextInput) around a shared NumberForm.Error primitive so error rendering is no longer conditional.
```
jmusial: Can the TODO be adressed before proceeding ?

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
```diff
@@ -0,0 +1,237 @@
...
+        return;
+    }
+
+    if (ref) {
+        // eslint-disable-next-line no-param-reassign
+        ref.current = value;
+    }
+}
+
+function useNumberFormInputLogic({
```
jmusial: Seems odd that you added tests for components but not for the hook holding logic

jmusial (commented):
The code feels bit artificially split @ times (fe. context). 

I think it would be cleaner if you avoid `{...props}` spreading.

Also would suggest to migrate at least 1 call site with it

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
jmgraa: I have removed those props completely in the latest commit

src/components/NumberForm/context/NumberFormContext.ts
jmgraa: I kept the split because the input value changes on every keystroke, while the action callbacks do not. However, I removed the combined useNumberFormContext hook, so consumers now subscribe to state or actions separately

src/components/NumberForm/index.tsx
jmgraa: I’d keep these exports since they’re part of NumberForm’s public API - this matches the pattern used by `ButtonComposed`, which re-exports `ButtonProps` from its index

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
jmgraa: Most of the hook logic is covered through the `NumberForm.SymbolInput` and `NumberForm.TextInput` tests. This also verifies that both primitives are wired to the hook correctly. I don’t think we need a separate test suite that duplicates all component tests. I added an additional forward-delete test for a hook branch that was not covered before. We can add more focused hook-level tests later if we find logic that cannot be tested naturally through the public primitives

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
jmgraa: I would rather leave this TODO for a follow-up PR. Extracting the error rendering into a shared `NumberForm` primitive would require changes to the existing `TextInput`, since `NumberForm.TextInput` currently delegates error rendering, while `NumberForm.SymbolInput` renders `FormHelpMessage` separately (based on old component behaviour)

src/components/NumberForm/context/types.ts
```diff
@@ -0,0 +1,31 @@
+import type {NumberFormRef} from '@components/NumberForm/types';
+import type {BaseTextInputProps, BaseTextInputRef} from '@components/TextInput/BaseTextInput/types';
+
+import type {ForwardedRef} from 'react';
+
+type SetValueOptions = {
+    /** Whether `onInputChange` should be called with the new value. Defaults to `true`. */
+    notify?: boolean;
+};
```
dariusz-biela: This is a strange prop. Can you rename it or give an example in the comments of what it's used for?

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,221 @@
...
+            clearNegative();
+        }
+
+        inputHandleKeyPress(event);
+    };
+
+    // TODO: Unify both input paths (NumberForm.SymbolInput and NumberForm.TextInput) around a shared NumberForm.Error primitive so error rendering is no longer conditional.
+    return (
+        <>
+            <TextInputWithSymbol
```
Guccio163: No  setMouseDown/setMouseUp props

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    return value;
+};
+
+type NumberFormSymbolInputContentProps = NumberFormSymbolInputProps & {
+    isNegative: boolean;
+    clearNegative: () => void;
+};
+
+function NumberFormSymbolInputContent({
```
Guccio163: No shouldUseDynamicFontSize prop, breaking f.ex. MoneyRequestAmountForm.tsx:302

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,221 @@
...
+    onFocus,
+    onPress,
+    prefixContainerStyle,
+    shouldAllowFocusInLandscapeMode,
+    shouldApplyPaddingToContainer,
+    shouldUseDefaultLineHeightForPrefix,
+    submitBehavior,
+    testID,
+    touchableInputWrapperStyle,
+    isSymbolPressable = false,
```
Guccio163: Wasn't this one previously true by default?

src/components/NumberForm/types.ts
```diff
@@ -0,0 +1,91 @@
...
+        containerStyle?: StyleProp<ViewStyle>;
+
+        /** Style applied to the symbol. */
+        symbolTextStyle?: StyleProp<TextStyle>;
+
+        /** Style applied to the negative symbol. */
+        negativeSymbolStyle?: StyleProp<TextStyle>;
+    };
+
+type NumberFormTextInputProps = NumberFormInputBaseProps & Pick<BaseTextInputProps, 'inputMode' | 'label' | 'onSubmitEditing' | 'prefixStyle' | 'suffixStyle'>;
```
Guccio163: Why is onSubmitEditing included, when NumberFormSymbolInput.tsx doesn't use it?

src/components/NumberForm/primitives/NumberFormSymbolInput.tsx
```diff
@@ -0,0 +1,221 @@
...
+    // NumberForm.SymbolInput owns the separate-minus presentation. The root value remains the canonical signed value,
+    // while the shared input logic receives only the magnitude it renders.
+    useLayoutEffect(() => {
+        canonicalValueRef.current = value;
+    }, [value]);
+
+    const isNegative = value.startsWith('-');
+    const displayValue = getMagnitude(value);
+
+    const setValue = (nextValue: string, options?: SetValueOptions) => {
```
Guccio163: Might be a good idea to include a comment as to why does setValue use `getSignedValue(nextValue, previousValue.startsWith('-'))`, when  updateNumber only `canonicalValueRef.current = newNumber`.

Guccio163 (approved):
Generally LGTM,I left some comment and one more thing though: Since some props/code fragments  are not present anymore, it would be a good idea to create a list of all the intentional discrepancies and why is there the change, f.ex. prop isn't used no more, because special components covers the case or its behaviour will be covered by injecting inline-styles at callsite - Overall it would be a lot clearer for reviewers as to what is intentional and easier in the migration phase to remember about everything 👍

src/components/NumberForm/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    /** Renders a numeric input with a symbol beside it. */
+    SymbolInput: NumberFormSymbolInput,
+
+    /** Renders a numeric input using the standard text input component. */
+    TextInput: NumberFormTextInput,
+});
```
dariusz-biela: I still think we should break this down into two separate components. Right now, the problem is that there’s practically no logic in the composition root, and it’s the subcomponents that determine the component’s main behavior, which clearly shows that they should be independent.

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
```diff
@@ -0,0 +1,187 @@
...
+    start: number;
+    end: number;
+};
+
+const getNewSelection = (oldSelection: NumberSelection, previousLength: number, newLength: number): NumberSelection => {
+    const cursorPosition = oldSelection.end + (newLength - previousLength);
+    return {start: cursorPosition, end: cursorPosition};
+};
+
+function useNumberFormInputLogic({decimals = 0, maxLength, position = 'prefix', ref, onBlur: inputOnBlur, onKeyPress}: NumberFormInputBaseProps) {
```
dariusz-biela: If possible, it would be great to break down this hook, or at least isolate the part of it responsible for selection.

src/components/NumberForm/index.tsx
dariusz-biela, line 1: While I was going through the PR, I had Claude run a review.

He found a few interesting things. Some of them should resolve themselves once this code is split into two independent components. I’m sharing what he wrote to me more as a way to check for potentially unusual areas or to use it as context for your own agent. 

After you do this refactor, I’ll be digging deeper into the details anyway, so you don’t need to attach too much importance to any of this. 

<details>
<summary>AI Review</summary>

### 1 - `NumberFormSymbolInput.tsx:15`

> `getSignedValue` re-applies the previous sign unconditionally, so clearing a negative field commits `"-"` as the canonical value. I reproduced it on this branch: with `value="-12.50"` and `allowNegative`, `changeText(input, '')` fires `onInputChange('-')`. The field still looks empty because `getMagnitude('-')` is `''`, but `parseFloat('-')` is `NaN`. `clearNegative` only covers backspace-on-empty, not select-all-delete or pasting an empty value. Would a guard for the empty magnitude be enough here?

### 2 - `NumberFormSymbolInput.tsx:15`

> `getSignedValue` never looks at `allowNegative`, so a value that starts out negative keeps its sign even when negatives are disallowed. Reproduced with `value="-12"` and no `allowNegative`: typing `13` emits `"-13"`. `validateAmount` can't catch it either, since the hook only ever sees the magnitude. Is `allowNegative` meant to apply to `SymbolInput` at all?

### 3 - `AmountForm.tsx:117`

> This switch is the only production change in the PR and I don't see a test covering it - `grep -rln AmountForm tests` comes back empty. The prop mapping just below is complete today, but dropping one line would stay green on CI and break the 14 pages on this path. Separately, the legacy suite has `strips decimals when the decimals prop changes to a lower precision` (`NumberWithSymbolFormTest.tsx:175`), and the new suites never change `decimals` on a mounted `NumberForm` even though that effect was rewritten. Could we port those two over?

### 4 - `NumberForm.tsx:49`

> This reverses a contract the legacy component has an explicit test for: `does not update the internal number when the value prop changes to another non-empty value` (`NumberWithSymbolFormTest.tsx:161`). The new root resyncs on every prop change. It looks safe for the pages being switched, since `FormProvider` echoes the string back verbatim, so mostly I think it deserves a line in the PR description and a test. It also touches the next migration step: `MoneyRequestAmountInput` feeds a derived value and drives the field through `updateNumber`, which writes with `{notify: false}`, so the prop stays stale and the next prop change would undo the imperative write. How do you see those two fitting together?

### 5 - `NumberFormSymbolInput.tsx:191`

> `SymbolInput` re-provides both root contexts with different semantics - `value` becomes the magnitude and `setValue` takes a magnitude. `useNumberFormState` / `useNumberFormActions` are exported from `index.tsx` for custom primitives, so they now return two different things under the same field name with nothing at the type level to tell them apart. The cost for what is a two-line sign mapping is also fairly high: the component split in two, a second value mirror, a second `useImperativeHandle` layer and the `context-provider-split-values` suppression. Would passing the mapping into the hook, or a small `useSignedNumber` adapter, get the same result more cheaply?

### 6 - `useNumberFormInputLogic.ts:41`

> `selection` lives in the hook's `useState` and the real input handle is a local `useRef` in the primitive, so a sibling can neither read the caret nor focus the field. That's exactly what `updateValueNumberPad` and `updateLongPressHandlerState` need (`NumberWithSymbolForm.tsx:376-406`), along with the `shouldUpdateSelection` guard at `:449`. Is the plan to lift those into the root before `NumberForm.NumberPad` lands?

### 7 - `NumberFormSymbolInput.tsx:163`

> Legacy keeps the sign outside the value: the component emits the magnitude and the parent owns `isNegative` / `toggleNegative` / `clearNegative`, with `handleNegativeAmountFlipping` toggling the parent's flag. This primitive sets the sign instead of toggling it, so typing `-` on an already negative value no longer flips it, and `updateNumber('12.50')` - which is what `MoneyRequestAmountInput.tsx:202` passes - drops the sign from `getNumber()`. There's also no `isNegative` / `toggleNegative` seam in `NumberFormSymbolInputProps`. Which side do you want owning the sign going forward?

### 8 - `useNumberFormInputLogic.ts:30`

> This hook covers locale normalization and validation, caret state and the manual-update guard, the focus-return selection reset, decimals renormalization, forward-delete detection, and the ref and blur plumbing. That makes the hook the real unit of design rather than the component tree, so a future variant that needs different validation has to take all six or fork it. The return shape leans the same way: `onSubmitEditing` can't be used by `SymbolInput` at all, and `numberFormat` / `errorText` are read from context anyway. Any appetite for pulling the caret and forward-delete pieces out?

### 9 - `useNumberFormInputLogic.ts:25`

> `getNewSelection`, the forward-delete detection block and `NumberFormRef` are character-for-character copies of the legacy versions, and both implementations will coexist for the whole migration since `AmountForm.tsx:147` still routes currency-button variants to legacy. This is the code that historically needed fixes - the comments still reference #16385 and #16974 - so each future fix has to be found and applied twice. Would it be worth extracting the pure helpers into a shared module for the transition?

### 10 - `NumberForm.tsx:29`

> `onSubmitEditing` is declared on the root, put into the actions context, re-provided by `SymbolInput` and returned by the hook, but `NumberFormSymbolInputContent` never forwards it - so `<NumberForm onSubmitEditing={submit}><NumberForm.SymbolInput/></NumberForm>` typechecks and does nothing. It isn't a one-liner either, since `BaseTextInputWithSymbolProps` doesn't include `onSubmitEditing`. Related: `onBlur` is merged in the hook while `onSubmitEditing` is merged in `TextInput`, so there's no single owner for merging form callbacks. Drop it from the root contract for now, or wire it through?

### 11 - `context/types.ts:28`

> Two small things on the context contract. `setValue` returns the previous committed value, and its only consumer is the caret math at `useNumberFormInputLogic.ts:84` - but honouring it means the root carries `committedValueRef` plus a layout effect mirroring its own state, and `SymbolInput` repeats the whole pattern with `canonicalValueRef`. Separately, `externalValue` is public state purely so one effect can read the raw prop. Both are things every future provider layer has to reproduce. Could the caret keep the previous length locally, and could the decimals renormalization move to the layer that owns the prop?

### 12 - `NumberFormSymbolInput.tsx:100`

> `errorText` is part of the shared state contract but has no presentation owner: `TextInput` delegates to the input's built-in `errorText`, while `SymbolInput` returns a bare `FormHelpMessage` with no `isError` and no styling prop. Legacy rendered it as an absolutely positioned sibling of the amount row (`NumberWithSymbolForm.tsx:753`), so dropping `SymbolInput` into that `flexRow` would put the message beside the amount with no prop to correct it. Your TODO already points at `NumberForm.Error` - is it worth doing while there are no consumers yet?

### 13 - `NumberForm/types.ts:65`

> `NumberFormSymbolInputProps` has `prefixContainerStyle` but not `prefixCharacter`, `prefixStyle`, `onMouseDown` or `onMouseUp`. `SplitAmountInput` uses the first two through `MoneyRequestAmountInput`, and the mouse handlers call `setMouseDown` / `setMouseUp` with `stopPropagation` so `BaseListItem` can suppress the row press. Without them the split-amount rows would lose the currency prefix, and selecting text in the amount field would trigger the row's `onPress`. Are those planned for a follow-up?

### 14 - `NumberFormTextInput.tsx:11`

> Both primitives destructure and re-list roughly 25-30 props, though only a handful are actually transformed; `AmountTextInput.tsx:79` does the same job with `...rest`. Nothing is dropped today, but adding a prop takes three coordinated edits and missing the third produces no type error - the prop is documented, typechecks and is silently ignored. That is what happened with `onSubmitEditing` above. Would a rest spread for the pure pass-through props be easier to keep honest?

### 15 - `useNumberFormInputLogic.ts:102`

> The manual-selection guard now clears in a layout effect after commit, whereas legacy only cleared it once `handleSelectionChange` had actually swallowed an event (`NumberWithSymbolForm.tsx:342-346`, `:445`). If a platform delivers `onSelectionChange` in a later JS batch than `onChangeText`, the stale caret would win. If both land in the same batch the behaviour is identical, so this may well be fine - could you confirm on a physical iOS and Android device by typing into the middle of an existing number on one of the migrated pages?

</details>

dariusz-biela (commented):
Overall, it looks okay, but I think we could achieve much more by splitting this component into two independent components. 

Just to test it out, I wrote a quick prototype:

  - NumberInputCore
      - Contains shared, presentation-independent number logic
      - Handles draft parsing, normalization, validation, selection, and keyboard behavior
      - Is reused by both input variants

  - NumberField
      - Covers standard text-input use cases
      - Owns its draft value, constraints, selection, and input reference
      - Provides compound primitives for the input, error, flip button, and symbol action

  - NumberComposer
      - Covers symbol-based and number-pad use cases
      - Owns the complete signed value, selection, constraints, and input reference
      - Provides compound primitives for the symbol input, number pad, controls, error, and responsive layout
      - Supports custom composition for portrait/landscape layouts, actions, and footers

Overall, there are probably a whole bunch of shortcomings in it, but visually, this split makes sense to me. If you’d like, I can share this preliminary version with you, but I think it would be worth it for you to form your own opinion as well.

I believe this architecture should provide us with: 
  1. Clear responsibilities — standard inputs and number-pad experiences no longer share an ambiguous API.
  2. Single state owner — each root controls its draft, sign, constraints, selection, and input reference.
  3. Reusable shared logic — normalization, parsing, validation, and selection behavior are implemented once in NumberInputCore.
  4. Flexible composition — compound primitives allow each consumer to assemble only the UI elements it needs.
  5. Safer evolution — changes to one interaction model do not increase complexity or introduce regressions in the other.

src/components/NumberForm/context/types.ts
jmgraa: Prop is no longer part of public context, explained in code

src/components/NumberForm/hooks/useNumberFormInputLogic.ts
jmgraa: The given logic has been moved to `useNumberEditController.ts`. However, I didn't break down this hook. In my opinion, functionalities like selection, text editing, validation are tightly coupled, and splitting them would cause more confusion

src/components/NumberInput/hooks/useNumberEditController.ts
```diff
@@ -0,0 +1,255 @@
...
+const toCanonicalValueDefault = (displayText: string) => displayText;
+
+/** Returns the new selection based on length delta. */
+const getNewSelection = (oldSelection: NumberInputSelection, previousLength: number, newLength: number): NumberInputSelection => {
+    const cursorPosition = oldSelection.end + (newLength - previousLength);
+    return {start: cursorPosition, end: cursorPosition};
+};
+
+/** Controller managing numeric input state, formatting, validation, and cursor selection. */
+function useNumberEditController({
```
dariusz-biela: This hook owns value editing, validation, the whole selection state machine, forward-delete detection and the decimals effect. #98170 and #98171 will grow it further unless we draw a line now.                    
                                                                                                                                                                                                                      
  Please split it by responsibility:                                                                                                                                                                                  
                                                                                                                                                                                                                      
  - Pure helpers: text normalization, key detection, selection math. No React.                                                                                                                                        
  - Selection hook: everything about the caret and native selection events.                                                                                                                                           
  - Controller: the value, its validation and commit. It talks to the selection hook only when an edit changes the displayed text.                                                                                    
                                                                                                                                                                                                                      
  Anything about buttons, currency or layout stays out of the controller and composes from its public API in the root or the primitive, like toggleSign already does.

src/components/NumericField/context/types.ts
```diff
@@ -0,0 +1,62 @@
...
+
+type NumberFormStateContextValue = {
+    /** The canonical signed value owned by the root. */
+    value: string;
+
+    /**
+     * The `value` prop exactly as passed to the root, before any internal editing state. The decimals-change logic
+     * reads this the way the legacy component read its own `number` prop, so an intentionally empty field is left alone.
+     */
+    externalValue: string;
```
dariusz-biela, line 15: Nothing reads `externalValue` from either context, only the two roots copy it in, and the decimals effect already reads the prop in the controller. Please drop it from both context types and the controller return. Add it back when a primitive needs it.

src/components/NumericInput/primitives/NumericSymbolInput.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const {formattedNumber, isNegative, selection} = useNumberComposerState();
+    const {clearSign, handleBlur, handleKeyPress, handleSelectionChange, inputRef, setNumber} = useNumberComposerActions();
+    const textInput = useRef<BaseTextInputRef | null>(null);
+
+    const inputPosition = position === 'suffix' ? CONST.TEXT_INPUT_SYMBOL_POSITION.SUFFIX : CONST.TEXT_INPUT_SYMBOL_POSITION.PREFIX;
+
+    const handleInputKeyPress = (event: NumberInputKeyPressEvent) => {
+        const key = event.nativeEvent.key.toLowerCase();
+
+        if (!textInput.current?.value && key === 'backspace' && isNegative) {
```
dariusz-biela, line 74: `textInput.current?.value` reads the native input while the displayed text is already in context. `formattedNumber === ''` gives the same answer, drops the local ref and the third `mergeRefs` argument, and does not depend on `BaseTextInputRef.value` per platform.

src/components/NumericField/primitives/NumericTextInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const handleSubmitEditing = (event: Parameters<NonNullable<BaseTextInputProps['onSubmitEditing']>>[0]) => {
+        inputOnSubmitEditing?.(event);
+        onSubmitEditing?.(event);
+    };
+
+    return (
+        <TextInput
+            accessibilityLabel={accessibilityLabel}
+            autoCapitalize="words"
```
dariusz-biela, line 70: `autoCapitalize="words"` on a decimal input looks like a mistake without the legacy comment. Please carry it over: on Android it is required with `keyboardType="decimal-pad"` / `inputMode="decimal"` to avoid input lag.

src/components/AmountForm.tsx
```diff
@@ -109,6 +110,39 @@ function AmountForm({
     const styles = useThemeStyles();
     const {getCurrencyDecimals} = useCurrencyListActions();
     const decimals = decimalsProp ?? getCurrencyDecimals(currency);
+    const symbol = getLocalizedCurrencySymbol(preferredLocale, currency) ?? '';
+
+    // Use the new composable NumberForm for the standard text-input path as part of the NumberWithSymbolForm migration.
+    // Keep the legacy component for currency-button variants until those additional controls are supported by NumberForm.
+    if (displayAsTextInput && !shouldShowCurrencyButton) {
```
dariusz-biela: Fine as a temporary fork. Please add a TODO pointing at #98170 and say the `NumberForm` branch has no trailing currency button yet, so `shouldShowCurrencyButton` keeps the legacy path.

src/components/NumberComposer/NumberComposer.tsx
```diff
@@ -0,0 +1,132 @@
...
+    /** Reference to the underlying text input, supplied by InputWrapper. */
+    ref?: ForwardedRef<BaseTextInputRef>;
+
+    /** Reference exposing the number editing imperative API. */
+    numberFormRef?: ForwardedRef<NumberInputRef>;
+
+    children: ReactNode;
+};
+
+function NumberComposer({value = '', onInputChange, allowNegative = false, decimals = 0, maxLength, errorText, onBlur, onSubmitEditing, ref, numberFormRef, children}: NumberComposerProps) {
```
dariusz-biela: Non-blocking: `stateContextValue`, the shared part of `actionsContextValue` and `useImperativeHandle` are copied between the roots. If the controller returned ready `state`/`actions`, each root would only add its own (`toggleSign/clearSign` here). Fine to defer to the pad follow-up.

Also `numberFormRef` on a Composer: `composerRef`, or one neutral name for both roots.

src/components/NumberInput/hooks/useNumberEditController.ts
```diff
@@ -0,0 +1,255 @@
...
+            return;
+        }
+
+        const operatingSystem = getOperatingSystem();
+        const isMacOrIOS = operatingSystem === CONST.OS.MAC_OS || operatingSystem === CONST.OS.IOS;
+        // Control-D is a macOS/iOS hardware keyboard shortcut for forward-delete.
+        forwardDeletePressedRef.current = key === 'delete' || (isMacOrIOS && !!event.nativeEvent.ctrlKey && key === 'd');
+    };
+
+    useLayoutEffect(() => {
```
dariusz-biela: NIT: `applyValue` already writes `committedValueRef`. Only the reset branch at line 89 bypasses it; set the ref there and this layout effect can go. Defensive as is, fine to keep.

src/components/NumberComposer/primitives/NumberComposerError.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const styles = useThemeStyles();
+    const {errorText} = useNumberComposerState();
+
+    if (!errorText) {
+        return null;
+    }
+
+    return (
+        <FormHelpMessage
+            style={[styles.ph5, styles.w100, style]}
```
dariusz-biela: NIT: `ph5 w100` matches both legacy layouts, so fine for #98171. Once the layout primitive exists it may be cleaner for the layout to own padding and keep this primitive style-free. No action now.

src/components/NumberInput/hooks/useNumberEditController.ts
```diff
@@ -0,0 +1,255 @@
...
+    const adjustAmountOnDecimalsChange = useEffectEvent((newDecimals: number) => {
+        // Skip if empty or already valid for the new decimal precision.
+        if (externalValue === '' || validateAmount(currentValue, newDecimals, maxLength, allowNegative)) {
+            return;
+        }
+
+        setNumber(toDisplayText(stripDecimalsFromAmount(currentValue)));
+    });
+
+    useEffect(() => {
```
dariusz-biela: NIT: the hook tracks previous values in render elsewhere (`previousExternalValue`, `previousIsFocused`); the decimals change uses ref + effect + `useEffectEvent`. The effect is needed (it notifies the parent), so consistency only.

src/components/NumberInput/types.ts
```diff
@@ -0,0 +1,64 @@
+import type {BaseTextInputProps, BaseTextInputRef} from '@components/TextInput/BaseTextInput/types';
+
+import type {ForwardedRef} from 'react';
+import type {StyleProp, TextStyle} from 'react-native';
+
+type NumberInputPosition = 'prefix' | 'suffix';
+
+type NumberInputRef = {
```
dariusz-biela: NIT: `getNumber` exists only because the value is semi-controlled. A one-line JSDoc saying it is legacy parity and new consumers should read their own state would stop it from spreading.

src/components/NumberComposer/context/types.ts
```diff
@@ -0,0 +1,68 @@
...
+
+type NumberComposerStateContextValue = {
+    /** The canonical signed value owned by the root. */
+    value: string;
+
+    /**
+     * The `value` prop exactly as passed to the root, before any internal editing state. The decimals-change logic
+     * reads this the way the legacy component read its own `number` prop, so an intentionally empty field is left alone.
+     */
+    externalValue: string;
```
dariusz-biela: If possible, I would try to avoid this hack and come up with something better in terms of code quality.

jmgraa:
This PR is being split into two smaller PRs following code review. The first follow-up covers the `NumberForm`/`NumericField` foundation, shared numeric editing controller and standard text-input integration for https://github.com/Expensify/App/issues/98169. The second PR will handle the remaining symbol controls and number-pad composition.

Follow-up PR: https://github.com/Expensify/App/pull/99772

PR #98971, opened 2026-08-19
Fix staging server toggle resetting on cold start in release builds

mikolajpochec:
### Explanation of Change

Fixes the staging toggle resetting to production on Android GH-release builds, and being off by
default on a fresh install of a staging build.

Staging and production ship the same binary, so the app has to work out at runtime which one it is.

The old version compared version from `package.json` with the latest GH release. It now looks at how the build was installed. Anything the Play Store didn't install was handed to a tester directly, so it is a staging build. That is fixed at install time, needs no network, and never expires. Play Store installs still go through the version comparison, because
they can be on a testing track.

Two smaller fixes: `ApiUtils` derives the flag when asked instead of caching it, and `getEnvironment`
always settles its promise.

OldDot part: Expensify/Mobile-Expensify#14062 — both are needed for HybridApp.

Behavior changes vs main:

- Installer ≠ `com.android.vending` (sideload, `adb install`, null → `"unknown"`) resolves STAGING
  regardless of version — including a sideloaded production APK. `com.android.vending` is
  characteristic of apps distributed via Play Store.
- Play Store installs still fall through to the version comparison. That is what covers testers on
  the Play alpha/beta/Internal Testers tracks, whose installer is `com.android.vending` like a real
  user's.
- `ONYXKEYS.IS_BETA` preserved across sign-out, so a resolved verdict survives a re-login.
- A GitHub response with no usable tag keeps the last saved verdict instead of overwriting it with
  `false`. It previously resolved `false` and then fell into `semver.gt(version, undefined)`, which
  threw — the `.catch` fallback never applied, since the promise had already settled. A rate-limited
  403 is exactly this case, so one throttled request no longer demotes a tester to production.
- `undefined` values in the OldDot transition payload are skipped rather than merged as `{}`, which
  left a truthy non-boolean in `SHOULD_USE_STAGING_SERVER`. `null` passes through, since Onyx reads
  a merged `null` as a key removal.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97891
PROPOSAL:

MOBILE-EXPENSIFY: https://github.com/Expensify/Mobile-Expensify/pull/14062

### Tests

Android only — iOS reads the TestFlight receipt and is unchanged by this PR. Install by tapping the
Android build link on this PR, or build an apk package and sideload it via `adb`.

 When testing HybridApp behaviour, the OldDot PR has to be in the same build — linked above with `MOBILE-EXPENSIFY`: so CI builds both halves.

**Warning**: the build triggered from this PR is an `adhoc` build. Both apps return early on `ENVIRONMENT=adhoc`, before the install-source check this PR adds, and `adhoc` already defaults to the staging server — so the toggle will look correct without exercising any of the new code. The persistence and transition sections below are still valid on it; the install-source and GitHub sections are not.

To test the detection itself, build release locally (`./gradlew assembleRelease` from `Mobile-Expensify/Android`, which uses `ENVIRONMENT=production`) and sideload that. Then a non-Play installer resolves `STAGING`. Test cases are described below:

<details>
<summary>Toggle defaults on, and stays on</summary>

1. Settings → Troubleshoot: **Use Staging Server** is ON, and the API root pill reads staging.
2. Force-close, reopen. Still ON, still staging.

</details>

<details>
<summary>A deliberate opt-out outlives the default</summary>

1. Turn the toggle OFF.
2. Force-close, reopen. Still OFF.
3. Sign out, sign back in. Still OFF.

</details>

<details>
<summary>OldDot does not overwrite the choice</summary>

1. Note the toggle state in New Expensify.
2. Switch to Expensify Classic, then back. Unchanged.
3. In Classic, flip its **Use Staging Server** row, then switch to New Expensify. The flip carries
   over — a deliberate change propagates, an absent value never overwrites.

</details>

<details>
<summary>Play Store installs are unchanged</summary>

Needs a build installed through a Play Store. This can be emulated by uninstalling the app first, then installing the apk with `adb install -r -i com.android.vending <apk>` — `-i` only takes effect on a fresh install.

1. Sign in. The install source no longer decides; the version comparison does, exactly as before
   this PR.
2. Which way it lands depends on the build: a real staging build is ahead of the latest release and
   resolves staging, while a build behind it resolves production and hides the toggle. Compare
   `package.json` against `https://api.github.com/repos/expensify/app/releases/latest` before reading
   the result.

</details>

### Offline tests

The environment verdict must not move when the network does. On a sideloaded release build:

1. **Offline cold start.** With the toggle ON, enable airplane mode, force-close, reopen. Environment
   still staging, toggle still ON — the install source is read locally, so nothing waits on GitHub.
2. **Offline flip.** Still offline, turn the toggle OFF. Force-close, reopen. Still OFF.
3. **Back online.** Re-enable the network and reopen. Nothing changes: a successful GitHub response
   cannot override the install source.
4. **Spotty / throttled.** Exhaust the unauthenticated GitHub limit (see the Tests section), then
   cold start. The stored verdict is kept instead of being overwritten with production, so one
   throttled request no longer demotes a tester. Before this PR the same response wrote
   `IS_BETA = false`, which also poisoned later launches.

### QA Steps

Android only — iOS is unchanged by this PR. Unlike the adhoc build above, the staging build QA
receives is `ENVIRONMENT=production`, so it does exercise the new code. No shell access needed:
"force-close" means swiping the app away in recents.

<details>
<summary>Toggle is on by default</summary>

1. Install the staging build and sign in.
2. Settings → Troubleshoot: **Use Staging Server** is ON.
3. Force-close and reopen. Still ON.

</details>

<details>
<summary>Turning it off sticks</summary>

1. Turn **Use Staging Server** OFF.
2. Force-close and reopen. Still OFF.
3. Sign out and sign back in. Still OFF.

</details>

<details>
<summary>Classic and New Expensify agree</summary>

Needs a staff account — the row is hidden in Classic otherwise.

1. Switch to Expensify Classic. Its **Use Staging Server** row matches New Expensify.
2. Flip it in Classic, then switch back to New Expensify. The flip carried over.

</details>

<details>
<summary>Offline</summary>

1. Enable airplane mode.
2. Force-close and reopen. Toggle and API root unchanged.

</details>

<details>
<summary>Regression check: production build</summary>

1. On a production build from the Play Store, **Use Staging Server** is not offered and requests go
   to production. Nothing in this PR changes that.

</details>

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


Julesssss:
Hi @mikolajpochec could you add the test steps and fix checks please, I'll generate builds in the meantime

mikolajpochec:
> Hi @mikolajpochec could you add the test steps and fix checks please, I'll generate builds in the meantime

Hi @Julesssss, I'm still waiting for our SWM internal review. I'll add test steps soon, and I'll open this draft

src/libs/actions/Session/index.ts
```diff
@@ -792,10 +793,17 @@ function setupNewDotAfterTransitionFromOldDot(hybridAppSettings: HybridAppSettin
             ];
 
             for (const [key, value] of Object.entries(newDotOnyxValues)) {
+                // `?? {}` used to stand in for a missing value, but these keys are not all objects —
+                // merging `{}` into SHOULD_USE_STAGING_SERVER leaves a truthy non-boolean behind. There is
+                // nothing to apply when OldDot sends nothing, so skip the key instead.
```
war-in: This sounds like a comment you could add here in GH to explain why the change was made, but it shouldn't land in the code. Once the PR is merged, everyone will wonder what `?? {}` was and why this is mentioned here

src/libs/actions/Session/index.ts
```diff
@@ -792,10 +793,17 @@ function setupNewDotAfterTransitionFromOldDot(hybridAppSettings: HybridAppSettin
...
+                if (value === undefined || value === null) {
```
war-in: I wonder if we sometimes would want to send `value: null` to clear the onyx entry 🤔 if yes, the value won't be applied and we'll waste time debugging
Shouldn't we only check for `undefined`?

src/libs/Environment/betaChecker/index.android.ts
```diff
@@ -24,18 +25,47 @@ Onyx.connectWithoutView({
     },
 });
 
+/**
+ * Whether something other than the Play Store put this build on the device. Production builds only ever reach a
+ * device through the Play Store, so any other installer means the build was sideloaded, i.e. downloaded from a
+ * GitHub release. Unlike the version comparison below, the answer does not change once a production release catches
+ * up with the build's version — though Android does rewrite the installer if a different one later updates the app.
+ */
```
war-in, line 10: I wouldn't compare this function to the other in the description comment. Someone might change the other function and the comment would be outdated

src/libs/Environment/betaChecker/index.android.ts
```diff
@@ -24,18 +25,47 @@ Onyx.connectWithoutView({
...
+function isSideloadedBuild(): boolean {
```
war-in: How about naming it `isProductionBuild`?

src/libs/Environment/betaChecker/index.android.ts
```diff
@@ -24,18 +25,47 @@ Onyx.connectWithoutView({
...
+        return false;
+    }
+}
+
 /**
  * Check the GitHub releases to see if the current build is a beta build or production build
  */
 function isBetaBuild(): IsBetaBuild {
     return new Promise((resolve) => {
+        // A sideloaded build is a beta build: testers install those from GitHub prereleases. Answering here also
```
war-in: It could land in the function description

src/libs/Environment/betaChecker/index.android.ts
```diff
@@ -24,18 +25,47 @@ Onyx.connectWithoutView({
...
+
+        // Otherwise compare our version against the latest production release
         fetch(CONST.GITHUB_RELEASE_URL)
             .then((res) => res.json())
             .then((json: GithubReleaseJSON) => {
                 const productionVersion = json.tag_name;
-                if (!productionVersion) {
+
+                // A rate limited or malformed response carries no usable tag. Production is the safe answer when
+                // we cannot tell — the missing `return` here used to fall through into semver.gt(version, undefined),
```
war-in: Wasn't that intentional? To fall back to `isLastSavedBeta`? Also, this comment will get outdated when we merge the PR

war-in (commented):
Good job overall! I'd just review all comments and check if they can be shortened/removed

src/libs/actions/Session/index.ts
mikolajpochec: Missed this case, thanks

src/libs/Environment/betaChecker/index.android.ts
mikolajpochec: Good idea, more intuitive. I went further and renamed it to `isPlayStoreInstall` to better reflect what this function actually checks

src/libs/Environment/betaChecker/index.android.ts
mikolajpochec: I think that original solution did that by mistake, notice how there wasn't a `return` after `resolve(false);`. That probably caused error being thrown on `const isBeta = semver.gt(pkg.version, productionVersion);`, which made this fallback possible.
Nevertheless I think that we should honor tester's choice in this case, and I'll introduce this in the incoming commits.

I'm addressing the AI comment too.

src/libs/Environment/betaChecker/index.android.ts
```diff
@@ -25,17 +26,41 @@ Onyx.connectWithoutView({
...
+/**
+ * Whether this build is a beta (staging) build.
+ *
+ * A sideloaded build is answered straight away, both because testers install those from GitHub prereleases and
+ * because it keeps them off the rate limited GitHub API. Anything the Play Store installed can still be on a
+ * tester track, which only a comparison against the newest production release recognizes.
  */
 function isBetaBuild(): IsBetaBuild {
     return new Promise((resolve) => {
+        if (isPlayStoreInstall() === false) {
```
marufsharifi: **Behavioral change to consciously sign off on:** any non-Play-Store install now resolves to **STAGING regardless of version — including a sideloaded *production* APK.**

```ts
if (isPlayStoreInstall() === false) {
    AppUpdate.setIsAppInBeta(true);
    resolve(true);
    return;
}
```

Per the PR description this is intended ("sideload = tester = staging"), and dev/adhoc builds short-circuit earlier in `getEnvironment`, so only `ENVIRONMENT=production` native builds reach this path. But the practical consequence is that **anyone who sideloads a production APK for any reason (repro, QA, debugging) will now transparently talk to staging servers and see the toggle ON.** That changes what "install the production APK to reproduce" means going forward. Not a code defect — just the single most important semantic shift in this PR, so let's make sure it's an explicit product decision and not an accidental side effect.

src/libs/actions/Delegate.ts
```diff
@@ -49,6 +49,7 @@ const KEYS_TO_PRESERVE_DELEGATE_ACCESS = [
     // This allows the report screen to load correctly when the delegate token expires and the delegate is returned to their original account.
     ONYXKEYS.RAM_ONLY_IS_SIDEBAR_LOADED,
     ONYXKEYS.NETWORK,
+    ONYXKEYS.IS_BETA,
```
marufsharifi: **Test-coverage gap.** `ONYXKEYS.IS_BETA` is added to four preserve-lists, but only `App.ts`'s exported `KEYS_TO_PRESERVE` is unit-tested (`KeysToPreserveTest`). The three siblings are untested:

- `KEYS_TO_PRESERVE_DELEGATE_ACCESS` — here (`src/libs/actions/Delegate.ts:52`)
- `KEYS_TO_PRESERVE_SUPPORTAL` — `src/libs/actions/Session/index.ts:346`
- the inline `keysToPreserve.push(...)` — `src/libs/actions/SignInRedirect.ts:61`

These are exactly the "one line in a list that is easy to drop by accident" the test comment describes, and three of the four copies aren't guarded. Low risk (they're not exported, so a shared test needs a small refactor), but worth closing since a delegate/supportal sign-out would silently lose the verdict without any failing test.

src/libs/actions/Session/index.ts
```diff
@@ -792,10 +793,14 @@ function setupNewDotAfterTransitionFromOldDot(hybridAppSettings: HybridAppSettin
             ];
 
             for (const [key, value] of Object.entries(newDotOnyxValues)) {
+                if (value === undefined) {
+                    continue;
+                }
```
marufsharifi, line 829: **Verify the blast radius — this branch changes behavior for _every_ key in the transition payload, and isn't covered by a test.**

```ts
if (value === undefined) {
    continue;
}
```

`newDotOnyxValues` is `{...hybridAppSettings}` minus `hybridApp` (line 683), i.e. every NVP key OldDot sends — not just `SHOULD_USE_STAGING_SERVER`. Skipping `undefined` (while letting `null` pass through as an Onyx key-removal) is strictly safer than the old `value ?? {}`, which clobbered a boolean key with a truthy `{}`. But since it applies to the whole payload, please sanity-check that no downstream consumer relied on an omitted key being initialized to `{}` rather than staying `undefined`. Since no test exercises this loop, a one-off unit test asserting (a) `undefined` is skipped, (b) `null` produces a removal, (c) a real value merges — would lock in the fix.

src/libs/Environment/betaChecker/index.android.ts
mikolajpochec: It is an intentional decision, not a side effect. Every option here has trade-offs, but I'd argue the install-source check is more robust than the GitHub fetch it replaces.

Separately, and not blocking this PR: I'd like to go further and drop the fetch entirely, leaving one deterministic signal. The catch is that Play testing-track testers are exactly who the fetch is exists for. Their installer is `com.android.vending`, same as a real user's. So the question is whether anyone actually installs staging builds from the Play alpha/beta/Internal Testers tracks, or whether testers use the GitHub APK. If it's the latter, the fetch has no population left to serve and can go. Do you know who owns that?

src/libs/actions/Session/index.ts
mikolajpochec: Checked, nothing relied on it. Added tests too.

marufsharifi (approved):
LGTM

src/libs/Environment/betaChecker/index.android.ts
Julesssss: > anyone who sideloads a production APK for any reason (repro, QA, debugging)

Hmm yeah that is quite a large change. Staging makes more sense but devs/QA do depend on this being prod so we'd need to let everyone know at a minimum. But @mikolajpochec this still overridable by the `use staging` checkbox, right? Which makes this a lot less of a problem

luacmartins (dismissed):
LGTM

src/libs/Environment/betaChecker/index.android.ts
Julesssss: The check also excludes 3rd party stores I guess

Julesssss:
> @Julesssss all yours

Just waiting on some feedback to the linked Mobile-Expenisfy [PR](https://github.com/Expensify/Mobile-Expensify/pull/14062)

JmillsExpensify (commented):
No product review required.

src/libs/Environment/betaChecker/index.android.ts
mikolajpochec: > Hmm yeah that is quite a large change. Staging makes more sense but devs/QA do depend on this being prod so we'd need to let everyone know at a minimum. But @mikolajpochec this still overridable by the use staging checkbox, right? Which makes this a lot less of a problem

Yeah, this is completely overridable. Moreover, there’s always the four‑finger gesture on ND that allows you to toggle this at any time

> The check also excludes 3rd party stores I guess

Hm, you're right. This kind of check could be added for all major third party stores. Should I do that?

mikolajpochec: > Hm, you're right. This kind of check could be added for all major third party stores. Should I do that?

I’ve done a bit of research, and I don’t think we distribute the production app outside the Play Store. Is it really a concern in that case @Julesssss?

Julesssss:
Discussion [here](https://github.com/Expensify/Mobile-Expensify/pull/14062#issuecomment-5458216084).

src/libs/Environment/betaChecker/index.android.ts
Julesssss: Nah agree it's not a problem, it would only be unauthorized uploads

Julesssss:
@mikolajpochec sorry for the delay, could you please merge main and resolve conflicts

luacmartins (dismissed):

Julesssss (dismissed):
Looking good, we've discussed the tradeoffs. Just need to resolve the conflicts now.

luacmartins (dismissed):

luacmartins:
@mikolajpochec conflicts

luacmartins:
@mikolajpochec eslint is failing too

mikolajpochec:
@luacmartins Seems to be alright now

luacmartins (dismissed):

luacmartins:
The failed workflow was discussed [here](https://expensify.slack.com/archives/C03TQ48KC/p1788976681377559). We'll proceed with a merge and stop this workflow on forks as a follow up.

luacmartins:
lol nvm, @mikolajpochec more conflicts

mikolajpochec:
@luacmartins @Julesssss Ready to merge now. [The second PR](https://github.com/Expensify/Mobile-Expensify/pull/14062) too.

Julesssss (approved):
Thanks, just waiting on Mobile-Expensify checks 🤞

mikolajpochec:
> Thanks, just waiting on Mobile-Expensify checks 🤞

@Julesssss Checks passed! :)

kavimuru:
Covered with https://github.com/Expensify/Mobile-Expensify/pull/14062

PR #97883, opened 2026-08-05
Land Looking-Around users in Spend > Expenses and add "Send to someone" to a tracked expense's More menu

MelvinBot:
### Explanation of Change

Fixes two issues a brand-new "Something else" (`LOOKING_AROUND`) user hits after creating their first expense.

**Bug 1 — new user lands in Personal Space instead of Spend › Expenses.**
A `LOOKING_AROUND` user has no report or workspace, so after onboarding they land on the **Home** page (Inbox). When they then create an expense from Home, that expense lands in their self-DM, and the app was dropping them into that self-DM (Personal Space) instead of Spend › Expenses.

Fix: when a `LOOKING_AROUND` user creates an expense whose sole destination is their self-DM, route them to **Spend › Expenses** (Search) instead of the self-DM. This is scoped to `isSelfDMDestination`, so a `LOOKING_AROUND` user who later has a workspace and submits to a real report/friend keeps their normal destination. The routing decision is threaded through all three post-create decision points that must agree with each other — `getSubmitExpensePreMountDestinationRoute` (what is pre-mounted behind the RHP), `getSubmitHandler` (dismiss strategy), and `navigateAfterExpenseCreate` (final landing) — and through every create path: the confirmation step **and** the skip-confirmation flows (scan-and-skip, distance-skip, and quick-amount entry). The self-DM is never pre-inserted behind the RHP for these users, so there's no Personal Space flash before the switch to Search on narrow layout. Onboarding landing (`navigateAfterOnboarding`) is unchanged — these users still land on Home after signup.

The "route to Search only when online" gate is now the single shared helper `isLookingAroundSearchRoutingActive` in `IOUUtils`, so every create path computes the routing flag identically and the decision points cannot disagree. It is offline-suppressed (Search reads a server-populated snapshot unavailable offline, so offline these users fall back to the self-DM landing where the optimistic expense is visible), and it is unit-tested.

**Bug 2 — "Send to someone" missing from a tracked expense's More menu.**
When viewing an unreported self-tracked expense (Spend › View Expense), the More menu had no way to send it to someone, even though the track-expense actionable whisper offers exactly that from the Inbox.

[...]

PR #99247, opened 2026-08-22
Add Send to my employer option to expense More menu

shawnborton:
### Explanation of Change

When a "Looking Around" user (someone who chose the "Something else" onboarding intent) creates an expense, they land on the Spend page. Opening the expense and its **More** menu previously offered only **Send to someone** — but the personal-space Concierge in Inbox offers two options: **Send to someone** and **Submit to my employer**.

This PR adds a **Submit to my employer** option to the expense More menu, reusing the exact same flow as the Inbox Concierge "Submit to my employer" action. Both call `createDraftTransactionAndNavigateToParticipantSelector` with `actionName: SUBMIT` and `submitDestination: EMPLOYER`, so the destination resolves identically:
- No accessible workspace: a new Submit workspace is created and the expense drops into its draft confirmation screen.
- Exactly one workspace: skips the picker and goes straight to that workspace's confirmation screen.
- Multiple workspaces: opens the workspaces-only destination picker.

The new row is gated on the same condition as "Send to someone" (an unreported self-tracked expense in personal space, with write access), and sits in the same top section of the menu.

### Fixed Issues

$ https://github.com/Expensify/App/issues/97881
PROPOSAL:

### Tests

1. Sign up as a new user and select the **Something else** onboarding intent (this qualifies you as a "Looking Around" user).
2. Create an expense — you land on the Spend page.
3. Open the expense, then open the **More** menu.
4. Verify a **Submit to my employer** option appears in the top section, alongside **Send to someone**.
5. Tap **Submit to my employer** and verify it behaves the same as the Inbox Concierge "Submit to my employer" option:
   - With no workspace: a new Submit workspace is created and you land on the confirmation screen.
   - With exactly one workspace: you go straight to that workspace's confirmation screen.
   - With multiple workspaces: the workspaces-only destination picker opens.
6. Open the same expense from the Inbox personal space via Concierge and confirm the "Submit to my employer" outcome matches the More-menu one.

### Offline tests

1. Go offline.
2. Open a self-tracked expense's **More** menu and verify **Submit to my employer** still appears (it is built from local Onyx state).
3. Select it and verify the flow proceeds and the request is queued, matching the Inbox behavior offline.

### QA Steps

Same as tests.

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/src/components/MoneyRequestHeaderSecondaryActions.tsx
+++ b/src/components/MoneyRequestHeaderSecondaryActions.tsx
@@ -122,7 +122,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
     const theme = useTheme();
     const {translate, localeCompare, formatPhoneNumber, dateFnsLocale} = useLocalize();
     const isInSidePanel = useIsInSidePanel();
-    const {login: currentUserLogin, accountID, localCurrencyCode, displayName: currentUserDisplayName} = useCurrentUserPersonalDetails();
+    const {login: currentUserLogin, email: currentUserEmail, accountID, localCurrencyCode, displayName: currentUserDisplayName} = useCurrentUserPersonalDetails();
     const delegateAccountID = useDelegateAccountID();
     const personalDetails = usePersonalDetails();
 
@@ -190,7 +190,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
     const activePolicy = useActivePolicy();
     const lastWorkspaceNumber = useLastWorkspaceNumber();
     const {isRestrictedToPreferredPolicy, preferredPolicyID} = usePreferredPolicy();
-    const filteredPoliciesInfoSelector = useMemo(() => createFilteredPoliciesInfoSelector(currentUserLogin), [currentUserLogin]);
+    const filteredPoliciesInfoSelector = useMemo(() => createFilteredPoliciesInfoSelector(currentUserEmail), [currentUserEmail]);
     const [filteredPoliciesInfo] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {selector: filteredPoliciesInfoSelector});
     const draftTransactionIDs = useMemo(() => Object.keys(transactionDrafts ?? {}), [transactionDrafts]);
 
@@ -354,7 +354,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
     })();
 
     // Shared params for converting a self-tracked expense through the participant selector, used by both
-    // "Send to someone" (friend) and "Send to my employer" (employer) menu actions. Mirrors the Inbox
+    // "Send to someone" (friend) and "Submit to my employer" (employer) menu actions. Mirrors the Inbox
     // track-expense whisper flow (see ChatActionableButtons' baseDraftTransactionParams).
     const sendTrackedExpenseParams = {
         reportID: parentReport?.reportID,
@@ -371,7 +371,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
         preferredPolicyID,
         transaction,
         currentUserAccountID: accountID,
-        currentUserEmail: currentUserLogin ?? '',
+        currentUserEmail: currentUserEmail ?? '',
         currentUserLocalCurrency: localCurrencyCode ?? CONST.CURRENCY.USD,
         filteredPoliciesCount: filteredPoliciesInfo?.filteredPoliciesCount ?? 0,
         firstPolicyID: filteredPoliciesInfo?.firstPolicyID,
@@ -663,7 +663,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
             },
         },
         [CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER]: {
-            text: translate('iou.sendToEmployer'),
+            text: translate('iou.submitToEmployer'),
             icon: expensifyIcons.Building,
             value: CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER,
             onSelected: () => {
@@ -675,7 +675,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
                 createDraftTransactionAndNavigateToParticipantSelector({
                     ...sendTrackedExpenseParams,
                     submitDestination: CONST.IOU.SUBMIT_DESTINATION.EMPLOYER,
-                    defaultWorkspaceName: generateDefaultWorkspaceName(currentUserLogin ?? '', lastWorkspaceNumber, translate, currentUserDisplayName),
+                    defaultWorkspaceName: generateDefaultWorkspaceName(currentUserEmail ?? '', lastWorkspaceNumber, translate, currentUserDisplayName),
                 });
             },
         },
--- a/src/components/MoneyRequestHeaderSecondaryActions.tsx
+++ b/src/components/MoneyRequestHeaderSecondaryActions.tsx
@@ -10,6 +10,7 @@ import useEnvironment from '@hooks/useEnvironment';
 import useGetIOUReportFromReportAction from '@hooks/useGetIOUReportFromReportAction';
 import useHasMultipleSplitChildren from '@hooks/useHasMultipleSplitChildren';
 import useIsInSidePanel from '@hooks/useIsInSidePanel';
+import useLastWorkspaceNumber from '@hooks/useLastWorkspaceNumber';
 import {useMemoizedLazyExpensifyIcons} from '@hooks/useLazyAsset';
 import useLocalize from '@hooks/useLocalize';
 import useMoneyRequestPolicyTagsForReport from '@hooks/useMoneyRequestPolicyTagsForReport';
@@ -33,6 +34,7 @@ import useTransactionViolations from '@hooks/useTransactionViolations';
 import {duplicateExpenseTransaction as duplicateTransactionAction} from '@libs/actions/IOU/Duplicate';
 import {deleteTrackExpense} from '@libs/actions/IOU/TrackExpense';
 import {setupMergeTransactionDataAndNavigate} from '@libs/actions/MergeTransaction';
+import {generateDefaultWorkspaceName} from '@libs/actions/Policy/Policy';
 import initSplitExpense from '@libs/actions/SplitExpenses';
 import {setNameValuePair} from '@libs/actions/User';
 import getNonEmptyStringOnyxID from '@libs/getNonEmptyStringOnyxID';
@@ -120,13 +122,14 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
     const theme = useTheme();
     const {translate, localeCompare, formatPhoneNumber, dateFnsLocale} = useLocalize();
     const isInSidePanel = useIsInSidePanel();
-    const {login: currentUserLogin, accountID, localCurrencyCode} = useCurrentUserPersonalDetails();
+    const {login: currentUserLogin, accountID, localCurrencyCode, displayName: currentUserDisplayName} = useCurrentUserPersonalDetails();
     const delegateAccountID = useDelegateAccountID();
     const personalDetails = usePersonalDetails();
 
     const expensifyIcons = useMemoizedLazyExpensifyIcons([
         'ArrowCollapse',
         'ArrowSplit',
+        'Building',
         'Checkmark',
         'DocumentMerge',
         'ExpenseCopy',
@@ -185,6 +188,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
     const isTrackIntentUser = isTrackOnboardingChoice(introSelected?.choice);
 
     const activePolicy = useActivePolicy();
+    const lastWorkspaceNumber = useLastWorkspaceNumber();
     const {isRestrictedToPreferredPolicy, preferredPolicyID} = usePreferredPolicy();
     const filteredPoliciesInfoSelector = useMemo(() => createFilteredPoliciesInfoSelector(currentUserLogin), [currentUserLogin]);
     const [filteredPoliciesInfo] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {selector: filteredPoliciesInfoSelector});
@@ -348,6 +352,30 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
             hasWorkspaceToSubmitTo,
         });
     })();
+
+    // Shared params for converting a self-tracked expense through the participant selector, used by both
+    // "Send to someone" (friend) and "Send to my employer" (employer) menu actions. Mirrors the Inbox
+    // track-expense whisper flow (see ChatActionableButtons' baseDraftTransactionParams).
+    const sendTrackedExpenseParams = {
+        reportID: parentReport?.reportID,
+        actionName: CONST.IOU.ACTION.SUBMIT,
+        reportActionID: getTrackExpenseActionableWhisper(transaction?.transactionID, parentReport?.reportID, parentReportActions)?.reportActionID,
+        reportActions: parentReportActions,
+        introSelected,
+        draftTransactionIDs,
+        activePolicy,
+        userBillingGracePeriodEnds,
+        amountOwed,
+        ownerBillingGracePeriodEnd,
+        isRestrictedToPreferredPolicy,
+        preferredPolicyID,
+        transaction,
+        currentUserAccountID: accountID,
+        currentUserEmail: currentUserLogin ?? '',
+        currentUserLocalCurrency: localCurrencyCode ?? CONST.CURRENCY.USD,
+        filteredPoliciesCount: filteredPoliciesInfo?.filteredPoliciesCount ?? 0,
+        firstPolicyID: filteredPoliciesInfo?.firstPolicyID,
+    };
 
     const secondaryActionsImplementation: Partial<
         Record<ValueOf<typeof CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS>, DropdownOption<ValueOf<typeof CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS>>>
@@ -631,25 +659,23 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
                     return;
                 }
 
+                createDraftTransactionAndNavigateToParticipantSelector(sendTrackedExpenseParams);
+            },
+        },
+        [CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER]: {
+            text: translate('iou.sendToEmployer'),
+            icon: expensifyIcons.Building,
+            value: CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER,
+            onSelected: () => {
+                if (isDelegateAccessRestricted) {
+                    showDelegateNoAccessModal();
+                    return;
+                }
+
                 createDraftTransactionAndNavigateToParticipantSelector({
-                    reportID: parentReport?.reportID,
-                    actionName: CONST.IOU.ACTION.SUBMIT,
-                    reportActionID: getTrackExpenseActionableWhisper(transaction?.transactionID, parentReport?.reportID, parentReportActions)?.reportActionID,
-                    reportActions: parentReportActions,
-                    introSelected,
-                    draftTransactionIDs,
-                    activePolicy,
-                    userBillingGracePeriodEnds,
-                    amountOwed,
-                    ownerBillingGracePeriodEnd,
-                    isRestrictedToPreferredPolicy,
-                    preferredPolicyID,
-                    transaction,
-                    currentUserAccountID: accountID,
-                    currentUserEmail: currentUserLogin ?? '',
-                    currentUserLocalCurrency: localCurrencyCode ?? CONST.CURRENCY.USD,
-                    filteredPoliciesCount: filteredPoliciesInfo?.filteredPoliciesCount ?? 0,
-                    firstPolicyID: filteredPoliciesInfo?.firstPolicyID,
+                    ...sendTrackedExpenseParams,
+                    submitDestination: CONST.IOU.SUBMIT_DESTINATION.EMPLOYER,
+                    defaultWorkspaceName: generateDefaultWorkspaceName(currentUserLogin ?? '', lastWorkspaceNumber, translate, currentUserDisplayName),
                 });
             },
         },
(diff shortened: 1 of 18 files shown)


src/libs/ReportSecondaryActionUtils.ts
```diff
@@ -1252,17 +1252,19 @@ function getSecondaryTransactionThreadActions({
...
+    // access (like MOVE_EXPENSE) so the rows hide on an archived self-DM.
     const {isExpenseSplit: isSelfDMExpenseSplit} = getOriginalTransactionWithSplitInfo(reportTransaction, originalTransaction);
     if (
         isTrackExpenseReportNew(transactionThreadReport, parentReport, reportAction) &&
         (!isSelfDMExpenseSplit || hasWorkspaceToSubmitTo) &&
         canUserPerformWriteActionReportUtils(parentReport, isChatReportArchived)
     ) {
         options.push(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_SOMEONE);
+        options.push(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER);
     }
```
ikevin127: #### 🟠 `src/libs/ReportSecondaryActionUtils.ts:1261-1268`: `SEND_TO_SOMEONE` should not share the employer gate for splits

The PR claims parity with the Inbox Concierge flow. It does not have parity for one case, and the new comment now asserts the divergent behavior is intentional.

Here is what Inbox actually does (`src/pages/inbox/report/actionContents/ChatActionableButtons.tsx:225-256`):

```tsx
const isSplitExpense = isSplitChildTransaction(trackExpenseTransaction);
const shouldShowSubmitButtons = !isSplitExpense || !!hasWorkspaceToSubmitTo;

{shouldShowSubmitButtons && (
    <>
        {!isSplitExpense && (
            <Button onPress={() => submit(CONST.IOU.SUBMIT_DESTINATION.FRIEND)}>…</Button>
        )}
        <Button onPress={() => submit(CONST.IOU.SUBMIT_DESTINATION.EMPLOYER)}>…</Button>
    </>
)}
```

Note the extra `{!isSplitExpense && …}` guard on the friend button. Inbox hides `Submit to a friend` for a split **unconditionally**, workspace or not.

The More menu pushes both under one gate:

```ts
if (
    isTrackExpenseReportNew(transactionThreadReport, parentReport, reportAction) &&
    (!isSelfDMExpenseSplit || hasWorkspaceToSubmitTo) &&
    canUserPerformWriteActionReportUtils(parentReport, isChatReportArchived)
) {
    options.push(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_SOMEONE);
    options.push(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER);
}
```

| case | Inbox friend | Inbox employer | More menu `SEND_TO_SOMEONE` | More menu `SEND_TO_EMPLOYER` |
| --- | --- | --- | --- | --- |
| not a split | shown | shown | shown | shown |
| split, has workspace | **hidden** | shown | **shown** | shown |
| split, no workspace | hidden | hidden | hidden | hidden |

The employer column matches exactly, which is the part this PR added, so the new code is right. The mismatch is on `SEND_TO_SOMEONE`, and it is pre-existing from #97883 rather than introduced here.

Two reasons I am still raising it on this PR:

1. **The comment rewrite actively codifies it as correct**: *"for a split we still surface both rows when the user has a workspace (the participant selector routes the split into it)"*. That is a new claim, and `ChatActionableButtons` says the opposite for the friend destination.
2. The original gate reads as the **employer** rule applied to the friend row. `"A self-DM split can only go to a workspace, so hide it for a split unless the user has one"` is the reasoning for a workspace destination. Now that `SEND_TO_EMPLOYER` exists to cover the split case properly, the friend row no longer needs to borrow it.

**Suggested split of the two conditions:**

```ts
const {isExpenseSplit: isSelfDMExpenseSplit} = getOriginalTransactionWithSplitInfo(reportTransaction, originalTransaction);
const canConvertFromTrack = isTrackExpenseReportNew(transactionThreadReport, parentReport, reportAction) && canUserPerformWriteActionReportUtils(parentReport, isChatReportArchived);

if (canConvertFromTrack) {
    // A self-DM split has no personal destination, so it can never go to a friend (matches ChatActionableButtons).
    if (!isSelfDMExpenseSplit) {
        options.push(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_SOMEONE);
    }
    // A split can still go to a workspace, but only one that already exists: the create-a-workspace
    // fallback in createDraftTransactionAndNavigateToParticipantSelector is not wired for splits.
    if (!isSelfDMExpenseSplit || hasWorkspaceToSubmitTo) {
        options.push(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER);
    }
}
```

**Bug this causes today:** a user with a self-DM split expense and at least one workspace opens **More** and sees **Send to someone**. Tapping it runs `createDraftTransactionAndNavigateToParticipantSelector` with `submitDestination` defaulting to `FRIEND`, which skips the employer branch at `ReportUtils.ts:12174` and opens the people picker.

**Inbox deliberately never offers that path for a split. Either the More menu is wrong or Inbox is, and right now they disagree.**

If the issue / PR owner decides the current behavior is deliberate, please say so in the comment with the reason rather than the `participant selector routes the split into it` phrasing, because that is the claim I could not confirm anywhere in the flow.

src/components/MoneyRequestHeaderSecondaryActions.tsx
```diff
@@ -185,6 +188,7 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
     const isTrackIntentUser = isTrackOnboardingChoice(introSelected?.choice);
 
     const activePolicy = useActivePolicy();
+    const lastWorkspaceNumber = useLastWorkspaceNumber();
```
ikevin127, line 191: ### 🟠 `src/components/MoneyRequestHeaderSecondaryActions.tsx:191`: `useLastWorkspaceNumber()` puts a whole-`POLICY`-collection regex scan on every expense header render

This component is not lazy. `MoneyRequestHeaderActions.tsx:50` renders it inline in the header for every expense, not on menu open. So the hook runs whenever an expense header renders.

What the hook costs (`src/hooks/useLastWorkspaceNumber.ts`):

```ts
function useLastWorkspaceNumber(email?: string) {
    const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
    const lastWorkspaceNumberSelectorWithEmail = (policies: OnyxCollection<Policy>) => lastWorkspaceNumberSelector(policies, email ?? sessionEmail ?? '');
    const [lastWorkspaceNumber] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {selector: lastWorkspaceNumberSelectorWithEmail});
    return lastWorkspaceNumber;
}
```

and what the selector does (`src/selectors/Policy.ts:330-348`):

```ts
const workspaceRegex = isSMSDomain ? new RegExp(…) : new RegExp(`^(?=.*${escapedName})(?:.*(?:${WORKSPACE_TRANSLATIONS})\\s*(\\d+)?)`, 'i');
const workspaceNumbers = Object.values(policies ?? {})
    .map((policy) => workspaceRegex.exec(policy?.name ?? ''))
    …
```

So this constructs a `RegExp` and executes it against every policy name in the account. Two things make it worse here:

1. `lastWorkspaceNumberSelectorWithEmail` is an inline arrow recreated on every render, so the selector identity is never stable. This same file already shows the right pattern eleven lines below:

    ```tsx
    const filteredPoliciesInfoSelector = useMemo(() => createFilteredPoliciesInfoSelector(currentUserLogin), [currentUserLogin]);
    const [filteredPoliciesInfo] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {selector: filteredPoliciesInfoSelector});
    ```

2. This is now the **third** `COLLECTION.POLICY` subscription in one component, alongside `useActivePolicy()` and `filteredPoliciesInfo`.

The value is consumed in exactly one place, and only matters when the user has zero workspaces:

```tsx
defaultWorkspaceName: generateDefaultWorkspaceName(currentUserLogin ?? '', lastWorkspaceNumber, translate, currentUserDisplayName),
```

`ReportUtils.ts:12176` only reads `defaultWorkspaceName` inside `if (filteredPoliciesCount === 0)`. So for every user who already has a workspace, this whole scan is computed on every header render and then thrown away.

`ChatActionableButtons` calls the hook the same way, but that component only mounts for a track-expense whisper. The expense header is a much hotter surface, and this is the area the INP work has been focused on.

Minimum fix, memoize the selector inside the hook so at least it only re-runs on `POLICY` changes rather than on every render:

```ts
function useLastWorkspaceNumber(email?: string) {
    const [sessionEmail] = useOnyx(ONYXKEYS.SESSION, {selector: emailSelector});
    const resolvedEmail = email ?? sessionEmail ?? '';
    const selector = useMemo(() => (policies: OnyxCollection<Policy>) => lastWorkspaceNumberSelector(policies, resolvedEmail), [resolvedEmail]);
    const [lastWorkspaceNumber] = useOnyx(ONYXKEYS.COLLECTION.POLICY, {selector});
    return lastWorkspaceNumber;
}
```

Better, skip the subscription entirely when it cannot be used, since you already have the count in this component:

```tsx
const hasNoAccessibleWorkspace = (filteredPoliciesInfo?.filteredPoliciesCount ?? 0) === 0;
const lastWorkspaceNumber = useLastWorkspaceNumber(hasNoAccessibleWorkspace ? undefined : SKIP);
```

That needs a hook change, so the memoization fix alone is a reasonable scope for this PR with a follow-up for the rest.

**Bug this causes:** no functional bug, but on a high-traffic account with many workspaces every expense header render pays a regex pass over the full policy collection for a value most users never consume.

The PR checklist item *"I tested this PR with a High Traffic account against the staging or production API"* is unchecked, and this is exactly what it would surface.

tests/unit/ReportSecondaryActionUtilsTest.ts
```diff
@@ -4998,6 +4998,22 @@ describe('getSecondaryTransactionThreadActions', () => {
...
+            expect(getSendToSomeoneResult(true, false)).not.toContain(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER);
+        });
+
+        it('includes SEND_TO_EMPLOYER for a self-DM split expense when the user has a workspace to submit to', () => {
+            expect(getSendToSomeoneResult(true, true)).toContain(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER);
+        });
+
+        it('hides SEND_TO_EMPLOYER on an archived self-DM (no write access)', () => {
+            expect(getSendToSomeoneResult(false, false, true)).not.toContain(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_EMPLOYER);
+        });
```
ikevin127, line 5016: ### 🟡 `tests/unit/ReportSecondaryActionUtilsTest.ts:5002-5016`: The new tests cannot fail on the thing most likely to be wrong

The four new tests are exact mirrors of the four existing `SEND_TO_SOMEONE` tests, asserted through the same helper. That is expected given both options are pushed under one condition, but it means the suite asserts the shared gate rather than checking the two options can diverge. If the fix in the [first finding](https://github.com/Expensify/App/pull/99247/changes#r3837295870) lands, these tests keep passing while the friend row silently keeps appearing for splits.

The test that would earn its place is the negative one:

```ts
it('hides SEND_TO_SOMEONE for a self-DM split expense even when the user has a workspace, since a split has no personal destination', () => {
    expect(getSendToSomeoneResult(true, true)).not.toContain(CONST.REPORT.TRANSACTION_SECONDARY_ACTIONS.SEND_TO_SOMEONE);
});
```

Also, `getSendToSomeoneResult` (line 4962) is now driving employer assertions too, and its inline comment still reads *"mirror that so the write-action guard on SEND_TO_SOMEONE is exercised"*.

Worth renaming to `getSelfDMConvertActionsResult` or similar and updating the comment, otherwise the next person reading `getSendToSomeoneResult(true, true)).toContain(SEND_TO_EMPLOYER)` has to go read the helper to understand what is being set up.

tests/unit/ReportSecondaryActionUtilsTest.ts
```diff
@@ -4998,6 +4998,22 @@ describe('getSecondaryTransactionThreadActions', () => {
...
     });
```
ikevin127, line 5018: ### 🟡 No coverage for anything in `MoneyRequestHeaderSecondaryActions.tsx` or `PopoverMenuSections.ts`

All four new tests sit on the gating util. Untested:

- that `SEND_TO_EMPLOYER`'s `onSelected` passes `submitDestination: EMPLOYER` and a non-empty `defaultWorkspaceName`
- that `SEND_TO_SOMEONE`'s `onSelected` still passes neither, so it keeps defaulting to `FRIEND`
- that both rows land in the top section of `TRANSACTION_MORE_MENU_SECTIONS` with the divider beneath
- that the delegate guard fires on the new row

The params object is the part a future refactor is most likely to break, since it is now shared between two callers and spread with an override. A small assertion on the spread would be cheap:

```ts
it('passes the employer destination and a generated workspace name, and leaves Send to someone on the default friend destination', () => {
    // assert createDraftTransactionAndNavigateToParticipantSelector was called with
    // {submitDestination: 'employer', defaultWorkspaceName: expect.any(String)} for the employer row,
    // and with submitDestination undefined for the friend row
});
```

**Bug this prevents:** if someone later reorders the spread to `{submitDestination: EMPLOYER, ...sendTrackedExpenseParams}`, the override is silently dropped and **Send to my employer** quietly behaves as **Send to someone**. Nothing in the current suite catches that.

src/languages/en.ts
```diff
@@ -1633,6 +1631,7 @@ const translations = {
         enableWallet: 'Enable wallet',
         hold: 'Hold',
         sendToSomeone: 'Send to someone',
+        sendToEmployer: 'Send to my employer',
```
ikevin127: ### 🟡 Copy differs from the Inbox flow it mirrors, and the Design label is unchecked

`en.ts:9741-9742` already has the Inbox strings:

```ts
actionableMentionTrackExpense: {
    submitToFriend: 'Submit to a friend',
    submitToEmployer: 'Submit to my employer',
```

This PR adds `iou.sendToEmployer: 'Send to my employer'`. The PR body itself calls the Inbox action *"Submit to my employer"*, so the same action is `Submit` in one surface and `Send` in another.

`Send` is internally consistent with `Send to someone` in this menu, so this is **probably deliberate** since it matches `sendToSomeone: 'Send to someone'`, but it is new user-facing copy and the checklist item *"I added `Design` label and/or tagged `@Expensify/design`"* is unchecked. Worth getting an explicit ack rather than assuming.

The ten translations themselves all read correctly and are placed in the right `iou` block in each file. No missing locale, no punctuation issues, no placeholder left behind ✅

src/components/MoneyRequestHeaderSecondaryActions.tsx
```diff
@@ -631,25 +659,23 @@ function MoneyRequestHeaderSecondaryActions({reportID, onBackButtonPress}: Money
...
-                    preferredPolicyID,
-                    transaction,
-                    currentUserAccountID: accountID,
-                    currentUserEmail: currentUserLogin ?? '',
-                    currentUserLocalCurrency: localCurrencyCode ?? CONST.CURRENCY.USD,
-                    filteredPoliciesCount: filteredPoliciesInfo?.filteredPoliciesCount ?? 0,
-                    firstPolicyID: filteredPoliciesInfo?.firstPolicyID,
+                    ...sendTrackedExpenseParams,
+                    submitDestination: CONST.IOU.SUBMIT_DESTINATION.EMPLOYER,
+                    defaultWorkspaceName: generateDefaultWorkspaceName(currentUserLogin ?? '', lastWorkspaceNumber, translate, currentUserDisplayName),
```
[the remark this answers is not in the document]
ikevin127: **🔴 Confirmed, this needs fixing before merge!**

The premise holds: `CurrentUserPersonalDetailsProvider.tsx:23-31` sources `email` from `SESSION` but spreads `login` from `PERSONAL_DETAILS_LIST`, so there is a real window where the user is signed in, `email` is set, and `login` is still `undefined`.

In that window `generateDefaultWorkspaceName(currentUserLogin ?? '', ...)` hits the `emailParts?.length !== 2` guard at `Policy.ts:2450` and returns `''`, and `currentUserEmail: currentUserLogin ?? ''` flows through `createDraftWorkspace` into an optimistic policy with `name: ''`, `owner: ''`, `approver: ''` and `employeeList: {'': {...}}` (`Policy.ts:3306+`). That is a malformed workspace that gets sent to the server, not just a cosmetic blank.

The reason this belongs in this PR rather than a follow-up, even though two of the three `currentUserLogin` uses are pre-existing: `ReportUtils.ts:12174` gates the workspace-creation branch on `submitDestination === EMPLOYER`, and `Send to someone` leaves it at the `FRIEND` default. Nothing in this component could reach `createDraftWorkspaceAndNavigateToConfirmationScreen` before. **Adding `Send to my employer` is what makes the path reachable**, so the latent inconsistency becomes a live bug in this diff.

`generateDefaultWorkspaceName` has ten call sites and `MoneyRequestHeaderSecondaryActions.tsx:678` is the only one passing `login`. `DynamicReportDetailsPage.tsx:245,568,575` runs this exact employer flow on `email` with the same four-argument shape, so switching to `email` just matches existing precedent.

**One correction on the report**: the policy-count part is overstated. `getPolicyRole` returns `policy.role` before it ever checks login (`PolicyUtils.ts:632-634`), so for policies where the server populated `role` the count is still correct. The count only collapses for policies falling through to the `employeeList` lookup. The malformed-data outcome is the unconditional one and is what makes this worth blocking on.

`useLastWorkspaceNumber()` needs no change, it already falls back to the session email internally.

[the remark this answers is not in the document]
shawnborton: Confirmed and fixed in 02c61ccf2b0. The employer flow now reads `email` from `useCurrentUserPersonalDetails()` (which always comes from `SESSION`) instead of `login`, so `generateDefaultWorkspaceName` and `currentUserEmail` no longer collapse to `''` during the window where `login` is still undefined. I also switched `createFilteredPoliciesInfoSelector` to `email` to match the Inbox/report-details precedent. Left `useLastWorkspaceNumber()` untouched since it already falls back to the session email internally. Thanks for the `getPolicyRole` correction on the count — agreed the malformed-workspace outcome was the real blocker.

src/libs/ReportSecondaryActionUtils.ts
shawnborton: Fixed in 02c61ccf2b0. You're right — I could not confirm the "participant selector routes the split into it" claim either, and the FRIEND default just opens the people picker. Split the gate into two conditions: `SEND_TO_SOMEONE` now pushes only when `!isSelfDMExpenseSplit` (friend hidden for all splits, matching `ChatActionableButtons`), while `SEND_TO_EMPLOYER` keeps `!isSelfDMExpenseSplit || hasWorkspaceToSubmitTo`. Rewrote the comment to state the real reasons and dropped the incorrect claim.

src/components/MoneyRequestHeaderSecondaryActions.tsx
shawnborton: Fixed in 02c61ccf2b0 with the memoization approach — the selector is now `useMemo`'d on `resolvedEmail` inside `useLastWorkspaceNumber`, so the POLICY-collection regex scan only re-runs when the email changes rather than on every render of every consumer (including the expense header). Kept the "skip the subscription entirely when the user already has a workspace" variant out of scope since it needs a hook API change — happy to do that as a follow-up.

tests/unit/ReportSecondaryActionUtilsTest.ts
shawnborton: Fixed in 02c61ccf2b0. Renamed the helper to `getSelfDMConvertActionsResult` and updated its comment (it now drives both options, so the SEND_TO_SOMEONE-specific wording was misleading). Replaced the split+workspace "includes SEND_TO_SOMEONE" case with the negative assertion you suggested: `hides SEND_TO_SOMEONE for a self-DM split expense even when the user has a workspace, since a split has no personal destination`. That test now fails if the friend row ever starts sharing the employer gate again.

src/languages/en.ts
shawnborton: Changed to **Submit to my employer** in 02c61ccf2b0. Renamed the key `iou.sendToEmployer` -> `iou.submitToEmployer` and reused each locale's already-established `actionableMentionTrackExpense.submitToEmployer` copy, so the menu now matches the Inbox wording exactly rather than introducing new copy. (Shawn confirmed this direction over keeping "Send to my employer".)

tests/unit/ReportSecondaryActionUtilsTest.ts
shawnborton: Partially addressed in 02c61ccf2b0, with a follow-up proposed for the rest.

Added a `TRANSACTION_MORE_MENU_SECTIONS` test in `PopoverMenuSectionsTest.ts` asserting both rows sort into the top section together, in order, with no divider between them and a divider before the next section.

For the `onSelected` params + delegate-guard assertions: there is no existing render harness for `MoneyRequestHeaderSecondaryActions`, and standing one up (mocking ~20 Onyx keys + navigation + opening the dropdown) is a disproportionate lift for this PR. I'd like to handle that as a follow-up so the params-spread regression you flagged gets a real render test. Does that work for you? If so I'll open a follow-up issue and link it here.

ikevin127:
## Reviewer Checklist

    - [ ] Android: HybridApp
    - [ ] Android: mWeb Chrome
    - [ ] iOS: HybridApp
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari

### Screenshots/Videos

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

ikevin127:
## Re-review of #99247 at `02c61ccf`

@shawnborton All six items from the previous round are addressed. One thing in the PR body that will bite QA 👇 

### ⚠️ The PR body needs updating before QA sees it

This is the one thing I would not merge without. The `### Tests` and `### Offline tests` steps still say **Send to my employer** in four places, but the shipped copy is now **Submit to my employer**. QA follows those steps literally, so step 4 ("Verify a **Send to my employer** option appears") will be reported as a failure against a working build.

Also outstanding, and what the failing `checklist` CI job is flagging:

- all PR Author Checklist checkboxes are unchecked at the moment which is failing PR Author Checklist CI

shawnborton:
@MelvinBot please implement the feedback [here](https://github.com/Expensify/App/pull/99247#issuecomment-5401677275)

ikevin127 (approved):
**🟢 LGTM** - PR tests well, all review-comments were addressed

francoisl (approved):

Issue #99410, opened 2026-08-25
Add category tax defaults to Expense defaults to reach parity with Classic


JmillsExpensify:
### Background

R1 (#92739) centralized workspace merchant rules into the `Expense defaults` tab, and R2 (#92741) added the 12 locked `Merchant types` defaults. Every rule in that tab matches on a merchant or a merchant type.

Classic lets an admin set a default tax rate on a category, and we have no equivalent on the new platform. Any workspace using that setting today loses it when they move over, which makes this a backward-compatibility gap and a blocker on removing the beta.

This issue adds `Category` as a second condition for `Apply expense defaults`, so an admin can set a default tax rate for a category. `Expense defaults` has not yet been migrated to the new rules system, so the rule carries real constraints: one condition per rule, tax as the only available default, and no retroactive application. We express all three with the lock affordance R1 and R2 already established.

### Scope

- Add `Category` as a second condition to the `Apply expense defaults` branch of `New rule`, alongside the existing `Merchant` condition.
- Allow exactly one condition per rule. Setting one locks the other, because `Expense defaults` has not been migrated to the new rules system.
- When the condition is `Category`, allow only `Tax` in the defaults to apply and lock every other field.
- Lock `Apply to existing unsubmitted expenses` for category rules, since a category tax default only applies to expenses created after the rule is saved.
- Show category rules in a new `Categories` subsection of the `Expense defaults` table.
- No change to the `New rule` option list or its copy. This extends the existing `Apply expense defaults` branch; it does not add a fifth rule type.

### Product and design details

<img width="2800" height="2000" alt="Image" src="https://github.com/user-attachments/assets/5631e076-a0bc-4eb4-a3eb-5faa77da65f3" />

<img width="2410" height="1624" alt="Image" src="https://github.com/user-attachments/assets/1e96a2fa-267f-48af-af5d-3b386c9b3bd9" />

#### Applicability

The `Category` condition is available to workspace admins when taxes are enabled on the workspace. Taxes are what the rule sets, so with taxes off there is nothing to configure. In that case the `Category` condition row renders locked, and the `Merchant` condition continues to work as it does today.

#### Condition: one of `Merchant` or `Category`

The `If any expense matches:` section has two rows, `Merchant` and `Category`, in that order. Each shows its selected value beneath its label, and nothing beneath it when unset.

Setting a value in one row locks the other:

- The locked row replaces its chevron with the lock icon and no longer opens a picker.
[...]

PR #99420, opened 2026-08-25
Add category tax defaults to Expense defaults

Krishna2323:
### Fixed Issues
$ https://github.com/Expensify/App/issues/99410
PROPOSAL:

### Create — choose the type

1. Rules → `+ Rule` → `Apply expense defaults`: verify the chooser shows `Apply expense defaults` / `What should this rule match on?` with `Merchant` and `Category` cards.
2. With taxes off/no tax rates: verify `Category` is locked and tapping it shows `Turn on taxes first`.
3. Deep-link to `workspaces/<id>/rules/merchant-rules/new`: verify it redirects to the chooser.

### Create — category rule

1. Chooser → `Category`: verify required `Category` condition and only `Tax` under defaults. No merchant/default rows or `Preview matches`.
2. Verify apply-to-existing is locked and tapping it shows `Category rules apply going forward`.
3. Open `Tax`: verify no `None` option and workspace default rate isn't listed.
4. Select 3 categories + 1 tax → Save: verify 3 rows are created with the correct category/tax.
5. Search categories: verify filtering works, selected categories stay pinned, and select-all only selects filtered results.
6. With taxes off → `Tax`: verify `Taxes are not enabled` empty state with `Enable taxes`, and enabling returns to a working picker.

### Create — merchant rule

1. Chooser → `Merchant`: verify required `Merchant`, all defaults available, and `Preview matches` + apply-to-existing are usable.
2. Save a merchant rule with a category default: verify it appears under `Merchants`.

### Reset

1. Set a condition on an unsaved rule → `Reset`: verify condition/defaults clear while staying scoped to the same type.
2. With nothing selected or on a saved rule: verify `Reset` is absent.

### Edit

1. Open a `Categories` row: verify category + tax are pre-filled and are the only rows.
2. Change category (verify **radios**, not checkboxes) → Save: verify old row is removed and new category keeps the rate, with no not-found flash.
3. Change only tax → Save: verify the row updates.
4. Deep-link to `workspaces/<id>/rules/category-tax-rules/edit/<category>`: verify correct values load.
5. As an **auditor**, open a `Categories` and `Merchants` row: verify both are read-only — no chevrons, Save, or Delete, and no not-found page.

### Delete

1. Delete a category rule: verify it disappears.
2. Bulk delete 2 merchant + 1 category rule: verify all disappear.
3. Verify `Merchant types` rows are locked/non-selectable.
4. Delete a category rule → create an expense in that category: verify tax falls back to the workspace default.
5. Delete the category or tax rate used by a category rule: verify the rule disappears with it.
6 Delete a category that a merchant rule matches on: verify the merchant rule stays listed and still opens

### Sections / sorting

Unchanged — all four existing steps still apply.

### Validation

1. Merchant rule, save with nothing: verify `Enter a merchant, and apply at least one default`.
2. Merchant rule, default without merchant: verify `Please enter a merchant`.
3. Category rule, save without category: verify `Please select a category`.
4. Category rule, category without tax: verify `Please select a tax rate`.

### Offline tests
1. Delete merchant rule offline: verify row greys out and stays. Online: verify it disappears immediately.
2. Delete category rule: verify it disappears both offline and online — expected since category tax rules can't carry a pending state.
3. Move a category rule's category offline with a failing connection: verify neither rule disappears and the local array reverts to its pre-move state.

### QA Steps
- Same as tests

### PR Author Checklist

    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] MacOS: Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/5660f02a-7574-406c-b60e-2243bfec1305

</details>

<details>

https://github.com/user-attachments/assets/57a1eda2-a041-4ae6-a367-9cf53db342fd

</details>

<details>

https://github.com/user-attachments/assets/41a9ab00-bea3-4c35-9061-06f7b831ebde

</details>

<details>

https://github.com/user-attachments/assets/b001d660-ad28-4d4b-9a93-fc0ac41ce281

</details>

<details>

https://github.com/user-attachments/assets/8af6c192-3dd0-443c-8299-f66251ba9ab7

</details>


src/pages/workspace/rules/MerchantRules/EditCategoryTaxRulePage.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import MerchantRulePageBase from './MerchantRulePageBase';
+
+type EditCategoryTaxRulePageProps = PlatformStackScreenProps<SettingsNavigatorParamList, typeof SCREENS.WORKSPACE.RULES_CATEGORY_TAX_EDIT>;
+
+function EditCategoryTaxRulePage({route}: EditCategoryTaxRulePageProps) {
+    return (
+        <MerchantRulePageBase
+            policyID={route.params.policyID}
+            editCategoryTaxRuleFor={route.params.categoryName}
```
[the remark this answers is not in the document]
Krishna2323: fixed. `RULES_MERCHANT_TAX` now takes an optional `categoryName` and routes back to `RULES_CATEGORY_TAX_EDIT`.

src/libs/actions/Policy/Category.ts
```diff
@@ -1901,6 +1938,280 @@ function setPolicyCategoryTax(policy: OnyxEntry<Policy>, categoryName: string, t
...
+                {
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {
+                        rules: {
+                            // Clear the pending state for everything settled so far. Categories still in flight keep
+                            // theirs, so an early success can't make a later one look finished.
+                            expenseRules: pendingExpenseRules.map((rule) =>
+                                categoryNames.some((name) => settledSoFar.has(name) && matchesCategoryTaxRule(rule, name)) ? {...rule, pendingAction: null, errors: null} : rule,
+                            ),
```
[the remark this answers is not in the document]
Krishna2323: Valid, and not fixable with precomputed snapshots since they can't know which sibling fails. Removed `successData` entirely; `failureData` now rolls back only its own category.

src/libs/actions/Policy/Category.ts
```diff
@@ -1901,6 +1938,280 @@ function setPolicyCategoryTax(policy: OnyxEntry<Policy>, categoryName: string, t
...
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {rules: {expenseRules: markedExpenseRules}},
+                },
+            ],
+            successData: [
+                {
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {rules: {expenseRules: buildStage(removedSoFar)}},
```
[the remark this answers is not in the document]
Krishna2323: Same root cause as above, fixed the same way. No `successData`, and `failureData` restores only its own category.

JmillsExpensify:
> > Let's also update the list to this exact order:
> 
> The order of the list currently on prod was chosen because that's how the timeline of an expense goes. From spend -> submission. Why are we updating this order and what's the logic for what you propose?

Gotcha. Thanks, I had forgotten the history. I'm not passionate about the order. Interestingly though, the reason that I suggested the above is that my thought was, well defaults come first and that's also the most common rule that customers are asking about in research. 

No right answer here though, so fine keeping the order. We definitely need to update the language though so we have a parallel structure with verb/noun agreement.

JmillsExpensify:
Then as for the copy, I'm good with whatever you prefer. I don't think we should hold up the PR though, so depending on when we have C+ review we can address it in this PR or another one.

JmillsExpensify:
@situchan mind testing this PR first thing today?

situchan:
## Reviewer Checklist

    - [ ] Android: HybridApp
    - [ ] Android: mWeb Chrome
    - [ ] iOS: HybridApp
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

Krishna2323:
@dubielzyk-expensify @JmillsExpensify 6 strings are new and need sign-off:

String | Shown when
-- | --
Please select a merchant or category | Save with no condition, but at least one default is set
Select a merchant or category, and apply at least one default | Save with nothing filled in
Please select a category | Merchant-type rule saved with no category
Please select a tax rate | Category rule saved with no tax rate
Clear the other defaults? | Tapping Category with merchant-only defaults set
Clear fields | Confirm button on the clear-fields dialog

The first 4 are validation errors, following the existing pattern. The last 2 are for the new clear-fields dialog. Everything else is copied verbatim from the issue.

Krishna2323:
@JmillsExpensify @dubielzyk-expensify Two small guards I added while testing, both for the same underlying issue: category tax defaults are stored differently from merchant rules, so some UI states can ask for something the storage can't do.

**1. Hide the workspace default tax rate for category rules.**
- Deleting a category tax rule means writing the workspace default rate. So selecting that rate would look like saving a rule but actually delete it. It's now hidden for category rules and blocked on save too, since the shared merchant rule draft could otherwise set it as a merchant default and then add a category condition. 

**2. Reset is only available when creating a rule.**
- Resetting a saved rule clears the form but keeps its `ruleID`. Adding a category condition and saving could then create a category tax rule while leaving the original merchant rule in place. Since merchant and category rules are stored separately, changing the rule type isn't a normal edit. 

Editing still works normally. You can change the merchant or edit individual defaults, and Delete is still available. Only the clear-all Reset is removed from edit mode, which was the path to the broken state.

Krishna2323:
@codex review

JmillsExpensify:
> @dubielzyk-expensify @JmillsExpensify 6 strings are new and need sign-off:
> 
> String	Shown when
> Please select a merchant or category	Save with no condition, but at least one default is set
> Select a merchant or category, and apply at least one default	Save with nothing filled in
> Please select a category	Merchant-type rule saved with no category
> Please select a tax rate	Category rule saved with no tax rate
> Clear the other defaults?	Tapping Category with merchant-only defaults set
> Clear fields	Confirm button on the clear-fields dialog
> The first 4 are validation errors, following the existing pattern. The last 2 are for the new clear-fields dialog. Everything else is copied verbatim from the issue.

All these looks great to me.

JmillsExpensify:
> @JmillsExpensify @dubielzyk-expensify Two small guards I added while testing, both for the same underlying issue: category tax defaults are stored differently from merchant rules, so some UI states can ask for something the storage can't do.
> 
> **1. Hide the workspace default tax rate for category rules.**
> 
> * Deleting a category tax rule means writing the workspace default rate. So selecting that rate would look like saving a rule but actually delete it. It's now hidden for category rules and blocked on save too, since the shared merchant rule draft could otherwise set it as a merchant default and then add a category condition.
> 
> **2. Reset is only available when creating a rule.**
> 
> * Resetting a saved rule clears the form but keeps its `ruleID`. Adding a category condition and saving could then create a category tax rule while leaving the original merchant rule in place. Since merchant and category rules are stored separately, changing the rule type isn't a normal edit.
> 
> Editing still works normally. You can change the merchant or edit individual defaults, and Delete is still available. Only the clear-all Reset is removed from edit mode, which was the path to the broken state.

Both make sense. The first is a little tricky, though I agree with you that we should start here and see how it goes with users.

JS00001:
@Krishna2323 some checks are failing

Krishna2323:
> some checks are failing

@JS00001 Was waiting for confirmation on the new string. Please run the translations script, the errors are due to missing translations.

JS00001:
Ahh sorry, I thought I ran them, running!

dubielzyk-expensify:
Yeah that all makes sense to me 👍

situchan:
Bug: wrong modal when click locked category

There's no merchant rule set but the modal says "This rule already matches on a **merchant**. **Reset** the rule to match on a category instead."

2nd copy is also wrong. There's no reset button but it says "Reset"

https://github.com/user-attachments/assets/1440fb9e-f4c3-401b-b926-7a63d5edb3df

situchan:
Bug: wrong modal on the locked category row when taxes are off

I think the reason should be "This rule already matches on a merchant."

https://github.com/user-attachments/assets/64c0fa0c-d21f-4ce9-86c6-ff72b9ff3850

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
```diff
@@ -355,10 +355,20 @@ function PolicyRulesPageRevamp({route}: PolicyRulesPageRevampProps) {
...
+        <>
+            {rulesTabSelector}
+            {separateLineButtons}
+        </>
+    );
     const sharedTableTabProps = {
         policyID,
         canWriteRules,
-        headerComponent: rulesTabSelector,
+        headerComponent: listHeaderComponent,
```
situchan: On mobile, header buttons are no longer sticked to the top so user doesn't know how many selected without scrolling up to the top

https://github.com/user-attachments/assets/84d28ee5-3ce2-4e35-90d3-87174bfac42b

src/pages/workspace/rules/MerchantRules/AddCategoryToMatchPage.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+    useFocusEffect(() => {
+        fetchPolicyCategories();
+    });
+
+    // Only spin while a fetch can actually resolve. Offline there is nothing to wait for, so fall through to the
+    // list instead of a spinner that never goes away.
+    const arePolicyCategoriesLoading = areCategoriesEnabled && policyCategories === undefined && !isOffline;
+
+    const categoryItems: CategoryListItem[] = Object.values(policyCategories ?? {})
```
situchan: This code block is duplicated across 3 sites:
- RequireFieldsRuleCategoryPageBase.tsx
- FlagForReviewRuleCategoryPageBase.tsx
- And here

How about inroducing `usePolicyCategoryPickerItems({policyID, isEligible})` hook returning `{items, areCategoriesEnabled, isLoading}`?

src/libs/actions/Policy/Category.ts
```diff
@@ -1833,72 +1832,161 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
     API.write(WRITE_COMMANDS.SET_POLICY_CATEGORY_APPROVER, parameters, onyxData);
 }
 
-function setPolicyCategoryTax(policy: OnyxEntry<Policy>, categoryName: string, taxID: string) {
-    if (!policy?.id) {
+/** Whether an expense rule is the tax default for this category. `applyWhen` is typed as required but comes from the
+ * backend, and Classic writes this same array, so a rule without one must not take the whole list down. */
+function matchesCategoryTaxRule(rule: ExpenseRule, categoryName: string): boolean {
+    return !!rule.applyWhen?.some((when) => when.value === categoryName);
+}
```
situchan, line 1904: There are 3 competing definitions of "this rule is category X's tax default"

- https://github.com/Expensify/App/pull/99420/changes#diff-555d7a50adbb70ce577834409d86a957820aed37b37b92b61b64d45d6b6fffc8R31-R33
- https://github.com/Expensify/App/blob/de2eda378093175c28d85dd0d4a0ae947d23d20c/src/libs/CategoryUtils.ts#L92-L97
- and here

`getCategoryExpenseRule` already did this match strictly (`condition === MATCHES && field === CATEGORY && value === name`). 
The PR adds `getRuleCategoryName` (strict, reversed direction) and `matchesCategoryTaxRule` (`when.value === categoryName` with no field check).

so the actions and the table disagree about what they are operating on.


`matchesCategoryTaxRule` can be `(rule, name) => getRuleCategoryName(rule) === name`.

src/libs/CategoryTaxRulesUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+ * The `Name (Value)` label used for a tax rate everywhere in the rules UI. A tax rate deleted from the workspace
+ * leaves a rule pointing at nothing, so fall back to the raw ID rather than rendering an empty label.
+ */
+function getTaxRateDisplayName(policy: Policy | undefined, taxID: string | undefined): string {
+    if (!taxID) {
+        return '';
+    }
+    const taxRate = policy?.taxRates?.taxes?.[taxID];
+    return taxRate ? `${taxRate.name} (${taxRate.value})` : taxID;
+}
```
situchan: The `Name (Value)` tax label is formatted in three places:

- MerchantRulePageBase
- MerchantTypeRulesUtils
- and here

`getTaxRateDisplayName`: this seems the only one with a fallback for a deleted tax rate, which is the right behaviour. So export this and use it in all other places.

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
Krishna2323: fixed.

Krishna2323:
> Bug: wrong modal when click locked category
> 
> There's no merchant rule set but the modal says "This rule already matches on a **merchant**. **Reset** the rule to match on a category instead."
> 
> 2nd copy is also wrong. There's no reset button but it says "Reset"

I think the category should be editable, so I've made it editable. The optimistic state doesn't work perfectly yet, but we can address that in a follow-up.

> Bug: wrong modal on the locked category row when taxes are off
> 
> I think the reason should be "This rule already matches on a merchant."


Updated, it now says: `This rule matches on a category/merchant. Create a new rule to match on a merchant/category.`

src/pages/workspace/rules/MerchantRules/AddCategoryToMatchPage.tsx
Krishna2323: Added `usePolicyCategoryPickerCategories({policyID, isEligible}).` Kept items per-page since single vs multi select need different shapes.

src/libs/actions/Policy/Category.ts
Krishna2323: Good catch, fixed. `getCategoryExpenseRule` left alone as pre-existing.

src/libs/CategoryTaxRulesUtils.ts
Krishna2323: Exported and used in both. Added an optional saved-label fallback.

Krishna2323:
@dubielzyk-expensify @JmillsExpensify 2 more strings need sign-off:
String | Shown when
-- | --
This rule matches on a merchant. Create a new rule to match on a category. | Tapping the locked Category row on an already-saved rule
This rule matches on a category. Create a new rule to match on a merchant. | Tapping the locked Merchant row on an already-saved rule

These replace the existing copy on saved rules only, since <strong>Reset</strong> is now only shown while creating a rule. Unsaved rules keep the original copy.

JmillsExpensify:
> @dubielzyk-expensify @JmillsExpensify 2 more strings need sign-off:
> 
> String	Shown when
> This rule matches on a merchant. Create a new rule to match on a category.	Tapping the locked Category row on an already-saved rule
> This rule matches on a category. Create a new rule to match on a merchant.	Tapping the locked Merchant row on an already-saved rule
> These replace the existing copy on saved rules only, since **Reset** is now only shown while creating a rule. Unsaved rules keep the original copy.

LGTM

Krishna2323:
Hey @jason @dubielzyk-expensify, I was thinking we could simplify the Expense defaults flow by asking users to choose the rule type first. IMO this makes the UX much clearer and avoids a lot of confusing locked states. I've already applied the changes for testing, here's the recording:


https://github.com/user-attachments/assets/d88907e7-0840-4439-a2a8-1f6621ea002b



**How it works:**

**New rule → Apply expense defaults → What should this rule match on?**

- **Merchant** — update fields for a specific merchant
- **Category** — set a default tax rate for a category

The editor then only shows what applies to that rule type:

|  | Merchant rule | Category rule |
|---|---|---|
| Merchant condition | Shown | Hidden |
| Category condition | Hidden | Shown |
| Defaults | All | Tax only |

I think this removes a lot of the complexity we currently have around locked rows and explainer modals for invalid combinations.

JmillsExpensify:
I think it's an interesting idea, and to your point, better UX until we add more rules in due time. @jon thoughts on this direction?

situchan:
> > Bug: wrong modal when click locked category
> > There's no merchant rule set but the modal says "This rule already matches on a **merchant**. **Reset** the rule to match on a category instead."
> > 2nd copy is also wrong. There's no reset button but it says "Reset"
> 
> I think the category should be editable, so I've made it editable. The optimistic state doesn't work perfectly yet, but we can address that in a follow-up.

"I've made it editable" - This introduced another bugs:

- when select another category, old category is no longer displayed in category selection list
- 2 rows appear until refresh from backend response


https://github.com/user-attachments/assets/2b54eace-f222-4c99-9aa1-af8198ea552a

src/pages/workspace/rules/MerchantRules/AddCategoryToMatchPage.tsx
```diff
@@ -0,0 +1,199 @@
...
+            />
+            {arePolicyCategoriesLoading ? (
+                <View style={[styles.flex1, styles.justifyContentCenter, styles.alignItemsCenter]}>
+                    <ActivityIndicator size={CONST.ACTIVITY_INDICATOR_SIZE.LARGE} />
+                </View>
+            ) : (
+                <SelectionList
+                    canSelectMultiple={!isEditingCategoryTaxRule}
+                    shouldUpdateFocusedIndex
+                    ListItem={MultiSelectListItem}
```
situchan: In edit mode, single-select mode renders checkboxes instead of radioboxes.
Need design confirmation

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
```diff
@@ -384,8 +387,8 @@ function PolicyRulesPageRevamp({route}: PolicyRulesPageRevampProps) {
                 headerContent={!shouldDisplayButtonsInSeparateLine && headerButtons}
             >
                 <View style={[styles.flex1, styles.w100, styles.mnh0]}>
+                    {separateLineButtons}
                     {!isTableTab && !isAgentsTab && rulesTabSelector}
```
situchan: Bug?: In **General** tab, the header buttons are now scrollable inside list. This doesn't happen on other tabs.

And the scroll behaviors are inconsistent in empty state pages:
The tabs are scrollable in **Expense defaults**, **Agent** empty pages but not in **Card restrictions**, **Field requirements**, **Flag for review** empty pages

https://github.com/user-attachments/assets/15700a40-786f-45c5-a321-746f6252e3fa

situchan:
There's no handling of existing rules for disabled/deleted categories/taxes. They're now stale


https://github.com/user-attachments/assets/bb5d7b25-1fb5-4fa2-843d-856c0d745c82

----

Same issue when entire Categories / Taxes feature is disabled

https://github.com/user-attachments/assets/f60c9fba-6c25-468c-941a-03e5ef8fea38

dannymcclain:
>I was thinking we could simplify the Expense defaults flow by asking users to choose the rule type first. IMO this makes the UX much clearer and avoids a lot of confusing locked states.
>
>>I think it's an interesting idea, and to your point, better UX until we add more rules in due time.

I'm down with this. I love removing the ability for users to get into weird sticky spots, so this seems like a pretty good update in that regard (less confusing locked states).

For the merchant illustration here:
CleanShot 2026-08-31 at 11 54 40@2x

Can we update to the [taco truck illustration](https://github.com/Expensify/App/blob/main/assets/images/simple-illustrations/simple-illustration__foodtruck-tacos.svg)? `assets/images/simple-illustrations/simple-illustration__foodtruck-tacos.svg`

That's the same illo we use in the empty state of allowed/blocked merchants here:
CleanShot 2026-08-31 at 11 56 08@2x

cc @JmillsExpensify @Expensify/design

JmillsExpensify:
Sounds great to me! @Krishna2323 back to you.

Krishna2323:
@JmillsExpensify could you please advise on how we should handle [these cases](https://github.com/Expensify/App/pull/99420#issuecomment-5480313505)?

Krishna2323:
> "I've made it editable" - This introduced another bugs:
> 
> * when select another category, old category is no longer displayed in category selection list
> * 2 rows appear until refresh from backend response


Fixed.

> In edit mode, single-select mode renders checkboxes instead of radioboxes.

Updated.


> Bug?: In General tab, the header buttons are now scrollable inside list. This doesn't happen on other tabs.
> 
> And the scroll behaviors are inconsistent in empty state pages:
> The tabs are scrollable in Expense defaults, Agent empty pages but not in Card restrictions, Field requirements, Flag for review empty pages

I’ve made the General tab’s tab selector scrollable like the other tabs, and I think we can leave the empty-page state as it is.

JmillsExpensify:
> There's no handling of existing rules for disabled/deleted categories/taxes. They're now stale

For disabled, I think that case is fine. That is, the rule still exists and can fire any time the category is re-enabled. So I think we should keep it as is. One question though: In the category list for rule creation, are we showing all categories or only enabled ones? It'd be good to align on that. 

The deleted category case is more tricky. I would expect that if a category is deleted, the underlying rule for it is also deleted. After all, in that case, the category no long exists on the workspace and you couldn't select said category when trying to create/edit a rule.

JmillsExpensify:
> Same issue when entire Categories / Taxes feature is disabled

Ugh that pattern continues to irk me. That said, I think that it's similar to the disabled category case. So my expectation is that we'd faithfully show the rule, because you'd just turned off the ability to see the categories/page. We don't actually delete any underlying categories or tags in this case. Right?

JS00001:
> The deleted category case is more tricky. I would expect that if a category is deleted, the underlying rule for it is also deleted. 

I agree with this. I also agree that we should still show the rule if the category/feature is only disabled

Krishna2323:
@JmillsExpensify @JS00001 The picker only shows **enabled categories** and hides ones pending deletion while online. So a disabled category can't be picked for a new rule, but its existing rule stays listed.

Deleted categories/taxes are already handled by the backend: deleting a category or tax rate used by a category rule deletes the rule, and deleting a category used by a merchant rule deletes the whole merchant rule. Disabling anything — either a single category or the Categories/Taxes feature, keeps the rule as-is, so it works again once re-enabled. Confirmed nothing gets deleted underneath in that case.

The only gap was the optimistic state: rows didn't grey out while a delete was in flight or offline. I fixed this by having the rule row borrow the pending state of the category/tax it depends on. No new optimistic data is needed, so there's nothing to roll back.

One more thing I noticed while testing: with Taxes disabled, a category rule showed an empty **"Then apply the following defaults"** section, since tax is the only default available for category rules. I've added a disabled-feature empty state, matching the existing Categories one, with an **Enable taxes** CTA.

New copy that needs sign-off:

* **Title:** Taxes are not enabled
* **Subtitle:** Enable taxes to document and reclaim eligible taxes.
* **Button:** Enable taxes

The title mirrors **"Categories are not enabled"** and the subtitle is borrowed from the existing More features taxes copy, so hopefully this is straightforward. Just flagging it so we can include it in the translation run.



https://github.com/user-attachments/assets/bb9b8fc0-6e44-4501-94ba-fd8daaf207c0

JmillsExpensify:
> The picker only shows enabled categories and hides ones pending deletion while online. So a disabled category can't be picked for a new rule, but its existing rule stays listed.

> Deleted categories/taxes are already handled by the backend: deleting a category or tax rate used by a category rule deletes the rule, and deleting a category used by a merchant rule deletes the whole merchant rule. Disabling anything — either a single category or the Categories/Taxes feature, keeps the rule as-is, so it works again once re-enabled. Confirmed nothing gets deleted underneath in that case.

Perfect, that all sounds great. 

> The only gap was the optimistic state: rows didn't grey out while a delete was in flight or offline. I fixed this by having the rule row borrow the pending state of the category/tax it depends on. No new optimistic data is needed, so there's nothing to roll back.

Great catch. That all looks good in the video. 

> One more thing I noticed while testing: with Taxes disabled, a category rule showed an empty "Then apply the following defaults" section, since tax is the only default available for category rules. I've added a disabled-feature empty state, matching the existing Categories one, with an Enable taxes CTA.

All that sounds great to me, including the copy.

Krishna2323:
@JS00001 please run the translations script when you get a chance. Thanks!

JS00001:
ran!

src/languages/en.ts
```diff
@@ -8230,6 +8230,22 @@ const translations = {
                 expenseDefaultsSubtitle: 'Update fields without submitter doing anything',
                 ifAnyExpenseMatches: 'If any expense matches:',
                 thenApplyFollowingDefaults: 'Then apply the following defaults:',
+                confirmErrorCategory: 'Please select a category',
+                confirmErrorCategoryTax: 'Please select a tax rate',
+                confirmErrorCondition: 'Please select a merchant or category',
+                confirmErrorConditionAndDefault: 'Select a merchant or category, and apply at least one default',
```
situchan: Image

`or category` should be removed

situchan:
Bug: Merchant rule disappears and then reappears after deleting category

Repro step:
1. Workspace → Rules → Expense defaults → + Rule → Apply expense defaults → Merchant. Set Merchant to match Uber, then under the defaults set Category to Travel and Description to ride. Save.
2. Confirm the row Merchant is "Uber" / Update category to "Travel", update description to "ride" under Merchants.
3. Workspace → Categories → select Travel → Delete, then immediately return to Rules → Expense defaults.
4. The row disappears entirely while the delete is in flight (and greys out instead if you do the same offline). Once the delete lands it reappears, now pointing at a category that no longer exists.



https://github.com/user-attachments/assets/fad042d9-c7b6-48ab-b346-3e19c26d5110

situchan:
Bug: not found page briefly shows when update category and save


https://github.com/user-attachments/assets/2f951504-ab65-488b-bb3e-9121ac61eda3

situchan:
Bug: None shows instead of default tax rate. This doesn't happen on category setting


https://github.com/user-attachments/assets/228b9897-4de7-4958-a6c5-902b95ae2126

Krishna2323:
> Bug: Merchant rule disappears and then reappears after deleting category

Fixed, the BE does not delete the merchant rule when category or tax is deleted, I was [wrong here](https://github.com/Expensify/App/pull/99420#issuecomment-5488965338).

> Bug: not found page briefly shows when update category and save

Fixed

> Bug: None shows instead of default tax rate. This doesn't happen on category setting

This is expected since the default rate is used to delete the rule.

situchan:
Bug: Merchant rule is not deleted immediately, while Category rule is


https://github.com/user-attachments/assets/c3bf08ed-42be-4f50-a16c-72e4a5a69126

situchan:
Reversed Bug: Category rule is immediately deleted when offline while Merchant rule is optimistically deleted


https://github.com/user-attachments/assets/9e34fa14-bcaf-464a-ae07-9cb212926964

src/languages/en.ts
```diff
@@ -8230,6 +8230,22 @@ const translations = {
                 expenseDefaultsSubtitle: 'Update fields without submitter doing anything',
                 ifAnyExpenseMatches: 'If any expense matches:',
                 thenApplyFollowingDefaults: 'Then apply the following defaults:',
+                confirmErrorCategory: 'Please select a category',
+                confirmErrorCategoryTax: 'Please select a tax rate',
+                confirmErrorCondition: 'Please enter a merchant',
+                confirmErrorConditionAndDefault: 'Enter a merchant, and apply at least one default',
```
situchan, line 8263: Make sure to update other languages as well

situchan:
Bug: not found show pages when click on category rule as auditor. Merchant rule seems fine as read-only mode


https://github.com/user-attachments/assets/4849455a-a840-42d2-b7ed-763b30e3c1bd

situchan:
Bug: Created merchant rule disappears after refresh


https://github.com/user-attachments/assets/dcc9902d-dfd3-416a-80aa-f0e505c526c5

situchan:
Bug: category rule is not grayed out when optimistically added offline


https://github.com/user-attachments/assets/06bf0522-93ad-4148-a58d-a0e31dd891e1

situchan:
Bug: merchant rule is not optimistically deleted (no ~cross out~) when the linked tax rate is deleted offline


https://github.com/user-attachments/assets/a34465ca-be55-413b-b034-d63ee55f7112

situchan:
Review complete. Let's make sure all offline test cases work perfect

Krishna2323:
> Reversed Bug: Category rule is immediately deleted when offline while Merchant rule is optimistically deleted

Category rules live in a plain array, so Onyx can't mark a single row as deleting. Clearing that flag requires `successData`, which can replay stale arrays on reconnect. So we drop the row instead. Known tradeoff.


cc: @JS00001

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1889,207 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
-            applyWhen: [
+            ],
+            failureData: [
                 {
-                    condition: CONST.POLICY.RULE_CONDITIONS.MATCHES,
-                    field: CONST.POLICY.FIELDS.CATEGORY,
-                    value: categoryName,
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {rules: {expenseRules: failureExpenseRules}},
```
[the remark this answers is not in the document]
Krishna2323: Fixed. All writes in a save now roll back to the pre-save array, so arrival order can't make a rejected category look saved.

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1889,207 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
-                        expenseRules,
+            ],
+            failureData: [
+                {
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {
+                        rules: {
+                            // Put back only this category, so one rejection doesn't restore rows that were deleted fine.
+                            expenseRules: [...optimisticExpenseRules, ...expenseRules.filter((rule) => matchesCategoryTaxRule(rule, categoryName))],
```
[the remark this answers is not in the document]
Krishna2323: Fixed the same way. One shared baseline per delete, since Onyx replaces the array wholesale and can't patch one rule.

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1889,207 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
+                {
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {rules: {expenseRules: originalExpenseRules}},
+                },
+            ],
+        };
 
-    API.write(WRITE_COMMANDS.SET_POLICY_CATEGORY_TAX, parameters, onyxData);
+        API.write(WRITE_COMMANDS.SET_POLICY_CATEGORY_TAX, {policyID, ...write}, onyxData);
```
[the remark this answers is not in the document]
Krishna2323: The API is per-category. A compensating write can fail too, and offline queues behind it. Documented; next policy read reconciles.

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1889,207 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
+    const policyID = policy.id;
+    const expenseRules = policy.rules?.expenseRules ?? [];
+    const targets = categoryNames.filter((categoryName) => expenseRules.some((rule) => matchesCategoryTaxRule(rule, categoryName)));
+
+    if (targets.length === 0) {
+        return;
+    }
+
+    // Shared like the save path: the writes are enqueued together, so each needs the same end state.
+    const optimisticExpenseRules = expenseRules.filter((rule) => !targets.some((categoryName) => matchesCategoryTaxRule(rule, categoryName)));
```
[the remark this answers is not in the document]
Krishna2323: Clearing a per-rule flag needs `successData`, which replays stale arrays on reconnect. Fixing the misleading comment instead.

Krishna2323:
> Bug: Created merchant rule disappears after refresh

Can't reproduce this bug

Krishna2323:
> Bug: merchant rule is not optimistically deleted (no ~cross out~) when the linked tax rate is deleted offline

Expected, the BE does not delete the merchant rule when category or tax is deleted.

src/languages/en.ts
Krishna2323: updated.

situchan:
> > Reversed Bug: Category rule is immediately deleted when offline while Merchant rule is optimistically deleted
> 
> Category rules live in a plain array, so Onyx can't mark a single row as deleting. Clearing that flag requires `successData`, which can replay stale arrays on reconnect. So we drop the row instead. Known tradeoff.
> 
> > Bug: category rule is not grayed out when optimistically added offline
> 
> Same
> 
> cc: @JS00001



> > Bug: merchant rule is not optimistically deleted (no ~cross out~) when the linked tax rate is deleted offline
> 
> Expected, the BE does not delete the merchant rule when category or tax is deleted.

ok please update QA Steps to avoid confusion

situchan:
Bug: Offline renaming a category destroys another category rule

Repro step:
1. Go offline
2. Create 2 category rules with the same tax rate selected
3. Rename each category to something else
4. Observe category rules in Expense defaults table

https://github.com/user-attachments/assets/36cf9e1f-2d93-410d-92c2-1f6add900482

JS00001:
> Category rules live in a plain array, so Onyx can't mark a single row as deleting. Clearing that flag requires successData, which can replay stale arrays on reconnect. So we drop the row instead. Known tradeoff.

I agree, we can stick with this for now until we migrate to the new rules format

situchan:
Please review and update all outdated test cases.

e.g.

> Select a category → Save: verify the category shows, Merchant locks, Reset appears, all defaults except Tax lock, apply-to-existing is locked/off, and Preview matches is hidden.

Merchant, tax lock, preview matches no longer exist

Krishna2323:
> Bug: Offline renaming a category destroys another category rule

Fixed. Also updated QA Steps


https://github.com/user-attachments/assets/0f83c4e2-1069-4b96-b185-1674290991ec

src/libs/actions/Policy/Category.ts
```diff
@@ -955,16 +954,22 @@ function renamePolicyCategory(policyData: PolicyData, policyCategory: {oldName:
...
+    if (expenseRuleIndex !== -1 && ruleToRename) {
+        updatedExpenseRules[expenseRuleIndex] = {
+            ...ruleToRename,
+            applyWhen: ruleToRename.applyWhen.map((applyWhen) => ({
+                ...applyWhen,
+                ...(applyWhen.field === CONST.POLICY.FIELDS.CATEGORY &&
+                    applyWhen.value === policyCategory.oldName && {
+                        value: policyCategory.newName,
+                    }),
+            })),
```
situchan, line 971: A failed category rename leaves the tax rule pointing at a name that no longer exists

restore `expenseRules` alongside `approvalRules`, the way the optimistic block sets both

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
```diff
@@ -401,6 +402,7 @@ function PolicyRulesPageRevamp({route}: PolicyRulesPageRevampProps) {
                                 canWriteRules={canWriteRules}
                                 isAgentsRulesBannerDismissed={isAgentsRulesBannerDismissed}
                                 onOpenAgentsTab={() => handleTabPress(RULES_TAB.AGENTS)}
+                                headerComponent={rulesTabSelector}
```
situchan, line 406: Bug: Rules tab bar shrinks to 720px on the General tab so Agents tab is cut off

https://github.com/user-attachments/assets/6c6eec8b-fe8f-487d-9ddf-c30cc10b9923

situchan:
Bug: Editing a merchant rule whose tax rate was deleted shows the raw tax ID

Screenshot 2026-09-02 at 10 09 59 PM

https://github.com/user-attachments/assets/4f99de27-11fd-4bbd-908f-e0f22257295a

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1894,215 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
+        }
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        return {...rule, tax: {field_id_TAX: {...rule.tax?.field_id_TAX, externalID: taxID}}};
+    });
+
+    const added = [...taxRatesByCategory.entries()].flatMap(([categoryName, taxID]) =>
+        !taxID || expenseRules.some((rule) => matchesCategoryTaxRule(rule, categoryName)) ? [] : [buildCategoryTaxRule(categoryName, taxID)],
+    );
+
+    return [...updated, ...added];
```
situchan, line 1945: `matchesCategoryTaxRule(rule, name)` is exactly `getRuleCategoryName(rule) === name`, so searching a map's keys for the one that matches is the same as reading the rule's own category out of the map:

```
const categoryName = [...taxRatesByCategory.keys()].find((name) => matchesCategoryTaxRule(rule, name));
const taxID = categoryName ? taxRatesByCategory.get(categoryName) : undefined;
if (!categoryName || !taxID) { return rule; }
```
becomes:
```
const categoryName = getRuleCategoryName(rule);
const taxID = categoryName ? taxRatesByCategory.get(categoryName) : undefined;
if (!taxID) { return rule; }

```

This drops a key-array rebuild per rule and a redundant guard.

The same shape repeats four more times:

- [added](https://github.com/Expensify/App/pull/99420/changes#diff-fe9bfd658e57c4e63236b195c38af47ab024fc4428a9264c931adcd60c760c19R1936)
- [targets](https://github.com/Expensify/App/pull/99420/changes#diff-fe9bfd658e57c4e63236b195c38af47ab024fc4428a9264c931adcd60c760c19R2017)
- [optimisticExpenseRules](https://github.com/Expensify/App/pull/99420/changes#diff-fe9bfd658e57c4e63236b195c38af47ab024fc4428a9264c931adcd60c760c19R2024)

(each an `Array.some` inside a `filter`/`flatMap`)

One new `Set(expenseRules.map(getRuleCategoryName))` (or `new Set(targets)`) makes all of them single-pass, and shorter.

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1894,215 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
     API.write(WRITE_COMMANDS.SET_POLICY_CATEGORY_APPROVER, parameters, onyxData);
 }
 
-function setPolicyCategoryTax(policy: OnyxEntry<Policy>, categoryName: string, taxID: string) {
-    if (!policy?.id) {
+/** Whether an expense rule is this category's tax default. Shares `getRuleCategoryName` so writes and the table agree. */
+function matchesCategoryTaxRule(rule: ExpenseRule, categoryName: string): boolean {
+    return getRuleCategoryName(rule) === categoryName;
+}
```
situchan: Maybe better to define this in `CategoryTaxRulesUtils`?
Similar to `getRuleCategoryName` and the near-identical `getCategoryTaxRule`

src/pages/workspace/rules/MerchantRules/ExpenseDefaultTypePage.tsx
```diff
@@ -0,0 +1,131 @@
...
+                            />
+                        ))}
+                    </View>
+                </ScrollView>
+            </ScreenWrapper>
+        </AccessOrNotFoundWrapper>
+    );
+}
+
+ExpenseDefaultTypePage.displayName = 'ExpenseDefaultTypePage';
```
situchan: We don't need `displayName` in function components if it's exactly same as component name.

src/pages/workspace/rules/MerchantRules/EditCategoryTaxRulePage.tsx
```diff
@@ -0,0 +1,25 @@
...
+            titleKey="workspace.rules.merchantRules.editRuleTitle"
+            testID="EditCategoryTaxRulePage"
+        />
+    );
+}
+
+EditCategoryTaxRulePage.displayName = 'EditCategoryTaxRulePage';
```
situchan: Same here

src/pages/workspace/rules/MerchantRules/ExpenseDefaultTypePage.tsx
```diff
@@ -0,0 +1,131 @@
...
+            icon: illustrations.FolderOpen,
+            title: translate('workspace.rules.expenseDefaultType.category'),
+            description: translate('workspace.rules.expenseDefaultType.categoryDescription'),
+            onPress: areTaxesEnabled ? () => openEditorScopedTo('category') : showTurnOnTaxesFirstExplainer,
+            isLocked: !areTaxesEnabled,
+        },
+    ];
+
+    return (
+        <AccessOrNotFoundWrapper
```
situchan, line 90: This render block is much overlapped with `RulesNewPage`.

- same `AccessOrNotFoundWrapper` props
- same `ScreenWrapper`/`HeaderWithBackButton`/`ScrollView`
- same subtitle `<Text>`
- same `<View style={styles.mh5}>`
- same `MenuItem` with `displayInDefaultIconColor`
- `iconSizeExtraLarge` twice and `styles.rulesNewMenuItem`

Only `testID`, two translation keys and `iconRight` differ.

A shared `RuleChooserPage`/option-list component would collapse both.

src/types/form/MerchantRuleForm.ts
```diff
@@ -24,11 +26,15 @@ type MerchantRuleForm = Form<
...
         [INPUT_IDS.CATEGORY]: string;
+        [INPUT_IDS.CATEGORIES_TO_MATCH]: string[];
         [INPUT_IDS.DESCRIPTION]: string;
         [INPUT_IDS.MATCH_TYPE]: ValueOf<typeof CONST.SEARCH.SYNTAX_OPERATORS>;
         [INPUT_IDS.MERCHANT]: string;
         [INPUT_IDS.MERCHANT_TO_MATCH]: string;
         [INPUT_IDS.REIMBURSABLE]: boolean;
+        /** Which kind of expense default this is, chosen before the editor opened. Kept in the draft so it survives a
+         * trip to any picker, rather than having to ride on every picker's route. */
+        [INPUT_IDS.RULE_TYPE]: 'merchant' | 'category';
```
situchan: `'merchant' | 'category'` is a raw literal in many places

re-declared inline at:
- here
- MerchantRulePageBase:276
- ExpenseDefaultTypePage:54

then compared against bare strings at:
- MerchantRulePageBase:286
- AddTaxPage:32
- ExpenseDefaultTypePage:74-82

We can apply same as [CONST.POLICY.EXPENSE_DEFAULTS_SECTION](https://github.com/Expensify/App/pull/99420/changes#diff-ea78c710861a60147d2f3000e07ce41004df40cf732560c51cb1e714080ed69eR4180-R4184)

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
```diff
@@ -231,15 +266,52 @@ function MerchantRulePageBase({policyID, ruleID, initialCategoryName, titleKey,
...
-        const tax = policy.taxRates.taxes[form.tax];
-        return tax ? `${tax.name} (${tax.value})` : undefined;
+        if (isEditing) {
+            return 'merchant';
+        }
+        return form?.ruleType;
+    };
+    const scopedRuleType = getScopedRuleType();
+    const isScopedToCategory = scopedRuleType === 'category';
+    const isCategoryRule = hasCategoryCondition || isEditingCategoryTaxRule || isScopedToCategory;
```
situchan: **Two shared predicates, each computed differently per file**

This `isCategoryRule` logic is different from [AddTaxPage.tsx:32](https://github.com/Expensify/App/blob/6c99aa5308efb03c820f619fff4199aa92230340/src/pages/workspace/rules/MerchantRules/AddTaxPage.tsx#L32) (`!!categoryName || form?.ruleType === 'category' || !!form?.categoriesToMatch?.length`).
Same three disjuncts, different order, different sources, maintained separately in two files that have to agree.

"Are taxes usable" is worse — three copies, two of which are identical and one of which isn't:

| | check |
|---|---|
| [MerchantRulePageBase.tsx:254-259](https://github.com/Expensify/App/blob/6c99aa5308efb03c820f619fff4199aa92230340/src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx#L254-L259) | `trackingEnabled && Object.keys(taxes).length > 0` |
| [ExpenseDefaultTypePage.tsx:50](https://github.com/Expensify/App/blob/6c99aa5308efb03c820f619fff4199aa92230340/src/pages/workspace/rules/MerchantRules/ExpenseDefaultTypePage.tsx#L50) | `trackingEnabled && Object.keys(taxes).length > 0` |
| [AddTaxPage.tsx:29](https://github.com/Expensify/App/blob/6c99aa5308efb03c820f619fff4199aa92230340/src/pages/workspace/rules/MerchantRules/AddTaxPage.tsx#L29) | `trackingEnabled` only |

Both belong in `CategoryTaxRulesUtils` next to the other rule helpers.

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
```diff
@@ -231,15 +266,52 @@ function MerchantRulePageBase({policyID, ruleID, initialCategoryName, titleKey,
...
+        setDraftMerchantRule(scopedRuleType ? {ruleType: scopedRuleType} : {});
+        setShouldShowError(false);
+        setShouldUpdateMatchingTransactions(false);
     };
 
+    // One rule per category is saved, so the condition row lists every category the admin picked.
+    const categoriesToMatchDisplayName = hasCategoryCondition ? categoriesToMatch.map(getDecodedCategoryName).join(', ') : undefined;
+    const categoryDisplayName = form?.category ? getDecodedCategoryName(form.category) : undefined;
+    // Stay undefined when empty so the row reads as unset rather than blank.
+    const taxDisplayName = () => getTaxRateDisplayName(policy, isCategoryRule ? categoryTaxID : form?.tax) || undefined;
```
situchan: `taxDisplayName` is now a zero-arg function wrapping a single expression, called once at [:539](https://github.com/Expensify/App/blob/6c99aa5308efb03c820f619fff4199aa92230340/src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx#L539).
It was a function when it had early returns; it should be a plain `const` now.

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
```diff
@@ -231,15 +266,52 @@ function MerchantRulePageBase({policyID, ruleID, initialCategoryName, titleKey,
...
+            return 'category';
         }
-        const tax = policy.taxRates.taxes[form.tax];
-        return tax ? `${tax.name} (${tax.value})` : undefined;
+        if (isEditing) {
+            return 'merchant';
+        }
+        return form?.ruleType;
+    };
+    const scopedRuleType = getScopedRuleType();
```
situchan: Same here

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
```diff
@@ -444,7 +574,8 @@ function MerchantRulePageBase({policyID, ruleID, initialCategoryName, titleKey,
                           icon: getItemIcon(icons.Paycheck),
                       }
                     : undefined,
-            ],
+                // Tax is the only default a category rule can set, so the other rows are dropped rather than shown.
+            ].filter((item) => !isCategoryRule || !item || item.key === 'tax'),
```
situchan: Remove `!item`. Instead use optional chaining `item?.key`

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1894,215 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
+                    value: {rules: {expenseRules: optimisticExpenseRules}},
+                },
+            ],
+            failureData: [
+                {
+                    onyxMethod: Onyx.METHOD.MERGE,
+                    key: `${ONYXKEYS.COLLECTION.POLICY}${policyID}`,
+                    value: {rules: {expenseRules: originalExpenseRules}},
+                },
+            ],
```
situchan: Move `onyxData` out of the `for` loop.

src/libs/actions/Policy/Category.ts
Krishna2323: Fixed. `expenseRules` now restored alongside `approvalRules`.

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
Krishna2323: Fixed.

Krishna2323:
> Bug: Editing a merchant rule whose tax rate was deleted shows the raw tax ID

Fixed, it will show empty row for tax in that case.

src/libs/actions/Policy/Category.ts
Krishna2323: Done, as suggested. All five now read the rule's own category or share one Set, and the redundant guard is gone.

src/libs/actions/Policy/Category.ts
Krishna2323: Moved, next to `getRuleCategoryName`.

src/pages/workspace/rules/MerchantRules/ExpenseDefaultTypePage.tsx
Krishna2323: Agreed on the overlap, but I’d defer it for now. `RulesNewPage` has five beta-gated options vs. two with a lock state, so the shared part is pretty thin once parameterized.

src/types/form/MerchantRuleForm.ts
Krishna2323: Done. Added `CONST.POLICY.EXPENSE_DEFAULT_RULE_TYPE` plus an `ExpenseDefaultRuleType` type, all six comparison sites go through it.

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
Krishna2323: Both moved to `CategoryTaxRulesUtils`.

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
Krishna2323: Done. It had early returns again from the deleted-rate fix, so it's now two consts and an expression.

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
Krishna2323: Literals now go through `CONST.POLICY.EXPENSE_DEFAULT_RULE_TYPE`. Left it a function though, has three branches, and no-nested-ternary rejects the const version.

src/pages/workspace/rules/MerchantRules/MerchantRulePageBase.tsx
Krishna2323: Done. Undefined entries were already dropped by the type guard before render.

src/libs/actions/Policy/Category.ts
Krishna2323: Done.

situchan:
While it's expected to hide default tax rate in selection list, it's a bit confusing for the users:

1. Workspace with Taxes on and exactly one rate, which is the workspace default.
2. Rules → New rule → Apply expense defaults → Category
3. Pick any category, then tap Tax.
4. The picker is empty - no rates, no None. Back out and tap Save: "Please select a tax rate", with nothing selectable.


https://github.com/user-attachments/assets/2f45d360-9f12-49b7-840d-2c2493889b10

situchan:
Please make sure **Delete** steps are correct.

e.g. these steps failing and not make sense:

> Delete a category rule: verify it disappears. Delete rule should be absent when the workspace has no default tax rate.
> Delete the category set by a merchant rule: verify the merchant rule stays and only that default is dropped.


https://github.com/user-attachments/assets/f5874ade-97fc-48d1-8691-e07e280266cb


https://github.com/user-attachments/assets/805755aa-9fb4-441a-b813-617b2ae3a673

situchan (approved):
Looks good other than test steps above

Krishna2323:
Test steps updated.

JS00001:
This is my next PR to review

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1895,224 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
+                value: {rules: {expenseRules}},
+            },
+        ],
+    };
 
-        if (expenseRule && indexToUpdate !== -1) {
-            expenseRule.tax.field_id_TAX.externalID = taxID;
-        }
+    for (const categoryName of categoryNames) {
+        API.write(WRITE_COMMANDS.SET_POLICY_CATEGORY_TAX, {policyID, categoryName, taxID}, onyxData);
```
JS00001, line 1993: should we do this independently, where we remove the one entry from the `rules.expenseRules` if it fails? It feels strange to remove all changes when only one API Req failed

src/libs/actions/Policy/Category.ts
```diff
@@ -1889,72 +1895,224 @@ function setPolicyCategoryApprover(policyID: string, categoryName: string, appro
...
+                value: {rules: {expenseRules}},
             },
         ],
     };
 
-    const parameters: SetPolicyCategoryTaxParams = {
-        policyID,
-        categoryName,
-        taxID,
+    for (const categoryName of targets) {
```
JS00001, line 2053: Same here, should we be reverting all changes to expenseRules when only one request fails?

src/libs/RequireFieldsRulesUtils.ts
```diff
@@ -672,7 +672,9 @@ function createRequireFieldsTableItem({
         typeLabel,
         conditionText,
         ruleDescription,
-        searchTokens: [decodedCategoryName, ruleDescription, typeLabel, ...ruleDescriptions],
+        // The Type column reads "Require" now, but the rules are still called field requirements everywhere else, so
+        // the old name stays searchable rather than making admins learn the column's shorthand to find a row.
+        searchTokens: [decodedCategoryName, ruleDescription, typeLabel, translate('workspace.rules.requireFieldsRule.title'), ...ruleDescriptions],
```
JS00001, line 677: to clarify, this is the value in the table, right? The new translation here?

src/pages/workspace/rules/MerchantRules/AddCategoryToMatchPage.tsx
```diff
@@ -0,0 +1,204 @@
...
+    });
+
+    const categoryItems: CategoryListItem[] = categories.map((category) => ({
+        keyForList: category.name,
+        text: getDecodedCategoryName(category.name),
+        value: category.name,
+        isSelected: selectedCategories.includes(category.name),
+    }));
+
+    const filterCategory = (item: CategoryListItem, searchInput: string) => (item.text ?? '').toLowerCase().includes(searchInput.toLowerCase());
```
JS00001: should we use a tokenized search?

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
```diff
@@ -384,15 +386,14 @@ function PolicyRulesPageRevamp({route}: PolicyRulesPageRevampProps) {
...
-                    {shouldDisplayButtonsInSeparateLine && !!headerButtons && <View style={[styles.flexShrink0, styles.pl5, styles.pr5, styles.pb5, styles.w100]}>{headerButtons}</View>}
+                    {separateLineButtons}
                     <View
                         style={[
                             styles.flex1,
                             styles.mnh0,
                             styles.w100,
                             shouldUseNarrowLayout ? styles.workspaceSectionMobile : styles.workspaceSection,
-                            (isTableTab || isAgentsTab) && styles.mw100,
+                            (isTableTab || isAgentsTab || activeTab === RULES_TAB.GENERAL) && styles.mw100,
```
JS00001: NAB we could use `isGeneralTab` here to match the other two

src/pages/workspace/rules/tabs/RulesGeneralTab.tsx
```diff
@@ -11,40 +13,54 @@ import IndividualExpenseRulesSectionRevamp from '@pages/workspace/rules/Individu
...
+import {View} from 'react-native';
 
 type RulesGeneralTabProps = {
     policyID: string;
     canWriteRules: boolean;
     isAgentsRulesBannerDismissed: boolean;
     /** Opens the Agents tab through the page's tab handler, so Collect gets the Control upgrade page instead. */
     onOpenAgentsTab: () => void;
+    /** The tab selector, rendered inside this tab's scroll like the table tabs render it inside their list. */
+    headerComponent?: React.ReactElement;
```
JS00001: nab why is this optional? We always need a header, right?

JS00001 (dismissed):
Bunch of NABs/questions, lmk if you think any should be changed

src/libs/actions/Policy/Category.ts
Krishna2323: `failureData` is fixed before the request goes out, and Onyx replaces arrays wholesale. Per-entry reverts can contradict each other: whichever failure lands last wins, so a rejected category can end up looking saved. One baseline makes the order irrelevant. The real fix is to key `expenseRules` like `codingRules` so each rule can carry its own pending state.

src/libs/actions/Policy/Category.ts
Krishna2323: > `failureData` is fixed before the request goes out, and Onyx replaces arrays wholesale. Per-entry reverts can contradict each other: whichever failure lands last wins, so a rejected category can end up looking saved. One baseline makes the order irrelevant. The real fix is to key `expenseRules` like `codingRules` so each rule can carry its own pending state.

same

src/pages/workspace/rules/PolicyRulesPageRevamp.tsx
Krishna2323: done.

src/pages/workspace/rules/tabs/RulesGeneralTab.tsx
Krishna2323: Yes, correct. Made it required.

src/pages/workspace/rules/MerchantRules/AddCategoryToMatchPage.tsx
Krishna2323: yes, updated.

src/libs/RequireFieldsRulesUtils.ts
Krishna2323: No, the other way around. `typeLabel` is the table value, the “Require” badge in the Type column, and it was already a token. The new one is “Field requirements”.

JS00001 (approved):

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100504 was identified to be related to this PR.

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100529 was identified to be related to this PR.

mitarachim:
Deploy Blocker https://github.com/Expensify/App/issues/100530 was identified to be related to this PR.

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100532 was identified to be related to this PR.

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100536 was identified to be related to this PR.

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100544 was identified to be related to this PR.

lanitochka17:
Deploy Blocker https://github.com/Expensify/App/issues/100545 was identified to be related to this PR.

jponikarchuk:
Deploy Blocker https://github.com/Expensify/App/issues/100546 was identified to be related to this PR.

nlemma:
Deploy Blocker https://github.com/Expensify/App/issues/100591 was identified to be related to this PR.

